PR #23, opened 2020-05-04
chore: test on node 14

### LINKED ISSUE #22: Build: Use node 14 instead of node 13
Similar to https://github.com/open-telemetry/opentelemetry-js/issues/996

/cc @dyladan

dyladan:
Fixes #22


dyladan:
hmm @mayurkale22 wdyt? AFAIK 13 essentially is 14, but with the API not completely finalized yet. Obviously there are differences though since our tests are passing in 13 but not 14.

mayurkale22:
> hmm @mayurkale22 wdyt? AFAIK 13 essentially is 14, but with the API not completely finalized yet. Obviously there are differences though since our tests are passing in 13 but not 14.

Maybe we can keep both and remove 13 after `2020-06-01`?

dyladan:
I will update the PR, and will make a PR to have same behavior in core. We still need to figure out why pg fails on 14

obecny:
@dyladan can you please resolve conflicts ?

dyladan:
> @dyladan can you please resolve conflicts ?

I'll do this today

Issue #20, opened 2020-04-28
create a plugins-all package


andrewhsu:
**NB:** Before opening a feature request against this repo, consider whether the feature should/could be implemented in the [other OpenTelemetry client libraries](https://github.com/open-telemetry/). If so, please [open an issue on opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/issues/new) first.

**Is your feature request related to a problem? Please describe.**
Would like an easy way to install plugins that would help to autoinstrument my app that would be sending opentelemetry traces.

**Describe the solution you'd like**
A plugins-all package that would instal the kitchen sink from opentelemetry-js-contrib.

**Describe alternatives you've considered**
Possible alternatives:
1. Manual instructions for the user to find the package and install themselves.
2. Utility tool to detect what npm package is already in the system and echo to the user a suggested npm command to run to install plugins that would help to auto-instrument.

**Additional context**
Talked about on 2020-04-22 [agenda](https://docs.google.com/document/d/1S_pN7sBU7wO6cBP2hDSAt3bjSWdWdm6jWeUikC9KrcY/edit#heading=h.lsz268gua808)

dyladan:
Per sig, I will create it with name `@opentelemetry/plugins-all-node`

vmarchaud:
This has been implemented since then (https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/plugins-node-core-and-contrib) so i'm closing this

PR #25, opened 2020-05-06
feat: add @opentelemetry/plugins-all-node-contrib

dyladan:
This is related to #20

### DIFF
--- a/metapackages/plugins-all-node-contrib/package.json
+++ b/metapackages/plugins-all-node-contrib/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@opentelemetry/plugins-all-node-contrib",
+  "version": "0.7.0",
+  "description": "Metapackage which bundles all opentelemetry contrib plugins",
+  "author": "OpenTelemetry Authors",
+  "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib#readme",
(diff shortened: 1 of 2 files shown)


metapackages/plugins-all-node-contrib/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/open-telemetry/opentelemetry-js-contrib.git"
+  },
+  "bugs": {
+    "url": "https://github.com/open-telemetry/opentelemetry-js-contrib/issues"
+  },
+  "dependencies": {
+    "@opentelemetry/plugin-dns": "^0.7.0",
```
dyladan, line 17: Remove for now until context based tracing disable is implemented

dyladan:
Add a README

andrewhsu:
@dyladan thanks for this! as discussed in the SIG mtg today, adding a README or documentation with this to reserve the right to update it to make sure stuff is added (or not added if it causes problems) would be good.

dyladan:
Closing in favor of only having `plugins-core` and `plugins-all`. The use case to only install contrib plugins probably is not that valuable and increases maintenance burden.

PR #33, opened 2020-05-15
feat(opentelemetry-plugin-ioredis): provide a custom serializer fn for db.statement

marcoreni:
IORedis plugin currently serializes the command name and arguments in the `db.statement` tag. This may yield to a big payload serialized inside the tag (for example, while `mget`ting many values or `set`ting large json objects). It may also be an issue when data saved inside Redis should not be accessible inside logs/traces. 

This PR adds an optional function inside the configuration that can be used to define a custom serialization function for `db.statement`. The function will receive cmdName and cmdArgs and will return a serialized string.
With this, one may decide (for example) to serialize only the `cmdName` for specific commands, or to have a separate handling for each command.

First contribution to the project so I hope I have done everything right :D

### DIFF
--- a/plugins/node/opentelemetry-plugin-ioredis/src/types.ts
+++ b/plugins/node/opentelemetry-plugin-ioredis/src/types.ts
@@ -31,6 +31,13 @@ export interface IORedisPluginClientTypes {
   options: ioredisTypes.RedisOptions;
 }
 
+/**
+ * Function that can be used to serialize db.statement tag
+ * @param cmdName - The name of the command (eg. set, get, mset)
+ * @param cmdArgs - Array of arguments passed to the command
+ *
+ * @returns serialized string that will be used as the db.statement attribute.
+ */
 export type DbStatementSerializer = (
   cmdName: IORedisCommand['name'],
   cmdArgs: IORedisCommand['args']
--- a/plugins/node/opentelemetry-plugin-ioredis/src/types.ts
+++ b/plugins/node/opentelemetry-plugin-ioredis/src/types.ts
@@ -15,6 +15,7 @@
  */
 
 import * as ioredisTypes from 'ioredis';
+import { PluginConfig } from '@opentelemetry/api';
 
 export interface IORedisCommand {
   reject: (err: Error) => void;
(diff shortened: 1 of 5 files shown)


plugins/node/opentelemetry-plugin-ioredis/test/ioredis.test.ts
```diff
@@ -511,6 +512,52 @@ describe('ioredis', () => {
       });
     });
 
+    describe('Instrumenting with specific configuration', () => {
```
dyladan, line 515: "specific configuration" isn't particularly descriptive. Maybe just `describe("Custom Serializer", () => {`

marcoreni: Done

plugins/node/opentelemetry-plugin-ioredis/src/types.ts
```diff
@@ -29,3 +30,16 @@ export interface IORedisPluginClientTypes {
   // https://github.com/luin/ioredis/blob/master/API.md
   options: ioredisTypes.RedisOptions;
 }
+
+export type DbStatementSerializer = (
```
obecny, line 34: pls add tsdoc

plugins/node/opentelemetry-plugin-ioredis/src/types.ts
```diff
@@ -29,3 +30,16 @@ export interface IORedisPluginClientTypes {
...
+  cmdName: IORedisCommand['name'],
+  cmdArgs: IORedisCommand['args']
+) => string;
+
+/**
+ * Options available for the IORedis Plugin (see [documentation](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-plugin-ioredis#ioredis-plugin-options))
+ */
+export interface IORedisPluginConfig extends PluginConfig {
+  /** Custom serializer function for the db.statement tag */
+  dbStatementSerializer?: DbStatementSerializer;
```
obecny, line 44: please add some example to documentation so it is easy to understand how this should look like without necessity of looking into source code

plugins/node/opentelemetry-plugin-ioredis/src/utils.ts
```diff
@@ -60,21 +65,35 @@ export const traceConnection = (tracer: Tracer, original: Function) => {
...
+  tracer: Tracer,
+  original: Function,
+  config?: IORedisPluginConfig
+) => {
   return function(
     this: ioredisTypes.Redis & IORedisPluginClientTypes,
     cmd?: IORedisCommand
   ) {
     if (arguments.length >= 1 && typeof cmd === 'object') {
+      const dbStatementSerializer =
```
obecny, line 86: I think this assignment for `dbStatementSerializer` should be done one level up (between line 80 and 81) so that it is assigned once only during the `patch` function

dyladan: Ah yes I like this optimization

marcoreni: Done

plugins/node/opentelemetry-plugin-ioredis/src/types.ts
marcoreni: Done. Let me know if it's enough

plugins/node/opentelemetry-plugin-ioredis/src/types.ts
marcoreni: Done

dyladan:
@naseemkullah any way to get your approval on this since you contributed the ioredis plugin?

naseemkullah:
LGTM 👍

dyladan:
@mayurkale22 this is ready to merge

PR #34, opened 2020-05-15
Enhanced Database Reporting for MongoDB

romil-punetha:

- Fixes [#1044](https://github.com/open-telemetry/opentelemetry-js/issues/1044)

- The `enhancedDatabaseReporting` variable is used to allow capturing entire database queries instead of obfuscating query parameteres


dyladan:
I renamed the PR because "0.7.1 proposal" doesn't really make sense.

dyladan:
Why did you change the image used for mongo in tests?

romil-punetha:
> Why did you change the image used for mongo in tests?

Have reverted that. By mistake.

examples/mongodb/client.js
```diff
@@ -0,0 +1,73 @@
...
+function makeRequest() {
+  // span corresponds to outgoing requests. Here, we have manually created
+  // the span, which is created to track work that happens outside of the
+  // request lifecycle entirely.
+  const span = tracer.startSpan('makeRequest');
+
+  let queries = 0;
+  let responses = 0;
+
+  tracer.withSpan(span, () => {
```
dyladan, line 17: Both of these calls can be done in the same `withSpan` call.

plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
```diff
@@ -192,13 +192,17 @@ export class MongoDBPlugin extends BasePlugin<typeof mongodb> {
...
     if (command === undefined) return;
-    const query = Object.keys(command.query ?? command.q ?? command).reduce(
-      (obj, key) => {
-        obj[key] = '?';
-        return obj;
-      },
-      {} as { [key: string]: string }
-    );
+
+    const query = this._config?.enhancedDatabaseReporting
```
dyladan, line 196: I think we should add a comment to explain why this is done

romil-punetha:
> Please add some documentation in the README that makes it clear what the enhanced db reporting flag does, and how to enable it.
> 
> I think the example can be drastically simplified. It doesn't need a client and server component, or to make http calls. A script that simply sets up the tracer and makes a call to mongodb should be enough to demonstrate the plugin.

I used the same format as mysql plugin.

plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
romil-punetha: done

dyladan:
> > Please add some documentation in the README that makes it clear what the enhanced db reporting flag does, and how to enable it.
> > I think the example can be drastically simplified. It doesn't need a client and server component, or to make http calls. A script that simply sets up the tracer and makes a call to mongodb should be enough to demonstrate the plugin.
> 
> I used the same format as mysql plugin.

Yea I see that. We've been making an effort to simplify the examples recently though as we have too many that show too many things. Smaller examples are easier for beginners to understand.

romil-punetha:
> i think we should document the option inside the readme of the plugin (like done with the http one [there](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-plugin-http#http-plugin-options))

I dont see a folder for mongo plugin there.

vmarchaud:
@romil-punetha I was refering to the documentation made **for** the http plugin options that you could re-use to document the `enhancedDatabaseReporting` inside the mongodb plugin's readme :)

romil-punetha:
Oh, ok.

examples/mongodb/README.md
```diff
@@ -0,0 +1,88 @@
+# Overview
+
+OpenTelemetry Mongodb Instrumentation allows the user to automatically collect trace data and export them to the backend of choice (we can use Zipkin or Jaeger for this example), to give observability to distributed systems.
+
+This is a modification of the HTTP example that executes multiple parallel requests that interact with a Mongodb server backend using the `mongo` npm module. The example displays traces using multiple connection methods.
+- Create Collection Query
+- Insert Document Query
+- Fetch All Documents Query
+
+### Http Plugin Options
```
vmarchaud, line 10: This should be inside the plugin readme ([there](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/master/plugins/node/opentelemetry-plugin-mongodb/README.md)), not the example. Also the option of the plugin only contains `enhancedDatabaseReporting` with your PR, why did you add the others ?

romil-punetha: Moving this. 
I thought it would be good to have all available options documented in one place, and because the plugin takes all of these as input.

romil-punetha: @vmarchaud should i only add `enhancedDatabaseReporting` ?

vmarchaud: @romil-punetha yeah, we generally document only the custom options.

romil-punetha: Ok. Editing that.

romil-punetha:
> lgtm, as @dyladan said the example could be simpler but thats not a blocker for me as generally user want examples that looks close to their use-cases which would be in this case a http server that make operation against mongodb.

Agreed. The same worked for me as well. The client-server example contains the minimum changes required instrument a server.

plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
```diff
@@ -192,13 +192,18 @@ export class MongoDBPlugin extends BasePlugin<typeof mongodb> {
...
-      (obj, key) => {
-        obj[key] = '?';
-        return obj;
-      },
-      {} as { [key: string]: string }
-    );
+
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
+    const query = this._config?.enhancedDatabaseReporting
+      ? command.query ?? command.q ?? command
```
obecny, line 198: Nit: line 198 and 199 has the same maybe consider this to put into some variable and then reuse ?
`const commandObj = command.query ?? command.q ?? command;`

plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
```diff
@@ -192,13 +192,18 @@ export class MongoDBPlugin extends BasePlugin<typeof mongodb> {
...
-    const query = Object.keys(command.query ?? command.q ?? command).reduce(
-      (obj, key) => {
-        obj[key] = '?';
-        return obj;
-      },
-      {} as { [key: string]: string }
-    );
+
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
+    const query = this._config?.enhancedDatabaseReporting
```
obecny, line 197: can you compare this directly with boolean expression
`this._config?.enhancedDatabaseReporting === true`
simply to avoid situation like this `enhancedDatabaseReporting = 'true'` or `enhancedDatabaseReporting='foo'` etc.

plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
romil-punetha: done

plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
romil-punetha: fixed this.

PR #42, opened 2020-05-30
improv(ioredis): require parent span

vmarchaud:
I've also fixed a missing dep in `test-utils` which was making my local install fail.

### DIFF
--- a/plugins/node/opentelemetry-plugin-ioredis/package.json
+++ b/plugins/node/opentelemetry-plugin-ioredis/package.json
@@ -48,7 +48,7 @@
     "@opentelemetry/node": "0.9.0",
     "@opentelemetry/test-utils": "^0.8.0",
     "@opentelemetry/tracing": "0.9.0",
-    "@types/ioredis": "4.16.6",
+    "@types/ioredis": "4.16.7",
     "@types/mocha": "7.0.2",
     "@types/node": "13.13.12",
     "@types/shimmer": "1.0.1",
--- a/plugins/node/opentelemetry-plugin-ioredis/src/utils.ts
+++ b/plugins/node/opentelemetry-plugin-ioredis/src/utils.ts
@@ -84,35 +84,41 @@ export const traceSendCommand = (
     this: ioredisTypes.Redis & IORedisPluginClientTypes,
     cmd?: IORedisCommand
   ) {
-    if (arguments.length >= 1 && typeof cmd === 'object') {
-      const span = tracer.startSpan(cmd.name, {
-        kind: SpanKind.CLIENT,
-        attributes: {
-          [AttributeNames.COMPONENT]: IORedisPlugin.COMPONENT,
-          [AttributeNames.DB_TYPE]: IORedisPlugin.DB_TYPE,
-          [AttributeNames.DB_STATEMENT]: dbStatementSerializer(
-            cmd.name,
-            cmd.args
-          ),
-        },
-      });
+    if (arguments.length < 1 || typeof cmd !== 'object') {
+      return original.apply(this, arguments);
+    }
+    // Do not trace if there is not parent span
+    if (tracer.getCurrentSpan() === undefined) {
+      return original.apply(this, arguments);
+    }
 
-      const { host, port } = this.options;
+    const span = tracer.startSpan(cmd.name, {
+      kind: SpanKind.CLIENT,
+      attributes: {
+        [AttributeNames.COMPONENT]: IORedisPlugin.COMPONENT,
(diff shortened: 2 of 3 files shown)


vmarchaud:
cc @naseemkullah since you wrote the plugin

naseemkullah:
Would this prevent background queue processing tasks from emitting spans? Not that that's necessarily bad

vmarchaud:
> Would this prevent background queue processing tasks from emitting spans? Not that that's necessarily bad

Well yes but if there aren't created inside a trace it doesn't make sense to trace these i believe.
If you want to trace a background execution, you would need to create a span and then launch your task inside a `withSpan`.

naseemkullah:
LGTM

plugins/node/opentelemetry-plugin-ioredis/src/utils.ts
```diff
@@ -84,35 +84,40 @@ export const traceSendCommand = (
...
-          [AttributeNames.DB_STATEMENT]: dbStatementSerializer(
-            cmd.name,
-            cmd.args
-          ),
-        },
-      });
+    if (arguments.length < 1 || typeof cmd !== 'object') {
+      return original.apply(this, arguments);
+    }
+    if (tracer.getCurrentSpan() === undefined) {
```
naseemkullah, line 90:
```suggestion
    // Do not trace if no parent span
    if (tracer.getCurrentSpan() === undefined) {
```

Just for ppl like me :)

plugins/node/opentelemetry-plugin-ioredis/src/utils.ts
```diff
@@ -84,35 +84,40 @@ export const traceSendCommand = (
...
-        attributes: {
-          [AttributeNames.COMPONENT]: IORedisPlugin.COMPONENT,
-          [AttributeNames.DB_TYPE]: IORedisPlugin.DB_TYPE,
-          [AttributeNames.DB_STATEMENT]: dbStatementSerializer(
-            cmd.name,
-            cmd.args
-          ),
-        },
-      });
+    if (arguments.length < 1 || typeof cmd !== 'object') {
```
naseemkullah, line 87:
```suggestion
    if (arguments.length < 1 || typeof cmd !== 'object') {
```

When does/could this happen? can we put a comment above e.g.

vmarchaud: Since we monkey patch the runtime we code we can never be sure how the module is used, those check are present in most plugins to prevent us from instrumenting a usage that we didn't anticipated.
I don't think its really worth to explain that since it's inherent of the monkeypatch method

naseemkullah: Ok thanks for explaining, so it's to protect from unanticipated use of the lib, but we may not have a concrete example (hence unanticipated).

plugins/node/opentelemetry-plugin-ioredis/src/utils.ts
```diff
@@ -84,35 +84,41 @@ export const traceSendCommand = (
...
+      return original.apply(this, arguments);
+    }
+    // Do not trace if there is not parent span
```
naseemkullah, line 90:
```suggestion
    // Do not trace if there is no parent span
```

naseemkullah:
@open-telemetry/javascript-approvers bump

@vmarchaud Could you kindly resolve conflicts?

vmarchaud:
I've rebased the PR

naseemkullah:
@open-telemetry/javascript-maintainers would it be possible to include this one in the next cut?

dyladan:
In the next contrib cut sure, but we're going to wait for 0.9.0 of the core first

plugins/node/opentelemetry-plugin-ioredis/package.json
```diff
@@ -48,7 +48,7 @@
     "@opentelemetry/node": "0.8.3",
     "@opentelemetry/test-utils": "^0.8.0",
     "@opentelemetry/tracing": "0.8.3",
-    "@types/ioredis": "4.16.5",
+    "@types/ioredis": "^4.16.6",
```
dyladan, line 51: Now that we are using dependabot, we are pinning versions.
```suggestion
    "@types/ioredis": "4.16.6",
```

naseemkullah:
> In the next contrib cut sure, but we're going to wait for 0.9.0 of the core first

Great, thanks!

vmarchaud:
I've rebased and bumped again the types, i think it good to merge cc @dyladan @mayurkale22

Issue #8, opened 2020-04-07
feat(express): add option to disable span creation without root


vmarchaud:
Similar as https://github.com/open-telemetry/opentelemetry-js/issues/931, i've added both the express and http plugin but added some pattern to ignore for the http plugin. However when the http plugin dont create a span because the url is ignored, the express plugin still create, resulting in single trace with one span for each middleware.
Two solutions:
- allow to ignore express layers depending on the url (but seems weird)
- allow to disable new span if there are no parent/root span

WDYT ?

vmarchaud:
Closing since #43 has been merged

PR #43, opened 2020-05-30
improv(express): require parent span #8

vmarchaud:


dyladan:
Looks like you've made this so it's not configurable?

vmarchaud:
@dyladan Yeah i intially wanted to make it an option but i don't see an use case where you want express to create root spans, do you think we should make it just in case ?

markwolff:
> i don't see an use case where you want express to create root spans, do you think we should make it just in case ?

Would it make sense to add warn logging here instead? e.g. we didn't create a span for you here because xyz. Would it get too noisy if someone does have this use case?

dyladan:
At this point our customer base isn't so large that having them create github issues and/or reach out on gitter if they need some use-case isn't a problem. I would say we don't need to log

johanneswuerbach:
We run into a similar issue with a custom [hapi](https://hapi.dev/) plugin we currently develop internally.

Instead of just not generating spans we would actually have found it more convenient to change the http plugin would generate a `NoopSpan` in those cases to not just propagate the parent span creation to the next layer, but actually ignore the entire trace.

Would that also make sense here? Would it be an option to add this to the http(s) plugin?

vmarchaud:
@johanneswuerbach I believe in your case it could be easier to use the API to stop tracing (https://github.com/open-telemetry/opentelemetry-js/issues/1040)

johanneswuerbach:
Shouldn't this be set by the http plugin in this case, so either express or hapi won't record spans?

Our use-case is a health check route, which is now ignored by the http plugin, but still generates a trace, just without the http root span. I guess every express user would have the same issue.

vmarchaud:
@johanneswuerbach Sorry i wasn't clear, you could indeed use the noop span but it would propagate to other service that trace the request too. I believe the use of the system to stop tracing via context would be more appriopriate since from my understanding it would propagate.

vmarchaud:
@dyladan What is the requirement of reviewer for the contrib repo ? This one has been opened for almost 3 weeks too :(

dyladan:
@vmarchaud let's talk about it at SIG today, but I think we can probably merge

vmarchaud:
@dyladan cant make it to this SIG sadly

dyladan:
I am going to try to come up with a contrib-specific merge guideline similar to what we have for core. For now, I'm ok to merge this if we get @mayurkale22 approval.

mayurkale22:
> I am going to try to come up with a contrib-specific merge guideline similar to what we have for core. For now, I'm ok to merge this if we get @mayurkale22 approval.

Sorry for the delay, this one slipped through my inbox.

PR #27, opened 2020-05-06
chore: bump to pg8.*

markwolff:
## Which problem is this PR solving?

- Fixes #24 

## Short description of the changes

- Adds `8.*` to versions matrix. Only test on `pg@8.*`

PR #49, opened 2020-06-09
chore(pg-pool): bump to latest pg versions

markwolff:

- Updates #23, #27

- `pg@7.x` does not work on node14+ and guidance from maintainers is to upgrade to `8.x`


dyladan:
@mayurkale22 think this is ok to merge if you're ok with it

Issue #87, opened 2020-06-17
npm run lint doesn't work


obecny:

rezakrimi:
I see that the the project is using tslint and it's been deprecated. Is it ok to setup eslint similar to the core repo?

dyladan:
@rezakrimi yes that would be great! Would you like to be assigned?

rezakrimi:
Sure

vmarchaud:
I believe we can close this since we moved to eslint and the lint run fine on the CI: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/workflows/lint.yml

PR #110, opened 2020-06-25
Setting up ESLint

rezakrimi:

- issue #87 

- Removing TSLint
- Setting up ESLint similar to the core repo
~- None of the files are modified by the linter to keep the PR small~


rezakrimi:
I think the lint check should install ESLint modules (parser, plugins, etc.) instead of TSLint.

dyladan:
I would look at the github action in the core repo. It may help you get this figured out, and it would be easier to maintain if both repos used the same mechanism.

obecny:
@rezakrimi can you please make sure the build actually passes ?

rezakrimi:
> @rezakrimi can you please make sure the build actually passes ?

I'm planning to update the lint checks over the weekend. However, some of the files have linting issues that are not fixable by the linter, so the check will still fail. How are we going to work around that? Do you think that fixing lint issues should be a separate issue or should it be part of this PR?

dyladan:
Fixing lint issues should be a part of this PR if it is not too many errors. If it is a large number of errors it may be done in a follow up, but the follow up should be done quickly.

rezakrimi:
Should I have to make a new PR to trigger github action? Because it runs when I push to my local repo.
Also, there's a long list of linting issues and I think they should be fixed in a follow up issue.

eslint.config.js
```diff
@@ -0,0 +1,49 @@
+module.exports = {
+  plugins: [
+    "@typescript-eslint",
+    "header"
+  ],
+  extends: [
+      "./node_modules/gts",
```
markwolff, line 7: I wonder if we could extend the eslint config from the main repo so that changes are automatically picked up after a main repo release? wdyt @dyladan?

dyladan: I wonder if there is a way to depend on the file directly in github instead of a release. Very interesting idea though. A package like `@opentelemetry/eslint-config` would be a welcome addition I think.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,29 @@
...
+          key: ${{ runner.os }}-${{ hashFiles('**/package.json') }}
+
+      - name: Bootstrap
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          npm install --only=dev --ignore-scripts
+          npx lerna bootstrap --no-ci --ignore-scripts -- --only=dev
+      - name: Lint
+        run: |
+          npm run lint
```
dyladan, line 29: newline at end of file please

eslint.config.js
markwolff: Yes we could do something like `npm install --save-dev https://github.com/opentelemetry-js/opentelemetry-js/tarball/master`

rezakrimi:
As mentioned [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories): "When you create a pull request from a forked repository to the base repository, GitHub sends the `pull_request` event to the base repository and no pull request events occur on the forked repository."
So I think because the workflow doesn't exist in the base repository, the new github workflow isn't triggered. I tries closing and reopening this PR but it didn't help.
You can check the new lint workflow running on my [branch](https://github.com/tinylox/opentelemetry-js-contrib/tree/lint).
I'll will move on to fixing the linting issues if that looks good.

rezakrimi:
[Here](https://docs.google.com/document/d/1bxrwuxEP6acxRXGe7hT5THJLjYXm3Bt924jZ72C_82s/edit?usp=sharing) is a list of all the linting issues that are not fixable by the linter. Please have a look and comment on how these should be resolved.

packages/opentelemetry-test-utils/testUtils.ts
```diff
@@ -15,13 +15,16 @@
  */
 
 import * as childProcess from 'child_process';
+// eslint-disable-next-line node/no-unpublished-import
```
obecny, line 18: is this really needed ? and then line 21 and 27 ?

rezakrimi: I followed what was suggested on this [doc](https://docs.google.com/document/d/1bxrwuxEP6acxRXGe7hT5THJLjYXm3Bt924jZ72C_82s/edit?usp=sharing). As Daniel mentioned, this could also be resolved by moving these modules from devDependencies to dependencies. What do you think?

plugins/node/opentelemetry-plugin-ioredis/src/types.ts
```diff
@@ -17,7 +17,7 @@
 import * as ioredisTypes from 'ioredis';
 import { PluginConfig } from '@opentelemetry/api';
 
-export interface IORedisCommand {
+export interface IoRedisCommand {
```
naseemkullah, line 20:
```suggestion
export interface IoredisCommand {
```

and all occurences of `IoRedis` as ioredis is one word https://github.com/luin/ioredis

packages/opentelemetry-test-utils/testUtils.ts
dyladan: This is a dependency of the package so it should be a real dependency.

rezakrimi:
@OlivierAlbertini Since you're the author of the DNS plugin, could you please help on resolving the issue mentioned [here](https://docs.google.com/document/d/1bxrwuxEP6acxRXGe7hT5THJLjYXm3Bt924jZ72C_82s/edit?usp=sharing) for that plugin?

dyladan:
@rezakrimi can you update the doc to make it clear which issues you have resolved and which you have not?

rezakrimi:
> @rezakrimi can you update the doc to make it clear which issues you have resolved and which you have not?

I did, it's on the header for each package

plugins/node/opentelemetry-plugin-dns/test/integrations/dnspromise-lookup.test.ts
```diff
@@ -122,52 +132,77 @@ describe('dns.promises.lookup()', () => {
...
         assert.strictEqual(spans.length, 1);
-        assertSpan(span, {
-          addresses: [],
-          hostname,
-          forceStatus: {
-            code: CanonicalCode.INVALID_ARGUMENT,
-            message: error!.message,
-          },
-        });
+        if (semver.lt(process.versions.node, '9.0.0')) {
```
rezakrimi, line 142: I created a new dns lookup method in this file that is compatible with node 8. The only thing is that the node `INVALID_ARGUMENT` code seemed to be different in node 8 so I had to make separate cases for this test. If this is fine then I'll open this PR to be merged.

eslint.config.js
obecny: @dyladan maybe we can create some general package that can be easily shared between contrib repo and here andd maybe even a 3rd party contrib. We could put there a bit more stuff, eslint, karma config, etc. basically all things that could be shared. WDYT ?

plugins/node/opentelemetry-plugin-dns/test/integrations/dnspromise-lookup.test.ts
dyladan: Any idea why the canonical code is different between them? I would rather have behavior consistent across versions if possible

rezakrimi: base on [this](https://nodejs.org/api/errors.html#errors_error_code) it seems like it's possible to have different error codes in different major releases. The error message was the same though.

PR #118, opened 2020-07-02
chore: update contrib repo

mayurkale22:
- Update codeowners file.
- Add `README.md` for propagators.
- Update top `README.md`.
- Remove duplicate example.


mayurkale22:
> should we list here only the plugins that are in this repo - currently we show all ?

If you don't mind, I would like to do it in the separate PR.

PR #139, opened 2020-07-10
feat: mimic core repo renovate config

naseemkullah:


naseemkullah:
My pleasure as always @mayurkale22

PR #144, opened 2020-07-17
feat: adding Koa instrumentation

### LINKED ISSUE #93: Plugin for Koa
Requesting a plugin for the [Koa](https://koajs.com/) framework

Node
Weekly Downloads:
[525k](https://www.npmjs.com/package/koa)

**Additional context**
Existing plugin for OpenTracing is shown [here](https://github.com/teambition/koa-opentracing)

carolinee21:
- Fixes #93
- Adds automatic instrumentation for Koa, a Node.js web framework.

- This instrumentation adds traces to middleware for both the Koa core and [Koa Router](https://github.com/koajs/router) modules.
- Some example traces are shown here: Example Koa Instrumentation in Zipkin

- User friendly example code is also added under [examples/koa](https://github.com/carolinee21/opentelemetry-js-contrib/blob/koa-plugin/examples/koa/)


plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { BasePlugin } from '@opentelemetry/core';
+import * as koa from 'koa';
```
dyladan, line 18:
```suggestion
import type * as koa from 'koa';
```

plugins/node/opentelemetry-koa-instrumentation/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Middleware, ParameterizedContext, DefaultState } from 'koa';
+import { RouterParamContext } from '@koa/router';
+import Router = require('@koa/router');
```
dyladan, line 18: Do you have to depend on `@koa/router`? Why require instead of regular import? Can this be 
```suggestion
import type * as Router from '@koa/router';
```

plugins/node/opentelemetry-koa-instrumentation/src/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+export const getMiddlewareMetadata = (
+  context: KoaContext,
+  layer: KoaMiddleware,
+  isRouter: boolean,
+  layerPath?: string
+): {
+  attributes: Attributes;
+  name: string;
+} => {
+  if (isRouter) {
```
dyladan, line 34: Would rather just have one function for router and another function for whatever not router is instead of this boolean switching.

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+        patchedFunction = plugin._patchLayer(middlewareFunction, false);
+      }
+
+      args[0] = patchedFunction;
+      const res = original.apply(this, args);
+
+      return res;
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any;
```
dyladan, line 87: No way to do this without any?

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+    super('@opentelemetry/koa-instrumentation', VERSION);
+  }
+
+  /**
+   * Patches Koa operations by wrapping the Koa.use function
+   */
+  protected patch(): typeof koa {
+    this._logger.debug('Patching Koa');
+
+    if (this._moduleExports === undefined || this._moduleExports === null) {
```
dyladan, line 45:
```suggestion
    if (this._moduleExports == null) {
```

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+      return this._moduleExports;
+    }
+    this._logger.debug('Patching Koa.use');
+    const appProto = this._moduleExports.prototype;
```
dyladan, line 49: Why allocate a new var for this instead of using it directly?

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+    shimmer.wrap(appProto, 'use', this._getKoaUsePatch.bind(this));
```
dyladan, line 50: use of `bind` is often confusing
```suggestion
    shimmer.wrap(appProto, 'use', (original) => this._getKoaUsePatch(original));
```

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+      ...args: Parameters<typeof original>
+    ) {
+      let patchedFunction;
+      if (middlewareFunction.router) {
+        patchedFunction = plugin._patchRouterDispatch(middlewareFunction);
+      } else {
+        patchedFunction = plugin._patchLayer(middlewareFunction, false);
+      }
+
+      args[0] = patchedFunction;
```
dyladan, line 81:
```suggestion
      args[0] = middlewareFunction.router ?
        plugin._patchRouterDispatch(middlewareFunction) :
        patchedFunction = plugin._patchLayer(middlewareFunction, false);
```

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+      const res = original.apply(this, args);
+
+      return res;
```
dyladan, line 84:
```suggestion
      return original.apply(this, args);
```

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
dyladan: Why are you overwriting `args[0]`? Did you mean to use something like `unshift` instead?

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
vmarchaud: We really need to find a consensus on loose comparison :/

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+    middlewareLayer: KoaMiddleware,
+    isRouter: boolean,
+    layerPath?: string
+  ): KoaMiddleware {
+    if (middlewareLayer[kLayerPatched] === true) return middlewareLayer;
+    middlewareLayer[kLayerPatched] = true;
+    this._logger.debug('patching Koa middleware layer');
+    return async (context: KoaContext, next: koa.Next) => {
+      if (this._tracer.getCurrentSpan() === undefined) {
+        return await middlewareLayer(context, next);
```
vmarchaud, line 137: I think you don't need the async/await here, since `middlewareLayer` should be async anyway you could just `return middlewareLayer(context, next)` and remove the `async` keyword above.

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
vmarchaud: To avoid overwriting `args`, you could also just do 
```ts
const patchedFunction = middlewareFunction.router ?
        plugin._patchRouterDispatch(middlewareFunction) :
        patchedFunction = plugin._patchLayer(middlewareFunction, false);
return original.apply(this, [ patchedFunction, ...args ])
```

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
vmarchaud: Not sure the bind is necessary at all

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,155 @@
...
+    shimmer.wrap(appProto, 'use', this._getKoaUsePatch.bind(this));
+
+    return this._moduleExports;
+  }
+
+  /**
+   * Unpatches all Koa operations
+   */
+  protected unpatch(): void {
+    const appProto = this._moduleExports.prototype;
```
vmarchaud, line 59: could you had a debug log to know that you are unpatching ?

vmarchaud: Also same as @dyladan said above, you don't really need to declare a variable for this

plugins/node/opentelemetry-koa-instrumentation/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Parameters<T> = T extends (...args: infer T) => any ? T : unknown[];
```
vmarchaud, line 21: I believe `Parameters` exist in the Typescript stdlib so you shouldn't need it there. (I know that i've used it in the express plugin, but i'll to admit i don't remember why ^^)

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: I agree you do not need the await here when you are not adding tracing and just returning `middlewareLayer(context, next)` (line 137). The await on line 148 is necessary though for the span timing to fully capture any asynchronous processes related to the middleware, so I will need to keep the async keyword above.

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
dyladan: Without the bind, if someone modifies the `_getKoaUsePatch` function later to reference `this`, it will fail and may not be obvious why.

carolinee21: Hmm okay true - would you prefer that I leave it in? It is actually unnecessary for the current implementation

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: Thanks for the suggestions! I ended up removing args entirely because the Koa `app.use` function does not allow any additional parameters anyways.

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: Will do, thanks!

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: Good catch - this was left over. Just fixed

plugins/node/opentelemetry-koa-instrumentation/src/types.ts
carolinee21: Just removed - actually no longer need this at all because of the change to args in koa.ts

plugins/node/opentelemetry-koa-instrumentation/src/utils.ts
carolinee21: This would require boolean switching on `isRouter` in the `_patchLayer` function of koa.ts - right now, the `_patchLayer` function is designed to instrument both types of middleware smoothly. I can definitely change to checking `isRouter` there, but I think it might be a bit less clean than deferring it to here in utils

examples/koa/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "url": "https://github.com/open-telemetry/opentelemetry-js-contrib/issues"
+  },
+  "dependencies": {
+    "@koa/router": "^9.3.1",
+    "@opentelemetry/api": "^0.8.3",
+    "@opentelemetry/exporter-jaeger": "^0.8.3",
+    "@opentelemetry/exporter-zipkin": "^0.8.3",
+    "@opentelemetry/node": "^0.8.3",
+    "@opentelemetry/plugin-http": "^0.8.3",
+    "@opentelemetry/koa-instrumentation": "../../plugins/node/opentelemetry-koa-instrumentation",
```
markwolff, line 40: If the lerna picks up your new instrumentation folder, you should be able to just provide the semver here instead.

plugins/node/opentelemetry-koa-instrumentation/src/types.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export type KoaMiddleware = Middleware<DefaultState, KoaContext> & {
+  [kLayerPatched]?: boolean;
+  router?: Router;
+};
+
+export type KoaContext = ParameterizedContext<DefaultState, RouterParamContext>;
+
+export enum AttributeNames {
+  COMPONENT = 'component',
```
markwolff, line 34: `'component'` attribute [is deprecated now](https://github.com/open-telemetry/opentelemetry-specification/blob/b8948d9cec3994e396898feca7a10025d08370db/CHANGELOG.md#v030-02-21-2020). It has no replacement, so you can delete here and anywhere else it was set

plugins/node/opentelemetry-koa-instrumentation/src/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  HTTP_ROUTE = 'http.route',
```
markwolff, line 35: You can grab this from [@opentelemetry/semantic-conventions](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-semantic-conventions) so you won't need to re-rev this file if names ever change.

examples/koa/package.json
carolinee21: Good to know, will fix!

plugins/node/opentelemetry-koa-instrumentation/src/types.ts
carolinee21: Okay, will do

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: For this case, should I apply the suggested changes?

markwolff: I think so. You can create a separate tracking issue off of @vmarchaud's comment

vmarchaud: yeah go for it, there already is an issue open there https://github.com/open-telemetry/opentelemetry-js/issues/1313

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
vmarchaud: @dyladan The `this` is correctly typed there https://github.com/open-telemetry/opentelemetry-js-contrib/pull/144/files#diff-58a707d20a03601b47963c43aae79ed7R66 so i think it's fine not binding it.

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,141 @@
...
+    for (const pathLayer of routesStack) {
+      const path = pathLayer.path;
+      const pathStack = pathLayer.stack;
+      for (let j = 0; j < pathStack.length; j++) {
+        const routedMiddleware: KoaMiddleware = pathStack[j];
+        pathStack[j] = this._patchLayer(routedMiddleware, true, path);
+      }
+    }
+    const dispatcher: KoaMiddleware = (context: KoaContext, next: koa.Next) => {
+      return dispatchLayer(context, next);
```
vmarchaud, line 99: Can you just call the `displayLayer` directly ? Not sure i get why `dispatcher` exists here

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
dyladan: That is only a hint to typescript telling it what the `this` object's type _should_ be. It does not change behavior at all. If you do not bind context properly, the typings will simply be wrong.

dyladan: Ah I see there is no reference into the plugin. Sorry for my hasty judgement. It is fine without bind or arrow.

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: Good call - I think this was left over from previous iterations

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
carolinee21: Okay, I have removed it

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const router = dispatchLayer.router;
+
+    const routesStack = router?.stack ?? [];
+    for (const pathLayer of routesStack) {
+      const path = pathLayer.path;
+      const pathStack = pathLayer.stack;
+      for (let j = 0; j < pathStack.length; j++) {
+        const routedMiddleware: KoaMiddleware = pathStack[j];
+        pathStack[j] = this._patchLayer(routedMiddleware, true, path);
```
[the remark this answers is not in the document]
carolinee21: Thank you for bringing this up! Both nested and prefixed routers are handled currently and I believe the span names and attributes should function as expected. These routes are instrumented at the time that the routes are actually added to the app itself (in `app.use`).

For example, for the following nested router:
```javascript
const nestedRouter = new Router();
nestedRouter.get('/post/:id', showPost);
router.use('/:first', nestedRouter.routes());
app.use(router.routes());
```
The span recorded for a request to `/testparam/post/2` looks as follows: 
nestedRouter

Prefixed routes such as the following are also instrumented the same way, and result in the same spans:
```javascript
router.get('/post/:id', showPost);
router.prefix('/:first');
app.use(router.routes());
```

Let me know if you have any feedback on these span attributes or if this does not match what you were thinking. Also, please let me know if you have any further questions!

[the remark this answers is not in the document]
michaelgoin: Great! I was interpreting that as patching the `router.use`, which was incorrect. I should have re-read to double-check.

The name in your screenshot looks like what I'd expect so I expect the rest is legit. Thanks!

I might suggest some tests to represent these sort of cases as example, and to prevent accidentally breaking in the future, but extremely minor suggestion.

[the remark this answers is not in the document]
michaelgoin: Looks like I can't mark this resolved, but good on my end.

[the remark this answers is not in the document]
carolinee21: Thanks for the feedback! I appreciate the time you took to review this. I'll add a few tests on these to help in the future!

plugins/node/opentelemetry-koa-instrumentation/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@opentelemetry/koa-instrumentation",
+  "version": "0.8.0",
```
dyladan, line 3: Version does not match current

plugins/node/opentelemetry-koa-instrumentation/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "doc",
+    "LICENSE",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@koa/router": "^9.3.1",
+    "@opentelemetry/context-async-hooks": "0.9.0",
```
dyladan, line 46: opentelemetry dependencies out of date

plugins/node/opentelemetry-koa-instrumentation/package.json
carolinee21: Fixed

examples/koa/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "node": ">=8"
+  },
+  "author": "OpenTelemetry Authors",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/open-telemetry/opentelemetry-js-contrib/issues"
+  },
+  "dependencies": {
+    "@koa/router": "^9.3.1",
+    "@opentelemetry/api": "^0.8.3",
```
dyladan, line 35: Opentelemetry core modules are on 0.12.2 now.

plugins/node/opentelemetry-koa-instrumentation/package.json
dyladan: Please bump core deps to 0.12.2

carolinee21: @dyladan I would love it if you could help me look into this if you get a chance - I'm getting errors from these new versions of the opentelemetry core modules across other plugins as well as my own while developing locally. I also tried recloning a fresh copy of the contrib repo, then testing other plugins such as Express and MongoDB and ran into these same errors. I sent more information and repro steps over Gitter and would really appreciate it if you could take a look

examples/koa/package.json
```diff
@@ -0,0 +1,50 @@
...
```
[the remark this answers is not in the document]
carolinee21: Hi @obecny! I am currently running into some errors with version 0.10.1, but not in my own plugin - 
For the Koa plugin, it is passing all tests using the core version 0.10.2, but was running into some errors with 0.10.1. I think there may be some issues with 0.10.1, at least in my local environment, because I also tried running tests in the Express plugin locally, and they failed as well (see image).

Screenshot 2020-08-05 at 9 49 01 AM

In the Express plugin, bumping the version to 0.10.2 actually did fix these errors, though I'm not sure it would solve the errors across all plugins currently using 0.10.1. 

Have you been experiencing any issues with the version bump to 0.10.1 in the contrib repo, or would you be willing to help me look into this further? I would really appreciate it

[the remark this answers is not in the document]
michaelgoin: @carolinee21 ran down the cause for Express, which might relate to what you've tried/tested.

For express, there was a mismatch between `@opentelemetry/tracing`, `@opentelemetry/core`, and `@opentelemetry/api`.

`@opentelemetry/tracing` was hard-coded to 0.10.1 in the express package.json dev dependencies. The other two are ^0.10.1. So in reality, they were running 0.10.2 while tracing was still on 0.10.1. With 0.10.2, core no-longer exports a `randomSpanId`. Instead, an ID generator is exported to be used (which 0.10.2 of tracing does).

When all of those match, one way or the other, the install/compile/test process passes for me.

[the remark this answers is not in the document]
michaelgoin: It looks like this PR *should* solve that for express: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/170/files#diff-d2aa817a652c61faac8e33c6b07d8654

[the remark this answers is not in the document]
carolinee21: @michaelgoin Thank you so much for looking into this! It would be really great if that PR does solve some of these issues, and as soon as it merges I believe this PR should be unblocked!

plugins/node/opentelemetry-koa-instrumentation/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@koa/router": "^9.3.1",
+    "@opentelemetry/context-async-hooks": "0.10.2",
+    "@opentelemetry/node": "0.10.2",
+    "@opentelemetry/tracing": "0.10.2",
```
michaelgoin, line 48: An open question/thought having dug through this stuff recently (to everyone, not necessarily the author)...

It seems to be a pattern to have these dev-dependencies locked at a single version while the regular deps do the typical ^version. That has recently caused issues and seems like something that would be regularly prone to, at least pre-major release. 

Why wouldn't we want these dev-deps to update per usual and catch things and/or automatically keep in sync with full publishes of everything in opentelemetry-js? Wouldn't we want to know if these plugins were broken from a minor/patch release in the future?

carolinee21: Great point for discussion! I'm interested in hearing feedback from everyone else

plugins/node/opentelemetry-koa-instrumentation/test/koa.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  };
+
+  const asyncMiddleware: koa.Middleware = async (ctx, next) => {
+    const start = Date.now();
+    await next();
+    const ms = Date.now() - start;
+    ctx.body = `${ctx.method} ${ctx.url} - ${ms}ms`;
+  };
+
+  describe('Instrumenting core middleware calls', () => {
```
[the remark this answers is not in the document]
carolinee21: Good call, thanks! Just extracted this logic and the server.close() into beforeEach and afterEach blocks.

plugins/node/opentelemetry-koa-instrumentation/package.json
markwolff: I could be wrong, but I think dependabot intentionally locks them to a single version and updates on its own PRs only

obecny:
> @dyladan @obecny @michaelgoin Now that the other PR has been merged, this one is unblocked and should be ready to go!

@dyladan requested changes, wait for him to approve first

carolinee21:
@dyladan Hi Daniel! Please let me know if there's anything else missing from me on this or the Hapi PR. Thanks! :)

PR #158, opened 2020-07-23
fix: patch removeEventListener to properly remove patched callbacks

johnbley:
We need to patch removeEventListener in order to compensate for wrapping callbacks passed to addEventListener.

### DIFF
--- a/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
+++ b/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
@@ -49,7 +49,10 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
   private _spansData = new WeakMap<api.Span, SpanData>();
   private _zonePatched = false;
   // for addEventListener/removeEventListener state
-  private _wrappedListeners = new WeakMap<Function, Map<String, Map<HTMLElement, Function>>>();
+  private _wrappedListeners = new WeakMap<
+    Function,
+    Map<string, Map<HTMLElement, Function>>
+  >();
 
   constructor() {
     super('@opentelemetry/plugin-user-interaction', VERSION);
@@ -167,7 +170,12 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     }
   }
 
-  private addPatchedListener(on: HTMLElement, type: String, listener: Function, wrappedListener: Function) {
+  private addPatchedListener(
+    on: HTMLElement,
+    type: string,
+    listener: Function,
+    wrappedListener: Function
+  ) {
     let listener2Type = this._wrappedListeners.get(listener);
     if (!listener2Type) {
       listener2Type = new Map();
@@ -181,12 +189,16 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     element2patched.set(on, wrappedListener);
   }
 
-  private removePatchedListener(on: HTMLElement, type: String, listener: Function) : Function|undefined {
-    let listener2Type = this._wrappedListeners.get(listener);
+  private removePatchedListener(
+    on: HTMLElement,
+    type: string,
+    listener: Function
+  ): Function | undefined {
+    const listener2Type = this._wrappedListeners.get(listener);
     if (!listener2Type) {
       return undefined;
     }
-    let element2patched = listener2Type.get(type);
+    const element2patched = listener2Type.get(type);
     if (!element2patched) {
       return undefined;
     }
@@ -251,7 +263,11 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
         listener: any,
         useCapture: any
       ) {
-        const wrappedListener = plugin.removePatchedListener(this, type, listener);
+        const wrappedListener = plugin.removePatchedListener(
+          this,
+          type,
+          listener
+        );
         if (wrappedListener) {
           return original.call(this, type, wrappedListener, useCapture);
         } else {
--- a/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
+++ b/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
@@ -170,12 +170,15 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     }
   }
 
+  /**
+   * Returns true iff we should use the patched callback; false if it's already been patched
+   */
   private addPatchedListener(
     on: HTMLElement,
     type: string,
     listener: Function,
     wrappedListener: Function
-  ) {
+  ): boolean {
     let listener2Type = this._wrappedListeners.get(listener);
     if (!listener2Type) {
       listener2Type = new Map();
@@ -186,9 +189,16 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
       element2patched = new Map();
       listener2Type.set(type, element2patched);
     }
+    if (element2patched.has(on)) {
+      return false;
+    }
     element2patched.set(on, wrappedListener);
+    return true;
   }
 
+  /**
+   * Returns the patched version of the callback (or undefined)
+   */
   private removePatchedListener(
     on: HTMLElement,
     type: string,
@@ -243,8 +253,9 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
             return listener.apply(target, args);
           }
         };
-        plugin.addPatchedListener(this, type, listener, patchedListener);
-        return original.call(this, type, patchedListener, useCapture);
+        if (plugin.addPatchedListener(this, type, listener, patchedListener)) {
+          return original.call(this, type, patchedListener, useCapture);
+        }
       };
     };
   }
--- a/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
+++ b/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
@@ -48,6 +48,7 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
   moduleName = this.component;
   private _spansData = new WeakMap<api.Span, SpanData>();
   private _zonePatched = false;
+  private _wrappedListeners = new WeakMap<Function, Function>();
 
   constructor() {
     super('@opentelemetry/plugin-user-interaction', VERSION);
@@ -193,7 +194,32 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
             return listener.apply(target, args);
           }
         };
+        plugin._wrappedListeners.set(listener, patchedListener);
         return original.call(this, type, patchedListener, useCapture);
+      };
+    };
+  }
+
+  /**
+   * This patches the removeEventListener of HTMLElement to handle the fact that
+   * we patched the original callbacks
+   * This is done when zone is not available
+   */
+  private _patchRemoveEventListener() {
+    const plugin = this;
+    return (original: Function) => {
+      return function removeEventListenerPatched(
+        this: HTMLElement,
+        type: any,
+        listener: any,
+        useCapture: any
+      ) {
+        const wrappedListener = plugin._wrappedListeners.get(listener);
+        if (wrappedListener) {
+          return original.call(this, type, wrappedListener, useCapture);
+        } else {
+          return original.call(this, type, listener, useCapture);
+        }
       };
     };
   }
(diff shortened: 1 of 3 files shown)


plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -193,11 +194,36 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
...
+  private _patchRemoveEventListener() {
+    const plugin = this;
+    return (original: Function) => {
+      return function removeEventListenerPatched(
+        this: HTMLElement,
+        type: any,
+        listener: any,
+        useCapture: any
+      ) {
+        const wrappedListener = plugin._wrappedListeners.get(listener);
```
dyladan, line 217: I would also remove the wrapped listener from the weakmap here

johnbley: Good call; patching now.

dyladan:
@rezakrimi the lint action is failing completely. Can you please look into this?

rezakrimi:
> @rezakrimi the lint action is failing completely. Can you please look into this?

Sure.

obecny:
Can you please provide more details what is the exact problem here ?, thx

johnbley:
> Can you please provide more details what is the exact problem here ?, thx

Sure thing!  The added test shows that the addEventListener instrumentation actually breaks apps because it doesn't have the compensating logic for removeEventListener.  A patched callback is actually added as the event listener, and when the app tries to remove the original one, it's not found/ignored, leaving our patched callback present (along with unexpected calls to the underlying original one).  For some apps, this doesn't matter, but for others this can be arbitrarily bad (breaking functionality, thrown errors, infinite loops, etc. etc.).

obecny:
it looks fine, the failing build might be connected with yesterday eslint merge, we have to check it @dyladan https://circleci.com/gh/open-telemetry/opentelemetry-js-contrib/1546

dyladan:
> it looks fine, the failing build might be connected with yesterday eslint merge, we have to check it @dyladan [circleci.com/gh/open-telemetry/opentelemetry-js-contrib/1546](https://circleci.com/gh/open-telemetry/opentelemetry-js-contrib/1546)
> EDIT:
> I might be wrong with the link, but when checking the lint in this PR something else is failing not related to this PR

It is definitely related. eslint isn't running at all because of a gts dependency error.

rezakrimi:
> > it looks fine, the failing build might be connected with yesterday eslint merge, we have to check it @dyladan [circleci.com/gh/open-telemetry/opentelemetry-js-contrib/1546](https://circleci.com/gh/open-telemetry/opentelemetry-js-contrib/1546)
> > EDIT:
> > I might be wrong with the link, but when checking the lint in this PR something else is failing not related to this PR
> 
> It is definitely related. eslint isn't running at all because of a gts dependency error.

yes the caching was wrong, I'll make a PR today to fix it

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -193,11 +194,37 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
             return listener.apply(target, args);
           }
         };
+        plugin._wrappedListeners.set(listener, patchedListener);
```
lykkin, line 197: It looks like the listeners are assumed to be unique. If so this wouldn't handle reusing the same listener, though I don't know if that's an edge case we want to worry about.

rezakrimi:
#159 should fix the lint check.

dyladan:
@johnbley if you update from master now the lint should run properly

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
johnbley: No, this is an important point, thanks.  In particular I'm going to tweak it to properly support
```
  element.addEventListener('type1', listener);
  element.addEventListener('type2', listener);
  element.removeEventListener('type1', listener);
```
Which the current instrumentation does not properly support.  Thanks for the careful review!

johnbley:
> @johnbley if you update from master now the lint should run properly

Will do, but please hold on merging until I find the time to properly handle some edge cases better.

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
obecny: When I start thinking more about this it is becoming more complicated when I originally thought and it will require to make few other changes because when you call AddEventListener with the same callback and type it will overwrite previous callback. Also it should only create span for click event. So when saving patchedListener you should also save the type

If I haven't missed anything the patch function should look like
```
  private _wrappedListeners = new WeakMap<Function, Record<string, ()=>{}>>();
  private _patchElement() {
    const plugin = this;
    return (original: Function) => {
      return function addEventListenerPatched(
        this: HTMLElement,
        type: string,
        listener: any,
        useCapture: any
      ) {
        const patchedListener = (...args: any[]) => {
          const target = this;
          if (type.toLowerCase() !== 'click') {
            return listener.apply(target, args);
          }
          const span = plugin._createSpan(target, 'click');
          if (span) {
            return plugin._tracer.withSpan(span, () => {
              const result = listener.apply(target, args);
              // no zone so end span immediately
              span.end();
              return result;
            });
          } else {
            return listener.apply(target, args);
          }
        };
        let listeners = plugin._wrappedListeners.get(original);
        if (!listeners) {
          listeners = {};
          listeners[type] = patchedListener;
          plugin._wrappedListeners.set(original, listeners);
        } else if (listeners[type]) {
          this.removeEventListener(type, listeners[type]);
          delete listeners[type];
        }
        listeners[type] = patchedListener;
        // not sure if the below line is yet needed
        plugin._wrappedListeners.set(original, listeners);
        return original.call(this, type, patchedListener, useCapture);
      };
    };
  }
```

After this you will also need to modify the patch for `removeEventListener` so that it will take into consideration the type.

johnbley: This is very similar to where I landed, with the added bit that, of course, the element it's on is a part of the registration.  I've got a separate patch for better handling of the event type and would rather not conflate the two.

johnbley:
Looks like the lint step in the build is failing again, but this time with out-of-date code... I ran lint:fix and pushed those changes.

rezakrimi:
> Looks like the lint step in the build is failing again, but this time with out-of-date code... I ran lint:fix and pushed those changes.

That's strange. Same thing happened to my PR on the main opentelemetry-js repo. For me it just got resolved after pushing another commit.

plugins/web/opentelemetry-plugin-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -89,6 +89,40 @@ describe('UserInteractionPlugin', () => {
       context.disable();
     });
 
+    it('should not break removeEventListener', () => {
+      let called = false;
+      const listener = function () {
+        called = true;
+      };
+      // add same listener three different ways
```
obecny, line 97: I'm curious about one scenario, to check that can you modify this a bit to test the case with overwriting the previous event with different callback
```
let called = false;
let called2 = false;
const listener = function () {
  called = true;
};
const listener2 = function () {
  called2 = true;
};
document.body.addEventListener('bodyEvent', listener);
document.body.addEventListener('bodyEvent', listener2);
document.body.addEventListener('bodyEvent2', listener);

// now fun
document.body.dispatchEvent(new Event('bodyEvent'));
assert.strictEqual(called, false);
assert.strictEqual(called2, true);
```

obecny: fixed my example

johnbley: Yep, good catch and an easy fix for this one based on the tracking state.  Pushing fix momentarily.

obecny: There is one more thing a bonus :). Currently you can call `addEventListener` providing a 3rd param `options`. In options you can set `once:true` which means the event will be removed automatically. It might be worth to test that too.

obecny:
```
let called2 = 0;
const listener2 = function () {
  called2++;
};
document.body.addEventListener('bodyEvent', listener2);
document.body.addEventListener('bodyEvent', listener2, { once: true });

// now fun
document.body.dispatchEvent(new Event('bodyEvent'));
assert.strictEqual(called2, 1);
document.body.dispatchEvent(new Event('bodyEvent'));
assert.strictEqual(called2, 2); // error ?
```

obecny: and you can revert it add `once` first and then add without options, heh small things and many edge cases

johnbley: Argh.  There's also the case that once:true shouldn't prevent us from allowing it to be adding later (it's not really a double-register).  Will dive in and fix this one tomorrow.  Thanks!

obecny: thx for taking this into deep details :)

johnbley: Couldn't leave it alone; handled the `once:true` callbacks properly being removed when an event is dispatched.

plugins/web/opentelemetry-plugin-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -89,6 +89,73 @@ describe('UserInteractionPlugin', () => {
...
+      let callCount = 0;
+      const listener = function () {
+        callCount++;
+      };
+      // addEventListener semantics treat the second call as a no-op
+      document.body.addEventListener('bodyEvent', listener, { once: true });
+      document.body.addEventListener('bodyEvent', listener); // considered a double-register
+      document.body.dispatchEvent(new Event('bodyEvent'));
+      assert.strictEqual(callCount, 1);
+      // now that it's been dispatched once, it's been removed; should be able to re-add
```
obecny, line 153: Nit: I would add here extra call and check
```
document.body.dispatchEvent(new Event('bodyEvent'));
assert.strictEqual(callCount, 1);
```

plugins/web/opentelemetry-plugin-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -89,6 +89,73 @@ describe('UserInteractionPlugin', () => {
...
+      document.body.addEventListener('bodyEvent', listener);
+      document.body.dispatchEvent(new Event('bodyEvent'));
+      assert.strictEqual(callCount, 2);
```
obecny, line 156: Nit: and the same here

```
document.body.dispatchEvent(new Event('bodyEvent'));
assert.strictEqual(callCount, 3);
```

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -165,6 +170,61 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
...
+   */
+  private addPatchedListener(
+    on: HTMLElement,
+    type: string,
+    listener: Function,
+    wrappedListener: Function
+  ): boolean {
+    let listener2Type = this._wrappedListeners.get(listener);
+    if (!listener2Type) {
+      listener2Type = new Map();
```
lykkin, line 184: Is this going to be coerced into a `WeakMap` when you set it on `_wrappedListeners`?

johnbley: No, though making the HTMLElement reference weak is a good idea.  Unfortunately I don't see a way to do it since there are no iterators/size/isEmpty methods available on WeakMap and without it we'd *definitely* leak.  I'll ponder this but I'd like to get this fix in as it is since it's clearly better than the current breaking of functional behavior.

johnbley:
> great job, thx for investigating this further :)

Thanks to all for spotting further issues and edge cases!

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
lykkin: Awesome, thanks for the clarification!

PR #160, opened 2020-07-23
fix typos

shivkanya9146:

-

-


obecny:
master is not failing, this is failing and it looks like it is merged with master, strange, @shivkanya9146 can you try to make sure it is merged with master, maybe some caching issue etc. ?

PR #162, opened 2020-07-24
chore: clean up span naming

johnbley:
- Clean up span naming to OpenTelemetry specifications (existing names are discouraged as having high cardinality)
- Centralize logic for what event types to allow, and allow `dblclick` as well as `click`
- (Bad form, but blergh): forgot to unpatch removeEventListener from yesterday's work.

### DIFF
--- a/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
+++ b/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
@@ -83,7 +83,7 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
    * Controls whether or not to create a span, based on the event type.
    */
   protected _allowEventType(eventType: string): boolean {
-    return eventType === 'click' || eventType === 'dblclick';
+    return eventType === 'click';
   }
   /**
    * Creates a new span
--- a/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
+++ b/plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
@@ -34,7 +34,6 @@ import { AttributeNames } from './enums/AttributeNames';
 import { VERSION } from './version';
 
 const ZONE_CONTEXT_KEY = 'OT_ZONE_CONTEXT';
-const EVENT_CLICK_NAME = 'event_click:';
 const EVENT_NAVIGATION_NAME = 'Navigation:';
 
 /**
@@ -81,6 +80,12 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
   }
 
   /**
+   * Controls whether or not to create a span, based on the event type.
+   */
+  protected allowEventType(eventType: string): boolean {
+    return eventType === 'click' || eventType === 'dblclick';
+  }
+  /**
    * Creates a new span
    * @param element
    * @param eventName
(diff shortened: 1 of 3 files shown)


kkruk-sumo:
@johnbley Is there a plan to do the same for `plugin-document-load` and `plugin-xml-http-request` (in both cases URLs are used as span operation names)?

johnbley:
> @johnbley Is there a plan to do the same for `plugin-document-load` and `plugin-xml-http-request` (in both cases URLs are used as span operation names)?

Yes, working my way through a number of smaller issues like this.  Doing one area at a time to not overwhelm the maintainers. 😄

dyladan:
@obecny particularly looking for your feedback on this one

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -80,6 +79,12 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     }
   }
 
+  /**
+   * Controls whether or not to create a span, based on the event type.
+   */
+  protected allowEventType(eventType: string): boolean {
```
dyladan, line 85: this should be private as it is not used by any subclass?

Prefix protected and private methods with _

johnbley: I wanted to leave a way for users to specialize their own list of "interesting enough" events (possibly even dynamic, unlike with a config item).  If that's not cool I can make it private.  Either way I'll prefix with `_`.

dyladan: By writing an instrumentation module which extends ours? that is an interesting idea.

johnbley: Also: perhaps the lint rules could be expanded to enforce the naming?

johnbley:
> We can either add such possibility to a plugin config. This can be a 2 callbacks
Sorry, not understanding this.  You would like `_updateInteractionName` to be protected as well?

> By default I would support click event only
Will do.

obecny:
> > We can either add such possibility to a plugin config. This can be a 2 callbacks
> 
> Sorry, not understanding this. You would like `_updateInteractionName` to be protected as well?
> 
> > By default I would support click event only
> 
> Will do.

I meant when you instantiate plugin

```
const providerWithZone = new WebTracerProvider({
  plugins: [
    new UserInteractionPlugin({
      allowEventType(eventType: string) {
        if (eventType === ....) { // write your own logic
          return true;
        }
      },
      updateSpanName(target: HTMLElement) {
       // write your own logic
       return 'spanName';
      }
    }),
  ]
});

```

and 2nd possibility

```
class MyUserInteractionPlugin extends UserInteractionPlugin {
  allowEventType(eventType: string) {
    if (eventType === ....) { // write your own logic
      return true;
    }
  },
  updateSpanName(target: HTMLElement) {
   // write your own logic
   return 'spanName';
  }
}

```

johnbley:
> class MyUserInteractionPlugin extends UserInteractionPlugin {

This (strongly typed) is my preference - easier for IDEs, static tools, etc.  to analyze/understand the specialization.  That's what the current patch uses.

PR #163, opened 2020-07-26
feat: add AWS Xray Propagator

EdZou:
As discussed on Node SIG meeting, adding AWS Xray propagator components under opentelemetry-js-contrib/propagators repository

1. Add AWS Xray propagator components
2. Add corresponding unit test for AWS Xray propagator
3. Add corresponding README.md file for AWS Xray propagator


propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,200 @@
...
+  private getSpanContextFromHeader(carrier: unknown, getter: GetterFunction): SpanContext {
+    const traceHeader = getter(carrier, AWSXRAY_TRACE_ID_HEADER);
+    // Only if the returned traceHeader is no empty string can be extracted
+    if (!traceHeader || typeof traceHeader !== 'string') return INVALID_SPAN_CONTEXT;
+
+    let pos = 0, trimmedPart: string;
+    let parsedTraceId = INVALID_TRACE_ID;
+    let parsedSpanId = INVALID_SPAN_ID;
+    let parsedTraceFlags = null;
+    while(pos < traceHeader.length) {
```
anuraaga, line 119: In Java, this is quite a bit faster than using `split`, but I don't know about JS, it may be better to stick to using `split` here.

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,200 @@
...
+const TRACE_ID_KEY = 'Root';
+const TRACE_ID_LENGTH = 35;
+const TRACE_ID_VERSION = "1";
+const TRACE_ID_DELIMITER = '-';
+const TRACE_ID_DELIMITER_INDEX_1 = 1;
+const TRACE_ID_DELIMITER_INDEX_2 = 10;
+const TRACE_ID_FIRST_PART_LENGTH = 8;
+
+const PARENT_ID_KEY = 'Parent';
+// const PARENT_ID_LENGTH = 16;
```
anuraaga, line 48: Remove

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,200 @@
...
+ * An example AWS Xray Tracing Header is shown below:
+ * X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1
+ */
+export class AWSXRayPropagator implements HttpTextPropagator {
+  inject(context: Context, carrier: unknown, setter: SetterFunction) {
+    const spanContext = getParentSpanContext(context);
+    if (!spanContext) return;
+    if (!VALID_TRACEID_REGEX.test(spanContext.traceId) ||
+      !VALID_SPANID_REGEX.test(spanContext.spanId) ||
+      (spanContext.traceFlags !== TraceFlags.NONE &&
```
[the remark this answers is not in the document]
EdZou: No, I try to verify the sampled Flag we get should be either TraceFlags.None or TraceFlags.sampled. Other flags are invalid

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
EdZou: I am also confused here, I initially tried to use split method but the Java implementation is in this way so I just followed it

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,200 @@
...
+ * Tracing header spec</a>
+ * 
+ * An example AWS Xray Tracing Header is shown below:
+ * X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1
+ */
+export class AWSXRayPropagator implements HttpTextPropagator {
+  inject(context: Context, carrier: unknown, setter: SetterFunction) {
+    const spanContext = getParentSpanContext(context);
+    if (!spanContext) return;
+    if (!VALID_TRACEID_REGEX.test(spanContext.traceId) ||
```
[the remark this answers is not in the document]
EdZou: OK, I will do it, thank you!

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
EdZou: will do it!

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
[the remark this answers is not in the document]
anuraaga: Ah I realized those are the only two values of the traceflags, and it seems `TraceFlags` is not defined using `?` so it's always set, line 81 seems suspicious. Just change to check whether the flags is sampled on that line and remove this check.

propagators/opentelemetry-propagator-aws-xray/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "opentelemetry-propagator-aws-xray",
```
dyladan, line 2:
```suggestion
  "name": "@opentelemetry/propagator-aws-xray",
```

dyladan: @open-telemetry/javascript-maintainers do we want to have another namespace or even just a naming convention for vendor-specific and contrib things? Maybe `@otel-contrib/propagator-aws-xray` or `@opentelemetry/contrib-propagator-aws-xray`?

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
dyladan: There is a benchmark utility you can use to test things like this. You can see an example at https://github.com/open-telemetry/opentelemetry-js/issues/1334

You can also use jsperf.com for browser benchmarking.

dyladan: Subjectively, I think `split` based implementations are typically easier to read, which is an important distinction. I would expect your implementation here to be slightly faster than a split implementation, but not appreciably.

dyladan: If the ids are fixed length, it might be easier and faster to use substring also.

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  private _parseSpanId(xrayParentId: string): string{
+    return (VALID_SPANID_REGEX.test(xrayParentId))? xrayParentId: INVALID_SPAN_ID;
+  }
+
+  private _parseTraceFlag(xraySampledFlag: string): TraceFlags | null {
+    if (xraySampledFlag.length === SAMPLED_FLAG_LENGTH && xraySampledFlag === NOT_SAMPLED) {
+      return TraceFlags.NONE;
+    } else if (xraySampledFlag === IS_SAMPLED) {
```
dyladan, line 194: No need for `else` after a `return`

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,200 @@
...
+    // Check the content of trace id
+    if (!VALID_TRACEID_REGEX.test(resTraceId)) {
+      return INVALID_TRACE_ID
+    }
+
+    return resTraceId;
+  }
+
+  private _parseSpanId(xrayParentId: string): string{
+    return (VALID_SPANID_REGEX.test(xrayParentId))? xrayParentId: INVALID_SPAN_ID;
```
dyladan, line 188: style nit: we typically use spaces surrounding the `?` and `:`
```suggestion
    return VALID_SPANID_REGEX.test(xrayParentId) ? xrayParentId : INVALID_SPAN_ID;
```

EdZou: Yes, will do it! Really thank you for your advice! 
Another thing about lint I am pretty confused is that, I followed file structure of jaegerHttpTracePropagator, but I cannot pass the ```npm run lint``` test of GA. I checked eslint file and found it seems to run ```eslint . --ext .ts``` by using eslint.config.js file. eslint.config.js file does exist but still fails, I am blocked here and have no idea what to do.

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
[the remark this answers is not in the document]
EdZou: Got it! I initially do this check in case if there is a traceFlags value is neither NONE or SAMPLED. I will remove it

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
EdZou: I now think Daniel's consideration makes more sense. Since we have a fixed size tracing header, it can be a little bit faster to use substring, and maybe that is why Java also implements in this way

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
EdZou: oh my fault, it seems ```npm run test``` command cannot be run locally but GA does this successfully and gives me error info, I will fix lint errors and make another commit.

propagators/opentelemetry-propagator-aws-xray/package.json
EdZou: No problem, which one should it be changed? ```@opentelemetry/contrib-propagator-aws-xray``` or ```@otel-contrib/propagator-aws-xray```

dyladan: For now, lets do `@opentelemetry/propagator-xray`. I will bring up the naming discussion at tomorrow's SIG meeting.

EdZou: Got it! Thank you so much!

EdZou:
Hello, Daniel @dyladan 
After discussion with my mentor Anuraag @anuraaga , we believe according to the specification here: https://github.com/open-telemetry/opentelemetry-specification/pull/735/files. We should add our AWS Xray propagator code under js-contrib repo anyway

dyladan:
I have discussed this with the other maintainers @obecny and @mayurkale22 and we would like to keep vendor-specific code out of our hosted repositories if at all possible. We are happy to code-review vendor propagators and exporters, but we would prefer that these code modules be hosted and deployed from vendor-owned repositories. As an example, you can look at what GCP has done for their cloud trace and cloud monitoring exporters and propagator here https://github.com/GoogleCloudPlatform/opentelemetry-operations-js. At the time this repository was created, we had similar discussions with the GCP folks.

anuraaga:
@dyladan Do you mind adding your perspective to the linked spec PR? 

https://github.com/open-telemetry/opentelemetry-specification/pull/735/files

It'd be better to do that before it gets merged - while yes it only says SHOULD, I think that's usually considered a recommendation since otherwise there's no reason to mention it at all. Not going with the recommendation is something that needs to have some real motivation I think.

AWS can host the module - if that's the final decision we're happy to. But I see discrepancy between JS, Java, and the spec in this area where there doesn't seem to be a huge reason to have that discrepancy. Propagators being an interop concern, therefore part of the SDKs does make sense to me, but they do feel vendorish and wanting to split makes sense to me too. My personal opinion is wanting consistency there between Java and JS one way or another though (those are the two languages I've looked at in depth so far).

EdZou:
Hello Daniel @dyladan , sorry to bother you again.
Actually When I trying to use ```npm link``` command to instrument our ```AWSXRayPropagator``` locally, it failed. And then I tried to use ```HttpTraceContext``` locally it also failed.
So in my comprehension, I believe to correctly instrument a propagator, we may need to publish the propagator first?
But when I tried to publish our ```AWSXRayPropagator```, it gave me 
```
npm ERR! code E403
npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/@opentelemetry%2fpropagator-aws-xray - Forbidden
npm ERR! 403 In most cases, you or one of your dependencies are requesting
npm ERR! 403 a package version that is forbidden by your security policy.
```
And I did not change the dependency of this package, would you mind helping me out here? Really appreciate it!

anuraaga:
@EdZou `npm link` is an alternative to publishing, it should generally work. When we publish that package it will be a different name and we'll have an aws namespace for it like `@aws-observability/opentelemetry-propagator-xray`.

Anyways, I believe `lerna link` should work well enough for your example project, but I notice now that at least in your PR the example doesn't have dependencies on the xray packages?

https://github.com/open-o11y/aws-opentelemetry-js/pull/3/files#diff-640d1aab881301a5fabb1482b0fa1d21R24

EdZou:
Yes, I did not explicitly add dependency in package.json, but it still works for instrumenting IdGenerator. Not so sure if that is the key probelm here @anuraaga

anuraaga:
@EdZou @dyladan I debugged the example app and while I didn't confirm for sure, I guess we're waiting on this fix to be released https://github.com/open-telemetry/opentelemetry-js/pull/1387 @dyladan do you think it's possible to make a release so propagators in external packages can work ok?

Issue #84, opened 2020-06-17
React/Redux plugin


thgao:
**Is your plugin request related to a problem? Please describe.**
Currently, React routing doesn't trigger tracing from other plugins like document load, but it'd be nice to be able to see this for React pages as well.

State storage in Redux would also be interesting to see traces of.

**Is it applicable for Node or Browser or both**
Browser

dyladan:
Assigned the issue to you as discussed in SIG. If you'd like any help or guidance feel free to reach out to me on gitter or ask in the https://gitter.im/open-telemetry/opentelemetry-node channel.

bbigras:
Any progress for functional components (like with react hooks)?

vmarchaud:
@bbigras Sorry for the late response but i don't think someone is working on this

vmarchaud:
I'm gonna close this one since initial plugin has been merged, functional component support will be tracked here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/178

PR #164, opened 2020-07-27
feat: Add React Plugin

thgao:

- #84 

- Adding plugin for React instrumentation + docs and tests
- This plugin instruments the lifecycle methods of a React component as defined here: https://projects.wojtekmaj.pl/react-lifecycle-methods-diagram/
**Note: Constructor & getDerivedStateFromProps are not instrumented because they are static methods.  getDerivedStateFromProps is considered an extremely [rare use case](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)**

## Usage
Once the tracer is defined, statically set logger & tracer:
```
BaseOpenTelemetryComponent.setLogger(logger);
BaseOpenTelemetryComponent.setTracer("name", "version");
```
Component definitions:
```
class export Component1 extend BaseOpenTelemetryComponent { ...}
class export Component2 extend BaseOpenTelemetryComponent { ...}
```
Any component extending `BaseOpenTelemetryComponent` will be instrumented.

## Example Traces
Example Attributes:
Screenshot 2020-07-27 at 4 01 43 PM

Mounting Flow:
Screenshot 2020-07-27 at 4 01 26 PM

Updating Flow:
Screenshot 2020-07-27 at 4 01 59 PM

Unmounting Flow:
Screenshot 2020-07-27 at 4 04 23 PM


plugins/web/opentelemetry-plugin-react-load/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```js
+import { BaseOpenTelemetryComponent } from '@opentelemetry/plugin-react-load';
+
+// Set once for the entire plugin
+BaseOpenTelemetryComponent.setLogger(logger);
+BaseOpenTelemetryComponent.setTracer('name', 'version');
+```
+
+To instrument components, extend `BaseOpenTelemetryComponent`:
```
markwolff, line 24: How can I instrument a functional component? Would I just wrap it as a HOC?

```tsx
const FuncComponent = ({...props}) => (<BaseOpenTelemetryComponent>
  <div {...props}>Hello World</div>
</BaseOpenTelemetryComponent>);
```

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```diff
@@ -0,0 +1,541 @@
...
+  }
+
+  /**
+   * Returns attributes object for spans
+   * @param react React component currently being instrumented
+   **/
+  private _getAttributes(react: React.Component) {
+    let state = '';
+    try {
+      state = JSON.stringify(react.state);
```
markwolff, line 162: @opentelemetry/javascript-maintainers
Do attributes need to be non-object types? I wonder if this stringify is something that should be left to the exporters. 
Else I think the attribute value type should be changed from `unknown` to clarify whether this stringify is necessary or not.

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```diff
@@ -0,0 +1,541 @@
...
+  public patch(): void {
+    BaseOpenTelemetryComponent._logger.debug(
+      'applying patch to',
+      this.moduleName,
+      this.version
+    );
+
+    if (isWrapped(this.render)) {
+      shimmer.unwrap(this, 'render');
+      BaseOpenTelemetryComponent._logger.debug(
```
markwolff, line 416: These seem like they should be warnings. If double patching is attempted, then something probably isn't right.

examples/react-load/src/Home.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { BaseOpenTelemetryComponent } from '@opentelemetry/plugin-react-load';
+
+class Home extends BaseOpenTelemetryComponent {
```
markwolff, line 5: Is another example using [hooks/functional components](https://reactjs.org/docs/hooks-intro.html) also possible with this plugin?

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
dyladan: We already have an issue to track this on our side https://github.com/open-telemetry/opentelemetry-js/issues/1340, and a spec PR which resolved this https://github.com/open-telemetry/opentelemetry-specification/pull/722, but the simple answer is that this is not allowed:

> - The attribute key, which MUST be a non-`null` and non-empty string.
> - The attribute value, which is either:
>   - A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed > 64 bit integer.
>   - An array of primitive type values. The array MUST be homogeneous,
>     i.e. it MUST NOT contain values of different types. For protocols that do
>     not natively support array values such values SHOULD be represented as JSON strings.

dyladan:
Has this been tested with [preact](https://preactjs.com/)? Would be interested to see if it works, and would be a nice addition to the README.

plugins/web/opentelemetry-plugin-react-load/src/types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/*
+ * method "render" from React.Component
+ */
+export type RenderFunction = () => React.ReactNode;
+
+/*
+ * method "componentDidMount" from React.Component
+ */
+export type ComponentDidMountFunction = (() => void) | undefined;
```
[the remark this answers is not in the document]
thgao: I had to add `| undefined` in order to match the function definition provided by React, otherwise when I tried to wrap a function using Shimmer it didn't compile since the types of the patch I'm supplying and actual function prototype passed in don't match

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```diff
@@ -0,0 +1,541 @@
...
+        ...args
+      ): React.ReactNode {
+        // Render is the first method in the mounting flow, if parent span wasn't named updating already then we're in the mounting flow
+        // The naming assigned here will only be reflected if the component throws and error when rendering
+        const parentData = plugin._getCreateParentSpan(this);
+        if (parentData && parentData.name === DEFAULT_PARENT_NAME) {
+          parentData.name = AttributeNames.MOUNTING_SPAN;
+        }
+        return plugin._instrumentFunction(this, 'render', () => {
+          return original!.apply(this, args);
```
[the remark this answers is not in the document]
thgao: Related to the comment below! The type of the original function can be undefined as specified by React, but in my code I guarantee it to be a no-op when undefined. This patch function has to have the same signature as the actual React method so that the shimmer wrap function can take them as arguments

plugins/web/opentelemetry-plugin-react-load/README.md
thgao: Right now it doesn't support functional components, I was hoping this was okay because functional components don't have lifecycle methods. If we wanted to add tracing to it I was thinking that could be a separate strain of work to instrument `useEffect`. I was also thinking this would be okay since functional components are typically less complicated than class components so it would likely be more important for a user to want the class component to be instrumented, what do you think?

examples/react-load/src/Home.jsx
thgao: Related to comment above, but currently my implementation doesn't support functional components

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```diff
@@ -0,0 +1,541 @@
...
+    try {
+      state = JSON.stringify(react.state);
+    } catch {
+      state = '{"message": "state could not be turned into string"}';
+    }
+    return {
+      [GeneralAttribute.COMPONENT]: this.moduleName,
+      [AttributeNames.LOCATION_URL]: window.location.href,
+      [AttributeNames.REACT_NAME]: react.constructor.name,
+      [AttributeNames.REACT_STATE]: state,
```
dyladan, line 170: Is this a requirement? I think the state can become quite large in some cases and this may cause quite some overhead.

thgao: I added in state as an attribute because I found it a good way to decipher where/when spans were coming from in an application, would it be better to not include this attribute or only conditionally include it based on length?

dyladan: A length heuristic is likely confusing for users. Maybe you could define a protected property on the base class like `getCustomSpanAttributes: undefined | (state: State) => Attributes`, which is called with state iff it is defined, and its returned attributes are attached to the span? Or a simpler property like `addStateAsSpanAttribute: boolean` might be easier... It depends what use-case you're actually trying to solve.

dyladan: If the use-case is to help you debug while building the module, I would say to strip it out before merge. If it is actually impossible to tell which component generated the span, then maybe there is some other attribute that could cover that use-case like the component's `constructor.name` property.

thgao: I see, I'm trying to solve the user side use case. I was thinking that since it's common in React to have multiple state updates in a row (for example setting setting the state to Loading, then setting it to Loaded once a request returns), these would all come from the same component so just having a name (which is an attribute right now too) might still make it hard for users to decipher the traces from each other. 
Could you clarify about your suggestion I think I'm misunderstanding it a bit, is the suggestion to still allow long states to be added as attributes but not add the state label at all when state is undefined?

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```diff
@@ -0,0 +1,541 @@
...
+      const plugin = this;
+      return function patchRender(
+        this: React.Component,
+        ...args
+      ): React.ReactNode {
+        // Render is the first method in the mounting flow, if parent span wasn't named updating already then we're in the mounting flow
+        // The naming assigned here will only be reflected if the component throws and error when rendering
+        const parentData = plugin._getCreateParentSpan(this);
+        if (parentData && parentData.name === DEFAULT_PARENT_NAME) {
+          parentData.name = AttributeNames.MOUNTING_SPAN;
```
obecny, line 208: you really don't need this `parentData` structure to save the name and setup in the end when you are ending the span. You can create the span with correct name at first name. This is valid for all patched function when you call `_getCreateParentSpan`

thgao:
> 2. React is still having some problem, although the XMLHttpRequest is getting a correct parent,
>    but as you can see you spans are splitted into 2 groups and also the click event is not a parent for start group.
>    Why those 2 groups are not connected whereas in preact it works fine
>    WIth regards to click event for me it looks like the react might either have some addition life cycle that the plugin is not taking into consideration or something else.
> 
> Finally the react should look like preact and I'm enclosing you the image how it should look like
> https://imgur.com/SpLw3nQ

After investigating a bit, I found that if the button handler is added directly using `addEventListener` like:
```
 document.getElementById('id').addEventListener('click', () => this.buttonHandler(this))
```
instead of setting the handler using the `onClick` method as part of the button, the React traces get nested under the the click span as expected.

As discussed, we think this is likely due to some React optimizations or a bug in zone.js, and out of scope of this PR.

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
obecny: @dyladan @thgao was this addressed already or there are still things that left ?

thgao: I haven't addressed this change yet, I was hoping to understand it a bit more and get some clarification first!

plugins/web/opentelemetry-plugin-react-load/README.md
panfiva: Most of components that we write today are functional components since react hooks make class-based components unnecessary in most cases. As such, instrumentation of functional components will be a big plus.

PR #165, opened 2020-07-30
chore: bump opentelemetry core dependencies

dyladan:


dyladan:
@obecny looks like the user interaction plugin is failing on the latest core

dyladan:
Looks like attributes.component is still in the assertions, but was removed from the xhr plugin in core.

obecny:
Missing: examples, lerna.json, main package.json

dyladan:
@obecny this is not the release proposal. this is pumping the dependencies

obecny:
> @obecny this is not the release proposal. this is pumping the dependencies

but in examples should we also update to latest ?

dyladan:
> > @obecny this is not the release proposal. this is pumping the dependencies
> 
> but in examples should we also update to latest ?

Done

PR #167, opened 2020-08-05
ioredis and redis DB semantic conventions

naseemkullah:
towards https://github.com/open-telemetry/opentelemetry-js-contrib/issues/116

Want to do this one first, will likely follow up with more.

N.B. Now nothing in attributes points to which specific library (e.g.: ioredis) is being traced.

PR #170, opened 2020-08-06
fix: various compilation errors

naseemkullah:

-  Unblocks https://github.com/open-telemetry/opentelemetry-js-contrib/pull/143

- Deals with other compilation errors that were encountered with `ZoneContextManager` and `Context.ROOT_CONTEXT`

- Use @types/mysql ConnectionConfig and PoolActualConfig to overcome compilation error with @types/mysql 2.15.15


naseemkullah:
@dyladan I went down a rabbit hole here, and it appears that many things were failing due to core and/or other packages being out of date. I am left with a remaining error with the `mysql` plugin tests:

```
         should ......:
     Uncaught TypeError: core_1.randomSpanId is not a function
```

Any idea what may be going on?

vmarchaud:
@naseemkullah I believe that someone implemented a interface to generate span/trace ids in the last release, it might be coming from that

dyladan:
> @naseemkullah I believe that someone implemented a interface to generate span/trace ids in the last release, it might be coming from that

This seems the most likely to me

michaelgoin:
@naseemkullah i believe they are right. will leave a comment in the code on your PR if i can but here's a related comment on a koa PR related also to express: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/144#discussion_r466645876

plugins/node/opentelemetry-plugin-mysql/package.json
```diff
@@ -46,7 +46,7 @@
     "@opentelemetry/test-utils": "^0.9.0",
     "@opentelemetry/tracing": "0.10.1",
```
michaelgoin, line 47: tracing being on 0.10.2 is what causes the `randomSpanId` issue. api/core will update to 0.10.2 on fresh installs due to being set to ^0.10.2.

looks like in related places you bumped everything to 0.10.2/^0.10.2, so i'd prob do that here as well.

naseemkullah: Ah right, thanks!

naseemkullah:
This is now passing and ready to merge, should unblock other PRs such as #143 and #167  @open-telemetry/javascript-approvers

plugins/node/opentelemetry-plugin-mysql/src/utils.ts
```diff
@@ -16,7 +16,8 @@
 
 import { Attributes } from '@opentelemetry/api';
 import { AttributeNames } from './enums';
-import { ConnectionConfig, Query } from './types';
+import { Query } from './types';
+import { ConnectionConfig, PoolActualConfig } from 'mysql';
```
dyladan, line 20:
```suggestion
import type { ConnectionConfig, PoolActualConfig } from 'mysql';
```

naseemkullah: Thanks, what's the difference?

naseemkullah: ah TIL https://www.typescriptlang.org/docs/handbook/modules.html#importing-types

vmarchaud:
@open-telemetry/javascript-maintainers I think we can merge this one so it can unblocks other PRs (ex: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/144)

PR #171, opened 2020-08-07
feat: Hapi auto-instrumentation

### LINKED ISSUE #92: Plugin for Hapi
Requested plugin for [Hapi](https://hapi.dev/) framework

Node
Weekly Downloads:
[250k](https://www.npmjs.com/package/@hapi/hapi)

carolinee21:
- Fixes #92
- Adds automatic instrumentation for Hapi, a Node.js web framework.

- This instrumentation provides automatic tracing for hapi server routes and (request lifecycle extensions)[https://github.com/hapijs/hapi/blob/master/API.md#request-lifecycle] defined either directly or via a Hapi plugin.
- Some example traces are shown here: Example Hapi Instrumentation in Zipkin

- User friendly example code is also added under [examples/hapi](https://github.com/carolinee21/opentelemetry-js-contrib/blob/hapi-plugin/examples/hapi/)
- A visual diagram of the structure of this new instrumentation is provided here:
Hapi Instrumentation (2)

- Note: this build is currently affected by the compilation errors which will be fixed via #170


carolinee21:
@dyladan @obecny If someone could please add the tag "Internship Ending Soon" tag and add a requested review from @open-telemetry/javascript-maintainers, that would be great!

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
```diff
@@ -0,0 +1,376 @@
...
+   * via a call to the @function _wrapRegisterHandler function.
+   * @param {RegisterFunction<T>} original - the original register function which
+   * registers each plugin on the server
+   */
+  private _getServerRegisterPatch<T>(
+    original: RegisterFunction<T>
+  ): RegisterFunction<T> {
+    const instrumentation = this;
+    this._logger.debug('Patching Hapi.Server register function');
+    return async function register(
```
vmarchaud, line 136: async/await can be removed

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
```diff
@@ -0,0 +1,376 @@
...
+      options?: Hapi.ServerRegisterOptions
+    ) {
+      if (Array.isArray(pluginInput)) {
+        for (let i = 0; i < pluginInput.length; i++) {
+          instrumentation._wrapRegisterHandler(pluginInput[i].plugin);
+        }
+      } else {
+        instrumentation._wrapRegisterHandler(pluginInput.plugin);
+      }
+      await original.call(this, pluginInput, options);
```
vmarchaud, line 148: we should return the result of the call

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
```diff
@@ -0,0 +1,376 @@
...
+   * accepted input into the standard Hapi `server.ext` function. For each extension,
+   * it adds instrumentation to the handler via a call to the @function _wrapExtMethods
+   * function.
+   * @param original - the original ext function which adds the extension method to the server
+   * @param {string} [pluginName] - if present, represents the name of the plugin responsible
+   * for adding this server extension. Else, signifies that the extension was added directly
+   */
+  private _getServerExtPatch(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    original: (...args: any) => void,
```
vmarchaud, line 164: should by `unknown` instead of `any` so you'll be forced to check types

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
```diff
@@ -0,0 +1,376 @@
...
+          args[0] = lifecycleEventObj;
+        }
+      } else {
+        const type = args[0];
+        if (isLifecycleExtType(type)) {
+          const method: PatchableExtMethod = args[1];
+          args[1] = instrumentation._wrapExtMethods(method, type, pluginName);
+        }
+      }
+      await original.apply(this, args);
```
vmarchaud, line 213: same as before we should return the result and we shouldn't need async/await keywords

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
```diff
@@ -0,0 +1,376 @@
...
+          route[i] = newRoute;
+        }
+      } else {
+        route = instrumentation._wrapRouteHandler.call(
+          instrumentation,
+          route,
+          pluginName
+        );
+      }
+      original.apply(this, [route]);
```
vmarchaud, line 252: return the result

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
```diff
@@ -0,0 +1,376 @@
...
+          pluginName
+        );
+      });
+      shimmer.wrap(server, 'ext', original => {
+        return instrumentation._getServerExtPatch.bind(instrumentation)(
+          original,
+          pluginName
+        );
+      });
+      return await oldHandler(server, options);
```
vmarchaud, line 282: no need for async/await keywords

plugins/node/opentelemetry-hapi-instrumentation/src/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+  AttributeNames,
+  HapiLayerType,
+  HapiLifecycleMethodNames,
+} from './types';
+
+export function getPluginName<T>(plugin: Hapi.Plugin<T>): string {
+  if ((plugin as Hapi.PluginNameVersion).name) {
+    return (plugin as Hapi.PluginNameVersion).name;
+  } else {
+    return (plugin as Hapi.PluginPackage).pkg.name;
```
vmarchaud, line 30: I believe there should be no mention of `plugin` now, should be `instrumentation` right ?

carolinee21: This is actually referring to the Hapi.Plugin object - confusing overlap in naming schemas, I know!
I will scan over the rest of my code to make sure it is not present elsewhere

carolinee21:
Hapi itself is incompatible with Node versions <12.x.x, so the builds using node8 and node10 are expected to fail. Is there a recommended way to handle this and update the CircleCI config so it just checks node12 (and node14 if possible)?

dyladan:
> Is there a recommended way to handle this and update the CircleCI config so it just checks node12 (and node14 if possible)?

This would be done directly in the `.test.ts` files.

For an example of how to skip a test you can look at mysql/pg/mongo which are skipped during normal operation. It goes something like this:

```typescript
describe('this suite only runs in node 12+', () => {
  if (!semver.satisfies(process.version, '>=12.0.0')) {
    this.skip();
  }

  it('this test is skipped in node 10', () => {});
});
```

carolinee21:
> This would be done directly in the `.test.ts` files.

>   it('this test is skipped in node 10', () => {});
> });
> ```

Okay, thank you for the response, I'll add that now!

plugins/node/opentelemetry-hapi-instrumentation/src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  HAPI_TYPE = 'hapi.type',
+  PLUGIN_NAME = 'hapi.plugin.name',
+  EXT_TYPE = 'server.ext.type',
+}
+
+export enum HapiLayerType {
+  ROUTER = 'router',
+  PLUGIN = 'plugin',
+  EXT = 'server.ext',
+}
```
markwolff, line 61: nit: prefer const object over enum

plugins/node/opentelemetry-hapi-instrumentation/src/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+export const isLifecycleExtType = (
+  variableToCheck: unknown
+): variableToCheck is Hapi.ServerRequestExtType => {
+  return (
+    typeof variableToCheck === 'string' &&
+    HapiLifecycleMethodNames.includes(variableToCheck)
```
markwolff, line 39: Is this the only place `HapiLifecycleMethodNames` is used? If so, I think it should be an object instead to not do linear array searching

dyladan:
2952602 seems to work, but is there some reason you decided to do it in the mocha config rather than in the tests like I suggested? I'm not saying your approach is worse, just interested in why you decided to do it that way.

carolinee21:
> [2952602](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/2952602e05fba73a2c86062318fd54c7d1a3d602) seems to work, but is there some reason you decided to do it in the mocha config rather than in the tests like I suggested? I'm not saying your approach is worse, just interested in why you decided to do it that way.

@dyladan Yes, there is! The build failure earlier was happening when the Hapi module was imported at the top of each test file, so it was still failing when I tried skipping the individual tests. With the mocha config, it checks the node version before loading any of the test files, so it doesn't run into the same build issues.

plugins/node/opentelemetry-hapi-instrumentation/src/hapi.ts
carolinee21: Thanks for your help with this!

plugins/node/opentelemetry-hapi-instrumentation/src/utils.ts
carolinee21: Yes, this is. Did you mean something more like defining the type as:
```
export const HapiLifecycleMethodNames = {
  onPreAuth: 'onPreAuth',
  onCredentials: 'onCredentials',
  onPostAuth: 'onPostAuth',
  onPreHandler: 'onPreHandler',
  onPostHandler: 'onPostHandler',
  onPreResponse: 'onPreResponse',
  onRequest: 'onRequest'
 };
```
and then here when checking a given variable, instead checking `variableToCheck in HapiLifecycleMethodName`?

Otherwise, I am not sure if there is a cleaner way to perform this check (there was a [similar post on stackoverflow](https://stackoverflow.com/questions/36836011/checking-validity-of-string-literal-union-type-at-runtime) which I was looking at for reference. Please let me know if there's something I'm missing!

dyladan: Yes, that is what he meant. You could also have them just be `true`. The reason is to prevent having to search through an array each time, which is slower (`O(n)`) than an object access (`O(1)`).

Another option is to use a `Set`, which also has `O(1)` existence checking.

```typescript
const hapiLifecycleMethodNames = new Set([
  'onPreAuth',
  'onCredentials',
  'onPostAuth',
  // ... and so on
]);

console.log(hapiLifecycleMethodNames.has('onPreAuth')) // true
console.log(hapiLifecycleMethodNames.has('someOtherMethod')) // false
```

carolinee21: Okay thanks, that makes sense!

PR #172, opened 2020-08-10
[Plugin User Interaction]: Improve causality of spans from bubbled events

johnbley:

- Currently when bubbled events trigger multiple listeners, we create separate spans/traces for each of them, even though they were causally related.  It also makes it hard to see whether multiple "clicks" happened or just one.

- Track event bubbling and assign parent span relationship (lowest traced handler on event bubble gets top-most span, other higher-element handlers are tracked as child spans recursively).


plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -239,11 +244,19 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
         const once = useCapture && useCapture.once;
         const patchedListener = (...args: any[]) => {
           const target = this;
+          let parentSpan: Span | undefined;
+          const event: Event = args.length > 0 ? args[0] : undefined;
```
dyladan, line 248: `args[0]` is the same as `args.length > 0 ? args[0] : undefined`. Out of bounds access simply returns `undefined`.

`const event: Event = ...` ignores the possibility of `event` being undefined.

```suggestion
          const event: Event | undefined = args[0];
```

johnbley: TIL; thanks!

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -52,6 +53,8 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     Function,
     Map<string, Map<HTMLElement, Function>>
   >();
+  // for event bubbling
+  private _event2span: WeakMap<Event, Span> = new WeakMap<Event, Span>();
```
obecny, line 57: At first I thought it was a method and I couldn't find it. Can you change the name to  something like for example `_eventsSpanMap` or `_mapOfEventsWithSpan` etc. ?

johnbley: Sure thing; changing to _eventsSpanMap.

johnbley:
> overall lgtm, I just would like to ask you to update the image `images/main-sync.jpg` to show how the trace will look like now after this change.

There's no difference in the examples because of this change - this only affects scenarios where event listeners on bubbling events are added at multiple levels of the DOM tree, which the examples don't do.

PR #175, opened 2020-08-18
fix: accept EventListener callbacks

johnbley:

- UserInteractionPlugin's add/removeEventListener patching doesn't account for the EventListener option for callbacks, documented in https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

- Add unit test showing the issue and a utility method to invoke a listener through its appropriate interface.


plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -227,6 +227,13 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     return patched;
   }
 
+  // utility method to deal with the Function|EventListener nature of addEventListener
+  private _invokeListener(listener: any, target: any, args: any[]): any {
```
dyladan, line 231: Use of any type here for listener is not ideal. If you really don't know what will be sent in, an unknown would be better which would require you to check instanceof EventEmitter or typeof === 'function' and prevent errrors.

johnbley: I wanted to type it as `Function|EventListener` but found compilation problems; I just now discovered that what I meant was `Function|EventListenerObject` so I've changed all the types to that instead.  Better?

plugins/web/opentelemetry-plugin-user-interaction/src/userInteraction.ts
```diff
@@ -227,6 +227,19 @@ export class UserInteractionPlugin extends BasePlugin<unknown> {
     return patched;
   }
 
+  // utility method to deal with the Function|EventListener nature of addEventListener
+  private _invokeListener(
+    listener: Function | EventListenerObject,
+    target: any,
```
dyladan, line 233: Shouldn't these `any` be `unknown`?

johnbley: I was just copying through the types from the parent function (`addEventListenerPatched`) - just trying to keep the PR small rather than revisit types all through the file.  Happy to expand scope or file a separate PR if you'd like to go that way.

obecny (approved):
lgtm

dyladan (approved):

johnbley:
@obecny @dyladan I'm guessing this fell through the cracks?  Would be nice to have a merged fix for this.

PR #176, opened 2020-08-19
feat: reduce root span cardinality in express plugin

gecgooden:

Fixes https://github.com/open-telemetry/opentelemetry-js/issues/897 for the express plugin. Span names should be low cardinality, and indicate that a span belongs to a class of spans. 
Spans generated by incoming requests in the http plugin create spans with names such as `GET /users/123`, where `123` is a user identifier. 
This has a high cardinality but the http plugin doesn't have the context to set a low cardinality name such as `GET /users/:id`.

When patching the request handler in express, we call `updateName` on the root span with the method of the request (eg `GET`), and the parameterized route (eg `/users/:id`) to rename it.

### DIFF
--- a/plugins/node/opentelemetry-plugin-express/src/express.ts
+++ b/plugins/node/opentelemetry-plugin-express/src/express.ts
@@ -187,6 +187,16 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
         if (plugin._tracer.getCurrentSpan() === undefined) {
           return original.apply(this, arguments);
         }
+        // Rename the root http span once we reach the request handler
+        if (
+          metadata.attributes[AttributeNames.EXPRESS_TYPE] ===
+          ExpressLayerType.REQUEST_HANDLER
+        ) {
+          const parent = plugin._tracer.getCurrentSpan();
+          if (parent) {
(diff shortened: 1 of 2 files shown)


gecgooden:
The failures are related to an missing dependency in the user interaction plugin (https://github.com/angular/angular/issues/38526). Should I fix this by pinning the version of the dependency in this PR or should it be addressed another specific fix PR?

vmarchaud:
> The failures are related to an missing dependency in the user interaction plugin ([angular/angular#38526](https://github.com/angular/angular/issues/38526)). Should I fix this by pinning the version of the dependency in this PR or should it be addressed another specific fix PR?

Not sure but i believe https://github.com/open-telemetry/opentelemetry-js-contrib/pull/175 solves the issue

vmarchaud (approved):

dyladan (approved):

dyladan:
It looks like the build fails because our plan with circleci no longer supports the machine type we were using. Were we downgraded to a free plan? Perhaps we should consider moving tests fully to github actions as other sigs have done?

dyladan:
@gecgooden the tests are failing because they are running in your personal CircleCI account. You need to update your account in circleci to follow this repo.

PR #177, opened 2020-08-20
fix: pin zone.js version and fix tsconfig reference

gecgooden:
@opentelemetry/plugin-user-interaction was failing to build due to a change in the `zone.js` dependency.

- Pin the version of the `zone.js` dependency (`0.11.1`)
- Fix the reference to `zone.js.d.ts` in `tsconfig.json` to the new path in `0.11.1`.

### DIFF
--- a/plugins/web/opentelemetry-plugin-user-interaction/package.json
+++ b/plugins/web/opentelemetry-plugin-user-interaction/package.json
@@ -86,7 +86,7 @@
     "shimmer": "^1.2.1"
   },
   "peerDependencies": {
-    "zone.js": "^0.10.3"
+    "zone.js": "0.11.1"
   },
   "sideEffects": false
 }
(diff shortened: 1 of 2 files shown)


dyladan:
> lgtm but I defer to @obecny expertise here

+1

obecny (changes requested):
To achieve that we will have to upgrade first the `opentelemetry-context-zone` and `opentelemetry-context-zone-peer-dep` in main repo, then wait until they are released and make upgrade together with changing the zone.js to newer version. Otherwise the plugin will be using a different zone then context manager which heavily depends on that. I have to block this request until we upgrade and verify the context manager is working fine on main repo.

gecgooden:
I'll close this PR for now until the package has been updated in the main repo then 👍

Issue #116, opened 2020-07-01
[pg + pg-pool] Update DB semantic conventions


dyladan:
https://github.com/open-telemetry/opentelemetry-specification/pull/575

naseemkullah:
This may fall out of scope but would it make sense to have a shared `opentelemetry-semantic-conventions` package that exports the enum of all attribute names and such? If so should this be package be hosted in the core repo?

dyladan:
There is already such a package. https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-semantic-conventions

naseemkullah:
perfect!

naseemkullah:
@dyladan re: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#mysql

Please take note of span name and db statement.

Shall we proceed like as such? Currently span name is  e.g.: `mysql:query:SELECT`, the value of db statement is that of span name and we have a custom attribute MYSQL_VALUES with the  values. To adhere to spec, `MYSQL_VALUES` would be gone, they would be templated into the sql for db statement, and the non templated sql statement would be the span name.

If we proceed as per spec, should we worry about sanitization?

dyladan:
We should follow the spec

naseemkullah:
FYI I don't plan on tackling `pg` or `pg-pool` so those are up for grabs.

PR #183, opened 2020-08-25
[mysql] implement semantic conventions

naseemkullah:

partially #116 

Using the semantic-conventions package and following specification. A function `getDbStatement` was extracted for building the statement by formatting sql with values.


plugins/node/opentelemetry-plugin-mysql/src/mysql.ts
```diff
@@ -17,21 +17,19 @@
 import { BasePlugin, isWrapped } from '@opentelemetry/core';
 import { CanonicalCode, Span, SpanKind } from '@opentelemetry/api';
 import * as mysqlTypes from 'mysql';
+import { format } from 'mysql';
```
dyladan, line 20: Do not import a concrete implementation from the plugin being loaded or you will cause a circular require. if you need something from mysql, get it from `this._moduleExports`

naseemkullah: Oh shoot thanks! TIL

naseemkullah: I seemed to have had to thread through `format` from `_patchQuery` into `getDbStatement`...WDYT?

dyladan: probably fine

plugins/node/opentelemetry-plugin-mysql/src/mysql.ts
```diff
@@ -16,22 +16,19 @@
...
   readonly supportedVersions = ['2.*'];
 
   static readonly COMPONENT = 'mysql';
-  static readonly DB_TYPE = 'sql';
 
   static readonly COMMON_ATTRIBUTES = {
-    [AttributeNames.COMPONENT]: MysqlPlugin.COMPONENT,
-    [AttributeNames.DB_TYPE]: MysqlPlugin.DB_TYPE,
-    [AttributeNames.PEER_SERVICE]: MysqlPlugin.COMPONENT,
+    [DatabaseAttribute.DB_SYSTEM]: MysqlPlugin.COMPONENT,
```
markwolff, line 31: I think this [is deprecated](https://github.com/open-telemetry/opentelemetry-specification/blob/813aa36d502006f2462eb16244f84fc459c3eb41/CHANGELOG.md#v030-02-21-2020)

dyladan: The `component` attribute is deprecated. This is this `db.system` attribute which is [alive and well])https://github.com/dynatrace-oss-contrib/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#connection-level-attributes)

naseemkullah: I think the confusion arises from `MysqlPlugin` having a `COMPONENT` property that is used as the value of the attribute as well as lower down as the `moduleName` value for the plugin.

Should we change the prop name? or get rid of it altogether since `db.system` attribute and `moduleName` are separate entities, that so happen to often share the same name but not always (e.g. redis db system for ioredis module)

markwolff: oops my mistake, misread component vs system

vmarchaud (approved):

plugins/node/opentelemetry-plugin-mysql/src/utils.ts
```diff
@@ -87,3 +68,23 @@ function getJDBCString(
...
+/**
+ * Conjures up the value for the db.statement attribute by formatting a SQL query.
+ *
+ * @returns the database statement being executed.
+ */
+export function getDbStatement(
+  query: string | Query | QueryOptions,
+  format: Function,
+  values?: any[]
+) {
```
obecny, line 81: 1. return type is missing
2. What if query is not a string and not a query, should this still return something ?

naseemkullah: Thanks for spotting this @obecny , I fixed `format`'s type, as well as the function's logic, as to always ensure a string is returned.

obecny: @naseemkullah can you please check again the point 1 - the return type of this function is still missing - line 86

naseemkullah: Oh I thought inferred type would suffice, I will go ahead an make it explicit

obecny (approved):

PR #187, opened 2020-08-28
[mongodb] implement semantic db conventions

naseemkullah:

- partially #116

- adhering to spec, worth noting: split `ns` to populate `db.name` and  `db.mongodb.collection` attributes

### DIFF
--- a/plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
+++ b/plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
@@ -16,7 +16,7 @@
 
 import { BasePlugin } from '@opentelemetry/core';
 import { CanonicalCode, Span, SpanKind } from '@opentelemetry/api';
-import * as mongodb from 'mongodb';
+import type * as mongodb from 'mongodb';
 import * as shimmer from 'shimmer';
 import {
   Func,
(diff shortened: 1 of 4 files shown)


plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
```diff
@@ -175,19 +175,21 @@ export class MongoDBPlugin extends BasePlugin<typeof mongodb> {
...
     // add database related attributes
     span.setAttributes({
-      [AttributeNames.DB_INSTANCE]: `${ns}`,
-      [AttributeNames.DB_TYPE]: this._DB_TYPE,
-      [AttributeNames.COMPONENT]: this._COMPONENT,
+      [DatabaseAttribute.DB_SYSTEM]: 'mongodb',
+      // The namespace is a combination of the database name and the name of the
+      // collection or index, like so: [database-name].[collection-or-index-name].
+      [DatabaseAttribute.DB_NAME]: ns.split('.')[0],
+      [DatabaseAttribute.DB_MONGODB_COLLECTION]: ns.split('.')[1],
```
markwolff, line 192: nit: `const [dbName, dbCollection] = ns.split('.');`

obecny (approved):

obecny:
@naseemkullah pls have a loot why is it failing - looks like might be a valid failures now ?

naseemkullah:
Thanks @obecny 

Seems like the namespace can be a [MongoDBNamespace](https://github.com/mongodb/node-mongodb-native/blob/e3cd9e684aea99be0430d856d6299e65258bb4c3/src/utils.ts#L633-L649) which is not available in definitely typed. In any case, always coercing to a string seems to have fixed the issue.

vmarchaud (approved):

Issue #107, opened 2020-06-24
plugin-express: end() called more than once on async handlers


kellycampbell:
### What version of OpenTelemetry are you using?

0.9.0

plugin-express 0.8.0

### What version of Node are you using?
12.18.1

### What did you do?

Use express.static, or an async handler like a database lookup.

Modified `examples/express/server.js` with the following:

```
diff --git a/examples/express/server.js b/examples/express/server.js
index 93cc2ca..960fe56 100644
--- a/examples/express/server.js
+++ b/examples/express/server.js
@@ -6,6 +6,7 @@ const tracer = require('./tracer')('example-express-server');
 // Require in rest of modules
 const express = require('express');
 const axios = require('axios').default;
+const path = require('path');
 
 // Setup express
 const app = express();
@@ -18,7 +19,9 @@ const getCrudController = () => {
   router.get('/', (req, res) => res.send(resources));
   router.post('/', (req, res) => {
     resources.push(req.body);
-    return res.status(201).send(req.body);
+    setTimeout(() => {
+      res.status(201).send(req.body);
+    }, 10);
   });
   return router;
 };
@@ -35,6 +38,8 @@ const authMiddleware = (req, res, next) => {
 async function setupRoutes() {
   app.use(express.json());
 
+  app.use(express.static(path.resolve(__dirname, "static")));
+
   app.get('/run_test', async (req, res) => {
     const createdCat = await axios.post(`http://localhost:${PORT}/cats`, {
       name: 'Tom',
```

### What did you expect to see?

No errors in the console.

### What did you see instead?

```
Can not execute the operation on ended Span {traceId: 9f5d0387252f1a3d06d0e65cb4b690ee, spanId: 7a4f169ce42367d6}
You can only call end() on a span once.
Can not execute the operation on ended Span {traceId: 286e8be1b14600409880d60206959991, spanId: e46cea1d0722cb41}
You can only call end() on a span once.
```

### Additional context

Related to https://github.com/open-telemetry/opentelemetry-js/pull/910

One fix that removes the symptoms but I'm not sure if it's the correct solution since it may not trace the work done in the async part of the handler is to add `spanHasEnded = true;` to the `setImmediate` at the end of `_applyPatch`:

```
  setImmediate(() => {
      if (spanHasEnded === false) {
          span.end(startTime);
+         spanHasEnded = true;
      }
  }).unref();
```

dyladan:
@vmarchaud can you take a look at this?

[...]

PR #188, opened 2020-08-29
fix(express): listen for `finish` event on response for async express layer #107

vmarchaud:
I was hesitant of implement it like this from the start (for performance reason since we are adding a listener for every *async* layer) but seeing the half-broken (timing are not accurate) spans reporting, i think it's better like this for end user.

### DIFF
--- a/plugins/node/opentelemetry-plugin-express/src/express.ts
+++ b/plugins/node/opentelemetry-plugin-express/src/express.ts
@@ -214,9 +214,9 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
         if (callbackIdx >= 0) {
           arguments[callbackIdx] = function () {
             if (spanHasEnded === false) {
-              span.end();
               spanHasEnded = true;
               req.res?.removeListener('finish', onResponseFinish);
+              span.end();
             }
             if (!(req.route && arguments[0] instanceof Error)) {
               (req[_LAYERS_STORE_PROPERTY] as string[]).pop();
--- a/plugins/node/opentelemetry-plugin-express/src/express.ts
+++ b/plugins/node/opentelemetry-plugin-express/src/express.ts
@@ -227,7 +227,7 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
         }
         const result = original.apply(this, arguments);
         /**
-         * As this point if the callback wasn't called, that means either the
+         * At this point if the callback wasn't called, that means either the
          * layer is asynchronous (so it will call the callback later on) or that
          * the layer directly end the http response, so we'll hook into the "finish"
          * event to handle the later case.
--- a/plugins/node/opentelemetry-plugin-express/src/express.ts
+++ b/plugins/node/opentelemetry-plugin-express/src/express.ts
@@ -201,6 +201,13 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
           span.end(startTime);
           spanHasEnded = true;
         }
+        // listener for response.on('finish')
+        const onResponseFinish = () => {
+          if (spanHasEnded === false) {
+            spanHasEnded = true;
+            span.end(startTime);
+          }
+        };
         // verify we have a callback
         const args = Array.from(arguments);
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -90,8 +90,10 @@ describe('Express Plugin', () => {
       });
       const router = express.Router();
       app.use('/toto', router);
-      router.get('/:id', (req, res, next) => {
-        return res.status(200).end();
+      router.get('/:id', (req, res) => {
+        setTimeout(() => {
+          return res.status(200).end();
```
obecny, line 95: Nit: `return` is not needed

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -90,8 +90,10 @@ describe('Express Plugin', () => {
...
+        }, 5);
```
obecny, line 96: out of curiosity, would be "setImmediate" enough here, or it needs to wait for something else ?

plugins/node/opentelemetry-plugin-express/src/express.ts
```diff
@@ -218,9 +218,15 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
...
         const result = original.apply(this, arguments);
-        // If the callback is never called, we need to close the span.
-        setImmediate(() => {
+        /**
+         * As this point if the callback wasn't called, that means either the
+         * layer is asynchronous (so it will call the callback later on) or that
+         * the layer directly end the http response, so we'll hook into the "finish"
+         * event to handle the later case.
+         */
+        req.res?.once('finish', () => {
```
obecny, line 227: what if error happens, will the "finish" event cover that situation too ?

obecny (approved):
lgtm, just have few question to understand the changes better

vmarchaud:
@obecny Sorry for the force push, i didn't saw your review and wanted to make a review before someone would do one :/

plugins/node/opentelemetry-plugin-express/test/express.test.ts
vmarchaud: Yeah that would be enough, do you prefer using setImmediate ? It's moslty to be sure we have done at least one eventloop tick.

obecny: both works fine, was just curious why 5. But setImmediate would then explain it more straightforward I think

plugins/node/opentelemetry-plugin-express/src/express.ts
vmarchaud: Yeah that's the `finish` event from the underlying stream. It will be run when the `res.end()` is called.

plugins/node/opentelemetry-plugin-express/test/express.test.ts
vmarchaud: EDIT: I've updated it to `setImmediate`, i would prefer for people not to think it's done to wait for something.

dyladan (approved):

vmarchaud:
This PR has been open for a while now and it contains a important bug fix, can we merge it or do we need more review ?

plugins/node/opentelemetry-plugin-express/src/express.ts
```diff
@@ -218,12 +226,13 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
...
         }
         const result = original.apply(this, arguments);
-        // If the callback is never called, we need to close the span.
-        setImmediate(() => {
-          if (spanHasEnded === false) {
-            span.end(startTime);
-          }
-        });
+        /**
+         * As this point if the callback wasn't called, that means either the
```
naseemkullah, line 230: *At this point

naseemkullah (approved):

vmarchaud:
I think we can merge this one ? cc @obecny @dyladan

PR #189, opened 2020-08-29
chore(ci): run tests on node 14

vmarchaud:
The core repo test node 14 so i believe we should do it here as well


naseemkullah:
Seems like dupe of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/23

vmarchaud:
@naseemkullah You right, for some reason i didnt see it

PR #192, opened 2020-09-01
Handful of document-load fixes

johnbley:

- use otel-web's addSpanNetworkEvents rather than a local/modified copy.  This pulls in `http.request_content_length` for free.  Adjust event generation logic accordingly.
- Rename resource fetch spans `resourceFetch` to mirror `documentFetch` and bring span names into spec compliance (no raw URLs).  I'm open to other suggestions here.
- Fix missing docload events in Firefox, which sometimes has a `fetchStart` of 0.


plugins/web/opentelemetry-plugin-document-load/src/documentLoad.ts
```diff
@@ -184,7 +172,7 @@ export class DocumentLoad extends BasePlugin<unknown> {
       keys.forEach((key: string) => {
         if (hasKey(performanceNavigationTiming, key)) {
           const value = performanceNavigationTiming[key];
-          if (typeof value === 'number' && value > 0) {
+          if (typeof value === 'number' && value >= 0) {
```
obecny, line 175: why `=` ? From what I remember when value was a 0 it simply meant like it doesn't exist / is nto available or was not measured. And also having the information that something lasted 0s might be not really useful. WDYT ?

johnbley: Specifically FireFox seems to have legit 0s in there with some frequency (for `fetchStart` possibly when loaded from cache? - not sure) and the `> 0` was preventing any doc load spans from being created.  I'm open to another rearranging of the logic to simply always report a doc load span as "best effort" based on known data, rather than keying the whole apparatus on whether or not a `fetchStart` was found.

plugins/web/opentelemetry-plugin-document-load/src/documentLoad.ts
```diff
@@ -218,13 +206,14 @@ export class DocumentLoad extends BasePlugin<unknown> {
     spanOptions: SpanOptions = {}
   ) {
     const span = this._startSpan(
-      resource.name,
+      'resourceFetch',
```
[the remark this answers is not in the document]
johnbley: Oh, of course.  Will do.

plugins/web/opentelemetry-plugin-document-load/src/documentLoad.ts
obecny: hmm fair point - it was rather to avoid of adding information that was irrelevant. I think can try to leave it then if it mean it will add some value.

obecny (approved):

vmarchaud (approved):

PR #195, opened 2020-09-02
redis cleanup

naseemkullah:

- n/a

- Removing unneeded type `RedisPluginStreamTypes`
- only import required type from redis where applicable
- prefer import type where applicable

### DIFF
--- a/plugins/node/opentelemetry-plugin-redis/src/redis.ts
+++ b/plugins/node/opentelemetry-plugin-redis/src/redis.ts
@@ -15,7 +15,7 @@
  */
 
 import { BasePlugin } from '@opentelemetry/core';
-import type * as redis from 'redis';
+import type * as redisTypes from 'redis';
 import * as shimmer from 'shimmer';
 import {
   getTracedCreateClient,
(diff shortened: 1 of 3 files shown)


vmarchaud (approved):

obecny (commented):
I would prefer to just add `type` for all imports 
```
import type * as redisTypes from 'redis'
```
without really changing redisTypes into `redis` and be consistent everywhere so avoid things like
```
import type { Callback } from 'redis';
```
but rather have 
```
import type * as redisTypes from 'redis'

// and then 
callback: redisTypes.Callback<unknown>;
```
It is much more intuitive to see something like `redisTypes.Callback` instead of `Callback` or later 
`redisTypes.RedisClient` instead of `RedisClient` mixed with `RedisPluginClientTypes` and  trying to figure out which type comes from where

naseemkullah:
> I would prefer to just add `type` for all imports
> 
> ```
> import type * as redisTypes from 'redis'
> ```
> 
> without really changing redisTypes into `redis` and be consistent everywhere so avoid things like
> 
> ```
> import type { Callback } from 'redis';
> ```
> 
> but rather have
> 
> ```
> import type * as redisTypes from 'redis'
> 
> // and then 
> callback: redisTypes.Callback<unknown>;
> ```
> 
> It is much more intuitive to see something like `redisTypes.Callback` instead of `Callback` or later
> `redisTypes.RedisClient` instead of `RedisClient` mixed with `RedisPluginClientTypes` and trying to figure out which type comes from where

Alright I've adjusted as requested.

obecny (approved):
lgtm, thx

Issue #197, opened 2020-09-04
Remove non-inclusive language


flands:
**Is your feature request related to a problem? Please describe.**

This repository contains non-inclusive language:

```
$ > git grep -Ei '(master|slave)' | grep -Evi '(tree|blob)/master' | grep plugins | grep -v githubusercontent
plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts:    } else if (command.ismaster !== undefined) {
plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts:      return MongodbCommandType.IS_MASTER;
plugins/node/opentelemetry-plugin-mongodb/src/types.ts:  ismaster: boolean;
plugins/node/opentelemetry-plugin-mongodb/src/types.ts:  IS_MASTER = 'isMaster',
```

Note: this does not address the use of `master` branch.

**Describe the solution you'd like**
Push Mongo to change non-inclusive language and then remove non-inclusive language from this repository.

**Describe alternatives you've considered**

**Additional context**

dyladan:
Have you also created an issue on the mongo side?

vmarchaud:
I think we can close this since the PR has been closed (it need modification on the mongodb nodejs driver)

PR #198, opened 2020-09-04
Remove non-inclusive language

flands:

Addresses #197

Remove non-inclusive language

### DIFF
--- a/plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
+++ b/plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
@@ -150,8 +150,8 @@ export class MongoDBPlugin extends BasePlugin<typeof mongodb> {
       return MongodbCommandType.CREATE_INDEXES;
     } else if (command.findandmodify !== undefined) {
       return MongodbCommandType.FIND_AND_MODIFY;
-    } else if (command.ismaster !== undefined) {
-      return MongodbCommandType.IS_MASTER;
+    } else if (command.isprimary !== undefined) {
+      return MongodbCommandType.IS_PRIMARY;
     } else if (command.count !== undefined) {
       return MongodbCommandType.COUNT;
     } else {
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts
```diff
@@ -150,8 +150,8 @@ export class MongoDBPlugin extends BasePlugin<typeof mongodb> {
       return MongodbCommandType.CREATE_INDEXES;
     } else if (command.findandmodify !== undefined) {
       return MongodbCommandType.FIND_AND_MODIFY;
-    } else if (command.ismaster !== undefined) {
-      return MongodbCommandType.IS_MASTER;
+    } else if (command.isprimary !== undefined) {
```
vmarchaud, line 153: The `isprimary` property doesn't exist on the command (at least in the driver's repo: https://github.com/mongodb/node-mongodb-native/search?q=isprimary&unscoped_q=isprimary), so i'm guessing renaming this will break the condition

plugins/node/opentelemetry-plugin-mongodb/src/types.ts
```diff
@@ -19,7 +19,7 @@ export type MongoInternalCommand = {
   findandmodify: boolean;
   createIndexes: boolean;
   count: boolean;
-  ismaster: boolean;
+  isprimary: boolean;
```
vmarchaud, line 22: same here, i don't think this exist in the driver implementation

plugins/node/opentelemetry-plugin-mongodb/src/types.ts
```diff
@@ -43,7 +43,7 @@ export type MongoInternalTopology = {
 export enum MongodbCommandType {
   CREATE_INDEXES = 'createIndexes',
   FIND_AND_MODIFY = 'findAndModify',
-  IS_MASTER = 'isMaster',
+  IS_PRIMARY = 'isMaster',
```
vmarchaud, line 46: we could also rename the value here at its purely internal of the instrumentation

vmarchaud (changes requested):

flands:
Yeah, this change does not seem worth it currently -- closing for now and will update issue to reflect waiting on mongo

PR #199, opened 2020-09-07
v0.10.0 proposal

obecny:
1. Upgraded core packages to `0.11.0`
2. Fixes after upgrade
3. Added lerna changelog in package.json to support `plugin_request` label 
4. Renamed package `plugins-node-all` to `plugins-node-core-and-contrib`

Draft Release: https://github.com/open-telemetry/opentelemetry-js-contrib/releases/tag/untagged-4839a22120767d5485a8


vmarchaud (approved):

dyladan (approved):

naseemkullah (approved):

markwolff (approved):

mwear (approved):

michaelgoin:
@dyladan looks like the node-plugins-all package was renamed as a part of this PR... as such, should I still do a rename of that package only to the "instrumentation" convention like we discussed before OR should I just roll that in with the full rename of everything?

Originally, I was going to do that first to fix the versioning issues per our discussions in the meeting. Seems like that probably is no-longer needed?

PR #227, opened 2020-10-21
Host Metrics

obecny:
Runtime Metrics are added on top of Host Metrics so PR is here to be able to see the differences easily:
https://github.com/obecny/opentelemetry-js-contrib/pull/1


packages/opentelemetry-rca-metrics/README.md
```diff
@@ -0,0 +1,54 @@
+#OpenTelemetry RCA Metrics for Node.js
```
dyladan, line 1: The whole package including readme assumes the user already knows what RCA means. I think more explanation is needed as to what is collected.

packages/opentelemetry-rca-metrics/src/metric.ts
```diff
@@ -0,0 +1,291 @@
...
+    this._intervalExport =
+      typeof config.interval === 'number' ? config.interval : DEFAULT_INTERVAL;
+    this._exporter = config.exporter;
+    this._logger = config.logger;
+    this._name = config.name || DEFAULT_NAME;
+    this._maxTimeoutUpdateMS =
+      config.maxTimeoutUpdateMS || DEFAULT_MAX_TIMEOUT_UPDATE_MS;
+    this._metricNameSeparator =
+      config.metricNameSeparator || DEFAULT_METRIC_NAME_SEPARATOR;
+    this._meter = new metrics.MeterProvider({
```
dyladan, line 79: Why do you not use the global (or configured) meter provider the way tracing instrumentation works? Shouldn't the global meter provider already have exporting configured?

dyladan (commented):
First pass nothing looks bad. The structure of the RCAMetrics class is a little confusing to me, but after a little reading it seems fine.

packages/opentelemetry-rca-metrics/README.md
```diff
@@ -5,7 +5,10 @@
 [![devDependencies][devDependencies-image]][devDependencies-url]
 [![Apache License][license-image]][license-url]
 
-This module provides automatic collection of RCA Metrics
+This module provides automatic collection of RCA Metrics which includes basic metrics for:
```
dyladan, line 8: I still don't know what RCA stands for. Runtime [something] Analytics?

dyladan (approved):
Seems fine to me. The implementation of `metric.ts` seems more complicated than I would expect and I would be worried implementing something this complex on top of our current metrics implementation because the sdk is so incomplete and the spec is changing so much. But in general nothing seems wrong.

packages/opentelemetry-rca-metrics/README.md
obecny: I will change RCA everywhere for `Operation`

obecny: btw RCA - root cause analysis, but the term "Operation" has been already suggested to me by someone else and I think it will be easier to understand, unless someone else have a better name for that ?

mwear: I noticed that the Python project calls these "system metrics", and go divides their system metrics package into two packages, "host metrics", and "runtime metrics". We should see if we can align our naming with other projects, whichever we decide.

obecny: it will be "host" then

dyladan: I think root cause analysis implies some actual analysis of problems. To me, operation metrics sounds like it could include host metrics and runtime metrics. Since we're only capturing cpu, disk, and memory metrics, I think "Host Metrics" is a more accurate and descriptive term. If we add a runtime metrics package later, we could compose them both into an "Operations metrics" package if we want.

packages/opentelemetry-host-metrics/README.md
```diff
@@ -0,0 +1,61 @@
+#OpenTelemetry RCA Metrics for Node.js
```
mwear, line 1:
```suggestion
#OpenTelemetry Host Metrics for Node.js
```

mwear (approved):
LGTM. One question about the split between host and runtime metrics. Will it be possible to collect runtime metrics without collecting host metrics? I think this is use case that we'd like to support. The reason is that the OTel collector can generate host metrics, so some people will only want to generate runtime metrics from their applications.

packages/opentelemetry-host-metrics/README.md
obecny: yes the runtime metrics do not generate host metrics.

PR #235, opened 2020-10-29
WIP feat: runtime metrics

obecny:
Runtime metrics


packages/opentelemetry-runtime-metrics/src/metric.ts
```diff
@@ -0,0 +1,294 @@
...
+import { getStats } from './stats/native';
+
+import * as types from './types';
+
+/**
+ * Metrics Collector - collects metrics for CPU, Memory, Heap, Network, Event
+ * Loop, Garbage Collector, Heap Space
+ * the default label name for metric name is "name"
+ */
+export class RuntimeMetrics extends HostMetrics {
```
mwear, line 34: At first glance, it looks like we'll inherit the recording of HostMetrics, but after a closer look this doesn't appear to be the case, since we override the `_createMetrics()` function. Would it make more sense to have a common base class with shared functionality that HostMetrics and Runtime metrics can subclass? For example, `class RuntimeMetrics extends MetricsBase {}` and `class HostMetrics extends MetricsBase {}`.

packages/opentelemetry-runtime-metrics/src/metric.ts
```diff
@@ -0,0 +1,294 @@
...
+  private _memRuntimeValueObserver: ValueObserverWithObservations | undefined;
+  private _heapValueObserver: ValueObserverWithObservations | undefined;
+  private _procesUptimeValueObserver: ValueObserverWithObservations | undefined;
```
mwear, line 37:
```suggestion
  private _processUptimeValueObserver: ValueObserverWithObservations | undefined;
```

packages/opentelemetry-runtime-metrics/src/metric.ts
```diff
@@ -0,0 +1,294 @@
...
+  private _createHeapValueObserver() {
+    this._heapValueObserver = this._createValueObserver(
+      enums.METRIC_NAMES.HEAP,
+      Object.values(enums.HEAP_LABELS),
+      'Heap Data'
+    );
+  }
+
+  // PROCESS
+  private _createProcesUptimeValueObserver() {
```
mwear, line 62:
```suggestion
  private _createProcessUptimeValueObserver() {
```

mwear (approved):
LGTM. Just a couple of comments.

packages/opentelemetry-runtime-metrics/src/metric.ts
obecny: done

.circleci/config.yml
```diff
@@ -74,7 +78,7 @@ node_unit_tests: &node_unit_tests
           echo "CIRCLE_NODE_VERSION=${CIRCLE_NODE_VERSION}"
     - restore_cache:
         keys:
-          - npm-cache-01-{{ .Environment.CIRCLE_JOB }}-{{ checksum "/tmp/checksums.txt" }}-F267A71D
```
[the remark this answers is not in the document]
obecny: I have fixed also `checksums.txt` adding more stuff there. `F267A71D` is just a static value always the same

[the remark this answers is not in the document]
dyladan: Yes it is a static value. It is a random static value on purpose. You can change it to break the cache without changing dependencies to trigger a scratch build.

packages/opentelemetry-runtime-metrics/src/metric.ts
mwear: Thanks @obecny 👀 great!

dyladan (approved):

PR #241, opened 2020-11-03
fix(ioredis): set `net.peer.name` attribute according to spec

### LINKED ISSUE #240: ioredis plugin publish attribute `net.peer.host` instead of `net.peer.name`
Please answer these questions before submitting a bug report.

v0.12.0

v12.18.3

ioredis plugin is setting attribute `net.peer.host` while in the database specification it states that it should set `net.peer.name`
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#connection-level-attributes

blumamir:

fixes #240 

replace attribute `net.peer.host` with `net.peer.name` so it follows the specification [Semantic conventions for database client calls](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#connection-level-attributes)


naseemkullah:
Should we parse the host to see if it is an IP and if so set NET_PEER_IP instead?

naseemkullah (approved):

blumamir:
> Should we parse the host to see if it is an IP and if so set NET_PEER_IP instead?

In my opinion it makes sense this way, because we populate the attribute from what the user configured, and not extracting IP address from the socket itself like the HTTP plugin is doing.
But I can see how it also makes sense to use the NET_PEER_IP.
If you think NET_PEER_IP is preferable I'll change it.

naseemkullah:
What I meant is that a host set by the user can either be a name or an IP. If we choose one PEER NAME is good, but ideally we check if the host is an IP and if so we set pEER IP but if it's a name we set PEER NAME

naseemkullah:
If you do choose to implement this check the following SO post might help. If not it's fine this lgtm since it's already an improvement 
https://stackoverflow.com/questions/4460586/javascript-regular-expression-to-check-for-ip-addresses

naseemkullah:
In reality the spec should probably just have PEER HOST

vmarchaud (approved):

obecny (approved):

PR #273, opened 2020-12-04
feat: enable root span route instrumentation without any express layer spans 

shyimo:
### Which problem is this PR solving?
---
In addition to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/176 PR, We should able to instrument the root span name when ignoring all express layers (router, middleware and request_handler).
This significantly reduces the amount of spans for a large and complex express based micro-service with large amount of outgoing http requests for a single incoming http request and still able to set the route on the root span. 

### Short description of the changes
---
using express with only request_handler layer before the change:
Screen Shot 2020-12-04 at 11 25 16

using express without request_handler, middleware and route layer (root span name not modified) before the change:
Screen Shot 2020-12-04 at 11 28 59

using express without request_handler, middleware and route layer after the change:
Screen Shot 2020-12-04 at 11 30 19


plugins/node/opentelemetry-plugin-express/src/express.ts
```diff
@@ -180,23 +180,33 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
...
+        // once we reach the request handler
         if (
           metadata.attributes[AttributeNames.EXPRESS_TYPE] ===
           ExpressLayerType.REQUEST_HANDLER
         ) {
           const parent = plugin._tracer.getCurrentSpan();
-          if (parent) {
-            parent.updateName(`${req.method} ${route}`);
+          // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+          // @ts-ignore
```
dyladan, line 192: Why is this ts ignore here?

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -102,6 +102,8 @@ describe('Express Plugin', () => {
       await tracer.withSpan(rootSpan, async () => {
         await httpRequest.get(`http://localhost:${port}/toto/tata`);
         rootSpan.end();
+        // @ts-ignore
```
dyladan, line 105: Why ts ignore?

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -204,6 +206,65 @@ describe('Express Plugin', () => {
...
+      });
+      const server = http.createServer(app);
+      await new Promise(resolve => server.listen(0, resolve));
+
+      const port = (server.address() as AddressInfo).port;
+      assert.strictEqual(memoryExporter.getFinishedSpans().length, 0);
+      await tracer.withSpan(rootSpan, async () => {
+        await httpRequest.get(`http://localhost:${port}/toto/tata`);
+        rootSpan.end();
+        // @ts-ignore
```
dyladan, line 244: ts ignore?

dyladan (commented):
This is a very different behavior but it doesn't look like you added any configuration options. Am I missing something?

shyimo:
Hi @dyladan.
Thats true, i didn't add any new configuration to the plugin.
The change only verify that will modify the root span name with the current request path (for example: HTTP POST -> POST user/:userid) when not using the request_handler express layer.
currently, the root span name is modified only when request_handler layer is on and this adds many tiny spans that sometimes is not needed

plugins/node/opentelemetry-plugin-express/src/express.ts
shyimo: 'name' Property does not exist on the original type 'Span'. the name is added later on (in http instrumentation i think).
If i will not use ts-ignore annotation i will not pass the linter verification

plugins/node/opentelemetry-plugin-express/test/express.test.ts
shyimo: same as above

plugins/node/opentelemetry-plugin-express/test/express.test.ts
shyimo: same as above

plugins/node/opentelemetry-plugin-express/src/express.ts
obecny: You create a new interface that extends api.Span and adds just optional name, we can't have `@ts-ignore` in here.

plugins/node/opentelemetry-plugin-express/src/express.ts
```diff
@@ -180,23 +180,33 @@ export class ExpressPlugin extends BasePlugin<typeof express> {
...
+          const parentSpanName = parent && parent.name;
+          if (parent && parentSpanName) {
```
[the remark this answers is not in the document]
shyimo: @obecny looks good. fixed

plugins/node/opentelemetry-plugin-express/src/express.ts
shyimo: @obecny no problem. done

dyladan (approved):

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -204,6 +206,63 @@ describe('Express Plugin', () => {
       });
       server.close();
     });
+    it('should ignore all ExpressLayerType based on config. root span name should be modified when route exists', async () => {
```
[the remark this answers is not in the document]
shyimo: @obecny sure. on it

shyimo:
@obecny Hi. i fix your comments on the unit test structure

obecny (approved):
lgtm, thx for those changes

dyladan (approved):

Issue #258, opened 2020-11-09
plugin-express: EventEmitter leak


johanneswuerbach:
Please answer these questions before submitting a bug report.

### What version of OpenTelemetry are you using?

```
@opentelemetry/plugin-express: 0.11.0
```

### What version of Node are you using?

```
14.15.0
```

### What did you do?
If possible, provide a recipe for reproducing the error.

Upgrade`@opentelemetry/plugin-express` from `0.10.0` to `0.11.0` 

### What did you expect to see?

Nothing

### What did you see instead?

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 finish listeners added to [ServerResponse]. Use emitter.setMaxListeners() to increase limit
```

### Additional context

Downgrading the plugin resolved the issue for us, this was likely introduced in https://github.com/open-telemetry/opentelemetry-js-contrib/commit/ed0edf8f7fe29adb43f622600bff9c4e224e96cf

vmarchaud:
Well it isn't a leak we need to attach to the end of the response to end the spans. I'm not sure how we could fix that warning without increasing to arbitrary value :/

johanneswuerbach:
This https://github.com/open-telemetry/opentelemetry-js-contrib/pull/278 fixes the issue in our code base.

PR #278, opened 2020-12-11
fix(express): listener leak

johanneswuerbach:

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/258

When a middleware is calling the callback synchronous inside the middleware the span was already ended and the added `finish` listener below was never removed.

### DIFF
--- a/plugins/node/opentelemetry-plugin-express/test/express.test.ts
+++ b/plugins/node/opentelemetry-plugin-express/test/express.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { context } from '@opentelemetry/api';
+import { context, Span, Tracer } from '@opentelemetry/api';
 import { NoopLogger } from '@opentelemetry/core';
 import { NodeTracerProvider } from '@opentelemetry/node';
 import { AsyncHooksContextManager } from '@opentelemetry/context-async-hooks';
@@ -53,6 +53,37 @@ const httpRequest = {
   },
 };
 
+const serverWithMiddleware = async (
+  tracer: Tracer,
+  rootSpan: Span,
+  addMiddlewares: (app: express.Express) => void
+): Promise<http.Server> => {
+  const app = express();
+  if (tracer) {
+    app.use((req, res, next) => tracer.withSpan(rootSpan, next));
+  }
+
+  app.use(express.json());
+
+  addMiddlewares(app);
+
+  const router = express.Router();
+  app.use('/toto', router);
+  router.get('/:id', (req, res) => {
+    setImmediate(() => {
+      res.status(200).end(req.params.id);
+    });
+  });
+  const server = http.createServer(app);
+  await new Promise<void>(resolve =>
+    server.listen(0, () => {
+      resolve();
+    })
+  );
+
+  return server;
+};
+
 describe('Express Plugin', () => {
   const logger = new NoopLogger();
   const provider = new NodeTracerProvider();
@@ -82,31 +113,37 @@ describe('Express Plugin', () => {
       const app = express();
       app.use((req, res, next) => tracer.withSpan(rootSpan, next));
       app.use(express.json());
-      // eslint-disable-next-line prefer-arrow-callback
-      app.use(function customMiddleware(req, res, next) {
+      const customMiddleware: express.RequestHandler = (req, res, next) => {
         for (let i = 0; i < 1000000; i++) {
           continue;
         }
         return next();
-      });
+      };
+      app.use(customMiddleware);
       const router = express.Router();
       app.use('/toto', router);
       let finishListenerCount: number | undefined;
-      router.get('/:id', (req, res) => {
-        setImmediate(() => {
-          finishListenerCount = req.res?.listenerCount('finish');
-          res.status(200).end();
+      const server = await serverWithMiddleware(tracer, rootSpan, app => {
+        app.use((req, res, next) => {
+          res.on('finish', () => {
+            finishListenerCount = res.listenerCount('finish');
+          });
+          next();
         });
+        for (let index = 0; index < 15; index++) {
+          app.use(customMiddleware);
+        }
       });
-      const server = http.createServer(app);
-      await new Promise(resolve => server.listen(0, resolve));
       const port = (server.address() as AddressInfo).port;
       assert.strictEqual(memoryExporter.getFinishedSpans().length, 0);
       await tracer.withSpan(rootSpan, async () => {
-        await httpRequest.get(`http://localhost:${port}/toto/tata`);
+        const response = await httpRequest.get(
+          `http://localhost:${port}/toto/tata`
+        );
+        assert.strictEqual(response, 'tata');
         rootSpan.end();
         assert.strictEqual(rootSpan.name, 'GET /toto/:id');
-        assert.strictEqual(finishListenerCount, 1);
+        assert.strictEqual(finishListenerCount, 2);
         assert.notStrictEqual(
           memoryExporter
             .getFinishedSpans()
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -102,6 +104,10 @@ describe('Express Plugin', () => {
       await tracer.withSpan(rootSpan, async () => {
         await httpRequest.get(`http://localhost:${port}/toto/tata`);
         rootSpan.end();
+        assert.strictEqual(
+          finishListenerCount,
+          1
+        );
```
johanneswuerbach, line 110: I'm not sure about this test, it fails without the fix, but it feels somewhat brittle. 

You can replicate the reported warning by registering the middleware above 11 times, which will print `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 finish listeners added to [ServerResponse]. Use emitter.setMaxListeners() to increase limit`.

vmarchaud (approved):

naseemkullah (approved):

plugins/node/opentelemetry-plugin-express/test/express.test.ts
vmarchaud: Well it just make sure we are only adding a specific number of listener, one for each async middleware. 
Your changes only listen for `finish` for async middleware whereas it was for all middlewares before (my mistake) so if the user have more than 10 async middleware, it will still generate that warning.

johanneswuerbach: In this case wouldn't it be better to just use one listener and keep track of started spans? In case the listeners fires we could iterate through the list and close everything, which isn't closed yet.

vmarchaud: Yeah i agree that it would be much better

johanneswuerbach: After re-reading the code, I don't think we have an issue here.

According to the documentation under https://expressjs.com/en/guide/using-middleware.html:

> If the current middleware function does not end the request-response cycle, it must call next() to pass control to the next middleware function. Otherwise, the request will be left hanging.

As the `finish` listener is removed when calling `next()` and middlewares are called in serial, I don't think we can end up with more than one otel `finish` listener in this case.

vmarchaud: There are different possibilities:
1. sync middleware calling `next`
2. sync middleware ending the response
3. async middleware calling `next`
4. async middleware ending the reponse

With your fix, we correctly handle 1, 2 and 4. However when the middleware is doing async work before calling `next`, `spanHasEnded` is `false` since that part is run synchronously after calling the middleware, therefore its adding a listener.

I think storing spans is the only way to do this properly :/

johanneswuerbach: Sorry for the late reply, I now added tests for all cases you described, but can't reproduce the problem in any of them (also not in 3).

johanneswuerbach: @vmarchaud could you take another look if I missed anything in those test cases? Thank you 🙇

vmarchaud: @johanneswuerbach Looks good to me

vmarchaud (approved):

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -138,6 +172,124 @@ describe('Express Plugin', () => {
...
+      const server = await serverWithMiddleware(tracer, rootSpan, app => {
+        app.use((req, res, next) => {
+          res.on('finish', () => {
+            finishListenerCount = res.listenerCount('finish');
+          });
+          next();
+        });
+        for (let index = 0; index < 15; index++) {
+          // eslint-disable-next-line prefer-arrow-callback
+          app.use(function customMiddleware(req, res, next) {
```
obecny, line 266: please create an arrow function out of the loop scope and then just use it here, instead of creating 15 new callbacks functions

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -138,6 +172,124 @@ describe('Express Plugin', () => {
...
```
obecny, line 227: the same pattern here

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -138,6 +172,124 @@ describe('Express Plugin', () => {
...
```
obecny, line 187: and here

plugins/node/opentelemetry-plugin-express/test/express.test.ts
```diff
@@ -78,30 +109,33 @@ describe('Express Plugin', () => {
...
+      const server = await serverWithMiddleware(tracer, rootSpan, app => {
+        app.use((req, res, next) => {
+          res.on('finish', () => {
+            finishListenerCount = res.listenerCount('finish');
+          });
+          next();
         });
+        for (let index = 0; index < 15; index++) {
+          // eslint-disable-next-line prefer-arrow-callback
+          app.use(function customMiddleware(req, res, next) {
```
obecny, line 122: ditto

plugins/node/opentelemetry-plugin-express/test/express.test.ts
johanneswuerbach: I’m not entirely sure how the refactored code should look like. Could you elaborate? The test intentionally creates 15 middlewares to validate we don’t see any warnings with > 11 listeners.

While the middleware function could be shared, it somehow feels less real world like as usually each middleware is a different function with different functionality.

obecny: you declare a function out of the loop and then use it inside loop.

johanneswuerbach: I get that, but then we are calling effectively the same middleware multiple times, which isn’t really like something somebody would do in a real application.

dyladan: I agree with johannes here. I don't think we need to worry too much about micro-optimizing the test code.

dyladan (approved):

johanneswuerbach:
Fixed the merge conflicts, anything else required before getting this merged?

dyladan:
@johanneswuerbach I see there are open comments from @obecny. If they are resolved please mark them as resolved. If not, please make the suggested fixes or explain your reasoning not to. We typically do not merge any PRs that have open conversations so we can ensure all concerns are addressed before merging.

johanneswuerbach:
@dyladan all four comments are about the extracting the fn used in the loop or keeping it inside, which is being discussed here https://github.com/open-telemetry/opentelemetry-js-contrib/pull/278#discussion_r552698032.

As an outside contributor I'm not really sure whether your agreement to my argument and the one +1 is resolving the issue (and the other 3) or not and as it feels to me like a decision all involved maintainers need to agree upon. As already said I'm happy yo change the code, it just feels wrong, but I'm also not the person who is maintaining the code base so it isn't my call to make :-).

obecny:
> @dyladan all four comments are about the extracting the fn used in the loop or keeping it inside, which is being discussed here [#278 (comment)](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/278#discussion_r552698032).
> 
> As an outside contributor I'm not really sure whether your agreement to my argument and the one +1 is resolving the issue (and the other 3) or not and as it feels to me like a decision all involved maintainers need to agree upon. As already said I'm happy yo change the code, it just feels wrong, but I'm also not the person who is maintaining the code base so it isn't my call to make :-).

eslint has a rule `no-loop-func` which we don't have yet unfortunately (we should imho), once we enable it this code will need to be changed - someone will have to do it. Declaring a new function in a loop is a bad practice as it creates a new function with each iteration. So this isn't even about "micro-optimizing " this is just basic principle of "good code" which is even supported by the mentioned eslint rule - I'm surprised we don't have it enable by default as it should be recommended default and we would not have this discussion as the build would fail then. Having said I think that fixing this issue should be a 1 minute job as it is pretty straight forward and it will make sure the code quality is fine - even in test :).

johanneswuerbach:
👍  updated.

obecny (approved):
lgtm, thx for changes

PR #280, opened 2020-12-13
Include client ID in pg query span

jtfell:

- Adds visibility into the pooling/client management behaviour of the application being instrumented. It can be useful to see how the clients are being re-used when an application is under load.

- Whenever a new client object is used, it is assigned an incrementing ID. This is then included in all spans.

### DIFF
--- a/plugins/node/opentelemetry-plugin-pg/src/utils.ts
+++ b/plugins/node/opentelemetry-plugin-pg/src/utils.ts
@@ -49,9 +49,18 @@ function getJDBCString(params: PgClientConnectionParams) {
   return `jdbc:postgresql://${host}:${port}/${database}`;
 }
 
+// Keep track of individual clients in the pool, increment each time a new client is used. If
+// a pool is not used then this will be different for every query that is run.
+let CLIENT_ID = 1;
+
 // Private helper function to start a span
 function pgStartSpan(tracer: Tracer, client: PgClientExtended, name: string) {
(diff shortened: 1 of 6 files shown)


plugins/node/opentelemetry-plugin-pg/src/utils.ts
```diff
@@ -49,9 +49,18 @@ function getJDBCString(params: PgClientConnectionParams) {
   return `jdbc:postgresql://${host}:${port}/${database}`;
 }
 
+// Keep track of individual clients in the pool, increment each time a new client is used. If
+// a pool is not used then this will be different for every query that is run.
```
obecny, line 53: why would someone need something like this ?

obecny (commented):
Is this requested by spec if not why is this change ?

plugins/node/opentelemetry-plugin-pg/src/utils.ts
jtfell: For visibility into the connection pooling. It can helpful to see how many different connections were used over a time period and the percentage of requests each one served.

obecny: so this should be rather converted to a metric (possible after converting pg-pool plugin to instrumentation). not the information that will be added into each span.

Rauno56:
This can soon be done with hooks. https://github.com/open-telemetry/opentelemetry-js-contrib/pull/767

PR #283, opened 2020-12-17
chore: update to OTel v0.14.0

Flarna:
Update all dependencies to 0.14.0 and adapt code to changed APIs.

Most changes are straight forward. But it was needed to adapt the opentelemetry-host-metrics/test/metric.test as it seems the data is now in second export. Don't know if this is expected.


vmarchaud (approved):

obecny (commented):
It looks fine. I have checked the metrics and here are the findings.

Please check line 118 of metric.test.ts
and change this line into 
`sandbox.clock.tick(INTERVAL * 1);`  ("1")
it will run once only , I have discovered that when it has 2 it timeouts for 1st one and thats why the values are from 2nd run
You will have to just update the values when comparing them because they will have now 1 cycle (which is correct, haven't spotted that before). and now it seems that batch is running for 1st time too.

obecny:
@Flarna if you can fix just this tests for metrics, I would like to make a release propose of contrib repo today yet ...

Flarna:
@obecny I tried that but the test still fails but now the expected values differ. e.g.
```
  1) Host Metrics
       should export CPU time metrics:

+ actual - expected

+ 0.842204
- 3.8422039999999997
```

Flarna:
adapted the values. seems this was just a side effect of the measurement interval so depending on how aggregation is done it was needed to remove 3 seconds, divide values by two or adapt to round-off errors.

obecny (approved):
lgtm

PR #299, opened 2021-01-05
chore: fixing parent span for graphql

### LINKED ISSUE #294: Many graphql.execute spans have the same parent span id
opentelemtry: `0.12.0`
node: `14.15.0`

Many `graphql.execute` spans from a given node process seem to have the same parent span id instead of the one I would expect (e.g. the koa request).

This means all the graphql operations get grouped together into a single trace giant long-lived instead of being attached to the actual trace they should be.  This makes the tracing data somewhat useless.

Example of how this looks in LightStep:

image

obecny:

- Fixes #294 

- Fixes the parent span for main spans


mwear (approved):

vmarchaud (approved):

dyladan (approved):

chrisdostert:
Awesome contribution; i've been seeing this as well. Is it possible to get a release cut?

PR #300, opened 2021-01-05
chore: fixing async resolvers for graphql

### LINKED ISSUE #295: GraphQL plugin: not waiting for promises in execute and resolve steps
Hello!

I am running into a little span/timing issue/question with the graphql plugin. From what I explored a bit, it feels like the promises in the execute and resolve steps are not `then`ed, which leads to the timings and the spans confusing me a bit.

Using the GraphQL example: https://github.com/open-telemetry/opentelemetry-js-contrib/tree/master/examples/graphql

v12.19.0

### What did you do?
I followed the example's installation and serve guidelines, went to the http://localhost:4000/graphql endpoint and played the following query:

```gql
query {
  authors {
    id
    name
    description
  }
}
```

The `author.description` field uses a Promise for its resolution, which is of interest to me for this issue 😉 

The call takes about 600ms according to my browser, most of which happens in the `execute` and also `resolve` steps. I was expecting to see a trace along the lines of:

image

In this trace, the `execute` span "includes" the resolve spans, and the timings look consistent with what my browser reports.

### What did you see instead?

The current version of the example pushes the following trace:

image

The call that issued that trace took ~350ms according to the browser, and only ~45ms according to the trace. In addition, in the trace above, it looks like some `resolve` steps happen after the `execute` step, others during.

### Additional context

It feels like the `execute` and `resolve` patches do not handle promises, making the spans look shorter than the actual execution.

The first screenshot was taken by playing a bit with the current implementation of `@opentelemetry/instrumentation-graphql`:

These lines:
- https://github.com/open-telemetry/opentelemetry-js-contrib/blob/master/plugins/node/opentelemetry-instrumentation-graphql/src/graphql.ts#L223 (for the `execute` step)
- https://github.com/open-telemetry/opentelemetry-js-contrib/blob/master/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L369 (for the `resolve` step)

were replaced by:
```ts
Promise.resolve(res)
  .then(() => endSpan(span, err)) // or field.span
  .catch((error) => endSpan(span, error)); // or field.span
```

I also found that the pg plugin handles promises that way: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/master/plugins/node/opentelemetry-plugin-pg/src/pg.ts#L136-L155, which lead me to believe the graphql plugin could/should do something similar 😇 

Let me know if you need any further information, I'll be happy to help/PR my updates.

obecny:

- Fixes #295 

- Fixing the async resolvers


mwear (approved):

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -377,3 +379,32 @@ export function wrapFieldResolver<TSource = any, TContext = any, TArgs = any>(
...
 }
+
+/**
+ * Async version of safeExecuteInTheMiddle from instrumentation package
+ * can be removed once this will be added to instrumentation package
+ * @param execute
+ * @param onFinish
+ * @param preventThrowingError
+ */
+async function safeExecuteInTheMiddleAsync<T>(
```
vmarchaud, line 390: nit but could we make a PR on the core to move this to the instrumentation package ?

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
obecny: yea will do it, see the description :)

vmarchaud: Yeah i was wondering if you planned to do it, thanks for the info

dyladan (approved):

Issue #288, opened 2020-12-21
ioredis: two spans are generated for operations in offline queue


blumamir:
### What version of OpenTelemetry are you using?
v0.13.0

### What version of Node are you using?
12

### What did you do?
If possible, provide a recipe for reproducing the error.
```js
// instrument application
const Redis = require("ioredis");
const redis = new Redis();
// call set before connection is established
redis.set('foo', 'bar');
```

### What did you expect to see?
As I am calling `set` function once, so I expect to see a single span\trace with ioredis `set` operation.

### What did you see instead?
Two different spans in two different traces with ioredis `set` operation.

### Additional context
This happens because of [ioredis offline queue](https://github.com/luin/ioredis#offline-queue).
The call to `set` creates a call to `sendCommand` which create the first span, but does not send it as [redis status is not ready yet](https://github.com/luin/ioredis/blob/master/lib/redis/index.ts#L708), it is instead being [pushed to the offline queue](https://github.com/luin/ioredis/blob/master/lib/redis/index.ts#L770).
When ioredis connect to redis, it [calls `sendCommand` again from the offline queue](https://github.com/luin/ioredis/blob/master/lib/redis/event_handler.ts#L315) which generate a second span for the same operation.

This can confuse the user, giving the impression that `set` (or any other operation) was called more times, and in a different context from what the instrumented code was actually calling.

dyladan:
Not stale afaik

PR #239, opened 2020-11-02
fix(plugin-ioredis): end span on response from the server and set span status according to response

blumamir:
## Which problem is this PR solving?

- fix: end span on response from server, so duration accurately include the server processing and networking time. Currently, the span is ended after *sending* the command to the server.
- fix: set span status to error if server response with error. Currently, span status is always set is OK, even if the server replied with an error. 

## Short description of the changes
patch command `resolve` and `reject` functions. Only when the response arrives from the server - close the span (so the duration is accurate), and set status code to error if the server responds with error.

PR #301, opened 2021-01-12
feat: add ioredis instrumentation

Flarna:
Created ioredis instrumentation because the plugin API will be removed from OTel.

fixes: #251


Flarna:
the failed test seems to be caused by reusing the redis instance. If I run test local they pass the first them but fail the second time - plugin and instrumentation.

Flarna:
plugin-ioredis test failed again because a single redis instance is used for testing serveral node versions in parallel.
`should create a child span for del` deletes the `test` key and if this happens at a bad time the `should create a child span for get promise` test fails.

As this is unrelated to this PR I don't think it should be fixed here.

Flarna:
test improvements are in #303

dyladan:
Is this a direct port of the Plugin or did you make any behavior changes?

Flarna:
It's a direct port. If there are behavior changes they are not intended.

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
```diff
@@ -0,0 +1,93 @@
...
+  typeof ioredisTypes
+> {
+  static readonly DB_SYSTEM = 'redis';
+  readonly supportedVersions = ['>1 <5'];
+
+  constructor(protected _config: IORedisInstrumentationConfig = {}) {
+    super('@opentelemetry/instrumentation-ioredis', VERSION, _config);
+  }
+
+  public setConfig(
```
dyladan, line 38: What is this function for?

Flarna: I used the grpc instrumentation as example how to port and I found it also there: https://github.com/open-telemetry/opentelemetry-js/blob/b8afb044ff3e4bef8d9e375317f2501e82d1ff1e/packages/opentelemetry-instrumentation-grpc/src/grpc.ts#L68-L72

I'm fine with removing it.

dyladan: In general I'm not sure it's a good idea to allow changing the configuration after construction.

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
```diff
@@ -0,0 +1,93 @@
...
+          this._logger.debug('Applying patch for ioredis');
+          if (isWrapped(moduleExports.prototype.sendCommand)) {
+            this._unwrap(moduleExports.prototype, 'sendCommand');
+          }
+          this._wrap(
+            moduleExports.prototype,
+            'sendCommand',
+            this._patchSendCommand()
+          );
+          if (isWrapped(moduleExports.prototype.connect)) {
```
Flarna, line 59: I'm not sure if/why this is needed. I added it because I saw it in grpc plugin: https://github.com/open-telemetry/opentelemetry-js/blob/b8afb044ff3e4bef8d9e375317f2501e82d1ff1e/packages/opentelemetry-instrumentation-grpc/src/grpc.ts#L82-L84

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
Flarna: Removed it. we can re-add it later if needed.

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
dyladan: Maybe @vmarchaud can explain the motivation there, but in general I've seen other plugins do this too. I think it's not needed, but some plugin authors have added checks like this because it feels like it would be safer. It is worth mentioning that because we use shimmer, we wouldn't be able to double wrap anyways and this behavior would be the same I think. Do the tests pass without these checks?

vmarchaud: @dyladan Pretty sure you can double wrap with shimmer (at least thats what i observed back when i migrated the http instrumentation) and thats i always added those checks. I just checked, there is just [this condition](https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L123) that should avoid this issue but thats not what i found, it may be a bug ?

obecny: you can wrap and unwrap with shimmer successfully many times as long as between wrap and unwrap nothing else will call wrap. With new instrumentation you can enable plugin, disable it, change something in config and the enable it again. I hope we could rethink again of replacing the shimmer with something I have proposed some months ago which allows safe wrapping without worrying if there was anything in the middle. And because all instrumentations are using the internal wrap methods the change will be easy.

Flarna: In the tests `isWrapped` returns always return `false` so from this point of view it could be removed.

The issue which came in my mind once seeing this was a different one: Consider someone else (e.g. some APM tool) wrapped the same API also using shimmer - in this case we would remove this tool.

Anyway, for now it seems these `if (isWrapped(` checks are used everywhere and if it turns out that this is problematic we should correct it everywhere instead of creating inconsistent instrumentations.

Another requirement for a safe unwrap/wrap sequence is that noone holds a reference to some of the old wrapped functions.

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
vmarchaud: I used this both in `http` and `grpc` instrumentation to be able to dinamically updated the config (only need a disable/enable to take effect) mostly for tests, not sure if it has real use cases in the wild though

Flarna: I see enable/disable quite risky as there is no gurantee that unwrap/wrap works as expected.

But if the config contains only stuff read once a API is called it should be ok to update without enable/disable. the more problematic part is then if a config changes happens in the middle of an transaction.

obecny: I see this as opportunity to give user full control over the plugin. You would be able to enable plugin with different setting whenever you want without the need on restarting the service. We just have to replace shimmer - (just checked just one year passed since I created https://github.com/obecny/mpwrapper as alternative). This would be great especially in web environment or on mobile as you don't want to ask user to reload the page or close and open the app again. You would be able to change settings for the plugins fully remotely - this gives great possibilities for end user, A/B testing, etc. etc. - you can't easily replace version on user phone for example. But you can have it then the same on server too.

vmarchaud: > But if the config contains only stuff read once a API is called it should be ok to update without enable/disable. the more problematic part is then if a config changes happens in the middle of an transaction.

If the user update the config and have a big enough traffic (>100req/s for ex) that will happen for sure however i'm not sure to get why it's problematic ? The end user know that he updated the config so he can expect weird traces that happened in between both state

Flarna: > however i'm not sure to get why it's problematic

For simple instrumentations like ioredis this is for sure no issue. But there might be more complex ones where several low level operations build up one span and the same configuration has to be used across these operations to get a consistent result. In these cases it's needed to pass the current config along with the operation.

If we re-add `setConfig` we should clearly document and test what is needed to get it active.
If it goes via disable/enable also the risks (e.g. disable will not unwrap if someone else wrapped after us) should be clearly state. I would prefer to do this in a followup PR.

dyladan: I agree that if we want a `setConfig` feature it should be documented and properly added to the base class in a way that can be implemented consistently (possibly a `config_changed` event or similar). For now, I don't see any problem with removing it.

vmarchaud: I'm fine with removing it from this PR too since it wasn't in place, however do we want to refactor other instrumentations to remove it too ?

plugins/node/opentelemetry-instrumentation-ioredis/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+To load a specific instrumentation (**ioredis** in this case), specify it in the Node Tracer's configuration
+
+```javascript
+const { NodeTracerProvider } = require('@opentelemetry/node');
+const { IORedisInstrumentation } = require('@opentelemetry/instrumentation-ioredis');
+
+const provider = new NodeTracerProvider({
+  // be sure to disable old plugin
+  plugins: {
```
obecny, line 33: you won't be able to load instrumentation using node plugin loader, please use `registerInstrumentations`

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
obecny: I think that we might refactor this later to unwrap automatically when user tries to wrap it - inside instrumentation package.

plugins/node/opentelemetry-instrumentation-ioredis/README.md
Flarna: I copied this over from e.g. https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-instrumentation-http/README.md#usage
It just shows that old plugin should be disabled and not how to load the instrumentation.

The instrumentation works by just creating it via `new IORedisInstrumentation()` as shown a few lines below.

vmarchaud: Yeah currently all the instrumentation on the core have this inside their readme, most users still use plugins anyway. @obecny Is the new loading system ready ? I didn't see any PR to update documentation on the core

obecny: New loading system is ready yet. I'm finishing removing PluginLoader from node tracer. If you don't want to add `registerInstrumentations` you should still show how to set a trace provider for the instrumentation and enable it so user knows how to use it - check [graphql](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/master/plugins/node/opentelemetry-instrumentation-graphql). Although the easiest and less code would be to use `registerInstrumentations` flow but it will depends on user.

Flarna: Instrumentations are enabled on default so no need to call `enable()`. Why should we teach users to call `enable()`?

I think the graphql sample is not showing how users should startup their app. First the `TracerProvider` should be created and registered and then the instrumentations should be created as they use the global `TracerProvider` at construction time. Manually setting the `TracerProvider` at several locations seems to be redundant and error prone to me.

There might be special cases where one would have several different `TracerProviders` and `ContextManagers` but this requires special care as there are a lot places in code where e.g. a tracer uses the global context even it was not created by the global `TracerProvider`. In short I would not recommend anyone to go this path.

I don't understand why `registerInstrumentations()` result in less code and is easer then just calling `new FooInstrumentation()` followed by `new BarInstrumentation()`,....

obecny: Instrumentation is not enabled by default nor the tracer provider or meter provider is set. You have to call all of that manually or you can use `registerInstrumentations` where all of that is happening automatically.

Flarna: > Instrumentation is not enabled by default

seems to be a bug then, see https://github.com/open-telemetry/opentelemetry-js/blob/27a8d4530c56257c191468b073669909bf8892a1/packages/opentelemetry-instrumentation/src/instrumentation.ts#L38-L41


>  nor the tracer provider or meter provider is set.

see https://github.com/open-telemetry/opentelemetry-js/blob/27a8d4530c56257c191468b073669909bf8892a1/packages/opentelemetry-instrumentation/src/instrumentation.ts#L44-L47
A `Tracer` using the current global `TracerProvider` is created, same for `Meter`.

obecny: hmm I have wrote that, how could I miss that omg, anyway I will fix the enable in the coming PR where I remove the PluginLoader

Flarna: Honestly speaking I find the current implementation by far better. What is the benefit of creating disabled instrumentations - just to require an extra enable step? But most likely better to move this discussion into a separate issue/PR.

obecny: `What is the benefit of creating disabled instrumentations?` sry what are you referring to ? I have already wrote that I miss that ...

Flarna: If I understood you correct you intend to change `InstrumentationAbstract` to use `enabled: false` as default.
This means every user has to enable them compared to now where they "just work".

obecny: no I didn't not I was not aware of that I have already added a default tracer and meter and simply missed the enabling plugin - as it is done by `registerIntrumentations`.

plugins/node/opentelemetry-instrumentation-ioredis/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "LICENSE",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@opentelemetry/context-async-hooks": "^0.14.0",
+    "@opentelemetry/node": "^0.14.0",
+    "@opentelemetry/semantic-conventions": "^0.14.0",
```
vmarchaud, line 49: `semantic-conventions` is a runtime deps, see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/310

vmarchaud (approved):

obecny (approved):
lgtm

Flarna:
@blumamir I merge your changes but now a test fails - but only in node 14.
My gut feeling tells that this is a race in the test and depends on the internal timing inside redis. Any hints how to improve this?

Flarna:
ok, it is a flaky test as rerun resulting in green CI. For the records [this ](https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-js-contrib/1264/workflows/bbf4e8c4-473a-47db-b949-2413fbd522ed/jobs/6080) was the failure:
```
  1) ioredis
       #send_internal_message()
           should create a child span for pubsub:
     AssertionError [ERR_ASSERTION]: ifError got unwanted exception: Expected values to be strictly equal:
+ actual - expected

+ 'subscribe'
- 'info'
      at /root/project/plugins/node/opentelemetry-plugin-ioredis/test/ioredis.test.ts:47:60
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at Context.<anonymous> (test/ioredis.test.ts:47:60)
```

dyladan:
@Flarna should we hold off merging this while you look at the flaky test?

Flarna:
I moved to other tasks meanwhile. I think this can be done via a separate PR.
I think the best solution would be to isolate the test runs for each node version and let them use a dedicated redis,... instance. But this is for sure out of scope for this PR.

vmarchaud:
I agree that it could be done in another PR, we'll also need to make other PR to add https://github.com/open-telemetry/opentelemetry-js-contrib/pull/239 (and other changes that have been suggested while reviewing).

Flarna:
changes from #239 are already integrated here via https://github.com/open-telemetry/opentelemetry-js-contrib/pull/301/commits/afde5531dac594f527041d089f7fa92b4e57bb45

blumamir:
> @blumamir I merge your changes but now a test fails - but only in node 14.
> My gut feeling tells that this is a race in the test and depends on the internal timing inside redis. Any hints how to improve this?

The test is creating two ioredis clients.
Since we are not setting the client option `enableReadyCheck` to false, the default is true, which means that following a `connect`, the client is sending an `info` command to the server.

Now we have the `pub` client which is sending connect -> info,
And we have the `sub` client which is sending connect -> info -> subscribe.
Since these happen asynchronously, we can get both:
1) connect, connect, info, info, subscribe ...
2) connect, connect, info, subscribe, info ... 

The double `subscribe` spans in the result for a single `subscribe` operation is a bug which is explained in issue #288 .

To fix the test - making it deterministic, we can do either:

1) start the clients with `enableReadyCheck: false`, which will remove the `info` span:
```js
            const pub = new ioredis(URL, {enableReadyCheck: false});
            const sub = new ioredis(URL, {enableReadyCheck: false});
```

2) create the `sub` client, let it `subscribe`, and only after we are subscribed, create the `pub` client:
```js
...
            const sub = new ioredis(URL);
            await sub.subscribe('news', 'music');
            const pub = new ioredis(URL);
            await pub.publish('news', 'Hello world!');
...
```

3) wait for the clients to connect successfully, and only then send the commands. This can be achieved by calling `connect` explicitly (after disabling auto connect) and awaiting the response:
```js
            const sub = new ioredis(URL, {lazyConnect: true});
            const pub = new ioredis(URL, {lazyConnect: true});
            await Promise.all([sub.connect(), pub.connect()]);
            await sub.subscribe('news', 'music');
            await pub.publish('news', 'Hello world!');
            await pub.publish('music', 'Hello again!');
```
There might be some other way to await on connection to the DB which I'm not aware of.


----------
Notice that some of the suggested fixes above are also fixing the double `subscribe` span bug (#288) because it syncs the `subscribe` to be called only after connection is completed, thus it's not going into the offline queue and only single-span is created for it. Those will require a fix in the test as well:
- remove one `subscribe` from the expected`spanNames` array.
- Expecting 10/11 spans instead of 11/12.
- changing the index of the `subscribe` span from 5 to 4 in the assertion below.

Hope it will get the flaky test fixed.

blumamir:
@Flarna If you want, I can fix this flaky test in a separate PR once merged, or you can fix it in your PR. whichever you prefer.

Flarna:
@blumamir Thanks for the infos! I think it's better to fix this in a separate PR to avoid that I have to touch files from plugin in this PR.

blumamir:
> @blumamir Thanks for the infos! I think it's better to fix this in a separate PR to avoid that I have to touch files from plugin in this PR.

opened an issue for that, so it won't get lost: #332

vmarchaud:
I think its good to go, /cc @dyladan are you good with merging it since you made about the flaky test ?

PR #303, opened 2021-01-13
fix: avoid races because of parallel ioredis tests

Flarna:
ioredis tests run parallel for serverl node versions using the same redis instance.

Add a random part to keys to avoid that one run impacts others.

Additionally improve lua test asserts as the script may be already cached by redis resulting in one span less.


dyladan (approved):

dyladan:
I was just working on this haha

naseemkullah (approved):

vmarchaud (approved):
:+1:

PR #318, opened 2021-01-22
feat: OpenTracing propagator

mwear:

- This adds support for the semi-official OpenTracing (aka `ot-tracer`) header format.

- This PR introduces an `OpenTracingPropagator` that handles context propagation for the `ot-tracer*` headers.
- Supports propagation of both trace context and baggage context
- Note: a partial implementation exists in the Java Contrib repo: https://github.com/open-telemetry/opentelemetry-java/blob/master/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/OtTracerPropagator.java


propagators/opentelemetry-propagator-opentracing/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^0.15.0"
+  },
+  "devDependencies": {
+    "@opentelemetry/context-base": "^0.14.0",
```
Flarna, line 45:
```suggestion
    "@opentelemetry/context-base": "^0.15.0",
```

propagators/opentelemetry-propagator-opentracing/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@opentelemetry/propagator-opentracing",
+  "version": "0.12.0",
```
Flarna, line 3:
```suggestion
  "version": "0.12.1",
```
But most likely it will move further to 0.13.0

propagators/opentelemetry-propagator-opentracing/test/OpenTracingPropagator.test.ts
```diff
@@ -0,0 +1,304 @@
...
+  INVALID_TRACEID,
+  getSpanContext,
+  setSpanContext,
+  SpanContext,
+  TraceFlags,
+  Baggage,
+  setBaggage,
+  getBaggage,
+} from '@opentelemetry/api';
+import { ROOT_CONTEXT } from '@opentelemetry/context-base';
```
Flarna, line 30: I think `ROOT_CONTEXT` is also exported by API so you could get rid of context-base

mwear: i got rid of the dependency. thanks for pointing that out!

propagators/opentelemetry-propagator-opentracing/package.json
mwear: I preemptively went to 0.13.0.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
```diff
@@ -0,0 +1,111 @@
...
+      if (Object.keys(baggage).length > 0) {
+        context = setBaggage(context, baggage);
+      }
+    }
+
+    return context;
+  }
+
+  fields(): string[] {
+    return FIELDS;
```
Flarna, line 109: The baggage keys are missing here but spec requires them.
But honestly speaking I have no idea how to return them because they seem to depend on the context which is not available here.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
```diff
@@ -0,0 +1,111 @@
...
+  TextMapPropagator,
+  TextMapSetter,
+  TraceFlags,
+} from '@opentelemetry/api';
+
+/** OT header keys */
+export const OT_TRACE_ID_HEADER = 'ot-tracer-traceid';
+export const OT_SPAN_ID_HEADER = 'ot-tracer-spanid';
+export const OT_SAMPLED_HEADER = 'ot-tracer-sampled';
+export const OT_BAGGAGE_PREFIX = 'ot-tracer-baggage-';
```
Flarna, line 37: Is there somewhere a specification for baggage in opentracing? I haven't found that much about this but at some places prefix `ot-baggage-` is used instead.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
```diff
@@ -0,0 +1,111 @@
...
+      carrier,
+      OT_SAMPLED_HEADER,
+      spanContext.traceFlags === TraceFlags.SAMPLED ? 'true' : 'false'
+    );
+
+    const baggage = getBaggage(context);
+    if (!baggage) return;
+
+    Object.entries(baggage).forEach(([k, v]) => {
+      setter.set(carrier, `${OT_BAGGAGE_PREFIX}${k}`, v.value);
```
Flarna, line 72: Should we do some checking that this results in valid HTTP header names?
What about corner cases with key which differ only in casing? Should we serialize them?

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
mwear: I don't think this is possible because they are per carrier instance. The jaeger format has the same issue. I took a look at the JS implementation of the Jaeger propagator, and it does not have baggage support. This is likely a bug and I can file an issue. The java implementation does support baggage, but does not include it in fields, see: https://github.com/open-telemetry/opentelemetry-java/blob/master/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/JaegerPropagator.java#L68.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
mwear: `ot-baggage-` is correct, thanks!

propagators/opentelemetry-propagator-opentracing/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "clean": "tsc --build --clean",
+    "test": "nyc ts-mocha -p tsconfig.json test/**/*.test.ts",
+    "codecov": "nyc report --reporter=json && codecov -f coverage/*.json -p ../../",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "eslint . --ext .ts --fix",
+    "version": "node ../../scripts/version-update.js",
+    "watch": "tsc --build --watch"
+  },
+  "keywords": [
+    "opentelemetry",
```
obecny, line 19: add keyword `opentracing` and `propagator`

propagators/opentelemetry-propagator-opentracing/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "name": "@opentelemetry/propagator-opentracing",
+  "version": "0.13.0",
+  "description": "OpenTelemetry propagator OpenTracing provides context propagation for systems that are using the OpenTracing header format",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "repository": "open-telemetry/opentelemetry-js-contrib",
+  "scripts": {
+    "compile": "tsc --build",
+    "clean": "tsc --build --clean",
+    "test": "nyc ts-mocha -p tsconfig.json test/**/*.test.ts",
```
obecny, line 11: what about browser test, should this be run also for browser ?

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
```diff
@@ -0,0 +1,111 @@
...
+  TraceFlags,
+} from '@opentelemetry/api';
+
+/** OT header keys */
+export const OT_TRACE_ID_HEADER = 'ot-tracer-traceid';
+export const OT_SPAN_ID_HEADER = 'ot-tracer-spanid';
+export const OT_SAMPLED_HEADER = 'ot-tracer-sampled';
+export const OT_BAGGAGE_PREFIX = 'ot-baggage-';
+
+const FIELDS = [OT_TRACE_ID_HEADER, OT_SPAN_ID_HEADER, OT_SAMPLED_HEADER];
```
dyladan, line 39: The multi-header baggage breaks the fields usecase. Clearing fields out of an object won't work if we don't know all the fields that were used. Saving used fields might result in a memory leak. Not sure how to resolve this. It should at least be documented.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
dyladan: We should probably bring this up with spec. The usecase for fields is broken if we don't return all used fields, but saving fields that have been used could cause a memory leak.

One possible solution would be to add the list of baggage fields used to the context returned by `inject`.

dyladan: Avoid letting the caller mutate our fields list

```suggestion
    return FIELDS.slice();
```

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
mwear: I've updated this to drop keys and values that would be invalid for HTTP transport as that could cause an HTTP request to fail. Case-sensitivity will depend on use case. HTTP should treat these as case insensitive, but a carrier can be used for other purposes. For this reason, I'm leaving the casing as-is.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
mwear: I added a comment further down above the `fields()` method.

propagators/opentelemetry-propagator-opentracing/karma.conf.js
```diff
@@ -0,0 +1,26 @@
+/*!
+ * Copyright 2020, OpenTelemetry Authors
```
Flarna, line 2:
```suggestion
 * Copyright The OpenTelemetry Authors
```

propagators/opentelemetry-propagator-opentracing/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[gitter-image]: https://badges.gitter.im/open-telemetry/opentelemetry-js.svg
+[gitter-url]: https://gitter.im/open-telemetry/opentelemetry-node?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/master/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/status.svg?path=packages/opentelemetry-propagator-opentracing
```
Flarna, line 51: Some urls should be updated to point to opentelemetry-js-contrib

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
Flarna: usually I would recommend to log dropped entries but propagators have no logger as far as I know.

Maybe add a note in readme that such entries are silently discarded.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
Flarna: I think this should be also mentioned in readme similar as the restrictions regarding traceId which are already there.

obecny: can we do opposite then, keep the names of fields that should be there and clear all the rest that are not on that list?

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
mwear: I added a note.

propagators/opentelemetry-propagator-opentracing/src/OpenTracingPropagator.ts
mwear: I don't think that we actually use the `fields` method in the API or SDK, but it is there for users or instrumentation authors to use. I documented the special case on the readme. The specification seems to hint at this possibility as well:

> Observe that some Propagators may define, besides the returned values, additional fields with variable names. To get a full list of fields for a specific carrier object, use the Keys operation.

From https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#fields

obecny (approved):
lgtm

Flarna (approved):

vmarchaud (approved):

Issue #248, opened 2020-11-05
Convert pg plugin to instrumentation 


obecny:
- Change folder name from `opentelemetry-plugin-*` to `opentelemetry-instrumentation-*`
- Change package name from `@opentelemetry/plugin-*` to `@opentelemetry/instrumentation-*`
- Change implementation to use [InstrumentationBase](https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L30)

vmarchaud:
As linked in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/321#pullrequestreview-577734674, wehn converting we could use `safeExecuteInTheMiddle` to be safer

PR #321, opened 2021-01-26
feat(plugin-pg): add post-query custom attributes hook for spans

PocketMobsters:

- There is currently no way to add custom attributes to Postgres spans.

- Add a hook `applyCustomAttributesOnSpan` that functions the same way as the HTTP Plugins


obecny:
Personally I would create a mechanism an "interceptor". This interceptor would be then called a bit earlier with information about query and span - not sure if something else is there. This way the user should have more information and can cover more different cases. The current solution allows you to add some extra attributes but I think you don't have the full picture of what was happening for example you don't have query etc.. WDYT ?

plugins/node/opentelemetry-plugin-pg/src/pg.ts
```diff
@@ -78,8 +78,14 @@ export class PostgresPlugin extends BasePlugin<typeof pgTypes> {
               query,
               params
             );
+            if (plugin._config.applyCustomAttributesOnSpan) {
+              plugin._config.applyCustomAttributesOnSpan(span, query, params);
```
obecny, line 82: would pass it as an object to avoid problems with checking params
```suggestion
              plugin._config.applyCustomAttributesOnSpan({ span, query, params });
```

PocketMobsters: Just want to clarify real quick before I make this tweak.

You mean wrapping the arguments in an object for consumption in the function, perhaps something like:

```js
export interface PostgresCustomAttributesFunction {
  (ctx: QueryContext): void;
}

export interface QueryContext {
  span: Span;
  query: string | NormalizedQueryConfig;
  params?: unknown[];
}

export interface PostgresPluginConfig extends PluginConfig {
  applyCustomAttributesOnSpan?: PostgresCustomAttributesFunction;
}
```

I think that'd be a good way to avoid positional issues with the arguments. 

I'm also curious if it makes sense to split the text query and the `NormalizedQueryConfig` into separate fields. I reckon it would make for a more intuitive way to check what kind of query was being executed than the type checking that would need to be done in my most recent changes.

```js
export interface QueryContext {
  span: Span;
  query?: string;
  config?: NormalizedQueryConfig;
  params?: unknown[];
}
```

obecny: yes 
```
export interface QueryContext {
  span: Span;
  query?: string;
  config?: NormalizedQueryConfig;
  params?: unknown[];
}
```
+1

obecny: I'm not best with naming things but perhaps the name `applyCustomAttributesOnSpan` can be changed to something more meaningful now like `interceptOnQuery` etc. ?

PocketMobsters: Perhaps the name `postQueryHook` would work? There's a bit of precedent because that's also a [function in the Http plugin](https://github.com/open-telemetry/opentelemetry-js/blob/5f4e479b2754e6dda2823999bb8419ea484ee9b2/packages/opentelemetry-plugin-http/src/types.ts#L77-L81)

obecny: `postQueryHook` +1

obecny (approved):
lgtm, thx for the changes

vmarchaud (approved):
As the http instrumentation ([here](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-instrumentation-http/src/http.ts#L324)), i think it would be better to try/catch any errors with [`safeExecuteInTheMiddle`](https://github.com/open-telemetry/opentelemetry-js/blob/8f29081f9b56c49541d0beff986f7eca3a09ba52/packages/opentelemetry-instrumentation/src/utils.ts#L24). 
Since this utils is exposed through the instrumentation package i believe its better to tackle this by implementing it when we migrate the pg plugin to the instrumentation system (as tracked in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/248)


PR is good to go for me though (i'm only posting this comment to reference it on the other issue).

PR #327, opened 2021-01-28
feat: enable root span to contain route

DinaYakovlev:

In addition to [#176](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/176) PR and to [#273](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/273) We should able to instrument the root span name. This way it would be clear from the beginning on what request the span is about, and if in the future there would be an option to remove layers (like in the express plugin) the route will be in the root span.

using koa plugin before the change:
Before
using koa plugin after  the change:
After


plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
```diff
@@ -118,7 +121,8 @@ export class KoaInstrumentation extends BasePlugin<typeof koa> {
     middlewareLayer[kLayerPatched] = true;
     this._logger.debug('patching Koa middleware layer');
     return async (context: KoaContext, next: koa.Next) => {
-      if (api.getSpan(api.context.active()) === undefined) {
+      const parent = api.getSpan(api.context.active()) as KoaPluginSpan;
```
vmarchaud, line 124: i'm not sure why you are extending the span api, can't you use [`context()`](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-api/src/trace/span.ts#L42) instead ?

DinaYakovlev: I have to find out somehow who is the root parent span.. I do it by checking that the parentSpanId is undefined..

shyimo:
@vmarchaud Hi !
any news regrading the review ?

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
vmarchaud: I remembered that we do the same in the express plugin so thats fine

vmarchaud:
@shyimo sorry for the delay, could you rebase the PR ?

DinaYakovlev:
@vmarchaud rebased, please have another look :)

vmarchaud (approved):

plugins/node/opentelemetry-koa-instrumentation/src/koa.ts
dyladan: This isn't safe. An instrumentation may be used by a third party SDK which doesn't have the same private properties we do. The safer (but more roundabout) way to check if it is the root would be to check if there is a span in context before the first span is created

shyimo: thanks @dyladan 
fixed

obecny (approved):

vmarchaud:
@DinaYakovlev Thanks for the contribution and sorry for waiting so long

PR #272, opened 2020-12-02
Migrate CircleCI unit test workflow to Github  (contrib)

shovnik:
## Which problem is this PR solving?
- This addresses [issue](https://github.com/open-telemetry/opentelemetry-js/issues/1259#event-4038172785) by migrating the only existing opentelemetry-js-contrib workflow from CircleCI to GitHub Actions (config.yml) 

## Short description of the changes
- Migrate node-unit-tests and browser-unit-tests contained in config.yml file to GitHub Actions
- Change lint workflow trigger to be on pull_request and push as well to avoid overlapping triggers
- Minor unrelated lint fix so that the lint test stops failing for every PR on the contrib repo

## Migration Plan
I suggest having CircleCI and GitHub Action jobs run in parallel for a few weeks. After the GitHub Action jobs are running fine for a week or so and then remove the CircleCI workflows from config.yml

cc - @alolita

PR #328, opened 2021-01-28
added github workflow for unit tests

willarmiros:

- https://github.com/open-telemetry/opentelemetry-js/issues/1259

Mostly just a copy of #272, with added paths for caching. I'm leaving CircleCI config in for now to allow both tests to run in parallel, but will remove them once I confirm everything passes on GH Actions. From there I believe it's up to the maintainers to remove CircleCI checks from branch protection & disable the workflow on CircleCI.

Next steps are to migrate container images test images from CircleCI to GHCR, then make a CD workflow using GH Actions which will require more discussion.


willarmiros:
Fixed coverage to only run once (on Node 12) and removed CircleCI (which caused the CI failure). This should be ready for review!

johnbley (approved):
Looks like a good port to me, nice work.

willarmiros:
Thanks @johnbley! I have no idea why the codecov is dropping on this, I thought running coverage on all Node versions instead of just Node 12 might help, but I guess not. Any ideas/does this impact merging?

vmarchaud (approved):

johnbley:
> Thanks @johnbley! I have no idea why the codecov is dropping on this, I thought running coverage on all Node versions instead of just Node 12 might help, but I guess not. Any ideas/does this impact merging?

It doesn't look like a real problem to me, possibly just a blip owing to something not lining up right.  The root checkout directories changed from`/root/project/detectors/node/` to  `/__w/` and that might be confusing the codecov report?  In any case it doesn't show any specific instances of actual decreased coverage, so I'm not particularly worried.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,131 @@
...
+            propagators/*/node_modules
+          key: ${{ runner.os }}-${{ matrix.container }}-${{ hashFiles('**/package.json') }}
+      - name: Install Root Dependencies
+        run: npm install --ignore-scripts
+      - name: Boostrap Dependencies
+        run: npx lerna bootstrap --no-ci
+      - name: Unit tests
+        run: npm run test
+      - name: Report Coverage
+        run: npm run codecov
```
dyladan, line 99: Looks like this will report coverage for each version of node, not just 14

willarmiros: Oh yeah I removed the conditional here to see if it would fix the codecov diff, I'll add it back in.

dyladan: I'm going to approve with the assumption you will fix this before it gets merged

dyladan (approved):

.github/workflows/unit-test.yml
willarmiros: Should be good now :) thanks for the catch

PR #330, opened 2021-02-01
Allow negative performance timings

mhennoch:
- As mentioned in [1769](https://github.com/open-telemetry/opentelemetry-js/pull/1769) PerformanceNavigationTiming.fetchStart can be negative by a tiny amount ( seen values -1,-2, -3 during automated testing). Currently these entries get filtered out and no spans are created. We would like to see these spans and decide what to do with them in the backend.
- Remove `value >= 0` checks from both entry sources (PerformanceNavigationTiming and legacy perf.timing ). If they are negative then nothing should break:
  - If PerformanceNavigationTiming.fetchStart is -1 then when being converted to hrTime it will just be `perf.timeOrigin - 1` which is close to the correct timestamp.
  - Same with legacy timings, although I've never seen them being negative usually just out of order (We've seen responseEnd before fetchStart for example : ) ) or 0.
- Add and refactor tests. Replace Object.defineProperty usage with stubs(patch perf.timing getter) so that test order is not important anymore.


johnbley (approved):
👍

dyladan (approved):

obecny (approved):

PR #331, opened 2021-02-02
Document Load Instrumentation

### LINKED ISSUE #243: Convert document-load plugin to instrumentation
- Change folder name from `opentelemetry-plugin-*` to `opentelemetry-instrumentation-*`
- Change package name from `@opentelemetry/plugin-*` to `@opentelemetry/instrumentation-*`
- Change implementation to use [InstrumentationBase](https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-instrumentation/src/platform/browser/instrumentation.ts#L23)

obecny:

- Fixes #243

- Convert document-load plugin to instrumentation
- added example for document-load

addons:
- result of auto `npm run lint:examples`


plugins/web/opentelemetry-instrumentation-document-load/src/documentLoad.ts
```diff
@@ -256,22 +254,23 @@ export class DocumentLoad extends BasePlugin<unknown> {
...
       window.addEventListener('load', this._onDocumentLoaded);
     }
   }
 
   /**
    * implements patch function
    */
-  protected patch() {
+  enable() {
+    this.disable();
```
vmarchaud, line 266: not sure why you are disabling the plugin when its enable, thats seems counter intuitive ? i think its worth to explain it in a comment

vmarchaud (approved):

PR #336, opened 2021-02-05
chore: update typescript to 4.x and gts to 3.x

Flarna:
Update typescript and gts to new major versions.

Additionally update eslint and plugins to latest versions and fix
some other minor issues like repository field in package.json.


johnbley (approved):
Looks great, thanks for improving the tooling on both these repos.

dyladan (approved):

PR #337, opened 2021-02-09
chore: update contrib modules

dyladan:


detectors/node/opentelemetry-resource-detector-github/package.json
```diff
@@ -60,8 +60,8 @@
     "typescript": "4.1.3"
   },
   "dependencies": {
-    "@opentelemetry/api": "^0.15.0",
-    "@opentelemetry/core": "^0.15.0",
-    "@opentelemetry/resources": "^0.15.0"
+    "@opentelemetry/api": "^0.16.0",
+    "@opentelemetry/core": "^0.16.0",
```
Flarna, line 64: it looks strange to have `"@opentelemetry/core": "^0.16.0"` here and `"@opentelemetry/core": "0.16.0"` in dev dependencies.
is there a reason why we use fixed versions in dev dependencies but not in prod dependencies?

examples/dns/package.json
```diff
@@ -26,12 +26,12 @@
...
   "dependencies": {
-    "@opentelemetry/api": "^0.15.0",
-    "@opentelemetry/exporter-jaeger": "^0.15.0",
-    "@opentelemetry/exporter-zipkin": "^0.15.0",
-    "@opentelemetry/node": "^0.15.0",
+    "@opentelemetry/api": "^0.16.0",
+    "@opentelemetry/exporter-jaeger": "^0.16.0",
+    "@opentelemetry/exporter-zipkin": "^0.16.0",
+    "@opentelemetry/node": "^0.16.0",
     "@opentelemetry/plugin-dns": "^0.13.0",
```
Flarna, line 33: Is the update to 0.14.0 a followup PR?

dyladan: I was going to cut a 0.14.0 release after this PR

detectors/node/opentelemetry-resource-detector-github/package.json
dyladan: it should only depend on resources anyways. good catch

examples/dns/package.json
dyladan: devDependencies are pinned so the CI is faster. prod dependencies are ranged so customers can receive important bugfixes without manual intervention.

detectors/node/opentelemetry-resource-detector-github/package.json
dyladan: devDependencies are pinned so the CI is faster. prod dependencies are ranged so customers can receive important bugfixes without manual intervention.

dyladan: devDependencies are pinned so the CI is faster. prod dependencies are ranged so customers can receive important bugfixes without manual intervention.

vmarchaud (approved):
Not really linked to this PR but i just remembered that both koa and hapi plugins were named with `instrumentation` even though there were based on the plugin system. It will introduces hard breaking changes for those who directly imported them

obecny (commented):
looks like many things has changed already with this

dyladan:
> looks like many things has changed already with this

yeah it's quite behind now. working on a new version

obecny:
as spoken with @dyladan closing this PR and will prepare a new one

PR #339, opened 2021-02-10
fix: graphql operation.name span attribute

davazp:

- The operation name is recorded as `query` (or `mutation`) instead of the provided name from the query.

- Uses the name as provided in the query, like NAME in:

```
query NAME {
  ...
}
```

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/graphql.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/graphql.ts
@@ -348,8 +348,8 @@ export class GraphQLInstrumentation extends InstrumentationBase {
 
     const span = this.tracer.startSpan(SpanNames.EXECUTE, {});
     if (operation) {
-      const name = (operation as graphqlTypes.OperationDefinitionNode)
-        .operation;
+      const name = (operation as graphqlTypes.OperationDefinitionNode)?.name
+        ?.value;
       if (name) {
         span.setAttribute(SpanAttributes.OPERATION, name);
       }
--- a/plugins/node/opentelemetry-instrumentation-graphql/test/graphql.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/test/graphql.test.ts
@@ -55,7 +55,7 @@ const sourceBookById = `
 `;
 
 const sourceAddBook = `
-  mutation {
+  mutation Mutation1 {
     addBook(
       name: "Fifth Book"
       authorIds: "0,2"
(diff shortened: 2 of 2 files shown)


davazp:
I did not change it in this pull request, but the "type", (that is, `query` or `mutation`) would still useful to record, though. Should we add another attribute for it?  What would be a good name?  is having:

```
operation: query
operation.name: NAME
```

a good idea? or should we flat it in another field? 

```
operation.type: query
operation.name: NAME
```

Flarna:
It's not directly related to this PR but in general I would expect that graphql instrumentation follows the specified semantic conventions as they are what backend tools will/should rely on. Adding more graphql specific attributes is still fine.

see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md

davazp:
I fixed the unit tests. But there you can see how this was intentional.

I still think `operation.name` should be what Graphql uses (the name provided in the query/mutation), but I could also expose it with another attribute name if preferred.

obecny (commented):
After your change the operation name is undefined in many cases which is wrong. If you want to log additional information you should do it using an extra argument and not replacing the existing one.

davazp:
Yes, one option would be to use

```
operation.name = query | mutation
operation.operationName.                   # new field, better naming?
```

for instance. But it is confusing. The current attribute name is not right in my opinion. I would prefer

```
operation.type = query | mutation
operation.name
```

It is indeed a breaking change, but I think it would be better to do it now before a stable release?

I am good to make the change with another attribute though if that is preferred.

obecny:
> Yes, one option would be to use
> 
> ```
> operation.name = query | mutation
> operation.operationName.                   # new field, better naming?
> ```
> 
> for instance. But it is confusing. The current attribute name is not right in my opinion. I would prefer
> 
> ```
> operation.type = query | mutation
> operation.name
> ```
> 
> It is indeed a breaking change, but I think it would be better to do it now before a stable release?
> 
> I am good to make the change with another attribute though if that is preferred.

I'm not the best with figuring out the perfect name, but if I understand correctly the new attribute will only exist if this is a query, so maybe `queryName` ?. Besides that please consider what Flarna mentioned, maybe renaming attributes to be closer to semantic conventions without figuring out the best name, WDYT ?.

davazp:
hm, mutations can also have names thought. Like

```graphql
mutation Name {
...
}
```

aitskovi:
Any update on this PR? I'd happily take it over to get it merged in. We just starting using opentelemetry and ran into the same issue.

PR #343, opened 2021-02-10
feat(instrumentation-ioredis): add response hook for custom attributes

blumamir:
This feature was originally sent in [this PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/239), and I was asked to split it into a different PR. Now that the original PR is merged, I am able to send it once again.

- feat: add responseHook to instrumentation library config so user can register hook to add custom attributes to span when successful response arrives from the server.
This is similar to the hooks functions in [instrumentation-http](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation-http#http-instrumentation-options), and to the plugins in [this repo](https://github.com/aspecto-io/opentelemetry-ext-js) (which I maintain).
Possible usage: add a custom attribute to span which tells if the command is a key HIT or key MISS.

- added the optional function to the `IORedisInstrumentationConfig`, and call the function when the command is successfully `resolve`d.
- Added unit tests for the new feature which test both success and failure.


plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -121,6 +121,12 @@ export const traceSendCommand = (
       const origResolve = cmd.resolve;
       /* eslint-disable @typescript-eslint/no-explicit-any */
       cmd.resolve = function (result: any) {
+        try {
```
vmarchaud, line 124: could you use https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-instrumentation/src/utils.ts#L24 so we are sure we can't crash user code ?

blumamir: Thanks, food point. Updated the PR

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -121,11 +122,11 @@ export const traceSendCommand = (
...
       /* eslint-disable @typescript-eslint/no-explicit-any */
       cmd.resolve = function (result: any) {
-        try {
-          config?.responseHook?.(span, cmd.name, cmd.args, result);
-        } catch {
-          // we have nothing to do with exception from hook
-        }
+        safeExecuteInTheMiddle(
+          () => config?.responseHook?.(span, cmd.name, cmd.args, result),
+          () => {},
```
vmarchaud, line 127: Sorry i didn't catch this before but i think it could be better if we could log an error so the user know that its hook is failing

plugins/node/opentelemetry-instrumentation-ioredis/src/types.ts
```diff
@@ -39,11 +40,32 @@ export type DbStatementSerializer = (
...
+ * @param response - The response object which is returned to the user who called this command.
+ *  Can be used to set custom attributes on the span
+ */
+export interface RedisResponseCustomAttributeFunction {
+  (
+    span: Span,
+    cmdName: IORedisCommand['name'],
+    cmdArgs: IORedisCommand['args'],
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    response: any
```
Flarna, line 57: Is there a better type available then `any`?
Maybe we should we use `unknown` instead to "force" the user to do some type check/cast.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
blumamir: makes sense. added logging on error :)

plugins/node/opentelemetry-instrumentation-ioredis/src/types.ts
Flarna: If it turns out that `unknown` is problematic/cumbersome to use we can still change it to `any`.
`any` to `unknonw` would be breaking for users.

blumamir: Cool so `unknown` is better in that case 👍 
Learned something new :)
Thanks for the review

vmarchaud (approved):

obecny (approved):

PR #354, opened 2021-02-20
refactor: migrate mongodb to instrumentation #250

vmarchaud:
I migrated the plugin to instrumentation class, however i also updated the patch to hook to different files (which are more up to date with latest version and more precise).
It support `mongodb` starting `3.3` which was released in the summer of 2019. I though adding back support for `< 3.3` but i think it make more sense in a seperate PR + see if supporting those old versions is actually needed.


plugins/node/opentelemetry-instrumentation-mongodb/README.md
```diff
@@ -0,0 +1,69 @@
...
+To load a specific instrumentation (**mongodb** in this case), specify it in the Node Tracer's configuration.
+
+```javascript
+const { NodeTracerProvider } = require('@opentelemetry/node');
+const { MongoDBInstrumentation } = require('@opentelemetry/instrumentation-mongodb');
+
+const provider = new NodeTracerProvider({
+  // be sure to disable old plugin
+  plugins: {
+    ioredis: { enabled: false, path: '@opentelemetry/plugin-mongodb' }
```
Flarna, line 33:
```suggestion
    mongodb: { enabled: false, path: '@opentelemetry/plugin-mongodb' }
```

plugins/node/opentelemetry-instrumentation-mongodb/README.md
```diff
@@ -0,0 +1,69 @@
...
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[gitter-image]: https://badges.gitter.im/open-telemetry/opentelemetry-js.svg
+[gitter-url]: https://gitter.im/open-telemetry/opentelemetry-node?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[license-url]: https://github.com/open-telemetry/opentelemetry-js/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/status.svg?path=packages/opentelemetry-instrumentation-mongodb
+[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js?path=packages%2Fopentelemetry-instrumentation-mongodb
+[devDependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/dev-status.svg?path=packages/opentelemetry-instrumentation-mongodb
+[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js?path=packages%2Fopentelemetry-instrumentation-mongodb&type=dev
```
Flarna, line 69:
```suggestion
[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=packages%2Fopentelemetry-instrumentation-mongodb&type=dev
```

plugins/node/opentelemetry-instrumentation-mongodb/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[gitter-image]: https://badges.gitter.im/open-telemetry/opentelemetry-js.svg
+[gitter-url]: https://gitter.im/open-telemetry/opentelemetry-node?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[license-url]: https://github.com/open-telemetry/opentelemetry-js/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/status.svg?path=packages/opentelemetry-instrumentation-mongodb
+[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js?path=packages%2Fopentelemetry-instrumentation-mongodb
+[devDependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/dev-status.svg?path=packages/opentelemetry-instrumentation-mongodb
```
Flarna, line 68:
```suggestion
[devDependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib/dev-status.svg?path=packages/opentelemetry-instrumentation-mongodb
```

plugins/node/opentelemetry-instrumentation-mongodb/README.md
```diff
@@ -0,0 +1,69 @@
...
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[gitter-image]: https://badges.gitter.im/open-telemetry/opentelemetry-js.svg
+[gitter-url]: https://gitter.im/open-telemetry/opentelemetry-node?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[license-url]: https://github.com/open-telemetry/opentelemetry-js/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/status.svg?path=packages/opentelemetry-instrumentation-mongodb
+[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js?path=packages%2Fopentelemetry-instrumentation-mongodb
```
Flarna, line 67:
```suggestion
[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=packages%2Fopentelemetry-instrumentation-mongodb
```

plugins/node/opentelemetry-instrumentation-mongodb/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[gitter-image]: https://badges.gitter.im/open-telemetry/opentelemetry-js.svg
+[gitter-url]: https://gitter.im/open-telemetry/opentelemetry-node?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[license-url]: https://github.com/open-telemetry/opentelemetry-js/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/status.svg?path=packages/opentelemetry-instrumentation-mongodb
```
Flarna, line 66:
```suggestion
[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib/status.svg?path=packages/opentelemetry-instrumentation-mongodb
```

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "LICENSE",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@opentelemetry/context-async-hooks": "^0.17.0",
+    "@opentelemetry/node": "0.17.0",
+    "@opentelemetry/tracing": "^0.17.0",
```
Flarna, line 46:
```suggestion
    "@opentelemetry/tracing": "0.17.0",
```

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "build/src/**/*.d.ts",
+    "doc",
+    "LICENSE",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@opentelemetry/context-async-hooks": "^0.17.0",
```
Flarna, line 44:
```suggestion
    "@opentelemetry/context-async-hooks": "0.17.0",
```

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,410 @@
...
+    return [
+      new InstrumentationNodeModuleDefinition<typeof mongodb>(
+        'mongodb',
+        ['>=3.3 <4'],
+        undefined,
+        undefined,
+        [
+          new InstrumentationNodeModuleFile<WireProtocolInternal>(
+            'mongodb/lib/core/wireprotocol/index.js',
+            ['>=3.3 <4'],
```
Flarna, line 65: Please move the versions array into a const as it is used at more places.

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,410 @@
...
+          }
+        );
+        instrumentation._populateAttributes(
+          span,
+          ns,
+          server,
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          operationName !== 'insert' ? (ops[0] as any) : undefined
+        );
+        const patchedCallback = instrumentation._patchEnd(span, resultHandler);
```
Flarna, line 165: Not sure here but I think this is problematic.
Callbacks are optional in mongodb APIs (see e.g. http://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#remove) and if no callback is given a promise is returned. Looks like we may create a callback here which would change behavior for users.

vmarchaud: From what i saw all the internal here are using callback-based system ([here](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/core/wireprotocol/write_command.js#L9)) so i believe the promise are handled much higher in the API.

plugins/node/opentelemetry-instrumentation-mongodb/package.json
vmarchaud: Same as express, not sure why you'd want to pin it ?

vmarchaud:
About [`6d7367b` (#354)](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/354/commits/6d7367b52f05fa3b3f56ef346eaea46101aeddeb): For some obscure reason, tests were failing if `supportedVersions` was a property, i tested with typescript 4.2.2 with the same result. I converted it to a getter which works for some reason.

dyladan:
> About [`6d7367b` (#354)](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/354/commits/6d7367b52f05fa3b3f56ef346eaea46101aeddeb): For some obscure reason, tests were failing if `supportedVersions` was a property, i tested with typescript 4.2.2 with the same result. I converted it to a getter which works for some reason.

Can you paste the output of the broken test?

vmarchaud:
> Can you paste the output of the broken test?

You can see the tests output here: https://github.com/open-telemetry/opentelemetry-js-contrib/runs/1977512692?check_suite_focus=true

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,414 @@
...
+            this._supportedVersions,
+            patch,
+            unpatch
+          ),
+        ]
+      ),
+    ];
+  }
+
+  private get _supportedVersions() {
```
dyladan, line 74: There is a built-in protected `supportedVersions` array https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-instrumentation/src/types.ts#L53

Is there a reason you chose to do it manually in the `init`?

vmarchaud: What do you mean by manually ? I believe we give it to the constructor for every instrumentation

dyladan: bah i'm sorry i got my wires crossed

obecny: in fact the `supportedVersions` might be wrong if instrumentation patches multiple packages for example pg and pg-pool, http and https, so maybe we should remove this optional param anyway and then add version only when you create a new `InstrumentationNodeModuleDefinition` the same you do with name "mongodb" ?

dyladan: i'm fine with that but it might be confusing for the average instrumentation author who isn't a JS maintainer and is looking for the simplest way to patch a single package

obecny: instead of
```
new InstrumentationNodeModuleDefinition<typeof mongodb>(
        'mongodb',
        this._supportedVersions,
        undefined,
```
do
```
new InstrumentationNodeModuleDefinition<typeof mongodb>(
        'mongodb',
        ['>=3.3 <4'],
        undefined,

```
I think that is simple enough ?

vmarchaud: @obecny I agree that we could remove it from the instrumentation interface.

> I think that is simple enough ?

I did that in the first place (and it worked) but i needed it twice so @Flarna suggested to have a single variable to avoid duplication.

obecny: yea for version we can just declare some const that not necessary sits on class directly, but the main point is that it doesnt need to be declared as `_supportedVersions`. And because it was declared as optional we could remove it easily right ?

dyladan: We can remove anything at 0.x 🤣 

I think it is a relatively painless change for what we have so far.

Flarna: I haven't really understood why this fails but to me it looks like `InstrumentationNodeModuleDefinition` should copy the array instead of storing a reference.

I tried to reproduce this locally but failed. plugin-mongodb tests run fine so it doesn't seem to be a general setup issue on my machine.

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,414 @@
...
+    const { patch, unpatch } = this._getPatches();
+    return [
+      new InstrumentationNodeModuleDefinition<typeof mongodb>(
+        'mongodb',
+        this._supportedVersions,
+        undefined,
+        undefined,
+        [
+          new InstrumentationNodeModuleFile<WireProtocolInternal>(
+            'mongodb/lib/core/wireprotocol/index.js',
```
Flarna, line 64: This doesn't work on windows because of the `/`. Using `\\` works (but most likely breaks non windows).

dyladan: Have you tested this on windows? I think this uses the node module resolver which I thought normalized paths.

vmarchaud: Even though i think its handled by node, if thats an actual bug we should fix this directly in the instrumentation base package because the mongodb plugin isn't the only one to use internals require

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
vmarchaud: After re-reading this it might comes from the fact that i was using a private property even though a readonly is exposed on the abstract class ? I will try tomorrow but anyway that should not behave and "unset" the property

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
Flarna: Yes, tested on windows.

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
Flarna: It's not caused by the name. The problem is the sequence of calling `super` in constructor and initializing the properties of the derived class. `InstrumentationBase` calls `this.init()`.
See the generated JS code then it's clear why it's `undefined`:
```js
constructor(_config = {}) {
  super('@opentelemetry/instrumentation-mongodb', version_1.VERSION, _config);
  this._config = _config;
  this._supportedVersionsX = ['>=3.3 <4'];
}
```
It's similar as in C++ - never call virtual functions in a constructor...
In contrast to C++ JS calls the correct virtual function but on an object which is not completely constructed.

dyladan: We should add windows to the testing matrix

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
vmarchaud: @Flarna Could you open an issue on the main repo with the details ? we should fix this before RC of SDK for sure

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
vmarchaud: > It's not caused by the name. The problem is the sequence of calling super in constructor and initializing the properties of the derived class. InstrumentationBase calls this.init().

Right i see, i will fix this then

Flarna: created https://github.com/open-telemetry/opentelemetry-js/issues/1989 as this seems to be an issue in instrumentation module.

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
Flarna: https://github.com/open-telemetry/opentelemetry-js/issues/1990

dyladan:
Can you document in the readme and in the package.json that this package is mac/linux only for now until the instrumentation package is updated?

vmarchaud:
> Can you document in the readme and in the package.json that this package is mac/linux only for now until the instrumentation package is updated?

I think it would be better to make a dedicated PR for that, the graphql and grpc instrumentation use internal require for sure and i'm not sure about others. WDYT ?

Flarna:
> I think it would be better to make a dedicated PR for that, the graphql and grpc instrumentation use internal require for sure and i'm not sure about others. WDYT ?

I can confirm that grpc instrumenation tests fail on windows because of the slashes issue.

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
Flarna: What about using `path.normalize('mongodb/lib/core/wireprotocol/index.js'),` for now? The turnaround time for a fix in core until it arrives in contrib is not that short.

vmarchaud: I'm fine doing that in another PR so we can do one PR for all contrib, WDYT ?

Flarna (approved):

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
obecny: I think we should make a fix in core repo when `InstrumentationNodeModuleFile` is being created us path.normalize ass @Flarna mentioned

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,412 @@
...
+  GeneralAttribute,
+} from '@opentelemetry/semantic-conventions';
+import {
+  InstrumentationBase,
+  InstrumentationNodeModuleDefinition,
+  InstrumentationNodeModuleFile,
+  isWrapped,
+} from '@opentelemetry/instrumentation';
+
+const supportedVersions = ['>=3.3 <4'];
```
obecny, line 46: Should we still support version `2.x` as previously we did. It is quite old though (older than 3 years) and what about version  >=3.0.0 <=3.2.x ?
Is this something we should really worry about or we are can drop it ?

obecny (commented):
generally looks fine, I have just one concern with dropping support for older version we did support with plugin

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
vmarchaud: I commented on the PR description that IMO it seems fine and if its requested we could re-add support, WDYT ?

obecny: sry I missed that, not a blocker

obecny (approved):

vmarchaud:
Please don't merge the PR as i need to bump api dependency to 0.18

dyladan (approved):

PR #355, opened 2021-02-21
refactor: migrate express to instrumentation #252

### LINKED ISSUE #252: Convert express plugin to instrumentation
- Change folder name from `opentelemetry-plugin-*` to `opentelemetry-instrumentation-*`
- Change package name from `@opentelemetry/plugin-*` to `@opentelemetry/instrumentation-*`
- Change implementation to use [InstrumentationBase](https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L30)

vmarchaud:
No specific changes apart the migration to the instrumentation class.

Fixes #252


plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+This module provides automatic instrumentation for [`express`](https://github.com/expressjs/express).
+
+For automatic instrumentation see the
+[@opentelemetry/node](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-node) package.
+
+## Installation
+
+```bash
+npm install --save @opentelemetry/plugin-express
```
Flarna, line 15:
```suggestion
npm install --save @opentelemetry/instrumentation-express
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+To load a specific plugin (express in this case), specify it in the Node Tracer's configuration.
+```js
+const { NodeTracerProvider } = require('@opentelemetry/node');
+
+const provider = new NodeTracerProvider({
+  plugins: {
+    express: {
+      enabled: true,
+      // You may use a package name or absolute path to the file.
+      path: '@opentelemetry/plugin-express',
```
Flarna, line 33: I think this should be changed to something like this:
```
provider.register();
new ExpressInstrumentation();
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+});
+```
+
+To load all the [supported plugins](https://github.com/open-telemetry/opentelemetry-js#plugins), use below approach. Each plugin is only loaded when the module that it patches is loaded; in other words, there is no computational overhead for listing plugins for unused modules.
```
Flarna, line 39: As far as I know express is not loaded on default. Therefore I guess that this variant is not really helpful for users.

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+const provider = new NodeTracerProvider();
+```
+
+See [examples/express](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/express) for a short example.
+
+### Caveats
+
+Because of the way express works, it's hard to correctly compute the time taken by asynchronous middlewares and request handlers. For this reason, the time you'll see reported for asynchronous middlewares and request handlers will only represent the synchronous execution time, and **not** any asynchronous work.
+
+### Express Plugin Options
```
Flarna, line 52:
```suggestion
### Express Instrumentation Options
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+Express plugin has few options available to choose from. You can set the following:
```
Flarna, line 54:
```suggestion
Express instrumentation has few options available to choose from. You can set the following:
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+- For more about OpenTelemetry JavaScript: <https://github.com/open-telemetry/opentelemetry-js>
+- For help or feedback on this project, join us on [gitter][gitter-url]
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[gitter-image]: https://badges.gitter.im/open-telemetry/opentelemetry-js.svg
+[gitter-url]: https://gitter.im/open-telemetry/opentelemetry-node?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[license-url]: https://github.com/open-telemetry/opentelemetry-js/blob/main/LICENSE
```
Flarna, line 77:
```suggestion
[license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/LICENSE
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/status.svg?path=packages/opentelemetry-plugin-express
```
Flarna, line 79:
```suggestion
[dependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib/status.svg?path=packages/opentelemetry-instrumentation-express
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js?path=packages%2Fopentelemetry-plugin-express
```
Flarna, line 80:
```suggestion
[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=packages%2Fopentelemetry-instrumentation-express
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+[devDependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js/dev-status.svg?path=packages/opentelemetry-plugin-express
```
Flarna, line 81:
```suggestion
[devDependencies-image]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib/dev-status.svg?path=packages/opentelemetry-instrumentation-express
```

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -0,0 +1,82 @@
...
+[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js?path=packages%2Fopentelemetry-plugin-express&type=dev
```
Flarna, line 82:
```suggestion
[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=packages%2Fopentelemetry-instrumentation-express&type=dev
```

plugins/node/opentelemetry-instrumentation-express/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "doc",
+    "LICENSE",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@opentelemetry/context-async-hooks": "0.17.0",
+    "@opentelemetry/node": "^0.17.0",
```
Flarna, line 46:
```suggestion
    "@opentelemetry/node": "0.17.0",
```

plugins/node/opentelemetry-instrumentation-express/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "@opentelemetry/tracing": "^0.17.0",
```
Flarna, line 47:
```suggestion
    "@opentelemetry/tracing": "0.17.0",
```

Flarna (approved):
Left some minor, editorial comments

plugins/node/opentelemetry-instrumentation-express/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "@types/express": "4.17.7",
+    "@types/mocha": "7.0.2",
+    "@types/node": "14.0.27",
+    "@types/shimmer": "1.0.1",
```
Flarna, line 51: I think this can be removed

plugins/node/opentelemetry-instrumentation-express/README.md
vmarchaud: It depend on which meta packages you used, i think `plugin-all` installed every contrib + core. I'll still remove this since we don't have auto loading for instrumentation yet

plugins/node/opentelemetry-instrumentation-express/package.json
vmarchaud: Why do we pin this version specially ?

Flarna: As far as I understood all dev-dependencies are pinned. If you don't do it some bot creates a PR to do it.

maybe @dyladan knows the background.

dyladan: This is the default behavior of the renovate bot we use. The reason to pin development dependencies is so you can be sure that every developer is using the same version which prevents things like my machine showing a linting error that another machine does not. The runtime dependencies are ranged with `^` so that we can be sure our customers receive any important security patches without us having to publish a new release.

johnbley (approved):
LGTM

plugins/node/opentelemetry-instrumentation-express/package.json
vmarchaud: I'll pin the deps tomorrow

obecny: shouldn't this be `0.15`, and if this is merged then it should be changed to 0.17 before release (0.16 will be an orphaned anyway)

vmarchaud: We cant wait for https://github.com/open-telemetry/opentelemetry-js-contrib/pull/337, i think that make more sense

obecny: the thing is that if you merge this, and then use a github directly instead of npm, this instrumentation will depend on different version then other instrumentation / plugins in repo. This will cause that they will be using 2 different version of api so you won't be able to use the plugins and instrumentation together even though they have the same version. So  that's why it is important to keep version aligned and change dependency once for all.

vmarchaud: I don't really don't get whats you are trying to explain. `0.17.0` is the latest version currently so i don't see why we shouldn't make instrumentation depend on it.

obecny: if this instrumentation depends on 0.17 and all the rest packages depends on 0.15 you won't be able to even test them together before they are released. Try to simply use now express with any other package from contrib. In mono repo all packages should depend on the same version of the core. This change 0.17 should be done once for all packages.

vmarchaud: They can't be used together since you'll need to choose an api version (either 0.15 for current plugins or 0.17 for this) but we will not publish them to npm so i don't see how thats a problem.

Anyway we can just wait for contrib to be updated to 0.17 which i believe daniel is working on before merging it.

dyladan (approved):

PR #358, opened 2021-02-25
chore: use instrumentation-dns in example

Flarna:
Port the dns example to use the instrumentation instead the plugin.

PR #362, opened 2021-02-25
Updating all packages and examples to ver 0.17 of core

obecny:


examples/dns/tracer.js
```diff
@@ -36,5 +28,18 @@ module.exports = (serviceName) => {
   // Initialize the OpenTelemetry APIs to use the NodeTracerProvider bindings
   provider.register();
 
+  registerInstrumentations({
+    instrumentations: [{
+      plugins: {
+        dns: {
+          enabled: true,
+          path: '@opentelemetry/plugin-dns',
```
Flarna, line 36: There is https://github.com/open-telemetry/opentelemetry-js-contrib/pull/358 to move to instrumentation. Would be good to integrate it.

examples/graphql/package.json
```diff
@@ -31,14 +31,15 @@
     "url": "https://github.com/open-telemetry/opentelemetry-js/issues"
   },
   "dependencies": {
-    "@opentelemetry/api": "^0.15.0",
-    "@opentelemetry/exporter-collector": "^0.15.0",
+    "@opentelemetry/api": "^0.17.0",
+    "@opentelemetry/exporter-collector": "^0.17.0",
+    "@opentelemetry/instrumentation": "^0.17.0",
     "@opentelemetry/instrumentation-graphql": "^0.13.0",
```
Flarna, line 37: We should either use a file reference here like in express or keep the old version. 0.13 will not work with 0.17
applies several examples

examples/mongodb/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "url": "https://github.com/open-telemetry/opentelemetry-js/issues"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^0.17.0",
+    "@opentelemetry/exporter-jaeger": "^0.17.0",
+    "@opentelemetry/exporter-zipkin": "^0.17.0",
+    "@opentelemetry/instrumentation": "^0.17.0",
+    "@opentelemetry/node": "^0.17.0",
+    "@opentelemetry/plugin-http": "^0.17.0",
+    "@opentelemetry/plugin-mongodb": "^0.10.0",
```
Flarna, line 37: 0.10.0 seems to be outdated

examples/mongodb/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "main": "index.js",
+  "scripts": {
+    "zipkin:server": "cross-env EXPORTER=zipkin node ./server.js",
+    "zipkin:client": "cross-env EXPORTER=zipkin node ./client.js",
+    "jaeger:server": "cross-env EXPORTER=jaeger node ./server.js",
+    "jaeger:client": "cross-env EXPORTER=jaeger node ./client.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/open-telemetry/opentelemetry-js.git"
```
Flarna, line 15:
```suggestion
    "url": "git+ssh://git@github.com/open-telemetry/opentelemetry-js-contrib.git"
```

examples/mongodb/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "keywords": [
+    "opentelemetry",
+    "mongodb",
+    "tracing"
+  ],
+  "engines": {
+    "node": ">=8"
```
Flarna, line 23:
```suggestion
    "node": ">=8.5.0"
```

examples/mongodb/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "author": "OpenTelemetry Authors",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/open-telemetry/opentelemetry-js/issues"
```
Flarna, line 28:
```suggestion
    "url": "https://github.com/open-telemetry/opentelemetry-js-contrib/issues"
```

examples/mongodb/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^0.17.0",
+    "@opentelemetry/exporter-jaeger": "^0.17.0",
+    "@opentelemetry/exporter-zipkin": "^0.17.0",
+    "@opentelemetry/instrumentation": "^0.17.0",
+    "@opentelemetry/node": "^0.17.0",
+    "@opentelemetry/plugin-http": "^0.17.0",
```
Flarna, line 36: use instrumentation instead plugin here?

examples/mongodb/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@opentelemetry/plugin-mongodb": "^0.10.0",
+    "@opentelemetry/tracing": "^0.17.0",
+    "mongodb": "^3.5.7"
+  },
+  "homepage": "https://github.com/open-telemetry/opentelemetry-js#readme",
```
Flarna, line 41:
```suggestion
  "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib#readme",
```

plugins/node/opentelemetry-instrumentation-ioredis/README.md
```diff
@@ -22,25 +22,31 @@ npm install --save @opentelemetry/instrumentation-ioredis
...
+      // see under for available configuration
+    }),
+    {
+      // be sure to disable old plugin but only if it was installed
+      plugins: {
+        ioredis: { enabled: false, path: '@opentelemetry/plugin-ioredis' }
+      },
+    }
+  ],
+  tracerProvider: provider,
```
Flarna, line 48: I would remove this option here as it is usually correct and best to let instrumentations pick up the global `TracerProvider`. I know there are special usecases where this is not intended but most users want the global and I think Readme should show the usual case and docs of instrumentation packages all the details/special cases.

This comment is for sure not blocking, so feel free to keep it as it is.

Flarna (approved):
Looks good! a lot changes, hope I haven't overlooks something.

examples/mongodb/package.json
obecny: nice findings, thx @Flarna

plugins/node/opentelemetry-instrumentation-ioredis/README.md
obecny: I was really struggling with this if either keep it or not, then I aligned this across all so user has better understanding that it can be done like this. We will have to make one more round with documentation anyway when we remove plugins, we can decide then again, but it was already many changes here. I have chosen to add it after all so that it is easy to understand that this is possible. But we might add maybe some more examples or explanation too that it can be done differently. For now this code will work for 100%

vmarchaud (approved):
lgtm

dyladan (approved):

PR #372, opened 2021-03-03
feat(instrumentation-ioredis): add requireParentSpan option to config

### LINKED ISSUE #344: ioredis: add requireParent config option

ioredis instrumentation library is not creating a trace if the redis operation has no parent.
While this behaviour makes sense for some users, it might not be the desired behaviour for other, as for example redis commands might be sent from timers and then they will not be instrumented. I want to add a new config option `requireParent` similar to the one in [instrumentation-http](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation-http#http-instrumentation-options).

Currently, HTTP instrumentation library is by default not requiring parent span, and ioredis does.
I would like to make them behave the same way, so by default all operations are instrumented, and user can choose not to instrument orphaned redis commands by setting the `requireParent` option to `true`.

Since this is a breaking change for the library, I would like to discuss it first and hear your thought about it.

blumamir:

- fixes #344
- As discussed in the issue, I've implemented this option so there is no behavior change from how instrumentation is currently operating. This PR does not normalize the `requiredParent` behavior with other plugins which have different defaults. 
- The new config option name is `requiredParentSpan`, as suggested by @obecny .

- added `requiredParentSpan` to instrumentation config, and skip instrumentation only if this config option was set to `false`.
- I have added unittests for both cases (true and false). existing tests are verifying the behavior when this config option is unset.
- I added the new config option and description to the instrumentation README.md


vmarchaud (approved):

naseemkullah (approved):

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -94,7 +94,10 @@ export const traceSendCommand = (
       return original.apply(this, arguments);
     }
     // Do not trace if there is not parent span
-    if (getSpan(context.active()) === undefined) {
+    if (
+      config?.requireParentSpan !== false &&
```
obecny, line 98: Because in description it says that it defaults to `true`
I think it should be

```suggestion
      config?.requireParentSpan === true &&
```

Then in ioredis.ts
```
const DEFAULT_CONFIG: IORedisInstrumentationConfig = {
  requireParentSpan: true,
}
```
and in ioredis contructor
```
  constructor(protected _config: IORedisInstrumentationConfig = {}) {
    super('@opentelemetry/instrumentation-ioredis', VERSION, Object.assign({}, DEFAULT_CONFIG, _config));
  }
```
WDYT ?

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -94,7 +94,10 @@ export const traceSendCommand = (
...
+      getSpan(context.active()) === undefined
+    ) {
```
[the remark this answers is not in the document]
blumamir: Done :)

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
blumamir: Done :)

naseemkullah (approved):

obecny (approved):

PR #389, opened 2021-03-17
feat: net module instrumentation

### LINKED ISSUE #291: NodeJS Net library
It would be good to instrument the TCP socket connections that are made from NodeJS applications (whether outgoing or incoming).

For eg. The plugin can wrap the [createConnection](https://nodejs.org/api/net.html#net_net_createconnection) API to create a span whenever socket connection is established

NodeJS

Since this would autoinstrument core functionality, I believe it would be commonly used

**What version of plugin are you interested in using**
Versions:

**Additional context**

- **Is there a reference you could point for the well-defined lifecycle methods**

seemk:

net module instrumentation (fixes #291)

- Instrumentation for `Socket.connect` (thus also providing instrumentation for `net.connect` and `net.createConnection`)
- In case of successful TCP connections, the local socket attributes (host, port, ip) are added.
- `timeout` event is not handled on purpose as this event is not related to `connect`. When `connect` fails with a timeout, it will emit an error instead (e.g. `EATIMEOUT`).


plugins/node/opentelemetry-instrumentation-net/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "connect",
+    "nodejs",
+    "tracing",
+    "profiling",
+    "instrumentation"
+  ],
+  "author": "OpenTelemetry Authors",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=8.0.0"
```
vmarchaud, line 32: need to be `8.5.0`

plugins/node/opentelemetry-instrumentation-net/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@opentelemetry/instrumentation-net",
+  "version": "0.1.0",
```
vmarchaud, line 3: you can match the current version of other packages

plugins/node/opentelemetry-instrumentation-net/test/instrument.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  const PORT = 42123;
+  let instrumentation: NetInstrumentation;
+  let socket: net.Socket;
+  let tcpServer: net.Server;
+
+  before(() => {
+    instrumentation = new NetInstrumentation();
+    instrumentation.setTracerProvider(provider);
+    require('net');
+    assert.strictEqual(net.Socket.prototype.connect.__wrapped, true);
```
vmarchaud, line 42: you could use `isWrapped` from `@opentelemetry/instrumentation`

plugins/node/opentelemetry-instrumentation-net/src/version.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// this is autogenerated file, see scripts/version-update.js
+export const VERSION = '0.1.0';
```
vmarchaud, line 18: same, it should match other packages version number

plugins/node/opentelemetry-instrumentation-net/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@opentelemetry/core": "0.18.0",
+    "@opentelemetry/node": "0.18.0",
+    "@opentelemetry/tracing": "0.18.0",
+    "@types/mocha": "7.0.2",
+    "@types/node": "14.0.27",
+    "@types/shimmer": "1.0.1",
```
vmarchaud, line 50: i don't think this type is used

vmarchaud (commented):
overall looks good, few nitpicks

plugins/node/opentelemetry-instrumentation-net/src/version.ts
seemk: Thanks!

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+  isWrapped,
+  safeExecuteInTheMiddle,
+} from '@opentelemetry/instrumentation';
+import { GeneralAttribute } from '@opentelemetry/semantic-conventions';
+import { Net } from './types';
+import { VERSION } from './version';
+import { platform } from 'os';
+import { Socket } from 'net';
+
+const IPC_TRANSPORT = platform() == 'win32' ? 'pipe' : 'Unix';
```
obecny, line 37:
```suggestion
const IPC_TRANSPORT = platform() === 'win32' ? 'pipe' : 'Unix';
```
please declare some const for 'pipe' and 'Unix' I would also keep names for them in upper case so `PIPE` and `UNIX`

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  private _getPatchedConnect() {
+    return (original: (...args: unknown[]) => void) => {
+      const plugin = this;
+      return function patchedConnect(this: Socket, ...args: unknown[]) {
+        const options = normalizedArgs(args);
+
+        if (!options) {
+          startGenericSpan(plugin.tracer, this);
+          return original.apply(this, args);
```
johnbley, line 79: Why not just let the error-handling code a few lines below (in `safeExecuteInTheMiddle`) also apply to this?  Something like 
```
const span = !options 
                     ? startGeneric(...) 
                     : options.path 
                            ? startIpc() 
                             : startTcp()`
```
?

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+              span.end();
+            }
+          }
+        );
+      };
+    };
+  }
+}
+
+interface NormalizedOptions {
```
obecny, line 103: please move all interfaces to `types.ts`

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+  host?: string;
+  port?: number;
+  path?: string;
+}
+
+function normalizedArgs(args: unknown[]): NormalizedOptions | null | undefined {
```
obecny, line 109: please move this function to `utils.ts`

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+
+function spanEndHandler(span: Span) {
+  return () => {
+    span.end();
+  };
+}
+
+function spanErrorHandler(span: Span) {
+  return (e: Error) => {
+    span.setStatus({
```
obecny, line 145: please add `span.recordException` so that the stack will be saved too

johnbley (approved):
Looks really good; one small question about error handling which I don't consider a blocker.

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+      message: e.message,
+    });
+  };
+}
+
+interface ListenerOpts {
+  hostAttributes?: boolean;
+}
+
+function registerListeners(
```
obecny, line 156: most of this functions heavily depends either on instrumentation properties or the property from the instrumented package, why not keeping them as private methods of `NetInstrumentation` if you already keep them in the same file ?

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,237 @@
...
+    case 'string':
+      return {
+        path: opt,
+      };
+    default:
+      return;
+  }
+}
+
+const SOCKET_EVENTS = ['connect', 'error', 'close'];
```
obecny, line 135: please convert this to some enum / const so that it can be referenced instead of adding them directly later when for example you are removing events or listening to `connect` etc. etc.

plugins/node/opentelemetry-instrumentation-net/test/instrument.test.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as Sinon from 'sinon';
+import * as net from 'net';
+
+const memoryExporter = new InMemorySpanExporter();
+const provider = new NodeTracerProvider();
+const tracer = provider.getTracer('default');
+provider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
+
+describe('NetInstrumentation', () => {
+  const PORT = 42123;
```
obecny, line 34: I would keep const outside `describe` block

plugins/node/opentelemetry-instrumentation-net/test/connect.test.ts
```diff
@@ -0,0 +1,257 @@
...
+function assertClientSpan(span: ReadableSpan) {
+  assert.strictEqual(span.kind, SpanKind.CLIENT);
+}
+
+function assertAttrib(span: ReadableSpan, attrib: string, value: any) {
+  assert.strictEqual(span.attributes[attrib], value);
+}
+
+describe('NetInstrumentation', () => {
+  const PORT = 42123;
```
obecny, line 44: can you please move const and functions outside of describe block - const to top, functions below or in separate `utils.ts` file (one or two are fine, otherwise would prefer moving them to separate file)

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: Good point, fixed it.

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: I took the attribute names directly from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes where they have the same case, so I'm not sure which way to go here

obecny: can you add then const values to general semantic conventions?.  I see there are already 3 values, you can add the one you need here
```
export const GeneralAttribute = {
  // These are used as potential values to NET_TRANSPORT
  ...
  UNIX: 'Unix',
  PIPE: 'pipe',
};

```

obecny: here -> https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-semantic-conventions/src/trace/general.ts#L35

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -0,0 +1,236 @@
...
+
+function startTcpSpan(
+  tracer: Tracer,
+  options: NormalizedOptions,
+  socket: Socket
+) {
+  const span = tracer.startSpan('tcp.connect', {
+    kind: SpanKind.CLIENT,
+    attributes: {
+      [GeneralAttribute.NET_TRANSPORT]: 'IP.TCP',
```
obecny, line 227: this value is also available in semantic conventions `GeneralAttribute. IP_TCP`

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: Yeah, sure. Thanks :slightly_smiling_face:

seemk: Although on another note, they are not really attributes, but attribute values. I guess it's fine to just declare the constants here?

seemk: https://github.com/open-telemetry/opentelemetry-js/pull/2026

obecny: I think for now you can create some const locally on top with `@TODO` note to replace them later with semantic conventions as they will be available after releasing

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: Moved the `_start*Span` methods taking a tracer argument to private methods.

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: Done!

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: Done: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/389/commits/ac287749eb686eb9aac8455866f87deb7c7487f1

Also the other issues brought up should be resolved now as well.

obecny:
thx for all changes, I'm missing two final things: 
- a `README.md` with example usage
- LICENSE - please copy it from any other instrumentation

Extra:
An image showing the span that is being added by this plugin would be a nice addition but it is not a blocker.

obecny:
btw I tried to assign you the issue you worked on but it seems you have to comment on that before I can do it

seemk:
Added a LICENSE and README.md.

For an example how this looks like in Zipkin when using an express application which is doing an outbound request: 
image


Do note that the `connect` span is attached to the wrong parent, but this is an issue with the `http` instrumentation where it does not activate the correct context. It did once, but it was removed with https://github.com/open-telemetry/opentelemetry-js/issues/1479 (PR https://github.com/open-telemetry/opentelemetry-js/pull/1546) I think it was wrong to remove it and will create PR with a fix for `http` instrumentation.

obecny (approved):
thx a lot for your contribution, this looks great!

vmarchaud (approved):

johanneswuerbach:
This looks great, would you be open to accept a contribute to also add a span for the `lookup` event https://nodejs.org/api/net.html#net_event_lookup, which is generally useful to troubleshoot connection performance as especially k8s was prone to DNS timeouts?

vmarchaud:
@johanneswuerbach shouldn't this be handled by the dns instrumentation ?

PR #394, opened 2021-03-23
fix: net instrumentation - change span kind of connect to internal

seemk:

From the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind):
* `CLIENT` Indicates that the span describes a synchronous request to some remote service. This span is the parent of a remote SERVER span and waits for its response.

However there is no corresponding server span for `connect`. Thus I think the least spec violating span kind for `connect` should be `INTERNAL`.

Change span kind of `connect` from `CLIENT` to `INTERNAL`.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-net/src/net.ts
+++ b/plugins/node/opentelemetry-instrumentation-net/src/net.ts
@@ -92,7 +92,7 @@ export class NetInstrumentation extends InstrumentationBase<Net> {
   /* It might still be useful to pick up errors due to invalid connect arguments. */
   private _startGenericSpan(socket: Socket) {
     const span = this.tracer.startSpan('connect', {
-      kind: SpanKind.CLIENT,
+      kind: SpanKind.INTERNAL,
     });
 
     registerListeners(socket, span);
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -92,7 +92,7 @@ export class NetInstrumentation extends InstrumentationBase<Net> {
   /* It might still be useful to pick up errors due to invalid connect arguments. */
   private _startGenericSpan(socket: Socket) {
     const span = this.tracer.startSpan('connect', {
-      kind: SpanKind.CLIENT,
+      kind: SpanKind.INTERNAL,
```
Flarna, line 95: `SpanKind.INTERNAL` is the default so you could omit the second argument.

Flarna (approved):

dyladan (approved):

plugins/node/opentelemetry-instrumentation-net/src/net.ts
seemk: Thanks, good point :slightly_smiling_face:  Fixed it

vmarchaud (approved):

PR #399, opened 2021-03-30
feat: add instrumentation of AWS Lambda handlers to generate FaaS spans

anuraaga:

- Allow AWS Lambda users to have spans generated for function invocations

This instrumentation patches the AWS Lambda handler (found via the _HANDLER env variable which is always set by Lambda) to generate spans around invocations. It is the first in a series of PRs implementing the spec defined here

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md

This currently only decorates the invocations with spans following FaaS conventions - future PRs will add reading of HTTP for proxy requests and messaging for SQS requests with appropriate parenting.

I actually am not sure if this approach works I will try it out with an actual lambda function, sending out for an initial look at the code and maybe suggestions on how to work around this problem. I noticed that the patching infrastructure, via `require-in-middle` can only patch core modules or modules in `node_modules` - it literally checks the path of a required file to see if it is in `node_modules` and bails if it isn't. 

https://github.com/watson/module-details-from-path/blob/master/index.js#L8

I have worked around it in testing here by copying handlers into `node_modules` - I need to check whether Lambda actually houses a user handler function inside a folder called `node_modules`. There's a reasonable chance it doesn't - if anyone can provide some tips on whether it's possible to work around this, that would be great.


anuraaga:
@willarmiros @wangzlei

anuraaga:
I found how lambda requires modules (absolute path) and the way to get it to work with the patching infrastructure https://github.com/open-telemetry/opentelemetry-js/pull/2069

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
```diff
@@ -0,0 +1,187 @@
...
+    // Lambda loads user function using an absolute path.
+    let filename = path.resolve(taskRoot, moduleRoot, module);
+    if (!filename.endsWith('.js')) {
+      // Patching infrastructure currently requires a filename when requiring with an absolute path.
+      filename += '.js';
+    }
+
+    return [
+      new InstrumentationNodeModuleDefinition(
+        // NB: The patching infrastructure seems to match names backwards, this must be the filename, while
```
anuraaga, line 58: I found this strange pattern that did allow patching the lambda handler.

anuraaga (commented):
@dyladan @willarmiros Updated, now the tests pass even with the HEAD opentelemetry-js code.

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
vmarchaud: I'm not sure what you meant by matching name backwards ? It matches absolute path (https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L107)

plugins/node/opentelemetry-instrumentation-awslambda/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "awslambda",
+    "nodejs",
+    "tracing",
+    "profiling",
+    "instrumentation"
+  ],
+  "author": "OpenTelemetry Authors",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=8.0.0"
```
vmarchaud, line 31: should be `8.5.0`

plugins/node/opentelemetry-instrumentation-awslambda/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@opentelemetry/instrumentation-awslambda",
+  "version": "0.14.0",
```
vmarchaud, line 3: `0.15.0` to match current version

plugins/node/opentelemetry-instrumentation-awslambda/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "ts-node": "9.0.0",
+    "tslint-consistent-codestyle": "1.16.0",
+    "tslint-microsoft-contrib": "6.2.0",
+    "typescript": "3.9.7"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^0.18.0",
+    "@opentelemetry/instrumentation": "^0.18.0",
+    "@opentelemetry/semantic-conventions": "^0.18.0",
+    "semver": "^7.3.2"
```
vmarchaud, line 70: i don't think this is used

plugins/node/opentelemetry-instrumentation-awslambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for [`AWS Lambda`](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html).
+
+This module is currently under active development and not ready for general use.
+
+## Installation
+
+```bash
+npm install --save @opentelemetry/instrumentation-dns
```
vmarchaud, line 15: should be `@opentelemetry/instrumentation-awslambda`

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
anuraaga: Yeah, I mean that we pass the filename to a class called `InstrumentationNodeModuleDefinition`, but pass a module name to a class called `InstrumentationNodeModuleFile`. I'd expect file to refer to a filename :)

anuraaga:
By the way I could confirm this instrumentation works to instrumentent lambda invocations by building into a lambda layer in https://github.com/open-telemetry/opentelemetry-lambda/pull/25

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
vmarchaud: Yeah thats weird, i'm not sure if it's because of the resolving done by lamda. With other instrumentation we do the opposite (module is a name and file is an relative path to the module): https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts#L59

vmarchaud (approved):
lgtm

plugins/node/opentelemetry-instrumentation-awslambda/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@opentelemetry/instrumentation-awslambda",
```
obecny, line 2: I think the name should be 
```suggestion
  "name": "@opentelemetry/instrumentation-aws-lambda",
```
you are patching `aws-lambda` and those are 2 words anyway

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
```diff
@@ -0,0 +1,178 @@
...
+      });
+
+      // Lambda seems to pass a callback even if handler is of Promise form, so we wrap all the time before calling
+      // the handler and see if the result is a Promise or not. In such a case, the callback is usually ignored.
+      const wrappedCallback = plugin._wrapCallback(callback, span);
+      const maybePromise = safeExecuteInTheMiddle(
+        () => original.apply(this, [event, context, wrappedCallback]),
+        error => {
+          if (error != null) {
+            span.setStatus({
```
obecny, line 120: use also `span.recordException` to save the whole stack of error

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
```diff
@@ -0,0 +1,178 @@
...
+          () => span.end(),
+          (err: Error | string) => {
+            let errMessage;
+            if (typeof err === 'string') {
+              errMessage = err;
+            } else if (err) {
+              errMessage = err.message;
+            }
+            if (errMessage) {
+              span.setStatus({
```
obecny, line 142: + `span.recordException`

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
obecny: this indeed looks strange as it should be a module name if user requires for example `http` then the module name is `http` so in this case it should be `aws-lambda` not a file name. If you need to patch some additional files then this is what `InstrumentationNodeModuleFile` is for

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
```diff
@@ -0,0 +1,178 @@
...
+      context: Context,
+      callback: Callback
+    ) {
+      const name = context.functionName;
+      const span = plugin.tracer.startSpan(name, {
+        kind: SpanKind.SERVER,
+        attributes: {
+          [FaasAttribute.FAAS_EXECUTION]: context.awsRequestId,
+          [FaasAttribute.FAAS_ID]: context.invokedFunctionArn,
+          // TODO(anuraaga): Add cloud.account.id when there is a JS accessor for it.
```
obecny, line 109: can you please create a ticket for that so it is not lost ?

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
```diff
@@ -0,0 +1,178 @@
...
+      }
+      return maybePromise;
+    };
+  }
+
+  private _wrapCallback(original: Callback, span: Span): Callback {
+    return function wrappedCallback(this: {}, err, res) {
+      diag.debug('executing wrapped lookup callback function');
+
+      let errMessage;
```
obecny, line 160: this part of code is the same as in line 135. Please create a private method for that something like
`this._addError(span)`
please also use `recordException` if you have an object error.

plugins/node/opentelemetry-instrumentation-awslambda/src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Handler } from 'aws-lambda';
+
+export interface LambdaModule {
```
obecny, line 19:
```suggestion
export type LambdaModule = Record<string, Handler>;
```

plugins/node/opentelemetry-instrumentation-awslambda/test/lambdatest/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "lambdatest",
```
obecny, line 2:
```suggestion
  "name": "lambda-test",
```

plugins/node/opentelemetry-instrumentation-awslambda/test/integrations/lambda-handler.test.ts
```diff
@@ -0,0 +1,262 @@
...
+        err = e;
+      }
+      assert.strictEqual(err!, 'handler error');
+      const spans = memoryExporter.getFinishedSpans();
+      const [span] = spans;
+      assertSpanFailure(span);
+    });
+  });
+
+  context('sync success handler', () => {
```
obecny, line 142: please add block `describe('sync success handler')` and similar block to async

obecny (commented):
Please change names everywhere from `awslambda` to `aws-lambda` including the folder name

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
dyladan: Can this not be extracted from the ARN?

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
dyladan: recordException I don't think sets status does it? You may need to do both if that is the behavior you want.

dyladan (commented):
How are you handling flushing? Are you just hoping that the lambda runtime isn't paused at an inconvenient time?

vmarchaud:
> Please change names everywhere from awslambda to aws-lambda including the folder name

Is this a requirement that i missed for vendor components or just a better naming suggestion ?

obecny:
> aws-lambda

better naming and consistency, `aws-lambda` is the official name of the package so I think we should keep the exact name as we do with other things. It is also 2 words so one more reason for having a dash like we do with everything else with regards to naming.

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
willarmiros: +1 to getting account from ARN

plugins/node/opentelemetry-instrumentation-awslambda/src/awslambda.ts
```diff
@@ -0,0 +1,178 @@
...
+        kind: SpanKind.SERVER,
+        attributes: {
+          [FaasAttribute.FAAS_EXECUTION]: context.awsRequestId,
+          [FaasAttribute.FAAS_ID]: context.invokedFunctionArn,
+          // TODO(anuraaga): Add cloud.account.id when there is a JS accessor for it.
+        },
+      });
+
+      // Lambda seems to pass a callback even if handler is of Promise form, so we wrap all the time before calling
+      // the handler and see if the result is a Promise or not. In such a case, the callback is usually ignored.
```
willarmiros, line 114: > In such a case, the callback is usually ignored.

What if the callback isn't ignored? Would the span be ended twice/would that matter?

willarmiros (approved):
LGTM in general

anuraaga (commented):
Thanks all applied cleanups

anuraaga:
> How are you handling flushing? Are you just hoping that the lambda runtime isn't paused at an inconvenient time?

Thanks a lot for the reminder, added a forceFlush at the end of the function invocation.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -0,0 +1,216 @@
...
+      });
+    }
+
+    span.end();
+    // Lambda only runs on Node so always true in practice.
+    if (this._tracerProvider instanceof NodeTracerProvider) {
+      this._tracerProvider
+        .getActiveSpanProcessor()
+        .forceFlush()
+        .finally(() => callback());
```
Flarna, line 203: I don't think Node.js 8 supports `finally`. But AWS doesn't support node 8 anymore so no issue. But the supported version in package.json should be adapted and maybe exclude this module also from tests with node 8.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -0,0 +1,216 @@
...
+    if (errMessage) {
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: errMessage,
+      });
+    }
+
+    span.end();
+    // Lambda only runs on Node so always true in practice.
+    if (this._tracerProvider instanceof NodeTracerProvider) {
```
Flarna, line 199: I don't think we need to be that strict. Someone could implement it's own `TracerProvider` by extending `BasicTracerProvider`.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -0,0 +1,216 @@
...
+      return this._getPatchHandler(original);
+    };
+  }
+
+  private _getPatchHandler(original: Handler) {
+    diag.debug('patch handler function');
+    const plugin = this;
+
+    return function patchedHandler(
+      this: {},
```
Flarna, line 108: Maybe use `this: never` as you just forward. This protects from accidentally using it.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -0,0 +1,216 @@
...
+        () => original.apply(this, [event, context, wrappedCallback]),
+        error => {
+          if (error != null) {
+            // Exception thrown synchronously before resolving callback / promise.
+            plugin._endSpan(span, error, () => {});
+          }
+        }
+      ) as Promise<{}> | undefined;
+      if (
+        typeof maybePromise !== 'undefined' &&
```
Flarna, line 140: maybe use `typeof maybePromise?.then`?

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -0,0 +1,216 @@
...
+export class AwsLambdaInstrumentation extends InstrumentationBase {
+  private _tracerProvider: TracerProvider | undefined;
+
+  constructor() {
+    super('@opentelemetry/instrumentation-awslambda', VERSION);
+  }
+
+  init() {
+    // _HANDLER and LAMBDA_TASK_ROOT are always defined in Lambda.
+    const taskRoot = process.env.LAMBDA_TASK_ROOT!;
```
Flarna, line 51: Maybe skip init if this is not true? e.g. if AWS changes this or someone loads this extension somewhere outside of AWS?

plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
```diff
@@ -64,10 +63,10 @@
     "typescript": "3.9.7"
   },
   "dependencies": {
-    "@opentelemetry/api": "^0.18.0",
-    "@opentelemetry/instrumentation": "^0.18.0",
-    "@opentelemetry/node": "^0.18.0",
-    "@opentelemetry/resources": "^0.18.0",
-    "@opentelemetry/semantic-conventions": "^0.18.0"
+    "@opentelemetry/api": "0.18.0",
```
Flarna, line 66: I don't think we pin dependencies to other otel modules in general.

plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
```diff
@@ -43,7 +43,6 @@
   "devDependencies": {
     "@opentelemetry/core": "0.18.0",
     "@opentelemetry/node": "0.18.0",
```
Flarna, line 45: Looking on tip of main it seems we are on 0.18.2 for dev dependencies now.

anuraaga (commented):
Thanks @Flarna - also, do you happen to know how to ignore a test on nodejs 8 in CI? If it's not easy I'm thinking of just going with `then(() => callback(), () => callback())`

Flarna:
> do you happen to know how to ignore a test on nodejs 8 in CI?

Sorry, don't know. One simple solution would be to call `this.skip()` in the test suite for old node versions. But maybe there is a better solution.

anuraaga:
@flarna Thanks! Skipping sounds good

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+```bash
+npm install --save @opentelemetry/instrumentation-aws-lambda
+```
+
+## Usage
+
+Create a file to initialize the instrumentation, such as `lambda-wrapper.js`.
+```js
+const { NodeTracerProvider } = require('@opentelemetry/node');
+const { AwsLambdaInstrumentation } = require('@opentelemetry/instrumentation-awslambda');
```
obecny, line 23:
```suggestion
const { AwsLambdaInstrumentation } = require('@opentelemetry/instrumentation-aws-lambda');
```

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+- For help or feedback on this project, join us in [GitHub Discussions][discussions-url]
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[discussions-url]: https://github.com/open-telemetry/opentelemetry-js/discussions
+[license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda
```
obecny, line 54:
```suggestion
[dependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-aws-lambda
```

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda
+[devDependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda&type=dev
+[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda&type=dev
```
obecny, line 57:
```suggestion
[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-aws-lambda&type=dev
```

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+[npm-url]: https://www.npmjs.com/package/@opentelemetry/instrumentation-awslambda
```
obecny, line 58:
```suggestion
[npm-url]: https://www.npmjs.com/package/@opentelemetry/instrumentation-aws-lambda
```

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+[npm-img]: https://badge.fury.io/js/%40opentelemetry%2Finstrumentation-awslambda.svg
```
obecny, line 59:
```suggestion
[npm-img]: https://badge.fury.io/js/%40opentelemetry%2Finstrumentation-aws-lambda.svg
```

plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "lint:fix": "eslint . --ext .ts --fix",
+    "codecov": "nyc report --reporter=json && codecov -f coverage/*.json -p ../../",
+    "precompile": "tsc --version",
+    "prepare": "npm run compile",
+    "version:update": "node ../../../scripts/version-update.js",
+    "compile": "npm run version:update && tsc -p ."
+  },
+  "keywords": [
+    "opentelemetry",
+    "awslambda",
```
obecny, line 22:
```suggestion
    "aws-lambda",
```

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -0,0 +1,216 @@
...
+
+import { LambdaModule } from './types';
+import { VERSION } from './version';
+import { BasicTracerProvider } from '@opentelemetry/tracing';
+
+export class AwsLambdaInstrumentation extends InstrumentationBase {
+  private _tracerProvider: TracerProvider | undefined;
+
+  constructor() {
+    super('@opentelemetry/instrumentation-awslambda', VERSION);
```
obecny, line 46:
```suggestion
    super('@opentelemetry/instrumentation-aws-lambda', VERSION);
```

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[discussions-url]: https://github.com/open-telemetry/opentelemetry-js/discussions
+[license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda
+[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda
```
obecny, line 55:
```suggestion
[dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-aws-lambda
```

plugins/node/opentelemetry-instrumentation-aws-lambda/README.md
```diff
@@ -0,0 +1,59 @@
...
+[devDependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-awslambda&type=dev
```
obecny, line 56:
```suggestion
[devDependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-aws-lambda&type=dev
```

anuraaga:
@obecny >< Thanks for the help on cleanup!

obecny (approved):

anuraaga:
@obecny I ended up switching from `finally` to `.then(repeat, repeat)` because skipping the test caused code coverage to not be computed correctly. Hopefully it's not a big deal.

I have a couple of important followups planned, including implementing the parenting mechanism from the spec and HTTP attributes for proxy events. If this looks OK appreciate a merge :)

PR #403, opened 2021-03-31
chore: fixing context propagation on mongo callback

### LINKED ISSUE #390: mongodb instrumentation: nested span is not generated when using callback
    "@opentelemetry/api": "0.18.0"
    "@opentelemetry/core": "0.18.0"
    "@opentelemetry/instrumentation-mongodb": "0.14.0"

12.20.0

### What did you do?

```
import { SimpleSpanProcessor } from '@opentelemetry/tracing';
import { NodeTracerProvider } from '@opentelemetry/node';
import { ZipkinExporter } from '@opentelemetry/exporter-zipkin';

import { registerInstrumentations } from '@opentelemetry/instrumentation';
import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
import { MongoDBInstrumentation } from '@opentelemetry/instrumentation-mongodb';

const serviceName = 'otel-mongodb-test';

registerInstrumentations({
    instrumentations: [
        new HttpInstrumentation(),
        new MongoDBInstrumentation(),
    ],
    tracerProvider: provider,
});

const zipkinExporter = new ZipkinExporter({
    serviceName,
    url: 'http://localhost:9411/api/v2/spans'
});

provider.addSpanProcessor(new SimpleSpanProcessor(zipkinExporter));
provider.register();
const tracer = provider.getTracer(serviceName);

import * as http from 'http';
import * as mongodb from 'mongodb';

let collection;
const mongodbConnect = async () => {
    try {
        const client = await mongodb.MongoClient.connect('mongodb://localhost:27017');
        collection = client.db('test-db').collection('users');
        console.log('\nSuccessfully connected to mongoDB!\n');
    } catch (err) {
        console.log(`Error while connecting to mongoDB ${err}`, err);
        throw err;
    }
};

mongodbConnect();

http.createServer(function (_, res) {
    const doc = { name: 'Rick' };

    collection.insertOne(doc, function () {
        collection.findOne({}, function () {
            res.end();
        });
    });
}).listen(9090);
```

The new `mongodb.find` span being generated as a child of the `mongodb.insert` span.

### What did you see instead?
Screenshot 2021-03-17

### Additional context
It happens when a callback for a parent `mongodb.insert` operation is used.

When I'm using `Promise.then` instead of a callback a span for a `find` operation is generated as expected.

The following example works fine:
```
collection.insertOne(doc).then(function () {
        collection.findOne({}, function () {
            res.end();
        });
    });
```

obecny:

- Fixes #390 - fix is only in instrumentation package as plugin is going to be removed soon anyway

- when using callback instead of promise mongodb is using tick when calling callback. This causing the context to be not propagated correctly


plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -406,7 +409,10 @@ export class MongoDBInstrumentation extends InstrumentationBase<
         });
       }
       span.end();
-      return resultHandler.apply(this, args);
+
+      return context.with(activeContext, () => {
```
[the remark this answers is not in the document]
dyladan: we could also use `context.with(activeContext, resultHandler, this, ...args)`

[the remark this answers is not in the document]
obecny: Hmm all 3 options would work, so I would like to stay with the gate #1 if you don't mind ;)

dyladan (approved):

examples/mongodb/package.json
```diff
@@ -33,8 +33,8 @@
     "@opentelemetry/exporter-zipkin": "^0.18.0",
     "@opentelemetry/instrumentation": "^0.18.0",
     "@opentelemetry/instrumentation-http": "^0.18.0",
+    "@opentelemetry/instrumentation-mongodb": "^0.14.0",
```
jtmalinowski, line 36: Could be `^0.15.0`, potentially will fix itself when merging `main`

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
```diff
@@ -14,8 +14,7 @@
  * limitations under the License.
  */
 
-// for testing locally use this command to run docker
-// docker run -e MONGODB_DB=opentelemetry-tests -e MONGODB_PORT=27017 -e MONGODB_HOST=localhost -p 27017:27017 --name otmongo mongo
+// for testing locally "npm run docker:start" / "npm run docker:stop"
```
jtmalinowski, line 17: It doesn't start detached, so `docker:stop` doesn't make sense.

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -6,6 +6,8 @@
   "types": "build/src/index.d.ts",
   "repository": "open-telemetry/opentelemetry-js-contrib",
   "scripts": {
+    "docker:start": "docker run -e MONGODB_DB=opentelemetry-tests -e MONGODB_PORT=27017 -e MONGODB_HOST=localhost -p 27017:27017 --name otmongo mongo",
```
jtmalinowski, line 9: User experience would be better if we got rid of the name (`--name otmongo`) and instead auto-remove (`--rm`). If you run the command as it is, then stop the container, then run again, you get: 
```
docker: Error response from daemon: Conflict. The container name "/otmongo" is already in use by container
```

obecny: thx for tip it works better now

vmarchaud:
cc @jtmalinowski i believe your comments were addressed, are you okay to merge it ?

jtmalinowski (approved):

PR #406, opened 2021-04-01
chore: fix hapi compilation issue

dyladan:
Hapi instrumentation is failing to compile due to upstream `@types` issues. Disabling the check for the lib.

PR #407, opened 2021-04-01
fix: hapi compilation

Flarna:

Fix compilation issues by pinning @hapi/podium to 4.1.1 as later version has it's own typings which is not compatible with remaining dependencies.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-hapi/package.json
+++ b/plugins/node/opentelemetry-instrumentation-hapi/package.json
@@ -42,6 +42,7 @@
   },
   "devDependencies": {
     "@hapi/hapi": "20.0.1",
+    "@hapi/podium": "4.1.1",
     "@opentelemetry/context-async-hooks": "0.18.0",
     "@opentelemetry/node": "0.18.0",
     "@opentelemetry/tracing": "0.18.0",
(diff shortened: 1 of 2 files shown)


dyladan:
Once you get past the hapi stage you will need to apply https://github.com/open-telemetry/opentelemetry-js-contrib/pull/406/commits/a3f5b0d50516c8c3acc977b575dc4287eef0b6d2 too probably

Flarna:
An alternative to this change would be most likely that one described in hapi issue tracker: https://github.com/hapijs/hapi/issues/4240#issuecomment-810860745

Flarna:
> Once you get past the hapi stage you will need to apply [a3f5b0d](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/a3f5b0d50516c8c3acc977b575dc4287eef0b6d2) too probably

I cherry-picked this change. But honestly speaking I'm not a big fan of this. I think we should move to 0.18.2 everywhere instead at a few locations only.

dyladan:
> > Once you get past the hapi stage you will need to apply [a3f5b0d](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/a3f5b0d50516c8c3acc977b575dc4287eef0b6d2) too probably
> 
> I cherry-picked this change. But honestly speaking I'm not a big fan of this. I think we should move to 0.18.2 everywhere instead at a few locations only.

I agree, but this is the minimum to fix the currently broken compile and it is a dev dependency only

dyladan (approved):

obecny (approved):

PR #408, opened 2021-04-01
chore: bump otel dependencies to latest patch

dyladan:


vmarchaud (approved):
do you want to include this in the 0.15 contrib release ?

Flarna (approved):

Flarna:
Would be be possible to use a bot for tasks like this? Currently renovate bot tries to update all deps in one shot but this easily fails.
Maybe create a dedicated config to update only otel core and one more for otel api?

dyladan:
> Would be be possible to use a bot for tasks like this? Currently renovate bot tries to update all deps in one shot but this easily fails.
> Maybe create a dedicated config to update only otel core and one more for otel api?

likely possible. i'll look into it

obecny:
you have missed propagators, examples, metapackages (auto), detectors

PR #413, opened 2021-04-06
document: fix express example

### LINKED ISSUE #376: Express example doesn't work as expected
v0.18.0

v14.15.4 (npm v6.14.10)

### What did you do?
If possible, provide a recipe for reproducing the error.
1. Try to run the express example from => https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/express

Running the zipkin:client npm script thrown an error! 

```sh
npm run zipkin:client

> express-example@0.13.1 zipkin:client /Users/bojanche/private/opentelemetry-js-contrib/examples/express
> cross-env EXPORTER=zipkin node ./client.js
(node:35519) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'ERROR' of undefined
    at /Users/bojanche/private/opentelemetry-js-contrib/examples/express/client.js:19:45
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
(node:35519) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:35519) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Easy fix tho `StatusCode` should be `SpanStatusCode`
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e3c1359649afaeda35fda35f460c5cc91326b9e4/examples/express/client.js#L19 

However, after fixing the issue above I'm still not getting the desired outcome with trace and propper express spans...

Something like the example expected images for Express
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/examples/express/images/zipkin.jpg

### What did you see instead?
Only one `client.makeRequest` span
image

### Additional context

Willing to help and create PR during the weekend for the `SpanStatusCode` issue and the example partially not working... however will need some guidance for the Express plugin and why it is not working properly (maybe some changes introduced with 0.17.0 or 0.18.0?)

Rauno56:

The express example was erroring and used soon-to-be-deprecated API. I'm bringing that up to date.

----

Mainly just removing the usage of the old plugins API and replacing that with `registerInstrumentations`. I also added logger configuration because especially for the newcomers it will definitely be helpful to see those messages and know how to configure the logger. Other than that it's functionally the same.

----

A question for the original express instrumentation developer(@vmarchaud, I believe): **Is it expected that if HTTP layer is not instrumented, express instrumentation is a no-op? Should we document that? Error? Somehow make HTTP instrumentation a dependency for express?**

If that looks alright, I could work on others as well.

Fixes #376

### DIFF
--- a/examples/express/client.js
+++ b/examples/express/client.js
@@ -13,10 +13,11 @@ function makeRequest() {
   api.context.with(api.setSpan(api.ROOT_CONTEXT, span), async () => {
     try {
       const res = await axios.get('http://localhost:8080/run_test');
-      span.setStatus({ code: api.StatusCode.OK });
-      console.log(res.statusText);
+      console.log('status:', res.statusText);
+      span.setStatus({ code: api.SpanStatusCode.OK });
     } catch (e) {
-      span.setStatus({ code: api.StatusCode.ERROR, message: e.message });
+      console.log('failed:', e.message);
+      span.setStatus({ code: api.SpanStatusCode.ERROR, message: e.message });
     }
     span.end();
     console.log('Sleeping 5 seconds before shutdown to ensure all records are flushed.');
--- a/examples/express/package.json
+++ b/examples/express/package.json
@@ -32,15 +32,14 @@
     "@opentelemetry/exporter-jaeger": "^0.18.0",
     "@opentelemetry/exporter-zipkin": "^0.18.0",
     "@opentelemetry/instrumentation": "^0.18.0",
+    "@opentelemetry/instrumentation-express": "^0.15.0",
+    "@opentelemetry/instrumentation-http": "^0.18.2",
     "@opentelemetry/node": "^0.18.0",
-    "@opentelemetry/plugin-express": "file:../../plugins/node/opentelemetry-plugin-express",
-    "@opentelemetry/plugin-http": "^0.18.0",
     "@opentelemetry/tracing": "^0.18.0",
     "axios": "^0.19.0",
+    "cross-env": "^7.0.3",
     "express": "^4.17.1"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js#readme",
-  "devDependencies": {
-    "cross-env": "^6.0.0"
-  }
+  "devDependencies": {}
 }
--- a/examples/express/server.js
+++ b/examples/express/server.js
@@ -31,27 +31,24 @@ const authMiddleware = (req, res, next) => {
   }
 };
 
-async function setupRoutes() {
-  app.use(express.json());
-
-  app.get('/run_test', async (req, res) => {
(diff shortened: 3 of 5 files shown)


jtmalinowski (approved):

dyladan (approved):

vmarchaud:
> A question for the original express instrumentation developer(@vmarchaud, I believe): Is it expected that if HTTP layer is not instrumented, express instrumentation is a no-op? Should we document that? Error? 


I indeed wrote the instrumentation and its expected that without the http instrumentation, the instrumentation does nothing. I don't think this is a problem right now (either in term of documentation or behavior) but maybe i don't see it, why do you think it should be documented ?

> [...] Somehow make HTTP instrumentation a dependency for express?

Thats not really possible, instrumentation should not depend on each user, specially because you need to be able to configure them. If the express instrumentation load its http instrumentation itself, the user can't control its config.

vmarchaud (approved):

Rauno56:
> [...] why do you think it should be documented ?

Because the alternative - registering express instrumentation (without HTTP inst) and not seeing anything - is worse and will definitely frustrate people who expect at least _something_ to show up in their tracing UI when they enable their first instrumentation. As a very fundamental use-case, it will disproportionally affect people less experienced with OTel, which is something we want to especially avoid as a relatively young effort.

> Thats not really possible, instrumentation should not depend on each user, specially because you need to be able to configure them. If the express instrumentation load its http instrumentation itself, the user can't control its config.

The way to do that might be unprecedented or unobvious, but I'm sure we can figure something out! :) _OR_ if this really is the case (for now) it should at least be documented IMHO. Do we have any other takes on that?

I'll write something up regardless, but leave the behaviour as it is for now. **Please hold the merge until the commit has been added.**

Flarna:
> I indeed wrote the instrumentation and its expected that without the http instrumentation, the instrumentation does nothing.

That's not really true. The express instrumentation creates only spans if there is a current active span but there is no check that this span was created by http instrumentation. It will create a child on whatever parent is present.

Rauno56:
> That's not really true. The express instrumentation creates only spans if there is a current active span but there is no check that this span was created by http instrumentation. It will create a child on whatever parent is present.

True, good point! What could be the other possible expected sources for that parent span? Might there also exist a case where the spans created by express instrumentation are attached to an unexpected parent span if the HTTP instrumentation is not there?

PR #415, opened 2021-04-07
feat(instrumentation-ioredis): add request hook

### LINKED ISSUE #411: Capturing instrumented module version

Hello
I want to open a small discussion regarding capturing instrumented library version, so in case of issues, the data is available for debugging and investigation.
Our current convention for this in [opentelemetry-ext-js](https://github.com/aspecto-io/opentelemetry-ext-js) is to add `moduleVersionAttributeName` option to instrumentation config, and if this option is set, a span attribute is automatically added by the plugin for each span. The key of this span attribute is the string supplied via the `moduleVersionAttributeName` config option, and the value is the instrumented module version.

Usage example:
```js
export const amqplibInstrumentationConfig: AmqplibInstrumentationConfig = {
    moduleVersionAttributeName: 'instrumentedmodule.version',
};
// will add to each span an attribute {'instrumentedmodule.version': '1.2.3'} (for example)
```

Since we are also using the `ioredis` instrumentation from this repo, we would like to create PR and add this feature to it as well. Before I invest time in a PR, I thought to discuss about how to implement it.

Another option would be to pass the version as a function parameter to an external instrumentation hook, and let the user add the attribute as he like, but that would require another hook to be added to `ioredis` instrumentation (and possibly other instrumentations in the future).
Something like this:
```js
const ioredisInstrumentationConfig: IORedisInstrumentationConfig = {
    requestHook: (span: Span, instrumentedModuleVersion: string) => {
        span.setAttribute('instrumentedmodule.version', instrumentedModuleVersion);
    }
}
```

blumamir:

fixes #411 
- `ioredis` instrumentation: support capturing the instrumented `ioredis` package version and set it as span attribute.

- added the request hook which user can use to capture additional data from db request, such as moduleVersion, command name, command arguments


dyladan:
Is there semconv spec for this?

blumamir:
> Is there semconv spec for this?

no, this is why we allow the user to specify the attribute name and not using hardcoded value

dyladan:
There used to be a semconv for the `component` attribute and it was removed. There were quite some discussions about this attribute in the spec at the time, but I think I would prefer to come to some specification consensus on this.

blumamir:
> I think I would prefer to come to some specification consensus on this.

It could be great. Do you mean the general opentelemetry specification? or convention only for the javascript?

dyladan:
> > I think I would prefer to come to some specification consensus on this.
> 
> It could be great. Do you mean the general opentelemetry specification? or convention only for the javascript?

I mean the opentelemetry specification

blumamir:
> > > I think I would prefer to come to some specification consensus on this.
> > 
> > 
> > It could be great. Do you mean the general opentelemetry specification? or convention only for the javascript?
> 
> I mean the opentelemetry specification

https://github.com/open-telemetry/opentelemetry-specification/issues/1605

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
```diff
@@ -81,9 +81,14 @@ export class IORedisInstrumentation extends InstrumentationBase<
   /**
    * Patch send command internal to trace requests
    */
-  private _patchSendCommand() {
+  private _patchSendCommand(moduleVersion?: string | undefined) {
```
vmarchaud, line 84: could be `moduleVersion?: string`

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -85,7 +85,8 @@ const defaultDbStatementSerializer: DbStatementSerializer = (
 export const traceSendCommand = (
   tracer: Tracer,
   original: Function,
-  config?: IORedisInstrumentationConfig
+  config?: IORedisInstrumentationConfig,
+  moduleVersion?: string | undefined
```
vmarchaud, line 89: same here

plugins/node/opentelemetry-instrumentation-ioredis/src/ioredis.ts
blumamir: Fixed, thanks.
Sorry for the delay, I only now seen this comment

blumamir:
I think it might take time for this feature to enter spec.
@dyladan in the meantime, if I'll change this PR to send the module version via request hook, so instrumentation user can add it as span attribute under whatever attribute name he desires, will it be ok for you?

dyladan:
@blumamir sorry for slow response i was out for part of last week.

>in the meantime, if I'll change this PR to send the module version via request hook, so instrumentation user can add it as span attribute under whatever attribute name he desires, will it be ok for you?

I'm not sure what exactly you mean by this. The current implementation looks fine to me and I was about to approve actually before I saw this comment.

Do you mean like this?

```typescript
class IoredisInstrumentation extends ... {
  // ...

  // register a hook to add custom attributes to spans
  onRequest((req: Request, span: Span) => void): void;
}
```

blumamir:
> Do you mean like this?

I was thinking to add it to the hook function parameters like this:
```js
export interface IORedisInstrumentationConfig extends InstrumentationConfig {
  requestHook?: (span: Span, ioredisVersion: string, /** other params */);
}
```

and then instrumentation user can write hook that adds the span attribute under whatever name he desires:
```js
new IORedisInstrumentation({
    requestHook: (span: Span, ioredisVersion: string) => {
         span.setAttribute('whatever.name.user.want', ioredisVersion);
    }
});
```

Hope the syntax is right, I wrote it here without validating it.

I'm ok with both the `moduleVersionAttributeName` approach and the `requestHook` approach. Whichever one you think is better.

dyladan:
> Would actually make sense to make it on instrumentation level a general hook ?.
> Maybe something like `onSpanStarted` where we could pass an object param with information about instrumented library name and version ? There are more and more instrumentation where people want to add some additional information, it doesn't has to work in every edge case, but majority of usage will be to just add something to the span and I would like NOT to do this for every single instrumentation.

Think this is quite dependent on each instrumentation I think in terms of what would be useful (http instrumentation hook might need the request object, mysql hook might need the query and the connection). If only the span is sent to the hook, it might not be enough information for the user to add the things that are important to them.

obecny:
whoops @dyladan I think I edited yr response, fixed :/

> Think this is quite dependent on each instrumentation I think in terms of what would be useful (http instrumentation hook might need the request object, mysql hook might need the query and the connection). If only the span is sent to the hook, it might not be enough information for the user to add the things that are important to them.

Yea but we can think of general implementation in instrumentation
```
onSpanStarted(span, somegeneralParams)
```
and if you need more information from certain instrumentation then extend it
```
onSpanStarted(span, someGeneralParams, someAdditionalParams....)
```
I'm not saying we have to do it now, but to avoid adding more less the same to many instrumentations we can think if creating some general and simply add more info in specific instrumentations. The advantage would be that you will have one hook that exists always, so it will be easier to unify things

blumamir:
@obecny - if we can generalize it that would be awesome.
How do you imagine the general invocation of a hook when the instrumentation is patching multiple different modules, or patch the same module but with multiple installed versions?

In the past, I was experimenting with collecting stack trace per span (to help correlate the span to the code), and I did it with "require-in-the-middle" patching the tracer `startSpan` function. For those use cases adding a general start span hook can be very useful as well.

obecny:
> startSpan

Maybe create some observer / subscriber / interceptor solution that would allow to make such things. In the worst scenario basically patching our tracer startSpan, but I didn't analyse it deeply yet. I know this can be done and could be quite useful for end users.

obecny:
@blumamir with regards to 
```
new IORedisInstrumentation({
    requestHook: (span: Span, ioredisVersion: string) => {
         span.setAttribute('whatever.name.user.want', ioredisVersion);
    }
});
```
I think this is the best approach I would just change one thing

```
new IORedisInstrumentation({
    requestHook: (span: Span, hookInfo: IoRedisHookInformation) => {
         span.setAttribute('whatever.name.user.want', hookInfo.ioredisVersion);
    }
});
```
The reason is that it will be easier to add / change things in future and to avoid having too many params being added as next param. I can imagine someone else would be willing to export more data one day.

blumamir:
> @blumamir with regards to
> 
> ```
> new IORedisInstrumentation({
>     requestHook: (span: Span, ioredisVersion: string) => {
>          span.setAttribute('whatever.name.user.want', ioredisVersion);
>     }
> });
> ```
> 
> I think this is the best approach I would just change one thing
> 
> ```
> new IORedisInstrumentation({
>     requestHook: (span: Span, hookInfo: IoRedisHookInformation) => {
>          span.setAttribute('whatever.name.user.want', hookInfo.ioredisVersion);
>     }
> });
> ```
> 
> The reason is that it will be easier to add / change things in future and to avoid having too many params being added as next param. I can imagine someone else would be willing to export more data one day.


@obecny  Cool I think it's a good approach as well. Thank you for reviewing.
I'll update the PR next week as I'm having very busy days.

blumamir:
@obecny  - Update the PR to work with hooks like you suggested

vmarchaud (approved):

weyert (approved):

plugins/node/opentelemetry-instrumentation-ioredis/README.md
```diff
@@ -49,6 +49,7 @@ IORedis instrumentation has few options available to choose from. You can set th
 | Options | Type | Description |
 | ------- | ---- | ----------- |
 | `dbStatementSerializer` | `DbStatementSerializer` | IORedis instrumentation will serialize db.statement using the specified function. |
+| `requestHook` | `RedisRequestCustomAttributeFunction` | Function for adding custom attributes on db request |
```
Flarna, line 52: Maybe add some more info in readme how this hook is used and which data is given to it.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-ioredis/README.md
blumamir: @Flarna added the function signature to README.
I'm open to suggestions on how to improve it further.

blumamir:
The failing checks are not related to this PR

plugins/node/opentelemetry-instrumentation-ioredis/README.md
```diff
@@ -49,7 +49,8 @@ IORedis instrumentation has few options available to choose from. You can set th
 | Options | Type | Description |
 | ------- | ---- | ----------- |
 | `dbStatementSerializer` | `DbStatementSerializer` | IORedis instrumentation will serialize db.statement using the specified function. |
-| `responseHook` | `RedisResponseCustomAttributeFunction` | Function for adding custom attributes on db response |
+| `requestHook` | `RedisRequestCustomAttributeFunction` (function) | Function for adding custom attributes on db request. Receives params: `(span: Span, { moduleVersion, cmdName, cmdArgs } : ExpressRequestHookInformation)` |
```
Flarna, line 52:
```suggestion
| `requestHook` | `RedisRequestCustomAttributeFunction` (function) | Function for adding custom attributes on db request. Receives params: `span, { moduleVersion, cmdName, cmdArgs }` |
```

plugins/node/opentelemetry-instrumentation-ioredis/README.md
Flarna: A short sample like for `dbStatementSerializer` could be added. I assume you have some use of this already somewhere in your samples/tests/... you could add there.

plugins/node/opentelemetry-instrumentation-ioredis/README.md
blumamir: Done, thanks

plugins/node/opentelemetry-instrumentation-ioredis/README.md
blumamir: added requestHook example to README

plugins/node/opentelemetry-instrumentation-ioredis/README.md
```diff
@@ -85,6 +114,6 @@ Apache 2.0 - See [LICENSE][license-url] for more information.
 [dependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis
 [dependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis
 [devdependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev
-[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev
+[devdependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev
```
obecny, line 117: hmm why this change ?

plugins/node/opentelemetry-instrumentation-ioredis/src/types.ts
```diff
@@ -40,6 +40,16 @@ export type DbStatementSerializer = (
   cmdArgs: IORedisCommand['args']
 ) => string;
 
+export interface IoRedisRequestHookInformation {
```
obecny, line 43: Shouldn't this be 
```suggestion
export interface IORedisRequestHookInformation {
```
as everywhere you have `IORedis....`

plugins/node/opentelemetry-instrumentation-ioredis/README.md
```diff
@@ -59,11 +62,37 @@ It is also possible to define a custom serialization function. The function will
...
+const { IORedisInstrumentation } = require('@opentelemetry/instrumentation-ioredis');
+
+const ioredisInstrumentation = new IORedisInstrumentation({
+requestHook: function (
+    span: Span,
+    requestInfo: IoRedisRequestHookInformation
+  ) {
+    if (requestInfo.moduleVersion) {
+      span.setAttribute(
+        "instrumented_library.version",
```
obecny, line 92:
```suggestion
        'instrumented_library.version',
```

obecny (commented):
lgtm, just the name of interface should start with `IO` not the `Io`, the rest are minor

plugins/node/opentelemetry-instrumentation-ioredis/README.md
blumamir: weird, I don't remember changing it. maybe some linter or something
Anyhow, the current README in main looks broken, at least currently in my browser, so maybe we can keep it? I don't care to revert it also if you prefer.

image

plugins/node/opentelemetry-instrumentation-ioredis/README.md
blumamir: fixed. and I also fixed it in the 'db.statement` part of the README.
Thanks for spotting that

plugins/node/opentelemetry-instrumentation-ioredis/README.md
vmarchaud: I think there some kind of bug with david dm here: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev

blumamir: > I think there some kind of bug with david dm here: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev

I'm not very familiar with this README badges and how they work. @vmarchaud do you know how to fix it? should I revert the README badge change in this PR?

vmarchaud: No worries, i don't think the problem comes from us anyway so i think you should just revert that changes (unless thats the linter but i don't remember that our linter also touch markdown files)

blumamir: I don't mind reverting it, but the current version in **main** is not consistent. at the top of the README it says:
```
[![devDependencies][devDependencies-image]][devDependencies-url]
```

while at the bottom the image tag is different:
```
[devdependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev
[devDependencies-url]: https://david-dm.org/open-telemetry/opentelemetry-js-contrib?path=plugins%2Fnode%2Fopentelemetry-instrumentation-ioredis&type=dev
```

devdependencies-image is capitalized at one place and lower case at the other...

obecny: this is a link showing all dependencies but for some reason recently "david" just showing errors randomly. Then it works again etc. I was fixing all the links some time ago and they were working again, now might be either temporary or something else has changed.

obecny (approved):
lgtm, thx for changes

plugins/node/opentelemetry-instrumentation-ioredis/README.md
blumamir: reverted this change. thanks

blumamir: figured out where it came from.
When I press `option` + `shift` + `F` in visual studio code (which I do regularly) it does this modification to the markdown. Wonder why.
Anyway, it's not relevant to this PR :)

PR #416, opened 2021-04-08
feat: restify instrumentation

Rauno56:

This is adding `restify` instrumentation.

Restify has a somewhat similar API to `express` and is created for the same purpose - building APIs. This instrumentation wraps all handlers to generate a span for each of them.


plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
```diff
@@ -0,0 +1,166 @@
...
+        const attributes = {
+          [types.CustomAttributeNames.VERSION]: this._moduleVersion || 'n/a',
+          [types.CustomAttributeNames.TYPE]: metadata.type,
+          [types.CustomAttributeNames.METHOD]: metadata.methodName,
+          [HttpAttribute.HTTP_ROUTE]: route,
+        }
+        const span = this.tracer.startSpan(spanName, {
+          attributes,
+        }, api.context.active()); // TODO: <- with this I intend to find and attach all consecutive handlers to HTTP span
+        // .. but instead all spans are attached to the previous handler's span.
```
Rauno56, line 136: I actually don't specifically mind the fact that the parent spans are automatically set to the (already ended) span of the last handler, but there should be a way, for at least the instrumentation of the first handler for the every request to access the HTTP span it's tied to, right? How?

examples/restify/client.js
```diff
@@ -0,0 +1,39 @@
...
+// eslint-disable-next-line import/order
+const http = require('http');
+
+/** A function which makes requests and handles response. */
+function makeRequest(path) {
+  // span corresponds to outgoing requests. Here, we have manually created
+  // the span, which is created to track work that happens outside of the
+  // request lifecycle entirely.
+  const span = tracer.startSpan(`makeRequest to ${path}`);
+  api.context.with(api.setSpan(api.context.active(), span), () => {
```
owais, line 14: Do we need to include manual span creation in the example? I think we should keep the client as simple as possible for an example focusing on server/restify instrumentation. HTTP instrumentation should generate a client span anyway.

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
```diff
@@ -0,0 +1,204 @@
...
+  InstrumentationBase,
+  InstrumentationNodeModuleDefinition,
+  InstrumentationNodeModuleFile,
+  isWrapped,
+} from '@opentelemetry/instrumentation';
+import { HttpAttribute } from '@opentelemetry/semantic-conventions';
+
+const { diag } = api;
+const RESTIFY_MW_METHODS = ['use', 'pre'];
+const RESTIFY_METHODS = ['del', 'get', 'head', 'opts', 'post', 'put', 'patch'];
```
owais, line 33: Does restify support custom HTTP methods? If so, will this instrumentation be able to work with them?

Rauno56: No. Everything that allows adding a handler is listed here. There's a "wildcard"-way of reaching to the API of [find-my-way](https://github.com/delvedor/find-my-way) which is mentioned [here](http://restify.com/docs/server-api/#servermethodopts) in `restify` docs, but even that has to be done through one of these functions.

However, I'm not testing for that shortcut. Will add a test for that.

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
```diff
@@ -0,0 +1,204 @@
...
+            }
+          }
+        }
+      )
+    );
+
+    return module;
+  }
+
+  private _middlewarePatcher(original: Function, methodName?: string) {
```
seemk, line 107: I think middleware spans should have middleware name as the span name (not sure if possibly with restify), e.g. the operation tree looks like this currently, where each middleware operation name is the route name:
image

However the express instrumentation contains middleware names:
image

Also I think middleware spans should have the incoming request span as its parent.

seemk:
Seems to work fine on `5.2.1`

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
Rauno56: Yep, it's possible. Express takes the name of the function it seems, which can be undefined at times, but still better than nothing. Thanks!

> Also I think middleware spans should have the incoming request span as its parent.

I generally agree but don't want to change that because there might not be a request span if HTTP instrumentation is not enabled, so the structure would be different between HTTP instrumentation enabled and not.

examples/restify/client.js
Rauno56: True. It would make sense if the HTTP instrumentation was not there. I'll remove that.

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
Rauno56: Restify does not support custom HTTP methods, it turns out. It overwrites it with the one specified with function used to register the handler. Added a test to make sure the more verbose API works as expected.

johnbley (approved):
In general, LGTM.  Thanks for this significant feature contribution!

dyladan (approved):
Looks awesome thanks for contributing

vmarchaud (approved):
lgtm !

examples/restify/tracer.js
```diff
@@ -0,0 +1,50 @@
...
+const { diag, DiagConsoleLogger, DiagLogLevel } = opentelemetry;
+diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.VERBOSE);
+
+const { registerInstrumentations } = require('@opentelemetry/instrumentation');
+const { NodeTracerProvider } = require('@opentelemetry/node');
+const { SimpleSpanProcessor, ConsoleSpanExporter } = require('@opentelemetry/tracing');
+const { JaegerExporter } = require('@opentelemetry/exporter-jaeger');
+const { ZipkinExporter } = require('@opentelemetry/exporter-zipkin');
+
+const { RestifyInstrumentation: Instrumentation } = require('../../plugins/node/opentelemetry-instrumentation-restify');
```
obecny, line 14: why renaming this to `Instrumentation` ?
can you also use `@opentelemetry/instrumentation-restify` instead as well as `@opentelemetry/instrumentation-http` the last path probably exists only on yr local env.

plugins/node/opentelemetry-instrumentation-restify/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@opentelemetry/instrumentation-restify",
+  "version": "0.14.0",
```
obecny, line 3:
```suggestion
  "version": "0.15.0",
```

plugins/node/opentelemetry-instrumentation-restify/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "tslint-consistent-codestyle": "1.16.0",
+    "tslint-microsoft-contrib": "6.2.0",
+    "typescript": "4.1.3"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^0.18.0",
+    "@opentelemetry/core": "^0.18.0",
+    "@opentelemetry/instrumentation": "^0.18.0",
+    "@opentelemetry/semantic-conventions": "^0.18.0",
+    "lodash.once": "^4.1.1"
```
obecny, line 68: I would like to remove any 3rd party dependency

plugins/node/opentelemetry-instrumentation-restify/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Usage
+
+```js
+const { NodeTracerProvider } = require('@opentelemetry/node');
+const RestifyInstrumentation = require('@opentelemetry/instrumentation-restify');
+
+const provider = new NodeTracerProvider();
+provider.register();
+new RestifyInstrumentation();
```
obecny, line 29: trying some unification in examples / readme with `registerInstrumentations` like in core repo (soon in contrib too when plugins will be removed) for example
https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation-http.
Can you please make an example with `registerInstrumentations`

obecny (commented):
overall lgtm, just few issues

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
```diff
@@ -0,0 +1,224 @@
...
+    const instrumentation = this;
+    return function (
+      this: Server,
+      path: any,
+      ...handler: types.NestedRequestHandlers
+    ) {
+      return original.call(
+        this,
+        path,
+        ...instrumentation._handlerPatcher(
```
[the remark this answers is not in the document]
Rauno56: Thanks! :)

plugins/node/opentelemetry-instrumentation-restify/package.json
Rauno56: I get where you are coming - I'd like to avoid adding any deps to instrumentations whenever possible. I would not write a `lodash` utility function myself either, because I would most certainly do a worse job at this than lodash developers. It's a very well-trusted library and lightweight one-function-package.

More importantly: do we actually need this? I'm using it to avoid calling `span.end()` multiple times, but feel like `span.end()` should be noop anyways after the first time it's called.

plugins/node/opentelemetry-instrumentation-restify/package.json
Rauno56: This is outside the scope of this PR, but could we potentially remove the need for keeping this in sync with `./src/version.ts`?
Updated it.

plugins/node/opentelemetry-instrumentation-restify/README.md
Rauno56: Sure! Updated the example.

examples/restify/tracer.js
Rauno56: > why renaming this to Instrumentation?

To say "in this context, this is **the** instrumentation we're introducing to you, dear developer". It also makes less ctrl-f-replacing if the following code is moved somewhere.

> can you also use @opentelemetry/instrumentation-restify

`@opentelemetry/instrumentation-restify` simply does not exist and would be broken until a publish. #422

> @opentelemetry/instrumentation-http the last path probably exists only on yr local env.

Good catch! Fixed that.

Rauno56:
I think I addressed everything except two totally valid points @obecny, brought up, I don't necessarily think are worth changing and would like a second opinion on:

1. __Renaming `c` to `constants`__. Unnecessarily verbose in such a use-case in my view.
2. __Use of `lodash.once`__. I think we should use quality well-known packages and not reimplement utilities, which would be more costly in the long run, especially in the case where some(most?) distros ship tons of instrumentations, used or not, anyways. We cannot avoid bringing in any 3rd party packages and we have done so in the past(`shimmer`, `require-in-the-middle`, etc.).

I don't have a strong technical preference in either of those points, more of a question of how much bikeshedding we want to go through in the long run and for me to calibrate how important maintainers generally think such questions are.

plugins/node/opentelemetry-instrumentation-restify/package.json
obecny: avoiding any dependency in instrumenting has much more sense if you think that any other library can be instrumented as well. We also in general avoid any 3rd party dependency unless this is really necessary and provides a core functionality. Here you just want to prevent the span.end being called twice which can be done without it.

plugins/node/opentelemetry-instrumentation-restify/package.json
obecny: the `version.ts` is being generated based on the version from package.json during compilation, did you create the `version.ts` manually ?

examples/restify/tracer.js
obecny: if you use lerna for that it will work fine pointing to yr local version instead of npm where this package is not released. We are doing like this in our all examples in core and contrib repo and it works fine.

vmarchaud: I don't think renaming it is a good idea, generally users will have multiple instrumentations so they will almost never does this directly

vmarchaud: I think this need be adressed before merging, we discussed this between maintainers and we expect users to be able to just copy/paste example together. Having to rename the import to do so add a friction that isn't needed

plugins/node/opentelemetry-instrumentation-restify/package.json
vmarchaud: > This is outside the scope of this PR, but could we potentially remove the need for keeping this in sync with ./src/version.ts?

As bart said this is automatically generated when publishing new versions, we generally ask them to be up to date for new instrumentations so people doesn't get confused when looking at unpublishe code

vmarchaud:
I agree with @obecny here on both points:
- i prefer verbosity for the sake of readability
	- our project is open source, so we expect a lot of people will review our code having `constants` over `c` make it easy to understand what the variable is specially when reading on github and not within an IDE)
- i prefer to not use a dependency to avoid having few more lines in our code
	- I don't agree with you by comparing `lodash.once` (which is [5 lines long](https://github.com/lodash/lodash/blob/master/before.js#L21) that anyone could write without specific knowledge) and `shimmer`/`require-in-the-middle` which are battle tested with complex problem within NodeJS
	- Specially in this case i don't think its actually needed, the next callback shouldn't be called twice right ? And even if it did, nothing will happens (it will generate a warning log on the diagnostic logger which isn't logged by default) so its safe to remove it altogether

> I don't have a strong technical preference in either of those points, more of a question of how much bikeshedding we want to go through in the long run and for me to calibrate how important maintainers generally think such questions are.

I agree with you on this however: that there are both technical preference. I don't link blocking PR for those kind of choices (so i'll not) but i strongly recommend addressing them.

plugins/node/opentelemetry-instrumentation-restify/package.json
Rauno56: I did create it manually, by copying another instrumentation before I started. Should I remove it?

plugins/node/opentelemetry-instrumentation-restify/package.json
Rauno56: Clear! I'll just remove it then and live with having the warning messages.

plugins/node/opentelemetry-instrumentation-restify/package.json
vmarchaud: Its not really important as long as version are sync between the version and the entry in the package.json, so i dont think so

Rauno56:
> Specially in this case i don't think its actually needed, the next callback shouldn't be called twice right ?

Yes, it is not supposed to, but `restify` doesn't error if it is called twice, just ignores it. There are other cases where it's not called at all. That did make me think about the cases with promises. Added checks for those.

Removed `lodash.once` and renamed `c`.

Thanks a lot @vmarchaud and @obecny for your input. It's highly appreciated! :pray:

Rauno56:
> I don't agree with you by comparing lodash.once (which is 5 lines long that anyone could write without specific knowledge) and shimmer/require-in-the-middle which are battle tested with complex problem within NodeJS

I do wonder whether we should come up with another solution than implementing stuff ourselves if we have a utility in 3 instrumentations? 5? 15?

examples/restify/tracer.js
Rauno56: > if you use lerna for that it will work fine pointing to yr local version instead of npm where this package is not released. We are doing like this in our all examples in core and contrib repo and it works fine.

Examples are not linked. So it will not work until the first publish and will only ever work with the latest published release of the instrumentation. I broke this for consistency, but imo this is not OK.

vmarchaud:
> I do wonder whether we should come up with another solution than implementing stuff ourselves if we have a utility in 3 instrumentations? 5? 15?

We generally have them on the `core` package (https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-core/src/utils) or the instrumentation package itself (https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-instrumentation/src/utils.ts)

vmarchaud (approved):
thanks taking the time to addressing all feedback :)

examples/restify/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "restify-example",
+  "private": true,
+  "version": "0.18.0",
```
obecny, line 4: Nit: 0.15.0 in core we have 0.18.x

obecny (approved):
lgtm, thx for changes

examples/restify/package.json
vmarchaud: Version in examples are not kept in sync in this repo, i think it would be better to do it with lerna instead

Issue #402, opened 2021-03-31
MongoDB Instrumentation causes type error on command callback


jamesalucas:
Please answer these questions before submitting a bug report.

### What version of OpenTelemetry are you using?
"@opentelemetry/instrumentation-mongodb": "0.14.0",
"@opentelemetry/instrumentation": "0.18.2",

### What version of Node are you using?
14.15.4

### What did you do?
Instrumented code using MongoDB, at runtime, after being up for a few seconds the app crashes at mongodb/lib/cmap/connection.js:277 as the callback it attempts to invoke is undefined.

### What did you expect to see?
No TypeError

### What did you see instead?
 TypeError: callback is not a function
     at MessageStream.messageHandler (/app/node_modules/mongodb/lib/cmap/connection.js:277:5)
     at MessageStream.emit (events.js:315:20)
     at processIncomingData (/app/node_modules/mongodb/lib/cmap/message_stream.js:144:12)
     at MessageStream._write (/app/node_modules/mongodb/lib/cmap/message_stream.js:42:5)
     at writeOrBuffer (internal/streams/writable.js:358:12)
     at MessageStream.Writable.write (internal/streams/writable.js:303:10)
     at Socket.ondata (internal/streams/readable.js:719:22)
     at Socket.emit (events.js:315:20)
     at addChunk (internal/streams/readable.js:309:12)
     at readableAddChunk (internal/streams/readable.js:284:9)
 [ERROR] 14:52:57 TypeError: callback is not a function

### Additional context

From debugging, the command being invoked is from `measureRoundTripTime`:

```
rttPinger[kConnection].command('admin.$cmd', { ismaster: 1 }, err => {
    if (err) {
      rttPinger[kConnection] = undefined;
      rttPinger[kRoundTripTime] = 0;
      return;
    }

    measureAndReschedule();
  });
```

The command is { ismaster:1 } and the callback follows, but it seems that the patching of the server command may be passing in the callback incorrectly?

If I change this:
```
if (!currentSpan ||
                    typeof resultHandler !== 'function' ||
                    typeof cmd !== 'object') {
                    return original.call(this, server, ns, cmd, typeof options === 'function' ? callback: options, callback);
                }
```
to:
```
if (!currentSpan ||
                    typeof resultHandler !== 'function' ||
                    typeof cmd !== 'object') {
                    return original.call(this, server, ns, cmd, options, callback);
                }
```

[...]

PR #418, opened 2021-04-11
fix(mongodb): use correct args when calling mongodb driver #402

vmarchaud:
Sadly i wasn't able to find a case which reprocuce the reported error, the original code was wrong so i'm fixing it but i would have liked to be sure to fix the problem :/

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
+++ b/plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
@@ -142,14 +142,11 @@ export class MongoDBInstrumentation extends InstrumentationBase<
           typeof resultHandler !== 'function' ||
           typeof ops !== 'object'
         ) {
-          return original.call(
-            this,
-            server,
-            ns,
-            ops,
-            typeof options === 'function' ? callback : options,
-            callback
-          );
+          if (typeof options === 'function') {
+            return original.call(this, server, ns, ops, options);
+          } else {
+            return original.call(this, server, ns, ops, options, callback);
+          }
         }
         const span = instrumentation.tracer.startSpan(
           `mongodb.${operationName}`,
@@ -165,14 +162,12 @@ export class MongoDBInstrumentation extends InstrumentationBase<
           operationName !== 'insert' ? (ops[0] as any) : undefined
         );
         const patchedCallback = instrumentation._patchEnd(span, resultHandler);
-        return original.call(
-          this,
-          server,
-          ns,
-          ops,
-          typeof options === 'function' ? patchedCallback : options,
-          patchedCallback
-        );
+        // handle when options is the callback to send the correct number of args
+        if (typeof options === 'function') {
+          return original.call(this, server, ns, ops, patchedCallback);
+        } else {
+          return original.call(this, server, ns, ops, options, patchedCallback);
+        }
       };
     };
   }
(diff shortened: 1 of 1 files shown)


plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -197,14 +197,7 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
         ) {
-          return original.call(
-            this,
-            server,
-            ns,
-            cmd,
-            typeof options === 'function' ? callback : options,
-            callback
-          );
+          return original.call(this, server, ns, cmd, options, callback);
```
Flarna, line 200:
```suggestion
          return original.apply(this, arguments);
```
Consider callback is not given by caller. We would see `arguments.length === 4` here but mongodb would see `arguments.length === 5`

dyladan (approved):

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
dyladan: 👍 to this

The types might complain so you might even want to use `Reflect.apply`

ghsyeung:
Just wondering, should the same fix be applied to `_getPatchOperation` as well?

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
vmarchaud: You are right, i fixed it

vmarchaud: I can't use `arguments` because typescript think its type is `IArguments` instead of the concrete type i wrote. I fixed the `arguments.length` issue thought

vmarchaud:
Rebased and addressed comments (cc @Flarna)

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
Flarna: usually it works also to use rest arguments instead relying on `arguments` because rest arguments can be typed as needed.

vmarchaud: I'm not sure to see how we can type rest arguments in this case, is there an example somewhere ? thanks

Flarna: According to ts doc it should work, see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#rest-parameters-with-tuple-types
But not sure if it would really improve the code here as we maybe just move complexity from A to B.

vmarchaud: I see indeed it would be possible but i also don't think its improve the readability of the code. I think the current version is simple enough (even though it can break if they changes the number of arguments, which should not happen since they have been rewriting from scratch the module recently)

plugins/node/opentelemetry-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -142,14 +142,7 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
         ) {
-          return original.call(
-            this,
-            server,
-            ns,
-            ops,
-            typeof options === 'function' ? callback : options,
-            callback
-          );
+          return original.call(this, server, ns, ops, options, callback);
```
Flarna, line 145: shouldn't we the same if as in the other cases also here?

vmarchaud: Done

PR #419, opened 2021-04-11
feat: bunyan instrumentation

seemk:

Adds injection of trace context to bunyan records for improved log correlation as specified by the spec [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#trace-context-in-legacy-formats)

I.e. injects `trace_id`, `span_id`, `trace_flags` to bunyan records.


plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        if (!isSpanContextValid(spanContext)) {
+          return original.apply(this, args);
+        }
+
+        record['trace_id'] = spanContext.traceId;
+        record['span_id'] = spanContext.spanId;
+        record['trace_flags'] = `0${spanContext.traceFlags.toString(16)}`;
+
+        instrumentation.getConfig().logHook!(record, span);
```
blumamir, line 95: It is usually common to surround hook functions with `safeExecuteInTheMiddle`, so if the hook function throws, the exception will be catched, ignored, and app will not crash.

plugins/node/opentelemetry-instrumentation-bunyan/src/types.ts
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Span } from '@opentelemetry/api';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type LogHookFunction = (record: Record<string, any>, span: Span) => void;
```
blumamir, line 21: In many instrumentations the "hook" function signature's first parameter is `span`, and then the other specific datarelated to the hook:
[HTTP](https://github.com/open-telemetry/opentelemetry-js/blob/9371029ee22b2a28c118a80173c2a8990f90255a/packages/opentelemetry-instrumentation-http/src/types.ts#L54)
[ioredis](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1381a9a0f9e45a7dae53239489c68b0de95e2819/plugins/node/opentelemetry-instrumentation-ioredis/src/types.ts#L52)
and all the instrumentation [here](https://github.com/aspecto-io/opentelemetry-ext-js)

Might be worth keeping this structure if you find it appropriate here as well

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
seemk: Although is it good to have silent failures? Of course the instrumentation should not throw anything, but if it is the user code and user code throws, then I'm not so sure :thinking:

seemk: I'm not against it if it is the convention here, but is it The Right Thing :tm:? :smile:

blumamir: In ioredis instrumentation I was asked in the review to log an error in case **hook function** throws:
```js
        safeExecuteInTheMiddle(
          () => config?.responseHook?.(span, cmd.name, cmd.args, result),
          e => {
            if (e) {
              diag.error('ioredis response hook failed', e);
            }
          },
          true
        );
```

As far as I understand, If **`original.apply(this, arguments)`** throws, then `safeExecuteInTheMiddle` should have the parameter `preventThrowingError` set to false, so the exception can propagate back to the caller, and the `onFinish` function is used to close spans or set them to error.

seemk: I guess, that would be better indeed, thanks :slightly_smiling_face: Adding a fix

plugins/node/opentelemetry-instrumentation-bunyan/README.md
```diff
@@ -0,0 +1,71 @@
+# OpenTelemetry instrumentation for bunyan
+
+[![NPM Published Version][npm-img]][npm-url]
+[![dependencies][dependencies-image]][dependencies-url]
+[![devDependencies][devDependencies-image]][devDependencies-url]
+[![Apache License][license-image]][license-image]
+
+This module provides injection of trace context to [`bunyan`](https://www.npmjs.com/package/bunyan).
```
obecny, line 8:
```suggestion
This module provides automatic instrumentation for injection of trace context to [`bunyan`](https://www.npmjs.com/package/bunyan).
```

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  protected init() {
+    return [
+      new InstrumentationNodeModuleDefinition<typeof BunyanLogger>(
+        'bunyan',
+        ['*'],
+        logger => {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const proto = logger.prototype as any;
```
obecny, line 44: isn't the logger a typeof `BunyanLogger` ?

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  getConfig(): BunyanInstrumentationConfig {
+    return this._config;
+  }
+
+  setConfig(config: BunyanInstrumentationConfig) {
+    this._config = Object.assign({ logHook: defaultHook }, config);
+  }
+
+  private _getPatchedEmit() {
```
obecny, line 74: can this have better return typing (if possible) ?

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,112 @@
...
+    return (original: (...args: unknown[]) => void) => {
+      const instrumentation = this;
+      return function patchedEmit(this: BunyanLogger, ...args: unknown[]) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const record = args[0] as any;
```
obecny, line 79:
```suggestion
        const record = Recod<string, string>;
``` 
?

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        if (!isSpanContextValid(spanContext)) {
+          return original.apply(this, args);
+        }
+
+        record['trace_id'] = spanContext.traceId;
+        record['span_id'] = spanContext.spanId;
+        record['trace_flags'] = `0${spanContext.traceFlags.toString(16)}`;
+
+        safeExecuteInTheMiddle(
```
obecny, line 96:
```suggestion
  const logHook = instrumentation.getConfig().logHook;
  if (typeof logHook === 'function') {
        safeExecuteInTheMiddle(
        ....
```

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
seemk: Fixed, good point

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
seemk: Added a check for existence of the log hook

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
seemk: Think it's kinda awkward to add the return type here, the original function is not exposed via types. :thinking:

plugins/node/opentelemetry-instrumentation-bunyan/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "version:update": "node ../../../scripts/version-update.js",
+    "compile": "npm run version:update && tsc -p ."
+  },
+  "keywords": [
+    "opentelemetry",
+    "logging",
+    "nodejs",
+    "tracing",
+    "profiling",
+    "instrumentation"
```
obecny, line 26: probably adding word `bunyan` might be good idea, wdyt ?

plugins/node/opentelemetry-instrumentation-bunyan/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@opentelemetry/instrumentation-bunyan",
+  "version": "0.15.0",
+  "description": "OpenTelemetry instrumentation for bunyan",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "repository": "open-telemetry/opentelemetry-js-contrib",
+  "scripts": {
+    "test": "nyc ts-mocha -p tsconfig.json 'test/**/*.test.ts'",
```
obecny, line 9: nit: keep it alphabetically

obecny (approved):

plugins/node/opentelemetry-instrumentation-bunyan/package.json
seemk: Thanks, added it :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,123 @@
...
+
+        return original.apply(this, args);
+      };
+    };
+  }
+
+  private _callHook(span: Span, record: Record<string, string>) {
+    const hook = this.getConfig().logHook;
+
+    if (!hook) {
```
obecny, line 109: just last minor things
```suggestion
    if (typeof hook !== 'function') {
```

seemk: Done!

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
```diff
@@ -0,0 +1,123 @@
...
+> {
+  constructor(config: BunyanInstrumentationConfig = {}) {
+    super('@opentelemetry/instrumentation-bunyan', VERSION, config);
+  }
+
+  protected init() {
+    return [
+      new InstrumentationNodeModuleDefinition<typeof BunyanLogger>(
+        'bunyan',
+        ['*'],
```
vmarchaud, line 45: would be safer to use `<2` since they might change internals between majors version so the instrumentation doesn't break user applications in the future

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
vmarchaud: @seemk You marked the conversion as resolved without answering nor modifying anything, is there any reason why this isn't typed ?

seemk: > @seemk You marked the conversion as resolved without answering nor modifying anything, is there any reason why this isn't typed ?

My bad, thought I replied to this.

But the issue is that `logger` is indeed typeof `BunyanLogger`, however `prototype` does not expose `_emit` in the types thus the need for any.

plugins/node/opentelemetry-instrumentation-bunyan/src/bunyan.ts
seemk: Added it, thanks

vmarchaud (approved):
lgtm

PR #421, opened 2021-04-13
feat: add xray propagator

anuraaga:

- Unable to propagate context through AWS services (e.g., S3, SQS, Lambda)

- Add AWS X-Ray propagator for AWS users


anuraaga:
In case it helps this has been reviewed previously in #163 - at the time it wasn't known how to handle vendor propagators but since then we have the green light to add to contrib repos.

propagators/opentelemetry-propagator-aws-xray/README.md
```diff
@@ -0,0 +1,77 @@
...
+#### Sampled - The sampling decision*
+
+* Defined in the AWS X-Ray specification as a tri-state field, with "0", "1" and "?" as valid values. Only "0" and "1" are used in this propagator.
+* Populated from the [OpenTelemetry trace flags](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/overview.md#spancontext).
+
+Example of usage (will be added after published):
+### Useful links
+- For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
+- For more about OpenTelemetry JavaScript: <https://github.com/open-telemetry/opentelemetry-js>
+- For help or feedback on this project, join us on [gitter][gitter-url]
```
Flarna, line 63: I think we removed gitter everywhere

propagators/opentelemetry-propagator-aws-xray/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+Example header:`X-Amzn-Trace-Id: Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1`
+
+The header consists of three parts: the root trace ID, the parent ID and the sampling decision. The root is required, whereas the parent ID and sampling decision are optional.
+
+#### Root - The AWS X-Ray format trace ID
+
+* Format: (spec-version)-(timestamp)-(UUID)
+    * spec_version - The version of the AWS X-Ray header format. Currently, only "1" is valid.
+    * timestamp - 32-bit number in base16 format, encoded from time(second) when created. Populated by taking the first 8 characters of the OpenTelemetry trace ID.
```
Flarna, line 44: Does this mean the `AwsXRayIdGenerator` is mandatory for all monitored entities in case x-ray propagation is used?
I assume yes otherwise the timestamp would be garbage.

propagators/opentelemetry-propagator-aws-xray/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@opentelemetry/propagator-aws-xray",
+  "version": "0.15.0",
+  "description": "OpenTelemetry AWS Xray propagator provides context propagation for systems that are using AWS X-Ray format.",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "repository": "open-telemetry/opentelemetry-js-contrib",
+  "scripts": {
+    "precompile": "tsc --version",
+    "compile": "tsc --build",
```
Flarna, line 10: in plugins `npm run version:update && tsc -p .` is used. Not sure what is correct here or if it is don't care anyway.

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  fields(): string[] {
+    return [AWSXRAY_TRACE_ID_HEADER];
+  }
+
+  private getSpanContextFromHeader(
+    carrier: unknown,
+    getter: TextMapGetter
+  ): SpanContext {
+    const traceHeader = getter.get(carrier, AWSXRAY_TRACE_ID_HEADER);
```
Flarna, line 96: I think this wont work for NodeJs HTTP headers because all incoming header names are converted to lowercase but `AWSXRAY_TRACE_ID_HEADER` is mixed case.
Or is it intended that users use a custom getter here? If yes this should be mentioned in docs.

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,194 @@
...
+        pos = traceHeader.length;
+      }
+      const equalsIndex = trimmedPart.indexOf(KV_DELIMITER);
+
+      const value = trimmedPart.substring(equalsIndex + 1);
+
+      if (trimmedPart.startsWith(TRACE_ID_KEY)) {
+        parsedTraceId = AWSXRayPropagator._parseTraceId(value);
+      }
+      if (trimmedPart.startsWith(PARENT_ID_KEY)) {
```
Flarna, line 122:
```suggestion
      else if (trimmedPart.startsWith(PARENT_ID_KEY)) {
```

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,194 @@
...
+        parsedSpanId = AWSXRayPropagator._parseSpanId(value);
+      }
+      if (trimmedPart.startsWith(SAMPLED_FLAG_KEY)) {
```
Flarna, line 125:
```suggestion
      else if (trimmedPart.startsWith(SAMPLED_FLAG_KEY)) {
```

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -0,0 +1,194 @@
...
+        parsedTraceFlags = AWSXRayPropagator._parseTraceFlag(value);
+      }
+    }
+    if (parsedTraceFlags === null) {
+      return INVALID_SPAN_CONTEXT;
```
Flarna, line 130: According to readme the sampled part is optional but it seems it is required here.

propagators/opentelemetry-propagator-aws-xray/README.md
```diff
@@ -0,0 +1,77 @@
+# OpenTelemetry Propagator AWS X-Ray
+
+[![NPM Published Version][npm-img]][npm-url]
+[![dependencies][dependencies-image]][dependencies-url]
+[![devDependencies][devDependencies-image]][devDependencies-url]
+[![Apache License][license-image]][license-image]
+
+The OpenTelemetry Propagator for AWS X-Ray provides HTTP header propagation for systems that are using AWS `X-Amzn-Trace-Id` format.
+This propagator translates the OpenTelemetry SpanContext into the equivalent AWS header format, for use with the OpenTelemetry JS SDK.
```
Flarna, line 9: I think it should be mentioned that `TraceState` is not propagated

propagators/opentelemetry-propagator-aws-xray/test/AWSXRayPropagator.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      );
+
+      assert.deepStrictEqual(carrier, {});
+    });
+  });
+
+  describe('.extract()', () => {
+    it('extract nothing from context', () => {
+      // context remains untouched
+      assert.deepStrictEqual(
```
Flarna, line 124:
```suggestion
      assert.strictEqual(
```
we should verify that we return the original object not one which looks like the original.

propagators/opentelemetry-propagator-aws-xray/package.json
anuraaga: I copied this from the `-ot-trace` module so figure sticking with it is ok

propagators/opentelemetry-propagator-aws-xray/README.md
anuraaga: Nah - for propagation it's just an opaque value so it doesn't need to be a timestamp, if a user doesn't use X-Ray they would still use this propagator with random IDs to propagate through AWS services like SQS.

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
anuraaga: Hmm - I figured issues about lowercase vs uppercase would be handled at a different layer than the propagator. I assume that means it's ok for me to use all lowercase in this propagator, similar to the b3 one

https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-propagator-b3/src/constants.ts#L21

So switched to it

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
anuraaga: I have removed the text about it, since with OpenTelemetry all parts are required, we basically require them here (or start a new trace otherwise)

anuraaga (commented):
Thanks, updated

propagators/opentelemetry-propagator-aws-xray/README.md
Flarna: Just for my interest, not directly effecting the PR.
If I understand this correct users have to use this propagator if they use AWS services because W3C tracecontext is not propagate there. But tracestate is lost. Are there any plans to preserve tracestate?

anuraaga: Good question - we don't currently have a spec for modeling tracestate with the xray propagator. We should probably be able to do so at some point but haven't prioritized it - when we have a spec we'd apply it to all languages. Before tracestate, we're focusing on baggage which I'll add in a followup.

anuraaga:
codecov delta seems to be a flake related to metrics, not this PR

Flarna (approved):

propagators/opentelemetry-propagator-aws-xray/package.json
vmarchaud: Well generally the `version.ts` is used by instrumentations (because its required in the otel protocol). Since this package is not, its not needed in both the compile script or the `script.ts` file (so you can remove it)

vmarchaud (approved):

PR #424, opened 2021-04-15
chore: propagate context on AWS Lambda Instrumentation

gregoryfranklin:
Context is not propagated.  This means the span created by the AWS Lambda Instrumentation is not part of the same trace as spans created within the lambda function.

Add the span created by the AWS Lambda Instrumentation to the context that the lamdba function is run with.


anuraaga (approved):
Oops, thanks!

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/aws-lambda.ts
```diff
@@ -24,7 +24,9 @@ import {
   safeExecuteInTheMiddle,
 } from '@opentelemetry/instrumentation';
 import {
+  context as otelcontext,
```
obecny, line 27: a typo ?
```suggestion
  context as otelContext,
```

obecny (approved):
lgtm

dyladan (approved):

PR #425, opened 2021-04-15
feat: winston instrumentation

seemk:

Similarly to bunyan (#419), adds trace context to log metadata fields.

Injects `trace_id`, `span_id`, `trace_flags` to winston log metadata.

Note that testing with older winston versions (`1.1.2` and `2.4.4`) was done manually. To help with testing in the future, see https://github.com/open-telemetry/opentelemetry-js/issues/2110


vmarchaud (approved):

obecny (approved):
This is really nice piece of work, thx for contribution 

Would you be able to include the screenshots to show how the spans look like, can be in this or separate PR ?

seemk:
> This is really nice piece of work, thx for contribution
> 
> Would you be able to include the screenshots to show how the spans look like, can be in this or separate PR ?

Perhaps I understood the question wrong, but the instrumentation itself does not generate spans, only injects the active span context to log records. E.g. an example:

```
console.log(getSpanContext(context.active()));
logger.info('foobar');
logger.info('foobar', { something: 42 });
```

Without instrumentation:
image

With instrumentation:
image

dyladan (approved):

PR #432, opened 2021-04-19
feat: pino instrumentation

seemk:

Adds automatic injection of trace context to pino (`5.x` and `6.x`) similarly to bunyan and winston instrumentations.


naseemkullah (approved):

naseemkullah:
cc @matteocollina if you'd like to take look

naseemkullah:
cc @mcollina if you'd like to take a look

naseemkullah:
cc @mcollina @davidmarkclements if you'd like to take a look

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
```diff
@@ -0,0 +1,230 @@
...
+      new InstrumentationNodeModuleDefinition<Pino>(
+        'pino',
+        pinoVersions,
+        (pinoModule, version) => {
+          if (version === undefined) {
+            return pinoModule;
+          }
+
+          if (semver.satisfies(version, pino5Versions)) {
+            const proto = Object.getPrototypeOf(pinoModule());
```
vmarchaud, line 64: do the constructor have a side effect ? I mean we are creating a new instance here which can be problematic

seemk: Afaik it does not, it's just a factory function: https://github.com/pinojs/pino/blob/v5.17.0/pino.js#L63 so no global state gets set.

vmarchaud: I think what would make more sense is to retrieve the `lib/proto` file to patch it, WDYT ?

seemk: The thing with this is that the file hook gets called first, but in order to patch the prototype I need the value of `writeSym`. `lib/proto` gets loaded before `lib/symbols`, not sure if there is a non-hackish way to do it. Will give it some thought :slightly_smiling_face:

vmarchaud: Indeed, i don't think its possible then. Might be worth checking though

dyladan (approved):

mcollina:
I would like to ask why would you need to monkeypatch pino? Pino is inherently sync in how it logs, so there should not be no need to monkeypatch anything here.

(streams are async, but those are separate).

mcollina:
could this be turned off anyway?

dyladan:
I wonder if "Instrumentation" is a bad name for this, since it isn't actually instrumenting anything? Maybe we should use some other name like `LogInjection` or something like that?

dyladan:
Also, pino supports "mixins" as a first party API. Maybe we should just release an official opentelemetry pino mixin that injects the span data. It would likely be a much more stable interface than patching internal methods.

dyladan:
For now I'm retracting my approval until we are sure this is the direction we want to go.

seemk:
I don't get it, it's meant for automatic instrumentation and is completely optional. If you want to use a mixin and have separate configuration for this, fine.

Besides, the spec even has a section for auto instrumentation about this exact same use case: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#auto-instrumenting-existing-logging

Other languages already have similar mechanisms, e.g. [Java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/log4j/log4j-2.13.2/library)

@naseemkullah is completely correct on this one.

> I would like to ask why would you need to monkeypatch pino? Pino is inherently sync in how it logs, so there should not be no need to monkeypatch anything here.
> 
> (streams are async, but those are separate).

On a side note, what does it being sync or async have to do with this PR? It's about achieving log correlation.

Flarna:
To avoid monkey patching a pino API to register global mixins would be needed.

vmarchaud:
> I wonder if "Instrumentation" is a bad name for this, since it isn't actually instrumenting anything? Maybe we should use some other name like LogInjection or something like that?

i would prefer to keep them as instrumentation, someone could decide to add metrics for how much logs you are emitting etc.

mcollina:
The feature we added for supporting this use case was the `mixin` as @Flarna said! Can you please check if it's sufficient for the use case? In _theory_ you can use it to add the relevant data to all the log lines: it was added specifically to support the APM use case.

If mixin is enough, I would recommend either:

1. you just wrap the pino factory and add a `mixin()` option. A bit of care must be taken so that user-provided mixins will still be supported.
2. add the concept of a global `mixin` to pino itself and you just use that in this library.

I do not have much time to work on this right now but happy to review. I'm ok with both approach with a slight preference to 1.

seemk:
> The feature we added for supporting this use case was the `mixin` as @Flarna said! Can you please check if it's sufficient for the use case? In _theory_ you can use it to add the relevant data to all the log lines: it was added specifically to support the APM use case.
> 
> If mixin is enough, I would recommend either:
> 
> 1. you just wrap the pino factory and add a `mixin()` option. A bit of care must be taken so that user-provided mixins will still be supported.
> 2. add the concept of a global `mixin` to pino itself and you just use that in this library.
> 
> I do not have much time to work on this right now but happy to review. I'm ok with both approach with a slight preference to 1.

I'm all for option 1, but with our current instrumentation API how can we achieve this? I am perhaps missing something here.

naseemkullah:
> > The feature we added for supporting this use case was the `mixin` as @Flarna said! Can you please check if it's sufficient for the use case? In _theory_ you can use it to add the relevant data to all the log lines: it was added specifically to support the APM use case.
> > If mixin is enough, I would recommend either:
> > 
> > 1. you just wrap the pino factory and add a `mixin()` option. A bit of care must be taken so that user-provided mixins will still be supported.
> > 2. add the concept of a global `mixin` to pino itself and you just use that in this library.
> > 
> > I do not have much time to work on this right now but happy to review. I'm ok with both approach with a slight preference to 1.
> 
> I'm all for option 1, but with our current instrumentation API how can we achieve this? I am perhaps missing something here.
> 
> Edit: Can return the wrapper inside `patch` I guess. Reworking this a bit

You probably know this but I can confirm that mixin works great with the following option passed in to  pino :

```js
      mixin: () => (getSpan(context.active())?.context() ?? {}),
```

(though its missing the snakecasing of object keys)

Flarna:
Personally I'm in favor of option 2 as it requires no monkey patching of pino. Noone knows if we will be ever able to reliable monkey patch ESM modules therefore getting public APIs is future proof.

But clearly this can't happen within this PR.

dyladan:
>I don't get it, it's meant for automatic instrumentation and is completely optional. If you want to use a mixin and have separate configuration for this, fine.

**Instrument (verb)**: _equip (something) with measuring instruments._

Since we are not measuring anything, we are not instrumenting the module.

>Besides, the spec even has a section for auto instrumentation about this exact same use case: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#auto-instrumenting-existing-logging

If spec says this it's probably fine for now. I would also prefer to take advantage of public, supported APIs. As @Flarna says we will not be able to monkeypatch once ESM is widely supported.

I generally support option 2, but since it relies on options which don't yet exist 1 is probably a better solution at the moment.

dyladan:
I'm going to approve anyways as this can always be reworked later.

dyladan (approved):

seemk:
Changing it to mixin patching, don't merge yet. This will bring the min version to 5.14, but should be fine

dyladan:
Adding blocked label until you finish

seemk:
> Adding blocked label until you finish

@dyladan Done. Added additional tests regarding options and mixins handling. Upped the required version to 5.14.0 and ran both 5.14.0 and 6.x through an example app.

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
seemk: Changed it to mixins patching now, should be simpler.

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
```diff
@@ -0,0 +1,171 @@
...
+          return patchedPino;
+        },
+        () => {
+          this._isEnabled = false;
+        }
+      ),
+    ];
+  }
+
+  getConfig(): PinoInstrumentationConfig {
```
dyladan, line 85: Why public getter/setter for config here?

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
```diff
@@ -0,0 +1,171 @@
...
+              const optsOrStream = args[0] as any;
+              if (
+                typeof optsOrStream === 'string' ||
+                (optsOrStream !== undefined &&
+                  typeof optsOrStream['write'] === 'function')
+              ) {
+                args.splice(0, 0, {
+                  mixin: instrumentation._getMixinFunction(),
+                });
+                return pinoModule(...(args as Parameters<Pino>));
```
dyladan, line 65: This can be done in a single step I think

```suggestion
                return pinoModule({ mixin: instrumentation._getMixinFunction() }, ...args);
```

dyladan (approved):
Looks good

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
seemk: Good point, although this won't compile due to pino taking 0-2 args and `...(args as Parameters<Pino>)` already expands to 2 args and it can't be untyped :frowning:

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
seemk: I have found it useful sometimes to change the instrumentation's config on the fly. In this case changing the `logHook`, so it's not completely set on creation :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  enable() {
+    super.enable();
+    this._isEnabled = true;
+  }
+
+  disable() {
+    super.disable();
+    this._isEnabled = false;
```
obecny, line 100: hmm we already have such functionality, can you add todo here to remove it once the base class will be changed. I created an issue for that -> https://github.com/open-telemetry/opentelemetry-js/issues/2131

obecny (approved):
lgtm, thx for contribution, one comment

mcollina (approved):
lgtm, really good work!

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
seemk: Added a TODO, found it kinda awkward as well :smile:

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
blumamir: We are under the case of `args.length == 1`, so it can also be written as `return pinoModule({ mixin: instrumentation._getMixinFunction() }, args[0])` maybe?

blumamir (approved):

plugins/node/opentelemetry-instrumentation-pino/src/pino.ts
```diff
@@ -0,0 +1,172 @@
...
+              return pinoModule({ mixin: instrumentation._getMixinFunction() });
+            }
+
+            if (args.length == 1) {
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              const optsOrStream = args[0] as any;
+              if (
+                typeof optsOrStream === 'string' ||
+                (optsOrStream !== undefined &&
+                  typeof optsOrStream['write'] === 'function')
```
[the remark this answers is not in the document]
seemk: Indeed it can. Changed it, thanks :slightly_smiling_face:

PR #437, opened 2021-04-21
chore: clean some unused deps

Rauno56:
Net negative PR. Yaay! :tada: 

See commit messages. Mainly:

- `ts-node` which is a dep in `ts-mocha` and
- `@opentelemetry/core` in plugins, which, I believe is left there from debugging/testing/copy-pasting.

Definitely didn't catch all of them especially many occurrences of:

* `tslint-consistent-codestyle` and `tslint-microsoft-contrib` because I thought those might be magically loaded or considered somehow,
* `@types/*`, because IDEs load those magically,
* karma plugins, of which I wasn't sure either how they're loaded.

[`depcheck`](https://www.npmjs.com/package/depcheck) helped.


plugins/node/opentelemetry-instrumentation-hapi/package.json
```diff
@@ -42,7 +42,6 @@
   },
   "devDependencies": {
     "@hapi/hapi": "20.0.1",
-    "@hapi/podium": "4.1.1",
```
Flarna, line 45: I guess this needs to be kept here to avoid build problems caused by incompatible type definitions in various hapi components (see https://github.com/hapijs/hapi/issues/4240).

Flarna (approved):

plugins/node/opentelemetry-instrumentation-hapi/package.json
Rauno56: Thanks for the pointer! I would've never figured that out.

plugins/node/opentelemetry-instrumentation-hapi/README.md
```diff
@@ -68,6 +68,10 @@ registerInstrumentations({
 
 See [examples/hapi](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/hapi) for a short example using Hapi
 
+<!-- 
+The dependency of `@hapi/podium@4.1.1` is required to force the compatible type declarations. See: https://github.com/hapijs/hapi/issues/4240
```
Flarna, line 72:
```suggestion
The dev dependency of `@hapi/podium@4.1.1` is required to force the compatible type declarations. See: https://github.com/hapijs/hapi/issues/4240
```

vmarchaud (approved):

obecny (approved):
lgtm, nice findings

naseemkullah (approved):

dyladan (approved):

plugins/node/opentelemetry-instrumentation-hapi/package.json
dyladan: it would be nice if we could add comments in package.json

Issue #420, opened 2021-04-12
"host-metrics" should collect system (os) cpu metrics to comply with spec


blumamir:
## Edit:
This issue is for modifying the ["@opentelemetry/host-metrics" instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/opentelemetry-host-metrics) to use os metrics instead of process metrics, according to the specification for [Semantic Conventions for System Metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#semantic-conventions-for-system-metrics).

### What version of OpenTelemetry are you using?
v0.18.2

### What version of Node are you using?
v12.18.3

### What did you do?
I was playing with the `host-metrics` package and the metrics it produces.
I may be missing something, but it looks like cpu metrics are for the current process (`process.cpuUsage()` - "returns the user and system CPU time usage of the current process"), while the memory metrics are total os values (`os.totalmem()` - "Returns the total amount of system memory in bytes as an integer.").

Is it intentional? from [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/1afab39e5658f807315abf2f3256809293bfd421/specification/metrics/semantic_conventions/system-metrics.md#systemcpu---processor-metrics) it looks like they should both be system metrics.

obecny:
Nice spotted. When I implemented this for first time it wasn't yet spec'd, it had then few iterations and from what I see it should be probably using the `os.cpu()` instead of `process.cpuUsage()`.

blumamir:
> Nice spotted. When I implemented this for first time it wasn't yet spec'd, it had then few iterations and from what I see it should be probably using the `os.cpu()` instead of `process.cpuUsage()`.

@obecny thanks for looking into it, edited the issue. 

Maybe we can add "up-for-grabs" and "good-first-issue" labels?

svrnm:
I provided a PR for this, please let me know if this looks good or needs some changes.

PR #438, opened 2021-04-22
fix: "host-metrics" should collect system (os) cpu metrics to comply …

svrnm:
This is a fix for the problem described at #420 

This fix replaces `process.cpuUsage()` with `os.cpus()`. Since the latter returns cpu details in milliseconds per core, this change also required an update to report metrics on a cpu basis (which is also within the spec) and to convert from milliseconds instead of microseconds. Additionally interrupt and nice metrics are added. Tests have been updated accordingly.


vmarchaud (approved):

obecny (approved):
lgtm, thx for those changes

vmarchaud:
@svrnm Thanks for your contribution !

PR #441, opened 2021-04-22
feat(instrumentation-document-load): documentLoad attributes enhancement

kkruk-sumo:

This PR answers questions:
- which website has been loaded (added `http.url` ~and `page_title`~),
- who visited it (added `http.user_agent`).


plugins/web/opentelemetry-instrumentation-document-load/test/documentLoad.test.ts
```diff
@@ -492,31 +499,38 @@ describe('DocumentLoad Instrumentation', () => {
...
-        const rsEvents = rootSpan.events;
+        const fetchSpan = exporter.getFinishedSpans()[0] as ReadableSpan;
+        const rootSpan = exporter.getFinishedSpans()[1] as ReadableSpan;
         const fsEvents = fetchSpan.events;
+        const rsEvents = rootSpan.events;
 
-        assert.strictEqual(rootSpan.name, 'documentFetch');
-        assert.strictEqual(fetchSpan.name, 'documentLoad');
+        assert.strictEqual(fetchSpan.name, 'documentFetch');
+        assert.strictEqual(rootSpan.name, 'documentLoad');
```
kkruk-sumo, line 508: `documentLoad` span is the root one so I think previous names were mistaken?

kkruk-sumo, line 508: `documentLoad` span is the root one so I think previous names were mistaken?

plugins/web/opentelemetry-instrumentation-document-load/src/documentLoad.ts
```diff
@@ -115,6 +115,10 @@ export class DocumentLoadInstrumentation extends InstrumentationBase<unknown> {
         }
       });
 
+      rootSpan.setAttribute(HttpAttribute.HTTP_URL, location.href);
+      rootSpan.setAttribute(HttpAttribute.HTTP_USER_AGENT, navigator.userAgent);
+      rootSpan.setAttribute(AttributeNames.PAGE_TITLE, document.title);
```
johnbley, line 120: Not sure about this one... this could veer into PII or other sensitive info.  We generally take pains not to collect things like this out of the box (e.g., the use of xpaths rather than labels/text for clicked elements in `user-interaction`.

kkruk-sumo: Got it. I removed the page title.

obecny: I agree with page title it might contain the user name or name and surname etc.  With regards to `location.href` I think this is too much information it will also produce also very high cardinality. On the other hand the page load in majority of cases in SPA will be connected with the same address, unless you have old style page where each subpage is loaded. But then probably some mechanism for lowering the cardinality should be added similar it is done in express / http or for xmlhttprequest (we only save host). Can you please tell the real use case why do think it should be useful ?

kkruk-sumo: Right now I don't actually know which pages of a website is slow or loads more resources than others. I think it's a case for static websites where URL is predictable and different websites can produce significantly different traces (other resources, scripts, load times). In case of SPA applications there can occur high cardinality like random hashes in URLs but documentLoad spans will be generated only when page opens.
In XHR instrumentation I see the requested url is directly saved as the `http.url` attribute so there can occur the same problem (common case with adding timestamp as query param to prevent caching). So is preventing high cardinality in span attributes very crucial?

dyladan: In span attributes data can be high cardinality. In names it is low cardinality. You are correct `http.url` is already used in other areas. I think it is pretty well accepted these days too that URL is not private or PII typically.

dyladan:
Approved this user to run tests in our CI

plugins/web/opentelemetry-instrumentation-document-load/src/documentLoad.ts
johnbley: I'm fine with `location.href` in concept - Daniel's right on all counts, though I would point out that we (Splunk) apply it to *all* spans in our product and you might consider doing that too (e.g., in a `SpanProcessor`).  It does seem odd to have it just on one bit of instrumentation.

obecny (approved):
lgtm

kkruk-sumo:
> Approved this user to run tests in our CI

@dyladan I had to merge `main`. Could you approve CI for me again? Thanks :)

plugins/web/opentelemetry-instrumentation-document-load/src/documentLoad.ts
kkruk-sumo: @johnbley That's an interesting idea. Is it also called `http.url`? Did you think about proposing it to the standard (maybe `@opentelemetry/web`?). When I think about it, adding `location.href` by default makes a lot of sense.

dyladan (approved):

Issue #292, opened 2020-12-23
NodeJS TLS library


maxmil7:
**Is your plugin request related to a problem? Please describe.**
It would be good to autoinstrument functions in NodeJS TLS library to capture granular information related to establishing secure socket connections

For eg. The plugin can wrap the [tls.connect](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) API to create a span whenever a TLS handshake is successfully completed

When this plugin is combined with the https, dns and Net plugins it will be possible to get detailed trace information for an outgoing https request such as the type of the request, dns lookup time, socket connect time and TLS handshake time

**Is it applicable for Node or Browser or both**
NodeJS

**Do you expect this plugin to be commonly used**
Since this would autoinstrument core functionality, I believe it would be commonly used

**What version of plugin are you interested in using**
Versions:

**Additional context**

- **Is there a reference you could point for the well-defined lifecycle methods**

PR #447, opened 2021-04-23
feat(instrumentation-tls): wrap tls.connect API

svrnm:
This PR is addressing the instrumentation request at #292 

* Add plugin opentelemetry-instrumentation-tls 
* Add example "network" that combines http, dns, tls & net packages:

Screenshot 2021-04-23 at 09 46 34

## Open Questions
* For whatever reason I needed to use `prependOnceListener` to get the instrumentation for the `secureConnect` event working. I didn't run into any issues with that, but maybe someone can double check?

* This package has a kind of dependency on the "net" plugin. Since both instrument the connect method of the socket I had to add a workaround to make both work at the same time. I am not super happy with that solution: I check in the isWrapped conditional for the name of the existing wrapper and if it is 'patchedConnect' I save it and apply it later in a "chained" fashion. Is there already an approach for dependencies? 

```javascript
          if (isWrapped(moduleExports.TLSSocket.prototype.connect)) {
            if (
              moduleExports.TLSSocket.prototype.connect.name ===
              'patchedConnect'
            ) {
              netPatchedConnect = moduleExports.TLSSocket.prototype.connect;
            }
            this._unwrap(moduleExports.TLSSocket.prototype, 'connect');
          }
...
return safeExecuteInTheMiddle(
          () =>
            netPatchedConnect
              ? netPatchedConnect.apply(this, args)
              : original.apply(this, args),
...
```


plugins/node/opentelemetry-instrumentation-tls/test/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+export const PORT = 42124;
+export const HOST = 'localhost';
+
+export function assertSpanKind(span: ReadableSpan) {
+  assert.strictEqual(span.kind, SpanKind.INTERNAL);
+}
+
+export function assertTLSSpan(span: ReadableSpan, socket: TLSSocket) {
+  assertSpanKind(span);
+  /*assertAttrib(span, GeneralAttribute.NET_TRANSPORT, GeneralAttribute.IP_TCP);
```
obecny, line 30: remove comments

examples/network/client.js
```diff
@@ -0,0 +1,47 @@
...
+registerInstrumentations({
+  instrumentations: [
+    new NetInstrumentation(),
+    new TLSInstrumentation(),
+    new HttpInstrumentation(),
+    new DnsInstrumentation({
+      // Avoid dns lookup loop with http zipkin calls
+      ignoreHostnames: ['localhost'],
+    }),
+    // new TLSInstrumentation(),
```
obecny, line 35: remove

obecny (commented):
first pass, why the test coverage went down so much ?

obecny:
>This package has a kind of dependency on the "net" plugin. 

Maybe consider adding the tls under net plugin ?

obecny:
Patching the same thing twice for different instrumentations with shimmer it is tricky as unpatching cannot be done then 
properly in different order
plugin 1 patch -> plugin 2 patch -> plugin 2 unpatch -> plugin 1 unpatch / ok
plugin 1 patch -> plugin 2 patch -> plugin 1 unpatch // error

@open-telemetry/javascript-approvers replacing shimmer sooner or later is inevitable. I have mentioned this year ago, worth to mention it again, it will work fine with `mpwrapper` for example, allowing for such things without any problem

svrnm:
> Maybe consider adding the tls under net plugin ?
@obecny I'll give it a try. Checking if a Socket is an instance of TSLSocket and then hooking into it might be the better approach.

svrnm:
@obecny I figured out a way to combine TLS into the .net package (like http instrumentation does both http and https) 

I am not 100% sure how this works: the attributes for TLS (like cipher and certificate details) are not unique to Node.JS of course. If I understand it correctly, they might be a candidate for semantic-attributes, which are defined through the specification, right?

plugins/node/opentelemetry-instrumentation-net/src/net.ts
```diff
@@ -69,11 +70,11 @@ export class NetInstrumentation extends InstrumentationBase<Net> {
       return function patchedConnect(this: Socket, ...args: unknown[]) {
         const options = getNormalizedArgs(args);
 
-        const span = options
-          ? options.path
-            ? plugin._startIpcSpan(options, this)
-            : plugin._startTcpSpan(options, this)
-          : plugin._startGenericSpan(this);
+        // const span = plugin._startSpan(options, this)
```
vmarchaud, line 73: remove comment

vmarchaud:
> I am not 100% sure how this works: the attributes for TLS (like cipher and certificate details) are not unique to Node.JS of course. If I understand it correctly, they might be a candidate for semantic-attributes, which are defined through the specification, right?

There indeed defined at the spec level (here for [trace](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) and [here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions) for resources) which we then use to code-gen our package ([here](https://github.com/open-telemetry/opentelemetry-js/blob/main/scripts/semconv/generate.sh)), you should open an issue there to be able to "standardize" them for all SDKs. In the mean time, i think you can keep the one you used

plugins/node/opentelemetry-instrumentation-net/src/net.ts
svrnm: Removed

vmarchaud (approved):

dyladan:
Tests are timing out after 360 minutes. i'm going to manually trigger them again but if they time out again you probably have something to look into.

svrnm:
> Tests are timing out after 360 minutes. i'm going to manually trigger them again but if they time out again you probably have something to look into.

Thanks. I saw this issue but couldn’t figure out yet why this happens. I’ll try to reproduce it locally and come up with a fix.

plugins/node/opentelemetry-instrumentation-net/src/types.ts
```diff
@@ -28,4 +28,17 @@ export enum SocketEvent {
   CLOSE = 'close',
   CONNECT = 'connect',
   ERROR = 'error',
+  SECURE_CONNECT = 'secureConnect',
+}
+
+export enum TLSAttributes {
```
dyladan, line 34: Are any of these in the Semantic Conventions?

svrnm:
Sorry for triggering the tests multiple times, I assumed no issues with node:10 and node:8, hope it's fixed now...

plugins/node/opentelemetry-instrumentation-net/src/types.ts
svrnm: Not that I am aware of. I checked the specs and didn't find them there

svrnm: Sorry, I was confused. I thought you talk about the socket events, but you mean the TLSAttributes? I actually opened an issue already: https://github.com/open-telemetry/opentelemetry-specification/issues/1652

> can you add a comment that these are not "official"

Yes, I can do that.

I guess they are not the only candidates for that, e.g. the DNS package puts the name that was looked up into net.peer.name, which is probably not correct and their _might_ be a semantic convention for DNS.

Since this is kind of related to https://github.com/open-telemetry/opentelemetry-js/issues/2123: Would it help if I scan js and js-contrib (semi)automatically for attributes that are either in the semantic-specs already or that might be candidates for it?

dyladan: that would be great

plugins/node/opentelemetry-instrumentation-tls/README.md
```diff
@@ -0,0 +1,57 @@
...
+- For help or feedback on this project, join us in [GitHub Discussions][discussions-url]
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[discussions-url]: https://github.com/open-telemetry/opentelemetry-js/discussions
+[license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[dependencies-image]: https://status.david-dm.org/gh/open-telemetry/opentelemetry-js-contrib.svg?path=plugins%2Fnode%2Fopentelemetry-instrumentation-net
```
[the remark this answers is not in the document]
svrnm: Outdated since I moved all instrumentations to opentelemetry-instrumentation-net

plugins/node/opentelemetry-instrumentation-tls/README.md
```diff
@@ -0,0 +1,57 @@
+# OpenTelemetry Net module Instrumentation for Node.js
```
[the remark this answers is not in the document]
svrnm: Outdated since I moved all instrumentations to opentelemetry-instrumentation-net

plugins/node/opentelemetry-instrumentation-tls/src/tls.ts
```diff
@@ -0,0 +1,167 @@
...
+        'tls',
+        ['*'],
+        moduleExports => {
+          diag.debug('Applying patch for tls module');
+          let netPatchedConnect;
+          if (isWrapped(moduleExports.TLSSocket.prototype.connect)) {
+            // If NetInstrumentation was applied already, we unwrap
+            // the connect method and save the "patchedConnect" for later
+            // use (see _getPatchedConnect)
+            if (
```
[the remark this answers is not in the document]
svrnm: Outdated since I moved all instrumentations to opentelemetry-instrumentation-net

plugins/node/opentelemetry-instrumentation-tls/test/connect.test.ts
```diff
@@ -0,0 +1,118 @@
...
+import { TLSInstrumentation } from '../src/tls';
+// import { TLSSocketEvent } from '../src/types';
+import { assertTLSSpan, HOST, PORT, SERVER_CERT, SERVER_KEY } from './utils';
+// import { TLSSocket } from 'tls';
+
+const memoryExporter = new InMemorySpanExporter();
+const provider = new NodeTracerProvider();
+provider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
+
+function getSpan() {
```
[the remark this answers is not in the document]
svrnm: I renamed this to "getTLSSpans()", although getSpan() is still used in the original connect.test.ts for the net package. I would patch this in a separate way.

plugins/node/opentelemetry-instrumentation-net/src/types.ts
svrnm: @dyladan I found a bunch of them. There's four different cases:

1) Use of existing semantic conventions (like PG, see #467)
2) Use of potentially new semantic conventions, I mean things that are used outside of node.js as well (DNS, TLS, graphQL)
3) Use of custom attributes specific to the instrumented library (express, hapi, restify)
4) A corner case are browser specific attributes: while those are "unique" to javascript I would still argue that they should be covered by a semantic convention, since they are still universal

Since this is a discussion unrelated to this ticket, can you give me some guidance where to move this discussion? Open another issue with the specification repo?

dyladan: I would open an issue on this repo to track it at least. If you think some should be spec'd then you should open an issue on the spec repo too and link them.

dyladan (approved):

PR #450, opened 2021-04-26
chore: prefer use of global TracerProvider/MeterProvider

Flarna:
Remove setting of `TracerProvider` in samples where it is actually not needed as the global `TracerProvider` should be preferred on default.

I found it also inconsistent to set a `TracerProvider` but no `MeterProvider`.

Besides that I fixed some warnings in md files and net test on windows.

Refs.: https://github.com/open-telemetry/opentelemetry-js/pull/2127


naseemkullah (approved):

dyladan (approved):

vmarchaud (approved):

obecny (approved):
lgtm

Issue #451, opened 2021-04-26
express instrumentation does not construct span name correctly when ignoring middlewares


seemk:
Example - middlewares activated for a given path:

```javascript
const instrumentation = new ExpressInstrumentation({
  ignoreLayersType: [ExpressLayerType.MIDDLEWARE],
});
const app = express();
app.use('/mw', (req, res, next) => { next(); });
app.get('/mw', (req, res) => { res.send('ok'); });
```

When doing a request to the `/mw` endpoint, the span name will be `GET /mw/mw`.

This is repeated for every middleware on the same path and the final span name is a concatenation of middleware paths + the final route path and can cause some wonky route names, e.g. `GET /mw/mw/mw/mw/mw/mw`

seemk:
Closing it as the fix was merged

PR #452, opened 2021-04-26
fix(express): construct span names correctly with ignored middleware layers

seemk:

When middlewares are ignored the final span name is constructed from a concatenation of middleware paths. More here: #451


seemk:
Hmm, suddenly unrelated CI errors after merging https://github.com/open-telemetry/opentelemetry-js-contrib/commit/5ecccc916e8f4eecaf6629a0350fc672edcba1be

Flarna:
Most likely a side effect of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52509
Maybe updating the dependencies of @opentelemetry/instrumentation-hapi solves this.

dyladan:
build works locally on my machine but seems to complain about hapi/podium only in CI. any ideas? is it building successfully for others locally?

seems to affect all PRs not just this one

Flarna:
a fix should be included in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/455

dyladan:
Any idea why it works locally?

Flarna:
most likely because package-lock.json files are lying around and as a result transitive dependencies are not updated.
if I clean everything it fails locally.

dyladan:
same. i almost regret confirming this because now i can't build locally

plugins/node/opentelemetry-instrumentation-express/test/custom-config.test.ts
```diff
@@ -108,7 +125,51 @@ describe('ExpressInstrumentation', () => {
           .find(span => span.name === 'rootSpan');
         assert.notStrictEqual(exportedRootSpan, undefined);
       });
-      server.close();
+    });
+
+    it('should construct span name correctly', async () => {
```
dyladan, line 130: Please rename this to what is actually tested. All tests should be testing for "correctness"

seemk: Done

dyladan (approved):
Approving but would like @vmarchaud approval before merge (original plugin author)

plugins/node/opentelemetry-instrumentation-express/src/express.ts
```diff
@@ -209,6 +209,9 @@ export class ExpressInstrumentation extends InstrumentationBase<
 
         // verify against the config if the layer should be ignored
         if (isLayerIgnored(metadata.name, type, instrumentation._config)) {
+          if (type === ExpressLayerType.MIDDLEWARE) {
```
vmarchaud, line 212: i'm not sure that will do the trick, it only works in the test because we had ignored the `middleware` type which isn't the default. I think we should prefer detecting that the previous handler was mounted the same path as the current one, WDYT ?

seemk: What do you mean by default? If it's a route handler, then I guess it's fine to have the same path. Afaik the issue only comes to light with middleware layers.

vmarchaud: your code popup the middleware from the stack when its ignored, i was pointing that it might be not the case. in your tests those middleware are ignored so thats work, but that we'll not be the case for end users

seemk: Btw this fix is only for ignored layers, if the middleware layers are not ignored it works fine as is.

vmarchaud: >Btw this fix is only for ignored layers, if the middleware layers are not ignored it works fine as is. 

Thats the thing i don't get, with the following example, the span name is bugged too: 
```ts
// init without ignored layers
const express = new ExpressInstrumentation({})

app.use('/mw', (req, res, next) => {
	  next();
});
	  
app.get('/mw', (req, res) => {
	  res.send('ok');
});

// span name will be `GET /mw/mw` even though it should be `/mw`
```

If so, i think the fix should handle even when the middleware isn't ignored, WDYT ?

seemk: To clarify, this is with the mainline express instrumentation (not the changes in this PR)?

For example, my demo app:

```javascript
const app = express(); 
app.use('/mw1', function mw1(req, res, next) {
  next();
});
app.use('/mw1', function mw1(req, res, next) {
  next();
});
app.get('/mw1', (req, res) => {
  res.send('ok\n');
}); 
```

With the current main branch:
* `new ExpressInstrumentation()`: root span name = `get /mw1`
* `new ExpressInstrumentation({ ignoreLayersType: [ExpressLayerType.MIDDLEWARE] })`: `get /mw1/mw1/mw1`

With this PR:
* `new ExpressInstrumentation()`: root span name = `get /mw1`
* `new ExpressInstrumentation({ ignoreLayersType: [ExpressLayerType.MIDDLEWARE] })`: `get /mw1`

Or am I missing something here? :smile:

vmarchaud: No thats definitly me that forgot how the whole middleware stack was working, i was not understanding why with `new ExpressInstrumentation():` root span name = `get /mw1` in the main branch but i looked up the code and i understand know so everything is fine ! Sorry for bothering ahah

vmarchaud (approved):

vmarchaud:
@seemk Thanks again for taking the time to do this and being patient with me ahah

plugins/node/opentelemetry-instrumentation-express/src/express.ts
seemk: > No thats definitly me that forgot how the whole middleware stack was working, i was not understanding why with `new ExpressInstrumentation():` root span name = `get /mw1` in the main branch but i looked up the code and i understand know so everything is fine ! Sorry for bothering ahah

No problem, better to get it correct!

I was a bit confused, that perhaps I missed something when testing, so thanks for the double checking :slightly_smiling_face:

PR #458, opened 2021-04-27
feat: router instrumentation

Rauno56:

This adds auto instrumentation for [pillarjs/router](https://github.com/pillarjs/router), which is almost identical to the router implementation used in Express.

Ideally, if possible, those two instrumentations should share code(**thoughts on that?**), but this is not currently the case because

1. Router is more low level than Express, so the dependency makes sense to be other way around;
2. Express' instrumentation is currently [not measuring the execution time](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-express#caveats) for handlers correctly.

Instead of wrapping each handler passed to the router, I'm patching two functions on `Layer` to plug into the core of the routers work.


plugins/node/opentelemetry-instrumentation-router/src/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  span?: types.InstrumentationSpan,
+  method?: string,
+  route?: string
+) => {
+  if (
+    typeof method === 'string' &&
+    typeof route === 'string' &&
+    span?.name?.startsWith('HTTP ')
+  ) {
+    (span as types.InstrumentationSpan).updateName(
```
obecny, line 45: why do you have to cast ? it should not be need

plugins/node/opentelemetry-instrumentation-router/test/index.test.ts
```diff
@@ -0,0 +1,269 @@
...
+    router(req, res, err => {
+      if (err) {
+        res.statusCode = 500;
+        res.end(err.message);
+      }
+      if (!res.headersSent) {
+        res.statusCode = 404;
+        res.end('Not Found');
+      }
+      parentSpan && parentSpan.end();
```
obecny, line 97:
```suggestion
      parentSpan?.end();
```

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
```diff
@@ -0,0 +1,211 @@
+/*
```
obecny, line 1: we usually create a file with name from the instrumented library so in this case it would be `router.ts`

plugins/node/opentelemetry-instrumentation-router/test/index.test.ts
```diff
@@ -0,0 +1,269 @@
...
+  );
+  actualSpans.forEach((span, idx) => {
+    const expected = expectedSpans[idx];
+    if (expected === null) return;
+    try {
+      assert.notStrictEqual(span, undefined);
+      assert.notStrictEqual(expected, undefined);
+      assert.strictEqual(span.attributes['router.name'], expected.name);
+      assert.strictEqual(span.attributes['router.type'], expected.type);
+      assert.strictEqual(span.attributes['router.version'], '1.3.5');
```
obecny, line 129: is this mocked, or real value from package.json so after upgrade it will fail ?

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
```diff
@@ -0,0 +1,211 @@
...
+  ) {
+    const fnName = layer.handle.name || undefined;
+    const type = layer.method
+      ? types.LayerType.REQUEST_HANDLER
+      : types.LayerType.MIDDLEWARE;
+    const route = utils.getRoute(req);
+    const spanName =
+      type === types.LayerType.REQUEST_HANDLER
+        ? `request handler - ${route}`
+        : `middleware - ${fnName || '<anonymous>'}`;
```
obecny, line 173: the same variable `fName` is being adding to the spanName and as attribute, whereas in span name it might be set as anonymous. I think in line 165 you could do 
```
const fnName = layer.handle.name || '<anonymous>`;
```
and  get rid of adding this again as attribute.

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (fn.name === 'router' && fn.toString() === constants.ROUTE_ROUTER_FN) {
+    return true;
+  }
+  return false;
+};
+
+export const getRoute = (req: Router.RoutedRequest) =>
+  req.baseUrl + (req.route?.path || '') || '/';
+
+export const renameHttpSpan = (
```
obecny, line 35:
```suggestion
export const renameHttpSpanName = (
```

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
vmarchaud: I'm not sure `router.ts` name would be clearer, we sometimes have `instrumentation` instead so i don't think thats a problem here

Rauno56: I'd like to decrease the boilerplate everyone has to attend to when creating new instrumentations. Package-specific file names only increase that. The package name is already in the file path(`opentelemetry-instrumentation-router`) and `package.json`.

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
Rauno56: Doesn't "rename" already imply changing something's name?

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
Rauno56: Span name is intended to be used for humans, attributes for machines. I'd avoid fiddling with the "raw data"(attributes, meant for filtering, etc) as much as I can leaving it what the real value actually is.

In span name, I think it's fine to display `<anonymous>` because otherwise span name might read unusual having "- " at the end.

What do you think?

examples/router/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "bugs": {
+    "url": "https://github.com/open-telemetry/opentelemetry-js/issues"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^0.18.0",
+    "@opentelemetry/exporter-jaeger": "^0.18.0",
+    "@opentelemetry/exporter-zipkin": "^0.18.0",
+    "@opentelemetry/instrumentation": "^0.18.0",
+    "@opentelemetry/instrumentation-http": "^0.18.0",
+    "@opentelemetry/instrumentation-router": "file:../../plugins/node/opentelemetry-instrumentation-router",
```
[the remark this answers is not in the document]
Rauno56: That breaks `npm i` in the example, but I'll leave it broken as with `restify`.

obecny:
Please avoid force push after 1st review, it is impossible to track changes

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
obecny: you rename http span which sounds weird, because you rename the name of that span, not the span itself

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
obecny: well in yr case if `fnName` is undefined then the span name attribute will be undefined too. After proposed change if you leave setting span name as attribute it will be `anonymous` which at least give some more value then undefined.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
obecny: The whole purpose of having some boilerplate code is that at least it will be unified in some way so any1 new will have less things to learn and the things will be straightforward so If any1 needs to modify something it will be easier. So I see rather a huge benefit in unifying things then leaving this to be different. We usually name the file after the instrumented package, if we instrument more classes in one class then we use name `instrumentation.ts` for example grpc, you can check all of our plugins.

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
Rauno56: > rename: to give (someone or something) a new name
> — https://www.merriam-webster.com/dictionary/rename

In this case that "something" is a span. Not its name. Because we're not naming the name, but a span.

I mean... I'm more than happy to change this if we get another vote. I'm currently not convinced though.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
Rauno56: > well in yr case if fnName is undefined then the span name attribute will be undefined too.

Not exactly. **Span name** will be set - "<anonymous>" will be used in place of the function name. Function name attribute will *not* be set. And that is intentional.

> After proposed change if you leave setting span name as attribute it will be anonymous which at least give some more value then undefined.

My point above was that it is not useful for me(the instrumentation developer) to give the function an *arbitrary* name either because that function was in fact nameless.

The developer might not expect to see "anonymous" as its name. "anonymous" might actually collide with another name they use.

Rauno56:
Interesting that the lint fails on a file that no longer exists.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
vmarchaud: > My point above was that it is not useful for me(the instrumentation developer) to give the function an arbitrary name either because that function was in fact nameless.

This isn't arbitraty, [anonynmous function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#the_function_expression_function_expression) are a known type of function and everyone is using this name to recognize them so i think we should keep the anonymous name.

Rauno56: Yes, the term "anonymous function" absolutely applies here. Function **name** is very different though. `<anonymous>` is not even a valid name for a function, `anonymous`(without the brackets) is, but for example:

```
const anonymous = () => {
	// not actually anonymous, but named "anonymous"
};
router.use(anonymous); // case 1
router.use(() => {}); // case 2
```

both cases would get the same name, but in fact, only case 1 would be "correct". I'm just having a hard time seeing any good in filling in the name if there is in fact none. It would only collapse options further down by merging cases above into one.

vmarchaud: `<anonymous>` is a standard name for unnamed function, specially used for stacktraces (CTRL+F "anonymous", second reference in https://v8.dev/docs/stack-trace-api).

Rauno56: Yes! Absolutely correct! And inside the span name(just like inside of stack traces) that's exactly how I render the name as well. :slightly_smiling_face: 

However! Just like `fn.name` doesn't give you `"<anonymous>"`, but `undefined`, the attribute(raw data) should also be undefined. In my opinion.

I see I'm clearly not able to convince you two of that. We can end the discussion here. Thanks for your time, again, as always! I'll change the behaviour.

vmarchaud: > However! Just like fn.name doesn't give you "<anonymous>", but undefined, the attribute(raw data) should also be undefined. In my opinion.

I totally understand your point, but imagine if some vendors might use this to display the fn name in their UI they will need to display a fallback in their code even though `<anonymous>` should be its name, we have the JS specific knowledge so we should set coherent values when we know them so vendor don't have to.

(Althrough we don't currently have a standard attribute for this data, it might comes around in the future when vendors start to implement deep otel integration and use our conventions for their UI)

plugins/node/opentelemetry-instrumentation-router/README.md
```diff
@@ -0,0 +1,58 @@
+# OpenTelemetry router Instrumentation for Node.js
+
+[![Gitter chat][gitter-image]][gitter-url]
```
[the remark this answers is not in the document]
Rauno56: Used links based on Koa readme.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
Rauno56: I see where you are coming from as well.

As I understand the aspect we are disagreeing on is **when** should that substitution for the fact("the name is not set") be done to turn it into human-consumable information(dashboard listing the names of functions somewhere).

- You're saying that we, as instrumentation specialists, know the domain in general, the best practices and what the operators likely want to see.
- I'm saying that it's best to push that specific decision as far as down the line as possible.

Both have pros and cons imho.

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
iNikem: One "changes the name" and "renames the span".

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
Rauno56: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/479

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
Rauno56: Changed the behavior.

plugins/node/opentelemetry-instrumentation-router/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+import { Span } from '@opentelemetry/api';
+
+export enum LayerType {
+  MIDDLEWARE = 'middleware',
+  REQUEST_HANDLER = 'request_handler',
+}
+
+export enum CustomAttributeNames {
```
vmarchaud, line 23: for consistency with other packages, culd you move those custom attributes to `enums/AttributeNames.ts` ? https://github.com/open-telemetry/opentelemetry-js-contrib/commit/aec35d8a5e3d06c39db58c1ac7d20555163dc709

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import * as constants from './constants';
+import * as types from './types';
+import * as Router from 'router';
+
+// Detect whether a function is a router package internal plumming handler
+export const isInternal = (fn: Function) => {
+  // Note that both of those functions are sync
+  if (fn.name === 'handle' && fn.toString() === constants.ROUTER_HANDLE_FN) {
```
vmarchaud, line 24: why do we need to check the string body ? Isnt the name enough ?

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+    return true;
+  }
+  if (fn.name === 'router' && fn.toString() === constants.ROUTE_ROUTER_FN) {
+    return true;
+  }
+  return false;
+};
+
+export const getRoute = (req: Router.RoutedRequest) =>
+  req.baseUrl + (req.route?.path || '') || '/';
```
vmarchaud, line 34: prefer `??` over `||` when possible

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
```diff
@@ -0,0 +1,215 @@
...
+import {
+  InstrumentationBase,
+  InstrumentationNodeModuleDefinition,
+  InstrumentationNodeModuleFile,
+  isWrapped,
+} from '@opentelemetry/instrumentation';
+import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
+
+import * as http from 'http';
+import * as Router from 'router';
```
vmarchaud, line 27: could you use `import type` over `import` ? That will help to avoid using actual export at runtime so the instrmuentation can be loaded (by metapackages) even though the `router` package isn't installed.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
```diff
@@ -0,0 +1,215 @@
...
+
+import * as types from './types';
+import { VERSION } from './version';
+import * as constants from './constants';
+import * as utils from './utils';
+
+const { diag } = api;
```
vmarchaud, line 34: could we just use `api.diag` instead of destructing the api export ?

vmarchaud, line 34: could we just use `api.diag` instead of destructing the api export ?

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
```diff
@@ -0,0 +1,211 @@
...
+      spanName,
+      {
+        attributes,
+      },
+      parent
+    ) as types.InstrumentationSpan;
+
+    utils.renameHttpSpan(parentSpan, layer.method, route);
+    // make sure spans are ended at least when response is finished
+    res.prependOnceListener('finish', () => span.end());
```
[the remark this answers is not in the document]
vmarchaud: I couldn't find the discussion in #416 but ending twice the span isn't technically a noop because its log a warning to the diag logger. Now that isn't enabled by default isn't a problem for all users but i suspect it will for people having it enabled (see https://github.com/open-telemetry/opentelemetry-js-contrib/issues/107 for exemple).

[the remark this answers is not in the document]
vmarchaud: I couldn't find the discussion in #416 but ending twice the span isn't technically a noop because its log a warning to the diag logger. 
Now that isn't enabled by default isn't a problem for all users but i suspect it will for people having it enabled (see https://github.com/open-telemetry/opentelemetry-js-contrib/issues/107 or https://github.com/open-telemetry/opentelemetry-js/pull/2126 for exemple).

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
Rauno56: I'm afraid the name is too generic to be reliable in this case. Ideally, we'd reference compare the actual function called, but we do not have access to that.

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
Rauno56: Changed the first one only because the second one is functional/meaningful.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
Rauno56: You are right!

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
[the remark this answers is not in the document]
Rauno56: I'm going to add a check here, but I feel like it's such an widespread usecase, we should address it somehow better. **Any ideas?**

I cannot see a situation where this would even be something to be concerned about.

plugins/node/opentelemetry-instrumentation-router/src/types.ts
Rauno56: Moved that under enums. Stuff's changing underneath me, hard to keep up. :)

vmarchaud (approved):
lgtm, thanks for the changes

plugins/node/opentelemetry-instrumentation-router/test/index.test.ts
```diff
@@ -0,0 +1,270 @@
...
+  helloRouter.get('/:name', function announceRude(req, res, next) {
+    res.end('How rude!');
+  });
+
+  router.use('/hello', helloRouter);
+
+  const deepRouter = new Router();
+
+  deepRouter.use('/hello', helloRouter);
+  router.use('/deep', deepRouter);
```
johnbley, line 72: I like the thinking behind this test (setup)!

johnbley (approved):
lgtm, agree that `renameHttpSpanName` was a bit much.

plugins/node/opentelemetry-instrumentation-router/src/utils.ts
dyladan: I agree with rauno/inikem here. If I rename a thing, I am changing its name. I wouldn't say I renamed the name of something. It would be either `changeHttpSpanName` or `renameHttpSpan`.

obecny (approved):
lgtm

iNikem:
@open-telemetry/javascript-maintainers Can this be merged now?

PR #460, opened 2021-04-28
chore: moving dev dependency to main dependency due to upgrade to ts 4

obecny:
Because of upgrading to ts v4 the types dependency needs to be now a part of main dependency, otherwise there is an error during installation for example
```
node_modules/@opentelemetry/instrumentation-pg/build/src/pg.d.ts:2:26 - error TS2307: Cannot find module 'pg' or its corresponding type declarations.

2 import * as pgTypes from 'pg';
```

We need to make a patch release asap

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
@@ -42,7 +42,6 @@
   },
   "devDependencies": {
     "@opentelemetry/node": "0.19.0",
-    "@types/aws-lambda": "8.10.76",
     "@types/mocha": "7.0.2",
     "@types/node": "14.14.41",
     "codecov": "3.8.1",
@@ -60,6 +59,7 @@
     "@opentelemetry/instrumentation": "^0.19.0",
     "@opentelemetry/resources": "^0.19.0",
     "@opentelemetry/semantic-conventions": "^0.19.0",
-    "@opentelemetry/tracing": "^0.19.0"
+    "@opentelemetry/tracing": "^0.19.0",
+    "@types/aws-lambda": "8.10.76"
   }
 }
--- a/plugins/node/opentelemetry-instrumentation-bunyan/package.json
+++ b/plugins/node/opentelemetry-instrumentation-bunyan/package.json
@@ -45,7 +45,6 @@
     "@opentelemetry/context-async-hooks": "0.19.0",
     "@opentelemetry/node": "0.19.0",
     "@opentelemetry/tracing": "0.19.0",
-    "@types/bunyan": "1.8.6",
     "@types/mocha": "7.0.2",
     "@types/node": "14.14.41",
     "@types/sinon": "9.0.11",
@@ -63,6 +62,7 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.0.0-rc.0",
-    "@opentelemetry/instrumentation": "^0.19.0"
+    "@opentelemetry/instrumentation": "^0.19.0",
+    "@types/bunyan": "1.8.6"
   }
 }
--- a/plugins/node/opentelemetry-instrumentation-express/package.json
+++ b/plugins/node/opentelemetry-instrumentation-express/package.json
@@ -45,7 +45,6 @@
     "@opentelemetry/context-async-hooks": "0.19.0",
     "@opentelemetry/node": "0.19.0",
     "@opentelemetry/tracing": "0.19.0",
-    "@types/express": "4.17.11",
     "@types/mocha": "7.0.2",
     "@types/node": "14.14.41",
     "codecov": "3.8.1",
@@ -63,6 +62,7 @@
     "@opentelemetry/api": "^1.0.0-rc.0",
     "@opentelemetry/core": "^0.19.0",
     "@opentelemetry/instrumentation": "^0.19.0",
-    "@opentelemetry/semantic-conventions": "^0.19.0"
+    "@opentelemetry/semantic-conventions": "^0.19.0",
+    "@types/express": "4.17.11"
   }
 }
--- a/plugins/node/opentelemetry-instrumentation-graphql/package.json
+++ b/plugins/node/opentelemetry-instrumentation-graphql/package.json
@@ -44,7 +44,6 @@
   },
   "devDependencies": {
     "@opentelemetry/tracing": "0.19.0",
-    "@types/graphql": "14.5.0",
     "@types/mocha": "8.2.2",
     "@types/node": "14.14.41",
     "codecov": "3.8.1",
@@ -58,6 +57,7 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.0.0-rc.0",
-    "@opentelemetry/instrumentation": "^0.19.0"
+    "@opentelemetry/instrumentation": "^0.19.0",
+    "@types/graphql": "14.5.0"
   }
 }
--- a/plugins/node/opentelemetry-instrumentation-hapi/package.json
+++ b/plugins/node/opentelemetry-instrumentation-hapi/package.json
@@ -45,7 +45,6 @@
     "@opentelemetry/context-async-hooks": "0.19.0",
     "@opentelemetry/node": "0.19.0",
     "@opentelemetry/tracing": "0.19.0",
-    "@types/hapi__hapi": "20.0.8",
     "@types/mocha": "7.0.2",
     "@types/node": "12.20.10",
     "codecov": "3.8.1",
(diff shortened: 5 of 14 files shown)


Flarna:
Which installation are you referring here? I tried `npm install @opentelemetry/instrumentation-pg` but got no error.

dyladan:
If you have a project which depends on `@opentelemetry/instrumentation-pg` but not on `pg`, it will fail to compile if you import `PostgresInstrumentation`.

This is not a common use-case, but it is at least the case for the `instrumentations-all` package which depends on all instrumentations.

plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
```diff
@@ -60,6 +59,7 @@
     "@opentelemetry/instrumentation": "^0.19.0",
     "@opentelemetry/resources": "^0.19.0",
     "@opentelemetry/semantic-conventions": "^0.19.0",
-    "@opentelemetry/tracing": "^0.19.0"
+    "@opentelemetry/tracing": "^0.19.0",
+    "@types/aws-lambda": "8.10.76"
```
Flarna, line 63: Usually we don't pin dependencies, only dev-dependencies

obecny: I just moved them to different location, don't want to change them in  any other way

dyladan: I think we actually do pin all deps that aren't `@opentelemetry` deps usually iirc

Flarna: The only dependency I found is semver e.g. [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/45e875167e15f24003a2baf8b69263923b632e19/plugins/node/opentelemetry-instrumentation-dns/package.json#L67) which is not pinned.

We should be careful which version we use for type definitions. If npm is not able to de-duplicate them users end up in having two versions of e.g. @types/aws-lambda in the same project which usually not what someone wants.

I doubt that typescript is able to use version A for one file and version B for other files. And clearly types will differ between versions.

dyladan: We might want to think about using `*` if that is the case.

In this case, I think it's probably ok since we don't "leak" any of the types from the packages we wrap. If we use `@types/pg@1` and the user has `@types/pg@2`, ours will simply be installed in the deep node_modules directory at something like `node_modules/@opentelemetry/instrumentation-pg/node_modules/@types/pg` no?

Flarna: Hard to tell without trying.
Main question is what is typescript using - and remembering - when transpiling a file like this and the user has @types/pg installed in a significant different version then @opentelemetry/instrumentation-pg.
```typescript
import * from "pg"
import * from "@opentelemetry/instrumentation-pg"
```

I assume as long as the only types used are `protected init(): (InstrumentationNodeModuleDefinition<typeof pgTypes> | InstrumentationNodeModuleDefinition<typeof pgPoolTypes>)[];` in an protected method never used by end users it's don't care.

Once there are APIs using more types it may get problematic.

maybe we should somehow hide/remove the types from `InstrumentationAbstract#init` to avoid this at all?

obecny: Guys you are aware that because we upgraded to ts 4 now all the packages doesn't contain the types and if you don't install the package by yrself it cannot be used. This change is just fixing to how it was, should be already merged and released. Then we can discuss about all other things but currently the packages are broken especially the meta package.

dyladan: @obecny is right. this simple program fails to compile because of missing types currently:

```ts
import type {} from '@opentelemetry/auto-instrumentations-node';
```

dyladan: A simple local test seems to show that typescript correctly differentiates the types. I manually installed `@types/pg` under `node_modules/@opentelemetry/instrumentation-pg/node_modules/@types/pg` and installed it also at `node_modules/@types/pg`. I made a breaking incompatible change to the types manually to the `node_modules/@types/pg`. My program still compiled successfuly and didn't break the checks for `node_modules/@opentelemetry/instrumentation-pg` so I think @Flarna's version concerns are OK.

obecny: How would that be a different if you keep types in dev and compile using ts 3.x ? you would still break it exactly the same ?. The files will be in the same location if you either keep in dev with ts3 or move it to main packages and compile with ver. 4. You will have exactly the same pinned version of types right ?

dyladan:
I can't figure out why this is caused. My first thought was that it was the upgrade to tsc 4, but the output from tsc 3 and the output from tsc 4 both contain the broken line `import * as pgTypes from 'pg'`.

I think this has likely always been like this, but because we never had metapackages that had real code before we never noticed. The failure is only triggered if you install the instrumentation and try to import it without having the module being instrumented installed.

I think this dependency solution is a stopgap. I think the real fix might be to not import the instrumented module in the instrumentation package.

obecny:
> I can't figure out why this is caused. My first thought was that it was the upgrade to tsc 4, but the output from tsc 3 and the output from tsc 4 both contain the broken line `import * as pgTypes from 'pg'`.
> 
> I think this has likely always been like this, but because we never had metapackages that had real code before we never noticed. The failure is only triggered if you install the instrumentation and try to import it without having the module being instrumented installed.
> 
> I think this dependency solution is a stopgap. I think the real fix might be to not import the instrumented module in the instrumentation package.

This is happening not only with metapackage try import just any package it will not compile unless you add a types. And you have to add types into main because if you don't then your final consumer / user of yr package will hit exactly the same error.

dyladan:
> > I can't figure out why this is caused. My first thought was that it was the upgrade to tsc 4, but the output from tsc 3 and the output from tsc 4 both contain the broken line `import * as pgTypes from 'pg'`.
> > I think this has likely always been like this, but because we never had metapackages that had real code before we never noticed. The failure is only triggered if you install the instrumentation and try to import it without having the module being instrumented installed.
> > I think this dependency solution is a stopgap. I think the real fix might be to not import the instrumented module in the instrumentation package.
> 
> This is happening not only with metapackage try import just any package it will not compile unless you add a types. And you have to add types into main because if you don't then your final consumer / user of yr package will hit exactly the same error.

There is no reason to have `@instrumentation/pg` without having `pg` so that's not really a bug.

dyladan:
Can you change the title since this actually happens with older versions of typescript too (i checked)

Flarna:
I tried to reproduce this but failed. Not sure what differs from my setup compared to yours but `import * as i from "@opentelemetry/auto-instrumentations-node"` doesn't cause any problems for me.

I see also no difference between typescript 3 and 4.

obecny:
@Flarna @dyladan ok I will let you then figure out what is the best way to fix it. Bear in ming that currently the meta package doesn't work out of box so people won't be able to use it without fixing themselves.

dyladan:
@obecny we are simply trying to make sure we understand the problem so we can actually have a correct fix instead of applying fixes without fully understanding the issue.

@Flarna can you try to `npm i` and `npm test` with the attached sample project? It definitely is failing on my machine.

[tests.zip](https://github.com/open-telemetry/opentelemetry-js-contrib/files/6415934/tests.zip)

Flarna:
ok, can reproduce.

Reason why I haven't seen the problem before is that I did
```js
tsc --init
tsc -p .
```

The generated tsconfig.json includes `"skipLibCheck": true` which avoids this problem.

dyladan:
Now that we can reliably repro, do we think each plugin should declare the types package as a dependency, or should the metapackage declare them all? I would say the metapackage probably.

obecny:
having this on each package will be working correctly in each case, the meta will receive "fixed" package automatically then. Or any other meta package that any 3rd party might already have

dyladan:
I think that's fair. I dont have a strong opinion each way. I see you deleted your branch or I was about to reopen this PR

obecny:
> I think that's fair. I dont have a strong opinion each way. I see you deleted your branch or I was about to reopen this PR

it had conflicts anyway

Flarna:
I think we should put it into each package. But I still think we should not pin the versions instead use `*` like it is done in dependencies within `@types` packages.
For some types it happens automatically anyway. For example`@types/graphql` version 14.5.0 is just a stub which itself depends on `graphql` version `*` which includes the types now.

An I still think we should aim mid term to get rid of "leaking" such dev dependencies.

dyladan:
>An I still think we should aim mid term to get rid of "leaking" such dev dependencies.

I agree here. We should be able to move all of the types we need into our packages directly instead of depending on them. It's a painful process but it removes this dependency issue.

Flarna:
At least in most cases we leak only at one place, the protected `init` function of the Instrumentation. e.g. for pg: `protected init(): (InstrumentationNodeModuleDefinition<typeof pgTypes> | InstrumentationNodeModuleDefinition<typeof pgPoolTypes>)[];`

I don't think the base class actually needs the types. I have not tried yet but I think if we return `InstrumentationNodeModuleDefinition<unknown>` we could hide this.

PR #463, opened 2021-04-30
feat(express): allow rewriting span names

seemk:
Add an optional hook to rewrite the span names. For example this will allow updating wildcard `*` paths to contain more information from the request.

As per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name):
```
Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.
```

Miscellaneous:
- Deleted duplicate code in the tests


plugins/node/opentelemetry-instrumentation-express/src/express.ts
```diff
@@ -54,6 +54,14 @@ export class ExpressInstrumentation extends InstrumentationBase<
     );
   }
 
+  getConfig(): ExpressInstrumentationConfig {
```
dyladan, line 57: I have seen you add this mechanism to other instrumentations already. If this is a feature we want on all instrumentations we should implement it in the base class so it doesn't need duplicated implementations all over the place. If you think this is a feature all instrumentations should have please create an issue in core so it can be discussed.

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -80,6 +80,16 @@ export enum ExpressLayerType {
 
 export type IgnoreMatcher = string | RegExp | ((name: string) => boolean);
 
+export type SpanNameHook = (
+  request: Request,
+  route: string,
+  /** Type of the layer handling the request.
```
dyladan, line 86: Does tsdodc work inline in function arguments like this?

plugins/node/opentelemetry-instrumentation-express/src/express.ts
vmarchaud: https://github.com/open-telemetry/opentelemetry-js/issues/2174

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -88,6 +98,7 @@ export interface ExpressInstrumentationConfig extends InstrumentationConfig {
   ignoreLayers?: IgnoreMatcher[];
   /** Ignore specific layers based on their type */
   ignoreLayersType?: ExpressLayerType[];
+  spanNameHook?: SpanNameHook;
```
vmarchaud, line 101: We will need to update the doc in the readme too

plugins/node/opentelemetry-instrumentation-express/src/express.ts
```diff
@@ -270,4 +291,27 @@ export class ExpressInstrumentation extends InstrumentationBase<
...
+    defaultName: string
+  ) {
+    const hook = this.getConfig().spanNameHook;
+
+    if (!hook) {
+      return defaultName;
+    }
+
+    try {
+      return hook(req, route, type, defaultName);
```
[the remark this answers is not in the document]
seemk: But it's just a wrapper around try/catch anyway with a less clear API?

E.g.

```
safeExecuteInTheMiddle(
  () => config?.spanNameHook?.(),
  e => diag.error(e),
  true
);
```

Is there an actual necessity for it?

plugins/node/opentelemetry-instrumentation-express/src/types.ts
seemk: Updated README

plugins/node/opentelemetry-instrumentation-express/src/express.ts
```diff
@@ -270,4 +291,27 @@ export class ExpressInstrumentation extends InstrumentationBase<
       };
     });
   }
+
+  _spanName(
```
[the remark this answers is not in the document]
seemk: Renamed it

plugins/node/opentelemetry-instrumentation-express/src/types.ts
seemk: Moved the explanation to README, got rid of these comments

plugins/node/opentelemetry-instrumentation-express/src/express.ts
seemk: As it's not published yet, I'm keeping `getConfig` until then

plugins/node/opentelemetry-instrumentation-express/src/express.ts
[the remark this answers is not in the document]
vmarchaud: Its just boilerplate to avoid everytime the try/catch handling but it isn't mandatory if you prefer to write it without it

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -54,6 +54,14 @@ export class ExpressInstrumentation extends InstrumentationBase<
     );
   }
 
+  getConfig(): ExpressInstrumentationConfig {
+    return this._config;
+  }
+
+  setConfig(config: ExpressInstrumentationConfig) {
+    this._config = config;
+  }
```
dyladan, line 63: With the latest release of core these are no longer required

Rauno56:
Sorry, @seemk for waking up so late to this, I'd think this would come up with other instrumentations as well. Could we solve it in a Span Processor?

gamedevsam:
I just ran into an issue that would be solved by this PR. Any updates here?

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -70,6 +70,13 @@ export type LayerMetadata = {
 
 export type IgnoreMatcher = string | RegExp | ((name: string) => boolean);
 
+export type SpanNameHook = (
+  request: Request,
+  route: string,
+  layerType: ExpressLayerType | null,
+  defaultPath: string
+) => string;
```
blumamir, line 78: I suggest to pass a single `info` object so if we need to extend it in the future, we won't end up with huge funcion signatures with many unused variables.
This pattern is what we use for most instrumentation hooks

blumamir:
@seemk @gamedevsam 
Is anyone still interested in finalizing this PR?
Seems like it's been sitting here for a long time

gamedevsam:
I'd hate to see this be discarded, but don't feel sufficiently comfortable with this codebase to take it on.

seemk:
I'll take it up for cleaning at the beginning of next week. For some reason it wasn't merged and then it was forgotten

seemk:
@blumamir @Rauno56 Brought the PR up to date

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -75,6 +76,13 @@ Express instrumentation has few options available to choose from. You can set th
 - `middleware`,
 - `request_handler` is the name for anything that's not a router or a middleware.
 
+`spanNameHook` is invoked with the following information:
```
Rauno56, line 79: "... with an object containing following properties"* or similar. Essentially I'd love to see something that tells me there are two arguments and that the properties of the first one are X and the second one is a string, etc.

Rauno56 (approved):
Well done! Thanks for refoctoring the tests as well.

plugins/node/opentelemetry-instrumentation-express/README.md
seemk: updated the wording a bit

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -202,9 +211,14 @@ export class ExpressInstrumentation extends InstrumentationBase<
...
         ) {
-          rpcMetadata.span.updateName(
+          const name = instrumentation._getSpanName(
+            {
+              request: req,
+              route,
+            },
             `${req.method} ${route.length > 0 ? route : '/'}`
           );
+          rpcMetadata.span.updateName(name);
```
blumamir, line 221: Here we actually updating the name of the `http` span to [match the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name):
> Many REST APIs encode parameters into URI path, e.g. /api/users/123 where 123 is a user id, which creates high cardinality value space not suitable for span names. In case of HTTP servers, these endpoints are often mapped by the server frameworks to more concise HTTP routes, e.g. /api/users/{user_id}, which are recommended as the low cardinality span names

Do we intend this new hook to affect the `http` span name as well? If we do then I suggest to make it more clear in the docs and the hook parameters that this is the case.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -277,4 +299,22 @@ export class ExpressInstrumentation extends InstrumentationBase<
...
+
+  _getSpanName(info: ExpressRequestInfo, defaultName: string) {
+    const hook = this.getConfig().spanNameHook;
+
+    if (!(hook instanceof Function)) {
+      return defaultName;
+    }
+
+    try {
+      return hook(info, defaultName);
```
blumamir, line 311: Is it possible that the hook returns `undefined` or `null` values?
WDYT about returning the default value in this case as well? 
something like:
```js
return hook(info, defaultName) ?? defaultName;
```

blumamir (commented):
Sorry for the late review.
Added 2 optional comments.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
seemk: Might be possible, but only if the usage goes against the API. Should we be defensive in such cases? It could very well return an arbitrary type.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
seemk: Yes it is intentional (and perhaps even the most important part of the hook). Added a comment to `types.ts`

blumamir: Thanks :)

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
blumamir: mmm, not sure how typescript enforces that. I think returning `undefined` is easier to hit, but that was just a thought. I'm good with keeping it the way it is 👍

blumamir (approved):

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
dyladan: Not all our users will be typescript users, but I think going against the defined contract is just undefined behavior. One thing we do want to make sure is that we are guarding against a crash though.

dyladan: Looking through the code there appears to be no guard against returning null or undefined and it will be directly used with `updateName` which also has no guard. The `updateName` method not having a guard is OK I think because the user is explicitly calling a function with the wrong data, but accidentally returning nothing is an extremely easy mistake to make, and the error won't be visible until the exporter at least, if not the tracing backend.

This should definitely have a guard against undefined/null return at the least.

seemk: Added the check

dyladan (approved):

Issue #465, opened 2021-05-02
PostgreSQL instrumentation uses wrong attributes


pauldraper:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-pg/src/enums.ts

Instead of `net.peer.host` (the semantic convention) the instrumentation uses `peer.host`. Etc.

svrnm:
I can take this.

pauldraper:
While you're at it, probably makes sense to use SemanticAttributes in @opentelemetry/semantic-conventions

svrnm:
@pauldraper the PR #467 should cover that, please double check if I missed anything!:-)

obecny:
@svrnm can I assign this to you then ?

Issue #297, opened 2020-12-30
PG Pool Plugin - semantic conventions


obecny:
Currently pg pool plugin is not using correct semantic conventions, some things are missing, some has different name etc. It should be aligned with
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md

Both for database and general for example "net.peer.host" etc.

Issue #296, opened 2020-12-30
PG Plugin - semantic conventions


obecny:
Currently pg plugin is not using correct semantic conventions, some things are missing, some has different name etc. It should be aligned with
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md

Both for database and general for example "net.peer.host" etc.

cartermp:
I think this can be closed because https://github.com/open-telemetry/opentelemetry-js-contrib/pull/467 implements these. There's an outstanding issue around capturing the JDBC connection string but the rest appear correct specified as per the spec.

PR #467, opened 2021-05-04
Fix semantic conventions for PG and PG Pool

### LINKED ISSUE #116: [pg + pg-pool] Update DB semantic conventions

svrnm:
This should fix #116, #296, #297 and #465

Replace all official semantic conventions with their values from @opentelemetry/semantic-conventions package. 
Leave some of the old custom attributes as they are, but put them into a namespace db.postgresql (while not officialy stated in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) there are some other examples with db.<db.system>, e.g. db.mssql.instance_name or db.mongodb.collection

```
  PG_VALUES = 'db.postgresql.values',
  PG_PLAN = 'db.postgresql.plan',
  IDLE_TIMEOUT_MILLIS = 'db.postgresql.idle.timeout.millis',
  MAX_CLIENT = 'db.postgresql.max.client',
```

### DIFF
--- a/examples/postgres/package.json
+++ b/examples/postgres/package.json
@@ -35,7 +35,7 @@
     "@opentelemetry/exporter-zipkin": "^0.19.0",
     "@opentelemetry/instrumentation": "^0.19.0",
     "@opentelemetry/instrumentation-http": "^0.19.0",
-    "@opentelemetry/instrumentation-pg": "file:///Users/neumanns/Projects/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-pg",
+    "@opentelemetry/instrumentation-pg": "^0.15.0",
     "@opentelemetry/node": "^0.19.0",
     "@opentelemetry/tracing": "^0.19.0",
     "express": "^4.17.1",
--- a/examples/postgres/package.json
+++ b/examples/postgres/package.json
@@ -35,7 +35,7 @@
     "@opentelemetry/exporter-zipkin": "^0.19.0",
     "@opentelemetry/instrumentation": "^0.19.0",
     "@opentelemetry/instrumentation-http": "^0.19.0",
-    "@opentelemetry/instrumentation-pg": "^0.15.0",
+    "@opentelemetry/instrumentation-pg": "^0.16.0",
     "@opentelemetry/node": "^0.19.0",
     "@opentelemetry/tracing": "^0.19.0",
     "express": "^4.17.1",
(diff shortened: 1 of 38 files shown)


examples/postgres/package.json
```diff
@@ -35,11 +35,11 @@
     "@opentelemetry/exporter-zipkin": "^0.19.0",
     "@opentelemetry/instrumentation": "^0.19.0",
     "@opentelemetry/instrumentation-http": "^0.19.0",
-    "@opentelemetry/instrumentation-pg": "^0.15.0",
+    "@opentelemetry/instrumentation-pg": "file:///Users/neumanns/Projects/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-pg",
```
dyladan, line 38: Please undo this

plugins/node/opentelemetry-instrumentation-pg/src/enums.ts
```diff
@@ -14,27 +14,10 @@
...
-  PEER_SERVICE = 'peer.service',
-
-  // PG specific -- not specified by spec
-  PG_VALUES = 'pg.values',
-  PG_PLAN = 'pg.plan',
-
-  // PG-POOL specific -- not specified by spec
-  IDLE_TIMEOUT_MILLIS = 'idle.timeout.millis',
-  MAX_CLIENT = 'max',
+  PG_VALUES = 'db.postgresql.values',
```
dyladan, line 19: I think this one might actually be in semconv?

dyladan (changes requested):
Blocked because your local path is in the example package.json

examples/postgres/package.json
svrnm: 🤦‍♂️ ... good catch. I changed that back.

plugins/node/opentelemetry-instrumentation-pg/src/enums.ts
svrnm: No, there is none for storing the "values"

obecny (approved):
lgtm

plugins/node/opentelemetry-instrumentation-pg/src/enums.ts
svrnm: will keep this for now, maybe spec needs to add something here.

examples/postgres/package.json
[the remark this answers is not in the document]
svrnm: fixed

vmarchaud (approved):

PR #468, opened 2021-05-04
moving dev dependency for types to main dependency

obecny:
moving dev dependency for types to main dependency


plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
```diff
@@ -60,6 +59,7 @@
     "@opentelemetry/instrumentation": "^0.19.0",
     "@opentelemetry/resources": "^0.19.0",
     "@opentelemetry/semantic-conventions": "^0.19.0",
-    "@opentelemetry/tracing": "^0.19.0"
+    "@opentelemetry/tracing": "^0.19.0",
+    "@types/aws-lambda": "*"
```
dyladan, line 63: If we don't have at least `^` we might get a version that is incompatible with us and users will fail to compile until we release an update. This is very dangerous

obecny: > If we don't have at least `^` we might get a version that is incompatible with us and users will fail to compile until we release an update. This is very dangerous

`*` is what @Flarna requested looks like it is not working

plugins/web/opentelemetry-instrumentation-user-interaction/package.json
```diff
@@ -54,7 +54,6 @@
     "@types/jquery": "3.5.5",
     "@types/mocha": "7.0.2",
     "@types/node": "14.14.43",
-    "@types/shimmer": "1.0.1",
```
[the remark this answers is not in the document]
obecny: yes

plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
Flarna: I tried '*' locally in a hapi and aws-lambda instrumentation and it worked. But most likely I tried the wrong thing...
It's used on default in `@types` packages from DefinitelyTyped.

Which error happened?

dyladan: I wouldn't expect an error but it looks like there was a force push so i can't see the old test run.

@Flarna it was mentioned you requested `*`. Do you think we shouldn't worry about the issue I mentioned above where a breaking change might break our users until we push a new release?

Flarna: Ah ok, I misunderstood the message above.

If we use `*` it means we accept any version and if nothing is specified the latest available is used. On consume side it means it will be always deduped and they don't get two versions installed.

Clearly `*` has the potential to break our CI where the actual build step is running. But using `^` may also break our CI because `@types` packages don't follow semver therefore even a patch version may break something.

On user side the opentelemetry packages are not transpiled anymore therefore the version should not matter.

So far my understanding... but honestly speaking types I'm not sure if I get all corner cases.
`@types` packages can be beasts sometimes - as we have recently seen here with the hapi/hapi-podium types.

Flarna: Just to be clear regarding my request: I'm fine with all variants to solve this as all of them may require to tune later.

dyladan: maybe we should even do `~`. We have seen with the metapackage that the typescript compiler is checking transitive type dependencies otherwise we wouldn't have this issue.

Honestly I think keeping pinned versions is the safest for the time being.

dyladan (approved):

Flarna (approved):

vmarchaud (approved):

PR #470, opened 2021-05-04
chore: removing usage of timed event from api

obecny:

- Removing dependency on TimedEvent from api

- Fixes https://github.com/open-telemetry/opentelemetry-js-api/issues/37

### DIFF
--- a/packages/opentelemetry-test-utils/testUtils.ts
+++ b/packages/opentelemetry-test-utils/testUtils.ts
@@ -16,11 +16,11 @@
 
 import * as childProcess from 'child_process';
 import {
-  SpanKind,
-  SpanAttributes,
+  HrTime,
   Span,
+  SpanAttributes,
+  SpanKind,
   SpanStatus,
-  TimedEvent,
 } from '@opentelemetry/api';
 import * as assert from 'assert';
 import { ReadableSpan } from '@opentelemetry/tracing';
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-test-utils/testUtils.ts
```diff
@@ -126,3 +126,15 @@ export const assertPropagation = (
   );
   assert.notStrictEqual(targetSpanContext.spanId, sourceSpanContext.spanId);
 };
+
+/**
+ * Represents a timed event.
+ * A timed event is an event with a timestamp.
+ */
+export interface TimedEvent {
```
dyladan, line 134: should be able to use the type from the `tracing` module here

obecny: the tracing doesn't have this interface yet

dyladan: I think it should be in tracing since it's a tracing sdk specific thing

dyladan: if this is just temporary you can just put a comment to remove later

obecny: but it doesn't exist there yet, besides it is just one interface needed for test utils, we can change it later or leave it as it is. I'm fine for both, but for now we cant use it.

dyladan: ok

Flarna (approved):

vmarchaud (approved):

PR #482, opened 2021-05-12
Update documentation for new propagator API

willarmiros:

- The example code uses an old way of setting the propagator, which causes apps to crash with the latest SDK version

- Updates sample code to use correct API usage. Tested it on a sample app to verify.


Flarna (approved):

vmarchaud (approved):
lgtm but the direct register should also works, why does the SDK crash ?

willarmiros:
@vmarchaud it crashed for me when I did:
```js
propagation.setGlobalPropagator(new AWSXRayPropagator());
provider.register();
```

In previous versions, I believe we had to `register()` the provider even after setting the global propagator, but perhaps now it is no longer necessary to also call `register()`? Regardless, I still prefer the updated approach from a style perspective.

Flarna:
`NodeTracerProvider.register()` will try to set `CompositePropagator` for Baggage and TraceContext. This will fail if there is already a propagator registered but I don't think it should crash.

There were some changes regarding this a few versions ago. Would be nice to get some infos in the versions of API/SDK/... used.

willarmiros:
@Flarna you're right, it doesn't crash it just gives me this stack trace, my mistake:

```
Error: @opentelemetry/api: Attempted duplicate registration of API: propagation
    at Object.registerGlobal (aws-otel-js/sample-apps/node_modules/@opentelemetry/api/build/src/internal/global-utils.js:35:19)
    at PropagationAPI.setGlobalPropagator (aws-otel-js/sample-apps/node_modules/@opentelemetry/api/build/src/api/propagation.js:41:24)
    at NodeTracerProvider.register (aws-otel-js/sample-apps/node_modules/@opentelemetry/tracing/build/src/BasicTracerProvider.js:77:31)
    at NodeTracerProvider.register (aws-otel-js/sample-apps/node_modules/@opentelemetry/node/build/src/NodeTracerProvider.js:28:15)
    at module.exports (aws-otel-js/sample-apps/tracer.js:43:18)
    at Object.<anonymous> (aws-otel-js/sample-apps/server.js:18:35)
    at Module._compile (internal/modules/cjs/loader.js:1201:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1221:10)
    at Module.load (internal/modules/cjs/loader.js:1050:32)
    at Function.Module._load (internal/modules/cjs/loader.js:938:14)
```

For what it's worth, 1.0.0-rc.0 of API, 0.19.0 of all core packages, but it sounds like this is expected behavior.

Flarna:
Yes, I think this is expected. Overwriting of the global `TracerProvider`, `ContextManager`, `Propagation` is not possible. It's needed to disable/remove the original first (e.g. `api.propagation.disable()`).

Exception is `diag.setLogger()` which allows overwrite.

obecny (approved):

PR #486, opened 2021-05-18
feat(aws-lambda): added request and response hooks

nirsky:
Hello, maintainers!

First wanted to thank you for this AWS Lamda instrumentation, very creative!

The attributes provided by this instrumentation are very basic (does not follow the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md)).
In this PR, I added simple request and response hooks. This way, each consumer can customize the attributes to his own needs.

Also added docs and tests.


blumamir (approved):

vmarchaud (approved):
overall lgtm, not sure some line pass the linter though

willarmiros (approved):
LGTM

anuraaga:
In Java, we are introducing an instrumenter API which provides hooks for extracting from requests / responses in a generic way. Is it something js may want to explore instead of one Instrumentation having them?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java#L65

Also, @nirsky you mentioned the spec, do you intend to use the hooks to implement spec attributes or your own? We should implement the missing spec attributes here natively rather than relying on hooks as they're the spec :) Only reason the HTTP ones aren't in yet is implementation time.

nirsky:
Hey @anuraaga, for the instrumenter API - that's a great suggestion! But I think it should be discussed in a larger audience.

As for the spec, I might implement it in the future, my hands are full now.
Nonetheless, the hooks are still needed, for example, some users may want to collect the HTTP request body.

nirsky:
I think unit tests are breaking due to https://github.com/open-telemetry/opentelemetry-js/issues/2212

nirsky:
Any timeline on when I can expect to get this merged?

PR #489, opened 2021-05-18
feat: implement instrumentation for generic-pool

Rauno56:

It adds an instrumentation for [`generic-pool`](https://github.com/coopernurse/node-pool).
Nothing extravagant. The package is quite simple. The instrumentation follows the theme by creating spans for the most important operation - acquiring a resource. Supports 3 somewhat different APIs through v2 to the latest(v3.7.8).


vmarchaud (approved):
lgtm

dyladan (approved):
Way to support multiple previous versions too. Looks good.

PR #501, opened 2021-05-25
chore: fix tls tests in net instrumentation

vmarchaud:
Original tests had a TLS certificate only valid for 1 month so tests started to fail may 22th. I changed to certificate to be valid until 2121 :)


vmarchaud:
Coverage is broken, i didn't change the test at all so i think we can ignore it

Rauno56 (approved):
Awesome! We have certs at
- https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-exporter-collector/test/certs
- https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-exporter-collector-grpc/test/certs
- https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-exporter-collector-proto/test/certs

Since your keyboard is still hot from creating all the certs here, could you take care of those as well if it makes sense?

vmarchaud:
> Since your keyboard is still hot from creating all the certs here, could you take care of those as well if it makes sense?

It does not i believe, those certs are used just to secure the communication with grpc and not https so validity doesn't make sense.

Flarna (approved):

dyladan (approved):

Issue #512, opened 2021-06-03
Express Instrumentation Incorrectly Names Spans for Base URLs


astorm:
### What version of OpenTelemetry are you using?

It's not clear to me what constitutes an OpenTelemetry version, but here are the package versions from my lock file.

    "@opentelemetry/api-metrics": "0.19.0"
    "@opentelemetry/context-async-hooks": "0.19.0"
    "@opentelemetry/core": "0.19.0"
    "@opentelemetry/instrumentation": "0.19.0",
    "@opentelemetry/propagator-b3": "0.19.0",
    "@opentelemetry/propagator-jaeger": "0.19.0",
    "@opentelemetry/resources": "0.19.0",
    "@opentelemetry/semantic-conventions": "0.19.0"
    "@opentelemetry/tracing": "0.19.0"

### What version of Node are you using?

Node v16.2.0 and Node v12.22.1

### What did you do?

The express instrumentation appears to incorrectly name spans for routes using the base/root URL.  Instead of naming the span something like

    'GET /'
    
The span is named

    'GET '    

Reproduction:

1. Create the following program 

   ```
    'use strict';
    //File: repro.js
    const { diag, DiagConsoleLogger, DiagLogLevel } = require("@opentelemetry/api");
    const { NodeTracerProvider } = require("@opentelemetry/node");
    const { registerInstrumentations } = require("@opentelemetry/instrumentation");
    const { HttpInstrumentation } = require("@opentelemetry/instrumentation-http");
    const { ConsoleSpanExporter, SimpleSpanProcessor} = require('@opentelemetry/tracing');
    const { ExpressInstrumentation } = require('@opentelemetry/instrumentation-express');
    const core = require("@opentelemetry/core");

    const provider = new NodeTracerProvider();

    diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.ALL);

    provider.addSpanProcessor(
      new SimpleSpanProcessor(
        new ConsoleSpanExporter()
      )
    );

    provider.register();

    registerInstrumentations({
      instrumentations: [
        new HttpInstrumentation(),
        new ExpressInstrumentation({
          ignoreLayersType: ['request_handler','middleware']
        }),
      ],
    });

    const express = require('express')
    const app = express()
    const port = 3000

    app.get('/', (req, res) => {
      res.send('Hello Homepage!')
    })

    app.get('/hello', (req, res) => {
      res.send('Hello World!')
    })

    app.listen(port, () => {
[...]

PR #517, opened 2021-06-04
feat: use rpcMetadata to update http span name #464

### LINKED ISSUE #464: Express instrumentation depends on our SDK
The express instrumentation has a type `ExpressInstrumentationSpan` which extends `Span` with `name: string`, which is not exposed by the API type. This means it can only be used with SDKs that provide a public accessor for name.

At the very least it should be documented. I would prefer to remove the type.

vmarchaud:
Fixes #464 #512

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
-import { hrTime } from '@opentelemetry/core';
+import {
+  hrTime,
+  setRPCMetadata,
+  getRPCMetadata,
+  RPCType,
+} from '@opentelemetry/core';
 import { trace, context, diag, SpanAttributes } from '@opentelemetry/api';
 import * as express from 'express';
 import {
@@ -23,7 +28,6 @@ import {
   PatchedRequest,
   _LAYERS_STORE_PROPERTY,
   ExpressInstrumentationConfig,
-  ExpressInstrumentationSpan,
 } from './types';
 import { ExpressLayerType } from './enums/ExpressLayerType';
 import { AttributeNames } from './enums/AttributeNames';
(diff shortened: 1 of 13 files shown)


plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -192,19 +196,15 @@ export class ExpressInstrumentation extends InstrumentationBase<
 
         // Rename the root http span in case we haven't done it already
         // once we reach the request handler
+        const rpcMetadata = getRPCMetadata(context.active());
         if (
           metadata.attributes[AttributeNames.EXPRESS_TYPE] ===
-          ExpressLayerType.REQUEST_HANDLER
+            ExpressLayerType.REQUEST_HANDLER &&
+          rpcMetadata?.type === RPCType.HTTP
```
dyladan, line 203: Odd spacing here, wonder if lint will fail?

plugins/node/opentelemetry-instrumentation-express/test/custom-config.test.ts
```diff
@@ -175,5 +178,58 @@ describe('ExpressInstrumentation', () => {
         }
       );
     });
+
+    it('should correctly set the http path', async () => {
```
dyladan, line 182: What case is this testing? Is this the root path test?

dyladan (approved):
Looks good thanks :)

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
vmarchaud: i actually don't know, i applied the eslint `--fix` option through my IDE and i passed lint locally

plugins/node/opentelemetry-instrumentation-express/test/custom-config.test.ts
vmarchaud: Its the test case for #512, do you want to update the name ? It just verify that the http path of the span is `GET /` and not `GET` like the user reported

dyladan: It's just that it could apply to any name or path case. maybe "should correctly name http root path"?

vmarchaud: I've updated it

PR #521, opened 2021-06-04
feat: mysql2 instrumentation

Rauno56:

Adds autoinstrumentation for `mysql2`.

This is pretty much the same as `opentelemetry-instrumentation-mysql2`, but I simplified it by only wrapping `Connection.prototype`. This exact code might actually work for `mysql` as well.


vmarchaud:
If patched module are close, i think we should just add support for mysql2 inside `mysql` instrumentation, WDYT ?

Rauno56:
Yes... It would be totally fine, unless... Types.
Could still be done, but I'd take it on outside of this PR.

vmarchaud:
> Yes... It would be totally fine, unless... Types.

Not sure to understand why types are a problem ?

Rauno56:
We don't have a good precedent of 2-package instrumentation - I mainly don't want to start that discussion here.

To answer your question: Because those do not match between the two packages. Not saying it's not doable, it absolutely is with unions and whatnot, but transforming the code to work with `mysql2` was already a task and I don't see a good reason to push more into a PR.

vmarchaud:
FYI we have both http (that handle https too) and grpc (that handle grpc-js and grpc c++) that have this behavior and we don't have any issue with it.
I totally think that since you already written as a separate package it doesnt matter but i was more asking if you tried and there was issue.

vmarchaud (approved):
lgtm

vmarchaud (approved):
lgtm

Rauno56:
Thanks for pointing this out, @vmarchaud. I'll look into that at some point! :)

vmarchaud:
cc @open-telemetry/javascript-approvers would appreciate some review so we can ship this with the next release :) thanks

johnbley (approved):

PR #525, opened 2021-06-07
feat(peer-api-check): error differently if devDep not pinned

Rauno56:
I bumped into this in my PRs when I had both(devDep and peerDep) versions with `^`.

### DIFF
--- a/scripts/peer-api-check.js
+++ b/scripts/peer-api-check.js
@@ -26,6 +26,12 @@ if (pjson.dependencies && pjson.dependencies['@opentelemetry/api']) {
 const peerVersion = pjson.peerDependencies && pjson.peerDependencies['@opentelemetry/api'];
 const devVersion = pjson.devDependencies && pjson.devDependencies['@opentelemetry/api'];
 if (peerVersion) {
+	// error if not pinned
+	if (!/^[0-9]/.test(devVersion)) {
+		throw new Error(
+			`Package ${pjson.name} does't have API version pinned in dev dependencies: ${devVersion}`
+		);
+	}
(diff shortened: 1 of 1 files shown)


Flarna:
I copied the script from opentelemetry-js repo and only adapted it to the  dependency style here. in contrib API is pinned in dev dependencies but in opentelemetry-js we don't pin it.

Personally I would prefer to use the same solution in both repos as I see no reason for a difference.

Please note also that renovate bot has it's own opinion/config regarding what to pin and this should also match.

@open-telemetry/javascript-maintainers I think it's up to you to decided where and how we pin/not pin.

Rauno56:
Yes, I don't mind whichever way we go either, but it does make sense *not* to pin them here to give flexibility for the end user. Actually... for the same reason it should be unpinned in `opentelemetry-js` as well, but I guess there is a good reason not to?

The setups are different and I wanted to just make the error message more descriptive.

Flarna:
pinning dev-dependency has no impact on end-user and peer-dependency is not pinned.

As far as I know dev-dependencies are pinned to reduce the number of times CI becomes red because a new version of a dependency is published. renovate should ensure that they get updated frequently but via a PR which shows CI state with new versions.

Clearly transitive dev-dependencies are not covered by this and we had already issues with them. We tried also package-lock.json files but it turned out to be a pain and use of them was reverted.

Flarna (approved):

vmarchaud:
I'm not sure why the API isn't pinned in core, i think it should be pinned when in dev deps (like others). this PR lgtm though

vmarchaud (approved):

dyladan (approved):

PR #526, opened 2021-06-09
Baggage propagation issue

amitgud-doordash:

-

-

### DIFF
--- a/examples/graphql/server-express.js
+++ b/examples/graphql/server-express.js
@@ -5,10 +5,33 @@ require('./tracer');
 const express = require('express');
 const { graphqlHTTP } = require('express-graphql');
 const buildSchema = require('./schema');
+const otel = require('@opentelemetry/api');
 
 const schema = buildSchema();
 
+const testMw = function (req, res, next) {
+  otel.context.with(
+    otel.propagation.setBaggage(
+      otel.context.active(),
+      otel.propagation.createBaggage({
+        'test': { value: 'testValue' },
+      })
+    ),
+    () => {
+      console.log('attached baggage', otel.propagation.getBaggage(otel.context.active()));
+      next();
+    }
+  );
+}
+
+const printMw = function (req, res, next) {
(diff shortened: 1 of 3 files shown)


examples/graphql/server-express.js
```diff
@@ -5,10 +5,33 @@ require('./tracer');
...
+    ),
+    () => {
+      console.log('attached baggage', otel.propagation.getBaggage(otel.context.active()));
+      next();
+    }
+  );
+}
+
+const printMw = function (req, res, next) {
+  console.log('printMw baggage', otel.propagation.getBaggage(otel.context.active()));
```
amitgud-doordash, line 28: This is 'undefined'.

johnbley: Hi Amit - 
It looks like you're trying to report a bug, rather than contributing a code fix?  It would help us if you would [file an issue](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/new/choose) with your sample code rather than a pull request.

examples/graphql/server-express.js
```diff
@@ -5,10 +5,33 @@ require('./tracer');
...
+  otel.context.with(
+    otel.propagation.setBaggage(
+      otel.context.active(),
+      otel.propagation.createBaggage({
+        'test': { value: 'testValue' },
+      })
+    ),
+    () => {
+      console.log('attached baggage', otel.propagation.getBaggage(otel.context.active()));
+      next();
```
vmarchaud, line 22: Express break context propagation (see https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L263) so you'll need to use `context.bind` when calling next here

vmarchaud:
I'm going to close this, please [open a discussion](https://github.com/open-telemetry/opentelemetry-js/discussions) on the main repo next time

PR #537, opened 2021-06-14
fix(instrumentation-user-interaction): support clicks in React apps

### LINKED ISSUE #168: onclick - react, zone.js
Investigate problem with onclick event for react when using plugin for user interaction , react plugin, xmlhttprequest plugin, and zone manager.
More context here -> https://github.com/open-telemetry/opentelemetry-js-contrib/pull/164#issuecomment-668811140

kkruk-sumo:

This fixes #168.

React <17 listens on `document`. It didn't work previously, because we were only listening on `HTMLElement` instances.
React >=17 listens on a page root (usually a div or `document.body`), so all spans had the same xpath.

**This PR changes current behaviour**.

The good news is that the current behaviour is undocumented - I didn't have to change existing unit tests.

Previously the element on which the listener has been added was taken into a span (in `div.addEventListener` - the `div` was always in span's xpath, not the clicked child).
After this change, the `event.target` will be taken (the actually clicked element).


plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -265,16 +265,19 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
+          const target = event?.target;
           if (event) {
             parentSpan = plugin._eventsSpanMap.get(event);
           }
           if (once) {
             plugin.removePatchedListener(this, type, listener);
           }
-          const span = plugin._createSpan(target, type, parentSpan);
+          const span =
+            target instanceof HTMLElement
```
johnbley, line 278: `_createSpan` already has a number of "does this have enough information to produce a good span? -> if not, `undefined`" checks, and I think this one belongs inside `_createSpan` rather than outside it, to keep said logic centralized.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -439,10 +442,14 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
-        const target: HTMLElement | undefined = task.target;
+        const event =
+          Array.isArray(applyArgs) && applyArgs[0] instanceof Event
+            ? applyArgs[0]
+            : undefined;
+        const target = event?.target;
         let span: api.Span | undefined;
         const activeZone = this;
-        if (target) {
+        if (target instanceof HTMLElement) {
```
johnbley, line 452: See previous comment - if `_createSpan` does this check, it doesn't need to be repeated elsewhere.

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -410,6 +415,122 @@ describe('UserInteractionInstrumentation', () => {
...
+            }).then(() => {});
+            break;
+        }
+      };
+
+      document.addEventListener('click', listener);
+
+      try {
+        btn1.click();
+        btn2.click();
```
johnbley, line 445: I appreciate this test (having two targets and ensuring that the span names properly line up)!

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
```diff
@@ -40,6 +40,12 @@ const FILE_URL =
   'https://raw.githubusercontent.com/open-telemetry/opentelemetry-js/main/package.json';
 
 describe('UserInteractionInstrumentation', () => {
+  afterEach(() => {
+    while (document.body.lastChild) {
+      document.body.removeChild(document.body.lastChild);
```
johnbley, line 45: I'm not sure why this is necessary?  Perhaps a comment?

johnbley (changes requested):

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
kkruk-sumo: Comment added.

johnbley (approved):
Looks good now, thanks for the edits!

MSNev (approved):

obecny (approved):

PR #538, opened 2021-06-15
feat: add cassandra-driver instrumentation

seemk:

Adds [cassandra-driver](https://www.npmjs.com/package/cassandra-driver) instrumentation.

* Instruments `execute`, `batch`, `stream` calls. `stream` span ends when all of the data has been received.
* Added support for Cassandra docker images to `test-utils`.
* Supported versions: `>=4.4 <5.0`


plugins/node/opentelemetry-instrumentation-cassandra/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@opentelemetry/instrumentation-cassandra-driver",
+  "version": "0.20.0",
```
vmarchaud, line 3: should be 0.21 now

plugins/node/opentelemetry-instrumentation-cassandra/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^0.21.0"
+  },
+  "devDependencies": {
+    "@opentelemetry/api": "0.21.0",
```
vmarchaud, line 48: you can directly put api 1.0 (PR to update contrib https://github.com/open-telemetry/opentelemetry-js-contrib/pull/540 will likely be merged next week)

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -0,0 +1,366 @@
...
+      original: (...args: unknown[]) => Promise<CassandraDriver.types.ResultSet>
+    ) => {
+      const plugin = this;
+      return function patchedExecute(
+        this: CassandraDriver.Client,
+        ...args: unknown[]
+      ) {
+        const span = startSpan(
+          plugin.tracer,
+          'execute',
```
vmarchaud, line 131: i think the span name should include that its a span for cassandra ? What about `cassandra.execute` ?

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -0,0 +1,366 @@
...
+            ) as Promise<CassandraDriver.types.ResultSet>;
+          },
+          error => {
+            if (error) {
+              failSpan(span, error);
+            }
+          }
+        );
+
+        const wrappedPromise = wrapPromise(span, batchPromise);
```
vmarchaud, line 240: since you are wrapping the promise, do we need `safeExecuteInTheMiddle` ? the promise can't throw if you catch it right ?

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -0,0 +1,366 @@
...
+      return function patchedBatch(
+        this: CassandraDriver.Client,
+        ...args: unknown[]
+      ) {
+        const queries = Array.isArray(args[0]) ? args[0] : [];
+        const combined = truncateQuery(
+          combineQueries(queries),
+          plugin.getMaxQueryLength()
+        );
+        const span = startSpan(plugin.tracer, 'batch', combined, this);
```
vmarchaud, line 191: same as above, what about naming it `cassandra.batch` ?

vmarchaud: Sorry i didn't see that you used a custom method

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  private _getPatchedStream() {
+    return (original: (...args: unknown[]) => EventEmitter) => {
+      const plugin = this;
+      return function patchedStream(
+        this: CassandraDriver.Client,
+        ...args: unknown[]
+      ) {
+        // Since stream internally uses execute, there is no need to add DB_STATEMENT twice
+        const span = plugin.tracer.startSpan('cassandra-driver.stream', {
```
vmarchaud, line 255: we not use `startSpan` here too ?

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -0,0 +1,366 @@
...
+
+function startSpan(
+  tracer: Tracer,
+  op: string,
+  statement: string,
+  client: CassandraDriver.Client
+) {
+  const attributes = {
+    [SemanticAttributes.DB_SYSTEM]: DbSystemValues.CASSANDRA,
+    [SemanticAttributes.DB_STATEMENT]: statement,
```
vmarchaud, line 308: we generally do not add the statement by default since it can contains sensitive values, could add `enhancedDatabaseReporting` like mongodb (https://github.com/open-telemetry/opentelemetry-js-contrib/blob/621f0141e406b75ef11a77fb85ce8caab9c8a272/plugins/node/opentelemetry-instrumentation-mongodb/src/types.ts#L25) ?

seemk: Disabled it by default and added the same option

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
seemk: Refactored it to use the internal `startSpan`

plugins/node/opentelemetry-instrumentation-cassandra/package.json
seemk: Upgraded to 1.0.0

plugins/node/opentelemetry-instrumentation-cassandra/package.json
seemk: Upgraded to 0.22 :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
seemk: The call itself might throw (e.g. on invalid input)

obecny:
it looks like version 3.x is still quite popular would you be able to add support for version 3 too ?

image

seemk:
> it looks like version 3.x is still quite popular would you be able to add support for version 3 too ?
> 
> image

As it's already 3 years old and gradually phased out, perhaps it would make sense to add it once someone explicitly requests it?

By the way, a CI job is stuck at `Starting workflow run` for node 14, is it possible to restart it?

vmarchaud (approved):
lgtnm

dyladan (approved):
I agree with seemk, there is no urgent need to support old versions unless someone requests it.

Please fix the conflicts and this can be merged

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  public startSpan(
+    { op, query }: { op: string; query?: unknown },
+    client: CassandraDriver.Client
+  ): Span {
+    const attributes: SpanAttributes = {
+      [SemanticAttributes.DB_SYSTEM]: DbSystemValues.CASSANDRA,
+    };
+
+    if (this._shouldIncludeDbStatement() && query !== undefined) {
```
johnbley, line 301: Standard PII concerns here -
- What normalization/sanitization is/should be performed on this string/set of strings?  Might it contain sensitive values?
- If it can contain sensitive values, then the documentation about the config option should note the risk.

johnbley (changes requested):

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
seemk: It can contain sensitive information if the queries are are not escaped, e.g. when using
```javascript
const creditCard = '1234';
client.execute(`insert into credit_cards (number) values ('${creditCard}')`);
```
instead of the usual interface
```javascript
client.execute('insert into credit_cards (number) values (?)', ['1234']);
```

With the first case you'd have more to worry about anyway :smile:

Anyway I added a comment to both code and README, thanks!

dyladan: Since enhanced reporting is off by default and the behavior is documented I think this is acceptable thanks.

johnbley: Seconded Daniel's viewpoint.  LGTM now

johnbley (approved):

PR #539, opened 2021-06-17
feat: memcached instrumentation

Rauno56:

Adds auto-instrumentation for [`memcached`](https://github.com/3rd-Eden/memcached).

We're monkey-pathing an public generic `command` function all methods use internally.


plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -0,0 +1,183 @@
...
+    return function (
+      this: Memcached,
+      queryCompiler: () => Memcached.CommandData,
+      server?: string
+    ) {
+      if (typeof queryCompiler !== 'function') {
+        return original.apply(this, arguments as any);
+      }
+      // The name will be overwritten later
+      const span = instrumentation.tracer.startSpan('<type> <key>', {
```
johnbley, line 96: I'd prefer `'unknown'` or `'unknown memcached command'` or something similar in case renaming somehow fails.

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -0,0 +1,183 @@
...
+      const callback = query.callback;
+
+      span.updateName(`${query.type} ${query.key}`);
+      span.setAttributes({
+        'db.memcached.key': query.key,
+        'db.memcached.lifetime': query.lifetime,
+        [SemanticAttributes.DB_OPERATION]: query.type,
+        'db.statement': (instrumentation._config as InstrumentationConfig)
+          .collectCommand
+          ? query.command
```
johnbley, line 141: I don't this API that well, and based on your tests it looks like the answer is 'no', but I have to ask: can this leak values, which might be sensitive information?  It looks like it's just operation and/or key names, but I want to make sure.

Rauno56: By default, it does not leak values.

However, if the configuration explicitly has `collectCommand` set to `true`, it does when setting values for example. Thoughts on that?

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
Rauno56: Me too actually!

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
johnbley: The behavior is fine and the existence of the config boolean is also fine; I might add a stronger disclaimer in the documentation - "Caution: setting `collectCommand=true` exposes values in command strings, which may contain sensitive information." or similar.

johnbley (approved):

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InstrumentationConfig as BaseInstrumentationConfig } from '@opentelemetry/instrumentation';
+
+export interface InstrumentationConfig extends BaseInstrumentationConfig {
+  collectCommand?: boolean;
```
vmarchaud, line 20: Could you use `enhancedDatabaseReporting` like other db instrumentation do (ex: [mongodb](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/621f0141e406b75ef11a77fb85ce8caab9c8a272/plugins/node/opentelemetry-instrumentation-mongodb/src/types.ts#L25)) so its consistent ?

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  constructor(config: InstrumentationConfig = Instrumentation.DEFAULT_CONFIG) {
+    super(
+      '@opentelemetry/instrumentation-memcached',
+      VERSION,
+      Object.assign({}, Instrumentation.DEFAULT_CONFIG, config)
+    );
+  }
+
+  setConfig(config: InstrumentationConfig = Instrumentation.DEFAULT_CONFIG) {
```
vmarchaud, line 46: I'm not sure to understand why you reset the default config here ? It doesn't make sense to set the config without actualling give it a config right ?

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -0,0 +1,186 @@
...
+    };
+  }
+
+  private ensureWrapped(
+    moduleVersion: string | undefined,
+    obj: any,
+    methodName: string,
+    wrapper: (original: any) => any
+  ) {
+    api.diag.debug(
```
vmarchaud, line 178: you should be able to use `this._diag` so the logger is namespaced with the instrumentation name

plugins/node/opentelemetry-instrumentation-memcached/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^0.21.0"
+  },
+  "devDependencies": {
+    "@opentelemetry/api": "0.21.0",
```
vmarchaud, line 49: you could use 1.0 directly (pr is already opened for contrib https://github.com/open-telemetry/opentelemetry-js-contrib/pull/540)

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
Rauno56: For resetting it to defaults, it does. I don't mind whichever way. Do you think it should be changed?

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
Rauno56: The name `enhancedDatabaseReporting`

1. is not descriptive by itself and;
2. sounds like it can affect multiple aspects of how the instrumentation works.

I'd personally stay away from such configuration options anywhere.
Happy to change it for you if that's a breaker though! :)

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
Rauno56: Fantastic! Thanks!

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
vmarchaud: We inherited this from Opencensus, i don't mind changing it but i would prefer to make it the same for all instrumentation anyway. Could you use `enhancedDatabaseReporting` and open an issue to discuss this ?

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
vmarchaud: I don't see why an user would want to reset to the default config at runtime (althrough its handy in tests). I would prefer to match the signature of the implementation in the `BaseInstrumentation` class

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
Rauno56: Could you please share a reference to that in OpenCensus?

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
Rauno56: OK! Changed it.

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
vmarchaud: > Could you please share a reference to that in OpenCensus?

I was wrong that wasn't from Opencensus directly but from the google cloud tracer sdk (https://github.com/googleapis/cloud-trace-nodejs/search?q=enhancedDatabaseReporting. 
You can lookup the initial proposal here: https://github.com/open-telemetry/opentelemetry-js/issues/214

> Since many of the auto-instrumentations will be provided by 3rd parties we can never ensure there is a finite list of options used across all instrumentations across all libs.

We cannot ensure consistency with 3rd party instrumentations but we can at least for 1st party instrumentation.

There was a discussion at the time here about this: https://github.com/open-telemetry/opentelemetry-js/issues/614

obecny:
in general lgtm, I'm missing just one thing to be able to verify this easily, a working example so that I can generate spans myself, can you please add the example ?

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
Rauno56: Thanks for the info, @vmarchaud! :pray: I renamed the config option.

Rauno56:
> in general lgtm, I'm missing just one thing to be able to verify this easily, a working example so that I can generate spans myself, can you please add the example ?

Added an example!

plugins/node/opentelemetry-instrumentation-memcached/src/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+        );
+        server = redundancy.shift();
+      } else {
+        server = client.HashRing.get(query.key);
+      }
+    }
+  }
+
+  if (typeof server === 'string') {
+    const [host, port] = server && server.split(':');
```
blumamir, line 47: if the first check fails (e.g. `server` is empty string), the right side expression will return an empty string which we will try to destruct into an array `[host, port]`. Not sure it will work (tried it locally and got the tsc error: `Type '""' is not an array type or a string type. Use compiler option '--downlevelIteration' to allow iterating of iterators.ts(2569)`

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -0,0 +1,186 @@
...
+    return function (this: Memcached) {
+      const query = original.apply(this, arguments as any);
+      const callback = query.callback;
+
+      span.updateName(`${query.type} ${query.key}`);
+      span.setAttributes({
+        'db.memcached.key': query.key,
+        'db.memcached.lifetime': query.lifetime,
+        [SemanticAttributes.DB_OPERATION]: query.type,
+        'db.statement': (instrumentation._config as InstrumentationConfig)
```
blumamir, line 142: you can replace with `SemanticAttributes.DB_STATEMENT` I think

vmarchaud (approved):
:+1:

examples/memcached/tracer.js
```diff
@@ -0,0 +1,37 @@
...
+const { SimpleSpanProcessor, ConsoleSpanExporter } = require('@opentelemetry/tracing');
+const { Resource } = require('@opentelemetry/resources');
+const { ResourceAttributes } = require('@opentelemetry/semantic-conventions');
+
+const { MemcachedInstrumentation } = require('@opentelemetry/instrumentation-memcached');
+
+module.exports = (serviceName) => {
+  const provider = new NodeTracerProvider({
+    resource: new Resource({
+      [ResourceAttributes.SERVICE_NAME]: 'memcached-example-service',
```
blumamir, line 19: nit: you can use the serviceName parameter to the function

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -0,0 +1,186 @@
...
+          .enhancedDatabaseReporting
+          ? query.command
+          : undefined,
+        ...utils.getPeerAttributes(client, server, query),
+      });
+
+      query.callback = api.context.bind(
+        context,
```
blumamir, line 150: did not actually run it, but wouldn't setting the context this way make it so that every span started inside the callback will be the child of the (KIND=CLIENT) memcached span. 
Not sure if it's intentional, but I think this is something we try to avoid in instrumentations.

plugins/node/opentelemetry-instrumentation-memcached/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "mocha": "7.2.0",
+    "nyc": "15.1.0",
+    "rimraf": "3.0.2",
+    "ts-mocha": "8.0.0",
+    "typescript": "4.3.2"
+  },
+  "dependencies": {
+    "@opentelemetry/instrumentation": "^0.22.0",
+    "@opentelemetry/semantic-conventions": "^0.22.0",
+    "@types/memcached": "^2.2.6"
```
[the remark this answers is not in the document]
Rauno56: Yes. This is the current practice. See references here: https://github.com/open-telemetry/opentelemetry-js/pull/2256

plugins/node/opentelemetry-instrumentation-memcached/src/utils.ts
blumamir: Since we already know that `server` is of type string at this point, I would just call split on it without checking that it's truth, which will work for empty strings as well

plugins/node/opentelemetry-instrumentation-memcached/src/types.ts
blumamir: I join the voice of not using `enhancedDatabaseReporting` parameter. Would be happy to see it replaced with meaningful names everywhere

plugins/node/opentelemetry-instrumentation-memcached/src/utils.ts
Rauno56: Typescript warns about that, yes, but actually, it's never gonna error as long as `server` is a string, which we check for.

```js
const [a, b] = '';
// a = b = undefined
```

Could make it more rigid, but I don't see a good reason to.

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
Rauno56: Yeah, I was thinking about it as well while implementing it. In some cases it makes sense, in others, not.

What's the correct way to implement it, @vmarchaud , @obecny ?

plugins/node/opentelemetry-instrumentation-memcached/package.json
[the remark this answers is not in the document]
obecny: since typescript 4+ if you use this instrumentation without the main package that was instrumented and types it will fail, strange behaviour but this is what we have, before ts 4 it was working fine as types could be references from dev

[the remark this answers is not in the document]
vmarchaud: When you define a type inside the `BaseInstrumentation` generic (to "try" to have type completion when patching/unpatching, it creates some sort of link that typescript will by default try to verify. If you don't bring the types as dependency, typescript will complain unless you add `skipLibCheck: true`

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
vmarchaud: Since you use `.with` when calling `wrapQueryCompiler` you shouldn't need to .bind here (except if the memcached client is loosing context, for example if they have internal queues), if it works without i believe can you can remove

blumamir: @vmarchaud - do you expect the callback to have the `memcached` span context? or the parent context?

vmarchaud: parent context, we want to have the memcached context only within the internal code that does the request.

blumamir: I agree, but current implementation is setting the Memcached span as context for the callback.
So do we agree to change it? @Rauno56

plugins/node/opentelemetry-instrumentation-memcached/package.json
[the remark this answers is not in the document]
blumamir: Thanks @vmarchaud @obecny 
Wasn't aware of it.
I Will fix it in my repo as well

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
blumamir: > Since you use `.with` when calling `wrapQueryCompiler` you shouldn't need to .bind here (except if the memcached client is loosing context, for example if they have internal queues), if it works without i believe can you can remove

@vmarchaud sorry I was reading your reply without full attention. I think it should still be binded, but to parent context, but not sure about it. better to try and see how it behaves

obecny (approved):
lgtm

vmarchaud:
@Rauno56 Should be good to merge after fixing the CI

blumamir (approved):

dyladan (approved):

PR #549, opened 2021-06-26
chore: enable typescript 4.3 option noImplicitOverride

Flarna:
Enable typescript 4.3 feature `noImplicitOverride` to improve code quality.
Adapt code to compile by adding override keyword or avoid overriding.

Update some dev-dependencies and correct some lint warnings.
Configure tsx files to use LF instead CRLF to stay in sync with ts files.

Refs: https://github.com/open-telemetry/opentelemetry-js/pull/2241


plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -71,11 +71,11 @@ export const traceContextEnvironmentKey = '_X_AMZN_TRACE_ID';
 export class AwsLambdaInstrumentation extends InstrumentationBase {
   private _tracerProvider: TracerProvider | undefined;
 
-  constructor(protected _config: AwsLambdaInstrumentationConfig = {}) {
+  constructor(protected override _config: AwsLambdaInstrumentationConfig = {}) {
```
Flarna, line 74: overriding `_config` here is solely done to get the right type.
In my opinion a better solution would be to use a generic for config (see https://github.com/open-telemetry/opentelemetry-js/issues/2258)

vmarchaud (approved):
lgtm

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
MSNev: overriding an existing variable from a base class is wasteful and potentially dangerous as the confusion as now your going to have 2 variable (the wasteful part) which (in theory) could reference 2 different objects (the dangerous part).

So if the base (or sub) class "reassigns" the config for some reason (perhaps we start supporting dynamic updating of config for example -- or they just call the existing setConfig(...) ) then you now have the 2 different variables referencing different objects  so overriding functions could now have extremely unexpected side effects etc.

Would prefer to either have this as a generic or have an accessor that does type casting to the required type rather than introducing this extra variable.

dyladan (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
dyladan: I agree that a generic is the "correct" way to solve the problem of getting the correct type, but this PR doesn't introduce a new override it just points out the already existing one which is an improvement in my opinion. After the base class gets a generic, these overrides can be removed.

MSNev: Are you sure... Just did a quick test with this
```
export class Test {
    constructor(protected _config: any) {

    }
}

export class ExtraTest extends Test {
    constructor(protected _config: any) {
        super(_config);
    }
}

export class OverrideTest extends Test {
    constructor(protected override _config: any) {
        super(_config);
    }
}
```

And it produces this which would result in properties ```_config``` and ```override``` in the OverrideTest "class"

```
var Test = /** @class */ (function () {
    function Test(_config) {
        this._config = _config;
    }
    return Test;
}());
export { Test };
var ExtraTest = /** @class */ (function (_super) {
    __extends(ExtraTest, _super);
    function ExtraTest(_config) {
        var _this = _super.call(this, _config) || this;
        _this._config = _config;
        return _this;
    }
    return ExtraTest;
}(Test));
export { ExtraTest };
var OverrideTest = /** @class */ (function (_super) {
    __extends(OverrideTest, _super);
    function OverrideTest(override, _config) {
        var _this = _super.call(this, _config) || this;
        _this.override = override;
        return _this;
    }
    return OverrideTest;
}(Test));
export { OverrideTest };
```

Flarna: @MSNev I assume you used TS 4.2 instead 4.3. `override` keyword has been added in 4.3.
your sample in [TS playground](https://www.typescriptlang.org/play?ts=4.3.4#code/KYDwDg9gTgLgBAYwDYEMDOa4BVhvgbwCg4TEIA7PKAVwRmgAowoIZg7gATOAfQQoBmASwDmALjgpyATwCUcIsVIBfQqsKhIsRKgxwAoiBhQUOPHFBtynTGYJKS-Ssdr0oTFmw7c+g0RKk5BQdSEjRqMGB3X3JhEVkAbhDVdU1oeGR0TAB5ADcoqCFOYDsLI2BrW1x7UKcqV0ZmVnY2bgh8qELi3ic4gJl5IlDQ8Mjo3tFE5LVCIA) shows that `override` has no impact to generated JS code

MSNev: So this will implicitly require that anyone who (wants to) consume the raw *.ts code, like people wanting to create their own distro without consuming the pre-compiled packages will need to use TS 4.3+...

If that is the case this needs to be documented.

Flarna: It's documented in the package.json like for all other dev-dependencies.

blumamir (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
dyladan: @MSNev are you on your team doing this (consuming the raw .ts)? If not, I'm not aware of anyone doing that and it seems like not an unreasonable requirement to have a minimum ts version. Also, it isn't _broken_ with earlier versions of typescript, it just has this small caveat.


There are many other places in the code where compiling with a different typescript version will produce different code. Anyone who recompiles our source code for their own purposes should be aware of what they're doing. I don't want to get into the game of making any guarantees that we will support compilation with any version of typescript other than what we are using.

MSNev: Not currently, but this was one of our fallbacks 🤕 when required.
However, we are also still using v2.5.3 (for internal team backward compatibility issues) and v3.9.9 for our different versions (internal and external) -- and I'm now exploring upgrading to 4.3.x.

> Also, it isn't broken with earlier versions of typescript, it just has this small caveat.
Looking (only) at the generated code then it does look wrong as it is now using 2 parameters for the constructor (I'd prefer to have the older TS versions just fail with a Typo style error -- but that's not going to happen)

obecny (approved):
lgtm

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
Flarna: I guess there are more problems with v2.5.3 then `override`. If we intend to support an older TS version we should actually use it everywhere and adapt bot to not update,...
But I would prefer to discuss this in a seperate issue. I fear that we will be not able to update TS for quite a while than and it would be breaking and therefore semver major...

API and Core use `override` already therefore I think we should use it also in contrib for consistency.

@MSNev Is it ok to resolve this conversation here?

vmarchaud: friendly ping @MSNev

MSNev: My preference would be to not live on the bleeding edge, you can use later versions of TS and simply avoid using the latest features. And the override issue above also exists in 3.9.9 (I didn't try 4.2, but it may also have the same problem).

In terms of this conversation, I'm going to leave it there but will refrain from approving as this is not my view of how this should be solved (because of this side effect)

Flarna: I created a followup issue in core to continue the discussion and sync all repos: https://github.com/open-telemetry/opentelemetry-js/issues/2329

I set this conversation to resolved. If decission is to go for compatibility I'm fine with reverting API/core but for now I think consistency is more important.

PR #553, opened 2021-06-30
feat: Add NestJS instrumentation

Rauno56:

Adds a auto-instrumentation for [NestJS](https://nestjs.com/). The current solution is still minimal and only traces controllers and the whole request context. Latter also includes guard, interception and pipe execution, which are currently not traced but can be added later without breaking what's already there.


plugins/node/opentelemetry-instrumentation-nestjs-core/src/version.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// this is autogenerated file, see scripts/version-update.js
+export const VERSION = '0.21.0';
```
vmarchaud, line 18: both should version in this file and in package.json should be 0.23.0 i believe, otherwise lgtm

Rauno56:
Note to myself. If [that](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/594) gets merged, I need to update versions again.

plugins/node/opentelemetry-instrumentation-nestjs-core/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "rxjs": "^7.2.0",
+    "rxjs-compat": "^6.6.7",
+    "semver": "^7.3.5",
+    "ts-mocha": "8.0.0",
+    "typescript": "4.3.2"
+  },
+  "dependencies": {
+    "@opentelemetry/instrumentation": "^0.23.0",
+    "@opentelemetry/semantic-conventions": "^0.23.0"
+  }
```
dyladan, line 77:
```suggestion
  "devDependencies": {
    "@nestjs/common": "^8.0.5",
    "@nestjs/core": "^8.0.5",
    "@nestjs/microservices": "^8.0.5",
    "@nestjs/platform-express": "^8.0.5",
    "@nestjs/websockets": "^8.0.5",
    "@opentelemetry/api": "1.0.1",
    "@opentelemetry/context-async-hooks": "0.24.0",
    "@opentelemetry/node": "0.24.0",
    "@opentelemetry/tracing": "0.24.0",
    "@types/mocha": "7.0.2",
    "@types/node": "14.17.2",
    "@types/semver": "^7.3.8",
    "@types/vinyl-fs": "^2.4.12",
    "codecov": "3.8.2",
    "cross-env": "7.0.3",
    "gts": "3.1.0",
    "mocha": "7.2.0",
    "nyc": "15.1.0",
    "reflect-metadata": "^0.1.13",
    "rimraf": "3.0.2",
    "rxjs": "^7.2.0",
    "rxjs-compat": "^6.6.7",
    "semver": "^7.3.5",
    "ts-mocha": "8.0.0",
    "typescript": "4.3.2"
  },
  "dependencies": {
    "@opentelemetry/instrumentation": "^0.24.0",
    "@opentelemetry/semantic-conventions": "^0.24.0"
  }
```

plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts
```diff
@@ -0,0 +1,252 @@
...
+import type { RouterExecutionContext } from '@nestjs/core/router/router-execution-context.js';
+import type { Controller } from '@nestjs/common/interfaces';
+import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
+import { VERSION } from './version';
+import { AttributeNames, NestType } from './enums';
+
+export class Instrumentation extends InstrumentationBase<typeof NestJS> {
+  static readonly COMPONENT = '@nestjs/core';
+  static readonly COMMON_ATTRIBUTES = {
+    component: Instrumentation.COMPONENT,
```
dyladan, line 35: We are not supposed to be doing the `component` attribute anymore I think.

plugins/node/opentelemetry-instrumentation-nestjs-core/src/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// util.types.isPromise is supported from 10.0.0
+export const isPromise = (value: any): value is Promise<unknown> => {
```
dyladan, line 18: This seems like something that could be useful to more than one instrumentation. Possibly a good candidate for an instrumentation-utils package? Doesn't affect this PR

dyladan (approved):

plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts
Rauno56: I'd prefer that (and adding the version number of the instrumented lib, which I currently tend to implement again and again) to be done somewhere centrally as well.

Do you have a reference to a discussion somewhere?

plugins/node/opentelemetry-instrumentation-nestjs-core/src/utils.ts
Rauno56: Agreed. I'd more specifically 1. see as few of those utilities managed centrally as possible; 2. keep them in the instrumentation package itself.

plugins/node/opentelemetry-instrumentation-nestjs-core/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "gts": "3.1.0",
+    "mocha": "7.2.0",
+    "nyc": "15.1.0",
+    "reflect-metadata": "^0.1.13",
+    "rimraf": "3.0.2",
+    "rxjs": "^7.2.0",
+    "rxjs-compat": "^6.6.7",
+    "semver": "^7.3.5",
+    "ts-mocha": "8.0.0",
+    "typescript": "4.3.2"
```
Flarna, line 72:
```suggestion
    "typescript": "4.3.5"
```
Maybe there are other deps which could be updated

plugins/node/opentelemetry-instrumentation-nestjs-core/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "nestjs",
+    "nodejs",
+    "tracing",
+    "profiling",
+    "instrumentation"
+  ],
+  "author": "OpenTelemetry Authors",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=8.0.0"
```
Flarna, line 32:
```suggestion
    "node": ">=8.5.0"
```

plugins/node/opentelemetry-instrumentation-nestjs-core/package.json
```diff
@@ -0,0 +1,78 @@
...
+  },
+  "files": [
+    "build/src/**/*.js",
+    "build/src/**/*.d.ts",
```
Flarna, line 36:
```suggestion
    "build/src/**/*.js.map",
    "build/src/**/*.d.ts",
```

plugins/node/opentelemetry-instrumentation-nestjs-core/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Instrumentation } from './instrumentation';
+
+export * from './instrumentation';
+export default Instrumentation;
```
Flarna, line 20: why do we add a default export here?

Rauno56: Why not? It's essentially one-class-package, so the default doesn't have any "competition".
In that way instrumentation can be just exported more conveniently without knowing the name I chose for the instrumentation class. Dynamic loading (if that will ever be a thing. Currently for many reasons, it's not obvious why it should) would be easier and it's just a small improvement on convenience in some cases. Interested in your thoughts though.

plugins/node/opentelemetry-instrumentation-nestjs-core/package.json
Rauno56: There were! Thanks!

plugins/node/opentelemetry-instrumentation-nestjs-core/package.json
Rauno56: There are quite a few packages that have `>=8.0.0` as the engine requirement. I'll make a note to have a PR to bump those to the actual supported version as well!

plugins/node/opentelemetry-instrumentation-nestjs-core/src/index.ts
Flarna: I don't care that much if it is done here or not.
I was just wondering as I haven't seen this at other places. Don't think there is a single export default in API or Core packages.

Rauno56: Yeah... And for API or Core, I understand why - they are kind of bundles of many useful blocks.

dyladan: I would prefer to do what the other instr packages do for the sake of consistency. If we want to change it then it should be changed for all of them in another PR.

plugins/node/opentelemetry-instrumentation-nestjs-core/test/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      assert.strictEqual(span.attributes.component, '@nestjs/core');
+      assert.strictEqual(span.attributes['nestjs.module'], expected.module);
+
+      assert.strictEqual(span.name, expected.name);
+
+      assert.strictEqual(span.attributes['http.method'], expected.method);
+      assert.strictEqual(span.attributes['http.url'], expected.url);
+      assert.strictEqual(span.attributes['http.route'], expected.path);
+      assert.strictEqual(span.attributes['nestjs.callback'], expected.callback);
+      assert.strictEqual(
```
johnbley, line 172: I don't see any asserts covering `nestjs.type` - it looks your tests cover all the types of spans your instrumentation creates, but I don't see anything validating this attribute?

johnbley (approved):

plugins/node/opentelemetry-instrumentation-nestjs-core/test/index.test.ts
Rauno56: That is right! Thanks!

plugins/node/opentelemetry-instrumentation-nestjs-core/src/index.ts
Rauno56: Even though it's additive - the "normal" way stays the same? For what it's worth I've added the default export before to instrumentations: Knex, Mysql2..

Rauno56: @dyladan, is this holding the merge back?

MSNev: If a package only exports 1 thing (and only ever will) then `default` make sense. But if something exports multiple keys as in this case, then (my preference) is to not use `default` mainly from an end user usage perspective.

Rauno56: I can be convinced, but I'd like a more compelling argument for "the end user usage perspective". Even if the instrumentation class is not strictly the only thing, it's **always** the **main** thing, don't you think?

Let me explain:

The vast majority of instrumentation packages will export one thing: The instrumentation class. Everyone hopefully knows how to use it - you can pass it to "registerInstrumentation" call to enable it, and call it a day.

With default export, there's one thing you need to know: the package name(which you needed to install it).
Without default export, you need to know **twice**(a little dramatization :wink:) as many things: the package name(to import) **and** the name that was given to the Instrumentation class inside the package(`NestInstrumentation`).

Some instrumentations also have types and enums that are exported but importing those is (hopefully - since we do want to default to the most useful configuration) an edge case.

vmarchaud: > it's always the main thing, don't you think?

While its true it isn't the elephant in the room IMO, i agree with @dyladan that we have done that since the opencensus days and if we want this to change, this would need be changed everywhere and not having specific package that does this and other use a named export.
It's a problem for me at least for consistency sake

dyladan:
Can you please add yourself to `.github/component_owners.yml`

plugins/node/opentelemetry-instrumentation-nestjs-core/src/index.ts
Rauno56: Just to be clear - I wasn't suggesting removing named import, but adding a default one for convenience.
The rule of "always have default export"(that's how I took your premise) cannot and should not be enforced generally because there are cases where it is not applicable so chasing that consistency - is that a good reason to never use it? I don't think so.

But the general sentiment from you is clear and I respect that - removed the default export.

MSNev (approved):

.github/component_owners.yml
```diff
@@ -1,19 +1,26 @@
...
-  plugins/node/opentelemetry-instrumentation-restify:
+  plugins/node/opentelemetry-instrumentation-generic-pool:
     - rauno56
-  plugins/node/opentelemetry-instrumentation-router:
+  plugins/node/opentelemetry-instrumentation-knex:
     - rauno56
   plugins/node/opentelemetry-instrumentation-memcached:
     - rauno56
-  plugins/node/opentelemetry-instrumentation-knex:
+  plugins/node/opentelemetry-instrumentation-nestjs-core:
```
dyladan, line 19: Can you update it to be `Rauno56`? At the current time the action is case sensitive

Issue #561, opened 2021-07-05
MySQL spans do not link to the active parent span


sstone1:
### What version of OpenTelemetry are you using?
`^0.21.0`

### What version of Node are you using?
`v14.17.1`

### What did you do?

We are using OpenTelemetry in a LoopBack 4 application, which connects to a MySQL database and exposes a CRUD API over a database table. LoopBack 4 is built on top of the `express`, `http` and `mysql` libraries, so we are using the instrumentation for these.

### What did you expect to see?
We expected to see a single trace for each HTTP request to the LoopBack 4 application, containing various spans - including a span for the request to the MySQL database. 

### What did you see instead?
We see two traces for each HTTP request to the LoopBack 4 application, one with all the HTTP spans, and another with just the MySQL span. 

### Additional context

I believe that this problem occurs as the context is getting lost when requesting a connection from the connection pool. I have a patch for the instrumentation code that fixes the problem by binding the callback for `getConnection` to the active context before passing it into the instrumented code.

PR #562, opened 2021-07-05
fix(mysql): bind get connection callback to active context

sstone1:
Signed-off-by: Simon Stone <simonstone1@gmail.com>

- The problem described in issue #561 

- The callback for `getConnection` is now bound to the active context before it is passed into the instrumented `mysql` code

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
@@ -34,7 +34,10 @@ import { VERSION } from './version';
 
 type formatType = typeof mysqlTypes.format;
 
-type getConnectionCallbackType = (err: mysqlTypes.MysqlError, connection: mysqlTypes.PoolConnection) => void;
+type getConnectionCallbackType = (
+  err: mysqlTypes.MysqlError,
+  connection: mysqlTypes.PoolConnection
+) => void;
 
 export class MySQLInstrumentation extends InstrumentationBase<
   typeof mysqlTypes
@@ -209,10 +212,17 @@ export class MySQLInstrumentation extends InstrumentationBase<
     };
   }
 
-  private _getConnectionCallbackPatchFn(cb: getConnectionCallbackType, format: formatType) {
+  private _getConnectionCallbackPatchFn(
+    cb: getConnectionCallbackType,
+    format: formatType
+  ) {
     const thisPlugin = this;
     const activeContext = context.active();
-    return function (this: any, err: mysqlTypes.MysqlError, connection: mysqlTypes.PoolConnection) {
+    return function (
+      this: any,
+      err: mysqlTypes.MysqlError,
+      connection: mysqlTypes.PoolConnection
+    ) {
       if (connection) {
         // this is the callback passed into a query
         // no need to unwrap
--- a/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
@@ -34,6 +34,8 @@ import { VERSION } from './version';
 
 type formatType = typeof mysqlTypes.format;
 
+type getConnectionCallbackType = (err: mysqlTypes.MysqlError, connection: mysqlTypes.PoolConnection) => void;
+
 export class MySQLInstrumentation extends InstrumentationBase<
   typeof mysqlTypes
 > {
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -203,6 +203,13 @@ export class MySQLInstrumentation extends InstrumentationBase<
 
   private _getConnectionCallbackPatchFn(cb: Function, format: formatType) {
     const thisPlugin = this;
+    // If a parent span exists, bind the callback so that the parent span
+    // is still visible after we've gone through the MySQL pool code.
+    const activeContext = context.active();
+    const parentSpan = trace.getSpan(activeContext);
+    if (parentSpan && typeof cb === 'function') {
```
vmarchaud, line 210: I don't believe you should check for span here, context can propagate other things (like baggage) without necessarily having a span, checking that cb is a fn should be enough !

vmarchaud: Also you could move the bind to when the callback is actually called line 226

sstone1: @vmarchaud thanks for taking a look! I've removed the check for the parent span, and used `with` instead of `bind` at the point that the callback gets called - I think `with` makes more sense now?

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
vmarchaud: @sstone1 Yes, lgtm now ! thanks for taking the time to fix this :)

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -216,7 +223,9 @@ export class MySQLInstrumentation extends InstrumentationBase<
         }
       }
       if (typeof cb === 'function') {
-        cb(...arguments);
+        context.with(activeContext, () => {
```
obecny, line 226: can you please cover that with unit test ?

sstone1: @obecny done, please can you take another look?

Flarna, line 226: `this` is not passed to the callback here.

I think following should work which additionally avoids the allocation of the arrow function:
```js
context.with(activeContext, cb, this, ...arguments)
```

vmarchaud: I prefer your version because its safer but shouldn't the correct `this` be inhered since its an arrow fn ?

Flarna: The `this` visible in the arrow function should be the correct one but it's not forwarded into `cb`. It would be needed use e.g. `Reflect.apply(cb, this, arguments)` to forward it.
But as mentioned above it's not needed to allocate the arrow function at all.

obecny: @vmarchaud pls avoid resolving comments which you are not author of - I have missed this conversation as it was resolved already :/

obecny: @sstone1 pls don't force push after receiving first review or comment from anyone, it is impossible to see changes. The PR is stashed once it is merged anyway

obecny (approved):
lgtm

MSNev (approved):

vmarchaud (changes requested):
requesting changes so we don't merge it before addresing flarna feedback

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
```diff
@@ -426,6 +426,30 @@ describe('mysql@2.x', () => {
...
+            const spans = memoryExporter.getFinishedSpans();
+            assert.strictEqual(spans.length, 1);
+            const querySpan = spans[0];
+            assert.strictEqual(
+              querySpan.parentSpanId,
+              parentSpan.spanContext().spanId
+            );
+            done();
+          });
+        });
```
blumamir, line 450: This is a very verbose way to check that the context is correctly set in the callback.
Alternative and more direct option is to extract `trace.getSpan(context.active())` in the callback and compare it to `parentSpan`

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
sstone1: @Flarna thanks, I've made changes to use `context.with` without the arrow function instead. I had to make some additional changes around types to make the compiler happy though.

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
sstone1: @blumamir thanks, I've changed the test.

Flarna (approved):

vmarchaud (approved):

PR #570, opened 2021-07-13
fix(package.json): publish source maps

blumamir:

- publish source maps files to npm. This is already enabled in some of the packages in this repo (but missing in others), and was [added to core repo](https://github.com/open-telemetry/opentelemetry-js/pull/1237) a year ago:

- add "build/src/**/*.js.map" to "files" section in package.json


Flarna:
Shouldn't we include the ts files also?
At least VsCode complains that it can't find ts files if there are source map files pointing to "nowhere".

blumamir:
> Shouldn't we include the ts files also?
> At least VsCode complains that it can't find ts files if there are source map files pointing to "nowhere".

Don't really know, 
I fix this due to warnings about missing source map files in one of our client's IDE, which was solved by adding the js.map files to npm publish.
I remember it was missing in core and then added in [the PR I mentioned above](https://github.com/open-telemetry/opentelemetry-js/pull/1237). Then we added it to the instrumentations in ext-js [here](https://github.com/aspecto-io/opentelemetry-ext-js/pull/147) and the warning went away for these packages.
The packages from the contrib still generate warnings about missing source map files which I believe will also be solved if we include the ".js.map" in the distribution.

MSNev:
> Shouldn't we include the ts files also?

If we change the ```sourceMap``` in the tsconfig.json to ```inlineSourceMap``` then no as the generated sourcemap will then contain the original source. While this makes the *.map files huge its a lot easier to handle as it includes the original code.

I'm not sure about any other packaging switches / tools that would also need additional config changes to retain these (like webpack for example).

obecny (approved):
lgtm

MSNev (approved):

vmarchaud (approved):

PR #571, opened 2021-07-13
fix(instrumentation-hapi): change root span name to route name

### LINKED ISSUE #453: use route as root span name in hapi instrumentation

Currently the root span name is set by the http instrumentation to a generic name like `HTTP GET`. Additionally, `http.route` of the root span is set to the actual http pathname and not the route. So, something like `http.route: /events/123` instead of `http.route: /events/{eventId}`.

This makes traces hard to identify (in Honeycomb in this case):
Bildschirmfoto 2021-04-26 um 14 44 19

The actual hapi route is buried somewhat deep in the trace:
Bildschirmfoto 2021-04-26 um 14 43 02

Similarly to the way it's done for Koa (#327) and Express (#176, #273) the root span name should be renamed once the route has been determined and `http.route` should be set to the actual `route` at that point.

None

Thank you for your work on OpenTelemetry and all these instrumentations :)

CptSchnitz:

- fixes #453 

- root span is renamed once the route information is known (name is like in the express instrumentation)
- http.route attribute is also set on the root span
- added unit test to cover this


plugins/node/opentelemetry-instrumentation-hapi/src/utils.ts
```diff
@@ -94,6 +94,20 @@ export const getRouteMetadata = (
...
+  route: Hapi.ServerRoute
+): {
+  attributes: SpanAttributes;
+  name: string;
+} => {
+  return {
+    attributes: {
+      [SemanticAttributes.HTTP_ROUTE]: route.path,
+    },
+    name: `${route.method} ${route.path}`,
```
dyladan, line 107: Is `route.path` the parameterized path like `/users/:id` or the full path like `/users/123`?

dyladan (approved):
LGTM thanks

MSNev (approved):

Flarna (approved):

CptSchnitz:
@dyladan happy to help ☺️.
As long as there are issues I'll try to help.

plugins/node/opentelemetry-instrumentation-hapi/src/utils.ts
obecny, line 107: can a `route.path` be empty if so then

```suggestion
    name: `${route.method} ${route.path || '/'}`,
```

CptSchnitz: from looking at hapi docs it looks like the path must start with `/` so I don't think this is necessary.
```
path - (required) the absolute path used to match incoming requests (must begin with '/'). Incoming requests are compared to the configured paths based on the server's router configuration. The path can include named parameters enclosed in {} which will be matched against literal values in the request as described in Path parameters.
```
https://github.com/hapijs/hapi/blob/master/API.md#-serverrouteroute

obecny: make sense, thx

obecny (approved):
lgtm

vmarchaud (approved):

klaemo:
Thank you so much for working on this @CptSchnitz ☺️🚀

PR #580, opened 2021-07-19
fix: redis instrumentation loses context when using callbacks

aspectom:

- When using redis instrumentation with callbacks instead of async await, original span context is lost.

- Call context.with when calling original callback


plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
```diff
@@ -146,7 +147,10 @@ export const getTracedInternalSendCommand = (
         }
 
         endSpan(span, err);
-        return originalCallback.apply(this, arguments);
+        const callbackThis = this;
+        return context.with(originalContext, () => {
+          return originalCallback.apply(callbackThis, arguments);
+        });
```
dyladan, line 153: You can also just do `return context.with(originalContext, originalCallback, callbackThis, ...arguments)` https://open-telemetry.github.io/opentelemetry-js-api/classes/contextapi.html#with

blumamir (approved):

plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
dyladan: This way you avoid creating an additional function call

aspectom: Wasn't aware of that one. thanks @dyladan , fixed :)

MSNev (approved):

plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
```diff
@@ -125,6 +125,7 @@ export const getTracedInternalSendCommand = (
       );
     }
 
+    const originalContext = context.active();
```
Flarna, line 128: You could move this line down into the `if (originalCallback) {`.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
aspectom: Thanks @Flarna, fixed.

vmarchaud (approved):

Rauno56 (approved):

obecny (approved):

PR #584, opened 2021-07-21
feat(pg): allow overriding command extraction

FauxFaux:

Command extraction (the code which generates `pg.query:select` from `select * from foo`) fails for us, as we comment our SQL, e.g. `/* getUser@main.js:17 */ select * from user where id=?`. Currently, all of our spans are named `pg.query:/*`, which is, uh, not useful.

This PR allows us to override the implementation with one which can extract `select` in this case.

- Config option allows specification of `getCommandFromText`.
- `getCommandFromText` defers to the config implementation, if it is set.
- Tests for before and after.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
@@ -44,7 +44,11 @@ function arrayStringifyHelper(arr: Array<unknown>): string {
 }
 
 // Helper function to get a low cardinality command name from the full text query
-function getCommandFromText(text?: string): string {
+function getCommandFromText(
+  config: Pick<PgInstrumentationConfig, 'getCommandFromText'>,
+  text?: string
+): string {
+  if (config.getCommandFromText) return config.getCommandFromText(text);
   if (!text) return 'unknown';
   const words = text.split(' ');
   return words[0].length > 0 ? words[0] : 'unknown';
@@ -81,7 +85,10 @@ export function handleConfigQuery(
   queryConfig: NormalizedQueryConfig
 ) {
   // Set child span name
-  const queryCommand = getCommandFromText(queryConfig.name || queryConfig.text);
+  const queryCommand = getCommandFromText(
+    instrumentationConfig,
+    queryConfig.name || queryConfig.text
+  );
   const name = PgInstrumentation.BASE_SPAN_NAME + ':' + queryCommand;
   const span = pgStartSpan(tracer, this, name);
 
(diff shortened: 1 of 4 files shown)


plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -44,7 +44,11 @@ function arrayStringifyHelper(arr: Array<unknown>): string {
 }
 
 // Helper function to get a low cardinality command name from the full text query
-function getCommandFromText(text?: string): string {
+function getCommandFromText(
+  config: Pick<PgInstrumentationConfig, 'getCommandFromText'>,
+  text?: string
+): string {
+  if (config.getCommandFromText) return config.getCommandFromText(text);
```
obecny, line 51: I think this function should not depend on config, it has name `getCommandFromText` so it extracts command from text and nothing more. I think this check / call for this function from config - this should be done in `upper level` or create a new helper for that if it will be repeated in more places. WDYT ?

FauxFaux: I went this way around as it's quite a bit of code to inline, especially if you think `(config.getCommandFromText ?? getCommandFromText)(text)` is ugly; you'd end up adding a function for `function getCommandFromTextMaybeFromConfig(config, text) { if (config.getCommandFromText) return config.getCommandFromText(text); return getCommandFromText(text);` which seemed excessive?

FauxFaux: Alternatively you could override the default value in the config at construct time, but nothing else seemed to be doing that.

Rauno56:
I wonder whether the operation might already be parsed and reliably available from somewhere in the client driver. I haven't looked into `pg` instrumentation, but `knex`(ORM on top of PG or MySQL drivers) provided it and I'd expect(not sure though) it got that from the actual driver it uses.

To me, it seems that the current implementation is buggy and the preferred solution is not to add an option to overwrite the parser but to fix the issue.

I think it's also worth noting that the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/f32b45b55a00edcd8616e0fa50768adf56bb0c0c/specification/trace/semantic_conventions/database.md) also votes against it(about `db.operation` property):

>  it is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if the operation name is provided by the library being instrumented.

FauxFaux:
Even if that was the implementation, I would still be raising a PR to override the operation name, because the default operation names are not useful. The function I've provided (for overriding) doesn't have to parse the query.

Issue #550, opened 2021-06-29
New Instrumentation - connect


obecny:
https://www.npmjs.com/package/connect

### Describe the solution you'd like to see
Add instrumentation for connect

### Additional context
weekly downloads: 4,5 mln

PR #588, opened 2021-07-23
chore: adding instrumentation for connect

obecny:

- Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/550

- Ads auto instrumentation for connect


vmarchaud (commented):
lgtm, just need to fix tests

examples/connect/server.js
```diff
@@ -0,0 +1,35 @@
...
+
+app.use((req, res, next) => {
+  next();
+});
+
+app.use('/run_test', async (req, res) => {
+  const result = await axios.get('https://raw.githubusercontent.com/open-telemetry/opentelemetry-js/main/package.json');
+  tracing.log('sending response');
+  res.end(`OK ${result.data.version}`);
+  if (tracing.connectInstrumentation.isEnabled()) {
```
Rauno56, line 26: I can only guess that the reason you disable the instrumentation is to, on every second request, showcase how the traces will look without Connect instrumentation.
A nit: Perhaps a comment this, or have another endpoint do that.

examples/connect/server.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+// eslint-disable-next-line
+const tracing = require('./tracing')('example-connect-server');
+
+// Require in rest of modules
+const connect = require('connect');
+const axios = require('axios');
+
+// Setup express
```
Rauno56, line 10:
```suggestion
// Setup connect
```
In the example package name I'd actually leave express there(or some other kind of specifier, since the package name is so generic).

examples/connect/server.js
```diff
@@ -0,0 +1,35 @@
...
+  next();
+});
+
+app.use((req, res, next) => {
+  next();
+});
+
+app.use('/run_test', async (req, res) => {
+  const result = await axios.get('https://raw.githubusercontent.com/open-telemetry/opentelemetry-js/main/package.json');
+  tracing.log('sending response');
```
Rauno56, line 24: I'd just use `console.log`.

plugins/node/opentelemetry-instrumentation-connect/test/instrumentation.test.ts
```diff
@@ -0,0 +1,219 @@
...
+  },
+};
+
+const getNextPort = (port: number, count = 0) => {
+  if (count > 10) {
+    return Promise.reject('too many retries');
+  }
+  return new Promise((resolve, reject) => {
+    const timer = setTimeout(() => {
+      getNextPort(++port, ++count).then(resolve, reject);
```
Rauno56, line 57: You can ask for a random number by using 0 for the port number and then later reading the actual used port from the initiated server.

plugins/node/opentelemetry-instrumentation-connect/README.md
```diff
@@ -0,0 +1,70 @@
...
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for [`connect`](https://github.com/senchalabs/connect).
+
+For automatic instrumentation see the
+[@opentelemetry/node](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-node) package.
+
+## Installation
+
+This instrumentation relies on HTTP calls to also be instrumented. Make sure you install and enable both, otherwise you will not see any spans being exported from the instrumentation.
```
Rauno56, line 15: Does it really absolutely require http instrumentation? I don't see that in the current implementation.

Rauno56 (approved):
Few nits, otherwise beautifully crafted! :+1:

examples/connect/server.js
obecny: it has additional time so it is easier to compare it with time of span

plugins/node/opentelemetry-instrumentation-connect/README.md
obecny: updated description basically it will export spans which will have no common parent, http is quite important here the same as in express

examples/connect/server.js
obecny: added

plugins/node/opentelemetry-instrumentation-connect/test/instrumentation.test.ts
obecny: it would be just additional logic, whereas it really doesn't matter, it is useful especially during testing when something hangs etc.

dyladan: You are searching for an open port which can fail. What @Rauno56 is saying is you can start the server with port 0 and it is guaranteed to be assigned to an open port.

Rauno56: The current solution(using another server to find the port and then closing it) works as well I guess, but you could just start the server you will be using for requests on port 0 and detect the port it's running instead of allocating, starting and closing another server.

Rauno56:
@obecny, see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/588#discussion_r681653581
There's an example of that in [koa tests](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-koa/test/koa.test.ts).

plugins/node/opentelemetry-instrumentation-connect/test/instrumentation.test.ts
obecny: I had already changed that 5 days ago guys .... :)

Rauno56: I know - my last comment was about that changed version.

vmarchaud (approved):

dyladan:
What should we do about the component_owners.yml in the case maintainers add the instrumentation? Is it ok to have an unowned instr or should you own it?

vmarchaud:
> What should we do about the component_owners.yml in the case maintainers add the instrumentation? Is it ok to have an unowned instr or should you own it?

Well i think we should add them for every instrumentation in case there is a change in maintainer ? Specially since its one of us that have wrote it, it will force us to switch ownership when stepping down.
However it can be done later on in another PR, we need to decide who own each of the instrumentations anyway

dyladan:
> > What should we do about the component_owners.yml in the case maintainers add the instrumentation? Is it ok to have an unowned instr or should you own it?
> 
> Well i think we should add them for every instrumentation in case there is a change in maintainer ? Specially since its one of us that have wrote it, it will force us to switch ownership when stepping down.
> However it can be done later on in another PR, we need to decide who own each of the instrumentations anyway

Stepping down as maintainer wouldn't mean stepping down as component owner I think.

vmarchaud:
> Stepping down as maintainer wouldn't mean stepping down as component owner I think

Not necessarely i agree, however if a maintainer don't have the time anymore to work on otel i suppose he would not have the time to review PRs about components he owns (in the case of a step down because of lack of time of course).

PR #589, opened 2021-07-24
fix(hapi-instrumentation): close spans on errors in instrumented functions

### LINKED ISSUE #572: route span is missing when route throws an error in hapi instrumentation
api - 1.0.1
core - 0.23.0
contrib - 0.23.0

v12.20.0

### What did you do?

I created a simple hapi server based on the example that only throws errors.
I made a few requests to it's endpoints, one throws a normal error, and the other using boom.
the server can be found here:
https://github.com/CptSchnitz/instrumented-hapi-with-errors

I expected to see the route span with its status set on error.

### What did you see instead?
the route span is missing.
image

### Additional context

I would happy to work on fixing this once you confirm its an actual bug.

CptSchnitz:

- closes #572 

- caught and recorded exceptions in extMethod wrapper and ended the span.
- caught and recorded exceptions in routeHandler wrapper and ended the span.
- added tests to cover both scenarios.


plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
```diff
@@ -386,10 +396,19 @@ export class HapiInstrumentation extends InstrumentationBase {
         const span = instrumentation.tracer.startSpan(metadata.name, {
           attributes: metadata.attributes,
         });
-        const res = await oldHandler(request, h, err);
-        span.end();
-
-        return res;
+        try {
+          const res = await oldHandler(request, h, err);
+          return res;
```
Rauno56, line 401: Don't need the assignment nor `await` here. Just `return oldHandler(...);`

plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
```diff
@@ -337,15 +337,25 @@ export class HapiInstrumentation extends InstrumentationBase {
...
+        } catch (err) {
+          span.recordException(err);
+          span.setStatus({
+            code: api.SpanStatusCode.ERROR,
+            message: err.message,
+          });
+          throw err;
+        } finally {
+          span.end();
+        }
```
Rauno56, line 358: `contex.with` passes return values through so the closure is not required here.
This or something similar should work:
```typescript
        try {
          return api.context.with(
            api.trace.setSpan(api.context.active(), span),
            () => method(...params)
          );
        } catch (err) {
          span.recordException(err);
          span.setStatus({
            code: api.SpanStatusCode.ERROR,
            message: err.message,
          });
          throw err;
        } finally {
          span.end();
        }
```
Additionally, `with` also takes arguments to be passed to the function, so it can be simplified even further:
```typescript
        try {
          return api.context.with(
            api.trace.setSpan(api.context.active(), span),
            method,
            ...params
          );
        } catch (err) {
          span.recordException(err);
          span.setStatus({
            code: api.SpanStatusCode.ERROR,
            message: err.message,
          });
          throw err;
        } finally {
          span.end();
        }
```
Test it out!
Docs: https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_context_async_hooks.asynchookscontextmanager.html#with

Rauno56 (commented):
Thanks for yet another contribution! The implementation can be simplified - see the comments.

plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
CptSchnitz: removed the unneeded assignment.
as for the await, it is needed as catch won't work on promises that are not awaited, and I also need to wait to close the span after the action was completed.
maybe i should add a test that includes an async handler?
Also, there is a  handy eslint rule that covers that https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/return-await.md.

plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
CptSchnitz: Thanks! it looks much better now.

blumamir (approved):

dyladan (approved):

MSNev (approved):

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
```diff
@@ -337,15 +337,21 @@ export class HapiInstrumentation extends InstrumentationBase {
...
-            res = await method(...params);
-          }
-        );
-        span.end();
-        return res;
+        try {
+          return await api.context.with(
+            api.trace.setSpan(api.context.active(), span),
+            () => method(...params)
+          );
```
dyladan, line 344: I think @Rauno56 was suggesting this, which avoids creating an additional closure and the overhead associated.

```suggestion
          return await api.context.with(
            api.trace.setSpan(api.context.active(), span),
            method,
            undefined,
            ...params
          );
```

dyladan: https://open-telemetry.github.io/opentelemetry-js-api/classes/contextapi.html#with

CptSchnitz: made the change.
For some reason typescript didn't infer the generic types correctly so I had to set them manually.

Issue #591, opened 2021-07-28
@opentelemetry/instrumentation-pino failed to compile


legendecas:
### What version of OpenTelemetry are you using?

- main branch

### What version of Node are you using?

- v14.17.0

### What did you do?

- `npm i` in a newly cloned repo

### What did you expect to see?

- Success to install and compile all packages

### What did you see instead?

```
> @opentelemetry/instrumentation-pino@0.23.0 version:update /Users/lucas/Developer/github/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-pino
> node ../../../scripts/version-update.js

node_modules/@types/pino/index.d.ts:197:8 - error TS2709: Cannot use namespace 'SonicBoom' as a type.

197     ): SonicBoom;
           ~~~~~~~~~

node_modules/@types/pino/index.d.ts:205:57 - error TS2709: Cannot use namespace 'SonicBoom' as a type.

205     function extreme(fileDescriptor?: string | number): SonicBoom;
                                                            ~~~~~~~~~
```

### Additional context

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/54756

PR #592, opened 2021-07-28
chore(deps): upgrade @types/pino to be compatible with latest sonic-stream types

legendecas:

- #591 

- upgrade @types/pino to 6.3.11

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pino/package.json
+++ b/plugins/node/opentelemetry-instrumentation-pino/package.json
@@ -66,7 +66,7 @@
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.23.0",
-    "@types/pino": "6.3.9",
+    "@types/pino": "6.3.11",
     "semver": "^7.3.5"
   }
 }


Flarna:
Does this still work for users using an older version of @types/pino?

A while ago we had a discussion if we should pin the version here or not. Or even use `*` to accept any version selected in the final app.

Flarna (approved):

legendecas:
> @Flarna Does this still work for users using an older version of @types/pino?

IIUC, types maintained in DefinitelyTyped versioned with the same `minor.patch` following the original package (see https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library). If there are no mistakes, the version of type packages should follow semantic versioning.

I'm not sure which condition you are referring to. If you mean users that declared their own @types/pino dependencies in their package, as long as the user-defined version is compatible with the version defined in @opentelemetry/instrumentation-pino, it should be fine.

legendecas:
> Or even use * to accept any version selected in the final app.

IIUC, the instrumentation package should set a version range of pino that it supports (like say, declare a set of version range in peer dependencies). A `*` sounds not great to me.

Flarna:
@types packages follow major.minor of their original. They may add/remove/rework definitions during this resulting in breakage like here. So in short @types packages don't follow semver (and this is documented at DefinitelyTyped).


> I'm not sure which condition you are referring to. If you mean users that declared their own @types/pino dependencies in their package, as long as the user-defined version range is compatible with the version defined in @opentelemetry/instrumentation-pino, it should be fine.

I mean the version selected by the end user. `@opentelemetry/instrumentation-pino` uses a fixed version therefore a mismatch is likely. In that case NPM installs both versions (one as child of `@opentelemetry/instrumentation-pino` and one in the root node_modules folder).

dyladan (approved):

dyladan:
Merging this since it is fixing the build which breaks in every PR

PR #606, opened 2021-08-06
chore: add ownership for aws propagator

dyladan:
First official component owners @willarmiros and @NathanielRN

### DIFF
--- a/.github/component_owners.yml
+++ b/.github/component_owners.yml
@@ -1,6 +1,12 @@
 # Test configuration
 
 components:
-  propagators/opentelemetry-propagator-aws-xray:
-    - willarmiros
+  packages/opentelemetry-id-generator-aws-xray:
     - NathanielRN
+    - willarmiros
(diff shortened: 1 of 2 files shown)


.github/component_owners.yml
```diff
@@ -0,0 +1,6 @@
+# Test configuration
+
+components:
```
willarmiros, line 3: Shall we also own https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/opentelemetry-id-generator-aws-xray?

willarmiros: And https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-lambda?

obecny (approved):
lgtm

.github/component_owners.yml
dyladan: yes

willarmiros (approved):
🎉

Issue #595, opened 2021-07-28
Publish test-utils as a package on NPM


jonchurch:
Per discussion at the JS SIG meeting on 7/28/2021, it was decided to publish the [test-utils](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/opentelemetry-test-utils) as a package to improve workflow of contributors.

Needs a readme, ~and to be moved out into it's own folder~ (should it be moved into the opentelemetry-js repo?)

@Rauno56 volunteered to work on this. 

Refs: https://github.com/open-telemetry/opentelemetry-js/discussions/2085

Rauno56:
Thanks for doing the legwork of documenting it!

It's already in its own package [here](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/opentelemetry-test-utils), but it just happens not to be published.

Rauno56:
Pumped into this one again, @dyladan. Please publish `@opentelemetry/test-utils`.

PR #621, opened 2021-08-15
chore: upstream mocha instrumentation testing plugin from ext-js

blumamir:
Read more about it [here](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-testing-utils).
Basically, it's a mocha plugin that implements many of the repeated boilerplate code needed for instrumentation testing, in a consistent, idiomatic fashion.
It makes writing instrumentation tests an easier task, with less code, and less room for doing things wrong, by ensuring that there is only a single instance of the instrumentation under test, and cleaning state between tests runs.

This PR is the first step for upstreaming Aspecto's opensource instrumentations from [ext-js repo](https://github.com/aspecto-io/opentelemetry-ext-js) to the contrib. I upstream it first in order to create smaller, more reviewable PRs, with less changes in each.

The next step will be to upstream the `aws-sdk` instrumentation which relies on this mocha plugin for it's unit tests

- When writing unit test for instrumentation, there is a lot of boilerplate things to do: create trace provider, memory exporter and span processor, context manager, create the instance of the instrumentation under test. These components need to be initialized, configured, cleaned between tests, enabled\disabled and managed in each instrumentation in the repo.
For one `.spec` file this is relatively easy, but when instrumentation has multiple `.spec.ts` files, each file has to set these up again and again, and due to the way mocha runs test files, it's very easy to mess it up and cause spans to be written to the wrong memory exporter, update the wrong instrumentation instance, use provider from other file etc.
This PR is a suggestion for mocha plugin and documented workflow which aims to solve many of these issues by using singleton instances of these components, similar to how the SDK is expected to be setup and run in an instrumented application.

- move source files in `@opentelemetry/test-utils` from root to `src` directory and adjust `package.json` and `tsconfig.json` accordingly, so the structure matches the rest of the packages in the repo
- add mocha plugin to `test-utils` package which can be used in instrumentation tests, along with documentation on how to use in the package README.
- refactor mongodb instrumentation tests to demonstrate the use of the plugin in an instrumentation from contrib


packages/opentelemetry-test-utils/README.md
```diff
@@ -2,6 +2,60 @@
 
 This is a internal utils package used across the contrib packages. No guarantees are given to uses outside of [open-telemetry/opentelemetry-js-contrib](https://github.com/open-telemetry/opentelemetry-js-contrib/) repository.
 
+## Instrumentation Testing
+This package exports a mocah [root hook plugin](https://mochajs.org/#root-hook-plugins), which implements common boilerplate code a developer probably needs for writing instrumentation unit tests in node.
```
obecny, line 6:
```suggestion
This package exports a mocha [root hook plugin](https://mochajs.org/#root-hook-plugins), which implements common boilerplate code a developer probably needs for writing instrumentation unit tests in node.
```

packages/opentelemetry-test-utils/src/instrumentations/instrumentation-singelton.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+const _global = global as OTelInstrumentationSingeltonGlobal;
+
+export const getInstrumentation = <T extends InstrumentationBase>():
+  | T
+  | undefined => {
+  return _global[OTEL_TESTING_INSTRUMENTATION_SINGLETON] as T;
+};
+
+export const registerInstrumentation = <T extends InstrumentationBase>(
```
obecny, line 33: we have `registerInstrumentations` which sounds very similar, to avoid confusion my suggestion would be to just rename to something more meaningful to avoid using this accidentally in real app.

```suggestion
export const registerInstrumentationTesting = <T extends InstrumentationBase>(
```

obecny (commented):
I like the idea, just one concern towards the name for registerInstrumentation, other than that lgtm

packages/opentelemetry-test-utils/src/instrumentations/instrumentation-singelton.ts
blumamir: Good point. I choose the name specifically so it matched the name of the function used in the regular, non-testing setup. But now that you mentioned it, I support changing the name to make it explicit that the function is only relevant for tests, and to avoid confusion.

blumamir:
> I like the idea, just one concern towards the name for registerInstrumentation, other than that lgtm

Fixed
Thanks for the review :)

dyladan (approved):

dyladan:
Did you mean to apply it to the mongodb instrumentation or was that just a test?

blumamir:
> Did you mean to apply it to the mongodb instrumentation or was that just a test?

I did, to demonstrate how to use it.
Is it ok?

dyladan:
Yes it's fine, just wanted to make sure it was intended

obecny (approved):

vmarchaud (approved):

dyladan:
Looks like node 16 is failing. Any idea why that's happening?

blumamir:
> Looks like node 16 is failing. Any idea why that's happening?

After merging with the lerna hoisting PR, now mocha is hoisted to the root node_modules, but it depends on the `test-utils` package which is not hosted. thus it fails to require it.

Adding `mocha` to the `--no-hoist` list will resolve the issue, but add an additional ~300MB of node_modules files which will increase CI time.
Still thinking of how to approach it

willarmiros:
What's strange that this seems to be an issue with lerna hoisted modules, but it only fails with Node 16. Do we know why that is? 

Also, regarding the fix for adding mocha to the `--no-hoist` list, I wonder what the behavior before hoisting was with mocha, and whether the added CI time is still much better than before and we could just live with it for now.

Alternatively, is there any way we could force the test-utils package to be hoisted? Perhaps adding it as a devDependency to the root package.json so it's always included with those root-level dependencies?

blumamir:
> What's strange that this seems to be an issue with lerna hoisted modules, but it only fails with Node 16. Do we know why that is?

Not actually. Tried to run it locally on node 14 and it also failed locally, so the mystery is even greater :/

> 
> Also, regarding the fix for adding mocha to the `--no-hoist` list, I wonder what the behavior before hoisting was with mocha, and whether the added CI time is still much better than before and we could just live with it for now.

Good question, I don't have the number on CI time, but in term of size, mocha is using 300 MB of node_modules out of the 1GB for non-hoisting setup, and hoisting reduced about 10 minutes from the CI time, so guess it will add about 30% of that back, about 3 minutes per run. If I don't find any other solution, we can consider that.

> Alternatively, is there any way we could force the test-utils package to be hoisted? Perhaps adding it as a devDependency to the root package.json so it's always included with those root-level dependencies?

That's a great idea. On ext-js repo, we use lerna with yarn workspaces which hoist the intra-repo dependency to the root node_modules automatically as you suggested. 
Unfortunately, npm fails when it tries to fetch the (private) package from the registry before lerna has the chance to link it properly. I remember [there was a discussion on publishing it to npm](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/595). Believe it can solve the issue. @dyladan do you support publishing test-utils to npm?

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
```diff
@@ -204,19 +167,17 @@ describe('MongoDBInstrumentation', () => {
             assert(secondElement !== null);
             // assert that we correctly got the first as a find
             assertSpans(
-              memoryExporter
-                .getFinishedSpans()
-                .filter(
-                  span => span.name.includes('mongodb.getMore') === false
-                ),
+              getTestSpans().filter(
+                span => span.name.includes('mongodb.getMore') === false
```
willarmiros, line 171:
```suggestion
                (span: ReadableSpan) => span.name.includes('mongodb.getMore') === false
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
```diff
@@ -204,19 +167,17 @@ describe('MongoDBInstrumentation', () => {
...
               'mongodb.find',
               SpanKind.CLIENT
             );
             // assert that we correctly got the first as a find
             assertSpans(
-              memoryExporter
-                .getFinishedSpans()
-                .filter(span => span.name.includes('mongodb.find') === false),
+              getTestSpans().filter(
+                span => span.name.includes('mongodb.find') === false
```
willarmiros, line 179:
```suggestion
                (span: ReadableSpan) => span.name.includes('mongodb.find') === false
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
```diff
@@ -256,18 +209,17 @@ describe('MongoDBInstrumentation', () => {
...
-      const span = provider.getTracer('default').startSpan('insertRootSpan');
+      const span = trace.getTracer('default').startSpan('insertRootSpan');
       context.with(trace.setSpan(context.active(), span), () => {
         collection.insertOne(object).then(() => {
           span.end();
-          const spans = memoryExporter.getFinishedSpans();
+          const spans = getTestSpans();
           const operationName = 'mongodb.insert';
           assertSpans(spans, operationName, SpanKind.CLIENT, false, false);
           const mongoSpan = spans.find(s => s.name === operationName);
```
willarmiros, line 225:
```suggestion
          const mongoSpan = spans.find((s: ReadableSpan) => s.name === operationName);
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
```diff
@@ -297,11 +248,11 @@ describe('MongoDBInstrumentation', () => {
...
-        const span = provider.getTracer('default').startSpan('insertRootSpan');
+        const span = trace.getTracer('default').startSpan('insertRootSpan');
         context.with(trace.setSpan(context.active(), span), () => {
           collection.insertOne(object).then(() => {
             span.end();
-            const spans = memoryExporter.getFinishedSpans();
+            const spans = getTestSpans();
             const operationName = 'mongodb.insert';
             assertSpans(spans, operationName, SpanKind.CLIENT, false, true);
             const mongoSpan = spans.find(s => s.name === operationName);
```
willarmiros, line 258:
```suggestion
            const mongoSpan = spans.find((s: ReadableSpan) => s.name === operationName);
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
```diff
@@ -17,19 +17,18 @@
...
-  SimpleSpanProcessor,
-} from '@opentelemetry/sdk-trace-base';
 import * as assert from 'assert';
 import { MongoDBInstrumentation, MongoDBInstrumentationConfig } from '../src';
 import { MongoResponseHookInformation } from '../src/types';
+import {
+  registerInstrumentationTesting,
+  getTestSpans,
+  resetMemoryExporter,
+} from '@opentelemetry/contrib-test-utils';
```
willarmiros, line 27:
```suggestion
} from '@opentelemetry/contrib-test-utils';
import { ReadableSpan } from '@opentelemetry/sdk-trace-base';
```

willarmiros (commented):
Well, this took me far longer than I care to admit, but I found the root cause for the `moduleNotFound` errors to be that you need to correct the `main` and `types` entrypoint in the test-utils `package.json`:

```
  "main": "build/src/index.js",
```

If you'd like, for further consistency with other packages, you can generate a `version.ts` at build-time with the following modifications to test-utils `scripts`. But after release-please is added I don't think this will be necessary:

```
    "version:update": "node ../../../scripts/version-update.js",
    "compile": "npm run version:update && tsc -p .",
```

Also added some other suggestions to resolve the easier compiler errors. @blumamir feel free to address these when you're able!

blumamir:
Fixed the issues, now tests are green.
Had to include `mocha` in the `--nohoist` list for it to work correctly.

If there are no additional comments after changes, this one is ready to get merged finally :)

willarmiros (approved):
LGTM! Ready to merge whenever possible @obecny @vmarchaud @dyladan

PR #627, opened 2021-08-18
fix(user-interaction): EventTarget is undefined in IE

t2t2:

#537 changed the patched interface from HTMLElement to EventTarget. Problem is EventTarget doesn't exist in Internet Explorer (11)

image

Instead addEventListener & removeEventListener are provided on interface closest to where EventTarget would be on prototype chain:

image

(for comparison same code in chrome confirming EventTarget on modern browsers)

image

Added check if browser has EventTarget, using a list of alternative targets if not.

There is a small difference between IE and modern browsers as EventTarget is used by more interfaces, this difference could be removed by [using a list of targets as zone.js uses](https://github.com/angular/angular/blob/3db35b360ab23fd046f1eed2234ad0de112a66c3/packages/zone.js/lib/browser/event-target-legacy.ts#L15) but I'm not seeing any of those being a target of user interactions so don't think it's worth it for now.

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
@@ -37,6 +37,20 @@ const ZONE_CONTEXT_KEY = 'OT_ZONE_CONTEXT';
 const EVENT_NAVIGATION_NAME = 'Navigation:';
 
 /**
+ * Most browser provide event listener api via EventTarget in prototype chain.
+ * Exception to this is IE 11 which has it on the prototypes closest to EventTarget:
+ *
+ * * - has addEventListener in IE
+ * ** - has addEventListener in all other browsers
+ * ! - missing in IE
+ *
+ * HTMLElement -> Element -> Node * -> EventTarget **! -> Object
+ * Document -> Node * -> EventTarget **! -> Object
+ * Window * -> WindowProperties ! -> EventTarget **! -> Object
+ */
+const EVENT_TARGETS = window.EventTarget ? [EventTarget.prototype] : [Node.prototype, Window.prototype];
+
+/**
  * This class represents a UserInteraction plugin for auto instrumentation.
  * If zone.js is available then it patches the zone otherwise it patches
  * addEventListener of HTMLElement
@@ -571,26 +585,28 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
       );
     } else {
       this._zonePatched = false;
-      if (isWrapped(EventTarget.prototype.addEventListener)) {
-        this._unwrap(EventTarget.prototype, 'addEventListener');
-        api.diag.debug('removing previous patch from method addEventListener');
-      }
-      if (isWrapped(EventTarget.prototype.removeEventListener)) {
-        this._unwrap(EventTarget.prototype, 'removeEventListener');
-        api.diag.debug(
-          'removing previous patch from method removeEventListener'
+      EVENT_TARGETS.forEach(target => {
+        if (isWrapped(target.addEventListener)) {
+          this._unwrap(target, 'addEventListener');
+          api.diag.debug('removing previous patch from method addEventListener');
+        }
+        if (isWrapped(target.removeEventListener)) {
+          this._unwrap(target, 'removeEventListener');
+          api.diag.debug(
+            'removing previous patch from method removeEventListener'
+          );
+        }
+        this._wrap(
+          target,
+          'addEventListener',
+          this._patchAddEventListener()
         );
-      }
-      this._wrap(
-        EventTarget.prototype,
-        'addEventListener',
-        this._patchAddEventListener()
-      );
-      this._wrap(
-        EventTarget.prototype,
-        'removeEventListener',
-        this._patchRemoveEventListener()
-      );
+        this._wrap(
+          target,
+          'removeEventListener',
+          this._patchRemoveEventListener()
+        );
+      })
     }
 
     this._patchHistoryApi();
@@ -619,12 +635,14 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
         this._unwrap(ZoneWithPrototype.prototype, 'cancelTask');
       }
     } else {
-      if (isWrapped(HTMLElement.prototype.addEventListener)) {
-        this._unwrap(HTMLElement.prototype, 'addEventListener');
-      }
-      if (isWrapped(HTMLElement.prototype.removeEventListener)) {
-        this._unwrap(HTMLElement.prototype, 'removeEventListener');
-      }
+      EVENT_TARGETS.forEach(target => {
+        if (isWrapped(target.addEventListener)) {
+          this._unwrap(target, 'addEventListener');
+        }
+        if (isWrapped(target.removeEventListener)) {
+          this._unwrap(target, 'removeEventListener');
+        }
+      });
     }
     this._unpatchHistoryApi();
   }


plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -36,6 +36,22 @@ import { VERSION } from './version';
...
+ * ** - has addEventListener in all other browsers
+ * ! - missing in IE
+ *
+ * HTMLElement -> Element -> Node * -> EventTarget **! -> Object
+ * Document -> Node * -> EventTarget **! -> Object
+ * Window * -> WindowProperties ! -> EventTarget **! -> Object
+ */
+const EVENT_TARGETS = window.EventTarget
+  ? [EventTarget.prototype]
+  : [Node.prototype, Window.prototype];
```
obecny, line 53: pls correct me if I'm wrong but wouldn't this produce a double events on IE ?

obecny (commented):
this is really nice finding, one question and I'm just missing some unit tests

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
t2t2: It wouldn't, much like pre-#537 targeted addEventListener calls to objects that had HTMLElement in it's prototype chain (HTMLAnchorElement -> **HTMLElement** -> Element -> Node -> EventTarget (if not IE) -> Object), anything in a different prototype chain that didn't have patched HTMLElement's addEventListener would not be affected (so document.addEventListener didn't produce spans as it's chain is Document -> Node -> EventTarget -> Object).

This behavior can actually be confirmed in any browser: https://jsfiddle.net/t2t2/cpw8svj4/

Same code ran in IE: (ofc jsfiddle is broken in IE)

image

(So in summary in IE Node.prototype has a reference to native addEventListener and Window.prototype has a reference to native addEventListener, changing one reference doesn't affect the other)

t2t2:
Don't think this can be tested with just adding an unit test but rather this requires a full cross-browser testing setup (as it's dependent on the browser / running environment). In fact the current tests, only ran in IE, would catch this issue as they fail due to non-existant EventTarget

johnbley (approved):

dyladan:
@obecny was working on multi-browser testing but i'm not sure what the current state is on that

dyladan (approved):

dyladan:
I'd like to wait for @obecny approval since he wrote this plugin

obecny (commented):
> Don't think this can be tested with just adding an unit test but rather this requires a full cross-browser testing setup (as it's dependent on the browser / running environment). In fact the current tests, only ran in IE, would catch this issue as they fail due to non-existant EventTarget

you can simply do a unit tests with mocking window object so that you can simulate this and then make sure it was called.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -571,26 +587,26 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
       this._zonePatched = false;
-      if (isWrapped(EventTarget.prototype.addEventListener)) {
-        this._unwrap(EventTarget.prototype, 'addEventListener');
-        api.diag.debug('removing previous patch from method addEventListener');
-      }
-      if (isWrapped(EventTarget.prototype.removeEventListener)) {
-        this._unwrap(EventTarget.prototype, 'removeEventListener');
-        api.diag.debug(
-          'removing previous patch from method removeEventListener'
+      EVENT_TARGETS.forEach(target => {
```
MSNev, line 590: IE doesn't have addEventListener() rather we should be using attachEvent(both of which hang off window (for IE)).

I think it would be better to check for the existence of "attachEvent" and if so wrap as before otherwise wrap addEventListener.

MSNev: Above is assuming that we want to support IE <= 8

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -36,6 +36,22 @@ import { VERSION } from './version';
...
+ *
+ * * - has addEventListener in IE
+ * ** - has addEventListener in all other browsers
+ * ! - missing in IE
+ *
+ * HTMLElement -> Element -> Node * -> EventTarget **! -> Object
+ * Document -> Node * -> EventTarget **! -> Object
+ * Window * -> WindowProperties ! -> EventTarget **! -> Object
+ */
+const EVENT_TARGETS = window.EventTarget
```
MSNev, line 51: As ```isWrapped``` handles detection of whether a function has already been wrapped, why not just always list the entry points that we want to wrap, rather than the split (which really would require tests to be run against IE to ensure this doesn't get broken).

MSNev: This also assumes (which is generally true), that the root object names "Node", "Window" "EventTarget" always exist in all runtimes (which they *may* not).

Would suggest that we have an additional "wrap" function that can take the ```target``` and ```funcName``` so that it can do the following and wrap the correct level.
- Check if ```target``` is not null/undefined (if true -- nothing to wrap => abort wrap)
- Check if ```target[funcName]``` is defined (effectively check for instance level override) (if true wrap instance function (for wrapping simple polyfill's); if false check next)
- Get ```prototype``` of ```target``` (this can be tricky for all environments) 
- Check if ```targetPrototype[funcName]``` exists (if true wrap; otherwise abort wrap)

Examples of getting the correct target are in the ApplicationInsights code I wrote, this is more than what OT needs as it supports a larger set of runtimes (ie. all browsers (IE7+), Node, React, Angular, ReactNative and variants etc.) https://github.com/microsoft/ApplicationInsights-JS/blob/master/shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts#L152-L181

MSNev: Just tracked through the ```this._wrap``` which is really the ```shimmer``` project dependency, it does almost all of the above except for the ```prototype``` lookup if it's not defined on the target. It also has a dependency on ```Object.defineProperty``` (which I assume is not the only place in the IT code 😄 ) so my previous comment about IE <= 8 is mute as this just won't work anyway.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
obecny: I don't think we need to worry about anything below ie11 tbh.

Rauno56 (approved):

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
obecny: ping @MSNev was your concerns resolved ?

MSNev: > Think a browser runtime that doesn't support DOM spec essentials like those has bigger issues than opentelemetry support

It's not the browser runtimes you need to be aware of as they generally have them all, it's other native runtimes that "support" JavaScript... 

> maybe this and IE 11 can be thrown away.

I'm still waiting for IE7/8 to be thrown away and as part of the ApplicationInsights work we have to make sure it keeps working (mostly for the health industry) as they tend to lag and still have older versions of Windows and IE. Additionally, when you run IE 11 as an "embedded" browser (create a C# app and drop a browser component on it), by default it runs in IE7 mode.

I haven't checked if you have created separate tests (I suspect not, based on comments and the runtime), so this would mean that this comment is still valid

> why not just always list the entry points that we want to wrap, rather than the split (which really would require tests to be run against IE to ensure this doesn't get broken).

So this really means that we are adding a code path this is never tested (the primary concern)

MSNev (approved):

PR #628, opened 2021-08-18
Build: Add test-all-versions scripts to redis & ioredis

YanivD:
This will let us run `test-all-versions` scripts for `redis` and `ioredis` plugins, and make sure tests are green for all versions supported by the instrumentations.

Currently these are just scripts in package.json which can be run manually on demand by user to verify compatibility. This is not integrated into CI or expected to impact project in any way.

After running locally I found that for `redis` all green. ✅ 

When running it on `ioredis` I found failed tests on `v2.x.x`, `v3.x.x`, it failed because those versions use BlueBird which breaks the trace context. I fixed that by configuring the instrumentation with `requireParentSpan=false`.
I found another issue with `v4.19.0`, script is keep running and never ends, probably caused by [this issue](https://github.com/luin/ioredis/pull/1219). So I excluded `v4.19.0` from TAV.


blumamir (approved):
LGTM

Me and Yaniv work together, and I helped in some parts of this PR, so will need more reviewers on this one.

Rauno56 (approved):
Well aware that it has nothing to do with TAV, but could you perhaps remove `console.log` from `ioredis` tests or replace them with meaningful asserts?

It took me few tries but worked fine when I finally installed the packages in the project root.

blumamir:
> Well aware that it has nothing to do with TAV, but could you perhaps remove `console.log` from `ioredis` tests or replace them with meaningful asserts?

Agree. I think it might be better to address that in a separate PR
WDYT?

> It took me few tries but worked fine when I finally installed the packages in the project root.

which packages were missing and how did it behave?

YanivD:
Created separated PR for removing `console.log` https://github.com/open-telemetry/opentelemetry-js-contrib/pull/632

Rauno56:
> Agree. I think it might be better to address that in a separate PR
> WDYT?

Saw the PR already, thanks!

> which packages were missing and how did it behave?

I don't remember exactly anymore, but it doesn't matter either.

YanivD:
Hi, do I need more approvers or is it ready to be merged?

package.json
```diff
@@ -26,6 +26,7 @@
     "codecov:ci:changed": "lerna run codecov --since origin/main --parallel",
     "codecov:browser": "lerna run codecov:browser",
     "changelog": "lerna-changelog",
+    "lerna:link": "lerna link",
```
obecny, line 29: Nit: is this really needed ?

obecny (approved):
lgtm, @YanivD  @blumamir is now owner of the redis, can decide about merging too

package.json
blumamir: yes, used [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/628/files#diff-d67e2325d9922fc89305ff873c2473de9d6f739f7e3ec5aa9a029615d929c6c5R6)

PR #632, opened 2021-08-20
test(ioredis): remove console.log

YanivD:
Removed unnecessary `console.log` from `ioredis` tests


Flarna (approved):

blumamir (approved):

plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
```diff
@@ -336,16 +336,12 @@ describe('ioredis', () => {
         context.with(trace.setSpan(context.active(), span), () => {
           const stream = client.scanStream();
           stream
-            .on('data', resultKeys => {
-              // `resultKeys` is an array of strings representing key names.
+            .on('data', _resultKeys => {
```
Rauno56, line 339: Is there a reason to keep the `data` listener at all?

Flarna: registering `data` listener puts the stream into flowing mode. An alternative would be to call `read` and consume the data or call `stream.resume()`.

Rauno56: Right. I'd prefer we use that instead of the empty listener actually. Other than that :+1:

Rauno56 (approved):

plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
YanivD: @Rauno56 
Pushed another commit, now it's using `stream.resume()` instead of the empty listener

dyladan:
Looks fine to me but it looks like it's doing a lot more than it says in the title (adds test all versions script)

dyladan:
@blumamir if you're ok with that feel free to merge this

YanivD:
@dyladan 
I accidentally based on TAV branch when creating this change.
If it's ok with you, please merge https://github.com/open-telemetry/opentelemetry-js-contrib/pull/628 first and than this PR.

blumamir:
> @dyladan
> I accidentally based on TAV branch when creating this change.
> If it's ok with you, please merge #628 first and then this PR.

After merging the [TAV PR](#628), this PR is only handling the console.log issue.
I'll give it few more hours and merge it.

PR #634, opened 2021-08-20
docs: Update README.md

Gkolman:
added https://reporoster.com/ contributor acknowledgment/shout out

-

- providing acknowledgment to open source contributors


obecny:
Tbh I don't think that providing this information to readme will give us any benefit as this info is already available through buttons on top right corner anyway.

blumamir:
> Tbh I don't think that providing this information to readme will give us any benefit as this info is already available through buttons on top right corner anyway.

Agree

dyladan:
I also don't really want to link out to an external site for functionality already easily available on github. Based on @obecny comment and @vmarchaud thumbs up on it, I think maintainers are united on this and I'm closing this as won't do.

PR #635, opened 2021-08-23
chore: update core dependencies to 0.25

dyladan:
What it says on the tin


Flarna (approved):

dyladan:
Anyone know why this is failing? i'm not really sure

Flarna:
Best guess is too high load because of parallel testing. I see no other reason for e.g. `"before all" hook in "WinstonInstrumentation"` not finishing within 2000ms.

dyladan:
> Best guess is too high load because of parallel testing. I see no other reason for e.g. `"before all" hook in "WinstonInstrumentation"` not finishing within 2000ms.

I agree. I can't reproduce it locally either.

dyladan:
Trying to limit concurrency to 4 (there are 2 cpus on the runners) to see if that helps. Probably we haven't seen this yet because we haven't had an update which touches so many packages.

dyladan:
Seems like the lambda failure is a legitimate failure

dyladan:
@willarmiros the AWS lambda tests are consistently failing and they pass locally. Any idea what might be up?

willarmiros:
Taking a look...seems like the problem is that the spans aren't getting exported. Were there any changes to the memory exporter or force flush?

willarmiros:
I can see no relevant changes were made, and that wouldn't really make sense since the tests are passing locally anyway. I can see that there many other tests that are using the in-memory exporter and then verifying the content of those exported spans. One difference is that we're using the BatchSpanProcessor and all others that I've seen are using the SimpleSpanProcessor, so maybe there's something to that

dyladan:
> Taking a look...seems like the problem is that the spans aren't getting exported. Were there any changes to the memory exporter or force flush?

Not that I'm aware of

dyladan:
Quick perusal of the release notes from 0.25 shows only this that might have made a difference?

https://github.com/open-telemetry/opentelemetry-js/pull/2396

willarmiros:
Yeah I saw that, don't think that's the culprit because it should have equally impacted the tests that use SimpleSpanProcessor... it could be something weird with the GitHub runner environment, just really not sure why we'd only see it with these tests on this particular PR.

willarmiros:
@anuraaga do you have any ideas?

anuraaga:
@willarmiros Since we don't have package-lock.json is it possible the dependencies aren't at the same version when run locally as what's running in CI? That's what comes to mind as a possible reason for not being able to reproduce.

We have this that relies on some weird behavior of the patching infrastructure but wonder if an update changed it.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L109

PR #639, opened 2021-08-25
build(test-utils): marking test-utils as non private so it can be published

blumamir:
Follows https://github.com/open-telemetry/opentelemetry-js-contrib/pull/596

- #595 
- #621 
To my knowledge, there was no objection for it to be published to npm.
We can discuss here or on the original issue if this is not ok for any reason

- mark test-utils as non private (so it can be published to npm)
- add missing LICENCE file
- add "files" section to package.json to state which files should be included in a release
- rename `test-utils` to `contrib-test-utils`


blumamir:
> LGTM. @dyladan hopefully this can go out in contrib release 0.25.0 so we can unblock the ext-js migrations ASAP!

@dyladan Fixed the conflicts and now tests are green

vmarchaud (approved):

PR #640, opened 2021-08-25
chore(mongodb): adding TAV script

YanivD:
This will let us run `test-all-versions` scripts for `mongodb` plugin, and make sure tests are green for all versions supported by the instrumentations.

Currently this is just a script in package.json which can be run manually on demand by user to verify compatibility. This is not integrated into CI or expected to impact project in any way.


YanivD:
The main purpose of TAV script is making it easy to validate tests are green for every supported version of instrumented library. 

Currently, it will let component authors to run it locally and verify nothing breaks. 

I'm separating this change to multiple pieces to get confidence with TAV and have smaller PR which is easier to review. Eventually, I think it will be great to run TAV before publishing new version and on a daily basis.

blumamir (approved):

vmarchaud (approved):

PR #641, opened 2021-08-25
chore(mysql2): adding TAV script

YanivD:
This will let us run `test-all-versions` scripts for `mysql2` plugin, and make sure tests are green for all versions supported by the instrumentations.

Currently this is just a script in package.json which can be run manually on demand by user to verify compatibility. This is not integrated into CI or expected to impact project in any way.


YanivD:
I'm putting this PR in draft mode, I found some issues for `<2.1.0` versions.
I'm investigating the issue

YanivD:
After fixing broken tests for older versions (v2.0.0, v2.0.1, v2.0.2, v2.1.0) 
PR is ready for review

dyladan (approved):

dyladan:
Might be worth making the fixes in a separate `fix:` PR so it can be properly detected as a bug by tooling

YanivD:
> Might be worth making the fixes in a separate `fix:` PR so it can be properly detected as a bug by tooling

The fix is relevant only for the tests, it doesn't change the instrumentation code.

Rauno56 (approved):

PR #646, opened 2021-08-26
feat(koa): add a config option to allow layers to be ignored by type

agustindaguerre:

- Adds `ignoreLayersType` option to instrumentation constructor config to allow layers to be ignored (as in express instrumentation)

- Defines `KoaInstrumentationConfig` with `ignoreLayersType` which is an array of `KoaLayerType`
- Utils module with `isLayerIgnored` to check if a layer should be ignored
- Modifies `_patchLayer` to check if the layer is ignored by the config


vmarchaud:
@agustindaguerre Could you sign the CLA please ?

plugins/node/opentelemetry-instrumentation-koa/README.md
```diff
@@ -41,6 +41,19 @@ registerInstrumentations({
 
 See [examples/koa](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/koa) for a short example using both Koa and @koa/router
 
+### Koa Instrumentation Options
+
+Koa instrumentation has few options available to choose from. You can set the following:
```
dyladan, line 46: I think this line isn't really saying anything. It can just be removed
```suggestion
```

plugins/node/opentelemetry-instrumentation-koa/README.md
```diff
@@ -41,6 +41,19 @@ registerInstrumentations({
...
+
+| Options | Type | Example | Description |
+| ------- | ---- | ------- | ----------- |
+| `ignoreLayersType`| `KoaLayerType[]` | `['middleware']` | Ignore layers of specified type. |
+
+`ignoreLayersType` accepts an array of following strings:
+
+- `router`,
+- `middleware`.
```
dyladan, line 55: `KoaLayerType` isn't really a useful type without explaining what it is. Either explain the type, or just say it is an array of string and list the possible values. Will koa users know what router and middleware layer types are or does that also need explanation?

plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
```diff
@@ -126,7 +126,13 @@ export class KoaInstrumentation extends InstrumentationBase<typeof koa> {
...
     layerPath?: string
   ): KoaMiddleware {
-    if (middlewareLayer[kLayerPatched] === true) return middlewareLayer;
+    const layerType = isRouter ? KoaLayerType.ROUTER : KoaLayerType.MIDDLEWARE;
+    // Skip patching layer if its ignored in the config
+    if (
+      middlewareLayer[kLayerPatched] === true ||
+      isLayerIgnored(layerType, this._config)
+    )
+      return middlewareLayer;
```
dyladan, line 135: With this if-statement on multiple lines I would prefer to use curly braces

```suggestion
    if (
      middlewareLayer[kLayerPatched] === true ||
      isLayerIgnored(layerType, this._config)
    ) {
      return middlewareLayer;
    }
```

plugins/node/opentelemetry-instrumentation-koa/README.md
agustindaguerre: @dyladan thanks for the feedback. Updated the README to be more specific around what the values of `KoaLayerType` are in reference of.

agustindaguerre:
@vmarchaud waiting on a check to sign the corporate CLA, I'll get authorized as soon that's done

plugins/node/opentelemetry-instrumentation-koa/test/utils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      assert.doesNotThrow(() =>
+        utils.isLayerIgnored(KoaLayerType.MIDDLEWARE, {
+          ignoreLayersType: {},
+        } as KoaInstrumentationConfig)
+      );
+      assert.doesNotThrow(() =>
+        utils.isLayerIgnored(KoaLayerType.MIDDLEWARE, {
+          ignoreLayersType: {},
+        } as KoaInstrumentationConfig)
+      );
```
Rauno56, line 40: The same thing seems to be included twice here.

plugins/node/opentelemetry-instrumentation-koa/test/utils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import * as utils from '../src/utils';
+import * as assert from 'assert';
+import { KoaInstrumentationConfig, KoaLayerType } from '../src/types';
+
+describe('Utils', () => {
+  describe('isLayerIgnored()', () => {
+    it('should not fail with invalid config', () => {
+      assert.doesNotThrow(() => utils.isLayerIgnored(KoaLayerType.MIDDLEWARE));
```
Rauno56, line 24: I understand the semantics here, but `assert.doesNotThrow` seems inaccurate here. If the function throws, the test would catch it anyway. It would be more expressive to just test the output to be `false` for the cases of invalid config because that's also included in the expectation for the function, isn't it?

agustindaguerre:
@Rauno56 updated and fixed linting errors 👍

Rauno56:
Fantastic, thanks!
I've approved the PR from my end.
Usually, we require at least one more approval and any discussions(@dyladan) to be settled before merging.

agustindaguerre:
@vmarchaud @dyladan rebased again :)

Rauno56:
@agustindaguerre, , no need to re-request the reviews after rebases - it removes the given reviews already given.
To make this process smoother the next time it helps if you enable "Allow edits and access to secrets by maintainers" - that means we can rebase it ourselves as well.

PR #652, opened 2021-09-02
Update express example to include route filtering

### LINKED ISSUE #651: ExpressJS instrumentation doesn't play nice with Sampler
I want to exclude certain routes from the ExpressJS instrumentation. However, this is made extra hard by the `updateName` usage in the instrumentation. If I look at the spans in our visualizer (Cloud Trace) the name of the span I want to exclude is "GET /health" (the Kubernetes health check), however, in the sampler the name of the span is still the name that the HTTP library instrumentation gives the span.

I'd like to see some possibility to exclude certain routes from being traced completely, however this does not seem possible now.

`updateName` culprit:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5fb3313b60a8d74af9ba8bacd260818545e4a1db/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L205-L207

0.24 of core and contrib packages.

Node 14

### What did you do?
```typescript
export class FilterSampler implements Sampler {
  constructor(private parent: Sampler, private filter: (spanName: string) => boolean) {}
  public shouldSample(ctx: Context, tid: string, name: string, kind: SpanKind, attr: SpanAttributes, links: Link[]) {
    if (!this.filter(name)) {
      process.stderr.write(`Excluding span with name "${name}"\n`); // debug
      return { decision: SamplingDecision.NOT_RECORD };
    }
    process.stderr.write(`Sampling span with name "${name}"\n`); // debug
    return this.parent.shouldSample(ctx, tid, name, kind, attr, links);
  }
  public toString(): string {
    return `FilterSampler(${this.parent.toString()})`;
  }
}

const filterFn = (spanName: string) => spanName.indexOf("GET /health") < 0 && spanName.indexOf("GET /") < 0;
const sampler = new FilterSampler(new AlwaysOnSampler(), filterFn);
const provider = new NodeTracerProvider({ sampler });
provider.register();
trace.setGlobalTracerProvider(provider);
registerInstrumentations({
  instrumentations: [
    new HttpInstrumentation(),
    new ExpressInstrumentation(),
  ],
});
```

```
Excluding span with name "GET /health"
Sampling span with name "middleware - query"
Sampling span with name "middleware - expressInit"
Sampling span with name "request handler - /health"
```

### What did you see instead?
```
Sampling span with name "HTTP GET"
Sampling span with name "middleware - query"
Sampling span with name "middleware - expressInit"
Sampling span with name "request handler - /health"
```

hermanbanken:
Fixes #651 where documentation was missing on how to exclude certain routes from ExpressJS instrumentation. This was not trivial due to the dependency on HTTP instrumentation. 

By having this example I hope to help people, so they don't have to figure this out themselves.

### DIFF
--- a/examples/express/tracer.js
+++ b/examples/express/tracer.js
@@ -6,13 +6,14 @@ const opentelemetry = require('@opentelemetry/api');
 const { diag, DiagConsoleLogger, DiagLogLevel } = opentelemetry;
 diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.INFO);
 
+const { AlwaysOnSampler } = require('@opentelemetry/core');
 const { registerInstrumentations } = require('@opentelemetry/instrumentation');
 const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
 const { SimpleSpanProcessor } = require('@opentelemetry/sdk-trace-base');
 const { JaegerExporter } = require('@opentelemetry/exporter-jaeger');
 const { ZipkinExporter } = require('@opentelemetry/exporter-zipkin');
 const { Resource } = require('@opentelemetry/resources');
-const { SemanticResourceAttributes: ResourceAttributesSC } = require('@opentelemetry/semantic-conventions');
+const { SemanticAttributes, SemanticResourceAttributes: ResourceAttributesSC } = require('@opentelemetry/semantic-conventions');
 
 const Exporter = (process.env.EXPORTER || '')
   .toLowerCase().startsWith('z') ? ZipkinExporter : JaegerExporter;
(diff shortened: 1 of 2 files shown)


Flarna (approved):

MSNev (approved):

examples/express/tracer.js
```diff
@@ -45,3 +47,21 @@ module.exports = (serviceName) => {
...
+      }
+      return parent.shouldSample(ctx, tid, name, kind, attr, links);
+    },
+    toString() {
+      return `FilterSampler(${parent.toString()})`;
+    }
+  }
+}
+
+function ignoreHealthCheck(spanName, spanKind, attributes) {
```
vmarchaud, line 65: you could also use https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-instrumentation-http/src/types.ts#L87 but this solution also works :)

vmarchaud (approved):

PR #653, opened 2021-09-02
feat(user-interaction): support for custom events and span enhancement

kkruk-sumo:

1. Only `click` events are instrumented right now. After this change, developers will have possibility to instrument any existing HTML events like `dblclick`, `touchstart`, `play` for whatever reason.
2. Created spans only have `target_element` and `target_xpath` attributes, but different websites often use custom attributes to describe elements. This change provides a callback `onSpan` which can be used to enhance created span with custom attributes, events, or even change span name. `target_xpath` is good for debugging, but when you are quickly checking out different spans, you need some more readable form of knowing which element has been clicked. Developers can also add `textContent` of clicked buttons if they want.

Added config for the `UserInteractionInstrumentation` class with two attributes `eventTypes` and `onSpan`.


plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -133,6 +140,8 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
           : undefined
       );
 
+      this._onSpan?.(eventName, element, span);
```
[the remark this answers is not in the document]
kkruk-sumo: That's a good idea! Added :)

plugins/web/opentelemetry-instrumentation-user-interaction/src/types.ts
```diff
@@ -14,7 +14,28 @@
...
+   * By default only 'click' event is instrumented.
+   */
+  eventTypes?: string[];
+
+  /**
+   * Callback function called each time new span is being created.
+   * You can use this handler to enhance created span with extra attributes.
+   * Return `false` to prevent span recording.
+   */
+  onSpan?: (
```
obecny, line 33: onSpan sounds more like an callback event approach which should return void type. I would name it like `shouldPreventSpanCreation` this way it sounds more intuitive and you can avoid adding `?` everywhere in a code by refactoring the piece in constructor - see my comment there

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -57,9 +55,13 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
     api.Span
   >();
+  private _eventTypes: Set<string>;
+  private _onSpan: UserInteractionInstrumentationConfig['onSpan'];
 
-  constructor(config?: InstrumentationConfig) {
+  constructor(config?: UserInteractionInstrumentationConfig) {
     super('@opentelemetry/instrumentation-user-interaction', VERSION, config);
+    this._eventTypes = new Set(config?.eventTypes ?? DEFAULT_EVENT_TYPES);
+    this._onSpan = config?.onSpan;
```
obecny, line 64:
```suggestion

// earlier
function defaultShouldPreventSpanCreation() { return false };

....


  this._ shouldPreventSpanCreation = 
    typeof config.shouldPreventSpanCreation === 'function' ? 
      config.shouldPreventSpanCreation : 
      defaultShouldPreventSpanCreation
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -546,6 +562,70 @@ describe('UserInteractionInstrumentation', () => {
...
+        eventTypes: ['play'],
+      });
+
+      fakeEventInteraction('play');
+      assert.strictEqual(exportSpy.args.length, 1, 'should export one span');
+      const span = exportSpy.args[0][0][0];
+      assertInteractionSpan(span, { name: 'play' });
+    });
+
+    it('should not export clicks if this event was not configured to be captured', () => {
```
obecny, line 585: whenever you use word "when" or "if" ettc. you should split the declaration with one extra level
```
describe('if event is not configured', ()=> {
  it('should not export clicks', ()=> {
    ...
  });
})
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -546,6 +562,70 @@ describe('UserInteractionInstrumentation', () => {
...
+
+      const element = createButton();
+      element.addEventListener('click', () => {});
+      element.click();
+
+      const span = exportSpy.args[0][0][0];
+      assert.deepStrictEqual(onSpan.args, [['click', element, span]]);
+    });
+
+    it('should not record span when onSpan returns false', () => {
```
obecny, line 612:
```
describe('when "shouldPreventSpanCreation" returns false', ()=> {
  it('should not create a span', ()=> {
    ...
  });
})
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
```diff
@@ -211,6 +230,70 @@ describe('UserInteractionInstrumentation', () => {
...
+        eventTypes: ['play'],
+      });
+
+      fakeEventInteraction('play');
+      assert.strictEqual(exportSpy.args.length, 1, 'should export one span');
+      const span = exportSpy.args[0][0][0];
+      assertInteractionSpan(span, { name: 'play' });
+    });
+
+    it('should not export clicks if this event was not configured to be captured', () => {
```
obecny, line 253: describe('when event was not configured to be captured', ()=> {
  it('should not export clicks', ()=> {
    ...
  });
})

plugins/web/opentelemetry-instrumentation-user-interaction/src/types.ts
```diff
@@ -14,7 +14,28 @@
...
+import { HrTime, Span } from '@opentelemetry/api';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+
+export interface UserInteractionInstrumentationConfig
+  extends InstrumentationConfig {
+  /**
+   * List of events to instrument (like 'mousedown', 'touchend', 'play' etc).
+   * By default only 'click' event is instrumented.
+   */
+  eventTypes?: string[];
```
obecny, line 26: if that is possible I would create a typed event here - so that you can have type validation and allows only certain events (click, double clicks etc.)

plugins/web/opentelemetry-instrumentation-user-interaction/src/types.ts
kkruk-sumo: The main purpose of this PR was to add ability to enhance spans and IMO using `shouldPreventSpanCreation` only to add some span attributes may looks unintuitive. On the other hand I agree that `onSpan` should not expect a boolean to be returned. I was inspired by a DOM's event listeners but the 'return false' thing to prevent default behaviour is not a good pattern.
What do you think about leaving `onSpan` with the initial version (without a boolean as a return value) AND adding `shouldPreventSpanCreation` in the current form?

kkruk-sumo: @obecny any thoughts on that? After some consideration, single `shouldPreventSpanCreation` will also work for me.

obecny: I'm just trying to think of different edge cases and having `onSpan` and then `shouldPreventSpanCreation` would confuse user more than it would help. I still think that having just one function might be easier for user and also for us to keep the logic simple. I just suggested to change a name a bit so that it is easier to understand and it will have correct default behaviour (reversing the logic) if you don't specify anything

kkruk-sumo: I thought that `shouldPreventSpanCreation` may not suggest that this can be also used to enhance spans, but now I think it will be fine. Thanks :)

kkruk-sumo:
@obecny I fixed the UTs. Are we good to merge?

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -546,6 +563,87 @@ describe('UserInteractionInstrumentation', () => {
...
+        eventNames: ['play'],
+      });
+
+      fakeEventInteraction('play');
+      assert.strictEqual(exportSpy.args.length, 1, 'should export one span');
+      const span = exportSpy.args[0][0][0];
+      assertInteractionSpan(span, { name: 'play' });
+    });
+
+    it('not configured spans should not be exported', () => {
```
obecny, line 586:
```suggestion
    it('should not be exported not configured spans', () => {
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -546,6 +563,87 @@ describe('UserInteractionInstrumentation', () => {
...
+      element.addEventListener('click', () => {});
+      element.click();
+
+      const span = exportSpy.args[0][0][0];
+      assert.deepStrictEqual(shouldPreventSpanCreation.args, [
+        ['click', element, span],
+      ]);
+    });
+
+    describe('when shouldPreventSpanCreation returns', () => {
```
obecny, line 615:
```suggestion
    describe('when shouldPreventSpanCreation returns true', () => {
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -546,6 +563,87 @@ describe('UserInteractionInstrumentation', () => {
...
+      it('true should not record span', () => {
```
obecny, line 616:
```suggestion
      it('should not record span', () => {
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.nozone.test.ts
```diff
@@ -546,6 +563,87 @@ describe('UserInteractionInstrumentation', () => {
...
+        element.click();
+
+        assert.strictEqual(
+          exportSpy.args.length,
+          0,
+          'should not export any spans'
+        );
+      });
+
+      it('false should record span', () => {
```
obecny, line 633:
```suggestion
  describe('when shouldPreventSpanCreation returns false', () => {
      it('should record span', () => {
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
```diff
@@ -55,6 +60,25 @@ describe('UserInteractionInstrumentation', () => {
     let dummySpanExporter: DummySpanExporter;
     let exportSpy: sinon.SinonSpy;
     let requests: sinon.SinonFakeXMLHttpRequest[] = [];
+
+    const registerInstrumentation = (
```
obecny, line 64: Nit: consider changing name as it may be confused at first look with `registerInstrumentations`

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
```diff
@@ -211,6 +231,87 @@ describe('UserInteractionInstrumentation', () => {
...
+      element.addEventListener('click', () => {});
+      element.click();
+
+      const span = exportSpy.args[0][0][0];
+      assert.deepStrictEqual(shouldPreventSpanCreation.args, [
+        ['click', element, span],
+      ]);
+    });
+
+    describe('when shouldPreventSpanCreation returns', () => {
```
obecny, line 283:
```suggestion
    describe('when shouldPreventSpanCreation returns true', () => {
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
```diff
@@ -211,6 +231,87 @@ describe('UserInteractionInstrumentation', () => {
...
+      it('true should not record span', () => {
```
obecny, line 284:
```suggestion
      it('should not record span', () => {
```

plugins/web/opentelemetry-instrumentation-user-interaction/test/userInteraction.test.ts
```diff
@@ -211,6 +231,87 @@ describe('UserInteractionInstrumentation', () => {
...
+        element.click();
+
+        assert.strictEqual(
+          exportSpy.args.length,
+          0,
+          'should not export any spans'
+        );
+      });
+
+      it('false should record span', () => {
```
obecny, line 301:
```suggestion
  describe('when shouldPreventSpanCreation returns false', () => {
      it('should record span', () => {
```

obecny (approved):
approving, just a few nitpicks, hopefully you can address them before merging  as I will be off for next 2 weeks and don't want to block it

kkruk-sumo:
@obecny I've fixed the failing UT's after merging main. Are we good to merge?

kkruk-sumo:
@obecny Looks like UTs are more stable now. Previously they failed because of timeout so I don't think it was caused by this PR.

dyladan (approved):
LGTM when the conflict is fixed

vmarchaud (approved):

PR #658, opened 2021-09-08
chore: bumped aws components to 1.0

willarmiros:

- Bumps AWS X-Ray ID generator, propagator, and resource detector packages to 1.0 since no breaking changes are anticipated.
- Will leave in draft until core SDK reaches 1.0 first.

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-aws/package.json
+++ b/detectors/node/opentelemetry-resource-detector-aws/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opentelemetry/resource-detector-aws",
-  "version": "0.24.0",
+  "version": "1.0.0",
   "description": "OpenTelemetry SDK resource detector for AWS",
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
--- a/packages/opentelemetry-id-generator-aws-xray/package.json
+++ b/packages/opentelemetry-id-generator-aws-xray/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opentelemetry/id-generator-aws-xray",
-  "version": "0.24.0",
+  "version": "1.0.0",
   "description": "AWS X-Ray ID generator for OpenTelemetry",
   "main": "build/src/index.js",
   "publishConfig": {
--- a/propagators/opentelemetry-propagator-aws-xray/package.json
+++ b/propagators/opentelemetry-propagator-aws-xray/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@opentelemetry/propagator-aws-xray",
-  "version": "0.24.0",
+  "version": "1.0.0",
   "description": "OpenTelemetry AWS Xray propagator provides context propagation for systems that are using AWS X-Ray format.",
   "main": "build/src/index.js",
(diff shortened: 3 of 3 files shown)


NathanielRN (approved):
🎉

dyladan:
This is in conflict with the current lerna.json which claims all packages to be the same version. I think this also has to wait until the release automation is updated to allow for releasing packages with independent versions.

The problems we had with release-please in core generally don't apply to this repo since modules do not depend on each other so it should be possible to use release-please here without issue. If you would like to take a crack at porting the PR I did in core to this repo it would be quite helpful.

willarmiros:
> This is in conflict with the current lerna.json which claims all packages to be the same version

Makes sense, I will leave it in draft until the release-please automation is in place and close if it's no longer necessary. I wasn't sure if release-please would be capable of automatically bumping to 1.0.0, since it's at 0.24.0 right now and even if we tagged a change as breaking, I don't know if that would bump the major or minor version.

> If you would like to take a crack at porting the PR I did in core to this repo it would be quite helpful.

Sure, we're talking about https://github.com/open-telemetry/opentelemetry-js/pull/2431 just to confirm right? I'll see if I have the bandwidth.

willarmiros:
> I wasn't sure if release-please would be capable of automatically bumping to 1.0.0

I guess this answers my question: https://github.com/open-telemetry/opentelemetry-js/blob/5fefbb274893a93033ef3786f75173c62682ade7/release-please-config.json#L4

So I think this PR will still be needed even after adding release-please.

willarmiros:
Opened #659 to add release-please to this repo.

Flarna:
Is it intended that v1.0 of these packages depends on 0.25.0 of core instead 1.0.0?

Flarna (approved):

willarmiros:
@dyladan any chance this could be merged to include in the next release?

packages/opentelemetry-id-generator-aws-xray/README.md
```diff
@@ -38,16 +38,15 @@ const tracerProvider = new NodeTracerProvider(tracerConfig);
 
 Example trace ID format: 58406520a006649127e371903a2de979
 
-A trace ID consists of two parts; the timestamp and the unique identifier.
+A trace ID consists of two parts: the timestamp and the unique identifier.
 
 #### Time Stamp
```
Rauno56, line 43: I know changing the docs wasn't the intention of the PR, but it caught my eye since you did changes in the readme file as well. :laughing: 

```suggestion
#### Timestamp
```

Rauno56: Since you were at it, I'd actually just format the whole section to be more concise:

---

A trace ID consists of two parts: 

1. **Timestamp**: The first 8 hexadecimal digits represent the time of the original request in Unix epoch time. For example, 10:00 AM December 1st, 2016 PST in epoch time is 1480615200 seconds, or 58406520 in hexadecimal digits.
2. **Unique Identifier**: The last 24 hexadecimal digits is an random identifier for the trace.

---

But that comes to style more than correctness.

PR #660, opened 2021-09-15
fix: Update aws-lambda-instrumentation to SDK v0.25.0

willarmiros:

- https://github.com/open-telemetry/opentelemetry-js-contrib/pull/635#issuecomment-904882232

- Upgrading the opentelemetry-instrumentation-aws-lambda package's core dependencies to `0.25.0`, and fixing associated breaking changes.


Flarna (approved):

willarmiros:
Converting to draft since this doesn't fix the bug (yet). Just wanted to test it with GH Actions, but am reproducing locally now.

willarmiros:
Now that #661 is merged this is ready for review :+1:

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -23,8 +23,8 @@ import { AwsLambdaInstrumentation } from '../../src';
 import {
   BatchSpanProcessor,
   InMemorySpanExporter,
-} from '@opentelemetry/tracing';
-import { NodeTracerProvider } from '@opentelemetry/node';
+} from '@opentelemetry/sdk-trace-base';
```
dyladan, line 26: This dependency is missing in package.json dependencies

dyladan (changes requested):

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
longility: This is in `devDependencies`. Basically, if the tests passes, then we are good as the dependencies are good.

dyladan: It was not, but it has been added. It was a transitive dependency of something else which is why the tests were passing but it is good practice to declare all of your dependencies.

dyladan (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
longility: Apparently, I'm blind. Thanks for sharing your perspective.

PR #661, opened 2021-09-15
fix(aws-lambda): BasicTracerProvider not force flushing

### LINKED ISSUE #647: I'm expecting node aws lambda to export spans properly
```
ih-otel-lambda-layer@1.0.0 /Users/longility/Repos/platform/ih-otel-lambda-layer
├── @opentelemetry/api@1.0.2
├── @opentelemetry/auto-instrumentations-node@0.24.0
├── @opentelemetry/exporter-collector-grpc@0.24.0
├── @opentelemetry/exporter-collector-proto@0.24.0
├── @opentelemetry/id-generator-aws-xray@0.24.0
├── @opentelemetry/instrumentation-aws-lambda@0.24.0
├── @opentelemetry/sdk-node@0.24.0
├── @tsconfig/node14@1.0.1
├── opentelemetry-instrumentation-aws-sdk@0.24.1
└── typescript@4.3.5
```

v14.17.4

### What did you do?
Create a lambda layer where I started from this: https://github.com/open-telemetry/opentelemetry-js#instantiate-tracing
```typescript
import * as opentelemetry from "@opentelemetry/sdk-node";
import { CollectorTraceExporter } from "@opentelemetry/exporter-collector-proto";
import { AwsInstrumentation } from "opentelemetry-instrumentation-aws-sdk";
import { AWSXRayIdGenerator } from "@opentelemetry/id-generator-aws-xray";
import { BatchSpanProcessor } from "@opentelemetry/tracing";
import { AwsLambdaInstrumentation } from "@opentelemetry/instrumentation-aws-lambda";
import { diag, DiagConsoleLogger } from "@opentelemetry/api";
import { getEnv } from "@opentelemetry/core";

// https://github.com/open-telemetry/opentelemetry-js-api/issues/102
diag.setLogger(new DiagConsoleLogger(), getEnv().OTEL_LOG_LEVEL);

const traceExporter = new CollectorTraceExporter();

  traceExporter,
  instrumentations: [
    new AwsLambdaInstrumentation(),
    new AwsInstrumentation(),
    getNodeAutoInstrumentations(),
  ],
});

// https://aws-otel.github.io/docs/getting-started/js-sdk/trace-manual-instr#configure-global-tracer-for-aws-x-ray
sdk.configureTracerProvider(
  {
    idGenerator: new AWSXRayIdGenerator(),
  },
  new BatchSpanProcessor(traceExporter)
);

// initialize the SDK and register with the OpenTelemetry API
// this enables the API to record telemetry
sdk
  .start()
  .then(() => console.log("Tracing initialized"))
  .catch((error) => console.log("Error initializing tracing", error));

// gracefully shut down the SDK on process exit
process.on("SIGTERM", () => {
  sdk
    .shutdown()
    .then(() => console.log("Tracing terminated"))
    .catch((error) => console.log("Error terminating tracing", error))
    .finally(() => process.exit(0));
});
```

I was expecting to see `statusCode: 200` to indicate that it got exported correctly. I saw this in the debug log when I was running a server, not lambda function. Also, I am not seeing any associated traces. I am testing against the same collector.

### What did you see instead?

Here is the actual log
```
@opentelemetry/api: Registered a global for diag v1.0.2.
--
Loading instrumentation for @opentelemetry/instrumentation-dns
Loading instrumentation for @opentelemetry/instrumentation-express
Loading instrumentation for @opentelemetry/instrumentation-graphql
Loading instrumentation for @opentelemetry/instrumentation-grpc
Loading instrumentation for @opentelemetry/instrumentation-http
Loading instrumentation for @opentelemetry/instrumentation-ioredis
Loading instrumentation for @opentelemetry/instrumentation-koa
Loading instrumentation for @opentelemetry/instrumentation-mongodb
Loading instrumentation for @opentelemetry/instrumentation-mysql
Loading instrumentation for @opentelemetry/instrumentation-pg
Loading instrumentation for @opentelemetry/instrumentation-redis
2021-08-27T21:10:45.953Z	undefined	DEBUG	@opentelemetry/instrumentation-http Applying patch for http@14.17.4
2021-08-27T21:10:45.968Z	undefined	DEBUG	@opentelemetry/instrumentation-http Applying patch for https@14.17.4
2021-08-27T21:10:46.611Z	undefined	DEBUG	Applying patch for express@4.17.1
2021-08-27T21:10:46.919Z	undefined	DEBUG	aws-sdk instrumentation: applying patch to aws-sdk v3 constructStack
2021-08-27T21:10:46.937Z	undefined	DEBUG	aws-sdk instrumentation: applying patch to aws-sdk v3 client send
2021-08-27T21:10:47.457Z	undefined	DEBUG	Applying patch for lambda handler
2021-08-27T21:10:47.457Z	undefined	DEBUG	patch handler function
2021-08-27T21:10:47.462Z	undefined	DEBUG	EnvDetector found resource. Resource {  attributes: {    'service.namespace': 'platform',    'service.name': 'playground-profile',    'service.version': '7c21b2955db42c81020bc11a693629cda5d95396'  }}
2021-08-27T21:10:47.462Z	undefined	DEBUG	ProcessDetector found resource. Resource {  attributes: {    'process.pid': 9,    'process.executable.name': '/var/lang/bin/node',    'process.command': '/var/runtime/index.js',    'process.command_line': '/var/lang/bin/node /var/runtime/index.js'  }}
2021-08-27T21:10:47.464Z	undefined	DEBUG	AwsEc2Detector failed: EC2 metadata api request timed out.
2021-08-27T21:10:47.465Z	undefined	DEBUG	GcpDetector failed: GCP Metadata unavailable.
2021-08-27T21:10:47.465Z	undefined	DEBUG	GcpDetector found resource. Resource { attributes: {} }
2021-08-27T21:10:47.465Z	undefined	TRACE	{  'service.name': 'playground-profile',  'service.namespace': 'platform',  'service.version': '7c21b2955db42c81020bc11a693629cda5d95396'}
2021-08-27T21:10:47.465Z	undefined	TRACE	{  'process.command': '/var/runtime/index.js',  'process.command_line': '/var/lang/bin/node /var/runtime/index.js',  'process.executable.name': '/var/lang/bin/node',  'process.pid': 9}
2021-08-27T21:10:47.467Z	undefined	DEBUG	@opentelemetry/api: Registered a global for trace v1.0.2.
2021-08-27T21:10:47.469Z	undefined	DEBUG	@opentelemetry/api: Registered a global for context v1.0.2.
2021-08-27T21:10:47.469Z	undefined	DEBUG	@opentelemetry/api: Registered a global for propagation v1.0.2.
2021-08-27T21:10:47.470Z	undefined	INFO	Tracing initialized
START RequestId: 08e980e0-95d4-4786-a349-244b666ede74 Version: $LATEST
END RequestId: 08e980e0-95d4-4786-a349-244b666ede74
REPORT RequestId: 08e980e0-95d4-4786-a349-244b666ede74	Duration: 397.18 ms	Billed Duration: 398 ms	Memory Size: 128 MB	Max Memory Used: 128 MB	Init Duration: 2099.94 ms	XRAY TraceId: 1-612954d5-516183ac74c9ce0079d6a7b1	SegmentId: 7b92173a4d75fd9e	Sampled: true
START RequestId: 038f80ac-612a-4605-bed7-215f046c5881 Version: $LATEST
2021-08-27T21:10:49.470Z	038f80ac-612a-4605-bed7-215f046c5881	DEBUG	@opentelemetry/instrumentation-http %s instrumentation outgoingRequest https
2021-08-27T21:10:49.529Z	038f80ac-612a-4605-bed7-215f046c5881	DEBUG	@opentelemetry/instrumentation-http http.ClientRequest return request
2021-08-27T21:10:49.810Z	038f80ac-612a-4605-bed7-215f046c5881	DEBUG	@opentelemetry/instrumentation-http outgoingRequest on response()
2021-08-27T21:10:49.850Z	038f80ac-612a-4605-bed7-215f046c5881	DEBUG	@opentelemetry/instrumentation-http outgoingRequest on end()
END RequestId: 038f80ac-612a-4605-bed7-215f046c5881
REPORT RequestId: 038f80ac-612a-4605-bed7-215f046c5881	Duration: 2067.00 ms	Billed Duration: 2067 ms	Memory Size: 128 MB	Max Memory Used: 128 MB	XRAY TraceId: 1-612954d8-28807c1012c024c619439d9b	SegmentId: 3ef94af35d3ab4f9	Sampled: true
```

### Additional context

Any ideas how to debug or understand what the issue is on why the spans are not exporting?

I was lead from here: https://github.com/open-telemetry/opentelemetry-js/issues/1295#issuecomment-907476081
/cc @willarmiros @anuraaga

longility:
When the tracer provider is set, the type is a ProxyTracerProvider that proxies to the BasicTracerProvider. This unwraps and gets the delegate to attempt to force flush.

fixes #647


plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -231,6 +232,23 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
 
+  private _getBasicTracerProvider() {
+    if(!this._tracerProvider) {
+      diag.error('Spans are not exported for the lambda function as the tracerProvider is undefined.');
+      return undefined;
+    }
+
+    let realTracerProvider = this._tracerProvider;
+
+    if (this._tracerProvider instanceof ProxyTracerProvider) {
```
Flarna, line 243: I'm not sure if the `instanceof` checks work if there are more instances of the `@opentelemetry/api` package installed.

I have following setup in mind where more instances (of the same/compatible version) are installed:
* a layer holding the AWS part of the instrumentation for AWS services
* user handler has it's own  `@opentelemetry/api` in it's dependencies as it also calls OTel APIs (and/or adds more instrumentations)

longility: @Flarna What do you suggest? For me, I believe @willarmiros may be on the same page is that I believe the happy path right now is broken and get that fix and figure out the proper fix later. This will be an incremental, low effort solution for now.

dyladan: What about `this._tracerProvider.getDelegate instanceof 'function'`?

longility: I think you meant `typeof this._tracerProvider.getDelegate === 'function'` @dyladan . I'll let y'all hash it out before I make any other changes. I'm going for the low effort, low risk solution. If desire any more, I may not have the time to help.

Flarna: I was also thinking about `typeof` function checks.

How is this instrumentation used in general? e.g. how is it instantiated?
If there is no risk that more `@opentelemetry/api` instances are present or if it is ensured from the initialization flow that the `@opentelemetry/api` instance referenced by this instrumentation is the same as that one used to create the global `TracerProvider` it's fine to use `instanceof`.

willarmiros: > If there is no risk that more @opentelemetry/api instances are present or if it is ensured from the initialization flow that the @opentelemetry/api instance referenced by this instrumentation is the same as that one used to create the global TracerProvider it's fine to use instanceof

I think since Lambda sandboxes are just that, sandboxes, we can make no guarantee of what users will have in them. So it's probably safer to go with the `typeof this._tracerProvider.getDelegate === 'function'` check that we all seem to be in agreement on. @longility could you update the PR? I'm comfortable with this solution for now since it's a pretty critical, blocking bug.

longility: Should we do function check for only `getDelegate` or should we do it for `getActiveSpanProcessor` and `forceFlush` as well?

dyladan: I would check all 3 since those methods aren't required by the API a third party sdk which doesn't have those functions could theoretically be used.

longility: I updated based on what is suggested, and I'm getting test failures to where I need more help.

```
25) lambda handler
       hooks
         responseHook
           sync - error:
     TypeError: Cannot read property 'attributes' of undefined
      at Context.<anonymous> (test/integrations/lambda-handler.test.ts:18:48)
      at processImmediate (internal/timers.js:464:21)
      at process.callbackTrampoline (internal/async_hooks.js:131:17)

  26) lambda handler
       hooks
         responseHook
           sync - error with callback:
     TypeError: Cannot read property '_spanProcessors' of undefined
      at forceFlush (node_modules/@opentelemetry/tracing/src/MultiSpanProcessor.ts:33:38)
      at AwsLambdaInstrumentation._endSpan (src/instrumentation.ts:66:4)
      at /Users/longility/Repos/longility-opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts:65:33
      at Object.safeExecuteInTheMiddle (node_modules/@opentelemetry/instrumentation/src/utils.ts:36:5)
      at /Users/longility/Repos/longility-opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts:64:15
      at async_hooks.js:312:14
      at AsyncResource.runInAsyncScope (async_hooks.js:197:9)
      at AsyncLocalStorage.run (async_hooks.js:310:35)
      at AsyncLocalStorageContextManager.with (node_modules/@opentelemetry/context-async-hooks/src/AsyncLocalStorageContextManager.ts:40:36)
      at ContextAPI.with (node_modules/@opentelemetry/api/src/api/context.ts:77:42)
      at Object.patchedHandler [as callbackerror] (src/instrumentation.ts:46:46)
      at /Users/longility/Repos/longility-opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts:18:48
      at new Promise (<anonymous>)
      at Context.<anonymous> (test/integrations/lambda-handler.test.ts:18:48)
      at processImmediate (internal/timers.js:464:21)
      at process.callbackTrampoline (internal/async_hooks.js:131:17)
```

willarmiros: Sorry for the late reply @longility, I've only just got around to debugging. It turns out we're just not passing in the `thisArg` when calling forceFlush. I'll make some suggestions on the PR to address.

Also, if you don't have the time to address this, I can roll these changes into #660 and give you a co-author :) up to you

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -231,6 +230,30 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
     this._tracerProvider = tracerProvider;
   }
 
+  private _getForceFlushFn() {
```
willarmiros, line 233:
```suggestion
  private _getActiveSpanProcessor() {
```

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -231,6 +230,30 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
+
+    if (typeof currentProvider.getDelegate === 'function') {
+      currentProvider = currentProvider.getDelegate();
+    }
+
+    if (typeof currentProvider.getActiveSpanProcessor === 'function') {
+      const activeSpanProcessor = currentProvider.getActiveSpanProcessor();
+      if (typeof activeSpanProcessor.forceFlush === 'function')
+        return activeSpanProcessor.forceFlush;
+    }
```
willarmiros, line 252:
```suggestion
    if (typeof currentProvider.getActiveSpanProcessor === 'function') {
      return currentProvider.getActiveSpanProcessor();
    }
```

Only get the span processor from this helper function.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -267,15 +290,17 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
-          () => callback(),
-          () => callback()
-        );
+
+    const forceFlush = this._getForceFlushFn();
+    if (forceFlush) {
+      forceFlush().then(
+        () => callback(),
+        () => callback()
+      );
```
willarmiros, line 299:
```suggestion
    const activeSpanProcessor = this._getActiveSpanProcessor();
    if (activeSpanProcessor && typeof activeSpanProcessor.forceFlush === 'function') {
      activeSpanProcessor.forceFlush().then(
        () => callback(),
        () => callback()
      );
```

willarmiros (commented):
@longility these comments should fix the unit tests (assuming no typos!)

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
longility: Thanks @willarmiros I committed your suggestions and that worked! I learned about suggested changes. Thanks for that. That is nice.

Also, can you elaborate on your comment as I would like to learn. The changes was not enough to understand why that worked and my changes does not.
> It turns out we're just not passing in the thisArg when calling forceFlush. I'll make some suggestions on the PR to address.

willarmiros: > Also, can you elaborate on your comment as I would like to learn. The changes was not enough to understand why that worked and my changes does not.

In `strict mode` JS, when you call a function of an object, it implicitly passes the object to the function which can be referenced as `this`. For instance, if you call `foo.bar()`, you can reference `foo` within the `bar` function via `this`.

In the original change, since we were calling `forceFlush` without a reference to its parent object, `this` was undefined within the `forceFlush` function, causing an error [here](https://github.com/open-telemetry/opentelemetry-js/blob/bebd82a7f12927bb8a486a252bb7c1cc12619294/packages/opentelemetry-sdk-trace-base/src/MultiSpanProcessor.ts#L33).

So my change allows the `this` reference to be defined correctly. Read more: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

longility: Makes sense, thanks! Alternatively, it has been a while since I had to care about `this`, but if I recall, we could explicitly `bind` to its original objects as well. Thanks for taking your time to explain a gotcha that I was unaware of @willarmiros !

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -231,6 +230,28 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
     this._tracerProvider = tracerProvider;
   }
 
+  private _getActiveSpanProcessor() {
```
dyladan, line 233: I wonder if you want to cache the unproxied tracer provider?

dyladan (approved):
Looks good to me. One minor suggestion up to you

willarmiros (approved):
Thanks @longility! It looks like there are some lint errors to fix quickly.

This will unblock #660. @Flarna please take a look when you get a chance.

longility:
> Thanks @longility! It looks like there are some lint errors to fix quickly.
> 
> This will unblock #660. @Flarna please take a look when you get a chance.

I’m not at my desk so maybe at least two hours before I can get to it

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
longility: I refactored and hopefully it is something we both agree is better. "Caches" without having to deal with multiple field states as we don't need a `_traceProvider` field, just `_forceFlush`. Since you helped me know what the problem is with `this`, I was able to just use `bind` to help with keeping nastiness in one method.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -44,7 +44,6 @@ import {
   SemanticAttributes,
   SemanticResourceAttributes,
 } from '@opentelemetry/semantic-conventions';
-import { BasicTracerProvider } from '@opentelemetry/tracing';
```
Flarna, line 47: This change allows to move `@opentelemetry/tracing` into dev-dependencies

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -228,7 +227,32 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
 
   override setTracerProvider(tracerProvider: TracerProvider) {
     super.setTracerProvider(tracerProvider);
-    this._tracerProvider = tracerProvider;
+    this._forceFlush = this._getForceFlush(tracerProvider);
+  }
+
+  private _getForceFlush(tracerProvider: TracerProvider) {
+    if (!tracerProvider) {
+      diag.error(
```
Flarna, line 235: not sure if `error` is the correct log level here because a user may decide to disable the sdk for whatever reason.
Or in other words: it's not an error of this instrumentation and nothing this instrumentation can change.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
longility: nice catch

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      );
+    });
+
+    assert.strictEqual(forceFlushed, true);
+  });
+
+  it('should force flush ProxyTracerProvider with NodeTracerProvider', async () => {
+    const nodeTracerProvider = new NodeTracerProvider();
+    nodeTracerProvider.addSpanProcessor(new BatchSpanProcessor(memoryExporter));
+    nodeTracerProvider.register();
```
Flarna, line 100: Why is the `NodeTracerProvider` registered here and not the `ProxyTracerProvider`?

Flarna (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
longility: You are right. Now that I refactored and this is happening during the setting of the tracer provider and not during the end span, this log does not make sense. Changing it to debug.

longility: Actually, I'm going to get rid of it as it doesn't make sense and hard to follow.

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
longility: `ProxyTracerProvider` does not have a method `register()`. https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/trace/ProxyTracerProvider.ts

willarmiros:
@Flarna is this ready for merge?

Flarna:
@willarmiros I have approved yes. Not sure about the general guidelines. At least in the past one or two of the otel-js maintainers had to approve but maybe this has changed in contrib repo.

longility:
> Ah I didn't see that @dyladan had been re-requested for a review nvm - my bad.

I figured I made enough changes for him to re-review to spot check. Sorry for the delay.

dyladan (approved):

PR #666, opened 2021-09-18
fix: redis instrumentation startup stream check

jonbcampos-alto:

- When instrumenting Redis I get the error that the stream property already exists and the current check doesn't find it. Switching to `hasOwnProperty` finds the property and properly stops the duplicate addition.

- change `if(!this.stream){` to `if(!this.hasOwnProperty('stream'){`


blumamir (commented):
Thanks for finding and creating a fix.
Is it possible to write a simple test for this case?

jonbcampos-alto:
@blumamir let me see what I can do really quickly

jonbcampos-alto:
@blumamir I am having issues creating a test case that successfully checks the negative case. However I know by changing the code in node_modules that this fixes the issue. I am happy to share what I have but would be curious if you or anyone else had thoughts on how to reproduce the issue in a unit test.

jonbcampos-alto:
@blumamir I found the real problem. This library doesn't play well with the [GCP cloud trace](https://github.com/googleapis/cloud-trace-nodejs#readme) specifically if you are using the redis plugin due to this define property issue. Any thoughts?

blumamir:
> @blumamir I found the real problem. This library doesn't play well with the [GCP cloud trace](https://github.com/googleapis/cloud-trace-nodejs#readme) specifically if you are using the redis plugin due to this define property issue. Any thoughts?

Now that I look at the code of GCP, I can see how this issue happens. Looks like the GCP instrumentation [is doing the same trick as the code for OTEL instrumentation](https://github.com/googleapis/cloud-trace-nodejs/blob/ec44590f85018ef5f760567b9d1f359452b78f2c/src/plugins/plugin-redis.ts#L32), which means they don't play along currently.
Even with your suggested fix, the code will still crash if you reverse the order (initialize GCP instrumentation before OTEL), since GCP will now try to define the existing property set by OTEL.
With your fix, and current initialization order, the otel instrumentation will now never have a chance to [bind its context to the event emitter](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/648ffb714b653347838c9fd051ae5001b0f5a2c7/plugins/node/opentelemetry-instrumentation-redis/src/utils.ts#L66) which is also something to consider.

Since this subject is broader than the specific fix for redis instrumentation crash, I would like to add it to the agenda for the SIG meeting on Wednesday. Is it ok with you to wait with this PR for Wedensday? will you be able to join the meeting and discuss it?

jonbcampos-alto:
Happily. I agree that this PR doesn't fix the real problem. Just let me
know how to know if that meeting and I'll be there.

Jonathan Campos
CTO and Vice-president of Engineering
ridealto.com

blumamir:
Great :)
You can find the zoom link and google doc link for the meeting notes in [opentelemetry-community](https://github.com/open-telemetry/community#special-interest-groups) under "JavaScript: SDK"

weyert:
@jonbcampos-alto There is also a Google Cloud Trace exporter for Opentelemetry that I have been using happily for a few months

jonbcampos-alto:
@weyert do you have a link to the one you've used? for some reason I can't get the one I've found working

weyert:
> @weyert do you have a link to the one you've used? for some reason I can't get the one I've found working

This is the one I am using: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js
Did you try this one?

blumamir (approved):
This fix will prevent the otel instrumentation from throwing in case it runs parallel with GCP redis instrumentation, and otel was setup before GCP, so I support it.
As stated in the comments and discussed in the SIG, they are still not compatible to work together, and it is advisable to have only a single instrumentation agent enabled for an app.

blumamir:
@jonbcampos-alto - can you please fix the lint issue?
it says: "Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins".
Probably [need to be changed to `Object.prototype.hasOwnProperty.call(this, 'stream')`](https://eslint.org/docs/rules/no-prototype-builtins)

jonbcampos-alto:
@blumamir shouldn't we just close this PR as we wouldn't want to support it?

blumamir:
> @blumamir shouldn't we just close this PR as we wouldn't want to support it?

I still think that checking for the `hasOwnProperty` is more "correct" than checking for `this.stream`. Although it won't fully fix the co-existing with GPC instrumentation, it will at least resolve the crash you were having in this specific setup.

jonbcampos-alto:
> This is the one I am using: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js
> Did you try this one? There 0.12 released yesterday should work as it accommodates for the package renames etc

I did. I released the update today and you can basically see when it went in from the attached chart.
Screen Shot 2021-09-30 at 3 00 48 PM

Can you think of any way to successfully debug what is going on and why it might not be providing the data necessary?

dyladan:
Please use conventional commits for the PR title as this is used to determine versions for release

jonbcampos-alto:
@dyladan you're right. missed that

vmarchaud:
@jonbcampos-alto I'm good to merge the PR but you'll need to fix the lint issue

vmarchaud (approved):

Rauno56:
Closing this for https://github.com/open-telemetry/opentelemetry-js-contrib/pull/818

PR #678, opened 2021-09-26
feat(instrumentation-aws-sdk): upstream aws-sdk instrumentation from ext-js

blumamir:
This PR is for upstreaming the aws-sdk instrumentation from [this repo](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-aws-sdk).

I copied the code here and made the required modification to match the contrib repo's tooling and conventions.

This plugin has been around for about a year and a half and has proven stable in production. it has >5K weekly downloads in npm and was actively maintained in the original repo. All issues and bug fixes along this period had been addressed and fixed.
The `aws-sdk` V2 and V3 codebase is kind of complex and was designed in a way that is not very instrumentation-friendly. I had to implement very complicated and non-trivial patches to be able to extract the needed attributes, as well as normalize the data across V2 and V3 - which can be quite hard to review and understand I suppose. Please share ideas for improvements if you have any or open discussions on things that require additional clarifications.

I am open to discuss any issue regarding the porting, in praticular:
- code and instrumentation conventions
- specification compliance and known gaps
- documentation
- testing strategy for this codebase
- breaking changes from the current published version (as this is a good chance to fix those with the new npm package name)


plugins/node/opentelemetry-instrumentation-aws-sdk/README.md
```diff
@@ -0,0 +1,120 @@
...
+```js
+const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
+const { registerInstrumentations } = require('@opentelemetry/instrumentation');
+const { AwsInstrumentation } = require('opentelemetry-instrumentation-aws-sdk');
+
+const traceProvider = new NodeTracerProvider();
+traceProvider.register();
+
+registerInstrumentations({
+  traceProvider,
```
Flarna, line 30: Please remove passing the traceProvider here like we did it in other samples. Instrumentations should pick up the global `TracerProvider`.

plugins/node/opentelemetry-instrumentation-aws-sdk/README.md
```diff
@@ -0,0 +1,120 @@
...
+npm install --save opentelemetry-instrumentation-aws-sdk
+```
+
+## Usage
+For further automatic instrumentation instruction see the [@opentelemetry/instrumentation](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation) package.
+
+```js
+const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
+const { registerInstrumentations } = require('@opentelemetry/instrumentation');
+const { AwsInstrumentation } = require('opentelemetry-instrumentation-aws-sdk');
```
Flarna, line 24:
```suggestion
const { AwsInstrumentation } = require('@opentelemetry/instrumentation-aws-sdk');
```

plugins/node/opentelemetry-instrumentation-aws-sdk/README.md
```diff
@@ -0,0 +1,120 @@
...
+[![Apache License][license-image]][license-image]
+
+[component owners](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/component_owners.yml): @willarmiros @NathanielRN @blumamir
+
+This module provides automatic instrumentation for [`aws-sdk` v2](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/) and [`@aws-sdk` v3](https://github.com/aws/aws-sdk-js-v3)
+
+## Installation
+
+```bash
+npm install --save opentelemetry-instrumentation-aws-sdk
```
Flarna, line 15:
```suggestion
npm install --save @opentelemetry/instrumentation-aws-sdk
```

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -0,0 +1,80 @@
...
+        "watch": "tsc -w"
+    },
+    "peerDependencies": {
+        "@opentelemetry/api": "^1.0.1"
+    },
+    "dependencies": {
+        "@opentelemetry/core": "^0.25.0",
+        "@opentelemetry/instrumentation": "^0.25.0",
+        "@opentelemetry/semantic-conventions": "^0.25.0",
+        "opentelemetry-propagation-utils": "^0.24.0"
```
Flarna, line 52: Should we move this package also?

blumamir: Yes you are right, I need to move it as well.

willarmiros (approved):
Code LGTM, apart from breaking config changes discussed offline. Looks like yargs-parser dropped Node 6 and 8 support a while ago in v19.0.0: https://github.com/yargs/yargs-parser/releases/tag/v19.0.0

So if we can force the package to adopt a version before then (`18.x`) it should fix the build issues. In the future we can hopefully drop support for Node 8 altogether.

willarmiros:
Looks like most of the linting errors can be fixed with `npm run lint:fix`

https://github.com/open-telemetry/opentelemetry-js-contrib/pull/678/checks?check_run_id=3825856708#step:5:647

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
```diff
@@ -0,0 +1,620 @@
...
+          commandName,
+          command.input,
+          undefined
+        );
+        const requestMetadata =
+          self.servicesExtensions.requestPreSpanHook(normalizedRequest);
+        const span = self._startAwsV3Span(normalizedRequest, requestMetadata);
+        const activeContextWithSpan = trace.setSpan(context.active(), span);
+
+        const handlerPromise = new Promise(async (resolve, reject) => {
```
willarmiros, line 433: Looks like another error is that this executor shouldn't be async... can you `await` the regionPromise outside of this executor to avoid making it async?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
```diff
@@ -0,0 +1,620 @@
...
+              resolve(res);
+            })
+            .catch(err => reject(err));
+        });
+
+        return requestMetadata.isIncoming
+          ? bindPromise(handlerPromise, activeContextWithSpan, 2)
+          : handlerPromise;
+      };
+      // @ts-ignore
```
willarmiros, line 516: Also need to remove this

willarmiros (commented):
I think you just need to remove the `ts-ignore` comments, which will obviously create new errors, but at least they'll be specific errors that we can address.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
blumamir: The `_getV3MiddlewareStackResolvePatch` is sync, so I cannot do it there.
What I can do is to convert it from await to Promise.finally, but that would create an ugly code, just to resolve the lint error, which is a bummer

blumamir: ended up refactoring this code to use promise callbacks instead of async/await to calm down the linter.
The code is less idiomatic and harder to read. we might want to refactor this one day

willarmiros (commented):
Great to see the CI checks passing, unfortunate we had to do some acrobatics to get lint passing but it still LGTM and should be ready for review by @open-telemetry/javascript-approvers if they want to.

The only thing blocking from merging is adding the new packages to the [release-please config](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/release-please-config.json)

plugins/node/opentelemetry-instrumentation-aws-sdk/doc/sqs.md
```diff
@@ -0,0 +1,56 @@
...
+According to open telemetry specification (and to reasonable expectation for trace structure), user of this library would expect to see one span for the operation of receiving messages batch from SQS, and then, **for each message**, a span with it's own sub-tree for the processing of this specific message. 
+
+For example, if a `receiveMessages` returned 2 messages: 
+* `msg1` resulting in storing something to a DB.
+* `msg2` resulting in calling an external HTTP endpoint.  
+
+This will result in a creating a DB span that would be the child of `msg1` process span, and an HTTP span that would be the child of `msg2` process span (in opposed to mixing all those operations under the single `receive` span, or start a new trace for each of them).
+
+Unfortunately, this is not so easy to implement in JS:
+1. The SDK is calling a single callback for the messages batch, and it's not straight forward to understand when each individual message processing starts and ends (and set the context correctly for cascading spans).
```
Rauno56, line 30:
```suggestion
1. The SDK is calling a single callback for the messages batch, and it's not straightforward to understand when each individual message processing starts and ends (and set the context correctly for cascading spans).
```

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -0,0 +1,80 @@
+{
+    "name": "@opentelemetry/instrumentation-aws-sdk",
+    "version": "0.25.0",
+    "description": "open telemetry automatic instrumentation for the `aws-sdk` package",
```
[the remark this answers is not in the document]
Rauno56: There are other places that need that fix as well...

packages/opentelemetry-propagation-utils/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "clean": "tsc --build --clean tsconfig.json tsconfig.esm.json",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "eslint . --ext .ts --fix",
+    "version": "node ../../scripts/version-update.js",
+    "watch": "tsc --build --watch tsconfig.json tsconfig.esm.json",
+    "prepare": "npm run compile"
+  },
+  "repository": "open-telemetry/opentelemetry-js-contrib",
+  "keywords": [
+    "opentelemetry",
```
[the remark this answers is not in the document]
blumamir: Changed. these utils are not only for aws, they are general for messaging systems

plugins/node/opentelemetry-instrumentation-aws-sdk/README.md
```diff
@@ -0,0 +1,124 @@
...
+## Span Attributes
+Both V2 and V3 instrumentations are collecting the following attributes:
+| Attribute Name | Type | Description | Example |
+| -------------- | ---- | ----------- | ------- |
+| `rpc.system` | string | Always equals "aws-api" | 
+| `rpc.method` | string | he name of the operation corresponding to the request, as returned by the AWS SDK. If the SDK does not provide a way to retrieve a name, the name of the command SHOULD be used, removing the suffix `Command` if present, resulting in a PascalCase name with no spaces. | `PutObject` |
+| `rpc.service` | string | The name of the service to which a request is made, as returned by the AWS SDK. If the SDK does not provide a away to retrieve a name, the name of the SDK's client interface for a service SHOULD be used, removing the suffix `Client` if present, resulting in a PascalCase name with no spaces. | `S3`, `DynamoDB`, `Route53` |
+| `aws.region` | string | Region name for the request | "eu-west-1" |
+
+### V2 attributes
```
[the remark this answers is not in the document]
willarmiros: yeah I'd always be in favor of adding them later if customers ask for them rather than adding them eagerly if they're awkward to support in v3.

mattfysh:
Hey @blumamir - thanks for this great contrib!

I had a question about using the table name as "DB_NAME" attribute, I wonder if this should be "DynamoDB" as shown in this screenshot: https://docs.aws.amazon.com/xray/latest/devguide/images/scorekeep-PUTrules-timeline.png

The table name would then be stored alongside the operation name, and rendered by the Xray UI in the segment description as "OperationName: TableName"

On a loosely related note, do you know if there is any documentation covering how Xray decides what to render as the grey description text (to the right of the coloured bar in the timeline)? Thanks!

blumamir:
> I had a question about using the table name as "DB_NAME" attribute, I wonder if this should be "DynamoDB" as shown in this screenshot: https://docs.aws.amazon.com/xray/latest/devguide/images/scorekeep-PUTrules-timeline.png

Currently, `db.name` is filled with the table name, and `db.system` contains the "dynamodb" value. I agree that maybe the table name is more appropriate in an attribute like `db.table`, but the spec only has [`db.sql.table`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes-for-specific-technologies) but dyanmodb is not an sql database.

> The table name would then be stored alongside the operation name, and rendered by the Xray UI in the segment description as "OperationName: TableName"

Not sure what you are suggesting here. Are you suggesting to add the table name in the span name? or to concat the table name into the `db.operation` attribute?

Regarding Xray, I have no experience with it. maybe @willarmiros  or @NathanielRN  can help

willarmiros:
> On a loosely related note, do you know if there is any documentation covering how Xray decides what to render as the grey description text (to the right of the coloured bar in the timeline)? Thanks!

I don't believe this is documented since it's more of an internal implementation detail - the closest documentation we have is [this on viewing traces](https://docs.aws.amazon.com/xray/latest/devguide/xray-console.html#xray-console-traces). However I can say that basically the gray text attempts to represent the operation performed, and depends on the `namespace` of a subsegment, and the attributes of that subsegment. For instance, if the namespace is `aws`, then the text is `aws.operation`, and for some cases it will also include a resource like for DynamoDB it will be `aws.operation: aws.table_name`. If the namespace is `remote`, the gray text will be something like `Remote: http.request.method`.

Issue #680, opened 2021-09-29
npm run compile fails on windows


Flarna:
### What version of OpenTelemetry are you using?

Latest main

### What version of Node are you using?

14.18.0

### What did you do?

`npm install` which triggers `npm run compile`

### What did you expect to see?

success compilation

### What did you see instead?

compilation failed with `ERR! lerna Unknown arguments: pkg, get, name)`

### Additional context

Seems to be caused by #659 which added linux style arguments `$(npm pkg get name)` to precompile script

dyladan:
Maybe we should replace it with a node script in scripts

Flarna:
Or we just copy paste them module name.

dyladan:
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/681

PR #682, opened 2021-09-29
chore: use package name in compile script

Flarna:

#680

Replace the linux style script to avoid problems on windows.


Flarna:
duplicate of #681 681

PR #689, opened 2021-10-05
fix: Add support for setConfig in ioredis instrumentation

### LINKED ISSUE #624: ioredis: SetConfig not supported
v0.24.0

12

### What did you do?
tried to call `setConfig` function from the `InstrumentationAbstract` api on an instance of the ioredis instrumentation to update the config.

instrumentation to use the new config on future operations

### What did you see instead?
previous config is in use

### Additional context
ioredis instrumentation is [using the `_config` object during patching](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts#L88) as closure param instead of reading the current attribute from `this` everytime it should be used. Thus instrumentation will not be affected when `setConfig` is called.

mustafain117:

- resolves #624 

- Override base setConfig function in `IORedisInstrumentation`
- Unit test added to verify setConfig function works as expected, asserts instrumentation uses the new config on future operations


mustafain117:
@Flarna Thank you for the feedback, I have updated the instrumentation to use the latest config from `this` instead and removed the override on `setConfig`.

plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts
```diff
@@ -96,4 +97,103 @@ export class IORedisInstrumentation extends InstrumentationBase<
...
+
+  private defaultDbStatementSerializer: DbStatementSerializer = (
+    cmdName,
+    cmdArgs
+  ) =>
+    Array.isArray(cmdArgs) && cmdArgs.length
+      ? `${cmdName} ${cmdArgs.join(' ')}`
+      : cmdName;
+
+  private traceSendCommand = (
```
blumamir, line 109: another small suggestion: since you copied the `traceSendCommand` from utils to instrumentation, I think it can be nice to also copy the `traceConnection` function, so the codebase is consistent

vmarchaud:
@mustafain117 You'll need to rebase the PR

Rauno56 (approved):

mustafain117:
@obecny Updated it

PR #690, opened 2021-10-06
fix: Fix restify instrumentation utils that might accept undefined types

faizhasim:

- Using restify instrumentation on middlewares that return nothing (return undefined) will break the instrumentation due to `typeof undefined.then`.

- This PR just add a simple defensive check on possibly undefined value.
- Additional coverage added for utils.ts


plugins/node/opentelemetry-instrumentation-restify/test/utils.test.ts
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { isAsyncFunction, isPromise } from '../src/utils';
+import * as assert from 'assert';
+
+describe('utils', () => {
+  describe('isPromise', () => {
+    it('should be able to validate a promise to be true', () => {
+      assert.strictEqual(isPromise(Promise.resolve()), true);
```
Flarna, line 22: Maybe add a testcase that validates that a non built in Promise (e.g. Bluebird or some home grown variant) is also detected as promise.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-restify/test/utils.test.ts
faizhasim: I've added Bluebird.

plugins/node/opentelemetry-instrumentation-restify/package.json
```diff
@@ -50,8 +50,10 @@
     "@opentelemetry/context-async-hooks": "0.25.0",
     "@opentelemetry/sdk-trace-base": "0.25.0",
     "@opentelemetry/sdk-trace-node": "0.25.0",
+    "@types/bluebird": "^3.5.36",
     "@types/mocha": "7.0.2",
     "@types/node": "14.17.9",
+    "bluebird": "^3.7.2",
```
Flarna, line 56:
```suggestion
    "bluebird": "3.7.2",
```
devDependencies are pinned in general

vmarchaud (approved):

Flarna:
Instead of bluebird or some other promise 3rd party you could write your own promise like object in a test.
My main point is that the instrumentation explicit tests on `typeof(foo.then) === "function"` and not for `instanceof Promise` therefore I assumed this should be verified.

But I fine with removing bluebird.

Issue #616, opened 2021-08-12
instrumentation-document-load doesn't support re-enables


kkruk-sumo:
### What version of OpenTelemetry are you using?
```
"@opentelemetry/api": "^1.0.2",
"@opentelemetry/instrumentation": "^0.24.0",
"@opentelemetry/instrumentation-document-load": "^0.24.0",
"@opentelemetry/web": "^0.24.0",
```

### What version of Node are you using?
`12.7.0`

### What did you do?
I enabled `DocumentLoadInstrumentation` twice.

### What did you expect to see?
Only one trace with `documentLoad` span to be created.

### What did you see instead?
Two traces were generated.

### Additional context
Instrumentations can be disabled and re-enabled in runtime. Other JS instrumentations like XHR or fetch supports re-enabling.
In my opinion the `instrumentation-document-load` should does nothing after being enabled again.

Issue #668, opened 2021-09-21
Instrumentation for capturing resource loads after load event


mhennoch:
Currently [document-load](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/web/opentelemetry-instrumentation-document-load) instrumentation captures resources(images, scripts etc. ) when load event fires. Everything happening after that moment is not captured. One example where this could be useful is pages with infinite scrolling eg. twitter where images etc. are loaded as you scroll. 

Here is a simplified version of what we internally use at the moment:

```
window.addEventListener('load', () => {
    this.observer = new PerformanceObserver((list) => {
          list.getEntries().forEach(entry => {
              this._createSpan(entry);
          });
      });
      this.observer.observe({ entryTypes: ['resource'] });
});
```

The idea is simple but there are some gotchas: making sure not to capture same resource twice(once on page load and once in this), configuration eg. which resource types are captured, every span is in a different trace as we won't have a parent, testing this is quite tricky etc.

We would ideally like to have this upstreamed either as a separate instrumentation or part of something else eg. document-load.
Thoughts?

obecny:
yes - create new plugin, don't try to squeeze too many edge cases into this plugin. The plugin name is document load and I think it does exactly what the name stands for.

Issue #320, opened 2021-01-25
Report web vitals in Document Load instrumentation


bradfrosty:
**Is your feature request related to a problem? Please describe.**
It's unrelated to a problem. This is an additional feature to add to the document load instrumentation.

**Describe the solution you'd like**
[Web Vitals](https://web.dev/vitals/) are a set of web performance metrics for analyzing user experience metrics rather than just navigation timing metrics. They are going to be used by [Google in the near future](https://developers.google.com/search/blog/2020/05/evaluating-page-experience) for evaluating SEO, meaning that many developers are optimizing for these metrics now. 

It would be useful to add these metrics into the document load plugin's existing span events.

**Describe alternatives you've considered**
It is possible for me to augment the plugin on the application side, but it feels beneficial for other OTel users to be in the library.

**Additional context**
[Here is the library](https://www.npmjs.com/package/web-vitals) for calculating these metrics on the client.

JapuDCret:
I think this would be more appropriate as it's own plugin.

Mainly because some of the metrics make use of vendor(/chrome)-specific APIs, which require either polyfills or don't work at all.
Excerpt from the current [README.md](https://github.com/GoogleChrome/web-vitals):

> Browser support for each function is as follows:
> 
>     getCLS(): Chromium,
>     getFCP(): Chromium, Safari Technology Preview
>     getFID(): Chromium, Firefox, Safari, Internet Explorer (with the polyfill)
>     getLCP(): Chromium
>     getTTFB(): Chromium, Firefox, Safari, Internet Explorer

obecny:
> I think this would be more appropriate as it's own plugin.
> 
> Mainly because some of the metrics make use of vendor(/chrome)-specific APIs, which require either polyfills or don't work at all.
> Excerpt from the current [README.md](https://github.com/GoogleChrome/web-vitals):
> 
> > Browser support for each function is as follows:
> > ```
> > getCLS(): Chromium,
> > getFCP(): Chromium, Safari Technology Preview
> > getFID(): Chromium, Firefox, Safari, Internet Explorer (with the polyfill)
> > getLCP(): Chromium
> > getTTFB(): Chromium, Firefox, Safari, Internet Explorer
> > ```

I cannot agree more, this is very good example, thx.

bradfrosty:
[...]

PR #702, opened 2021-10-15
feat: Capture Web vitals

mhennoch:
This PR is an initial try to upstream web-vitals capturing. 

[Web Vitals](https://web.dev/vitals/)  is an initiative by Google to provide unified guidance for quality signals that are essential to delivering a great user experience on the web. Currently the core web vitals are:

- Largest Contentful Paint (LCP)
- First Input Delay (FID)
- Cumulative Layout Shift (CLS)

Google has a [library](https://github.com/GoogleChrome/web-vitals)  that can be used to capture them. It currently has 1.4m weekly downloads. This instrumentation is basically just a wrapper to that library. Ran into several issues so opening this not so polished PR to get some feedback. Main Issues:

1. ~~Currently only signal that is stable enough is span but it doesn't really make sense to send them as spans. in theory LCP and FID could be made into spans as there is a start time and duration but CLS is just a score eg. 0.1234. We have been using 0 duration spans for these where the real value is sent as span attribute. Eventually these should probably be metrics/logs. Any better ideas for now?~~ Discussed at SIG, if I take ownership it is fine to use spans for now
2.  ~~I have no access to event handlers set by web-vitals library to remove them in disable method. Currently browser instrumentations are enabled twice if you don't use {enabled: false} in config. So in order to not have double the spans I keep the enable status myself. Can't access parent's _enabled or super.enable. Not sure what the usecase for enable/disable is especially in browser. Related issues #616, [#2410](https://github.com/open-telemetry/opentelemetry-js/issues/2410)~~ Fixed with [#2610](https://github.com/open-telemetry/opentelemetry-js/pull/2610)
3. Can't unit tests FID as it doesn't seem to happen if event is created by JS eg. document.body.click(). I think these should be tested with selenium only (needs [#2570](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/702)). Currently have some very basic unit tests only.


kkruk-sumo:
This is great and I was looking for it for some time. Did you think about adding these metrics into `documentLoad` spans from the `document-load` instrumentation. I added FP and FCP there.

Do you know the size of the library made by Google?

Also I was looking into this library long time ago and I remember that a user need to click on a page in order to get these metrics. Is is still true? Because of that I didn't manage to use it in `documentLoad` spans.

mhennoch:
> Did you think about adding these metrics into `documentLoad` spans from the `document-load` instrumentation. I added FP and FCP there.

I did think about it but I think the general consensus is to keep things separated. I asked a related question in [here](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/668#issuecomment-923960299). Also it doesn't really make sense to add them to the documentLoad spans. CLS can change much later than documentLoad  for example.

> Do you know the size of the library made by Google?

From their readme: The web-vitals library is a tiny (~1K) modular library.

> Also I was looking into this library long time ago and I remember that a user need to click on a page in order to get these metrics. Is is still true? Because of that I didn't manage to use it in `documentLoad` spans.

By definition you can't have FID without input. Also LCP can change until there is a click etc. on the document and CLS can change until page hidden event so I guess there is no other way.

I'll undraft it soon so we can push this forward.

dyladan:
I agree metrics would make much more sense for these types of things. Using the metrics API might be difficult because it is changing, but i'm not sure span is a good fit. Might you want to wait until the API stabilizes? It shouldn't be too long.

bradfrosty:
For some added context, I made an issue [last January](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/320) to add this feature. The reasoning at the time was due to browser specific polyfills needed to capture the metrics. 

I think it would make sense to add these as events on the document load plugin, or via the metrics API as @dyladan said.

dyladan:
Events on the document load make sense to me as well, but that makes analysis harder unless your backend understands what the events are. There are many off-the-shelf tools to process and analyze metrics. Far fewer can generate metrics from span events.

There is a module of the collector which generates metrics from spans (like request counts, request durations, etc). Maybe it could be extended to generate metrics from predefined span events? This would likely require a large spec effort, but would be very powerful.

bradfrosty:
Good call re: making the analysis harder since not all backends read the events. Useful for looking at when analyzing the trace, but these metrics are more useful in charts. 

> Maybe it could be extended to generate metrics from predefined span events?

This would be really really useful and remove some of the redundancy between span events and metric instrumentation.

mhennoch:
I think these should ideally be standalone events that get made into metrics in the backend also. You do want to see these as metrics but also on your session/trace timeline as events. They can also have some extra information eg. which element caused the paint etc. Don't think we can add them as events on docload spans as we would have to keep the span open for some unknown amount of time( first user interaction for lcp and fid or page hidden). Sending data on leaving the page can be unreliable and we might lose that span altogether. This was also discussed briefly in RUM SIG and we decided to make some examples and go to the spec SIG with them to find out what they think.

PR #714, opened 2021-10-23
feat: support fastify - instrumentation-nestjs-core

anchan828:

- Nestjs supports express and fastify. However, the @opentelemetry/instrumentation-nestjs-core package currently only supports requests of express.
  - For fastify:
    - Use [`req.url`](https://github.com/fastify/fastify/blob/505780bbfb0aa004537af4bf70d6ee5bdb228dc1/lib/request.js#L120) instead of `req.originalUrl`
    - Use [`req.routerPath`](https://github.com/fastify/fastify/blob/505780bbfb0aa004537af4bf70d6ee5bdb228dc1/lib/request.js#L130) instead of `req.route.path`

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/44d21fe5030d2f64bedde1efc50a1173e0c49f3d/plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts#L190-L191

- Access the property as a request of express, and if it does not exist, access it as a fastify.

I am wondering how to write the test. I think it's simple and I don't think needs a test if possible 🤔.


Rauno56 (approved):

Rauno56:
Thanks for your contribution, @anchan828!

Issue #396, opened 2021-03-23
add lint task for markdown


obecny:
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/395#issuecomment-805120282

Arbiv:
PR opened, ready for reivew

Arbiv:
@obecny can this issue be closed now that the PR is merged?

PR #716, opened 2021-10-28
ci: add markdown lint and fixed markdown files

Arbiv:
#396

Added lint for markdown


plugins/node/opentelemetry-instrumentation-fastify/src/version.ts
```diff
@@ -1,18 +0,0 @@
-/*
```
dyladan, line 1: Good catch but this might be better done in a separate PR since it has nothing to do really with what the PR is about.

obecny (approved):
lgtm

Arbiv:
UT failure and coverage decrease seems irrelevant to the changes in the PR

Arbiv:
@dyladan I fixed the comment, can you please re-review it?

vmarchaud (approved):

vmarchaud:
@Arbiv Could you please fix the conflict so we can merge this ?

Arbiv:
@vmarchaud Fixed conflict

PR #720, opened 2021-10-30
fix: adding nested input variables in graphql plugin

bodymindarts:

GraphQL inputs are frequently nested like:
```
{
  "input": {
    "value1": 1,
    "value2": 2
  }
}
```
Currently this would result the following call:
```
span.setAttribute("graphql.variables.input", Object({ value1: 1, value2: 2}))
```
But since `SpanAttribute` doesn't accept an `Object` type it silently fails
and doesn't add the attribute at all.

This PR goes depth-first into the nested structure and adds each
variable with the appropriate path like:
```
span.setAttribute("graphql.variables.input.value1", 1)
span.setAttribute("graphql.variables.input.value2", 2)
```

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
@@ -32,6 +32,31 @@ import {
 
 const OPERATION_VALUES = Object.values(AllowedOperationTypes);
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function addInputVariableAttribute(span: api.Span, key: string, variable: any) {
+  if (Array.isArray(variable)) {
+    variable.forEach((value, idx) => {
+      addInputVariableAttribute(span, `${key}.${idx}`, value);
+    });
+  } else if (variable instanceof Object) {
+    Object.entries(variable).forEach(([nestedKey, value]) => {
+      addInputVariableAttribute(span, `${key}.${nestedKey}`, value);
+    });
+  } else {
+    span.setAttribute(`${AttributeNames.VARIABLES}${String(key)}`, variable);
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function addInputVariableAttributes(
+  span: api.Span,
+  variableValues: { [key: string]: any }
+) {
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -32,6 +32,32 @@ import {
 
 const OPERATION_VALUES = Object.values(AllowedOperationTypes);
 
+function addInputVariableAttribute(span: api.Span, key: string, variable: any) {
+  switch (typeof variable) {
```
obecny, line 36: I think this would be enough 

```
      if (Array.isArray(variable)) {
        variable.forEach((value, idx) => {
          addInputVariableAttribute(span, `${key}.${idx}`, value);
        });
      } else if (variable instanceof Object) {
        Object.entries(variable).forEach(([nestedKey, value]) => {
          addInputVariableAttribute(span, `${key}.${nestedKey}`, value);
        });
      }
      else {
        span.setAttribute(`${AttributeNames.VARIABLES}${String(key)}`, variable);
      }
```

obecny (commented):
one concern to improve, other than that lgtm

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
bodymindarts: Fixed

bodymindarts:
Force pushed with suggested modification

dyladan:
Please use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) in your PR title as they are used to autogenerate new versions when packages are released. Your current title is not in the correct format and would be ignored, causing the package to not be released. If it is a bugfix that should only modify the PATCH version, please use `feat: message_here`. Otherwise, please use `feat: message_here`.

bodymindarts:
Force pushed to amend commit message and fix linting.

dyladan:
Thanks. Fixing the commit message doesn't matter as much as the PR title since the commits get squashed anyway and the PR title is used as the commit message.

bodymindarts:
Another attempt to fix linting

obecny (approved):

vmarchaud (approved):

PR #722, opened 2021-11-01
fix: increase ec2 resource detector timeout

### LINKED ISSUE #708: resource-detector-aws times out while acquiring token

* `@opentelemetry/sdk-node@0.26.0`
* `@opentelemetry/resource-detector-aws@0.24.0`

v14.15.4

### What did you do?

I'm using the Node SDK with automatic resource detection.

I expect to see my spans tagged with my AWS EC2 information (`cloud.*`, etc.).

### What did you see instead?

I do not see any resource attributes for my AWS EC2 information.

### Additional context

I added the following to my code to see additional debugging information:

```js
const { diag, DiagConsoleLogger, DiagLogLevel } = require('@opentelemetry/api');
diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
```

This results in logs like the following:

```
AwsEc2Detector failed: EC2 metadata api request timed out
```

I modified the code around [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/f5851e72512117dbce571a42930a90c560dbf63d/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsEc2Detector.ts#L122) to include additional information:

```diff
  private async _fetchString(options: http.RequestOptions): Promise<string> {
    return new Promise((resolve, reject) => {
+     const start = Date.now();
      const timeoutId = setTimeout(() => {
        req.abort();
        reject(new Error('EC2 metadata api request timed out.'));
-     }, 1000);
+     }, 10000);

      const req = http.request(options, res => {
+       console.log(`connection took ${Date.now() - start}ms to initialize`);
        clearTimeout(timeoutId);
        const { statusCode } = res;
        res.setEncoding('utf8');
        let rawData = '';
        res.on('data', chunk => (rawData += chunk));
        res.on('end', () => {
          if (statusCode && statusCode >= 200 && statusCode < 300) {
            try {
              resolve(rawData);
            } catch (e) {
              reject(e);
            }
          } else {
            reject(
              new Error('Failed to load page, status code: ' + statusCode)
            );
          }
        });
      });
      req.on('error', err => {
        clearTimeout(timeoutId);
        reject(err);
      });
      req.end();
    });
  }
```

With this, I now see logs like the following:

```
connection took 1924ms to initialize
connection took 9ms to initialize
connection took 2ms to initialize
```

So, it takes an oddly long amount of time to initialize the first connection to the metadata service to acquire a token, which exceeds the 1000ms timeout that's hardcoded for requests. This renders the detector unable to get any information, which means none of my spans have the correct attributes set.

I'm not sure what the right solution is here. Either we can increase the default timeout to something like 5000ms, or else make the timeout configurable. I'd prefer the former, as that would allow me to avoid adding additional config to the SDK, but it also feels like that might end up with a game of whack-a-mole if other folks encounter requests that take even longer.

willarmiros:
Fixes #708

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsEc2Detector.ts
+++ b/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsEc2Detector.ts
@@ -44,7 +44,7 @@ class AwsEc2Detector implements Detector {
   readonly AWS_INSTANCE_HOST_DOCUMENT_PATH = '/latest/meta-data/hostname';
   readonly AWS_METADATA_TTL_HEADER = 'X-aws-ec2-metadata-token-ttl-seconds';
   readonly AWS_METADATA_TOKEN_HEADER = 'X-aws-ec2-metadata-token';
-  readonly MILLISECOND_TIME_OUT = 1000;
+  readonly MILLISECOND_TIME_OUT = 5000;
 
   /**
    * Attempts to connect and obtain an AWS instance Identity document. If the


NathanielRN (approved):
Nice!

Flarna (approved):

dyladan (approved):

nwalters512:
Thanks for the fix! 🙌 any estimate on when a new version of the package will be released?

PR #724, opened 2021-11-04
feat(instrumentation-pino): bump supported pino version to v7.x

politician:

- When upgrading to the latest version of Pino (v7.x), the OpenTelemetry instrumentation doesn't work anymore

- There is a hard check on the Pino version in use to prevent errors in case of major upgrades. I tested with the latest Pino version and bumped that number.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pino/README.md
+++ b/plugins/node/opentelemetry-instrumentation-pino/README.md
@@ -55,7 +55,7 @@ When no span context is active or the span context is invalid, injection is skip
 
 ### Supported versions
 
-`>5.14.0` and `6.x`
+`>5.14.0` and `7.x`
 
 ## Useful links
 
--- a/plugins/node/opentelemetry-instrumentation-pino/README.md
+++ b/plugins/node/opentelemetry-instrumentation-pino/README.md
@@ -55,7 +55,7 @@ When no span context is active or the span context is invalid, injection is skip
 
 ### Supported versions
 
-`>5.14.0` and `7.x`
+`>=5.14.0 <8`
 
 ## Useful links
 
--- a/plugins/node/opentelemetry-instrumentation-pino/src/types.ts
+++ b/plugins/node/opentelemetry-instrumentation-pino/src/types.ts
@@ -16,7 +16,6 @@
 
 import { Span } from '@opentelemetry/api';
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
-import type * as pino from 'pino';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type LogHookFunction = (span: Span, record: Record<string, any>) => void;
@@ -25,4 +24,4 @@ export interface PinoInstrumentationConfig extends InstrumentationConfig {
   logHook?: LogHookFunction;
 }
 
-export type Pino = typeof pino;
+export type Pino = any;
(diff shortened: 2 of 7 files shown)


Flarna:
I think the dev-dependency should be also updated to pino 7. Or even better test with all supported versions.

dyladan:
There are other packages in this repo which use a module to test against multiple dependency versions. You may want to try that.

dyladan:
Here is an example. https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-ioredis/package.json#L68

politician:
@dyladan @Flarna I won't be updating the dev dependency to Pino 7.x because it requires more code changes and am not familiar with neither Pino nor OpenTelemetry sources.

That said, I have:
- Updated Pino to the latest 6.x version
- Configured the `test-all-versions` module and successfully ran the tests in all versions supported by this module

plugins/node/opentelemetry-instrumentation-pino/README.md
```diff
@@ -55,7 +55,7 @@ When no span context is active or the span context is invalid, injection is skip
 
 ### Supported versions
 
-`>5.14.0` and `6.x`
+`>5.14.0` and `7.x`
```
Rauno56, line 58:
```suggestion
`>=5.14.0 < 8`
```

"and" sounds misleading to me.

Rauno56 (approved):

dyladan:
If changing to the 7.x pino dev dependency causes code changes, then how can we claim to support it without making those code changes?

Rauno56:
The tests pass fine, but the issue is with the type definition mismatches.

Rauno56:
Merge https://github.com/politician/opentelemetry-js-contrib/pull/1, @politician and we should be good!

obecny (approved):

politician:
@Rauno56 thanks!

dyladan:
Looks like pino itself doesn't support node 8 and 10. Will probably have to skip testing the pino instrumentation for those old node versions.

plugins/node/opentelemetry-instrumentation-pino/src/types.ts
```diff
@@ -25,4 +24,4 @@ export interface PinoInstrumentationConfig extends InstrumentationConfig {
   logHook?: LogHookFunction;
 }
 
-export type Pino = typeof pino;
+export type Pino = any;
```
Flarna, line 27: This doesn't seem a good aproach. This leaks into the public exported class `PinoInstrumentation` therefore changing this back into a more concrete type could break user code.

politician: @Rauno56 Do you want to do something about it or should I revert your PR?

politician:
I have just reverted the types changes. Can we get this PR through so we can support Pino 7 already and defer refactoring with the Pino 7.x dev dependency to another PR?

politician:
I have just added the changes from @mentos1386 to upgrade Pino to the latest dependency.

@dyladan Can we merge this pull-request please?

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-pino/src/types.ts
Rauno56: Yeah, agreed - I thought I changed all the added `any`s to proper types in later commits, but missed this or left some commits unpushed. My bad. The changes from @mentos1386 did it the correct way regarding this so all good!

PR #737, opened 2021-11-14
feat: config option to extract sqs context from message payload

habmic:

- Context propagation between SNS to SQS didn't work without enabling AWS [raw message delivery](https://docs.aws.amazon.com/sns/latest/dg/sns-large-payload-raw-message-delivery.html) feature
- By default SNS message attributes are embedded within the payload when set to SQS

- Added a configuration (sqsExtractContextPropagationFromPayload ) that parse and extract the context propagation headers from the SQS payload ( off be default )


blumamir (approved):
LGTM

blumamir:
Is anyone else planning to review this? if not I'll merge today

plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts
```diff
@@ -36,12 +36,19 @@ import { ReadableSpan } from '@opentelemetry/sdk-trace-base';
...
 import * as expect from 'expect';
+import * as sinon from 'sinon';
+import * as messageAttributes from '../src/services/MessageAttributes';
 
 const responseMockSuccess = {
   requestId: '0000000000000',
   error: null,
 };
 
+const extractContextSpy = sinon.spy(
```
Flarna, line 47: Creating a spy on this scope effects all tests. You could move it into the relevant test/testsuite and instead of calling `resetHistory()` you could create a new spy for each test and use `sinon.restore()` after each test.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts
```diff
@@ -429,4 +436,87 @@ describe('SQS', () => {
...
+        .receiveMessage({
+          QueueUrl: 'queue/url/for/unittests1',
+        })
+        .promise();
+      expect(extractContextSpy.returnValues[0]?.traceparent).toBeUndefined();
+    });
+
+    it('should extract from payload', async () => {
+      const traceparent = {
+        traceparent: {
```
Flarna, line 460: Is this object structure documented somewhere?

Flarna:
Seems my review was too late :o) but there are anyway no blockers.

But I wonder that support for `extract` is added but not for `inject`.

blumamir:
@Flarna Thanks for adding a review :)
@habmic - interested in implementing the above sinon refactor?

blumamir:
> But I wonder that support for `extract` is added but not for `inject`.

Inject is always injecting the propagation headers into the message attributes at time of "Publish" \ "SendMessage"
When SNS forwards the message to SQS, it can place these headers as message metadata, or embed them into the payload. This is a configuration option for the SNS-SQS subscription.

This PR is about support for extracting the context for the second option - taking them from message payload, which is why only extract is relevant

PR #738, opened 2021-11-14
fix(aws-sdk): calc propagation fields count before context inject

YanivD:

_Background:_ As described [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-messages.html), SQS/SNS services limited up to 10 message metadata attributes.

The instrumentation is checking if there are enough available attributes before adding more metadata attributes.

Before this pr, it ignored how many fields are going be added by applied propagators. 
After this pr, it will check how many fields are going be added, to make sure not setting more than 10 message metadata attributes.


plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/MessageAttributes.ts
```diff
@@ -64,7 +64,7 @@ export const injectPropagationContext = (
   attributesMap?: SQS.MessageBodyAttributeMap | SNS.MessageAttributeMap
 ): SQS.MessageBodyAttributeMap | SNS.MessageAttributeMap => {
   const attributes = attributesMap ?? {};
-  if (Object.keys(attributes).length < MAX_MESSAGE_ATTRIBUTES) {
+  if (Object.keys(attributes).length + propagation.fields().length <= MAX_MESSAGE_ATTRIBUTES) {
```
blumamir, line 67: can you please add a simple unit test for that?

YanivD: @blumamir  👍
[Added tests](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/738/commits/75d653f6541ea0cb30ac1df39fb8db2456e338e4)

blumamir (approved):

Rauno56 (approved):

PR #743, opened 2021-11-16
feat(user-interaction): deduplicate interaction spans by grouping them

### LINKED ISSUE #548: User events getting registered twice
@opentelemetry/api: 1.0.0
@opentelemetry/context-zone: 0.22.0

12.14.0

### What did you do?
I used open telemetry with create-react-app

```javascript
// setupTelemetry.js
import { registerInstrumentations } from "@opentelemetry/instrumentation";
import { DocumentLoadInstrumentation } from "@opentelemetry/instrumentation-document-load";
import { WebTracerProvider } from "@opentelemetry/web";
import {
  ConsoleSpanExporter,
  SimpleSpanProcessor,
} from "@opentelemetry/tracing";
import { ZoneContextManager } from "@opentelemetry/context-zone";

// Minimum required setup - supports only synchronous operations
const provider = new WebTracerProvider();
provider.addSpanProcessor(
  new SimpleSpanProcessor(new ConsoleSpanExporter())
);
provider.register({
  contextManager: new ZoneContextManager(),
});

registerInstrumentations({
  instrumentations: [
    new DocumentLoadInstrumentation(),
    new UserInteractionInstrumentation(),
  ],
  tracerProvider: provider,
});

// index.js
import "./setupTelemetry"

```

Events would get logged only once in console by instrumentation-user-interaction

### What did you see instead?
Each click event is registered twice by instrumentation-user-interaction

### Additional context

https://user-images.githubusercontent.com/1942688/123246534-85b85380-d503-11eb-9f83-d281a0f74314.mov

bradfrosty:

Fixes #548 

- Deduplicate interaction spans when there are multiple event listeners for a single target

- Create a new zone that a single interaction will share with other handlers
- Capture all handlers in a 50ms window
- Fix task tracking logic for this case
- Update tests to flush macro tasks when necessary instead of a ticking a specific time

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
@@ -62,11 +62,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
   >();
 
   // Stores the zone that is being used to trace the current interaction
-  // Allows for multiple event handlers, even with different targets, to be collated into a single trace.
-  private _currentInteractionZone = new WeakMap<
-    EventTarget,
-    Zone
-  >();
+  private _currentInteractionZone?: Zone;
 
   constructor(config?: InstrumentationConfig) {
     super('@opentelemetry/instrumentation-user-interaction', VERSION, config);
@@ -144,7 +140,9 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
           : undefined
       );
 
-      this._spansData.set(span, { taskCount: 0, interactionTarget: element });
+      this._spansData.set(span, {
+        taskCount: 0,
+      });
 
       return span;
     } catch (e) {
@@ -161,7 +159,10 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
   private _decrementTask(span: api.Span) {
     const spanData = this._spansData.get(span);
     if (spanData) {
-      spanData.taskCount--;
+      if (spanData.taskCount > 0) {
+        spanData.taskCount--;
+      }
+
       if (spanData.taskCount === 0) {
         this._tryToEndSpan(span);
       }
@@ -196,10 +197,10 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
     spanData.taskCount++;
 
     // If a span completion timeout was scheduled, cancel it. There are more tasks to complete.
-    if (typeof spanData.taskCompleteTimeout === 'number') {
+    if (typeof spanData.interactionCompleteTimeout === 'number') {
       Zone.root.run(() => {
-        window.clearTimeout(spanData.taskCompleteTimeout);
-        spanData.taskCompleteTimeout = undefined;
+        window.clearTimeout(spanData.interactionCompleteTimeout);
+        spanData.interactionCompleteTimeout = undefined;
       });
     }
   }
@@ -432,19 +433,14 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
         this: Zone,
         task: AsyncTask
       ) {
-        let currentZone = Zone.current;
-
-        if (plugin._getCurrentSpan(currentZone)) {
-          task._zone = currentZone;
-        } else if (plugin._getCurrentSpan(task.zone)) {
-          currentZone = task.zone;
+        try {
+          return original.call(this, task) as T;
+        } finally {
+          const currentSpan = plugin._getCurrentSpan(Zone.current);
+          if (plugin._shouldCountTask(task) && currentSpan) {
+            plugin._decrementTask(currentSpan);
+          }
         }
-
-        const currentSpan = plugin._getCurrentSpan(currentZone);
-        if (currentSpan && plugin._shouldCountTask(task)) {
-          plugin._decrementTask(currentSpan);
-        }
-        return original.call(currentZone, task) as T;
       };
     };
   }
@@ -462,18 +458,15 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
         this: Zone,
         task: AsyncTask
       ) {
-        const currentZone = Zone.current;
-        const currentSpan = plugin._getCurrentSpan(currentZone);
-
-        if (currentSpan) {
-          task._zone = currentZone;
+        try {
+          return original.call(this, task) as T;
+        } finally {
+          const currentSpan = plugin._getCurrentSpan(Zone.current);
+          if (plugin._shouldCountTask(task) && currentSpan) {
+            plugin._incrementTask(currentSpan);
+            plugin._checkForTimeout(task, currentSpan);
+          }
         }
-
-        if (currentSpan && plugin._shouldCountTask(task)) {
-          plugin._incrementTask(currentSpan);
-          plugin._checkForTimeout(task, currentSpan);
-        }
-        return original.call(currentZone, task) as T;
       };
     };
   }
@@ -497,17 +490,17 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
         ? applyArgs[0]
         : undefined;
 
-        const target = event?.target;
+        const target = plugin._allowEventType(task.eventName) && event?.target;
         let span: api.Span | undefined;
 
-        if (target && plugin._allowEventType(task.eventName)) {
-          if (!plugin._currentInteractionZone.has(target)) {
+        if (target) {
+          if (!plugin._currentInteractionZone) {
             // create new span for interaction from current task zone
             span = plugin._createSpan(target, task.eventName);
 
             if (span) {
               api.context.with(api.trace.setSpan(api.context.active(), span), () => {
-                plugin._currentInteractionZone.set(target, Zone.current);
+                plugin._currentInteractionZone = Zone.current;
               });
 
               // Reset the current interaction zone after 50ms
@@ -515,18 +508,18 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
               const spanData = plugin._spansData.get(span);
               if (spanData) {
                 Zone.root.run(() => {
-                  spanData.interactionCompleteTimeout = window.setTimeout(() => {
-                    plugin._currentInteractionZone.delete(target);
+                  spanData.interactionResetTimeout = window.setTimeout(() => {
+                    plugin._currentInteractionZone = undefined;
+                    spanData.interactionResetTimeout = undefined;
                   }, RESET_ZONE_DELAY_MS);
                 });
               }
             }
           }
 
-          const currentInteractionZone = plugin._currentInteractionZone.get(target);
-          if (currentInteractionZone) {
+          if (plugin._currentInteractionZone) {
             // Override handler task zone with current interaction zone
-            task._zone = currentInteractionZone;
+            task._zone = plugin._currentInteractionZone;
             span = plugin._getCurrentSpan(task.zone);
           }
 
@@ -537,22 +530,13 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
           span = plugin._getCurrentSpan(task.zone);
         }
 
-        return task.zone.run(() => {
-          try {
-            return original.call(task.zone, task, applyThis, applyArgs);
-          } finally {
-            // If target is defined, task is decremented after running event handler.
-            // If _shouldCountTask() returns true, task is decremented after running other macro or micro tasks.
-            if (target || plugin._shouldCountTask(task)) {
-              if (span) {
-                plugin._decrementTask(span!);
-                // Ensure that repeat runs of the click handler (e.g. another click)
-                // are starting from the parent zone, since we started a new task from that parent.
-                task._zone = Zone.current.parent ?? Zone.root;
-              }
-            }
+        try {
+          return original.call(task.zone, task, applyThis, applyArgs);
+        } finally {
+          if (span && (target || plugin._shouldCountTask(task))) {
+            plugin._decrementTask(span);
           }
-        });
+        }
       };
     };
   }
@@ -564,7 +548,16 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
    * @private
    */
   private _shouldCountTask(task: AsyncTask): boolean {
-    if (!task.data || task.data.isPeriodic) {
+    if (!task.zone || !task.data || task.data.isPeriodic) {
+      return false;
+    }
+
+    const currentSpan = this._getCurrentSpan(task.zone);
+    if (!currentSpan) {
+      return false;
+    }
+
+    if (!this._spansData.get(currentSpan)) {
       return false;
     }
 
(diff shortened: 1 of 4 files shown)


crellison (approved):

obecny (changes requested):
I think you should find a way to fix it in a way to prevent creating 2 spans from one click instead of combining them into one extra span which sounds and looks very very wrong. For me this is blocker

dyladan:
> I think you should find a way to fix it in a way to prevent creating 2 spans from one click instead of combining them into one extra span which sounds and looks very very wrong. For me this is blocker

If a click event has multiple handlers, those handlers should be individual spans and you need to have a parent for them to be properly related. Imagine a scenario where there is a button with 2 handlers. One handler makes an API call and updates the page, and the other handler sends analytics data to your analytics service. In the current state, these are two separate traces which is obviously wrong. This PR suggests to make a "click" span, with a child span for each handler. You are suggesting the API call and the analytics should be a single span?

obecny:
> > I think you should find a way to fix it in a way to prevent creating 2 spans from one click instead of combining them into one extra span which sounds and looks very very wrong. For me this is blocker
> 
> If a click event has multiple handlers, those handlers should be individual spans and you need to have a parent for them to be properly related. Imagine a scenario where there is a button with 2 handlers. One handler makes an API call and updates the page, and the other handler sends analytics data to your analytics service. In the current state, these are two separate traces which is obviously wrong. This PR suggests to make a "click" span, with a child span for each handler. You are suggesting the API call and the analytics should be a single span?

This is a different case than this and what we discussed about
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/548#issuecomment-969116858

bradfrosty:
> This seems like a really unstable way to do this. How can you be sure you aren't going to grab unrelated handlers?

In practice, I think this timeout will rarely happen, if at all. It should be safe to remove it. Each task will run to completion and schedule a macro task (0ms timeout) to capture the next event handler. The other event handler tasks will now run before that timeout task, since they were scheduled before it (thus capturing them). Each time a new span is created, all subsequent event tasks that are click events will be considered part of that span. To increase confidence, we can compare event targets. Since it's a task queue, I believe this is safe even when it happens very quickly (e.g. user click, listeners run, user clicks, listeners run).

> I think you should find a way to fix it in a way to prevent creating 2 spans from one click instead of combining them into one extra span which sounds and looks very very wrong. For me this is blocker

This changeset does not combine them into one extra span. It wraps all interactions (e.g. deduplicates them) into a single span.

[I have integrated this changeset into an example here](https://stackblitz.com/edit/opentelemetry-js-interaction-with-gtm-akeucp?devtoolsheight=33&file=src/tracer.ts) and added verbose logging. Please look at this and compare the patch vs. the current version of this instrumentation (uncomment the instrumentations as needed in `tracer.ts`). 

As you will see, the current version creates **six listeners**, compared to the patch which creates **one**. The application only registers 4 listeners. The other two are from a third party script on the page (gtm.js). If you inspect the logs carefully, you can observe the flow through each patch of the zone.js methods (e.g. runTask, scheduleTask, cancelTask). It's been somewhat difficult to follow the flow through these callstacks, I found logs to be helpful when there are many tasks to run. Hopefully this explains the solution a bit more... 

Furthermore, please look at the test updates. It sets up two click handlers on the same target, and asserts that only one span is created.

> If a click event has multiple handlers, those handlers should be individual spans and you need to have a parent for them to be properly related. Imagine a scenario where there is a button with 2 handlers. One handler makes an API call and updates the page, and the other handler sends analytics data to your analytics service. In the current state, these are two separate traces which is obviously wrong. This PR suggests to make a "click" span, with a child span for each handler. You are suggesting the API call and the analytics should be a single span?

In this PR, there is only one span per interaction. The individual handlers do not create individual spans. All downstream spans from each event handler share this "root interaction span". In your example, there is no need for the span redundancy unless very granular span durations are needed on individual spans. The network (fetch or xhr) spans that would be created from those two requests (API call+update page and analytics request) would be properly nested. **This means that handlers that only interact with the client (e.g. just a page update) will not contain a new child span, but their duration is still captured in the "user interaction"**. However, it is possible to add this level of span granularity.

@obecny to answer your question, this is a bug fix to address (1). I am not suggesting (2). This instrumentation is called "user-interaction", but the spans being captured are "user-interaction-listeners". The extra spans introduce a lot of noise into span data, making it very difficult or impossible to analyze data.

bradfrosty:
@obecny @dyladan I have added new tests for the 50ms window logic, and enhanced its ability to handle concurrency. It is now grouped by the event target. If two clicks occur at the same time on different targets, it will still create two unique spans.

The 50ms timer is necessary for the case when a handler is long running. After 50ms, the instrumentation will clear the "current interaction", allowing a subsequent click to start a new span. This doesn't impact the original span, since it's carried on the task's zone.

bradfrosty:
I have removed the additional concurrency guard. In practice, this end up creating unwanted span duplicates. For our use case, we prefer there to be a single span per interaction and prefer to prevent span duplicates as much as possible. So I'm bringing my updated implementation back here. 

Any thoughts @dyladan @obecny? I'd prefer to not maintain my own fork of this instrumentation, and without this fix it is unusable for us. Given this issue was reported by another user it seems I'm not alone in thinking that duplicate spans are not desired. I spent a good chunk of time looking into this and would appreciate some feedback.

bradfrosty:
Hey @dyladan @obecny, any updates? This has been working well for us in production.

dyladan:
sorry for the unacceptably long period of non-response. @obecny recently stepped down as a maintainer and he the primary maintainer of the browser instrumentations so this slipped through the cracks a little on me. Please bear with me a bit since I'm not a react expert and I'm still working to understand the problem and your proposed solution, so let me know if any of this is incorrect:

This is a react-specific issue. This is caused by the react event delegation system. React creates an event listener for each event type on the root node. When a developer adds an event to a node like `<button onClick={handleClick}>`, react actually doesn't register a handler on that node. When a user clicks that button, the click event listener installed by react checks to see if the button clicked has a registered event listener in react, and calls it if appropriate. The duplication happens because we are tracing react's event listener as a span, but we are also tracing the callback registered by the user as a span.

Please let me know if anything in the above paragraph is incorrect, as it is crucial to fully understand the problem before I can evaluate a solution.

The proposed solution is to treat all click events that start within a 50ms window as the same click event, so if multiple click event listeners are called during this window, the span is started when the first event listener starts and ended when the last event listener ends. The underlying assumption is that there will never be two legitimate clicks within 50ms of each other, and that if two click event handlers are called, it must be because the first one has delegated the event to the second one.

Please let me know if anything in the above paragraph is incorrect.

obecny:
last thing have you considered using / creating a different context manager specific for react to handle those things differently ?

dyladan:
>Fixing this by grouping interactions within 50ms it doesn't really seems right for me. It might have different consequences and potential bugs once this is done and then we will have gordian knot.

I generally agree here. The underlying assumption that no clicks can happen within 50ms of each other seems like something that works 99.9% of the time but is very hard to debug when it actually breaks something.

I wonder if there is some way we can detect that an event handler is the react root node event handler and ignore it?

> last thing have you considered using / creating a different context manager specific for react to handle those things differently ?

I don't think the context manager is the correct place to fix this bug. It is not a context loss.

johnbley:
> I wonder if there is some way we can detect that an event handler is the react root node event handler and ignore it?

Not sure if this is what you meant, but I think that having react-specific instrumentation (rather than react-targeted heuristics) is the answer here.  If that instrumentation gets substantially more complex/different than this `user-interaction` code then we can look at splitting the react stuff into a separate plugin (and possibly altering/sharing code).

bradfrosty:
@dyladan this is in response to [your comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/743#issuecomment-1006926510) asking for clarification.

> This is a react-specific issue. This is caused by the react event delegation system. React creates an event listener for each event type on the root node. When a developer adds an event to a node like <button onClick={handleClick}>, react actually doesn't register a handler on that node. When a user clicks that button, the click event listener installed by react on the root node is called. That listener checks to see if the button clicked has a registered event listener in react, and calls it if appropriate. The duplication happens because we are tracing react's event listener as a span, but we are also tracing the callback registered by the developer as a span.

This is _not_ a react-specific issue. If you look at my example in the [previous comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/743#issuecomment-972462347), you'll see no React in use. It's all vanilla JS and `addEventListener`. 

The issue is applicable to all listeners that _share the same target_. [The example](https://stackblitz.com/edit/opentelemetry-js-interaction-with-gtm-akeucp?file=src/app.ts) shows some possible scenarios: 
- Scenario 1: Application has four independent listeners on a simple `button`. If the user is not leveraging some form of event delegation, this will generate four spans since each listener is uniquely registered. 
- Scenario 2: User has GTM installed, which likely loads GA. These scripts combined will register even more event listeners on all clicks. For example, clicking anywhere on the page will generate four independent spans. If you click on the button used in Scenario 1, you actually get **8 independent spans**. 

If a solution using the context manager makes more sense, I'm all for that. However, we mostly care about having usable data right now. Generating this number of spans makes the data from the current state of the user interaction instrumentation uninterpretable. Since they all share the same attributes (e.g. same xpath), there's nothing to correlate in our tracing backend. This is another negative impact of unique-per-listener solution, since it's possible the tracing backend will "mask" a slow performing listener with another one that is more performant since they share the same xpath. While this solution is heuristical, we have already found it immensely valuable since we can more effectively identify slow/erroring listeners. Knowing that something is performing slowly or erroring on a single event target significantly eliminates the number of places we need to look. It's a tradeoff in the end, I guess.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -179,8 +189,19 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
+      return;
+    }
+
+    spanData.taskCount++;
+
+    // If a span completion timeout was scheduled, cancel it. There are more tasks to complete.
+    if (typeof spanData.interactionCompleteTimeout === 'number') {
+      Zone.root.run(() => {
+        window.clearTimeout(spanData.interactionCompleteTimeout);
+        spanData.interactionCompleteTimeout = undefined;
```
dyladan, line 203: I think this is run in the root zone in order to exempt it from task count (because `_shouldCountTask` will return `false` because the root zone should never have a span?). I think there should be a comment to make that clear because otherwise this could be easily broken by someone in the future.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -464,41 +489,54 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
+            // create new span for interaction from current task zone
+            span = plugin._createSpan(target, task.eventName);
+
+            if (span) {
+              api.context.with(
+                api.trace.setSpan(api.context.active(), span),
+                () => {
+                  plugin._currentInteractionZone = Zone.current;
+                }
+              );
```
dyladan, line 507: Why is this required to be run in a context? Does it have something to do with `Zone.current`?

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -464,41 +489,54 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
...
+
+              // Reset the current interaction zone after 50ms
+              // This allows new clicks to create spans when the previous interaction is still not complete
+              const spanData = plugin._spansData.get(span);
+              if (spanData) {
+                Zone.root.run(() => {
```
dyladan, line 513: Same here. Please leave a comment why this needs to be run in the root zone

dyladan (approved):
Overall the implementation LGTM and I think I am convinced by the arguments you made for including this change.

stephencranedesign:
@bradfrosty 

Thanks for doing this! I just recently started looking into this package but walked away from it initially b/c I saw that a single click was always causing multiple traces in my react app. Will definitely use this once it is released.

stephencranedesign:
@dyladan / @bradfrosty 

Anything I can do to help here? I'd like to see this merged.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
bradfrosty: Yeah, that's the right idea. This is ensuring that instrumentation itself isn't tricked into thinking that tasks scheduled/executed/cancelled by the instrumentation _within the context of a click listener task_ should be counted. The task count stored on `spanData` is always in reference to the tasks scheduled/executed/cancelled in userland, so we make sure not to muddle that when registering async tasks in the instrumentation.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
bradfrosty: Your question is valid, it doesn't really need to run in context. I believe I did this to create a separate Zone from the listener task's Zone since the goal here is to share a Zone across multiple tasks that don't may not have common parent. 

At the time it felt semantically off to state the first registered listener's zone should be considered the zone for all other registered listeners. But in reality, it doesn't really matter since this execution context is never really used.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
bradfrosty: This is the same reasoning as above, but I'll make sure to mark all locations where the plugin registers an async task since by rule these run in the same execution context. Without doing this you're going to get incorrect taskCount state values.

bradfrosty:
@stephencranedesign of course, thanks for the appreciation! 😄  I also initially avoided this instrumentation (for over a year!) before finally taking a crack at it. But once I saw I realized its potential with more reasonable data, I decided the time investment to understand zone.js was worth it. 

We've been using this patch in production at Squarespace since November now. Funnily enough, within six hours of release our OTel instrumentation automatically detected a frontend outage (runtime error) for the first time! Without interaction tracing, you only get network tracing automatically (plus a small amount of non-network coverage via document load). Errors reported in a network trace on the frontend will typically be reported by backend code as well, so interaction tracing can really add immense value.

bradfrosty:
@dyladan @stephencranedesign I should be able to get the change in this week. I was thinking it might be useful to add more configuration, in the case that consumers want to tweak the heuristic settings or use the existing behavior today.

Something like:

```ts
interface UserInteractionInstrumentationConfig {
  // ...existing configuration
  joinInteractionSpans: boolean; // whether or not to attempt grouping similar listeners into a single span
  joinInteractionWindowMs: number; // time window after which similar listeners won't be grouped
}
```

Additionally, I would like to automatically record exceptions that occur in the interaction span's context.

stephencranedesign:
> I would like to automatically record exceptions that occur in the interaction span's context

that sounds reasonable & helpful.

> I was thinking it might be useful to add more configuration, in the case that consumers want to tweak the heuristic settings or use the existing behavior today

Based on the convo in this pr and my experience trying to use this plugin, it seems like this new implementation is a bug fix and not a new feature to opt into. So I'm not sure I'm on board with providing configuration to turn the bug fix off.

As far as configuring the ms delay, I'm leaning toward don't provide the option until someone asks for it.

That's my two cents :)

stephencranedesign:
@bradfrosty

Any updates here? Are you waiting on feedback from @dyladan or the javascript-maintainers?

bradfrosty:
@stephencranedesign yep. I think just feedback on the configuration is all we need.

While waiting, I've been thinking about the issue in general of why we need the heuristic in the first place. I think I've found a different approach to the implementation that doesn't patch zone.js, and leverages the public Zone API. It's almost a total rewrite since the approach is so different.

I'm still developing it, but I think it would enable accurate task tracking across async event listeners that are bound to the same EventTarget. It works by creating a new ZoneSpec (called UserInteractionZoneSpec) that contains task tracking state internally. When an event listener is invoked, the instrumentation checks if an active instance of this ZoneSpec exists for a given target xpath and event type, then runs the listener in that context. If it doesn't exist, then it forks that UserInteractionZoneSpec and clears the initial state (as it is the first listener to run), then stores this as the "active zone" for that target xpath/event type.

By running the subsequent listeners in the same zone, they will be able to share the task tracking state within the Zone. Thus we are able to detect when all async tasks complete across the listeners, and invoke a final callback `flush()` that will end the span created for this Zone instance. 

While we wait, I'm going to post a second PR demonstrating this solution since it might address all of concerns of this solution. It also simplifies the solution significantly (so far).

nvolker:
Really looking forward to having this merged. Great work here!

bradfrosty:
I have put up a PR with the alternative solution I discussed above (#980). It is still a work in progress but mostly feature complete. It needs additional test cases, verification of accuracy by hand with a timeline trace, and potentially some additional aggregates that would be useful for analyzing interaction traces. But as of right now it mostly replicates the existing behavior.

PR #747, opened 2021-11-16
feat(instrumentation-graphql): change execute and parse span names and attributes

### LINKED ISSUE #534: Add Resolver field name to graphql instrumentation span name
No issue

When looking at the spans using instrumentation-graphql, it would be helpful to see the field name in the span name when viewing, instead of having to drill down into attributes.  Today it just shows `graphql.resolve`.  If we could add the field name to this, it would be much more helpful.  Ex:  `graphql.resolve (field1)`

None

### LINKED ISSUE #705: Graphql should add operation name and hash in span

Currently the name of the operation and the hash are not provided in the attributes of the span. It just prints `query`, which is not even the operation name but the operation type. This makes it very difficult to find a given operation in all the traces.

By default the operation name should be provided as `operation.name` and the current `name` should be renamed to `type`.
Ideally the hash would also be provided. Getting a consistant is however difficult (each server implementation is different), so providing a hook for customization would be best with a sensible default.

I considered using `envelop` instead and skip the plugin from opentelemetry, but this requires significant changes in the code.

### Additional context

Screen Shot 2021-10-18 at 10 54 31 AM

mentos1386:

- Fixes partially #705 (without the `hash` request)
- Fixes #534

- Execute span names changed from `graphq.execute` to `graphql.execute.query`, where `query` represents an operation type.
- Prepare span names changed from `graphql.prepare` to `graphql.prepare.user` wher `user` is field name.
- Renamed `graphql.operation.name` to `graphql.operation.type` which represents `mutation`,`query` or `subscription`.
- Added `graphql.operation.name` which representes `Query1` as operation name.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
@@ -53,3 +53,20 @@ export enum SpanNames {
   SCHEMA_VALIDATE = 'graphql.validateSchema',
   SCHEMA_PARSE = 'graphql.parseSchema',
 }
+
+export const createResolveSpanName = (fieldName: string): string => {
+  return `${SpanNames.RESOLVE}.${fieldName}`;
+};
+
+export const createExecuteSpanName = (
+  operationType?: string,
+  operationName?: string
+): string => {
+  if (operationType) {
+    return `${SpanNames.EXECUTE}.${operationType}`;
+  }
+  if (operationType && operationName) {
+    return `${SpanNames.EXECUTE}.${operationType}.${operationName}`;
+  }
+  return `${SpanNames.EXECUTE}`;
+};
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
@@ -53,20 +53,3 @@ export enum SpanNames {
   SCHEMA_VALIDATE = 'graphql.validateSchema',
   SCHEMA_PARSE = 'graphql.parseSchema',
 }
-
-export const createResolveSpanName = (fieldName: string): string => {
-  return `${SpanNames.RESOLVE}.${fieldName}`;
-};
-
-export const createExecuteSpanName = (
-  operationType?: string,
-  operationName?: string
-): string => {
-  if (operationType) {
-    return `${SpanNames.EXECUTE}.${operationType}`;
-  }
-  if (operationType && operationName) {
-    return `${SpanNames.EXECUTE}.${operationType}.${operationName}`;
-  }
-  return `${SpanNames.EXECUTE}`;
-};
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
@@ -19,7 +19,8 @@ export enum AttributeNames {
   FIELD_NAME = 'graphql.field.name',
   FIELD_PATH = 'graphql.field.path',
   FIELD_TYPE = 'graphql.field.type',
-  OPERATION = 'graphql.operation.name',
+  OPERATION_TYPE = 'graphql.operation.type',
+  OPERATION_NAME = 'graphql.operation.name',
   VARIABLES = 'graphql.variables.',
   ERROR_VALIDATION_NAME = 'graphql.validation.error',
 }
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
@@ -393,7 +393,7 @@ export class GraphQLInstrumentation extends InstrumentationBase {
   private getOperationType(
     operation: graphqlTypes.DefinitionNode | undefined
   ): string | undefined {
-    if (operation && operation.kind == 'OperationDefinition') {
+    if (operation?.kind === 'OperationDefinition') {
       return operation.operation;
     }
 
@@ -404,15 +404,13 @@ export class GraphQLInstrumentation extends InstrumentationBase {
     operation: graphqlTypes.DefinitionNode | undefined,
     processedArgs: graphqlTypes.ExecutionArgs
   ): string | undefined {
-    if operation?.kind === 'OperationDefinition') {
+    if (operation?.kind === 'OperationDefinition') {
       if (operation.name) {
         return operation.name.value;
       }
     }
-    if (processedArgs.operationName) {
-      return processedArgs.operationName;
-    }
-    return undefined;
+
+    return processedArgs.operationName ?? undefined;
   }
 
   private _createExecuteSpan(
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
@@ -25,7 +25,7 @@ import {
 } from '@opentelemetry/instrumentation';
 import type * as graphqlTypes from 'graphql';
 import { GraphQLFieldResolver } from 'graphql/type/definition';
-import { SpanNames } from './enum';
+import { createExecuteSpanName, SpanNames } from './enum';
 import { AttributeNames } from './enums/AttributeNames';
 import { OTEL_GRAPHQL_DATA_SYMBOL } from './symbols';
 
@@ -39,7 +39,6 @@ import {
   GraphQLInstrumentationParsedConfig,
   OtelExecutionArgs,
   ObjectWithGraphQLData,
-  OPERATION_NOT_SUPPORTED,
   Maybe,
 } from './types';
 import {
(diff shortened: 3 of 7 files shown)


plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -390,29 +389,52 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+      span.setAttribute(AttributeNames.OPERATION_TYPE, operationType);
+      span.updateName(createExecuteSpanName(operationType));
+    }
+
+    if (operationName) {
+      span.setAttribute(AttributeNames.OPERATION_NAME, operationName);
+      if (operationType) {
+        span.updateName(createExecuteSpanName(operationType, operationName));
       }
-      operationName = OPERATION_NOT_SUPPORTED.replace(
```
mentos1386, line 411: I have removed this, as i don't believe it's needed to be formatted this way. Maybe we should rather add an attribute that would represent that the operation is "not supported"?

plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
```diff
@@ -53,3 +53,20 @@ export enum SpanNames {
   SCHEMA_VALIDATE = 'graphql.validateSchema',
   SCHEMA_PARSE = 'graphql.parseSchema',
 }
+
+export const createResolveSpanName = (fieldName: string): string => {
```
dyladan, line 57: this file is called enum and these are not enums. maybe rename the file to something more meaningful?

plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
```diff
@@ -53,3 +53,20 @@ export enum SpanNames {
...
+  return `${SpanNames.RESOLVE}.${fieldName}`;
+};
+
+export const createExecuteSpanName = (
+  operationType?: string,
+  operationName?: string
```
dyladan, line 63: Is name infinite cardinality or limited? The spec asks span names to have relatively low cardinality

plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
```diff
@@ -19,7 +19,8 @@ export enum AttributeNames {
   FIELD_NAME = 'graphql.field.name',
   FIELD_PATH = 'graphql.field.path',
   FIELD_TYPE = 'graphql.field.type',
-  OPERATION = 'graphql.operation.name',
+  OPERATION_TYPE = 'graphql.operation.type',
```
dyladan, line 22: Is this specified?

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -390,29 +389,52 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
-        span.setAttribute(AttributeNames.OPERATION, name);
-      }
-    } else {
-      let operationName = ' ';
-      if (processedArgs.operationName) {
-        operationName = ` "${processedArgs.operationName}" `;
+    const operationType = this.getOperationType(operation);
+    const operationName = this.getOperationName(operation, processedArgs);
+
+    const span = this.tracer.startSpan(createExecuteSpanName());
```
dyladan, line 426: Generally we like to avoid changing the name of a span after creation if at all possible. You can avoid updating the name if you just do this here.

```suggestion
    const span = this.tracer.startSpan(createExecuteSpanName(operationType, operationName));
```

`createExecuteSpanName` already gracefully handles missing type and name so it isn't a problem.

This would also require updating `createExecuteSpanName` or `getOperationType` to match null/undefined types but that's easy enough.

plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
mentos1386: I don't understand what you mean @dyladan.

plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
mentos1386: It's infinite, as clients can set anything as operation name.[1] I have removed it from execute span name.

[1] https://graphql.org/learn/queries/#operation-name

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
mentos1386: I have used your approach, decided on returning `undefined` as I've seen it being passed around.

plugins/node/opentelemetry-instrumentation-graphql/src/enum.ts
mentos1386: Moved to `utils.ts`.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -390,29 +390,48 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+    }
+
+    return undefined;
+  }
+
+  private getOperationName(
+    operation: graphqlTypes.DefinitionNode | undefined,
+    processedArgs: graphqlTypes.ExecutionArgs
+  ): string | undefined {
+    if (operation && operation.kind == 'OperationDefinition') {
```
obecny, line 407: please use es6 and strict comparison
```suggestion
    if operation?.kind === 'OperationDefinition') {
```

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
obecny: why do you think so ?
currently it will help to identify the operation name that is not supported, after yr change it will not be possible, so I see here a regression instead of improvement.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -390,29 +390,48 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+      if (operation.name) {
+        return operation.name.value;
+      }
+    }
+    if (processedArgs.operationName) {
```
obecny, line 412: I dont think you need that if you return undefined anyway so if operationName is undefined it will return undefined anyway
```suggestion
    return processedArgs.operationName;
```

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
mentos1386: That is true, but making `operationName` to be `Operation ${operationName} not supported` breaks what the "opeationName" field is supposed to be.

I would stick with "opertationName" to represent the name so the value is more predictable. Especially if you want to use it with some automation, where with current implementation you would have to parse the string to get out the actual operation name.

Maybe adding `operationUnsupported: true` or something in this sense, to indicate that operation is not supported?

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
mentos1386: Had to use `?? undefined` to make sure to return undefined, otherwise typescript complain about returning null.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -433,3 +433,16 @@ async function safeExecuteInTheMiddleAsync<T>(
...
+};
+
+export const createExecuteSpanName = (
+  operationType: string | undefined
+): string => {
+  if (operationType) {
+    return `${SpanNames.EXECUTE}.${operationType}`;
+  }
+  return `${SpanNames.EXECUTE}`;
+};
```
spencerwilson, line 448: What's your motivation in adding high-cardinality (field name) and unbounded-cardinality (operation name) data to these spans' names? I'm concerned that this change will make queries on these span names harder, since these names would now have two jobs: convey the span "type" + convey another aspect of the span. E.g., it's harder to refer to all the "resolver" spans. The added data is present in other attributes; is that not sufficient? Each attribute describing a single characteristic of the span? If one wished to do a query akin to  how "GROUP BY name" would be under these changes, one would do "GROUP BY name, field_name".

spencerwilson: Apologies; just saw the linked issue. Commented there.

mentos1386: I believe it's adding more visibility (without looking at attributes) at what part of "graphql resolving" does the span belongs to. Same as for HTTP there isn't just "http.resolving" but it includes method (GET, POST etc.) and also the "route" such as `/api/users/{user_id}`.

https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name

plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
spencerwilson: @mentos1386 I suspect @dyladan is asking whether there are OpenTelemetry semantic conventions for GraphQL trace data. (see here for what that means if you're unsure: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/705#issuecomment-952998047)

If so, the answer is "no", but progress toward having some is occurring here: https://github.com/open-telemetry/opentelemetry-specification/issues/1670

mentos1386:
Any update on this @obecny? Should we rather wait for a spec to be written (@spencerwilson)?

plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
dyladan: Yes @spencerwilson is correct that's exactly what I meant. In general we hesitate to add/change attributes without semantic conventions, but in this case it seems like the spec is moving quite slowly.

obecny:
> Any update on this @obecny? Should we rather wait for a spec to be written first (@spencerwilson)?

Mine main concern is that this is still not spec'ed and it might be tricky later to make necessary changes. I don't really have strong opinion here, will simply defer this to the @open-telemetry/javascript-maintainers to decide about this PR

PR #752, opened 2021-11-22
fix: early close of span with graphql execute

### LINKED ISSUE #750: Premature closing of execute span in graphql instrumentation
In the case of execute result being a promise, and if there is no responseHook defined endSpan is called too early

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e5a65c67ec63e9235f7ddf82b2cd75a5a926f7c8/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts#L248-L255

This is already handled correctly if there is a responseHook defined

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e5a65c67ec63e9235f7ddf82b2cd75a5a926f7c8/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts#L257-L263

samriley:

Fixes #750

Await promise resolution before calling endSpan

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
@@ -245,20 +245,24 @@ export class GraphQLInstrumentation extends InstrumentationBase {
     result?: PromiseOrValue<graphqlTypes.ExecutionResult>
   ) {
     const config = this._getConfig();
-    if (
-      typeof config.responseHook !== 'function' ||
-      result === undefined ||
-      err
-    ) {
+    if (result === undefined || err) {
       endSpan(span, err);
       return;
     }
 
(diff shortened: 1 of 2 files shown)


dyladan (approved):
Looks good. Would be nice if you could add a test for this

samriley:
Cheers, I'll take a look tomorrow at getting a unit test written and also rebasing it with the latest main branch

samriley:
I've fixed the failing tests - it was due to a change in the execution order changing. I've added a test for the span timings, let me know if there are any changes you'd like - eg duplicating the test in to the different cases eg when depth is set to 0.

vmarchaud (approved):

obecny (approved):

vmarchaud:
@samriley could you rebase your PR please ? it should be fix the lint

PR #755, opened 2021-11-24
prefer forceFlush from provider

kubawach:
Improvement for opentelemetry-instrumentation-aws-lambda

Currently the instrumentation gets `forceFlush` handler directly from the `activeSpanProcessor` of the current `TracerProvider`. The active span processor can, however, be changed during lifetime of a tracer provider. In such situation the instrumentation will call wrong `forceFlush` method. Also, a tracer provider can have multiple registered span processors and we should be flushing all of them.

Base implementation of a TraceProvider fortunately has the `forceFlush` method implemented. Therefore, if available, we should call TP's implementation and let the provider flush all span processors.


willarmiros:
> Base implementation of a TraceProvider fortunately has the forceFlush method implemented

Does the implementation just call `forceFlush` on all its span processors?

@anuraaga does this change make sense to you or would it have a risk of changing the behavior?

kubawach:
For a broader context - current solution seems to cause bug in opentelemetry-lambda usage. If the instrumentation is initialised before proper SpanProcessor is set (which can happen due to async nature of node.js execution) the first span is sent to Noop.

anuraaga (approved):

kubawach:
@willarmiros @NathanielRN any chances of looking into this? I see that @anuraaga already approved.

NathanielRN (approved):
This looks good to me because as @kubawach said the `BasicTracerProvider` in `opentelemetry-sdk-trace-base` has a `forceFlush` method that will call all the SpanProccesors in turn and `forceFlush` them.

See it [at this link](https://github.com/open-telemetry/opentelemetry-js/blob/5e9c7e1cb6ab57b92a59def81de2562760671c97/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L153).

It also falls back to calling `forceFlush` on the `activeSpanProcessor` if the above fails.

@kubawach can you _add_ a test case that confirms this fall back behavior? i.e. `it('should use the activeSpanProcessor forceFlush if one is not available on the TracerProvider')`?

kubawach:
@NathanielRN sure, will add a test for that.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -241,6 +241,10 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
       currentProvider = currentProvider.getDelegate();
     }
 
+    if (typeof currentProvider.forceFlush === 'function') {
+      return currentProvider.forceFlush.bind(currentProvider);
+    }
+
     if (typeof currentProvider.getActiveSpanProcessor === 'function') {
```
anuraaga, line 248: I have a feeling that at the time of writing, there wasn't any forceFlush on the tracer provider. Now that there is, it seems like this fallback isn't actually needed anymore (and would make the code closer to what the spec is, which is that SDK TracerProvider has forceFlush)

kubawach: Yeah, I agree. Neither "forceFlush" nor fallback property "getActiveSpanProcessor" is in the API (TraceProvider TS type) so since we depend on the SDK implementation we can simplify things.

mhennoch (approved):

johnbley (approved):
LGTM, though it might warrant a comment on `setTracerProvider` to the extent that the param should be a `BasicTracerProvider` or otherwise implement `forceFlush` to get the desired behavior.

willarmiros (approved):
lgtm based on conversation

kubawach:
@open-telemetry/javascript-maintainers any chances of getting this merged?

dyladan (approved):

dyladan:
LGTM i'll merge after the test run

vmarchaud (approved):

PR #757, opened 2021-11-25
feat: Long Tasks instrumentation

t2t2:

This PR adds instrumentation for Long Tasks API ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Long_Tasks_API), [w3c](https://github.com/w3c/longtasks)), automatically generating spans when a task takes longer than 50ms (as of current long task API spec).

Instrumentation wise it's pretty simple - it creates a `PerformanceObserver` for `longtask` type (including any buffered = before instrumentation is started) and then creates spans dumping all of the data from [`PerformanceLongTaskTiming`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceLongTaskTiming) entries

Adding @mhennoch as possible component co-maintainer as we both handle rum at splunk and more likely at least one of us will be available for maintenance (eg Long Task API v2).

resolves open-telemetry/opentelemetry-js#1722


vmarchaud (approved):
lgtm

johnbley (approved):

mhennoch (approved):

t2t2:
Seems like there was an update to [node.js (14 -> 16) & npm (6 -> 8) versions in github actions runners about 4 hours ago](https://github.com/actions/virtual-environments/commit/d0f20ddfa317289a4875d0c6916b285e8ebf8ca3#diff-6be9f9ce03ba57ca01e4ae937d5c2bb04893520fdf391ea87eb46acbc43ecdc5L25-L38) causing https://github.com/open-telemetry/opentelemetry-js/issues/2093 to happen on lint task as it uses whatever node/npm is installed ([CI run log](https://github.com/open-telemetry/opentelemetry-js-contrib/runs/4393271539?check_suite_focus=true)) however adding `--legacy-peer-deps` to lerna bootstrap's npm install arguments causes [a separate eslint module resolution error](https://github.com/open-telemetry/opentelemetry-js-contrib/runs/4394381502?check_suite_focus=true)

dyladan:
OK. We should work on the latest anyway so i'll look into updating our dependencies to get it up and running. Thanks for looking into it.

For now, we can probably use the setup-js action that the unit tests use to get around this.

plugins/web/opentelemetry-instrumentation-long-task/src/instrumentation.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  private _createSpanFromEntry(entry: PerformanceLongTaskTiming) {
+    const span = this.tracer.startSpan(LONGTASK_PERFORMANCE_TYPE, {
+      startTime: hrTime(entry.startTime),
+    });
+    span.setAttribute('component', this.component);
+    span.setAttribute('longtask.name', entry.name);
+    span.setAttribute('longtask.entry_type', entry.entryType);
+    span.setAttribute('longtask.duration', entry.duration);
```
dyladan, line 75: I assume there is no spec for these attributes. Can we see if there is any guidance from the spec/TC on how we should handle custom or non-specified attributes? If these are ever specified in the future in a different format it could make us non-compliant.

dyladan (approved):
This seems like an awesome addition and it is very easy to follow since the API is so straightforward. I made a comment about the non-specified attributes but if we don't get any guidance from the TC i'm happy to move forward with this since it is 0.x anyways.

plugins/web/opentelemetry-instrumentation-long-task/src/instrumentation.ts
vmarchaud: @dyladan instrumentation are not stable yet so this shouldn't be considered a breaking change if in the future they are in the spec ?

dyladan: No but in general I think we should try to be consistent.

Issue #654, opened 2021-09-02
Error trying to run the react-load demo


dariov1988:
### What version of OpenTelemetry are you using?
0.25.0

### What version of Node are you using?
v14.16.0

### What did you do?
I made a clean clone of the project
go to: examples/react-load/react and follow the Readme instructions:
run: npm install
run: npm run docker:start (zipkin start ok)
run: npm run build (complains about opentelemetry/api but after install it the build finish successfully)
run: npm install serve -g
run npm start

### What did you expect to see?
The demo app running.

### What did you see instead?
The app do not load and show an error in the console:
`Uncaught TypeError: Cannot read property 'debug' of undefined`

### Additional context
<img width="1919" alt="error" src="https://user-images.githubusercontent.com/10393939/131904952-ff1cab1d-5bb9-46ae-919f-e00827461a6a.png">

obecny:
just checked, the example is broken, but even though some strange thing is happening with react, can reproduce this bug

jussikuosa:
I added this to my `web-tracer.js` setup as a workaround: 

```
import { diag, DiagConsoleLogger, DiagLogLevel } from '@opentelemetry/api';
// ...
const diagLogger = new DiagConsoleLogger();
diag.setLogger(diagLogger, DiagLogLevel.ALL);
// ...
BaseOpenTelemetryComponent.setTracer(serviceName);
BaseOpenTelemetryComponent.setLogger(diagLogger); // provider.logger);
```

nemoshlag:
Hi I want to contribute to otl, and this seems like a good first issue, you can assign this for me thanks

dyladan:
Fixed in #764

PR #764, opened 2021-11-29
fix(init-logger): init logger for react-load example

nemoshlag:
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/654

-logger wasn't initiate properly on react implementation for react-load example

-imported diag from opentelemetry/api and set logger by it


naseemkullah (approved):
Thanks! Good to go after adjusting the import as per suggestion.

dyladan (approved):

PR #765, opened 2021-11-29
fix(user-interaction): handle null listener in addEventListener

t2t2:

Though useless, `null` is [also a valid listener argument for `addEventListener`](https://dom.spec.whatwg.org/#event-listener-callback). Currently user-interaction instrumentation however throws an exception as it tries to use null as a key in WeakMap.

Reproduction: https://codesandbox.io/s/otel-user-interaction-event-listener-null-ut8ed?file=/src/listener-null.js

Example of a valid use case in the wild:

https://github.com/angular/components/blob/59002e1649123922df3532f4be78c485a73c5bc1/src/cdk/platform/features/passive-listeners.ts#L21

(while this code is in angular, instrumentation then uses zone.js patching instead of addEventListener)

Skip patching logic when event listener argument is null, updated typings so typescript would also catch it


johnbley (approved):
Nice catch and analysis, thanks!

vmarchaud (approved):

PR #767, opened 2021-11-30
feat(instrumentation-pg): add post-query custom attributes hook for spans

### LINKED ISSUE #363: Copy post-query custom attributes hook from pg plugin to pg instrumentation
this is original PR
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/321
implement the same logic in PgInstrumentation

Alon-Katz:

- There is currently no way to add custom attributes to Postgres spans before the query is resolved

- add postQueryHook config to PgInstrumentationConfig
- When provided safely use it to add custom attributes to Postgres spans
- Fixes #363


vmarchaud (approved):

vmarchaud:
@Alon-Katz You'll need to rebase your PR against main

Alon-Katz:
@vmarchaud Sorry for the delay, I have rebased against main.

Alon-Katz:
@vmarchaud Sorry for the delay, I have rebased against main.

vmarchaud:
@Rauno56 could you take the time to review this PR ?

Rauno56:
I apologize for missing this one, @Alon-Katz. 20 days is really an unacceptable delay for a review.

This is no doubt an useful feature. Not only for pg but potentially every other instrumentation.

What kind of custom attributes are you looking to add to the spans from pg, @Alon-Katz?
From the test cases in the PR I'm _guessing_ it's about doing additional processing on what's already in the attributes and trying to provide additional convenience filtering to the o11y suite you are using?

vmarchaud:
@Rauno56 Judging from the fact that the PR has been opened from https://github.com/epsagon (which is an apm product) i guess they want to have more metadata to display.

Rauno56:
Sure, I was asking to validate whether it could be done somewhere else in the pipeline. It's a functionality commonly asked for - it doesn't scale well to add post-hooks to each of the instrumentations.
Currently the span processor provides a `onEnd` hook to a `ReadableSpan`, which, is not sufficient in this case - we should either reconsider that or add something into the instrumentation base.

vmarchaud:
> we should either reconsider that or add something into the instrumentation base.

There has been multiple discussions across different instrumentations to do this on the instrument base, the problem is that you generally want instrumentation specific context so using SpanProcessor isn't viable. While no one took the work to add this kind of mechanism on instrumentation base its okay for me to let people add those hooks in each instrumentation

Rauno56:
What to you mean by instrumentation specific context? If I understand you correctly, I can only come up with

1. the configuration for the instrumentation;
2. information that's already on the span.

vmarchaud:
> What to you mean by instrumentation specific context

Not every information is available as span attributes, for example one might want to add remote server version but we don't have it by default. I don't think its scalable to try add all data by default and let people use span processors. Specially since vendor might want other kind of data that we didn't think of.
I'm not sure to see why it's important anyway ? We already have those hooks in other instrumentation (so it solves a problem) and while there is no system on the base instrumentation class we can still merge this PR to allow this kind of use case ?

Rauno56:
> Not every information is available as span attributes, for example one might want to add remote server version but we don't have it by default. I don't think its scalable to try add all data by default and let people use span processors. Specially since vendor might want other kind of data that we didn't think of.

We agree on that - There has to be a mechanism for vendors or users to add their own attributes and this PR solves for that in the context of `pg` instrumentation.

However, now that we've seen the same feature added into different instrumentations with somewhat different semantics or ergonomics again and again we should know better. I think merging this does more harm than good long term and support solving this in a more general manner instead.

Do you think there's an usecase we cannot support by making it a general feature?

vmarchaud:
> Do you think there's an usecase we cannot support by making it a general feature?

I agree that we need to make it a general feature, but i don't agree with refusing PRs because contributors don't have the time to make a generic system.

dyladan:
I think as a feature it makes sense. There will always be information that people want to add to the spans with additional context that is not available at the span processor level. @Rauno56 is the owner of this module and I think he should have the final decision. When we decided to give ownership of modules to the contributors who maintain them, the original idea was that they would have the final say in decisions like this unless there was a security concern or similar extenuating circumstance. My personal suggestion would be to accept the feature because there is currently no generic system and the feature provides value, but ultimately the decision is Rauno's.

Creating a generic system for this feature might be difficult because of all the different types of modules involved, but maybe we could at least make the configurations consistent across instrumentations that hook similar modules. For example, all the DB instrumentations would probably benefit from the same hooks with consistent naming. It might make sense for us to make our own document in this repo which describes a set of suggested hooks for each type of module?

Rauno56:
> I agree that we need to make it a general feature, but i don't agree with refusing PRs because contributors don't have the time to make a generic system.

Yes, that makes sense actually, @vmarchaud. My bad. @Alon-Katz, could you please enable ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) so we wouldn't be stuck with the branch being out of date with the base branch.

In conclusion, I'd love to merge this for now with the hope to replace it with something more general in the long run.

I do not understand, however, why do you find it difficult to make something more general?

dyladan:
> I do not understand, however, why do you find it difficult to make something more general?

Simply because of the wide variety of modules that need to be instrumented and the resulting variety of useful hook points and their arguments. Maybe its easier than I'm thinking.

vmarchaud:
> I do not understand, however, why do you find it difficult to make something more general?

I don't think there is a lot of hidden complixity in this (not saying its trivial though) but we all have limited time to allocate and this wasn't a priority in the past (specially with the GA of api/sdk). There was discussion for a more generic instrumentation API that might have specified how to handle generic configuration like this (https://github.com/open-telemetry/opentelemetry-js/issues/2359), so generally we might want to have a complete API for instrumentation before trying to implement this feature in the base instrumentation class

Alon-Katz:
> Yes, that makes sense actually, @vmarchaud. My bad. @Alon-Katz, could you please enable "Allow edits from maintainers" so we wouldn't be stuck with the branch being out of date with the base branch.

@Rauno56 I followed the link for "Allow edits from maintainers", but for some reason it seems like I do not have this option. According to the documentation,  the option should be in the right section below the notifications, this is how it looks on my side: 
Screen Shot 2021-12-28 at 13 49 16

Flarna:
@Alon-Katz Most likely the problem is because the repo you use is not your personal fork, it seems to be a fork from an organization. For such forks it's not working to allow maintainers to edit via this checkbox.

So either move to a private fork in future PRs or take the burden to merge master regularly...

I think @dyladan found some rest call from github to enable this for a single branch/PR even on company forks.

dyladan:
You can do it with cURL:

```
curl -X PATCH -u dyladan:$GITHUB_TOKEN -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/vnd.github.v3+json" -d "{\"maintainer_can_modify\": true}"  "https://api.github.com/repos/open-telemetry/opentelemetry-js-contrib/pulls/767"
```

Alon-Katz:
> You can do it with cURL:
> 
> ```
> curl -X PATCH -u $GITHUB_USER:$GITHUB_TOKEN -H "Accept: application/vnd.github.v3+json" -H "Content-Type: application/vnd.github.v3+json" -d "{\"maintainer_can_modify\": true}"  "https://api.github.com/repos/open-telemetry/opentelemetry-js-contrib/pulls/767"
> ```

Thanks! I ran the command, hopefully it works.

Rauno56:
> [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/255986659/767/#/?version=2)
> 
>     * white_check_mark  Alon Katz ([d6f8af0](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/d6f8af008163a72c8b590ae002a7c780aab5b866), [85d5611](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/85d561152a84e0e63b094f64ae7ab958130ba63b), [27a4050](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/27a405076c3bf2214a6d8d9f34227a73e8b8cde1))
> 
>     * white_check_mark  Rauno Viskus ([f724864](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/f724864f9ac08e07a85406937f37f1c809d6db4c))
> 
> 
>     * x The commit ([07924d6](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/07924d68c84e601d308ac19b4369c3c6d3271497)) is missing the User's ID, preventing the EasyCLA check. [Consult GitHub Help](https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user) to resolve.For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

@Alon-Katz seems like your CLA is signed for the Cisco email, but one of the commits(https://github.com/open-telemetry/opentelemetry-js-contrib/commit/07924d68c84e601d308ac19b4369c3c6d3271497) is done from an Epsagon email address. That's why EasyCLA is probably upset.

I'll review the PR on Monday.

plugins/node/opentelemetry-instrumentation-pg/src/types.ts
```diff
@@ -27,12 +27,28 @@ export interface PgInstrumentationExecutionResponseHook {
...
 
+export interface QueryContext {
+  span: api.Span;
+  query?: string;
+  config?: NormalizedQueryConfig;
+  params?: unknown[];
+}
+
+export interface PgPostQueryHookFunction {
+  (ctx: QueryContext): void;
```
blumamir, line 38: Can you please change the hook signature to follow the other hook example in this file:
```js
export interface PgInstrumentationExecutionResponseHook {
  (span: api.Span, responseInfo: PgResponseHookInformation): void;
}
```

First parameter should be the span.
Second parameter is called `Info` on the other hook. Could be nice in my opinion to call it `queryInfo` (to stay consistent with the existing name. 
From what I remember, this is also the convention in other instrumentations.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -563,5 +563,113 @@ describe('pg', () => {
...
+
+    it('should call postQueryHook with query text if set', async () => {
+      instrumentation.disable();
+      let called = false;
+      const query = 'SELECT NOW()';
+      const config: PgInstrumentationConfig = {
+        postQueryHook: ctx => {
+          called = true;
+          assert.strictEqual(ctx.query, query);
+          assert.strictEqual(ctx.params, undefined);
```
blumamir, line 575: These assertions are performed in the hook body.
I think that if they fail, the exception will just be silently ignored and test will pass.
That is because the hook is protected with safeExecuteInTheMiddle which does not propagate exceptions to the caller.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -563,5 +563,113 @@ describe('pg', () => {
...
+            values: values,
+          });
+          assert.strictEqual(resPromise.command, 'SELECT');
+          runCallbackTest(span, attributes, events);
+        } catch (e) {
+          assert.ok(false, e.message);
+        }
+      });
+      assert.strictEqual(called, true);
+    });
```
blumamir, line 673: I also suggest having some negative tests for the hook - what happens if it throws? what if it's not a function?
This introduced new logic is currently not covered with tests.
You can see an example of that on the [responseHook tests here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/81b3190af64bda14f87c5b0cbd6172bafda26408/plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts#L461)

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -563,5 +563,113 @@ describe('pg', () => {
         client.query('SELECT NOW()').then(queryHandler);
       });
     });
+
+    it('should call postQueryHook with query text if set', async () => {
```
blumamir, line 567: You can wrap this section with a `describe` to group all these tests under `postQueryHook`.
This will also be consistent with the existing tests for `responseHook` [see here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/81b3190af64bda14f87c5b0cbd6172bafda26408/plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts#L397)

blumamir:
Hi @Alon-Katz 
Still interested in this PR?

Alon-Katz:
Hi @blumamir, thanks for time you put into making the CR but unfortunately I will not have the time to continue this PR at the moment, I will close it.

PR #776, opened 2021-12-05
chore(ci): fix lint workflow

vmarchaud:
Github Actions team updated their base ubuntu to ship with Node 16 instead of 14 like before, so we got an npm update that obviously broke the dependency resolution and resulted in broken deps install. I didn't have the time to (nor the will ahah)  to make it work with node 16 so i just downgraded to 14 :)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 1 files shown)


vmarchaud:
ping @dyladan blocking all PRs in contrib so would be cool to review and merge this asap !

Flarna (approved):

legendecas (approved):

Issue #756, opened 2021-11-24
Run a daily `test-all-versions` run


Rauno56:
### Is your feature request related to a problem? Please describe

Some of the contrib packages have a way to run the tests against multiple versions. That's super awesome and useful! We don't run those in our CI though, namely because 

### Describe the solution you'd like to see

We should schedule a regular run of the full test suite with TAV for packages that have it. + maybe for release PRs.

### Describe alternatives you've considered

Run it in every PR, but that could slow PR-work down too much.

vmarchaud:
I'm not against running it in every PRs, it may takes time but avoid problems after the PR is merged. To avoid too much impact we could just run tests on package that have been modified by the PR (since contribs pkgs don't depend on each other, i don't tihnk that would be an issue)

haddasbronfman:
I'll start looking at this issue if it's ok

Rauno56:
That's configured in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/823

PR #778, opened 2021-12-07
feat(add-tav-action): add test-all-versions as github action

haddasbronfman:

- This PR is solving this issue: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/756

I added a new github action, which is using dorny/paths-filter to get all the changed packages, and for each such package, I run the command "npm run test-all-versions" - if it exist.


dyladan:
it seems like you're filtering only the packages which have changed using some other action? Why not use the built-in lerna filtering functionality?

dyladan:
Looks like the check fails

.github/workflows/test-all-versions.yml
```diff
@@ -0,0 +1,16 @@
...
+  tav:
+    name: run test-all-version (tav)
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - run: npm install test-all-versions
+      - run: lerna run test-all-versions --since origin/main
```
dyladan, line 16: nit
```suggestion
      - run: lerna run test-all-versions --since origin/main

```

dyladan (commented):
It looks like you don't install dependencies or build the packages. Are you sure it will work? Can you please make a small change to a package with a test-all-versions script so we can verify that it runs properly? Modifying the readme slightly should be enough.

dyladan:
you need to install dependencies and build the project. You can see how this is done in the unit test workflow but usually you have to do `npm install --ignore-scripts`, then `lerna bootstrap --since origin/main --include-dependencies`, then `lerna run compile --since origin/main --include-dependencies`. Then you should be able to do the test all versions script.

dyladan:
the `feat` identifier in the title signifies the minor version number to be incremented in our release automation. Please use a more appropriate conventional commits identifier like `test`

dyladan:
I'm not sure why it says mongo isn't started. @Rauno56 is there some additional needed step for the tav script?

Rauno56:
Some tests need an service to be running in order to work: see "services" in [unit-test.yml](./.github/workflows/unit-test.yml).

blumamir:
Thanks for adding this.
The current action will run only against packages that had changes in each PR.
This will do a good job of catching errors with old versions that are introduced in the current PR.

Instrumentations are patching the internal logic of packages that are out of our control. It means that if the original instrumented library author published a new version to npm which breaks something in the instrumentation, no automatic process will catch that until someone makes a change to the instrumentation packages, or when a user complains (probably after something breaks in his deployment due to the OTEL installation).

This had happened before with `aws-sdk`, which has a complex internal logic and tends to publish new versions often.

At the [opentelemetry-ext-js project](https://github.com/aspecto-io/opentelemetry-ext-js) we run test-all-versions on all packages every day with [github actions scheduled-events](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#scheduled-events), so if someone pushes a new version to npm which is not compatible with the current instrumentation, we get notified (in slack) after at most 24 hours.
Github action configuration for reference: https://github.com/aspecto-io/opentelemetry-ext-js/blob/master/.github/workflows/daily-test.yml
sidenote: the daily tests currently fails due to kafkajs, we need to look at it and fix it

vmarchaud:
> At the opentelemetry-ext-js project we run test-all-versions on all packages every day with github actions scheduled-events, so if someone pushes a new version to npm which is not compatible with the current instrumentation, we get notified (in slack) after at most 24 hours.

We could do both but i worry it may require too much bandwith to fix them when alert comes off (i guess it depend of component maintainers) resulting in maintainer generally ignoring errors

dyladan:
> > At the opentelemetry-ext-js project we run test-all-versions on all packages every day with github actions scheduled-events, so if someone pushes a new version to npm which is not compatible with the current instrumentation, we get notified (in slack) after at most 24 hours.
> 
> We could do both but i worry it may require too much bandwith to fix them when alert comes off (i guess it depend of component maintainers) resulting in maintainer generally ignoring errors

I think not knowing about errors is a worse outcome than having errors that go unfixed for a while. I think this comes down to trusting the component authors and giving them a chance to succeed first before assuming they will fail.

vmarchaud:
Lets go both then ?

vmarchaud (approved):

.github/workflows/test-all-versions.yml
```diff
@@ -0,0 +1,92 @@
...
+      CASSANDRA_PORT: 9042
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - run: npm install --ignore-scripts
+      - run: lerna bootstrap --since origin/main --include-dependencies
+      - run: lerna run compile --since origin/main --include-dependencies
+      - run: lerna run test-all-versions --since origin/main --stream --concurrency=1
```
dyladan, line 92: nit:
```suggestion
      - run: lerna run test-all-versions --since origin/main --stream --concurrency=1

```

dyladan (approved):

vmarchaud (changes requested):
waiting for the cron scheduled workflow to be added before approving, otherwise seems good

dyladan:
@vmarchaud this particular action runs only the changed packages on pull requests. I think the full all-package cron can be done in a follow-up PR. @haddasbronfman do you agree?

vmarchaud:
> @haddasbronfman do you agree?

We discussed this privately on slack, i suggested to make only one for both (since it's closely related) but if you believe its bettter to make two i don't mind

vmarchaud (approved):

dyladan:
> > @haddasbronfman do you agree?
> 
> We discussed this privately on slack, i suggested to make only one for both (since it's closely related) but if you believe its bettter to make two i don't mind

No I just didn't realize you had already discussed this. Go ahead with whatever you already decided.

vmarchaud:
@haddasbronfman As you prefer then :)

haddasbronfman:
Hi,
I prefer to do it on another PR is this is ok with you.

PR #780, opened 2021-12-08
test: use `127.0.0.1` in mongo tests as the default host

Rauno56:

This fixes tests for mongodb versions 3.3 - 3.4.

We're currently running TAV tests for 37 versions, 8 of which do not pass because versions before 3.5 do not resolve the DNS name. This bypasses that and makes the tests pass on all of the supported versions

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb.test.ts
@@ -32,7 +32,7 @@ const instrumentation = registerInstrumentationTesting(
 );
 
 import * as mongodb from 'mongodb';
-import { assertSpans, accessCollection } from './utils';
+import { assertSpans, accessCollection, DEFAULT_MONGO_HOST } from './utils';
 import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 
 describe('MongoDBInstrumentation', () => {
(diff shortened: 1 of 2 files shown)


dyladan:
Curious how long the TAV script takes to complete with 37 versions?

Rauno56:
2:49 on my local machine.

Rauno56:
I've narrowed down the list of versions on the packages I've added the TAV setup to. I think whether we should care about that or not depends on how we plan to run those tests. Since this instrumentation already had TAV set up, I didn't change that. Yet. Any thoughts?

Flarna (approved):

dyladan:
> I've narrowed down the list of versions on the packages I've added the TAV setup to. I think whether we should care about that or not depends on how we plan to run those tests. Since this instrumentation already had TAV set up, I didn't change that. Yet. Any thoughts?

Sorry i'm not sure I understand what you're saying.

Rauno56:
My bad. In other words:

1. Since mongodb instrumentation already had TAV setup, I didn't change the range of versions we test.
2. In other packages(`pg`, `winston`), where I've had to introduce TAV setup I have been more conservative with the amount of versions me test and kept it down to the most popular ones.
3. Ideally we'd have the same approach across the packages. However, whether we should adopt "test all supported versions" or "test a practical number of versions" depends on how(and when and how often) we plan to run TAV tests.

Latter is being decided on here: #778

dyladan (approved):

PR #793, opened 2021-12-21
fix: fastify and browser autoinjection failed to compile

dyladan:
Fastify: a type update in a dependency caused the instrumentation to fail to compile

browser autoinjection: mixed 0.26.x and 0.27.x core dependencies caused the package to fail to compile.


packages/opentelemetry-browser-extension-autoinjection/package.json
```diff
@@ -71,9 +71,9 @@
     "@opentelemetry/exporter-otlp-http": "0.26.0",
```
Flarna, line 71: This seems to be also outdated

Flarna (approved):

legendecas (approved):

vmarchaud (approved):

Issue #577, opened 2021-07-17
feat(mongodb-instrumentation): support mongodb v4


vmarchaud:
The new mongodb driver (that has been rewritten from scratch) has been released, you can see the PR of renovate which obviously fails here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/575.
I expect people will start to request to support it in the near term so opening an issue for this

CptSchnitz:
Is the plan to support both v4 and v3 or just upgrade? (v3 will get bug fixes for a year).

vmarchaud:
I think we should support both of them since v3 is still massively used

osherv:
Hey guys :)
I can do that!
pls assign this to me.
@vmarchaud @CptSchnitz

PR #794, opened 2021-12-22
feat(mongodb4): added mongodb4 instrumentation

osherv:
MongoDB4 Instrumentation!


dyladan:
Is there a reason this is a new instrumentation instead of updating the existing mongodb instrumentation?

osherv:
@dyladan 
Yes :) Here's the related Issue:
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/577
'I think we should support both of them since v3 is still massively used'

dyladan:
I was not suggesting to drop support for v3, but to add support for v4 to the existing instrumentation using a new `InstrumentationNodeModuleDefinition`.

osherv:
@dyladan They were pretty much changed everything from 3 to 4. That will cause collisions in the depenencies.
I think it has to be in another package. Or, does otel have support for multiple versions dependencies in the same package?

osherv:
@tydhot 
I tried, but the new instrumentation won't be applyed
Screen Shot 2021-12-22 at 17 42 54
lied in that case.
Screen Shot 2021-12-22 at 17 44 09

Is there any option to fix that without create a new package?

dyladan:
In most other instrumentations where we patch multiple major versions, we only use the types from the latest version and use `any` or `unknown` or hand-rolled types for the other versions.

osherv:
@dyladan Thanks! ill try!

osherv:
@dyladan Hey :)
The issue is almost ready!
Two last fails:
1. mongo4 is not supported for node < 12
2. Code coverage is now lower because the mongo3 tests is only under test-all-versions and not under tests.

Any idea how to handle with these? Thanks alot

Rauno56 (approved):
Looks good! Thanks!

Please remove `console.log(getTestSpans());` from `mongodb-v3.test.ts` as well.

Some of the small changes(few newlines, typecast to any) could also be taken to v3 test file, but that's not a deal breaker.

Rauno56:
`mongodb@4` doesn't support node 8 and 10. Please ignore the package in the CI matrix via `lerna-extra-args` in `unit-test.yml`.

osherv:
@Rauno56 Fixed everything! Thanks a lot for the review :)

Rauno56:
Nice! Even the smallest of my nits now addressed! :+1: I'd love to get another review before we merge.

plugins/node/opentelemetry-instrumentation-mongodb/src/types.ts
```diff
@@ -161,3 +162,15 @@ export enum MongodbCommandType {
   COUNT = 'count',
   UNKNOWN = 'unknown',
 }
+
+// https://github.com/mongodb/node-mongodb-native/blob/v4.2.2/src/cmap/connection.ts
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export class V4Connection {
```
dyladan, line 168: I think this should just be an interface or a type. Its concrete implementation is never used.

osherv: Agree! fixed

dyladan (approved):

Rauno56:
:exclamation: This cannot yet be merged. For some reason, the tests stay pending forever. Work locally.

dyladan (changes requested):
@Rauno56 please use "request changes" to block a PR so that it is not accidentally merged

plugins/node/opentelemetry-instrumentation-mongodb/src/types.ts
```diff
@@ -16,6 +16,7 @@
 
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
 import { Span } from '@opentelemetry/api';
+import * as mongodb from 'mongodb';
```
blumamir, line 19: You cannot import from your instrumented library (mongodb)
This will crash applications that installed the instrumentation but not `mongodb` package (valid use case).
Use `import type * as mongodb from 'mongodb';`
This will remove the import fromo JS after transpilation.

.nycrc
```diff
@@ -6,8 +6,10 @@
     "**/*.d.ts",
     "build/**/**/*.js",
     "src/index.ts",
+    "src/version.ts",
     "karma.conf.js",
     "src/platform/browser/*.ts",
+    "test/*.ts",
```
blumamir, line 12: why ignore the test coverage on test files?

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -59,7 +60,7 @@
     "codecov": "3.8.3",
     "gts": "3.1.0",
     "mocha": "7.2.0",
-    "mongodb": "3.6.11",
+    "mongodb": "4.2.2",
```
blumamir, line 63: We are installing mongodb v4 as dev dependency, which means that v3 tests will no pass, only in tav, right?

blumamir (changes requested):
Wrote a few comments.
The only critical thing is not to `import * from 'mongodb'` as it can potentially break users. 
Other comments are recommended but not a blocker

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -7,7 +7,8 @@
   "repository": "open-telemetry/opentelemetry-js-contrib",
   "scripts": {
     "docker:start": "docker run -e MONGODB_DB=opentelemetry-tests -e MONGODB_PORT=27017 -e MONGODB_HOST=localhost -p 27017:27017 --rm mongo",
-    "test": "nyc ts-mocha --parallel -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/**/*.test.ts'",
+    "test": "nyc ts-mocha -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v4.test.ts'",
```
blumamir, line 10: this change means we are running PullRequests CI tests only for v4, and stop running v3 tests on PRs.
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/81b3190af64bda14f87c5b0cbd6172bafda26408/package.json#L28
something we might want to consider...

osherv: That is correct. I can't think about any other option. Suggestions?

plugins/node/opentelemetry-instrumentation-mongodb/package.json
osherv: Right :)

.nycrc
osherv: They are not representing code coverage. That caused my to fail duo to low code coverage.

plugins/node/opentelemetry-instrumentation-mongodb/package.json
blumamir: What do you think about documenting it somehow?
Maybe we can write a comment at the top of the v3 tests file with instructions on how to run them locally.
The next person who will try to run v3 test will see all tests failing and be surprised

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -370,22 +470,71 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+      // add database related attributes
+      span.setAttributes({
+        [SemanticAttributes.DB_SYSTEM]: 'mongodb',
+        [SemanticAttributes.DB_NAME]: ns.db,
+        [SemanticAttributes.DB_MONGODB_COLLECTION]: ns.collection,
+      });
+    }
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled
+    let commandObj: Record<string, unknown>;
+    if (command && command.documents) {
```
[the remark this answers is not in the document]
osherv: Fixed ty!

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -370,22 +470,71 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+        host = hostParts[0];
+        port = hostParts[1];
+      }
+
+      // add database related attributes
+      span.setAttributes({
+        [SemanticAttributes.DB_SYSTEM]: 'mongodb',
+        [SemanticAttributes.DB_NAME]: ns.db,
+        [SemanticAttributes.DB_MONGODB_COLLECTION]: ns.collection,
+      });
```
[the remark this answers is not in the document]
osherv: Fixed ty!

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -370,22 +470,71 @@ export class MongoDBInstrumentation extends InstrumentationBase<
   /**
    * Populate span's attributes by fetching related metadata from the context
    * @param span span to add attributes to
+   * @param ConnectionCtx mongodb internal connection context
    * @param ns mongodb namespace
    * @param command mongodb internal representation of a command
+   */
+  private _populateV4Attributes(
+    span: Span,
+    ConnectionCtx: any,
```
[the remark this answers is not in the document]
osherv: Fixed ty!

plugins/node/opentelemetry-instrumentation-mongodb/package.json
blumamir: Not a particularly good one
We can run `tav` here as well, possibly with just 2 versions. But that sounds clumsy and complex.
Let's wait for more minds on this one @Rauno56 @dyladan

osherv: btw, The tests are only can run locally (we need mongodb server). So anyway i dont think i would change much

plugins/node/opentelemetry-instrumentation-mongodb/src/types.ts
osherv: Good one! fixed

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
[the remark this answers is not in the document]
blumamir: amazing.
If you are up to it, you can also further shorten it to:
```js
const commandObj = command?.documents?.[0] ?? command?.cursors ?? command
```

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -394,31 +543,51 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
       [SemanticAttributes.DB_SYSTEM]: 'mongodb',
       [SemanticAttributes.DB_NAME]: dbName,
       [SemanticAttributes.DB_MONGODB_COLLECTION]: dbCollection,
     });
 
-    if (command === undefined) return;
-
-    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
-    const commandObj = command.query ?? command.q ?? command;
+    if (host?.length && port?.length) {
```
blumamir, line 584: another small nit: empty strings are falsy too, so you can just write `if (host && port)`

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -394,31 +543,51 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
     // The namespace is a combination of the database name and the name of the
     // collection or index, like so: [database-name].[collection-or-index-name].
     // It could be a string or an instance of MongoDBNamespace, as such we
     // always coerce to a string to extract db and collection.
     const [dbName, dbCollection] = ns.toString().split('.');
+    let commandObj: undefined | Record<string, unknown>;
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
+    if (command) {
+      commandObj = command.query ?? command.q ?? command;
+    }
```
blumamir, line 557: same here. a shortened writing of this code could look like this:
```js
const commandObj = command?.query ?? command?.q ?? command
```

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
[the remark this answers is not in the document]
dyladan: I would maybe think about putting each of the three possibilities on its own line for readability

plugins/node/opentelemetry-instrumentation-mongodb/package.json
blumamir: Yes, I usually spin up a mongo server locally with docker and run the tests against it.
CI tests also start a mongo server [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/81b3190af64bda14f87c5b0cbd6172bafda26408/.github/workflows/unit-test.yml#L27), so the tests can pass.

.nycrc
blumamir: I understand the problem but believe we should find a proper solution to this than disable test coverage altogether on this package.
It's related to the issue of not running v3 tests in `npm run test`.
Maybe when it is resolved, this one will be too,

blumamir: never mind my comment, I read it again and understood why it works.
sorry for the trouble :)

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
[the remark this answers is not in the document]
osherv: Done

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -216,8 +268,49 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+        const resultHandler = callback;
+        if (
+          !currentSpan ||
+          typeof resultHandler !== 'function' ||
+          typeof cmd !== 'object' ||
+          cmd.ismaster
+        ) {
+          return original.call(this, ns, cmd, options, callback);
+        }
+        const commandType = Object.keys(cmd)[0];
```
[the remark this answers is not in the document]
osherv: Dictionary like object with the command details

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Fixed

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Fixed

plugins/node/opentelemetry-instrumentation-mongodb/package.json
osherv: Added doc :)

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Fixed

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -370,22 +470,66 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+      if (hostParts.length === 2) {
+        host = hostParts[0];
+        port = hostParts[1];
+      }
+    }
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
+    let commandObj: Record<string, unknown>;
+    if (command?.documents) {
+      commandObj = command.documents[0];
+    } else if (command.cursors) {
```
blumamir, line 498: `command?.cursor`

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -370,22 +470,66 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+          : '';
+      if (hostParts.length === 2) {
+        host = hostParts[0];
+        port = hostParts[1];
+      }
+    }
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
+    let commandObj: Record<string, unknown>;
+    if (command?.documents) {
+      commandObj = command.documents[0];
```
blumamir, line 497: what is the meaning of this documents in this place? can it be empty array?

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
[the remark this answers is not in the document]
blumamir: Can you please check how this object is constructed and if keys[0] is guraenteed to contain the command type?

Rauno56:
MongoDB tests fail, but do not terminate the process. That's why the tests never finish:
https://github.com/Rauno56/opentelemetry-js-contrib/runs/4760394725?check_suite_focus=true

Rauno56:
Regarding :point_up: That definitely needs to be fixed in two aspects:

1. errors from the tests should always terminate the process. Some combination of MongoDB and mocha doesn't currently work well, because the test is marked failed by mocha internally but it doesn't really throw and terminate. I tried it with the instrumentation disabled(but the rest of the setup there) - same thing. To reproduce just throw in the callback of the mongo command. The only thing that comes to mind right now is to catch errors in callbacks and call `done(err);` ourselves. :disappointed: 
2. The thrown error comes from the fact that mongo resolves `localhost` and stores the IP on the field we check against `localhost`. To bypass that, we should set the host to `127.0.0.1` from the get-go to avoid the discrepancy. That can be done

plugins/node/opentelemetry-instrumentation-mongodb/package.json
dyladan: We talked about this in the SIG. We recommend to test the most popular major version (in this case v3) and test other versions with TAV.

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Fixed

zombispormedio:
any update about this PR? I am interested in

osherv:
> any update about this PR? I am interested in

Sry for the delay, i had to omicron virus -_-..
I'm working on it right now

zombispormedio:
> Sry for the delay, i had to omicron virus -_-.. I'm working on it right now

Sorry, just wondering. I hope you had a good recovery. Thank you.

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Fixed! ty

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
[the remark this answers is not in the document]
osherv: Hey, i looked all over their code and didnt saw any case that this is not the command type.
but, this is Document so nothing guraenteed. Do you want me to add verification?

osherv:
@Rauno56  @blumamir 
Hey! Good to be back after the Omicron XD.
Fix all of you comments. I need to fix my CLA, so the moment you approves my PR, ill open new branch with this changes.
:)

blumamir:
> @Rauno56 @blumamir Hey! Good to be back after the Omicron XD. Fix all of your comments. I need to fix my CLA, so the moment you approves my PR, ill open new branch with these changes. :)

Good to have you back and thanks for the great work. Hope you are feeling well now :)
You can also open a new PR with the fixed CLA and I'll finish the review there. It will probably take me a day or two to get to it.

.github/workflows/unit-test.yml
```diff
@@ -15,10 +15,12 @@ jobs:
             lerna-extra-args: >-
               --ignore @opentelemetry/instrumentation-aws-sdk
               --ignore @opentelemetry/instrumentation-pino
+              --ignore @opentelemetry/instrumentation-mongodb
               --ignore @opentelemetry/instrumentation-tedious
           - node: "10"
             lerna-extra-args: >-
               --ignore @opentelemetry/instrumentation-pino
+              --ignore @opentelemetry/instrumentation-mongodb
     runs-on: ubuntu-latest
```
blumamir, line 24: since we now run tests for version 3 in CI, we can keep running it on both node 8 and 10

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -370,22 +464,67 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+          : '';
+      if (hostParts.length === 2) {
+        host = hostParts[0];
+        port = hostParts[1];
+      }
+    }
+    // capture parameters within the query as well if enhancedDatabaseReporting is enabled.
+    let commandObj: Record<string, unknown>;
+    if (command?.documents && command.documents[0]) {
+      console.log(command.documents[0]);
```
blumamir, line 491: remove console.log

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -216,8 +268,43 @@ export class MongoDBInstrumentation extends InstrumentationBase<
...
+        options: undefined | unknown,
+        callback: any
+      ) {
+        const currentSpan = trace.getSpan(context.active());
+        const resultHandler = callback;
+        if (
+          !currentSpan ||
+          typeof resultHandler !== 'function' ||
+          typeof cmd !== 'object' ||
+          cmd.ismaster
```
blumamir, line 288: I think you are trying to filter out [this method](https://docs.mongodb.com/v4.2/reference/method/db.isMaster/)
According to the docs, it is deprecated and the new method is [hello](https://docs.mongodb.com/v4.2/reference/method/db.hello/#db.hello).

[This is the relevant part from the driver](https://github.com/mongodb/node-mongodb-native/blob/dd81438593f9900122ed16a583eb9c84d0408b76/src/cmap/connect.ts#L231):
```js
    [serverApi?.version ? 'hello' : LEGACY_HELLO_COMMAND]: true,
```

[where](https://github.com/mongodb/node-mongodb-native/blob/dd81438593f9900122ed16a583eb9c84d0408b76/src/constants.ts#L122)
```js
export const LEGACY_HELLO_COMMAND = 'ismaster';
```

I suggest adding `hello` as well to the check

.github/workflows/unit-test.yml
osherv: Fixed :)

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Oh -_- Fixed

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
osherv: Good point, Fixed

blumamir:
Thanks for addressing all the comments and for adding the support for v4 🥇 
After the CLA is fixed, this can be merged as far as I am concerned

blumamir (approved):

osherv:
Thanks! next pr will be muck more quicker :)
I'll open new pr with the fixed CLA

blumamir:
closing as this PR migrated to #869 due to CLA issues

PR #796, opened 2021-12-25
fix: NestJS Core Instrumentation: copy metadata to wrapped handler

matteoga:
NestJS makes extensive use of decorators that sets metadata on controllers  handlers functions, for example: https://docs.nestjs.com/guards#setting-roles-per-handler
As the NestJS Core Instrumentation wraps the original controller handlers with an additional function,
and do not copy the metadata on the wrapped handler, any functionality that relies on this data will not work properly.
In the "roler per handler settings" example, the guard will not detect any required roles, leaving the route un-protected.
Other side effects could happen with decorators following the same pattern.

### Reproduction of the issue
1) Initialize a new NestJS HTTP Application
2) Follow the "Guards Tutorial" https://docs.nestjs.com/guards#setting-roles-per-handler
3) Ensure that calling the route without the requested role results in HTTP 401
4) Add OT NestJS Core instrumentation
5) Perform the same call as in point 3, now the request will pass through succesfully
If needed i can build a minimal reproduction repository
 
- The change will get and copy all metadata key/values from the original handler onto the wrapped handler.
I appreciate the test could seem a bit hacky, but effectively ensures that metadata is copied, if anyone has suggestions on better way to test this i'll be happy to implemente the test in such way


dyladan (approved):

vmarchaud (approved):

Flarna (approved):

vmarchaud:
friendly ping @Rauno56

dyladan:
If @Rauno56 doesn't merge or block this by tomorrow I'll merge it

Rauno56:
Thanks for the contribution, @matteoga!

PR #797, opened 2021-12-27
docs: format supported versions the same way across the readmes

Rauno56:

Trying to parse supported versions from all the packages I found that some of the packages formatted their supported versions differently. Not a huge deal but changed them to a similar format.

Many of the packages have a section like that:

> This module provides automatic instrumentation for [`package name`](...url...).
> For automatic instrumentation see the [@opentelemetry/sdk-trace-node](...url...) package.

Which reads very weird to me - both are kind of true, but in a different way. Any ideas on how to make that section more clear?


dyladan:
> Many of the packages have a section like that:
> 
> > This module provides automatic instrumentation for [`package name`](...url...).
> > For automatic instrumentation see the [@opentelemetry/sdk-trace-node](...url...) package.
> 
> Which reads very weird to me - both are kind of true, but in a different way. Any ideas on how to make that section more clear?

Maybe something like this:

> This module provides automatic instrumentation for the ___ module, which may be loaded using the [@opentelemetry/sdk-trace-node](...url...) package and is included in the [@opentelemetry/auto-instrumentations-node](...url...) bundle.
>
> If total installation size is not constrained, it is recommended to use the [@opentelemetry/auto-instrumentations-node](...url...) bundle with [@opentelemetry/sdk-node](...url...) for the most seamless instrumentation experience.

plugins/node/opentelemetry-instrumentation-pg/README.md
```diff
@@ -50,11 +52,6 @@ PostgreSQL instrumentation has few options available to choose from. You can set
 | [`enhancedDatabaseReporting`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-pg/src/pg.ts#L48) | `boolean` | If true, additional information about query parameters and results will be attached (as `attributes`) to spans representing database operations |
 | `responseHook` | `PgInstrumentationExecutionResponseHook` (function) | Function for adding custom attributes from db response |
 
-## Supported Versions
-
-- [pg](https://npmjs.com/package/pg): `7.x`, `8.*`
-- [pg-pool](https://npmjs.com/package/pg-pool): `2.x`, `3.*` (Installed by `pg`)
```
vmarchaud, line 56: you removed the information about `pg-pool` module, we need it for people using it directly

Rauno56: What do you mean by "directly"?

There's only one instrumentation implemented in the package. It happens to be using hooks to two different package names, but a user cannot separate these anyways. The way it's implemented now, makes this information an implementation detail more than anything else.

Do you think `pg-pool` is ever used separately?
It's bundled into `pg` and used through `pg.Pool` - they're also version-locked(you cannot use `pg-pool@3` with `pg@7`).

All that made me conclude it's more of a noise than useful knowledge. I don't have strong feelings towards it, but if we mention those separately, I think we should also export 2 separate instrumentations.

vmarchaud: I'm not sure, i would guess its possible for `pg` to major bump `pg-pool` internally and doing a minor bump (which would break the instrumentation), WDYT ?

Rauno56: So you mean `pg@8.99.0` would bundle `pg-pool@4` for example? That wouldn't change much in the context of this discussion, because we specify the package and supported versions in the instrumentation definition. The `pg-pool` would **not** be instrumented(because it's unsupported) and access to `pg-pool` instrumentation separate from `pg` instrumentation would not change what the user is able to do about it.

Or maybe I misunderstood your case... :thinking:

vmarchaud: > So you mean pg@8.99.0 would bundle pg-pool@4 for example ?

Yes, we currently specify `pg 8.*` so if they did that and reading your changes, you could understand that the instrumentation support it while we don't because they bumped pg-pool.

dyladan: That's a difficult distinction to explain to our users who have simply installed the `pg` module and not thought about the fact that the `pg-pool` module even exists, let alone what version was automatically installed. It is much easier to claim we support a particular version range of `pg` and have the `pg-pool` instrumentation be an implementation detail (the same way `pg-pool` is an implementation detail of `pg`). The `tav` script should handle this for us, assuming the `pg-pool` instrumentation is covered by tests. We could also add a heading along the lines of "Spans not created when using connection pools" which explains the distinction for users who run into the issue.

vmarchaud: > Do you think pg-pool is ever used separately?

I though it was possible but checking the code `pg-pool` have a peer dependency on `pg` so technically it "should" never happens. I would think that its fine to let it be documented but if you agree that its a difficult distinction, i'm fine with removing it

legendecas (approved):

dyladan:
Hmm not really sure why the tests would fail because of readme changes. Any idea?

Rauno56:
Tests are now fixed. I'll do the readme changes separately.

vmarchaud (approved):

PR #800, opened 2021-12-28
fix(examples): fix React Load plugin example

kkhaidukov:

- I wasn't able to run the React Load plugin example, this PR fixes that

- updates the config
- uses a more recent version of `opentelemetry-collector-contrib`
- adds missing dependencies
- updates the README
- sets the service name for the example

### DIFF
--- a/examples/react-load/react/docker/docker-compose.yaml
+++ b/examples/react-load/react/docker/docker-compose.yaml
@@ -3,8 +3,8 @@ services:
 
   # Collector
   collector:
-    image: omnition/opentelemetry-collector-contrib:0.2.8
-    command: ["--config=/conf/collector-config.yaml", "--log-level=DEBUG"]
+    image: otel/opentelemetry-collector-contrib:0.40.0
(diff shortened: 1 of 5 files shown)


vmarchaud (approved):

examples/react-load/react/docker/docker-compose.yaml
```diff
@@ -3,8 +3,8 @@ services:
 
   # Collector
   collector:
-    image: omnition/opentelemetry-collector-contrib:0.2.8
-    command: ["--config=/conf/collector-config.yaml", "--log-level=DEBUG"]
+    image: otel/opentelemetry-collector-contrib:0.40.0
```
dyladan, line 6: Did you try this with collector 40? Our exporters are not on the latest proto so I want to make sure it works.

dyladan (approved):

examples/react-load/react/docker/docker-compose.yaml
kkhaidukov: yeah, I just tried it again on a clean repo

PR #806, opened 2021-12-29
chore: update actions/checkout to v2

### LINKED ISSUE #791: [Security] Workflow unit-test.yml is using vulnerable action actions/checkout
The workflow unit-test.yml is referencing action actions/checkout using references v1. However this reference is missing the commit [a6747255bd19d7a757dbdda8c654a9f84db19839](https://github.com/actions/checkout/commits/a6747255bd19d7a757dbdda8c654a9f84db19839) which may contain fix to the some vulnerability. 
The vulnerability fix that is missing by actions version could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider to update the reference to the action.

dyladan:
Fixes #791

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -113,7 +113,7 @@ jobs:
             propagators/*/node_modules
           key: ${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('**/package.json') }}
       - name: Legacy Peer Dependencies for npm 7
-        if: matrix.node == 'node:16'
+        if: matrix.node == '16'
         run: npm config set legacy-peer-deps=true
       - name: Install Root Dependencies
         run: npm install --ignore-scripts
@@ -126,7 +126,7 @@ jobs:
       - name: Unit tests
         run: npm run test:ci:changed -- ${{ matrix.lerna-extra-args }}
       - name: Report Coverage
-        if: matrix.node == 'node:14'
+        if: matrix.node == '14'
         run: npm run codecov:ci:changed
   browser-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test.yml
```diff
@@ -97,7 +95,12 @@ jobs:
       CASSANDRA_PORT: 9042
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
dyladan, line 100: fetch depth 0 needed to diff against main and only test changed packages

.github/workflows/unit-test.yml
```diff
@@ -134,10 +133,12 @@ jobs:
     env:
       NPM_CONFIG_UNSAFE_PERM: true
     steps:
-      - name: Checkout
-        uses: actions/checkout@v1
       - name: Permission Setup
         run: sudo chmod -R 777 /github /__w
+      - name: Checkout
```
dyladan, line 138: checkout moved to after the permission setup. We use the circleci container because it has chrome preinstalled, but it has weird permission issues.

.github/workflows/unit-test.yml
```diff
@@ -97,7 +95,12 @@ jobs:
       CASSANDRA_PORT: 9042
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v2
```
dyladan, line 101: use setup node instead of containers in order to avoid permission issues

legendecas (approved):

vmarchaud (approved):

Issue #804, opened 2021-12-29
Add `test-all-versions` setup for `express`


Rauno56:
We currently test the instrumentation only on one version of `express`. Adding just a second version would yield coverage over upwards of 4million downloads more.

Stats from 29. December 2021.

```
┌─────────┬──────────┬──────────────┬──────────────┬───────────┬──────────┐
│ (index) │ version  │     time     │     tags     │ downloads │ ratio(%) │
├─────────┼──────────┼──────────────┼──────────────┼───────────┼──────────┤
│    0    │ '4.17.1' │ '2019-05-26' │      []      │  8185193  │   57.2   │
│    1    │ '4.17.2' │ '2021-12-17' │ [ 'latest' ] │  4971813  │   34.8   │
│    2    │ '4.16.4' │ '2018-10-11' │      []      │  471561   │   3.3    │
│    3    │ '4.16.3' │ '2018-03-12' │      []      │  198109   │   1.4    │
│    4    │ '4.16.2' │ '2017-10-10' │      []      │  106531   │   0.7    │
│    5    │ '4.15.5' │ '2017-09-25' │      []      │   56242   │   0.4    │
│    6    │ '4.15.3' │ '2017-05-17' │      []      │   28954   │   0.2    │
│    7    │ '4.13.4' │ '2016-01-22' │      []      │   22930   │   0.2    │
│    8    │ '4.17.0' │ '2019-05-17' │      []      │   17352   │   0.1    │
│    9    │ '4.15.2' │ '2017-03-06' │      []      │   16637   │   0.1    │
│   10    │ '4.14.0' │ '2016-06-16' │      []      │   16287   │   0.1    │
└─────────┴──────────┴──────────────┴──────────────┴───────────┴──────────┘
Ratio of downloads through versions shown in the table: 98.5%
Total weekly downloads: 14,306,693

```

See #783, #792 or #786 for reference.

blumamir:
implemented in #811

PR #811, opened 2021-12-31
test(express-instrumentation): add test-all-versions

matteoga:
- Adds test-all-versions for `express` running tests on the previous minor version `4.17.1` solving https://github.com/open-telemetry/opentelemetry-js-contrib/issues/804


Rauno56:
Nice! :tada: How would you feel about broadening the tested version range a bit, @matteoga? To include the current, future versions and a couple more of the old, but still used ones.

matteoga:
sounds good, what do you suggest, adding `latest` versions tag and including also 
```
┌─────────┬──────────┬──────────────┬──────────────┬───────────┬──────────┐
│ (index) │ version  │     time     │     tags     │ downloads │ ratio(%) │
├─────────┼──────────┼──────────────┼──────────────┼───────────┼──────────┤
│    2    │ '4.16.4' │ '2018-10-11' │      []      │  471561   │   3.3    │
│    3    │ '4.16.3' │ '2018-03-12' │      []      │  198109   │   1.4    |
```
?

Rauno56:
Yes, possibly add `4.16.2` as well, because the number of downloads for that is more than for some of the other packages we support in total.
However, afaik, TAV doesn't support tags so `^4.16.2` would work well.

plugins/node/opentelemetry-instrumentation-express/.tav.yml
```diff
@@ -0,0 +1,15 @@
+express:
+  - versions: "^4.17.2"
+    commands: npm run test
+
+  - versions: "4.17.1"
```
Rauno56, line 5: Any reason you didn't use `^4.16.2` for the range which matches all the listed versions?

plugins/node/opentelemetry-instrumentation-express/.tav.yml
```diff
@@ -0,0 +1,15 @@
...
+    commands: npm run test
+
+  - versions: "4.16.4"
+    commands: npm run test
+
+  - versions: "4.16.3"
+    commands: npm run test
+
+  - versions: "4.16.2"
+    commands: npm run test
```
Rauno56, line 15: nit: newline

Rauno56 (approved):
Thanks!

plugins/node/opentelemetry-instrumentation-express/.tav.yml
blumamir: I too think it's more appropriate here

blumamir:
style: use a single semantic version range to test relevant versions

blumamir (approved):

vmarchaud (approved):

PR #813, opened 2022-01-01
chore: add health check for cassandra

Rauno56:

I'm experiencing some issues in #799 with the tests for Cassandra failing. Hard to believe that it's the root cause, but I'm a bit lost right now and want to rule at least something out.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -67,6 +67,11 @@ jobs:
         image: bitnami/cassandra:3
         ports:
         - 9042:9042
+        options: >-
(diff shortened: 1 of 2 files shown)


vmarchaud (approved):

Rauno56:
I'd really like to trigger Cassandra's test with this.

Rauno56:
OK, we def have a problem with the CI setup. :(

Rauno56:
Since we fixed the CI issue and Cassandra does boot up for a while, but not longer than your bootstrapping, I'll close this PR - it mainly just slows the CI process down for another 10s.

PR #814, opened 2022-01-03
ci: revert "feat(add-tav-action): add test-all-versions as github act…

Rauno56:

The CI seems to be unstable at this point and the only way I was able to get it working reliably was to revert the recent addition of TAV runs. We still definitely want to add TAV back but after we've found a way to work around the issues we're experiencing.
My theory is that the resource limits are shared between the workflows and we're hitting the limits of those with two processes of Cassandra running(each consuming ~4GB of memory, 7GB is the limit). I don't know how to make sure it's the case.

cc: @haddasbronfman


Rauno56:
Closing this now for it's not reliably fixed even when reverting TAV. Looking into it.

PR #816, opened 2022-01-03
fix: use localhost for services in CI

Rauno56:

The services in CI are not currently accessible.
The hostnames do not resolve because the step execution was recently moved from containers(which belong to the bridge network where the service names resolve) to the runner machine.
See c69c202a - @dyladan

The [docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idservices):

> If you configure the job to run directly on the runner machine and your step doesn't use a container action, you must map any required Docker service container ports to the Docker host (the runner machine). You can access the service container using localhost and the mapped port.

The above section from the docs makes total sense, however, I'm personally not clear on why is [`actions/setup-node@v2`](https://github.com/actions/setup-node) the recommended way to run node - it pollutes the runner machine and comes with an additional layer of caching.

Since just swapping back to running in a container, results in some FS permission issues, I'm just changing the hostnames to use to localhost.

**That means the TAV setup is broken as well. I'd make the same change to it, but seeing that it doesn't set up node, I cannot see how it could work even with the services being accessible. I'll disable it for now.**

Sorted the keywords in the packages mainly to trigger a change for Lerna for all the packages.

### DIFF
--- a/.github/workflows/test-all-versions.yml
+++ b/.github/workflows/test-all-versions.yml
@@ -5,6 +5,8 @@ on:
 
 jobs:
   tav:
+    # skip the workflow for now #816
+    if: false
     name: run test-all-version (tav)
     runs-on: ubuntu-latest
     services:
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -74,24 +74,24 @@ jobs:
       RUN_MYSQL_TESTS: 1
       RUN_POSTGRES_TESTS: 1
       RUN_REDIS_TESTS: 1
-      OPENTELEMETRY_MEMCACHED_HOST: memcached
+      OPENTELEMETRY_MEMCACHED_HOST: localhost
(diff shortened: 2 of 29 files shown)


dyladan:
I switched from containers to node when I updated `actions/checkout` from v1 to v2. The v2 checkout was running into filesystem permission issues you mentioned in this PR description. At first I tried to fix ownership/permissions of the directories, but switching to setup-node was just easier. At that time the tests passed so I didn't realize it would cause a networking issue.

We can go back to running in node containers, but we will need to solve the permission issue.

dyladan (approved):
Approving this because it solves our immediate problem of tests failing but we should look into container tests again

legendecas (approved):

naseemkullah (approved):

vmarchaud (approved):

PR #817, opened 2022-01-04
fix: rename lerna's --include-filtered-dependencies option

Rauno56:

`--include-filtered-dependencies` option is deprecated:
https://github.com/lerna/lerna/commit/f2c3a92fe41b6fdc5d11269f0f2c3e27761b4c85

Rename `--include-filtered-dependencies` -> `--include-dependencies`.
Also substituting `$(npm pkg get name)` in the `package.json` of `@opentelemetry/instrumentation-aws-sdk` - GHA logs seem to tell us it's not properly working.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,4 @@
 {
-  "lerna": "3.13.4",
   "npmClient": "npm",
   "packages": [
     "packages/*",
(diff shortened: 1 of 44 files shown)


vmarchaud (approved):

lerna.json
```diff
@@ -1,5 +1,4 @@
 {
-  "lerna": "3.13.4",
```
vmarchaud, line 2: why do we need to remove this ?

Rauno56: It's deprecated from v3 and duplicated with the one in `package.json`.
See: https://github.com/lerna/lerna/pull/1225

naseemkullah (approved):

legendecas (approved):

PR #821, opened 2022-01-05
ci: fix browser cache hashing issues

Rauno56:

In some PRs browser tests fail because hashing `package.json` files takes longer than 2 minutes.
The reason is probably that it's trying to hash all package.json files in the directory tree - including installed packages, etc.
That makes the cache very fragile and long to validate.

The hashing should be done only on `package.json` files belonging to our projects. `actions/cache@v2` actually [recommends caching npm's cache folder](https://github.com/actions/cache/blob/main/examples.md#node---npm) instead of the `node_modules` which makes sense:

> Note: It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci

We cannot use `npm ci` since we do not currently check lock files in.

- I'm unifying the CI step definitions between browsers and the node tests by using matrix in both cases but only running browser tests in node 12 as it was previously.
- Specifying a strict allow-list for the files to be hashed.
- Sorting env vars

Didn't do any reliable testing, but the CI times should not have gone worse since the hashing for the cache key was the only thing that changed.

#### Notes

- We do not have `package-lock.json` files. That's a reason we cannot use `npm ci` as well. `--no-ci` has to stay.
- The official example doesn't work example because of :point_up: 
- I don't see a reason to cache `node_modules` without caching `package-lock.json` file. npm, not finding lock files, has to do the same kind of computation every time with or without the node_modules.
- caching npm's cache is the simplest, solution, but not as fast as others, it seems.
- Even though the way we cache installs make a significant difference(compare caching via npm vs the installed modules on the fs), the current bottleneck is rather in the compilation step - that takes most of the CI time right now.
- We compile the packages though only to run the tests via `ts-mocha`, so the compilation step is mainly for checking types. That could definitely be more efficient.


.github/workflows/unit-test.yml
```diff
@@ -74,24 +74,24 @@ jobs:
...
       MYSQL_PORT: 3306
+      MYSQL_USER: otel
+      OPENTELEMETRY_MEMCACHED_HOST: localhost
+      OPENTELEMETRY_MEMCACHED_PORT: 11211
+      OPENTELEMETRY_REDIS_HOST: localhost
+      OPENTELEMETRY_REDIS_PORT: 6379
+      POSTGRES_DB: circle_database
+      POSTGRES_HOST: localhost
+      POSTGRES_PORT: 5432
+      POSTGRES_USER: postgres
```
dyladan, line 93: Is this actually changed or just sorted?

.github/workflows/unit-test.yml
```diff
@@ -125,32 +131,42 @@ jobs:
       - name: Report Coverage
         if: matrix.node == '14'
         run: npm run codecov:ci:changed
-  browser-tests:
+
+  browser-test:
+    strategy:
+      fail-fast: false
+      matrix:
```
dyladan, line 138: Why matrix when there is only one version? Shouldn't we just run setup-node with `12` directly?

johnbley (approved):
Looks great, thanks for looking into these flaky tests!

blumamir:
> `actions/cache@v2` actually [recommends caching npm's cache folder](https://github.com/actions/cache/blob/main/examples.md#node---npm) instead of the `node_modules` which makes sense:
> 
> > Note: It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci
> 
> We cannot use `npm ci` since we do not currently check lock files in.

My understanding from the text is that it's better to cache the `~/.npm` folder. Will it work in our case? why do we keep using the repo's `node_modules`?

.github/workflows/unit-test.yml
Rauno56: Just sorted. Mentioned that in the OP.

.github/workflows/unit-test.yml
Rauno56: I did it that way to make sure all the setup and env vars match between the jobs. To make it less error-prone to bring changes over from one job to another.

Actually, it does not make sense to have them as separate jobs at all - the bootstrapping and install phase is more or less the same. I'll see whether we can just skip one of those steps next.

Rauno56:
> My understanding from the text is that it's better to cache the ~/.npm folder. Will it work in our case? why do we keep using the repo's node_modules?

Good question. In the (very unscientific) tests I made, for some reason, for browser tests, it took 5min longer to do bootstrapping when caching was done like that. I figured I'll stop over-optimizing for now and fix the original issue first - hashing taking 20 minutes.

Once this is merged I want to try

1. to merge the bootstrapping steps and
2. test caching via npm cache again.

vmarchaud (approved):

blumamir (approved):
LGTM

Rauno56:
I now noticed that browser bootsrapping takes ages compared to node. That's probably because the cache doesn't work well for both of the scenarios because the bootstrapping is done differently - for one we hoist(node), for the other, we do not. I'm testing hoisting for the browser tests - if that doesn't yield, I'll make it to have a separate cache for the browser tests.

Rauno56:
Had to additionally disable running browser tests in parallel because of the Karma port conflicts. :facepalm:

.github/workflows/unit-test.yml
Rauno56: @dyladan, your thoughts?

dyladan: if we can merge them into a single matrix that would be nice but iirc finding an image that worked well with chrome and such was a pain and using different images for different branches of a matrix is also a pain.

I'm fine with the explanation you gave. There is no real _disadvantage_ to having a single-branch matrix.

dyladan:
> > My understanding from the text is that it's better to cache the ~/.npm folder. Will it work in our case? why do we keep using the repo's node_modules?
> 
> Good question. In the (very unscientific) tests I made, for some reason, for browser tests, it took 5min longer to do bootstrapping when caching was done like that. I figured I'll stop over-optimizing for now and fix the original issue first - hashing taking 20 minutes.
> 
> Once this is merged I want to try
> 
> 1. to merge the bootstrapping steps and
> 2. test caching via npm cache again.

Caching `~/.npm` caches the downloaded tarballs but still requires each one to be unpacked, have its package.json read, the npm server pinged to see what the latest version is, and so on. Caching the package-lock.json files and/or the node_modules directories skips a huge amount of work and remote calls to npm.

Rauno56:
> Caching `~/.npm` caches the downloaded tarballs but still requires each one to be unpacked, have its package.json read, the npm server pinged to see what the latest version is, and so on. Caching the package-lock.json files and/or the node_modules directories skips a huge amount of work and remote calls to npm.

Both points I made were for future explorations. In this PR I was mainly after fixing the hashing for the caching key taking 2 minutes and ending in failure. Tried to keep changes in other areas to a minimum.

I tested several solutions and honestly was surprised how well caching npm cache compared to the current solution. That is why I definitely want to explore it further - of course, it can prove not to be worth it in which case it'll be a lesson learned! In the end, it will be tradeoffs all the way down.

From my end, this PR is ready for a merge. @dyladan has pending comment threads - I'll leave it to you to merge then or continue on those threads.

dyladan:
Didn't realize i was the blocker on this

PR #823, opened 2022-01-06
wip: feat: re-enable TAV

Rauno56:

Had to temporarily disable TAV after some CI problems. Implementing some fixes and re-enabling it again.
After the discussion in SIG, the plan is:

- disable running on every PR - trust devs to do that locally;
- add a reminder of :point_up: in the PR template;
- run TAV setup scheduled;
- run TAV setup on release PRs.

- removed change-detection - doesn't make sense for scheduled runs;


.github/workflows/test-all-versions.yml
```diff
@@ -118,4 +118,4 @@ jobs:
       - name: Bootstrap Dependencies
         run: npx lerna bootstrap --no-ci --hoist --nohoist='zone.js' --nohoist='mocha' --nohoist='ts-mocha'
       - name: Run test-all-versions
-        run: lerna run test-all-versions -- ${{ matrix.lerna-extra-args }}
+        run: npx lerna run test-all-versions -- ${{ matrix.lerna-extra-args }}
```
dyladan, line 121: Maybe we should add a `test-all-versions` script to the root package.json that calls `lerna` so we don't have to use `npx`? I don't mind npx but this way we would guarantee a local version is used and we aren't installing over the network every time we run.

vmarchaud: I agree

Rauno56:
Status: tests for `ioredis@4.28.1` fail.

vmarchaud:
> Status: tests for ioredis@4.28.1 fail.

Is it flaky or just not working ? Seems weird that it doesn't work for node 16 but work for 8

Rauno56:
I'm not sure why, but the tests are actually not even running in node 8.

I think it's not flaky, but a bug. I just wanted to add a comment here for people who might have a question about why this is not moving along.

Rauno56:
Submitted fixes with #830, will turn those off until that's merged and this one here rebased.

Rauno56:
I'm afraid that running tests for packages concurrently doesn't work because of weird race conditions with `lerna` linking or something. Cryptic errors like "`requ` not defined" in the `test-utils` package.

dyladan (approved):

vmarchaud (approved):

blumamir (approved):

PR #825, opened 2022-01-10
fix: fix CI by forcing colors@1.4.0

Rauno56:

`colors` a widely used terminal styling package is causing our CI to crash. Not sure when that will be fixed and how so meanwhile I'll add a dev dependency on colors@1.4.0 to force npm to install an older version.


legendecas (approved):
Can we use the npm [overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides)? Though it is only available on npm v8.3.0 and later.

dyladan (approved):
LGTM % the question about overrides

Rauno56:
Yeah, it doesn't help us on node <16 :(

Rauno56:
I'll merge this for now. It's definitely not a permanent solution.
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/826

PR #835, opened 2022-01-14
chore: remove unused ts-node dependencies

blumamir:

- Following discussion [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/712), with the great eye of @Rauno56 , It was noticed that ts-node dependency is installed but not used in `instrumentation-aws-sdk` and `browser-extension-autoinjection`.

- This PR removes these unused dependencies which can also slightly improve install and CI times.

When this PR merges, there will be no dependencies on `ts-node` in the repo


plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -1,83 +1,82 @@
 {
-  "name": "@opentelemetry/instrumentation-aws-sdk",
```
dyladan, line 2: This whole file appears to be retabbed with 4 spaces instead of 2

Rauno56 (approved):

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
blumamir: Right, my IDE is automatically formatting my files. I'll fix that. Thanks for noticing

blumamir: Fixed

Flarna (approved):

dyladan (approved):

packages/opentelemetry-browser-extension-autoinjection/package.json
```diff
@@ -84,4 +83,4 @@
     "react": "17.0.2",
     "react-dom": "17.0.2"
   }
-}
+}
```
dyladan, line 86: missing trailing newline

PR #841, opened 2022-01-20
fix(aws-sdk): sns span name should be with low cardinality

YanivD:

- When using SNS to publish message to a phone number, span name contain the phone number.
According messaging spec, [span name should be of low cardinality](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-name)

- Span name will be `phone_number send` when publishing SNS to a phone number. As I implemented in [ruby's instrumentation](https://github.com/open-telemetry/opentelemetry-ruby/blob/658440f2f749f151fde393ce1654803a816f1c38/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb#L22-L23)


blumamir (approved):
LGTM

Rauno56 (approved):

PR #844, opened 2022-01-21
feat(aws-sdk): add http status code attribute to aws sdk span

YanivD:

- Add `http.status_code` attribute, as discussed in Ruby repo
https://github.com/open-telemetry/opentelemetry-ruby/issues/1094
 
## Checklist

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
@@ -57,6 +57,7 @@ import {
   removeSuffixFromStringIfExists,
 } from './utils';
 import { RequestMetadata } from './services/ServiceExtension';
+import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 
 const V3_CLIENT_CONFIG_KEY = Symbol(
   'opentelemetry.instrumentation.aws-sdk.client.config'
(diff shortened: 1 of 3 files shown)


YanivD:
This PR is not ready for review because I couldn't run `npm run test-all-versions`.
I'll mark this PR as ready for review once I'll verify it pass successfully.

blumamir (approved):

naseemkullah (approved):

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
```diff
@@ -469,9 +478,18 @@ export class AwsInstrumentation extends InstrumentationBase<typeof AWS> {
               const promiseWithResponseLogic = resultPromise
                 .then(response => {
                   const requestId = response.output?.$metadata?.requestId;
+                  const httpStatusCode =
```
NathanielRN, line 481: Nit: Wondering if we should move this variable down 2 lines so that it is closer to the if block where it is used? That’s consistent with the `extendedRequestId` Block below, and then we can leave a blank space between these?

NathanielRN (approved):
Thanks for this!

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
YanivD: Thanks for your comment. moved this variable

PR #845, opened 2022-01-23
fix(add-inst-to-dict): add more instrumentations to getNodeAutoInstru…

haddasbronfman:

The function getNodeAutoInstrumentations() doesn't return all the supported instrumentations, but only part of them. 

This PR added the missing instrumentations to the function getNodeAutoInstrumentations().

## Checklist


haddasbronfman:
This PR does not include a User ID. A new one will be opened instead.

PR #872, opened 2022-02-09
feat: implement auto-instrumentation for `fs`

Rauno56:

This implements partial auto-instrumentation for the native `fs` module.

This instruments most of the `fs` functions. Full list in `constants.ts` file.
Provides two hooks:

- `createHook: (functionName: string, info: { args: ArrayLike<unknown> }) => boolean;`
- `endHook: (functionName: string, info: { args: ArrayLike<unknown>; span: api.Span }) => void;`

Returning `false` from `createHook` skips creating the trace.

## Checklist

- [x] ~Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.~


plugins/node/instrumentation-fs/src/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  [K in keyof T]: T[K] extends Function ? K : never;
+}[keyof T];
+export type FunctionProperties<T> = Pick<T, FunctionPropertyNames<T>>;
+
+export type FMember = FunctionPropertyNames<typeof fs>;
+export type FPMember = FunctionPropertyNames<typeof fs['promises']>;
+
+export interface FsInstrumentationConfig extends InstrumentationConfig {
+  createHook?: (
+    functionName: string,
```
vmarchaud, line 31: I believe you could use `FMember | FPMember` to have better typechecking ?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+            original,
+            this,
+            ...args
+          );
+        }
+      }
+
+      const span = instrumentation.tracer.startSpan(`fs ${fnName}`) as api.Span;
+      try {
+        // QUESTION: Should we immediately suppress all internal nested calls?
```
vmarchaud, line 140: not sure which nested calls you are refering to ?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+} from './constants';
+import type * as fs from 'fs';
+import type { FsInstrumentationConfig } from './types';
+
+type FS = typeof fs;
+
+const hasPromises = parseInt(process.versions.node.split('.')[0], 10) > 8;
+
+export default class FsInstrumentation extends InstrumentationBase<FS> {
+  constructor(protected override _config: FsInstrumentationConfig = {}) {
```
blumamir, line 38: It's preferable to not use use default value.
[See here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/892#discussion_r805390997)

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+      }
+      const { createHook, endHook } =
+        instrumentation.getConfig() as FsInstrumentationConfig;
+      if (typeof createHook === 'function') {
+        if (
+          createHook(fnName, {
+            args: args,
+          }) === false
+        ) {
+          // return original.apply(this, args);
```
blumamir, line 279: remove comment?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+      if (isTracingSuppressed(api.context.active())) {
+        // Performance optimization. Avoid creating additional contexts and spans
+        // if we already know that the tracing is being suppressed.
+        return original.apply(this, args);
+      }
+      const { createHook, endHook } =
+        instrumentation.getConfig() as FsInstrumentationConfig;
+      if (typeof createHook === 'function') {
+        if (
+          createHook(fnName, {
```
blumamir, line 275: try / catch to protect against exception?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+            original,
+            this,
+            ...args
+          );
+        }
+      }
+
+      const span = instrumentation.tracer.startSpan(`fs ${fnName}`) as api.Span;
+      try {
+        // QUESTION: Should we immediately suppress all internal nested calls?
```
blumamir, line 291: I think we should. We should at least be consistent with the case above when `createHook` return false

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+        span.end();
+        return res;
+      } catch (err) {
+        span.recordException(err);
+        span.setStatus({
+          message: err.message,
+          code: api.SpanStatusCode.ERROR,
+        });
+        if (typeof endHook === 'function') {
+          endHook(fnName, { args: args, span });
```
blumamir, line 314: WDYT about also adding the err here to the info? I imagine it can be useful sometimes

blumamir, line 314: Should also be wrapped in try / catch. Otherwise the hook error will be the one returned to the user in case it throws

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+          ...args
+        );
+        if (typeof endHook === 'function') {
+          try {
+            endHook(fnName, { args: args, span });
+          } catch (e) {
+            instrumentation._diag.error('caught endHook error', e);
+          }
+        }
+        span.end();
```
blumamir, line 305: a bit safer alternative would be to place it in a finally block IMO

plugins/node/instrumentation-fs/src/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    info: { args: ArrayLike<unknown> }
+  ) => boolean;
+  endHook?: (
```
blumamir, line 34: I think in other places we usually set the first param to `span` and place function data in the second object

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+        return res;
+      } catch (err) {
+        span.recordException(err);
+        span.setStatus({
+          message: err.message,
+          code: api.SpanStatusCode.ERROR,
+        });
+        if (typeof endHook === 'function') {
+          endHook(fnName, { args: args, span });
```
blumamir, line 163: try / catch this hook

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+          ...args
+        );
+        if (typeof endHook === 'function') {
+          try {
+            endHook(fnName, { args: args, span });
+          } catch (e) {
+            instrumentation._diag.error('caught endHook error', e);
+          }
+        }
+        span.end();
```
blumamir, line 154: suggestion: move to finally block

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+  SYNC_FUNCTIONS,
+} from './constants';
+import type * as fs from 'fs';
+import type { FsInstrumentationConfig } from './types';
+
+type FS = typeof fs;
+
+const hasPromises = parseInt(process.versions.node.split('.')[0], 10) > 8;
+
+export default class FsInstrumentation extends InstrumentationBase<FS> {
```
blumamir, line 37: should we also add it to the `auto-instrument` package?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+        if (typeof endHook === 'function') {
+          endHook(fnName, { args: args, span });
+        }
+        span.end();
+        throw err;
+      }
+    };
+  }
+
+  protected _patchAsyncFunction<T extends (...args: any[]) => ReturnType<T>>(
```
blumamir, line 171: I would suggest naming it `_patchCallbackApiFunction`. When I read `Async` I think promises. This is also the name used in nodejs docs

plugins/node/instrumentation-fs/src/constants.ts
```diff
@@ -0,0 +1,136 @@
...
+  'writeFile',
+  // 'close', // functions on file descriptor
+  // 'fchmod', // functions on file descriptor
+  // 'fchown', // functions on file descriptor
+  // 'fdatasync', // functions on file descriptor
+  // 'fstat', // functions on file descriptor
+  // 'fsync', // functions on file descriptor
+  // 'ftruncate', // functions on file descriptor
+  // 'futimes', // functions on file descriptor
+  // 'lchmod', // only implemented on macOS
```
blumamir, line 87: is there a reason not to implement it nevertheless?

plugins/node/instrumentation-fs/test/index.test.ts
```diff
@@ -0,0 +1,366 @@
...
+import Instrumentation from '../src';
+import { Resource } from '@opentelemetry/resources';
+import { SemanticResourceAttributes as ResourceAttributesSC } from '@opentelemetry/semantic-conventions';
+import { JaegerExporter } from '@opentelemetry/exporter-jaeger';
+import * as sinon from 'sinon';
+import type * as FSType from 'fs';
+import tests, { TestCase, TestCreator } from './definitions';
+import type { FMember, FPMember } from '../src/types';
+
+const supportsPromises = parseInt(process.versions.node.split('.')[0]) >= 12;
```
blumamir, line 34: in `instrumentation.ts` the check for > 8.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: `fs.writeFile` for example calls `open` and `close` internally. Latter are the nested calls I'm referring to.

plugins/node/instrumentation-fs/src/types.ts
Rauno56: :+1: Removed all references to strings from the implmeentation.

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,321 @@
...
+  ASYNC_FUNCTIONS,
+  PROMISE_FUNCTIONS,
+  SYNC_FUNCTIONS,
+} from './constants';
+import type * as fs from 'fs';
+import type { FsInstrumentationConfig } from './types';
+
+type FS = typeof fs;
+
+const hasPromises = parseInt(process.versions.node.split('.')[0], 10) > 8;
```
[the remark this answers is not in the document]
Rauno56: I would use `semver` for anything more complex than that but consciously did not here. Do you think we should?

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: :+1:

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Done.

plugins/node/instrumentation-fs/src/instrumentation.ts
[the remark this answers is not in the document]
blumamir: I'm good with both options. Just think that using `semver` is more robust and readable. But I agree it's a bummer to import another package for such a simple task.
Feel free to resolve or change :)

plugins/node/instrumentation-fs/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import FsInstrumentation from './instrumentation';
+
+export { FsInstrumentation };
+export * from './instrumentation';
```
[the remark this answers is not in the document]
Rauno56: Absolutely, added exports for member, hook and config types.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: It's a bug here, yes. I refactored running hooks altogether and brought running those into one place and making them safe to run in one go.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Hmm... interesting. I don't see a strong reason to seek consistency with the `createHook` behavior since the scenarios are different.

1. The question in the comment concerns whether an user can expect to see only top-level calls which were initiated in the top level.
2. `createHook` behavior concerns the use-case when user has disabled the main call thus already signalling disinterest to the calls nested IMHO.

But you are right pointing out that `createHook` doesn't cover all the usecases: for example when I want to catch ALL(even nested) the "open" calls but nothing else.

I'd vote for
1. 70% removing suppressing nested calls and giving that control(and responsibility) to the user OR
2. 30% leaving it as is making it convenient to ignore noise.

I would want to avoid, however making the solution more complex than it absolutely has to be now and in the future. `fs` is very frequently called API and I would want to stay as slim as possible in such a hot path.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Yes. Addressed that in the hook refactor.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Addressed in hook refactor.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Added `error`.

plugins/node/instrumentation-fs/src/types.ts
Rauno56: I've prioritized consistency between the hooks over consistency between the other instrumentations here. But it does make sense to have Span up front for `endHook` - We have it the other way around as well(`responseHook` in aws sdk). Span is not available for `createHook`.

This would be an option:

```typescript
export type CreateHook = (
  info: { functionName: FMember | FPMember; args: ArrayLike<unknown> }
) => boolean;
export type EndHook = (
  span: api.Span,
  info: { functionName: FMember | FPMember; args: ArrayLike<unknown>; error?: Error }
) => void;
```

OR split the info object up to separate arguments altogether. I'd support the latter I think if we're going to move span to be the first argument.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Good idea.. Changed that everywhere.

plugins/node/instrumentation-fs/src/constants.ts
Rauno56: No good one.
Just wanted to get the core of it done and included everything that came "free" for the first version of it.

plugins/node/instrumentation-fs/test/index.test.ts
Rauno56: Unified those.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Changed to use `finally` for ending spans now. Running `endHook` cannot be moved to `finally` because we're now passing error to it.

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: I'd do it separately since packages take some time to get published.

Rauno56:
Thanks **a lot** @vmarchaud and @blumamir. Really valuable feedback! I addressed everything in one way or another and took the liberty to resolve all the threads I found trivial and resolved.

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -0,0 +1,322 @@
...
+              this._unwrap(fs, fName);
+            }
+            this._wrap(
+              fs,
+              fName,
+              <any>this._patchCallbackFunction.bind(this, fName)
+            );
+          }
+          if (supportsPromises) {
+            for (const fName of PROMISE_FUNCTIONS) {
```
vmarchaud, line 76: since some might be not available i think you should check that they are defined before wrapping them, it might be possible that user rely on the function existing or not for their apps

plugins/node/instrumentation-fs/src/instrumentation.ts
dyladan: Personally I would. Maybe we can introduce a verbose mode later if users ask for it

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: `shimmer` does that for us already and exits early. Do you think I should double-check it here before calling wrap?

> it might be possible that user rely on the function existing or not for their apps

The instrumentation doesn't change the API. I don't think I understood you correctly here, but if the user relies on `rm` function existing(added in v14) for example, but runs v12 we can't and should not do anything about it.

vmarchaud (approved):

plugins/node/instrumentation-fs/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@opentelemetry/instrumentation-fs",
+  "version": "0.27.0",
```
dyladan, line 3: Why starting at version 27?

plugins/node/instrumentation-fs/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "build/src/**/*.d.ts"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.0.2"
+  },
+  "devDependencies": {
+    "@opentelemetry/api": "1.0.2",
```
dyladan, line 45: may want to update to latest deps

plugins/node/instrumentation-fs/src/constants.ts
dyladan: Thinks its fine to get something out and iterate

plugins/node/instrumentation-fs/src/instrumentation.ts
dyladan: I'm going to let you make a decision here as i can see it both ways. As I said I would lean towards suppressing and we can always unsuppress or add a verbose mode later. Hard to go back the other way without breaking alerts and stuff.

dyladan (approved):
I'll let you decide on the suppressing. I'm going to approve, but not merge until you've made that decision.

plugins/node/instrumentation-fs/src/instrumentation.ts
blumamir: 👍

plugins/node/instrumentation-fs/src/constants.ts
blumamir: 👍

plugins/node/instrumentation-fs/package.json
blumamir: If the user uses an old 1.x.x version, this will force him to upgrade to a higher api version in order to use the instrumentation, when in practice older versions will work fine as well, right?
I find myself sometimes wondering if the version should be the minimum supported one (1.0.0), or the latest 🤔

blumamir (approved):
LGTM, great work. Was waiting for instrumentation for this package, thanks for taking it 🤩

release-please-config.json
```diff
@@ -16,6 +16,7 @@
     "packages/opentelemetry-id-generator-aws-xray": {},
     "packages/opentelemetry-propagation-utils": {},
     "packages/opentelemetry-test-utils": {},
+    "plugins/node/instrumentation-fs": {},
```
blumamir, line 19: based on #921, should it also be added to `.release-please-manifest.json`?

dyladan:
@Rauno56 you can merge this if you're happy with it. I think with release-please updated we should no longer need to manually add it to the manifest.

release-please-config.json
dyladan: I think we don't need that now that release-please is updated to v3

plugins/node/instrumentation-fs/package.json
Rauno56: Out of a habit. We used to version-match packages. I guess it can now be `0.1.0`, which makes a lot more sense.

plugins/node/instrumentation-fs/package.json
Rauno56: It makes sense for the implementors/libs to specify the smallest version supported to maximize compatibility across libs.

In that sense, having a core package expose interfaces to implement messes with the semver expectations.

But the I don't see a reason to hold back updating the dev deps.

release-please-config.json
Rauno56: Removed that change.

Rauno56:
> Hard to go back the other way without breaking alerts and stuff.

I think that's a good point. I'll change that to be strict for now.

Note to myself: test to make sure user calls in callbacks aren't suppressed.

plugins/node/instrumentation-fs/package.json
Flarna: As far as I know renovate bot will create a PR to pin it in case you would use e.g. `^1.0.2`. But on the other hand it seems renovate PRs are mostly ignored anyway meanwhile otherwise we would have 1.0.4 at other places.

blumamir: > But the I don't see a reason to hold back updating the dev deps.

I think CI enforces that the version of `@opentelemetry/api` in peerDependencies matches the version in devDependencies

Rauno56: That's true. I'll address that separately in #936 and update all the packages after that's merged. For now. I'll comply with the restrictions peer-check currently enforces.

> As far as I know renovate bot will create a PR to pin it in case you would use e.g. ^1.0.2. But on the other hand it seems renovate PRs are mostly ignored anyway meanwhile otherwise we would have 1.0.4 at other places.

I agree we could be better with updating the minor versions of packages. It's currently problematic because renovate bot bundles them all together and some of them fail tests - #618

Probably not a huge deal, but requires a significant contribution to resolve.

PR #890, opened 2022-02-12
add field name to graphql resolve span name

tamland:

#534

## Checklist

Screenshot from Elastic:

screenshot


dyladan:
Can you include a screenshot of what it looked like before? How high of cardinality would you expect this to be? If it is relatively low then this is probably ok, but if it is high then I worry we might be violating the spec.

tamland:
@dyladan I'm using this patch so I can't take a screenshot. But it looks as the above image except every span is named `graphql.execute`. With more than a few spans the trace view in elastic is pretty much useless.

The cardinality will be as many fields you are tracing in your graphql schema.

PR #892, opened 2022-02-13
feat: amqplib instrumentation

blumamir:
This is an instrumentation library for the [amqplib](https://www.npmjs.com/package/amqplib) package which is used for RabbitMQ.

It exists for a year and is currently hosted in [ext-js repo](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-amqplib), been used and working in production environments, and has 6.4K weekly downloads [from npm](https://www.npmjs.com/package/opentelemetry-instrumentation-amqplib).

The code is ported as-is with few changes to the hooks signatures.


plugins/node/instrumentation-amqplib/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
Flarna, line 189: Shound't we have a year and owner here?

plugins/node/instrumentation-amqplib/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+- For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
+- For more about OpenTelemetry JavaScript: <https://github.com/open-telemetry/opentelemetry-js>
+- For help or feedback on this project, join us in [GitHub Discussions][discussions-url]
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[discussions-url]: https://github.com/open-telemetry/opentelemetry-js/discussions
```
Flarna, line 95: is it intended to point to core repo here?

plugins/node/instrumentation-amqplib/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  moduleVersion: string | undefined;
+  exchange: string;
+  routingKey: string;
+  content: Buffer;
+  options?: amqp.Options.Publish;
+  isConfirmChannel?: boolean;
+}
+
+export interface PublishConfirmedInfo extends PublishInfo {
+  confirmError?: any;
```
Flarna, line 30:
```suggestion
  confirmError?: unknown;
```

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+  MESSAGE_STORED_SPAN,
+  normalizeExchange,
+  unmarkConfirmChannelTracing,
+} from './utils';
+import { VERSION } from './version';
+
+export class AmqplibInstrumentation extends InstrumentationBase<typeof amqp> {
+  protected override _config!: AmqplibInstrumentationConfig;
+
+  constructor(config: AmqplibInstrumentationConfig = {}) {
```
Flarna, line 64:
```suggestion
  constructor(config?: AmqplibInstrumentationConfig) {
```

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+            // `serverProperties` on the `conn` and not in a property `connection`.
+            // I don't have capacity to debug it currently but it should probably be fixed in @types or
+            // in the package itself
+            // currently setting as any to calm typescript
+            const serverAttributes = getConnectionAttributesFromServer(
+              conn as any
+            );
+            Object.defineProperty(conn, CONNECTION_ATTRIBUTES, {
+              value: { ...urlAttributes, ...serverAttributes },
+              enumerable: false,
```
Flarna, line 256:
```suggestion
              enumerable: false, configurable: true, writable: true
```
at least `configurable: true` should be added as non configureable properties are quite uncommon.

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+            });
+          }
+          openCallback.apply(this, arguments);
```
Flarna, line 259:
```suggestion
          Reflect.apply(openCallback, this, arguments);
```

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+      url: string | amqp.Options.Connect,
+      socketOptions: any,
+      openCallback: Function
+    ) {
+      return original.call(
+        this,
+        url,
+        socketOptions,
+        function (this: unknown, err, conn: amqp.Connection) {
+          if (err === null) {
```
Flarna, line 244:
```suggestion
          if (err == null) {
```

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+        break;
+      }
+      this.endConsumerSpan(
+        currMessage.msg,
+        null,
+        EndOperation.InstrumentationTimeout,
+        true
+      );
+    }
+    spansNotEnded.splice(0, i);
```
Flarna, line 747: Is it intended to mutate `spansNotEnded` during iteration? Won't this skip some elements as `i` is not adapted accordingly.

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+                  : requeue === false
+                  ? ' without requeue'
+                  : ''
+              }`
+            : operation,
+      });
+    }
+    this.callConsumeEndHook(storedSpan, message, isRejected, operation);
+    storedSpan.end();
+    delete message[MESSAGE_STORED_SPAN];
```
Flarna, line 689: from performance point of view setting it to `undefined` instead deleteing it is better

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+        !Object.prototype.hasOwnProperty.call(channel, CHANNEL_SPANS_NOT_ENDED)
+      ) {
+        if (self._config.consumeTimeoutMs) {
+          const timer = setInterval(() => {
+            self.checkConsumeTimeoutOnChannel(channel);
+          }, self._config.consumeTimeoutMs);
+          timer.unref();
+          Object.defineProperty(channel, CHANNEL_CONSUME_TIMEOUT_TIMER, {
+            value: timer,
+            enumerable: false,
```
Flarna, line 386: I'm not sure if it is really needed to use `Object.defineProperties` for symbol properties because they are never included in e.g. `Object.keys()` or in `for in` loops.
using `channel[CHANNEL_CONSUME_TIMEOUT_TIMER]` would be easier to read and faster

plugins/node/instrumentation-amqplib/LICENSE
blumamir: Good catch
I actually just copied this file from another instrumentation and never even looked at the content.
Fixed in amqplib. Documented for other places in #893

.release-please-manifest.json
```diff
@@ -1 +1,46 @@
...
+    "detectors/node/opentelemetry-resource-detector-aws": "1.0.3",
+    "detectors/node/opentelemetry-resource-detector-gcp": "0.26.2",
+    "detectors/node/opentelemetry-resource-detector-github": "0.26.1",
+    "metapackages/auto-instrumentations-node": "0.27.3",
+    "metapackages/auto-instrumentations-web": "0.27.2",
+    "packages/opentelemetry-browser-extension-autoinjection": "0.27.3",
+    "packages/opentelemetry-host-metrics": "0.27.1",
+    "packages/opentelemetry-id-generator-aws-xray": "1.0.1",
+    "packages/opentelemetry-test-utils": "0.29.0",
+    "plugins/node/instrumentation-amwplib": "0.27.0",
```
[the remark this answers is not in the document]
blumamir: I was following the change in #887 .
Should it not be added here?

plugins/node/instrumentation-amqplib/README.md
blumamir: I think we don't have a discussion page for contrib, but use the core repo for both.
Let me know if I am missing anything.
Thanks

plugins/node/instrumentation-amqplib/src/types.ts
blumamir: I am using the type as [published by `@types/amqplib`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ea0e0de908b17dd697e373da9cd0fca8d7e41c66/types/amqplib/callback_api.d.ts#L54):

```js
    publish(exchange: string, routingKey: string, content: Buffer, options?: Options.Publish, callback?: (err: any, ok: Replies.Empty) => void): boolean;
```

Should this be changed to `unknown` only in instrumentation?

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: Looks like there are many instrumentations using the default argument pattern and many that use the optional param pattern.
When should we use each?

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: fixed. thanks

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: This is done at the end of the iteration and uses the last value of `i` as used when the for loop is `break` or end.
Do you think it's incorrect?

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: changed to use undefined.
Thanks :)

.release-please-manifest.json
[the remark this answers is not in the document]
Flarna: honestly speaking I don't know. Release please does a bit too much magic for me at the moment.

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: These attributes are actually const once calculated and should not be altered at any time in the future.
So it does make sense logically.
But if it causes performance or readability concerns I am good with changing it.

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,749 @@
...
+      openCallback: (err: any, connection: amqp.Connection) => void
+    ) => amqp.Connection
+  ) {
+    return function patchedConnect(
+      this: unknown,
+      url: string | amqp.Options.Connect,
+      socketOptions: any,
+      openCallback: Function
+    ) {
+      return original.call(
```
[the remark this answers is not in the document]
blumamir: We use the `original.apply` everywhere in instrumentations, and `original.call` in dozens of places in the repo.
I couldn't find any use in `Reflect.apply`.
I want to stay consistent with the rest of the codebase, and maybe have a discussion regarding this in the SIG and then change it everywhere if it's prefered.
WDYT?

plugins/node/instrumentation-amqplib/README.md
Flarna: Ah yes, you are right. Sorry.

plugins/node/instrumentation-amqplib/src/types.ts
Flarna: `unknown` is more save. But I think it's ok stay inline with the original definitions.
But we should be aware that changing `any` to something else is usually breaking for some users and therefore most likely semver major.

plugins/node/instrumentation-amqplib/src/amqplib.ts
Flarna: A while ago I updated the core repo in this regard (see https://github.com/open-telemetry/opentelemetry-js/pull/2522) but I never found the time to do the same in contrib.

There are no rules as far as I know it's just not needed to use a default here.

plugins/node/instrumentation-amqplib/src/amqplib.ts
Flarna: seems I misread this during review, it looked like this line is inside the loop not after the loop.

plugins/node/instrumentation-amqplib/src/amqplib.ts
Flarna: As I'm coming from strong typed languages I agree. But JS is different as everything is mutable on default.
Just consider that everything what should not be altered would be really readonly/non configurable - not a single instrumentation would be possible.

plugins/node/instrumentation-amqplib/src/amqplib.ts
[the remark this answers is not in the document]
Flarna: As far as I know the initial instrumentations were copied from opencensus which had support for much older node.js versions and/or browsers. Therefore these had to use `.apply` and `.call`,... because of this.

Feel free to keep it. We can move forward in this independent of this PR, maybe even on a repo scope.

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: cool, I changed it in amqplib 👍

.release-please-manifest.json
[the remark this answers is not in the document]
blumamir: after a second look, I moved the new package from `.release-please-manifest.json` to `release-please-config.json`. Hope it's the right place to configure it

blumamir:
@Flarna Thanks again for the review. If I am not mistaken, everything is resolved

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,733 @@
...
+      ) {
+        // msg is expected to be null for signaling consumer cancel notification
+        // https://www.rabbitmq.com/consumer-cancel.html
+        // in this case, we do not start a span, as this is not a real message.
+        if (!msg) {
+          return onMessage.call(this, msg);
+        }
+
+        const headers = msg.properties.headers ?? {};
+        const parentContext = propagation.extract(context.active(), headers);
```
Flarna, line 401: Are you sure that `context.active()` is used here? in case the headers include a span it's used, otherwise we might link to whatever is active at this time.
Shouldn't we use `ROOT_CONTEXT` here instead to create a new root in case nothing is in the headers?

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -0,0 +1,733 @@
...
+        if (e) {
+          diag.error('amqplib instrumentation: consumerEndHook error', e);
+        }
+      },
+      true
+    );
+  }
+
+  private checkConsumeTimeoutOnChannel(channel: InstrumentationConsumeChannel) {
+    const currentTime = new Date().getTime();
```
Flarna, line 714: isn't `Date` sensitive regarding change of local system time? Maybe we should use the Otel build in performance timers instead which are guaranteed to be monotonic?

Flarna (approved):

Flarna:
@blumamir  I got interrupted a few times during review. Left two more comments, nothing blocking.

I have to admit that I didn't follow the amqplib patching details as I'm not familiar with this library.

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: Good catch, I agree about that. Changed to use `ROOT_CONTEXT`

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: fixed 👍

vmarchaud (approved):
overall lgtm (i'm no amplib expert though), i think you'll need to update the release please config as daniel did there though https://github.com/open-telemetry/opentelemetry-js-contrib/pull/921

blumamir:
> overall lgtm (i'm no amplib expert though), i think you'll need to update the release please config as daniel did there though #921

done 👍

Flarna:
@blumamir ci seems to be a bit unhappy on main:
https://github.com/open-telemetry/opentelemetry-js-contrib/runs/5535165359?check_suite_focus=true
https://github.com/open-telemetry/opentelemetry-js-contrib/runs/5535165302?check_suite_focus=true

blumamir:
> @blumamir ci seems to be a bit unhappy on main: https://github.com/open-telemetry/opentelemetry-js-contrib/runs/5535165359?check_suite_focus=true https://github.com/open-telemetry/opentelemetry-js-contrib/runs/5535165302?check_suite_focus=true

Oops, sorry about that. Checking what hppended

blumamir:
PR to fix the daily tests in CI: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/946

Issue #898, opened 2022-02-15
@opentelemetry/instrumentation-aws-sdk fails when using  BatchGetItem with ReturnConsumedCapacity set to NONE


pauloprestes:
### What version of OpenTelemetry are you using?
0.5.0

### What version of Node are you using?
v16.13.2

### What did you do?
Added @opentelemetry/instrumentation-aws-sdk with aws-sdk-js-v3, and doing a BatchGetItem to dynamo and setting [ReturnConsumedCapacity](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html#API_BatchGetItem_RequestSyntax) as NONE.

### What did you expect to see?
To open telemetry send status regarding my application.

### What did you see instead?
```
 TypeError: Cannot read properties of undefined (reading 'map')
    at DynamodbServiceExtension.responseHook (/workdir/node_modules/@opentelemetry/instrumentation-aws-sdk/src/services/dynamodb.ts:67:42)
    at ServicesExtensions.responseHook (/workdir/node_modules/@opentelemetry/instrumentation-aws-sdk/src/services/ServicesExtensions.ts:58:35)
    at /workdir/node_modules/@opentelemetry/instrumentation-aws-sdk/src/aws-sdk.ts:507:43
    at processTicksAndRejections (node:internal/process/task_queues:96:5) 
```
### Additional context
N/A

PR #899, opened 2022-02-15
 fix(opentelemetry-instrumentation-aws-sdk): error when ReturnConsumedCapacity is set to None

pauloprestes:

- Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/898

- check if function map exists on ConsumedCapacity property.

## Checklist


plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/dynamodb.ts
```diff
@@ -61,7 +61,7 @@ export class DynamodbServiceExtension implements ServiceExtension {
     const operation = response.request.commandName;
 
     if (operation === 'BatchGetItem') {
-      if ('ConsumedCapacity' in response.data) {
+      if ('ConsumedCapacity' in response.data && 'map' in response.data.ConsumedCapacity){
```
blumamir, line 64: Thanks for reporting it and creating a fix.
I think using the `in` operator to check if a function is found on the object is a bit non-ideomatic.
What do you think about changing it to: `if(Array.isArray(response.data?.ConsumedCapacity))`?
Also - is it possible to add a unit-test for this case?

pauloprestes: Thanks for your comment @blumamir . I've pushed some changes to address your comment. [667eade](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/899/commits/667eade433b7e9a49a756861b048b6ecbab9deef)

blumamir (approved):
LGTM, thanks again for the fix

blumamir:
@pauloprestes please run `npm run lint:fix`, thanks

PR #900, opened 2022-02-15
fix: install @types/form-data

Rauno56:

I honestly do not know what exactly caused the issue in the first place, because the first commit that had the issue on the `main` didn't have any code changes and I'm not able to repro that locally, but simply following the instructions `tsc` suggested seems to make the compiler chill.

Installed `@types/form-data` in the root.
Created https://github.com/open-telemetry/opentelemetry-js-contrib/issues/901 to remove it later.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "devDependencies": {
     "@commitlint/cli": "16.0.2",
     "@commitlint/config-conventional": "16.0.0",
+    "@types/form-data": "^2.5.0",
     "@typescript-eslint/eslint-plugin": "5.8.1",
     "@typescript-eslint/parser": "5.8.1",
     "colors": "1.4.0",


Flarna:
Most likely caused because `@types/node-fetch` version 2.6.0 was published which depends on `form-data` ^2.3.3 whereas the previous version 2.5.12 depended on `form-data` ^3.0.0.

I don't have this problem local therefore I assume some ci caching is additionally needed.

Rauno56:
> Most likely caused because @types/node-fetch version 2.6.0 was published which depends on form-data ^2.3.3 whereas the previous version 2.5.12 depended on form-data ^3.0.0.

I'm aware of that, but the confusing part is the fact that node-fetch is a dep of lerna and codecov - I wonder why does that affect our compilations.

Flarna:
there is also following in instrumentation tedious: tedious@14.1.0 => @azure/identity@1.5.2 => @azure/keyvault-keys@4.3.0 => @azure/core-http@2.2.4 => node-fetch@2.6.7 and @types/node-fetch@2.6.0

and the lerna hosting seems to move around/dedup stuff a bit.

Rauno56:
Thanks, @Flarna for your help bouncing this! I removed `codecov` instead of going with this hack for now. Removing it was due anyway because it has been deprecated for a while now.

Flarna:
welcome! I guess once it's done here we should do the same in opentelemetry-js and opentelemetry-js-api repos.

PR #906, opened 2022-02-16
chore: remove deprecated dependency of codecov

Rauno56:

`codecov` npm package is long deprecated and stated to stop working 1st of February(that's two weeks **ago**).
Coincidentally it is also solving our current issues with CI as well. #900

1. Generate the json report during testing itself. the amount of time it adds is negligible for local runs. Perhaps only in TAV runs one could potentially see any difference, but those should not be running nyc at all.
2. Use codecov GHA to upload the report. That means if one needs to upload reports locally, the official "binary" is ought to be used. I cannot imagine a usecase for that.
3. Added a `nyc` filter for `src/**` - we had tons of unrelated files in the source reports. Among others, the test files themselves.

## Checklist

- [ ] ~~Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.~~

closes #900


legendecas (approved):
This seems cool. We don't have to install codecov locally anymore.

Flarna (approved):

vmarchaud (approved):

PR #911, opened 2022-02-20
chore(aws-sdk): fix npm install on windows

trask:

- `npm install` fails on windows

- follow convention that is used in the other modules to hardcode the scope instead of using `$(npm pkg get name)`


NathanielRN (approved):
Thanks for finding this :)

blumamir (approved):

Flarna (approved):

Flarna:
same as #902

trask:
> same as #902

oh indeed! closing this in favor of #902

PR #912, opened 2022-02-21
chore: use db.system constants from semantic-conventions package

trask:

- DbSystemValues constants were used in some places but not in others

- use DbSystemValues everywhere for `db.system` attribute values

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
@@ -26,7 +26,10 @@ import {
   InstrumentationNodeModuleDefinition,
   isWrapped,
 } from '@opentelemetry/instrumentation';
-import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
+import {
+  DbSystemValues,
+  SemanticAttributes,
+} from '@opentelemetry/semantic-conventions';
 import type * as mysqlTypes from 'mysql';
 import { MySQLInstrumentationConfig } from './types';
 import { getConnectionAttributes, getDbStatement, getSpanName } from './utils';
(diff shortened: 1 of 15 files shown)


plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -44,7 +47,7 @@ export class MySQLInstrumentation extends InstrumentationBase<
 > {
   static readonly COMPONENT = 'mysql';
```
blumamir, line 48: I think it's not used anywhere else. Can be removed

blumamir (approved):

plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
```diff
@@ -102,7 +105,7 @@ export const getTracedInternalSendCommand = (
       {
         kind: SpanKind.CLIENT,
         attributes: {
-          [SemanticAttributes.DB_SYSTEM]: RedisInstrumentation.COMPONENT,
+          [SemanticAttributes.DB_SYSTEM]: DbSystemValues.REDIS,
```
Flarna, line 108: is `RedisInstrumentation.COMPONENT` still used anywhere?  if not it should be removed.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
trask: this one is still used

dyladan:
@trask just confirming `chore`. Our automation uses the title of the PR to automatically determine the semver version change and `chore` does not cause a release because there should be no change in behavior. It seems to me thats fine here, but I want to confirm this is what you intended.

trask:
> @trask just confirming `chore`. Our automation uses the title of the PR to automatically determine the semver version change and `chore` does not cause a release because there should be no change in behavior. It seems to me thats fine here, but I want to confirm this is what you intended.

ya, there should be no change in behavior, the code was already using the correct values, just not pulling them from the semantic-conventions package (whether I _intentionally_ used `chore` correctly is a different question 😂 so thx for checking)

PR #913, opened 2022-02-21
fix!(mysql*,redis): net.peer.ip -> db.connection_string

trask:

- mysql, mysql2, redis and ioredis instrumentation was capturing the connection string in the `net.peer.ip` attribute

- changes these instrumentations (except ioredis, because that change seems to be hanging CI not sure why, I'll send PR for that change separately) to capture the connection string in the `db.connection_string` attribute

[I have other thoughts about whether `db.connection_string` should be captured at all in these cases (when it's a synthetically reconstructed value from existing telemetry data like `net.peer.name`, `net.peer.port`, `db.name`), but I will seek to clarify that point in the specification as it's unclear (see also open-telemetry/opentelemetry-java-instrumentation#3080)]

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
+++ b/plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
@@ -33,7 +33,7 @@
     "compile": "npm run version:update && tsc -p .",
     "lint": "eslint . --ext .ts",
     "lint:fix": "eslint . --ext .ts --fix",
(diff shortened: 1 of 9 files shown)


dyladan (approved):

dyladan:
WRT `fix` conventional commit subject, I think it might be better to use `fix!` to indicate this is a breaking change which will bump the semver minor (treated as major before 1.0) version.

Flarna (approved):

PR #877, opened 2022-02-10
feat: allow custom aws sdk service extensions via config (#829)

chrisrichardsevergreen:
## Which problem is this PR solving?
Implements #829 - allowing the end user to provide custom service extensions, meaning trace semantic conventions can be added by the user to a much larger array of AWS services that could be reasonably expected to be built into this package. It also also for users to 'tweak' the semantics of existing service specific logic.

## Short description of the changes

- Add config parameter `customServiceExtensions`, and merge these into the existing service extensions (as implemented this allows overriding the built-in ones, however this is definitely open for debate!)
- Expose the `ServiceExtension` type and dependent types on the package interface, as well as the service specific types so these can be built upon
- Add tests to the handling of these custom service extensions

## Checklist

- [ x ] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.

PR #916, opened 2022-02-22
feat(aws-sdk): lambda client instrumentation

chrisrichardsevergreen:
Implements instrumentation of the AWS SDK 'Lambda' calls (invoke) using FAAS trace semantics - this follows on from discussions on PR #877 

- Add lambda service extension adding faas_* attributes to the request span, and propagating the context into the ClientContext custom field on the Invoke request.

## Checklist


plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -48,29 +48,32 @@
   "dependencies": {
     "@opentelemetry/core": "^1.0.0",
     "@opentelemetry/instrumentation": "^0.27.0",
+    "@opentelemetry/propagation-utils": "^0.26.0",
     "@opentelemetry/semantic-conventions": "^1.0.0",
-    "@opentelemetry/propagation-utils": "^0.26.0"
+    "aws-lambda": "^1.0.7"
```
blumamir, line 53: is it used anywhere?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+      case 'Invoke':
+        spanAttributes = {
+          [SemanticAttributes.FAAS_INVOKED_NAME]: functionName,
+          [SemanticAttributes.FAAS_INVOKED_PROVIDER]: 'aws',
+        };
+        if (request.region) {
+          spanAttributes = {
+            ...spanAttributes,
+            [SemanticAttributes.FAAS_INVOKED_REGION]: request.region,
+          };
```
blumamir, line 44: Can this be simplified to:
```js
spanAttributes[SemanticAttributes.FAAS_INVOKED_REGION] = request.region
```
?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+  responseHook(
+    response: NormalizedResponse,
+    span: Span,
+    tracer: Tracer,
+    config: AwsSdkInstrumentationConfig
+  ) {
+    const operation = response.request.commandName;
+
+    if (operation === 'Invoke') {
+      if (response.data && '$metadata' in response.data) {
```
blumamir, line 76: I think `$metadata` is only available for v3.
For v2 we need to [use the equivalent](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/760c4d4e9bd500ab65980f7414f72ad93da42e2f/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts#L331): `response.requestId`

blumamir: A better alternative if you are up for it, is to extract the `requestId` into the `normalizedResponse` so it can be used generically everywhere it is needed without repeating any code. This is recommended but optional.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+    if (encodedPayload.length <= 3583) {
+      carrier.ClientContext = encodedPayload;
+    }
+  }
+}
+const contextSetter = new ContextSetter();
+
+const injectPropagationContext = (
+  invocationRequest: Record<string, any>
```
blumamir, line 115: Can we use the actual type here? e.g. `InvocationRequest`

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  extractFunctionName = (commandInput: Record<string, any>): string => {
+    return commandInput?.FunctionName;
+  };
+}
+
+class ContextSetter implements TextMapSetter<Record<string, any>> {
+  set(carrier: Record<string, any>, key: string, value: string): void {
+    const parsedClientContext: ClientContext = JSON.parse(
+      carrier.ClientContext !== undefined
```
blumamir, line 93: Maybe we can place this check outside of the `JSON.parse`, so we don't need to execute redundant parse on `{"Custom":{}}`?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  extractFunctionName = (commandInput: Record<string, any>): string => {
+    return commandInput?.FunctionName;
+  };
+}
+
+class ContextSetter implements TextMapSetter<Record<string, any>> {
+  set(carrier: Record<string, any>, key: string, value: string): void {
+    const parsedClientContext: ClientContext = JSON.parse(
```
blumamir, line 92: Should we wrap it in try / catch?
What will happen if `JSON.parse` throws?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+      carrier.ClientContext !== undefined
+        ? Buffer.from(carrier.ClientContext, 'base64').toString('utf8')
+        : '{"Custom":{}}'
+    );
+    const updatedPayload = {
+      ...parsedClientContext,
+      Custom: {
+        ...parsedClientContext.Custom,
```
blumamir, line 100: Is it possible that `parsedClientContext.Custom` is undefined? In this case this code will throw I think

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+class ContextSetter implements TextMapSetter<Record<string, any>> {
+  set(carrier: Record<string, any>, key: string, value: string): void {
+    const parsedClientContext: ClientContext = JSON.parse(
+      carrier.ClientContext !== undefined
+        ? Buffer.from(carrier.ClientContext, 'base64').toString('utf8')
+        : '{"Custom":{}}'
+    );
+    const updatedPayload = {
+      ...parsedClientContext,
+      Custom: {
```
blumamir, line 99: Where is the code in `@opentelemetry/instrumentation-aws-lambda` that extracts the context from this `Custom` field?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+        ...parsedClientContext.Custom,
+        [key]: value,
+      },
+    };
+    const encodedPayload = Buffer.from(JSON.stringify(updatedPayload)).toString(
+      'base64'
+    );
+    if (encodedPayload.length <= 3583) {
```
blumamir, line 107: Can be great to add a comment with reference to this magic number :)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,119 @@
...
+      carrier.ClientContext = encodedPayload;
+    }
```
blumamir, line 109: Do we want to log anything to diag logger in case context cannot be injected?
Simialar to [this](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/760c4d4e9bd500ab65980f7414f72ad93da42e2f/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/MessageAttributes.ts#L80)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: I've got a test covering this (lambda.test.ts:302) - it didn't require any changes which I was anticipating. Happy to add a check in anyway as it's probably better to explicitly handle this case.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: Added

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: It doesn't throw if Custom is undefined.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: Went for the recommended alternative

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
chrisrichardsevergreen: It was used for the ClientContext type, however since this was removed I've now removed the package.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: The `eventContextExtractor` configuration parameter to `AwsLambdaInstrumentation ` in `@opentelemetry/instrumentation-aws-lambda` allows this to be extracted from context. There's an example implementation of this in the tests in [`@opentelemetry/instrumentation-aws-lambda`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/760c4d4e9bd500ab65980f7414f72ad93da42e2f/plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts#L660)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
```diff
@@ -503,6 +505,7 @@ export class AwsInstrumentation extends InstrumentationBase<typeof AWS> {
                   const normalizedResponse: NormalizedResponse = {
                     data: response.output,
                     request: normalizedRequest,
+                    requestId: requestId,
```
blumamir, line 508: If I remember correctly, the response sometimes only includes the `extendedRequestId` and lacks the `requestId` field. not sure what is the difference between them and when each one is used.
Just verifying that you tested it against aws and found that the request id is indeed the value to use for lambda client.
@NathanielRN @willarmiros - any insights on this maybe?

plugins/node/opentelemetry-instrumentation-aws-sdk/test/lambda.test.ts
```diff
@@ -0,0 +1,390 @@
...
+
+        expect(request).toBeDefined();
+        const requestHeaders = request!.headers;
+        const clientContext = JSON.parse(
+          Buffer.from(
+            requestHeaders['x-amz-client-context'],
+            'base64'
+          ).toString()
+        ) as Record<string, any>;
+        expect(clientContext.Custom).toHaveProperty('existing', 'data');
```
blumamir, line 243: maybe also expect `traceparent` here to make sure it's injected in this case?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,133 @@
...
+          [key]: value,
+        },
+      };
+      const encodedPayload = Buffer.from(
+        JSON.stringify(updatedPayload)
+      ).toString('base64');
+
+      // The length of client context is capped at 3583 bytes of base64 encoded data
+      // (https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_RequestSyntax)
+      if (encodedPayload.length > 3583) {
```
blumamir, line 111: If the function is invoked multiple times (for example, `traceparent` and `tracestate` for `W3CTraceContextPropagator`, or the 4 fields of `B3MultiPropagator` propagator), it is possible that only part of the context will be injected and some will be missing after the limit is reached.
For SQS / SNS, [the inject function is first checking if all context fields can be injected successfully and only then attempts any modifications to the request](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/986c3499b324ebdf49aabf35bc3711c91bb91ec8/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/MessageAttributes.ts#L74).

What do you think about using a similar approach here as well?

This will also be more performant, as currently, we decode the base64 string, JSON.parse it, then JSON stringify it and encode it back to base64, again and again for each field.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
blumamir: Amazing, thanks :)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
blumamir: So if the user will install the `aws-sdk` and `aws-lambda` instrumentations, he will not get automatic context propagation, but will have to add specific code for that, right?
I think we could add the extraction to the `aws-lambda` [_defaultEventContextExtractor](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/986c3499b324ebdf49aabf35bc3711c91bb91ec8/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L327) to give users a better experience. Not required for this PR, we can open an issue to add it and someone will do it in the future (you are welcome to do that yourself if you have motivation and time :) )

BTW, do you know maybe if this `Custom` field is something standard in AWS? Is it used in other contexts or was introduced only for OTEL? Do we know if instrumentations in other languages use it as well? @NathanielRN @willarmiros

blumamir (commented):
Thanks for addressing everything and for writing all these high-quality tests 🥇 
We are almost ready, added a few more minor comments.
@NathanielRN @willarmiros  do you want to review as well?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: I'll add it to my list of PRs integrating honeycomb tracing via OTEL has spawned 😄 

I believe `Custom` to be standard - it's not wonderfully documented from the invoker's perspective, but is mentioned in the [node developer guide for lambda](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html), and I've done some cross-language checks and e.g. Python also supports it.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: I noted that approach, but wasn't sure if there was a precise enough equivalent to the 'will this send it over the limit' question, when we first don't know the length of the key + value (the SQS/SNS equivalent is for max number of keys, the values are free in length, so don't need checking) and we would have to use a bodgy encoded => base64 length mapping formula. I'll give it a try to see if I can get all tests passing but it might need a final check before setting just in case despite best endeavours it still has managed to go over the limit.

blumamir: What I imagined is something like this (haven't really run it, it's just a scatch):
```js
const injectPropagationContext = (
  invocationRequest: InvocationRequest
): InvocationRequest => {
  try {
    const propagationKeysValues = {};
    propagation.inject(context.active(), propagationKeysValues);

    const parsedClientContext = invocationRequest.ClientContext
    ? JSON.parse(
        Buffer.from(invocationRequest.ClientContext, 'base64').toString('utf8')
      )
    : {};

    const updatedClientContext = {
      ...parsedClientContext,
      Custom: {
        ...parsedClientContext.Custom,
        ...propagationKeysValues,
      },
    };

    const encodedClientContext = Buffer.from(
      JSON.stringify(updatedClientContext)
    ).toString('base64');

    // The length of client context is capped at 3583 bytes of base64 encoded data
    // (https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_RequestSyntax)
    if (encodedClientContext.length > 3583) {
      diag.warn(
        'lambda instrumentation: cannot set context propagation on lambda invoke parameters due to ClientContext length limitations.'
      );
      return invocationRequest;      
    }

    return { 
      ...invocationRequest,
      ClientContext: encodedClientContext,
    }

  } catch(err) {
    diag.debug(
      'lambda instrumentation: failed to set context propagation on ClientContext',
      err
    );
    return invocationRequest;
  }
};
```

This code will either inject the entire propagation data, or none of it. It will also `JSON.parse` + `Buffer.from` and `JSON.stringify` + `toString('base64')` the client context at most once, regardless of the number of propagation keys.
It will also create a copy of the request params, instead of modifying the user's object which might introduce side effects

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
blumamir: awsome! thanks for the research and for the willingness to fix that in `aws-lambda` instrumentation as well :)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
willarmiros: Extended request ID is only included for S3 requests - at that I'm not even sure if the extended ID is included in responses or only errors.

See: AWS SDK v2 [error](https://github.com/aws/aws-sdk-js/blob/307e82673b48577fce4389e4ce03f95064e8fe0d/lib/error.d.ts#L44) vs. [response](https://github.com/aws/aws-sdk-js/blob/master/lib/response.js) (which has no extended ID field)

blumamir: So my tests were probably around S3 in V3 and that's why I saw extended id instead of request id. Thanks for clarifying :)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  responseHook(
+    response: NormalizedResponse,
+    span: Span,
+    tracer: Tracer,
+    config: AwsSdkInstrumentationConfig
+  ) {
+    const operation = response.request.commandName;
+
+    if (operation === 'Invoke') {
```
NathanielRN, line 81: Should this be a switch statement too?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,133 @@
...
+      isIncoming: false,
+      spanAttributes,
+      spanKind: SpanKind.CLIENT,
+      spanName,
+    };
+  }
+
+  requestPostSpanHook = (request: NormalizedRequest) => {
+    switch (request.commandName) {
+      case 'Invoke':
```
NathanielRN, line 63: Can this and the string below be a static constant at the top of the file?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    if (operation === 'Invoke') {
+      span.setAttribute(SemanticAttributes.FAAS_EXECUTION, response.requestId);
+    }
+  }
+
+  extractFunctionName = (commandInput: Record<string, any>): string => {
+    return commandInput?.FunctionName;
+  };
+}
```
NathanielRN, line 89: Probably a space below this would make sense?

NathanielRN (approved):
It looks good to me from a high level. I added a few comments, but I'll defer to @blumamir for the final review.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: The linter keeps removing that, so unfortunately I can't

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
```diff
@@ -0,0 +1,133 @@
...
+      case 'Invoke':
+        spanAttributes = {
+          [SemanticAttributes.FAAS_INVOKED_NAME]: functionName,
+          [SemanticAttributes.FAAS_INVOKED_PROVIDER]: 'aws',
+        };
+        if (request.region) {
+          spanAttributes[SemanticAttributes.FAAS_INVOKED_REGION] =
+            request.region;
+        }
+        spanName = `${functionName} invoke`;
```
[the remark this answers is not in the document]
chrisrichardsevergreen: Updated

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts
chrisrichardsevergreen: Updated

blumamir (approved):

PR #920, opened 2022-02-25
chore: release main

dyladan:
:robot: I have created a release *beep* *boop*
---

<details><summary>@opentelemetry/auto-instrumentations-node: 0.27.3</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/instrumentation-aws-sdk bumped from ^0.5.0 to ^0.5.1
    * @opentelemetry/instrumentation-fastify bumped from ^0.25.0 to ^0.26.0
    * @opentelemetry/instrumentation-graphql bumped from ^0.27.3 to ^0.27.4
    * @opentelemetry/instrumentation-ioredis bumped from ^0.27.1 to ^0.28.0
    * @opentelemetry/instrumentation-mongodb bumped from ^0.28.0 to ^0.29.0
    * @opentelemetry/instrumentation-mysql bumped from ^0.27.1 to ^0.28.0
    * @opentelemetry/instrumentation-mysql2 bumped from ^0.28.0 to ^0.29.0
    * @opentelemetry/instrumentation-redis bumped from ^0.28.0 to ^0.29.0
    * @opentelemetry/instrumentation-winston bumped from ^0.27.0 to ^0.27.2
</details>

<details><summary>browser-extension-autoinjection: 0.27.4</summary>

### Bug Fixes

* **browser-extension-autoinjection:** update eslint-plugin-json5 version to 0.1.4 ([#896](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/896)) ([e47fcaa](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/e47fcaa1a04de35b096f373f44abbf87ff4125f2))
</details>

<details><summary>@opentelemetry/instrumentation-aws-sdk: 0.5.0</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/propagation-utils bumped from ^0.26.0 to ^1.0.0
</details>

<details><summary>instrumentation-fastify: 0.26.0</summary>

### Features

* new fastify instrumentation ([#611](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/611)) ([77c215b](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/77c215bdd7adb76c8934028458a2b7f28e041f37))
[...]

PR #921, opened 2022-02-25
chore: update manifest to include missing packages

dyladan:
Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/pull/920#pullrequestreview-894051020


vmarchaud (approved):

blumamir (approved):
Thanks!
so when adding new packages, the manifest should be updated as well as the `release-please-config.json` file, right?
If so I'll update the `amqplib` PR

dyladan:
> Thanks! so when adding new packages, the manifest should be updated as well as the `release-please-config.json` file, right? If so I'll update the `amqplib` PR

if you leave it off it uses 1.0.0 as the default first version for release. this might be configurable.

PR #930, opened 2022-03-02
chore: remove unnecessary version:update scripts and don't check in fastify version.ts

dyladan:
The fastify version.ts file prevented the release from completing smoothly. When removing it, I noticed that the version scripts are inconsistent among instrumentations and unused in non-instrumentation packages. I removed the unused version:update scripts to prevent thrash in CI, fixed the fastify version script, and made the version update scripts consistent everywhere.


Flarna:
Why do we need the version update only in instrumentation packages but not in e.g. resource detectors?

dyladan:
> Why do we need the version update only in instrumentation packages but not in e.g. resource detectors?

The instrumentation packages need to know their version in order to get a meter using the `getMeter(name, version)` API. Other packages aren't using that API.

Flarna (approved):

vmarchaud (approved):

Rauno56 (approved):

Rauno56:
@dyladan, I don't have the permission to update the PR, otherwise would update and merge.

Issue #907, opened 2022-02-16
Add detector for docker container id for cgroup v1 to be added to resource metadata


abhee11:
### Is your instrumentation request related to a problem? Please describe

Currently there is no support for adding container id to resource metadata by itself. The logic exists in aws detector but not as a standalone module. 

### Is it applicable for Node or Browser or both?
Node

### Do you expect this instrumentation to be commonly used?
Weekly Downloads:

### What version of instrumentation are you interested in using?
Versions:

abhee11:
I am currently working on it. I should be able to open a PR soon.

abhee11:
This enhancement was added and merged .

PR #931, opened 2022-03-03
feat: add docker resource detector

abhee11:

- [abstracting the docker container id logic ](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/907)

- adding a new detector for docker container id cgroupov1 only 

## Checklist


detectors/node/opentelemetry-resource-detector-docker/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "typescript": "^4.5.5"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.0.4"
+  },
+  "dependencies": {
+    "@opentelemetry/core": "^1.0.1",
+    "@opentelemetry/resources": "^1.0.1",
+    "@opentelemetry/semantic-conventions": "^1.0.1",
+    "gts": "^3.1.0"
```
vmarchaud, line 56: this should be a dev dependencies

detectors/node/opentelemetry-resource-detector-docker/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@opentelemetry/resource-detector-docker",
+  "version": "1.0.0",
```
vmarchaud, line 3: not sure we can release directly as 1.0 ? waiting for other maintainers input on this

detectors/node/opentelemetry-resource-detector-docker/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "rimraf": "3.0.2",
+    "sinon": "^12.0.1",
+    "ts-mocha": "8.0.0",
+    "typescript": "^4.5.5"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.0.4"
+  },
+  "dependencies": {
+    "@opentelemetry/core": "^1.0.1",
```
vmarchaud, line 53: i believe this isn't used

vmarchaud (commented):
the detector itself looks good to me, you'll need to fix the CLA though

detectors/node/opentelemetry-resource-detector-docker/package.json
abhee11: done

detectors/node/opentelemetry-resource-detector-docker/package.json
abhee11: makes sense - done

vmarchaud (commented):
you'll also need to add there: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.release-please-manifest.json so it can be automatically published

dyladan:
> you'll also need to add there: [https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.release-please-manifest.json](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.release-please-manifest.json?rgh-link-date=2022-03-07T17%3A52%3A36Z) so it can be automatically published

I think this isn't true anymore with release-please v3

abhee11:
Just cleaning out some lint issues - waiting for jobs to run - @dyladan @vmarchaud 
I updated the .release-please-manifest.json as per the comment - let me know if you like me to revert it.

I appreciate the review on this - thank you

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
```diff
@@ -0,0 +1,91 @@
...
+  readonly DEFAULT_CGROUP_PATH = '/proc/self/cgroup';
+  readonly TIMEOUT_MS = 2000;
+  readonly UTF8_UNICODE = 'utf8';
+
+  private static readFileAsync = util.promisify(fs.readFile);
+  private static fileAccessAsync = util.promisify(fs.access);
+
+  async detect(_config?: ResourceDetectionConfig): Promise<Resource> {
+    try {
+      if (!(await this._isSupportedDocker())) {
```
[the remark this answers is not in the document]
Rauno56: [fs docs](https://nodejs.org/api/fs.html#fspromisesaccesspath-mode) seem to agree:

> Using fsPromises.access() to check for the accessibility of a file before calling fsPromises.open() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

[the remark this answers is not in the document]
abhee11: @Rauno56 That is a good insight , thank you for the comment. I will update the PR and notify .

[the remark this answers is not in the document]
abhee11: So - I will remove the check itself and if there is any issue - it should be caught in the catch. Right ?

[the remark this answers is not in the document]
Rauno56: Yes - The catch surrounding `readFile`.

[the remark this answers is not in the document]
abhee11: done

abhee11:
@dyladan @Rauno56 Seems like lint is failing - I tested my local with "lerna run lint" and it fails on the "aws" package - but in the logs it says it fails on "docker" package - not able to replicate ? any advice ?


npm run lint runs fine on my local

vmarchaud:
> I think this isn't true anymore with release-please v3

Didn't saw that sorry @abhee11

Rauno56:
> @dyladan @Rauno56 Seems like lint is failing - I tested my local with "lerna run lint" and it fails on the "aws" package - but in the logs it says it fails on "docker" package - not able to replicate ? any advice ?
> 
> npm run lint runs fine on my local

Messages you are referring to(from aws) are only warnings and do not cause the failure. The relevant part for you is this:

```
> @opentelemetry/resource-detector-docker@1.0.0 lint /__w/opentelemetry-js-contrib/opentelemetry-js-contrib/detectors/node/opentelemetry-resource-detector-docker
> eslint . --ext .ts


lerna ERR! npm run lint stderr:

Oops! Something went wrong! :(

ESLint: 8.7.0

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    npm init @eslint/config

ESLint looked for configuration files in /__w/opentelemetry-js-contrib/opentelemetry-js-contrib/detectors/node/opentelemetry-resource-detector-docker/src/detectors and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint chat room: https://eslint.org/chat/help
```

Make sure all the relevant files are checked in and paths in them specified correctly.

abhee11:
@Rauno56 Thanks - I was missing the eslintrc and ignore files - added them.

Rauno56 (approved):

detectors/node/opentelemetry-resource-detector-docker/.eslintignore
```diff
@@ -0,0 +1 @@
+build
```
dyladan, line 1: nit: add newline at the end

detectors/node/opentelemetry-resource-detector-docker/.npmignore
```diff
@@ -0,0 +1,4 @@
+/bin
+/coverage
+/doc
+/test
```
dyladan, line 4: nit: newline

detectors/node/opentelemetry-resource-detector-docker/package.json
dyladan: I would definitely prefer to start with 0.1.0

detectors/node/opentelemetry-resource-detector-docker/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@opentelemetry/resource-detector-docker",
+  "version": "1.0.0",
+  "description": "Opentelemetry resource detector to get docker containerId",
```
dyladan, line 4: There may be other things added in the future. I would make this more general
```suggestion
  "description": "OpenTelemetry resource detector for docker resource attributes",
```

detectors/node/opentelemetry-resource-detector-docker/README.md
```diff
@@ -0,0 +1,48 @@
...
+import { detectResources } from '@opentelemetry/resources';
+import { dockerCGroupV1Detector } from '@opentelemetry/resource-detector-docker'
+const resource = await detectResources({
+   detectors: [dockerCGroupV1Detector],
+})
+
+const tracerProvider = new NodeTracerProvider({ resource });
+```
+
+## Available detectors
```
dyladan, line 32: A detector per attribute? Seems like it would be better to have a `DockerDetector` and list the attributes it detects here

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  private async _getContainerId(): Promise<string | undefined> {
+    try {
+      const rawData = await DockerCGroupV1Detector.readFileAsync(
+        this.DEFAULT_CGROUP_PATH,
+        this.UTF8_UNICODE
+      );
+      const splitData = rawData.trim().split('\n');
```
dyladan, line 56: is it always \n or can it be \r\n?
```suggestion
      const splitData = rawData.trim().split(/\r?\n/);
```

dyladan (changes requested):
Please use 0.1.0 as the initial version. We can bump it to 1.0.0 when we know it is stable. Blocking for the 1.0.0 issue.

Flarna:
I guess release-please-config.json needs to be adapted to get this published after merge.

detectors/node/opentelemetry-resource-detector-docker/README.md
abhee11: @dyladan thank you for the comments
Currently I have named it as DockerCGroupV1Detector given it is specific to CGroup v1 
We currently just have container ID that is added as a part of resouce, In future we could definitely add more attributes but I think it is important to mark the name CGroupV1 

Do you agree ?

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
abhee11: Because docker is linux based we should be okay with \n

detectors/node/opentelemetry-resource-detector-docker/.npmignore
abhee11: done

detectors/node/opentelemetry-resource-detector-docker/package.json
abhee11: makes sense - done

abhee11:
@Flarna As per comments above - I believe we dont need that anymore ?
@dyladan Please review changes and my comments for \r and the READ.ME change - let me know what you think

Flarna:
@abhee11 no need anymore to touch `.release-please-manifest.json` but still needed to add a new package in `release-please-config.json`.

See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/938 which shows that this is still needed.

abhee11:
@Flarna my bad - pushed - thanks

detectors/node/opentelemetry-resource-detector-docker/README.md
dyladan: OK yeah thats fine

dyladan (approved):

vmarchaud (approved):

detectors/node/opentelemetry-resource-detector-docker/README.md
```diff
@@ -0,0 +1,48 @@
+# OpenTelemetry Resource Detector for Docker
+
+[![NPM Published Version][npm-img]][npm-url]
+[![dependencies][dependencies-image]][dependencies-url]
+[![devDependencies][devDependencies-image]][devDependencies-url]
+[![Apache License][license-image]][license-image]
+
+[component owners](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/component_owners.yml): @abhee11
+
+Resource detector for docker.
```
blumamir, line 10: According to [the semantic conventions for resource container](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/container.md), the following 5 attributes belong to the `docker` resource:
```
container.name, container.id, container.runtime, container.image.name, container.image.tag
```

The current implementation only extracts a single value (`container.id`) of these 5 and only for `CGroupV1` (not sure how many other types are possible).

I suggest to make that clear in the README, or support the other fields in the spec

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
```diff
@@ -0,0 +1,75 @@
...
+import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+
+import * as fs from 'fs';
+import * as util from 'util';
+import { diag } from '@opentelemetry/api';
+
+export class DockerCGroupV1Detector implements Detector {
+  readonly CONTAINER_ID_LENGTH = 64;
+  readonly DEFAULT_CGROUP_PATH = '/proc/self/cgroup';
+  readonly TIMEOUT_MS = 2000;
```
blumamir, line 31: If I'm not mistaken, this is not used anywhere

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
```diff
@@ -0,0 +1,75 @@
...
+  async detect(_config?: ResourceDetectionConfig): Promise<Resource> {
+    try {
+      const containerId = await this._getContainerId();
+      return !containerId
+        ? Resource.empty()
+        : new Resource({
+            [SemanticResourceAttributes.CONTAINER_ID]: containerId || '',
+          });
+    } catch (e) {
+      diag.warn('Process is not running on a supported docker version', e);
```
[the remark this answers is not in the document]
abhee11: So - the reason for this is - if you are installing DockerResourceV1Detector - that would mean - you are adding this dependency assuming you have v1 version running - hence the warning !

As this dependency would be of no use if it did get them the container id - hence we need to warn them that they need have a diff docker version running.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private async _getContainerId(): Promise<string | undefined> {
+    try {
+      const rawData = await DockerCGroupV1Detector.readFileAsync(
+        this.DEFAULT_CGROUP_PATH,
+        this.UTF8_UNICODE
+      );
+      const splitData = rawData.trim().split('\n');
+      for (const str of splitData) {
+        if (str.length >= this.CONTAINER_ID_LENGTH) {
```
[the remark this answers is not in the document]
abhee11: Yes - for group v1 - this seems to be the implementation across java , go etc - hence I adapted the same for JS

detectors/node/opentelemetry-resource-detector-docker/README.md
abhee11: currently in the readme it does say that we populate only container id - I will add more details in case if it is missing - thanks

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
```diff
@@ -0,0 +1,75 @@
...
+          return str.substring(str.length - this.CONTAINER_ID_LENGTH);
+        }
+      }
+    } catch (e) {
+      let errorMessage = '';
+      if (e instanceof Error) {
+        errorMessage = e.message;
+        diag.warn(
+          `Docker CGROUP V1 Detector failed to read container ID: ${errorMessage}`
+        );
```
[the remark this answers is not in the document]
abhee11: We need a way to let the user know that their dockerV1Detector is not doing as they hoped. I am wondering if their a another way - do you have any suggestions ?

detectors/node/opentelemetry-resource-detector-docker/test/DockerCGroupv1Detector.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  let readStub;
+  const correctCgroupData =
+    'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklm';
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe('Supported docker - Container ID ', () => {
+    it('should return a emoty resource - docker cgroup v1 detector', async () => {
```
blumamir, line 38: nit: `emoty` -> `empty` :)

detectors/node/opentelemetry-resource-detector-docker/test/DockerCGroupv1Detector.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    it('should return an empty resource when containerId is not valid', async () => {
+      const errorMsg = {
+        fileNotFoundError: new Error('cannot file in the path}'),
+      };
+
+      readStub = sinon
+        .stub(DockerCGroupV1Detector, 'readFileAsync' as any)
+        .onSecondCall()
+        .rejects(errorMsg);
```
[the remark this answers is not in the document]
blumamir: Maybe I misunderstand you, but the test uses:
```js
      const errorMsg = {
        fileNotFoundError: new Error('cannot file in the path}'),
      };
```

And it will not return an error, but an object with one key named `fileNotFoundError` which is an `Error`.
Is it consistent with what `fs` returns on `err`?
I would expect it to be something like:
```js
.rejects(new Error('foo bar'));
```

detectors/node/opentelemetry-resource-detector-docker/test/DockerCGroupv1Detector.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  const correctCgroupData =
+    'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklm';
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe('Supported docker - Container ID ', () => {
+    it('should return a emoty resource - docker cgroup v1 detector', async () => {
+      const resource: Resource = await dockerCGroupV1Detector.detect();
```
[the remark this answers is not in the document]
abhee11: Sorry - I don't understand this comment - can you explain a little more
It should return { attributes: {} } as resource if we do not give it the right file path ( in case of test - the right stub which resolves into the container id  ) ?

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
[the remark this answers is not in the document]
blumamir: I am wondering if including a resource detector means that the user is expecting that resource to be present. 

2 Examples: 
1. If the user installs an aws detector and then runs his service locally, I wouldn't expect any warnings to be printed to log as this is completely valid and the message will be confusing. Maybe the service is running inside a docker in production but not in other environments.
2. If someone creates a distribution of OpenTelemetry and includes many detectors, the expectation (as far as I am aware) is that only the relevant detectors will be applied. Detectors that are not relevant should be silently ignored. This is the way instrumentations are handled: if someone installs the `@opentelemetry/auto-instrumentations-node` package with all instrumentations, and then his app does not require `redis` package, then `instrumentation-redis` will just not load with no warning to diag channel

I think emitting it as `info` makes more sense, something like: "Docker CGROUP V1 Detector did not identify running inside a docker container, no docker attributes will be added to resource".

This is for the case that the file does not exist. If the file exists but exception is thrown for any other reason, I would expect an `error` with message that includes the `errorMessage`.

This is my take on it, but I'm open for other opinions

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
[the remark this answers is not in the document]
blumamir: See my comment on the other thread

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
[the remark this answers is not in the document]
blumamir: Quoting from the `DiagLogger` interface:

```js
  /**
   * Log a warning scenario to inform the developer of an issues that should be investigated.
   * The requested operation may or may not have succeeded or completed.
   */
  warn: DiagLogFunction;
```

I think when the file is missing it doesn't mean that the developer should investigate anything but only be informed

detectors/node/opentelemetry-resource-detector-docker/test/DockerCGroupv1Detector.test.ts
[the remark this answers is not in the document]
blumamir: Maybe it's not really a problem 🤷‍♂️, but I can't understand why the invocation of this detector does not return non-empty resource when run in CI.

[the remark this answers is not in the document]
abhee11: I have added a stub to return undefined - just to be more careful

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerCGroupV1Detector.ts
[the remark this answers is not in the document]
abhee11: This is very helpful - thank you for taking the time and helping me out with this. I appreciate it. 
Addressed.

abhee11:
@blumamir I have addressed your comments and pushed the changes. Let me know if there is anything else.

detectors/node/opentelemetry-resource-detector-docker/test/DockerCGroupv1Detector.test.ts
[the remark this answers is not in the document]
dyladan: Does the resource detector require some sort of elevated permission or security context? I'm also curious why this returns undefined in CI because our CI runs inside docker containers.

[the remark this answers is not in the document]
abhee11: what version of docker is it running on ? if it is v2 - then this is expected to return undefined as the cgroup path is different and may need permission.

abhee11:
@blumamir @dyladan Let me know if there is anything more I need to refactor - I have made the proposed changes

blumamir (approved):
LGTM. thanks for the patient and for addressing everything

abhee11:
@blumamir Thank you, feel free to merge.

PR #936, opened 2022-03-06
feat: loosen up peer api check to require dev ver to satisfy peer ver

Rauno56:

Ideally, we'd specify the peer dependency for OTel JS API to the lowest range that the lib works with to provide the user with the maximum flexibility possible. The `peer-api-check.js` script currently works against it by requesting the peer dependency to match the version used in development.

Loosen up peer API check script to allow

1. latest version in dev deps at the same time with
2. the lowest required API version in peer deps.

## Checklist

- [ ] ~Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.~


scripts/peer-api-check.js
```diff
@@ -32,10 +34,10 @@ if (peerVersion) {
       `Package ${pjson.name} does't have API version pinned in dev dependencies: ${devVersion}`
     );
   }
-  if (peerVersion !== `^${devVersion}`) {
+  if (!semver.satisfies(devVersion, peerVersion)) {
     throw new Error(
       `Package ${pjson.name} depends on peer API version ${peerVersion} ` +
-      `but version ${devVersion} in development`
+      `but version ${devVersion} in development which doesn't satisfy the peer API version`
```
Flarna, line 40:
```suggestion
      `but version ${devVersion} in development which doesn't satisfy the peer API version ${peerVersion}`
```

Rauno56: Thanks for the comment. The error message mentions `peerVersion` in the first part of the message already. Perhaps we could rewrite the whole thing better without duplicating?

Flarna: Sorry I saw this too late. and resolved the conversion just a minute ago.

vmarchaud (approved):

Flarna (approved):

Issue #552, opened 2021-06-29
Move package-specific examples under the respective packages


rauno56:
- [x] This only affects the JavaScript OpenTelemetry library
- [ ] This may affect other libraries, but I would like to get opinions here first

I'd like to propose moving package-specific examples(eg. [examples/express](examples/express)) to under the package itself(eg. [pkg/express](plugins/node/opentelemetry-instrumentation-express)), closer to the implementation, test, etc.

:+1: Less total packages to install;
:+1: Shorter bootstrap time for the whole repo;
:+1: More clear hierarchy to the repo;
:+1: Less work and boilerplate(package.json, docs) adding an example to an instrumentation;
:+1: Examples could reference unpublished `../`(built source), which means that they work right after the merge and do not need a separate publish and they wouldn't be tied to a specific version - Running examples would be as easy as running tests(that's also a :-1: if you are not used to doing that);

:-1: No *one* place for a user who just wants to explore the examples: Could add a README pointing to some of the examples under packages or the fact that most of the packages come with examples.
:-1: If `../` is referenced the example is not exactly copy-paste, because `../` has to be replaced by the package name.
:-1: Running most examples requires compilation.
:-1: Not all examples could reasonably be moved - some that are using multiple packages should still be left under the root examples.

dyladan:
Sounds like a good idea to me. I think our examples in general need some work and I would really like it if they could be compiled in order to be sure they match the current API

obecny:
I would keep examples for node the same way I did for web, one package.json and adding new example is easy, having example in each package might be not desired if we think about size etc., but it could be an improvement if you dont have to install this separately. Besides quite often for certain example you need more than just an instrumentation so it would mean keeping extra packages in every package, imho bad and hard to maintain. The easiest would be to have one node folder with one package.json and then each example will simply leave in separate folder - check this -> https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/web

Rauno56:
Let me see if I got this right, @obecny: your suggestion would be to take all `examples/*/package.json` files and merge them into one `examples/package.json`. That would:

:+1: dedupe many of the common 3rd party libs across the examples - less total deps.
:+1: keep the one place for all NodeJS examples - especially makes sense for cross-package examples.
:-1: have one massive `package.json` with basically all OTel libs installed(eventually) + all instrumentations + all instrumented libs + bits and bobs, utilities.
[...]

PR #939, opened 2022-03-09
chore: moves express examples into its package to establish pattern

andyfleming:
## Problem this solves

Examples get out of date quickly since they are at the top-level of this repository. There's no validation that they work correctly with the latest version of instrumentation packages.

## Changes in this PR

* Establishes new pattern of publishing examples in their respective instrumentation package directories (as discussed at the OTEL js SIG meeting on March 2nd).
* Moves example for express to the express instrumentation directory
* Converts express example to TypeScript to support a build step in CI as a technique for testing/validation of the example.
* Adds documentation in CONTRIBUTING.md about how to organize examples.
* Adds a README.md to the examples folder to note the migration (and can be used to document other standard examples in that folder).

## Testing approach

For validation of the instrumentation example code, I propose we leverage TypeScript compilation. It's not as exhaustive as some sort of functional tests, but I think that would be too costly at this point.

I've implemented that via the `npm run --if-present compile:examples` command in CI. 

We could later establish a pattern for CI to check for a "test:examples" script in the package.json of each instrumentation package to run (if present) in addition to the basic TypeScript compilation.

## Additional Notes

It was discussed that we could keep the top-level examples directory for 2 reasons:

1. To leave it as a home for examples that have not yet been migrated to their respective package directories.
2. To provide a basic instrumentation example and a README that guides users to the examples in the individual instrumentation packages.

## Checklist


plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -32,16 +33,18 @@
     "@opentelemetry/exporter-jaeger": "^0.25.0",
     "@opentelemetry/exporter-zipkin": "^0.25.0",
     "@opentelemetry/instrumentation": "^0.25.0",
-    "@opentelemetry/instrumentation-express": "^0.23.0",
     "@opentelemetry/instrumentation-http": "^0.25.0",
     "@opentelemetry/resources": "^0.25.0",
-    "@opentelemetry/semantic-conventions": "^0.25.0",
-    "@opentelemetry/sdk-trace-node": "^0.25.0",
     "@opentelemetry/sdk-trace-base": "^0.25.0",
+    "@opentelemetry/sdk-trace-node": "^0.25.0",
```
Flarna, line 39: Does the move really result in up to date samples if they don't use up to date dependencies? I think they should use the current instrumenation and the current core dependencies.

andyfleming: Good catch! I agree. Do you think some of them should be referenced with relative paths, version `*`, or something else?

Flarna: well, that's one of the point where it's easy to complain but hard to provide a good solution.
I think we should aim to depend on the same versions then the instrumentation does for that ones which are in common.
The sample should most likely depend on the instrumentation in the repo, maybe link via lerna or similar.
And renovate bot should include the samples.

But I don't know if this is easy to archive.

dyladan: If you add the express instrumentation to the package.json as a relative path it might be good enough. I would then use `*` dependencies for otel dependencies so we always get latest (or `^1.0.3` if you prefer).

Rauno56 (commented):
I support moving examples under respective packages and happy to see someone is getting the ball rolling.

How do you think converting the examples to Typescript will affect the benefit non-TS users will get from them?

A previous discussion on this: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/552

dyladan:
> I support moving examples under respective packages and happy to see someone is getting the ball rolling.
> 
> How do you think converting the examples to Typescript will affect the benefit non-TS users will get from them?
> 
> A previous discussion on this: #552

I think the benefit of having the examples type-checked and therefore easier to notice when they are outdated will outweigh the disadvantage of presenting typescript to a non-typescript user.

dyladan:
> A bit of topic regarding this specific PR. But what about adding an `InMemorySpanExporter`, run the samples in CI and verify that they create the required N spans?

I wonder how hard it would be to do that in a way where the example isn't unnecessarily complex for new users who are just trying to figure out how to use the instrumentation?

vmarchaud:
> A bit of topic regarding this specific PR. But what about adding an InMemorySpanExporter, run the samples in CI and verify that they create the required N spans?

Shouldn't the tests already verify that anyway ?

plugins/node/opentelemetry-instrumentation-express/examples/README.md
```diff
@@ -44,7 +44,7 @@ Setup [Jaeger Tracing](https://www.jaegertracing.io/docs/latest/getting-started/
 `zipkin:server` script should output the `traceid` in the terminal (e.g `traceid: 4815c3d576d930189725f1f1d1bdfcc6`).
 Go to Zipkin with your browser <http://localhost:9411/zipkin/traces/(your-trace-id)> (e.g <http://localhost:9411/zipkin/traces/4815c3d576d930189725f1f1d1bdfcc6>)
 
-<p align="center"><img src="./images/zipkin.jpg?raw=true"/></p>
+<p align="center"><img src="imagesipkin.jpg?raw=true"/></p>
```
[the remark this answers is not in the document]
andyfleming: Yeah, I think this was unintentional (and/or an IDE side-effect).

Rauno56 (approved):

Flarna:
As far as I know the samples are not executed at all in CI. The tests of e.g. express instrumentation verify that the instrumentation works but not that the example works.

andyfleming:
I added instructions for migrating examples in the examples/README file. I also created an example of what an issue would look like for a given package that needs converted. See [the example issue for koa instrumentation here](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/950).

andyfleming:
I believe this PR is ready to go now 👍

Rauno56:
> I believe this PR is ready to go now +1

Sorry, I merged another PR that caused conflicts here.

dyladan:
lint also fails on the markdown lint check

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "url": "https://github.com/open-telemetry/opentelemetry-js/issues"
+  },
+  "dependencies": {
+    "@opentelemetry/api": "^1.0.2",
+    "@opentelemetry/exporter-jaeger": "^0.26.0",
+    "@opentelemetry/exporter-zipkin": "^0.26.0",
+    "@opentelemetry/instrumentation": "^0.27.0",
+    "@opentelemetry/instrumentation-http": "^0.27.0",
+    "@opentelemetry/resources": "^0.26.0",
+    "@opentelemetry/sdk-trace-base": "^0.26.0",
```
Flarna, line 38:
```suggestion
    "@opentelemetry/sdk-trace-base": "^1.0.0",
```
same below for sdk-trace-node. Otherwise some packages like e.g @opentelemetry/core are installed twice.

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "express-example",
+  "private": true,
+  "version": "0.23.0",
```
Flarna, line 4:
```suggestion
  "version": "0.28.0",
```
I guess this should go in sync with express instrumentation itself

dyladan (approved):

plugins/node/opentelemetry-instrumentation-express/examples/package.json
Flarna: Could you please update all stable core Otel components to ^1.0.0? Mixing versions likely wont work and it is for sure not what we want that people do.

andyfleming:
Ok, lint should be resolved and latest changes from main merged.

vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -35,7 +35,7 @@
     "@opentelemetry/instrumentation": "^0.27.0",
     "@opentelemetry/instrumentation-http": "^0.27.0",
     "@opentelemetry/resources": "^0.26.0",
-    "@opentelemetry/sdk-trace-base": "^0.26.0",
+    "@opentelemetry/sdk-trace-base": "^1.0.0",
     "@opentelemetry/sdk-trace-node": "^0.26.0",
```
Flarna, line 39: I still think we should avoid 0.x and 1.x versions of core components.
But as far as I understood other comments the target is to update dependencies in followup PRs.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-express/examples/package.json
andyfleming: Good catch. I went ahead and bumped all of the otel dependency versions to their latest.

Rauno56:
Why eslint-ignore the examples?

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@opentelemetry/exporter-zipkin": "^1.1.1",
+    "@opentelemetry/instrumentation": "^0.27.0",
+    "@opentelemetry/instrumentation-http": "^0.27.0",
+    "@opentelemetry/resources": "^1.1.1",
+    "@opentelemetry/sdk-trace-base": "^1.1.1",
+    "@opentelemetry/sdk-trace-node": "^1.1.1",
+    "@opentelemetry/semantic-conventions": "^1.1.1",
+    "axios": "^0.21.1",
+    "cross-env": "^7.0.3",
+    "express": "^4.17.1"
```
blumamir, line 43: for typescript, you should also include `@types/express`

plugins/node/opentelemetry-instrumentation-express/examples/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "rootDir": ".",
+  },
+  "include": [
+    "src/**/*.ts",
+    "test/**/*.ts"
```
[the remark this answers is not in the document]
andyfleming: Will remove 👍

examples/README.md
```diff
@@ -0,0 +1,41 @@
...
+```json
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "rootDir": ".",
+  },
+  "include": [
+    "src/**/*.ts",
+    "test/**/*.ts"
```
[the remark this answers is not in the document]
andyfleming: Will remove 👍

plugins/node/opentelemetry-instrumentation-express/.eslintignore
```diff
@@ -1 +1,2 @@
 build
+examples
```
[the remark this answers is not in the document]
blumamir: cool

plugins/node/opentelemetry-instrumentation-express/examples/src/tracer.ts
```diff
@@ -1,26 +1,28 @@
...
+import { ZipkinExporter } from '@opentelemetry/exporter-zipkin';
+import { Resource } from '@opentelemetry/resources';
+import { SemanticAttributes, SemanticResourceAttributes as ResourceAttributesSC } from '@opentelemetry/semantic-conventions';
+import { SpanAttributes } from "@opentelemetry/api/build/src/trace/attributes";
 
-const Exporter = (process.env.EXPORTER || '')
-  .toLowerCase().startsWith('z') ? ZipkinExporter : JaegerExporter;
-const { ExpressInstrumentation } = require('@opentelemetry/instrumentation-express');
+const Exporter = (process.env.EXPORTER || '').toLowerCase().startsWith('z') ? ZipkinExporter : JaegerExporter;
+const { ExpressInstrumentation } = require('../../src');
```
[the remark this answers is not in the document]
andyfleming: As discussed at the SIG, that sounds like the right path here. I've updated the PR accordingly. Thanks!

.github/workflows/unit-test.yml
```diff
@@ -199,6 +199,8 @@ jobs:
         run: npx lerna bootstrap --no-ci
       - name: Unit tests
         run: npm run test:browser
+      - name: Build examples
+        run: npm run --if-present compile:examples
```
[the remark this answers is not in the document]
andyfleming: Updated 👍

blumamir (approved):
LGTM
Thanks @andyfleming for doing this

The CI currently fails due to unrelated issues.  After they are fixed, I suggest taking a quick look to verify that the "compile:examples" is run as expected in CI before merging.

vmarchaud:
friendly ping @andyfleming  if you find the time to wrap this one :)

andyfleming:
I'm having trouble reproducing the CI error locally. Here's the CI error. It doesn't look like an error directly in the example files. Any hunch on what's going wrong?

```
> opentelemetry-contrib@0.1.0 compile:examples
> lerna run compile:examples

lerna notice cli v3.22.1
lerna info ci enabled
lerna info Executing command in 1 package: "npm run compile:examples"
lerna ERR! npm run compile:examples exited 2 in '@opentelemetry/instrumentation-express'
lerna ERR! npm run compile:examples stdout:

> @opentelemetry/instrumentation-express@0.28.0 compile:examples
> cd examples && npm run compile


> express-example@0.28.0 compile
> tsc -p .

node_modules/@opentelemetry/sdk-trace-base/build/src/BasicTracerProvider.d.ts:1:29 - error TS2614: Module '"@opentelemetry/api"' has no exported member 'TracerOptions'. Did you mean to use 'import TracerOptions from "@opentelemetry/api"' instead?

1 import { TextMapPropagator, TracerOptions, TracerProvider } from '@opentelemetry/api';
                              ~~~~~~~~~~~~~


Found 1 error in node_modules/@opentelemetry/sdk-trace-base/build/src/BasicTracerProvider.d.ts:1


lerna ERR! npm run compile:examples exited 2 in '@opentelemetry/instrumentation-express'
Error: Process completed with exit code 2.
```

Flarna:
most likely caused by a mixmax of installed `@opentelemetry/api` versions. `TraceOptions` was added in API 1.1.0 so I guess in CI 1.0.x is installed next to SDK 1.1.0.
The reason for a diff between local run and ci is maybe the different lerna bootstrap commend used (see [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e99d6829604b90541e0bd33c426583565ac60167/.github/workflows/unit-test.yml#L153))

Flarna:
> Will [open-telemetry/opentelemetry-js#2892](https://github.com/open-telemetry/opentelemetry-js/pull/2892) help with that issue?

Yes I think so.

dyladan:
Once we have a release I think the CI will be much easier.

dyladan:
@andyfleming thank you for your patience. Release PR is prepared and should merge/release today I hope

andyfleming:
Thanks for following up on this @Rauno56 ! 🙂

Rauno56:
Sorry for this taking so long! Thanks and congrats for your first contribution!

PR #946, opened 2022-03-16
fix(amqplib): add missing rabbit service to CI daily run

blumamir:

- after amqplib is merged, the daily run of `test-all-versions` fails because I added the service only to `unit-test.yml` and forgot to add it to `test-all-versions.yml`.

- add the required rabbitmq service also to `test-all-versions.yml`. This should fix the CI.

## Checklist


Flarna (approved):

Rauno56 (approved):
:+1:

PR #948, opened 2022-03-16
fix: drop running amqplib tests on node 8 and add rabbit service for TAV

Rauno56:

CI is failing because of TAV runs of amqplib.

- skip on node 8
- add rabbitmq service


Rauno56:
duplicate of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/946

PR #953, opened 2022-03-23
chore: update OTel core deps and API

Rauno56:

Problem explained in https://github.com/open-telemetry/opentelemetry-js/pull/2863

Update everything to latest to make sure it's all in sync.

This is an alternative solution to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/954


Flarna:
I think this requires also a release of experimental core modules as e.g. `@opentelemetry/instrumentation` 0.27.0 has 1.0.x of stable pinned.

Rauno56:
Yes... and `@opentelemetry/exporter-trace-otlp-http`(for `@opentelemetry/browser-extension-autoinjection)` which we cannot release right now because of the json exporter, or is it separate? @dyladan

dyladan:
https://github.com/open-telemetry/opentelemetry-js/pull/2835 moves them all back to experimental, then we will make them not depend on each other, then we will move only proto and grpc to stable. JSON is being stabilized so the exporter can be released stable when that happens.

Rauno56:
> `@opentelemetry/instrumentation`

Rauno56:
Sorry, misclicked "close".

Actually, @Flarna, I take that back. We don't need to release `@opentelemetry/instrumentation` because it's `peerDependencies` are as such: `{ '@opentelemetry/api': '^1.0.0' }`. Which means they allow 1.x, not 1.0.x.

Rauno56:
It seems like I've run of "PR quota" or something, the CI just doesn't run.
:heart: GHA

Rauno56:
Between this and https://github.com/open-telemetry/opentelemetry-js-contrib/pull/954, this is

* probably safer because [#954](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/954) doesn't take into account packages that have invalid peer dep ranges like 1.0.x versions of trace SDK.
* more invasive for users because forces version 1.1.0 of the API.

Rauno56:
I'm going to close this. All the conflicting deps are updated in #954.

PR #954, opened 2022-03-24
chore: update OTel deps

Rauno56:

Problem explained in https://github.com/open-telemetry/opentelemetry-js/pull/2863
It's another experiment to see if we can get away without updating all the dependencies for now.

1. Leave ranges the same because they usually don't need to be stricter.
2. Update pinned API version in dev deps not to conflict with the updated SDK packages.

This is an alternative solution to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/953


Rauno56:
That's a good start. No incompatible packages anymore. I imagine api version range in peerDeps being invalid still for contrib packages. Will check those next.

naseemkullah (approved):

packages/opentelemetry-browser-extension-autoinjection/package.json
```diff
@@ -26,7 +26,7 @@
     "@opentelemetry/api": "^1.0.2"
   },
   "devDependencies": {
-    "@opentelemetry/api": "1.0.2",
+    "@opentelemetry/api": "1.1.0",
```
Flarna, line 29: below quite some deps are pinned to fixed versions (github doesn't allow to comment there). I think we should ranges there.

packages/opentelemetry-test-utils/package.json
```diff
@@ -38,7 +38,7 @@
     "@opentelemetry/api": "^1.0.2"
   },
   "devDependencies": {
-    "@opentelemetry/api": "1.0.2",
+    "@opentelemetry/api": "1.1.0",
```
Flarna, line 41: There are pinned versions in dependencies below.

Flarna:
There are still some inconsistencies in the ranges used, e.g. sometimes we have
```json
"peerDependencies": {
  "@opentelemetry/api": "^1.0.4"
},
"dependencies": {
  "@opentelemetry/resources": "^1.0.1",
  "@opentelemetry/semantic-conventions": "^1.0.1"
}
```

but sometimes like this:

```json
"peerDependencies": {
  "@opentelemetry/api": "^1.0.2"
},
"dependencies": {
  "@opentelemetry/resources": "^1.0.0",
  "@opentelemetry/semantic-conventions": "^1.0.0"
}
```

I guess we should do some cleanup in this regard also but no need to do this in this PR.

packages/opentelemetry-browser-extension-autoinjection/package.json
Rauno56: Changed those as well.

packages/opentelemetry-test-utils/package.json
Rauno56: And these.

packages/opentelemetry-test-utils/package.json
```diff
@@ -38,18 +38,18 @@
...
-    "@opentelemetry/resources": "1.0.1",
-    "@opentelemetry/sdk-trace-base": "1.0.1",
-    "@opentelemetry/semantic-conventions": "1.0.1"
+    "@opentelemetry/core": "^1.1.1",
+    "@opentelemetry/exporter-jaeger": "^1.1.1",
+    "@opentelemetry/instrumentation": "^0.27.0",
+    "@opentelemetry/sdk-trace-node": "^1.1.1",
+    "@opentelemetry/resources": "^1.1.1",
+    "@opentelemetry/sdk-trace-base": "^1.1.1",
+    "@opentelemetry/semantic-conventions": "^1.0.1"
```
Flarna, line 53: Why is here `^1.0.1` but the others are at `^1.1.1`?
At other places you moved to 1.1.x only for dep-dependencies but not for prod dependencies. So maybe it should be `^1.0.0`?

Rauno56: I only updated core packages that rely on api, semantic conventions does not.
Since it was `1.0.1` before, I didn't want to to downgrade the possible versions and just opened it up for upgrades.

Flarna:
seems long task is unhappy now because it uses 1.1.1 for sdk-trace-base but 1.0.1 for sdk-trace-web as latest tag was moved.

PR #963, opened 2022-04-08
feat: update webpack outside of examples

Rauno56:

Webpack hasn't been updated for a while. Testing out `pnpm`, I noticed webpack@4 doesn't work with it at all, so out of curiousity(the research of `pnpm` is still ongoing) had to hack it to work.

I'm not familiar with configuring webpack *at all* so any feedback to the changes are very welcome. The build results are identical.

- Updated all webpack- and karma-related packages.
- Changed `karma.conf.js` to include all plugins form the relative `package.json` - Karma failed autoloading all deps from the package.json if the deps were symlinked.
- Installed node polyfills for webpack, it used to include those, but doesn't in webpack@5.

## Checklist

- [ ] ~Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.~ Browser-based packages do not use TAV


Flarna:
I'm surpriesed that you got this running. A while ago I tried something similar in core repo but failed (created https://github.com/open-telemetry/opentelemetry-js/issues/2187 that time).

One of the reasons why I failed were that there are dependencies which seem to require webpack for, e.g. istanbul-instrumenter-loader has `"webpack": "^2.0.0 || ^3.0.0 || ^4.0.0"` in its [`peerDependencies`](https://github.com/webpack-contrib/istanbul-instrumenter-loader/blob/fe5fedb818cf2a96b3a5111ed4f2b04fd5ffd852/package.json#L76-L78).

Rauno56:
Those errors are ignored if legacy peer deps is turned on, so I guess that's why it hasn't popped onto my radar.

- [ ] Must check whether code coverage reporting still works. Thanks for the reference.

Rauno56:
Even though `instanbul-instrumenter-loader` worked with webpack@5, the peer dep range did not include v5. Replaced it with another package that's otherwise identical but doesn't have that problem.

vmarchaud (approved):

Rauno56:
Any additional comments/questions, @Flarna?

dyladan:
> No questions. Nice that this works. As I have not really an idea of these browser packages/setups I'm not sure if I'm the right one to approve here.

I think nobody is left who remembers when they were set up

PR #968, opened 2022-04-10
fix(instrumentation-aws-sdk): sqs message id missing on send command

### LINKED ISSUE #964: aws-sdk: sqs instrumentation is not setting messaging.message_id on send request
```
    "@opentelemetry/api": "^1.0.4",
    "@opentelemetry/api-metrics": "^0.27.0",
    "@opentelemetry/instrumentation-aws-sdk": "^0.5.1",
    "@opentelemetry/instrumentation-http": "^0.27.0"
```

`v16`

### What did you do?

We have instrumented our code with `aws-sdk` instrumentation.

When we are doing [sendMessage command](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sqs/classes/sendmessagecommand.html) we are expecting to se `messaging.message_id` as defined in [the messaging specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#messaging-attributes).

`messaging.message_id` attribute being set on spans for `sendMessage` command.

### What did you see instead?

We don't see this attribute.

### Additional context

aws-sdk-issue

mentos1386:

- Fixes #964 

- Added test that makes sure that message id attribute is added to span on send command
- Fixed response hook implementation to add message id attribute to span on send command

I am not sure about implementation for `SendMessageBatch` as there cannot be a single `message id` be set, as there are multiple of them.

## Checklist

   - I ran it, but it failed due to unrelated issue.


plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
```diff
@@ -121,45 +121,61 @@ export class SqsServiceExtension implements ServiceExtension {
...
-      pubsubPropagation.patchArrayForProcessSpans(
-        messages,
-        tracer,
-        context.active()
-      );
+    switch (response.request.commandName) {
+      case 'SendMessage':
+        span.setAttribute(
+          SemanticAttributes.MESSAGING_MESSAGE_ID,
+          response.data.MessageId
```
blumamir, line 128: Is it guaranteed that data is not nullish? I recommend using optional chaining, so we don't throw if `data` is not valid.

blumamir (commented):
Thanks

Did you check to verify it's working for v2 as well? I see that test-all-version failed in your setup but was it tested with at least one v2 version?

I see the unit-test is added for v3 only. Does it makes sense also adding it to v2 [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts)?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
mentos1386: Fixed. It's strange that `any` is used. It would be better to refactor the code to use `unknown` to avoid this kind of issues in the future.

mentos1386:
@blumamir I'm unable to run `test-all-versions`, i get weird issue:
```

File: plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts [hunk 2]
```diff
@@ -361,6 +363,49 @@ describe('SQS', () => {
...
+      expect(span.attributes[SemanticAttributes.MESSAGING_DESTINATION]).toEqual(
+        QueueName
+      );
+      expect(span.attributes[SemanticAttributes.MESSAGING_URL]).toEqual(
+        params.QueueUrl
+      );
+      expect(span.attributes[SemanticAttributes.MESSAGING_MESSAGE_ID]).toEqual(
+        response.MessageId
+      );
+      // expect(span.attributes[SemanticAttributes.HTTP_STATUS_CODE]).toEqual(200);
```
mentos1386, line 406: It seems to be `undefined` with aws sdk v2 and `200` with aws sdk v3. Haven't researched why.

blumamir: Might be related to the mock response returned for these tests [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/c5b9356d4eb925df64b66c859d1d8367ae57d437/plugins/node/opentelemetry-instrumentation-aws-sdk/test/testing-utils.ts#L52).

It's ok to merge as is, but if you have the motivation you are very welcome to see if it's an easy fix and help improve the tests :)

plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts
```diff
@@ -361,6 +363,49 @@ describe('SQS', () => {
   });
 
   describe('hooks', () => {
+    it('sqsResponseHook add messaging attributes', async () => {
```
blumamir, line 366: nit: this test is specifically for attributes added during `sendMessage`. Might be worth mentioning in the test name

blumamir (approved):
LGTM.
Thank you for the research and for working on a high-quality fix for that.

Added 2 nits but they are both optional - only if you have time and motivation

plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts
mentos1386: Fixed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/968/commits/0f6155c856f52fcdde568b1ea682c84f167c74f0

plugins/node/opentelemetry-instrumentation-aws-sdk/test/sqs.test.ts
mentos1386: You were right. Lack of types makes it hard to contribute to the repository :( Fixed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/968/commits/866e00f27ae50d78f5d9caf7e442c6b1dbe22e47.

blumamir: > Lack of types makes it hard to contribute to the repository

I agree. If you have some free time and want to contribute by adding them, that would be great :)

PR #969, opened 2022-04-11
docs: remove the lists of packages from the readme

Rauno56:

Every now and then we get a PR for updating the list in the Readme. I don't personally feel having a list in the readme is worth it if we can reference the folder the packages are all nested in but I'm also interested in what everyone else thinks.

1. Remove the lists of the included packages from the readme.
2. Reword some other parts about the support according to https://github.com/open-telemetry/opentelemetry-js/issues/2721#issuecomment-1076570990


vmarchaud:
Apart from the lint failing, i agree that its a problem and the information is generally available elsewhere (specially in the opentelemtry website) so i'm in favor of removing them as-is and maybe replace them by a link to https://opentelemetry.io/registry/ ?

dyladan:
I agree a link to the registry should be sufficient. I wonder if it would be possible to have a CI check that warns if the package is not in the registry?

Rauno56:
Added a link to the registry.

I think linking straight to Registry in the main content of the readme creates weird UX(repo -> registry -> repo) with little benefit(registry search vs Ctrl-F).

I have a feeling that most people will arrive to the repo through npm anyway

dyladan:
> Added a link to the registry.
> 
> I think linking straight to Registry in the main content of the readme creates weird UX(repo -> registry -> repo) with little benefit(registry search vs Ctrl-F).
> 
> I have a feeling that most people will arrive to the repo through npm anyway

I think its fine how you did it. The registry links directly to packages though. Unless someone is already familiar with our package structure, they may find the registry easier to navigate. In any case this LGTM

blumamir (approved):

PR #973, opened 2022-04-13
feat: add supported node versions for all packages

Rauno56:

In accordance with https://github.com/open-telemetry/opentelemetry-js/issues/2834 we've changed the support range to `>=8.12`.
https://github.com/open-telemetry/opentelemetry-js/commit/da22673d102d3ecb6364a39be968cd8d666cc136 changed that in Core.

Updated every `package.json` according to the issue above and out CI configuration, which skips tests for some of the packaged that have dropped the support for older node versions.


Flarna:
strictly speaking only modules using the performance times require 8.12 others work fine with lower versions. But not sure if there is any benefit of having e.g. a resource detector working with 8.0 but there is no SDK to use it with.

blumamir (approved):
LGTM.
This will break users if they are running on node >=8.5.0 && <=8.12.0
Not sure if there are any of these out there 🤷‍♂️

Could be nice to at least bump the minor version for publishing this change (e.g. label it as "feat"?)

Rauno56:
> strictly speaking only modules using the performance times require 8.12 others work fine with lower versions. But not sure if there is any benefit of having e.g. a resource detector working with 8.0 but there is no SDK to use it with.

That's right. In that sense we don't absolutely need to narrow down propagators and detectors. I think I'd approached it differently if we had SDK compatible with 8.5 and took it from module<>plugin direction making the version minimum of what either can support. Would you prefer it solved differently?

Flarna (approved):

Flarna:
I think it's fine. I doubt that it has much impact at all as people using that old node versions are most likely used to ignore warnings anyway :o).

PR #976, opened 2022-04-19
feat: use Koa router name as span name if available

lukiano:

- This allows to use the Koa router name (if available) as the span name. Usually this name is more meaningful than "method + path". In particular if the route name is the same as the OpenAPI's `operationId` for that route then it will be easier to associate spans with the OpenAPI spec.  

- Koa will store the route name in `_matchedRouteName`. Note that I'm not using the format `router - xxx` as I'd rather have the span name clean.

## Checklist

NPM failed to compile on the untouched `main` branch (some type errors related to React)

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
@@ -175,6 +175,9 @@ export class KoaInstrumentation extends InstrumentationBase<typeof koa> {
           span.recordException(err);
           throw err;
         } finally {
+          if (context._matchedRouteName) {
+            span.updateName(context._matchedRouteName);
+          }
           span.end();
         }
       });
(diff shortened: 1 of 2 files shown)


Rauno56:
@lukiano, please sign the CLA.

lukiano:
I've signed the CLA

Rauno56:
A named route refers to a route + a method so there's no way to use a named route with different methods?

1. The test should not be edited. Add another one for your usecase.
2. Nit: Your test says "insertOperation" even though it's a get operation for a post(not to mistaken for a POST method).
3. Current implementation already does gymnastics with naming the span, move your logic to where the name is calculated already. Do a else-if, for example.

[API docs for Koa](https://github.com/koajs/router/blob/master/API.md)

lukiano:
I'll work on these changes today. Sorry for the delay.

lukiano:
I've addressed the feedback and ran tav

Rauno56 (approved):

Rauno56:
Thanks and congrats about your first contribution, @lukiano!

PR #980, opened 2022-04-25
[WIP] feat(user-interaction): create one span per interaction

### LINKED ISSUE #548: User events getting registered twice

bradfrosty:

Fixes #548

This a WIP solution that replaces #548, and should account for some of the debate around the approach in that PR regarding heuristics. However, this solution is a much larger diff (albeit less code/simpler than existing solution by avoiding Zone patches).

[See StackBlitz for example](https://stackblitz.com/edit/opentelemetry-js-interaction-with-gtm-akeucp)

- Deduplicates user interaction spans per event target, rather than create one span per listener callback. Contrary to #743, this implementation does not rely on heuristics and accurately captures the execution context across multiple event handlers.
- Further aligns the instrumentation with the [upcoming Responsiveness CWV metric](https://web.dev/better-responsiveness-metric/). This will be baked in the PerformanceObserver in the future where we can get increased accuracy on span timestamps. However, the metrics they record will not account for async tasks due to the difficult implementation, [even though they would prefer that](https://web.dev/better-responsiveness-metric/#capture-the-full-event-duration). Because this solution is leveraging Zone.js, we are capable of capturing this powerful data today.

- Refactors Zone patching logic to a ZoneSpec implementation that performs async task tracking across all listeners for an event target, and provides onComplete/onError callbacks consumable from the instrumentation.
- Changes span duration characteristics. It now starts from the first `event.timeStamp`, and ends after all async tasks have been performed within all listener callbacks. Starting from `event.timeStamp` improves the recording accuracy as it now accounts for the input delay (aka First Input Delay but for all interactions)
- Calculates additional aggregates and events that may be useful when analyzing interactions, such as `processingStart` (when the browser is able to start executing), `processingEnd` (when all synchronous work is complete), `listenerCount` (total count of listeners that executed for the interaction), and detailed task counts (number of async tasks executed). The `processingStart` and `processingEnd` metrics are the same values found on the [spec'd PerformanceEventTiming interface](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEventTiming#properties).
- Silently records errors/exceptions on the interaction span without disrupting user-land error handling

## Checklist

- [ ] Add several new test cases (but having difficulty getting Karma to work right now)
- [ ] Explore additional aggregates such as longest event handler within interaction, total interaction latency over time, average interaction latency over time, high quantile aggregates
- [ ] Explore grouping `up` and `down` metrics (e.g. `mousedown` + `mouseup`, `keydown` + `keyup`) to further align with Responsiveness metric, since both events are guaranteed to occur and count as a single "interaction"
- [ ] Check if ZoneContextManager is being used rather assume it is being used by checking window.Zone 
- [ ] Verify accuracy by comparing with PerformanceTimeline values

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/src/enums/AttributeNames.ts
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/src/enums/AttributeNames.ts
@@ -15,11 +15,18 @@
  */
 
 export enum AttributeNames {
-  COMPONENT = 'component',
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions
+  INSTRUMENTATION_NAME = 'otel.scope.version',
+  INSTRUMENTATION_VERSION = 'otel.scope.version',
+  HTTP_URL = 'http.url',
+  HTTP_USER_AGENT = 'http.user_agent',
+
+  // NOT ON OFFICIAL SPEC
   EVENT_TYPE = 'event_type',
   TARGET_ELEMENT = 'target_element',
   TARGET_XPATH = 'target_xpath',
-  HTTP_URL = 'http.url',
-  // NOT ON OFFICIAL SPEC
-  HTTP_USER_AGENT = 'http.user_agent',
+  LISTENERS_COUNT = 'listeners.count',
+  TASKS_SCHEDULED_COUNT = 'tasks.scheduled.count',
+  TASKS_RAN_COUNT = 'tasks.ran.count',
+  TASKS_CANCELLED_COUNT = 'tasks.cancelled.count',
 }
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/src/enums/AttributeNames.ts
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/src/enums/AttributeNames.ts
@@ -16,7 +16,7 @@
 
 export enum AttributeNames {
   // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions
-  INSTRUMENTATION_NAME = 'otel.scope.version',
+  INSTRUMENTATION_NAME = 'otel.scope.name',
   INSTRUMENTATION_VERSION = 'otel.scope.version',
   HTTP_URL = 'http.url',
   HTTP_USER_AGENT = 'http.user_agent',
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
@@ -17,194 +17,218 @@
 import { isWrapped, InstrumentationBase } from '@opentelemetry/instrumentation';
 
 import * as api from '@opentelemetry/api';
-import { hrTime } from '@opentelemetry/core';
+import { isTracingSuppressed } from '@opentelemetry/core';
 import { getElementXPath } from '@opentelemetry/sdk-trace-web';
+import { UserInteractionZoneSpec } from './UserInteractionZoneSpec';
 import { AttributeNames } from './enums/AttributeNames';
-import {
-  AsyncTask,
-  EventName,
-  RunTaskFunction,
-  ShouldPreventSpanCreation,
-  SpanData,
-  UserInteractionInstrumentationConfig,
-  WindowWithZone,
-  ZoneTypeWithPrototype,
-} from './types';
+import { EventName, UserInteractionInstrumentationConfig } from './types';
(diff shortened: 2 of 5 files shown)


crellison (approved):

plugins/web/opentelemetry-instrumentation-user-interaction/src/enums/AttributeNames.ts
```diff
@@ -15,11 +15,18 @@
  */
 
 export enum AttributeNames {
-  COMPONENT = 'component',
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions
+  INSTRUMENTATION_NAME = 'otel.scope.version',
```
bradfrosty, line 19:
```suggestion
  INSTRUMENTATION_NAME = 'otel.scope.name',
```

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -17,194 +17,218 @@
...
-    Event,
-    api.Span
-  >();
-  private _eventNames: Set<EventName>;
-  private _shouldPreventSpanCreation: ShouldPreventSpanCreation;
-
-  constructor(config?: UserInteractionInstrumentationConfig) {
-    super('@opentelemetry/instrumentation-user-interaction', VERSION, config);
-    this._eventNames = new Set(config?.eventNames ?? DEFAULT_EVENT_NAMES);
-    this._shouldPreventSpanCreation =
```
stephencranedesign, line 70: ❓ 

why did we get rid of the `shouldPreventSpanCreation`?

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -17,194 +17,218 @@
...
-      this._spansData.set(span, {
-        taskCount: 0,
-      });
+            if (plugin._zonePatched) {
+              // create a new zone for listener for given target
+              activeInteractionContext = Zone.current.fork(
+                new UserInteractionZoneSpec(
+                  span,
+                  function onComplete(attributes, events) {
+                    console.log('onComplete', event);
```
stephencranedesign, line 126: When using this w/ react, I'm noticing that this onComplete function is executed right away and isn't being delayed until the async activity is completed. 

I updated your initial stack blitz with an [example here](https://stackblitz.com/edit/opentelemetry-js-interaction-with-gtm-iz9bho?file=src/app.tsx) if you'd like to take a look.

valerybugakov:
Hey @bradfrosty, thanks for sharing this work! I just recently started looking into this package and bumped into similar issues. Have you wholly switched to maintaining your forked internal version of user-interactions instrumentation? 

Let me know if I can help with anything here. I'd like to see this merged. 
cc @stephencranedesign if you have thoughts on how to push this forward.

valerybugakov:
@dyladan, @Rauno56, maybe?

Josh-a-e:
Has this fix/change been abandoned?

dyladan:
@Josh-a-e this component was maintained by a contributor who is no longer with the project, but if you're interested in this I would suggest you join the RUM SIG. We're holding off on many browser things while we wait for them to make some fundamental decisions about how they should work, including things like this.

PR #947, opened 2022-03-16
chore: release main

dyladan:
:robot: I have created a release *beep* *boop*
---

<details><summary>@opentelemetry/auto-instrumentations-node: 0.28.1</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/instrumentation-aws-lambda bumped from ^0.30.0 to ^0.31.0
    * @opentelemetry/instrumentation-winston bumped from ^0.27.3 to ^0.28.0
</details>

<details><summary>auto-instrumentations-web: 0.28.0</summary>

### Features

* remove colors dependency ([#943](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/943)) ([b21b96c](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/b21b96c1a3a4f871370f970d6b2825f00e1fe595)), closes [#826](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/826)

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/instrumentation-document-load bumped from ^0.27.1 to ^0.28.0
    * @opentelemetry/instrumentation-user-interaction bumped from ^0.28.1 to ^0.29.0
</details>

<details><summary>browser-extension-autoinjection: 0.28.0</summary>

### Features

* remove colors dependency ([#943](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/943)) ([b21b96c](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/b21b96c1a3a4f871370f970d6b2825f00e1fe595)), closes [#826](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/826)

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/instrumentation-document-load bumped from ^0.27.1 to ^0.28.0
</details>

<details><summary>id-generator-aws-xray: 1.1.0</summary>

### Features

* remove colors dependency ([#943](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/943)) ([b21b96c](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/b21b96c1a3a4f871370f970d6b2825f00e1fe595)), closes [#826](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/826)
</details>

<details><summary>instrumentation-aws-lambda: 0.31.0</summary>

### Features

[...]

PR #981, opened 2022-04-26
feat: add new instrumentations into `auto-instrumentations-node`

Rauno56:

As @pizzaz93 pointed out, some of the new instrumentation packages were missing from the auto-instrumentation-node package.

Added all contrib packages to the metapackage.

## Checklist


Flarna:
Isn't it needed to update also [utils.ts](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/src/utils.ts)?

Rauno56:
It is! I don't even have any excuses. Added a test for it for the future.

blumamir (approved):
Liked the new test.
LGTM

Flarna (approved):

Flarna:
Not related to this PR. It seems the dependencies are not updated automatically here if an instrumentation is published. Don't know if we can automate this or somehow tell lerna to do this.

Rauno56:
> Not related to this PR. It seems the dependencies are not updated automatically here if an instrumentation is published. Don't know if we can automate this or somehow tell lerna to do this.

Don't you think open ranges for those dependencies takes care of that?

Flarna:
I think this could work. But I guess the range needs to be `*` as other ranges may skip parts - in special for the 0.x ranges.
But this removes quite some control from end users.

Rauno56:
Not sure I understand what you mean exactly. What do you think would be the ideal version specifiers for the instrumentation deps?

Flarna:
I think the optimum would be that this module is automatically updated whenever another module is released. And the version should be pinned. I assume here that updates of dependencies like `@opentelemtry/instrumentation` are done atomic here.
This gives end users an easy way to select the instrumentation version.

If ranges like `^0.28.0` are used you get only new patch versions, for `^1.2.0` aditionally minors. only if you use `*` as range you get latest.
But if this is done users get new instrumenations automatically except if they use a lock file.

Rauno56:
I don't have a strong preference, but `*` feels a bit too broad to me during a major bump from 0 to 1. On the other hand as long as the meta package is unstable it should be fine either way.

Rauno56:
Going through [release-please logs](https://github.com/open-telemetry/opentelemetry-js-contrib/runs/6309076237?check_suite_focus=true) and [release PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/947/files#diff-177ba81cee88a5a6744999b2d1f843ec1d5d4d99af5a58a938a32a6d6edb6854L49-L79), the dependencies does seem to be updated, @Flarna. That was exactly what you had in mind, right?

Flarna:
Yes, that seems ok. But maybe remove the `^` prefix?
So pinning `@opentelemetry/auto-instrumentations-node` results in pinning all instrumentations.

Rauno56:
I've now learned that pinning is probably even better idea than I thought because the version `@opentelemetry/instrumentation` package has to match for all of the packages for the tests to run, otherwise we'll get a type mismatch because of the concrete classes used in the instrumentation package. Ideally those should be interfaces instead, but that's a fix for the future.

This is also something to keep in mind in the future whenever we update any of the deps in the meta-pacakges.

Rauno56:
@Flarna, had to revert some of the changes in 098c2ed6f to bring `@opentelemetry/instrumentation` to the same version.

Rauno56:
I'll try it out. I guess I must've had something left behind for the whole dep tree to end up in a mess.

Flarna:
Maybe some leftover package-lock.json files? maybe try `git clean -fdx`.

Maybe https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1012 helps also as I found npm is not dedup in all cases.

Rauno56:
> Maybe some leftover package-lock.json files? maybe try `git clean -fdx`.

The CI also has the same issue:

```
Argument of type 'AmqplibInstrumentation | AwsLambdaInstrumentation | AwsInstrumentation | BunyanInstrumentation | ... 26 more ... | WinstonInstrumentation' is not assignable to parameter of type 'Instrumentation'.
  Type 'AmqplibInstrumentation' is not assignable to type 'Instrumentation'.
```

Rauno56:
> There should be nothing left using 0.27.0. All the instrumentations here (expect http and grpc) should be linked by lerna and main should be on 0.28.0 meanwhile.
> http/grpc from npm in version 0.28.0 should match.

I could achieve that by updating all instrumentations here to use `^0.28`.
`^0.28` and `^0.27` have no overlap because of the different behavior of the tilde in the 0.x range.

blumamir:
@Rauno56 - This PR is still relevant, right? just needs to be updated (mostly version numbers I guess)

Rauno56:
Yep .. I'll rebase and let's get it merged.

dyladan (approved):

PR #982, opened 2022-04-26
feat(instrumentation-redis): add support for redis@^4.0.0

### LINKED ISSUE #834: redis instrumentation - support v4
Looks like the `redis` package released [v4](https://www.npmjs.com/package/redis/v/4.0.0)

Out instrumentation currently [only supports v2 and v3](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/9326c99f3cdf3e0166f74093a8093066be78bd0a/plugins/node/opentelemetry-instrumentation-redis/src/instrumentation.ts#L53)

This issue is to add support to this instrumentation for the new v4.
This will also resolve [this PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/758)

blumamir:

- fixes #834 

The new version splits the functionality into sub-packages.
I added patches for the relevant function in the subpackage: `@node-redis/client`:
- patch function `extendWithCommands`, to allow access to `executor` which is the function that records spans upon command execution
- patch `RedisClientMultiCommand.exec` function to correctly handle transactions and `MULTI` commands.
- patch `RedisClient.multi` function to allow setting the client options on the multi command object to allow recording network attributes for transactions.

I added support for all existing config options.
Added tests for v4 + added it to test all versions and made sure it's green
Since v4 supports node >= 12, the package.json and CI actions were updated accordingly. This might be breaking for users of redis 2-3 with node < 12.

## Checklist


Rauno56:
If the implementation doesn't share a lot of code I'd split it into a totally separate package. Do you know of any drawbacks that approach might have?

blumamir:
> If the implementation doesn't share a lot of code I'd split it into a totally separate package. Do you know of any drawbacks that approach might have?

I support this too.

The drawbacks I can spot:
- Does not enforce consistent `types` - e.g. the instrumentation config is declared in two different packages. Future fixes and features could be easily added to just one and make them diverge. Not sure how much that actually matters though. 
- Might be slightly more complicated for consumers - they will have to check versions in the app and choose the right package to use. I guess it shouldn't be a big deal to wrap them both into a single package that installs both and expose them as a single package that covers all versions.
- Extra mental step in maintaining these packages - need to support two packages instead of one.

Benefits:
- Both variants will have tests in CI
- We can state different node version support for each package (redis 3 supports node >=8 but v4 supports node >= 12)
- Smoother development - correct version is found in devDependencies
- Each instrumentation can depend on the right types for the versions it instruments
- Less confusion for the codecov tool - will see tests that cover both patches for old and new versions, as opposed to just running it on a single set.

What do you think a name for a package such as this should be? `@opentelemetry/instrumentation-redis-2`(covers both 2 and 3), `@opentelemetry/instrumentation-redis-4` (covers current 4 and maybe future 5,6,...)?

Rauno56:
Naming things - the hardest part.

I don't have a strong preference. If I can come with good heuristics on this those would be:

1. Don't change the one that's there if possible;
2. Make what's included as clear as possible;
3. Make the default clear.

Based on that, I'd leave `@opentelemetry/instrumentation-redis` to include instrumentation for versions 2 and 3(violates my point nr 3, but anything else would violate nr 1) and `@opentelemetry/instrumentation-redis-4` to include one for versions 4 and up until they make another very significant API change.

blumamir:
> Naming things - the hardest part.
> 
> I don't have a strong preference. If had to come up with good heuristics on this, those would be:
> 
> 1. Don't change the one that's there if possible;
> 2. Make what's included as clear as possible;
> 3. Make the default clear.
> 
> Based on that, I'd leave `@opentelemetry/instrumentation-redis` to include instrumentation for versions 2 and 3(violates my point nr 3, but anything else would violate nr 1) and `@opentelemetry/instrumentation-redis-4` to include one for versions 4 and up until they make another very significant API change.

Like, I'll update the PR to express that.
Thanks

blumamir:
> If the implementation doesn't share a lot of code I'd split it into a totally separate package. Do you know of any drawbacks that approach might have?

Done. There are now 2 packages: `@opentelemetry/instrumentation-redis`, `@opentelemetry/instrumentation-redis-4`

plugins/node/opentelemetry-instrumentation-redis-4/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "@opentelemetry/instrumentation-redis-4",
+  "version": "0.29.0",
+  "description": "OpenTelemetry redis automatic instrumentation package.",
```
habmic, line 4: Should we indicate in the description it is specific to version 4?

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -0,0 +1,408 @@
...
+
+const DEFAULT_CONFIG: RedisInstrumentationConfig = {
+  requireParentSpan: false,
+};
+
+export class RedisInstrumentation extends InstrumentationBase<any> {
+  static readonly COMPONENT = 'redis';
+
+  constructor(protected override _config: RedisInstrumentationConfig = {}) {
+    super('@opentelemetry/instrumentation-redis', VERSION, _config);
```
habmic, line 60: should be with 4

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -0,0 +1,408 @@
...
+        }
+
+        execRes.then((redisRes: unknown[]) => {
+          const openSpans = this[OTEL_OPEN_SPANS];
+          if (!openSpans) {
+            return plugin._diag.error(
+              'cannot find open spans to end for redis multi command'
+            );
+          }
+          if (redisRes.length !== openSpans.length) {
```
habmic, line 251: If this case is true, wouldn't it cause a potential memory leak?

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -0,0 +1,408 @@
...
+      [SemanticAttributes.DB_CONNECTION_STRING]: clientOptions?.url,
+    };
+
+    try {
+      const dbStatement = dbStatementSerializer(commandName, commandArgs);
+      if (dbStatement != null) {
+        attributes[SemanticAttributes.DB_STATEMENT] = dbStatement;
+      }
+    } catch (e) {
+      this._diag.error('dbStatementSerializer throw an exception', e);
```
habmic, line 346: It might make it easier to debug if we add the commandName and commandArgs

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    const span = this.tracer.startSpan(
+      `${RedisInstrumentation.COMPONENT}-${commandName}`,
```
habmic, line 350: I think that according to the specification it span name should be without the dash:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

plugins/node/opentelemetry-instrumentation-redis-4/package.json
blumamir: added

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
blumamir: good catch. fixed

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
blumamir: The spans are stored on the `RedisClientMultiCommand` object which should be garbaged collected as soon as the `multi` request is resolved.
Even when this `if` is not hit, we do not clear this array from the object which I think is fine.
But I might be missing something 🤷‍♂️

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
blumamir: You are right.
I kept it consistent with the current implementation which set the span name like "redis-SET".
According to the example in the spec, the span name for redis should look something like this: "HMSET myhash", which is not very trivial to achive.

As this is the existing behavior, I suggest implementing this change in a separate - dedicated PR, which will change the span name for both instrumentation packages (v2 and v4)

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
blumamir: added `commandName`. Args can contain non-bound payload size, and sensitive information, so I didn't add it here, but I amo open to discussing it if you think otherwise.

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
habmic: You are correct about the RedisClientMultiCommand object, but what will happen to the spans which aren't ended? if the this if statement isn't hit, we won't iterate over the spans and end them and that may leave hanging spans, no?

blumamir: You are right.

This `if` is not expected to be triggered ever. If it does, that means bug in instrumentation or `redis` package. In this case, I am not able to correlate the open spans array with the responses array for this mutli command, so proper status and reply attributes cannot be set on the spans.

Not `end`ing the spans means there will be "holes" in the trace (broken trace), which is not ideal but also something that could always happen regardless of this issue. Since we do not anticipate this to happen, I am not very worried. Emitting the error message to `diag` logger should cover the rare case where there is some bug.

An alternative approach would be to end the spans with some attributes to indicate that reply and status could not be obtained. It is not hard to implement but there are no semantics or prior art to base upon as far as I am aware.

I feel comfortable with the current implementation. WDYT?

habmic: I see your point, I think it is a valid solution.

habmic (approved):

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -0,0 +1,410 @@
...
+                  'internal instrumentation error, missing transformCommandArguments function'
+                );
+                return moduleExports;
+              }
+
+              // this is the function that extend a redis client with a list of commands.
+              // the function patches the commandExecutor to record a span
+              this._diag.debug('Patching redis commands executor');
+              if (isWrapped(moduleExports?.extendWithCommands)) {
+                this._unwrap(moduleExports, 'extendWithCommands');
```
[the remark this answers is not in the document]
blumamir: Unfortunately not :(
The package was refactored in a way that is very hard to patch and instrument. Had to do some workarounds to be able to make it work.

It could be that I missed something. If you spot a better alternative I'll be happy to discuss it.

plugins/node/opentelemetry-instrumentation-redis-4/README.md
```diff
@@ -0,0 +1,59 @@
+# OpenTelemetry redis Instrumentation for Node.js
+
+[![NPM Published Version][npm-img]][npm-url]
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for [`redis@^4.0.0`](https://github.com/NodeRedis/node_redis).
```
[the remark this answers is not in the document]
blumamir: done. Thanks :)

blumamir:
redis package released version `4.1.0` 10 days ago (on 02 May 2022) which changed the internal function names and the internal package names.

This currently breaks the tests (and support) for this new version.
I haven't yet checked what should be fixed in the instrumentation for this.
In the meantime, updating here about the status. Hope I can submit a fix soon.

blumamir:
@habmic @Rauno56 @dyladan 
After investigating the change, I found out that in version `4.1.0` the internal package namespace in npm changed from `@node-redis/client` to `@redis/client`, and the a patched function changed it's name from `extendWithCommands` to `attachCommands`.

I udpate the PR accordingly and now test-all-versions is green again for all versions on my local setup (node 16) and CI is green for the latests version (4.1.0) except for node 12.

For node 12 it seems to fail because of an issue with redis package itself, documented [here](https://github.com/redis/node-redis/issues/2119), [here](https://github.com/redis/node-redis/issues/2117) and [here](https://github.com/redis/node-redis/issues/2131). The package itself does not publish the supported node version anywhere in a clear way, but [I found this text in the changelog of the relevant subpackage of the latest version](https://github.com/redis/node-redis/releases/tag/client%401.1.0):

> Drop support for node 12

I am going to remove the tests for redis 4 for node 12

blumamir:
CI is now green.
Feel free to review again after the latest changes. This PR is ready to merge as far as I am concerned.

vmarchaud (approved):

Rauno56:
@blumamir, feel free to resolve the conflicts and merge right away.

PR #992, opened 2022-04-30
fix(nestjs): remove nestjs type from exported Instrumentation

### LINKED ISSUE #985: nestjs depends on types from devDependency, failing compilation
`1.0.1`

`v16.14.0`

### What did you do?
`yarn add @opentelemetry/instrumentation-nestjs-core` -> `"@opentelemetry/instrumentation-nestjs-core": "0.28.3"` and then:

```js
import { NestInstrumentation } from '@opentelemetry/instrumentation-nestjs-core';
new NestInstrumentation();
```

This package does not depend on nestjs directly.

When `yarn build`ing the package, following error printed to console:
```
../../node_modules/@opentelemetry/instrumentation-nestjs-core/build/src/instrumentation.d.ts:2:30 - error TS2307: Cannot find module '@nestjs/core' or its corresponding type declarations.

2 import type * as NestJS from '@nestjs/core';
                               ~~~~~~~~~~~~~~


Found 1 error.
```

This message is valid, as the nestjs instrumentation indeed [does this](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/cb915cfe018678060b014e1385fcefd0dc73c776/plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts#L25):
```js
import type * as NestJS from '@nestjs/core';
export class Instrumentation extends InstrumentationBase<typeof NestJS> {
```

which transpile into the `instrumentation.d.ts` file in `build` as:
```js
import type * as NestJS from '@nestjs/core';
export declare class Instrumentation extends InstrumentationBase<typeof NestJS> {
```

So in order for this package to work correctly in typescript, we will need to include these types in the regular "dependencies" which also means dragging a dependency on the entire `@nestjs/core` for all users just for the type.

Another alternative would be to remove the type from InstrumentationBase, replacing it with `any` and thus having no type import in the public interface. It seems to not be used for anything anyhow.
```js
export class Instrumentation extends InstrumentationBase<typeof NestJS> {

// Change to 
export class Instrumentation extends InstrumentationBase<any> {
```

[The other type imports from `@nestjs/*`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/cb915cfe018678060b014e1385fcefd0dc73c776/plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts#L26) are not exported to the public interface so I think they are safe to use in `instrumentation.ts`:
```js
import type { NestFactory } from '@nestjs/core/nest-factory.js';
import type { RouterExecutionContext } from '@nestjs/core/router/router-execution-context.js';
import type { Controller } from '@nestjs/common/interfaces';
```

If there is support for this change, I can work on a PR to add it.

@Rauno56

blumamir:

- fixes #985 for nestjs instrumentation, allowing the package to be used in services which do not import `@nestjs/core`.

- removed the generics type from `InstrumentationBase`. It was not used for anything anyway.

Before the change this import is found in `instrumentation.d.ts`:
```js
import type * as NestJS from '@nestjs/core';
export declare class Instrumentation extends InstrumentationBase<typeof NestJS> {
```

`@nestjs/core` is not a dependency of the instrumentation, and if the service does not import it, the compilation error is emitted.

After the change, there is no longer import of this package:
```js
export declare class Instrumentation extends InstrumentationBase<any> {
```

## Checklist


Flarna (approved):

plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts
```diff
@@ -22,15 +22,14 @@ import {
   InstrumentationNodeModuleFile,
   isWrapped,
 } from '@opentelemetry/instrumentation';
-import type * as NestJS from '@nestjs/core';
```
Rauno56, line 25: Aren't next two lines causing the same issue?

Flarna: no, because they are only used in file local functions and not in exported APIs. The .d.ts files hold only the exported types.

Rauno56 (approved):

Issue #784, opened 2021-12-10
Add support for graphql@16


Rauno56:
### Is your feature request related to a problem? Please describe

Tests currently fail on `graphql@16`. Haven't investigated further.

vmarchaud:
@obecny Do you still plan to do this ?

obecny:
> @obecny Do you still plan to do this ?

I don't recall I was planning to do this.

hilladalins:
Can you please assign this issue to me?

Rauno56:
Fixed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/998

PR #998, opened 2022-05-03
feat: support `graphql` v16

dchambers:

- Allow developers to use any version of the `graphql` library they want to by making `graphql` a peer dependency.
- Allow developers to use `graphql` v16 by introducing an adaptor function to abstract away the API change in v16.

- As per the problem solving statement, but defined as three distinct commits:
   1. Prove that `graphql` can be made a peer dependency. 🟢 
   2. Start testing against `graphql` v16. 🔴 
      - Green on CI because CI doesn't run `test-all-versions`, but verified as failing locally on my machine.
   3. Introduce the `graphql` adaptor function to fix the broken tests. 🟢 
      - Again, verified `test-all-versions` on my machine, but this time where all versions pass again.

## Checklist


dchambers:
This PR fixes #784.

Rauno56:
Thanks for the PR. I appreciate the clear PR description and commit structure. :+1: 

> Allow developers to use any version of the graphql library they want to by making graphql a peer dependency.

Instrumentation users can actually already use any version. Moving it to peer dependency only makes it installed by default on new npm versions.

> Green on CI because CI doesn't run test-all-versions, but verified as failing locally on my machine.

Do I understand you correctly that tests fail on graphql v16? How does that qualify supporting it in that case or is it still WIP?

dchambers:
> Instrumentation users can actually already use any version. Moving it to peer dependency only makes it installed by default on new npm versions.

Hi @Rauno56 👋. So prior to this PR users could only use versions other than `graphql` v15 by forcing it using something like `resolutions` (assuming they use Yarn), plus they had no way of knowing whether it would work (it didn't for v16 for example, and I haven't bothered to test v14 as it's now pretty old, and I don't need it).

The peer dependency approach, which includes a formal specification of which peer versions are supported, is what's used by the vast majority of `graphql` dependent NPM libraries that we currently use (50 out of 53 to be exact), with this being one of the three libraries we use that doesn't currently work that way.

> Do I understand you correctly that tests fail on graphql v16? How does that qualify supporting it in that case or is it still WIP?

No, `graphql` v16 is made to work in commit number 3. I only point out that it doesn't initially work (in commit number 2) to demonstrate the need for the changes in that third commit. This isn't reflected in the CI results however since `test-all-versions` is only run nightly, rather than being run on every single PR, and therefore I confirm that I've run it locally to make up for, what is otherwise, a lack of evidence for the utility of the PR.

plugins/node/opentelemetry-instrumentation-graphql/.tav.yml
```diff
@@ -1,4 +1,4 @@
 graphql:
-  # Taking a sample from the most downloaded versions in the range "14 || 15"
-  versions: "^15.8.0 || 15.7.2 || 15.6.1 || 15.6.0 || 15.5.3 || 15.5.1 || 15.5.0 || 15.4.0 || 15.3.0 || 14.6.0 || 14.5.8 || 14.5.0 || 14.0.0"
+  # Taking a sample from the most downloaded versions in the range "14 || 15 || 16"
+  versions: "^16.4.0 || ^16.0.0 || ^15.8.0 || 15.7.2 || 15.6.1 || 15.6.0 || 15.5.3 || 15.5.1 || 15.5.0 || 15.4.0 || 15.3.0 || 14.6.0 || 14.5.8 || 14.5.0 || 14.0.0"
```
Rauno56, line 3:
```suggestion
  versions: "^16.4.0 || 16.3.0 || 16.2.0 || ^15.8.0 || 15.7.2 || 15.6.1 || 15.6.0 || 15.5.3 || 15.5.1 || 15.5.0 || 15.4.0 || 15.3.0 || 14.7.0 || 14.6.0 || 14.5.8 || 14.0.0"
```

Please remove overlapping ranges. I additionally adjusted some of the ranges.
The rule here is
1. to have as many versions in as possible without making it too painful to run TAV for the next dev;
2. to have open ranges where we know that active development is still done - hence `^16.4.0` is open-ended.

plugins/node/opentelemetry-instrumentation-graphql/package.json
```diff
@@ -44,13 +44,15 @@
...
-    "@opentelemetry/api": "^1.0.0"
+    "@opentelemetry/api": "^1.0.0",
+    "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
   },
   "devDependencies": {
     "@opentelemetry/api": "1.0.2",
     "@opentelemetry/sdk-trace-base": "1.2.0",
     "@types/mocha": "8.2.3",
     "@types/node": "16.11.21",
+    "graphql": "^15.5.1",
```
Rauno56, line 55: Why not update it among the dev-deps as well to the latest and leave testing on v15 for TAV?

Rauno56 (commented):
Sorry, I missed the fact that we had `graphql` listed under the dependencies not only as a devDep. It's not common among the packages here.

Flarna:
graphql was move between dependency and devDependency a few times.
* if it is a dependency - in special with a pinned version - it tends to collide with the user version
* if it is a dep dependency it results in transpile problems if user has no graphql installed because it needs the types.

As it is part of the autoinstrument metapackage it's a common setup that users have the instrumentation installed even if they don't use graphql.

Refs: #754 and #850

plugins/node/opentelemetry-instrumentation-graphql/package.json
dchambers: Thanks for the great feedback @Rauno56. Those changes are now made.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -307,8 +307,8 @@ export class GraphQLInstrumentation extends InstrumentationBase {
         schema: graphqlTypes.GraphQLSchema,
         documentAST: graphqlTypes.DocumentNode,
         rules?: ReadonlyArray<graphqlTypes.ValidationRule>,
-        typeInfo?: graphqlTypes.TypeInfo,
-        options?: { maxErrors?: number }
+        options?: { maxErrors?: number },
+        typeInfo?: graphqlTypes.TypeInfo
```
dchambers, line 311: `graphql` v16 flips around these two parameters 🤯 . This has no effect at runtime when using v15 Vs v16 because they both pass whatever they receive through, but needs to be changed in the case we want to compile against v16.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -387,7 +387,7 @@ export function wrapFieldResolver<TSource = any, TContext = any, TArgs = any>(
             >
         >(
           () => {
-            return fieldResolver.call(this, source, args, contextValue, info);
+            return fieldResolver.call(this, source, args, contextValue, info) as any;
```
dchambers, line 390: The `any` used to be baked into the generic return type, whereas it now uses `unknown` instead, which is insufficient here.

plugins/node/opentelemetry-instrumentation-graphql/test/graphql.test.ts
```diff
@@ -976,7 +979,7 @@ describe('graphql', () => {
       assert.deepStrictEqual(spans.length, 1);
     });
 
-    it('should instrument parse with error', () => {
+    itMaybe('should instrument parse with error', () => {
```
dchambers, line 982: This test doesn't work on Node.js v8 and Node.js v10 when using `graphql` v16, so make only run it on more recent versions of Node.js.

plugins/node/opentelemetry-instrumentation-graphql/package.json
dchambers: > 

@Rauno56, I finally had some time to get the tests working while using v16 internally. However, since `graphql` v16 doesn't appear to support Node.js v8 or Node.js v10, that has meant conditionally disabling a few of the tests. Perhaps, therefore, you might prefer to use the latest `15.x` internally, so that running `npm test` locally exercises all tests.

BTW the failing code coverage check appears to be as a result of an `as Error` cast I perform, which is weird since casting is only a notional concept, and does not translate to code paths that can be tested.

dchambers:
HI @Rauno56 / @dyladan  👋 . A week has passed by. Anyway to get this PR moving again?

As a recap it's ready to merge now with some quirks due to upgrading internally to use v16, or there's the option to use v15 internally if you prefer not to have the quirks?

dchambers:
ℹ️  If you're using Yarn then you can actually just force it to use `graphql` v16 using the `resolutions` block since the code changes in this PR are only needed for the test code and for the TypeScript types, and are not required for the running of the plugin itself. Here's the config I used in my case:

```
"resolutions": {
  "@opentelemetry/instrumentation-graphql@npm:0.27.4/graphql": "^16.5.0",
},
```

vmarchaud:
I'm not the right person to review this since i don't even know how graphql works but @open-telemetry/javascript-approvers if someone can that would be nice

vmarchaud:
@Rauno56 i saw you made modification, are you ok with the current state of the PR ? I can review it if you can more eyes so we can merge this

Rauno56:
@dchambers , I removed the maybe-functionality in tests.

@vmarchaud, your review would be helpful for sure.

Rauno56 (approved):
Thanks for the contribution, @dchambers. Sorry, that it took so long to get merged.

@vmarchaud, nvm about my comment above. I think we can just merge it.

vmarchaud (approved):

PR #1002, opened 2022-05-04
feat: enable a way to disable amqp tests that require a server running

Rauno56:

Currently running `npm test` without any services running errors on `amqp` tests not being able to connect to the server.
Even though I'd generally do it the other way around - user having to disable the tests by default, not enable them - it's the current convention.

Match the test convention in amqp instrumentation with `process.env.RUN_*_TESTS`-approach.


vmarchaud (approved):

blumamir (approved):
LGTM
Thanks!

PR #1012, opened 2022-05-06
chore: depend on API ^1.0.0 also in devDependencies

Flarna:

Not sure if this actually solves a problem or creates new ones. But it seems pinning API results in duplicates installed by at least some NPM versions.

Avoid API duplication during npm install by allowing a range similar as in other modules.

## Checklist


scripts/peer-api-check.js
```diff
@@ -15,7 +15,6 @@
  */
 
 const path = require('path');
-const semver = require('semver');
```
blumamir, line 18: can also be removed from 'scripts/package.json`

blumamir (approved):

scripts/peer-api-check.js
Flarna: done

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -21,24 +21,24 @@
...
+    "@opentelemetry/api": "^1.0.0",
+    "@opentelemetry/exporter-jaeger": "^1.0.0",
+    "@opentelemetry/exporter-zipkin": "^1.0.0",
+    "@opentelemetry/instrumentation": "^0.28.0",
     "@opentelemetry/instrumentation-express": "^0.28.0",
-    "@opentelemetry/instrumentation-http": "^0.27.0",
-    "@opentelemetry/resources": "^1.1.1",
-    "@opentelemetry/sdk-trace-base": "^1.1.1",
-    "@opentelemetry/sdk-trace-node": "^1.1.1",
-    "@opentelemetry/semantic-conventions": "^1.1.1",
```
Rauno56, line 41: Here, because in other locations, sdk packages are not updated.

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -49,7 +49,7 @@
     "@opentelemetry/core": "^1.0.0",
     "@opentelemetry/instrumentation": "^0.28.0",
     "@opentelemetry/semantic-conventions": "^1.0.0",
-    "@opentelemetry/propagation-utils": "^0.27.0"
```
Rauno56, line 52: Here.

Rauno56 (commented):
Could you please comment on the changes other than making api `^1.0.0` in dev deps.  Not that they must not be here, but I'd just like to understand the premise.

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -49,7 +49,7 @@
...
+    "@opentelemetry/propagation-utils": "0.27.0"
```
Flarna, line 52: pin dependencies within the same learna monorep, forgot this one in a previous PR #984

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -21,24 +21,24 @@
     "tracing"
   ],
   "engines": {
-    "node": ">=8"
+    "node": ">=8.12.0"
```
Flarna, line 24: keep in sync with remaining modules

scripts/peer-api-check.js
```diff
@@ -34,16 +33,10 @@ if (pjson.dependencies && pjson.dependencies['@opentelemetry/api']) {
...
 const devVersion = pjson.devDependencies && pjson.devDependencies['@opentelemetry/api'];
 if (peerVersion) {
-  // error if not pinned
-  if (!/^[0-9]/.test(devVersion)) {
-    throw new Error(
-      `Package ${pjson.name} does't have API version pinned in dev dependencies: ${devVersion}`
-    );
-  }
-  if (!semver.satisfies(devVersion, peerVersion)) {
+  if (devVersion !== peerVersion) {
```
Flarna, line 36: update API peer dep check to allow ranges like ^1.0.0

plugins/node/opentelemetry-instrumentation-express/examples/package.json
Flarna: done, mostly cleanup oversights from previous PRs or merge races.

plugins/node/opentelemetry-instrumentation-express/examples/package.json
```diff
@@ -21,24 +21,24 @@
...
-    "@opentelemetry/instrumentation-http": "^0.27.0",
-    "@opentelemetry/resources": "^1.1.1",
-    "@opentelemetry/sdk-trace-base": "^1.1.1",
-    "@opentelemetry/sdk-trace-node": "^1.1.1",
-    "@opentelemetry/semantic-conventions": "^1.1.1",
+    "@opentelemetry/instrumentation-http": "^0.28.0",
+    "@opentelemetry/resources": "^1.0.0",
+    "@opentelemetry/sdk-trace-base": "^1.0.0",
+    "@opentelemetry/sdk-trace-node": "^1.0.0",
+    "@opentelemetry/semantic-conventions": "^1.0.0",
```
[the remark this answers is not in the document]
Rauno56: #984 has some of the same dependencies pinned for the [browser extension package](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/984/files#diff-d5c5343c2663b62222aa864aac1719f09508de66f31856492f984163908f10f0L69-L81) but not others, afaict. Why is that?
The extension pacakge also participates in hoisting.

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
Rauno56: Please help me understand why are we supposed to pin "monorepo-local" packages / why should we make those follow an different rule than external pacakges.

There are more:

- `@opentelemetry/propagator-aws-xray@^1.0.1` in `@opentelemetry/instrumentation-aws-lambda`
- `@opentelemetry/instrumentation-express@^0.28.0` in `express-example`
- `@opentelemetry/instrumentation-document-load@^0.27.1` in `@opentelemetry/browser-extension-autoinjection`
- all instrumentations in `@opentelemetry/auto-instrumentations-*`

Are these also supposed to be pinned? Why not?

Flarna: We do it in core repo and I see no reason to do it different here. They are released in one shot and user should update all of them or nothing. So it's always consistent as during tests here.
At least this is my understanding based on the setup in core repo.

Whenever I had a mixup of versions it ended up in a nightmare.

And yes, you are right, there are a few dependencies I haven't seen yet. In general I have the feeling that dependency management in this (and core) repo should be more automated but till now I found no time to get familiar with renovate bot.

plugins/node/opentelemetry-instrumentation-express/examples/package.json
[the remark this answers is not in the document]
Flarna: I pinned dev-dependencies from other repos as this seems to be the standard setup here and in core.

Not pinning dependencies to local repos was not my focus that time but I think it should be done. But I guess for dev-deps it's anyway don't care as they have no influence to end user and lerna links pinned and non pinned dependencies.

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
Flarna: I'm still not sure what is the best way to setup the dependencies. It seems that every release made has it's own followup problems regarding dependencies. Not sure if we are able to get to a setup which just works.

Rauno56: > We do it in core repo and I see no reason to do it different here. They are released in one shot and user should update all of them or nothing. So it's always consistent as during tests here.

I'm all for consistency, just trying to understand the reasoning behind it. So the main reason is to mimic the repository setup since they are linked in CI. That makes sense.

> And yes, you are right, there are a few dependencies I haven't seen yet. In general I have the feeling that dependency management in this (and core) repo should be more automated but till now I found no time to get familiar with renovate bot.

I'm also totally for automation, but currently I personally don't even always understand the rules to automate them. Hence me bombarding you with questions. :)

plugins/node/opentelemetry-instrumentation-express/examples/package.json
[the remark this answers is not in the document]
Rauno56: So should we achieve consistency between this here and browser extension package?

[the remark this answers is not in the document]
Flarna: Yes, I think we should be consistent.

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
Rauno56: By "examples I listed above" I meant:

> There are more:
>
> - @opentelemetry/propagator-aws-xray@^1.0.1 in @opentelemetry/instrumentation-aws-lambda
> - @opentelemetry/instrumentation-express@^0.28.0 in express-example
> - @opentelemetry/instrumentation-document-load@^0.27.1 in @opentelemetry/browser-extension-autoinjection
> - all instrumentations in @opentelemetry/auto-instrumentations-*

Which by the rule of "monorepo local deps should be pinned" are currently inconsistent.

What comes to `@opentelemetry/instrumentation` package, I'd leave that to `^0.27` for now **and** specify http and grpc instrumentations accordingly to a version that uses `^0.27` - we can then update that separately. What do you think?

Flarna: > What comes to @opentelemetry/instrumentation package, I'd leave that to ^0.27 for now and specify http and grpc instrumentations accordingly to a version that uses ^0.27 - we can then update that separately. What do you think?

If we depend on old http instrumentation this will pull in old SDK components and likely to new problems.

Flarna:
created a followup to harmonize more then API: #1020

PR #1015, opened 2022-05-08
fix(sns-publish-test-v3): add test for sns.publish for aws sdk v3

haddasbronfman:

- Add test for sns.publish() in aws sdk v3

- Created test (similar to sns.publish v2), that uses '@aws-sdk/client-sns' to call sns.publish()

## Checklist


plugins/node/opentelemetry-instrumentation-aws-sdk/test/sns.test.ts
```diff
@@ -58,7 +61,7 @@ describe('SNS', () => {
     } as AWS.SNS.Types.PublishResponse);
   });
 
-  describe('publish', () => {
+  describe('publish - v2', () => {
```
blumamir, line 64: nit: I think this current describe is already under the one named "SNS - v2", so we can omit the V2 here

plugins/node/opentelemetry-instrumentation-aws-sdk/test/sns.test.ts
```diff
@@ -22,6 +22,9 @@ const instrumentation = registerInstrumentationTesting(
   new AwsInstrumentation()
 );
 import * as AWS from 'aws-sdk';
+import { SNS } from '@aws-sdk/client-sns';
```
blumamir, line 25: I recommend renaming this import to include `v3` in the name. Since the client is called `SNS` for both v2 and v3, it can be confusing for future readers.
Something like:
```js
import { SNS as SNSv3 } from '@aws-sdk/client-sns';
```

Since this PR introduces testing of 2 versions in the same file, I also recommend renaming the v2 import:
```js
import * as AWSv2 from 'aws-sdk';
```

blumamir (commented):
Thanks for adding these tests.
LGTM. added minor nits.

PR #1020, opened 2022-05-12
chore: harmonize dependencies

Flarna:

Harmonize dependencies by using following rules:
* peer and dev depend on API ^1.0.0 (like in core repo)
* depend on SDK ^1.0.0 (don't limit user)
* dev-depend on latest SDK pinned (like other deps)
* pin dependencies within the lerna project (like in core repo)

This is a followup of #1012. I decided to create a new PR as it is no longer about API deps only

## Checklist


plugins/node/instrumentation-fs/package.json
```diff
@@ -42,14 +42,13 @@
...
   "devDependencies": {
-    "@opentelemetry/api": "1.0.4",
+    "@opentelemetry/api": "^1.0.0",
     "@opentelemetry/context-async-hooks": "1.2.0",
     "@opentelemetry/resources": "1.2.0",
     "@opentelemetry/sdk-trace-base": "1.2.0",
     "@opentelemetry/sdk-trace-node": "1.2.0",
     "@types/mocha": "7.0.2",
     "@types/node": "14.17.9",
-    "@types/shimmer": "1.0.2",
```
Flarna, line 52: seem unused

dyladan: shimmer shouldn't be used in any instrumentation directly

Flarna: I know, but it turned out that dns uses types only in tests and `@opentelemetry/plugin-react-load` really uses shimmer.

As far as I remember `@opentelemetry/plugin-react-load` is somewhat different as it doesn't depend on `@opentelemetry/instrumentation`. I think it's what the name tells - a plugin for react.

packages/opentelemetry-browser-extension-autoinjection/package.json
```diff
@@ -66,18 +66,18 @@
...
-    "@opentelemetry/exporter-zipkin": "1.2.0",
-    "@opentelemetry/instrumentation": "0.28.0",
-    "@opentelemetry/instrumentation-document-load": "^0.27.1",
+    "@opentelemetry/context-zone": "^1.0.0",
+    "@opentelemetry/core": "^1.0.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.28.0",
+    "@opentelemetry/exporter-zipkin": "^1.0.0",
+    "@opentelemetry/instrumentation": "^0.28.0",
+    "@opentelemetry/instrumentation-document-load": "0.27.1",
     "@opentelemetry/instrumentation-fetch": "0.28.0",
```
blumamir, line 75: I think this should be with caret as it's a core repo dependency

blumamir (approved):

Rauno56 (approved):

PR #1022, opened 2022-05-13
add span attributes hook to express middleware

dmathieu:

Our organization has an internal telemetry library that we use to configure the otel SDK on services. Then, services only have to use the API, and never directly the SDK.

We want to do some specific attribute additions to express spans globally. While we currently can set those up directly in every service, setting up this hook would make it easier to do the same across multiple services.

This sets up a new option to the express instrumentation plugin, `spanAttributesHook` which, when setup, will allow overriding the default attributes from created/updated spans.
In the case of created spans, the returned attributes will be set on span creation.
In the case of updated spans from the HTTP plugin, the returned attributes will be added to the existing span.

## Checklist


dyladan:
Hmm it sounds like you might be looking for resource attributes which are applied to every span implicitly? I'm hesitant to add a hook that applies attributes to every single span like this.

vmarchaud:
I agree with Daniel that if those attributes are static they might be better within the resources, however if they depend on the request i think its preferable to replace exisiting hook with a generic one that allows to customize both the name and attributes at once to avoid adding hooks for each possible mutation

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -97,4 +102,5 @@ export interface ExpressInstrumentationConfig extends InstrumentationConfig {
   /** Ignore specific layers based on their type */
   ignoreLayersType?: ExpressLayerType[];
   spanNameHook?: SpanNameHook;
+  spanAttributesHook?: SpanAttributesHook;
```
blumamir, line 105: In other instrumentations we usually prefer to use a hook that receives the `span` object itself and then the hook can call `span.setAttribute(...)` to modify the relevant attributes.

Can this pattern fit here as well? or is there a reason to implement it differently in this case?

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
```diff
@@ -169,4 +169,140 @@ describe('ExpressInstrumentation hooks', () => {
...
+          if (layerType) {
+            return {
+              route,
+            };
+          }
+
+          if (route === '*') {
+            return {
+              method: request.method,
+              url: request.url,
```
blumamir, line 206: It's completely valid to use any string for the attribute name, but in OpenTelemetry those specific attributes are defined in semantic conventions specification and are called `http.method`(`SemanticAttributes.HTTP_METHOD`) and `http.url`(`SemanticAttributes.HTTP_URL`).

I think it will be more appropriate to use semantic names here because current values promote non-spec compliant usage.

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
```diff
@@ -169,4 +169,140 @@ describe('ExpressInstrumentation hooks', () => {
...
+              'http.route': '*',
+            }
+          );
+        }
+      );
+    });
+
+    it('should use the base attributes when returning undefined from hook', async () => {
+      const spanAttributesHook: SpanAttributesHook = () => {
+        return undefined as unknown as SpanAttributes;
```
blumamir, line 276: Looks like implementation supports returning the value `undefined` from the hook to signal no change in attributes. What do you think about adding it to the function return type in `types.ts`. This way we don't have to do these non-idiomatic castings.

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -58,9 +58,10 @@ Express instrumentation has few options available to choose from. You can set th
 
 | Options | Type | Example | Description |
 | ------- | ---- | ------- | ----------- |
-| `ignoreLayers` | `IgnoreMatcher[]` | `[/^\/_internal\//]` | Ignore layers that by match. |
-| `ignoreLayersType`| `ExpressLayerType[]` | `['request_handler']` | Ignore layers of specified type. |
-| `spanNameHook` | `SpanNameHook` | `() => 'my-span-name'` | Can be used to customize span names by returning a new name from the hook. |
+| `ignoreLayers`       | `IgnoreMatcher[]`    | `[/^\/_internal\//]` | Ignore layers that by match. |
+| `ignoreLayersType`   | `ExpressLayerType[]` | `['request_handler']` | Ignore layers of specified type. |
+| `spanNameHook`       | `SpanNameHook`       | `() => 'my-span-name'` | Can be used to customize span names by returning a new name from the hook. |
+| `spanAttributesHook` | `SpanNameHook`       | `() => { 'key': 'value' }` | Can be used to customize span attributes by returning a new list of attributes from the hook. |
```
blumamir, line 64: [I think that the value `() => { 'key': 'value' }` is not valid in this context](https://stackoverflow.com/questions/28770415/ecmascript-6-arrow-function-that-returns-an-object), as `{` is interpreted as function's body.

You should use parentess like this:
```js
`() => ({ 'key': 'value' })`
```

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -196,9 +196,15 @@ export class ExpressInstrumentation extends InstrumentationBase<
...
           .filter(path => path !== '/' && path !== '/*')
           .join('');
-        const attributes: SpanAttributes = {
-          [SemanticAttributes.HTTP_ROUTE]: route.length > 0 ? route : '/',
-        };
+        const attributes: SpanAttributes = instrumentation._getSpanAttributes(
+          {
+            request: req,
+            route,
+          },
```
blumamir, line 203: `layerType` is missing on this object

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
```diff
@@ -169,4 +169,140 @@ describe('ExpressInstrumentation hooks', () => {
...
+    });
+
+    it('should set span attributes', async () => {
+      instrumentation.setConfig({
+        spanAttributesHook: ({ request, route, layerType }, baseAttributes) => {
+          if (layerType) {
+            return {
+              route,
+            };
+          }
```
blumamir, line 201: Can this `if` statement be  `truthy` in the test? Are there any assertions to verify this code path?

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
```diff
@@ -169,4 +169,140 @@ describe('ExpressInstrumentation hooks', () => {
...
+
+          if (route === '*') {
+            return {
+              method: request.method,
+              url: request.url,
+            };
+          }
+
+          return baseAttributes;
```
blumamir, line 210: Is this path being invoked in the test?

blumamir (changes requested):

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -220,6 +226,15 @@ export class ExpressInstrumentation extends InstrumentationBase<
             `${req.method} ${route.length > 0 ? route : '/'}`
           );
           rpcMetadata.span.updateName(name);
+          rpcMetadata.span.setAttributes(
```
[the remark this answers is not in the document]
dmathieu: Yes. Just like the `spanNameHook` will change the name of the `http` span.

plugins/node/opentelemetry-instrumentation-express/src/types.ts
dmathieu: I'm fine using this approach, but that means changing the API for the plugin, since it would also remove `spanHookName`.

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
dmathieu: Sure. I used those values here, same as they are for the name hook. I'll open another PR to change them in there as well.

dmathieu: Sure. I just reused this as it is for the name hook. I'll change it for the attributes hook in this PR and open another one for name.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
dmathieu: It's not passed either when setting the span name. Should I add it to both?

blumamir: It is [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5f0d1e4e43412d6a6598e2b6703f0bead4f60d13/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L239)

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
dmathieu: It isn't. I reused it from the span name hook. The tests that could use it set their own hook. So I've removed it.

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
dmathieu: It can't be. I reused it from the name hook. I've removed it.

dmathieu:
I'm not looking to add attributes on every span. I'm looking to add them on every express span.

plugins/node/opentelemetry-instrumentation-express/src/types.ts
blumamir: Yes it will be a breaking change. We can also add the new hook and keep `spanHookName` for now and mark it deprecated.
@vmarchaud @dyladan  WDYT?

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
dmathieu: Ah, indeed. I've changed the PR to properly include it (which brings back/explains the need for layerType in tests).

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
dmathieu: Even though it isn't used/usable, it's actually required to get the linter to pass.

Rauno56:
We can just deprecate the name hook for now and add the more generic one and refactor the code that's using the name-hook to use the more generic one internally. That's not a breaking change, but satisfies the convention and is more common amongst the instrumentations.

dmathieu:
I'll setup a more generic hook next week.

PR #1028, opened 2022-05-18
feat(mysql2): support Connection.execute

aptomaKetil:

The current instrumentation for mysql2 does not track queries made with [`Connection.execute`](https://www.npmjs.com/package/mysql2#using-prepared-statements), which is the helper method for preparing and executing a statement.

Since `Connection.query` and `Connection.execute` have the same signature and response formats (where it matters), all the existing wrapper code can simply be reused.

Tests for `Connection.query` and `Pool.query` have been mirrored for `.execute`.

## Checklist

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
@@ -68,6 +68,15 @@ export class MySQL2Instrumentation extends InstrumentationBase<
             this._patchQuery(moduleExports.format) as any
           );
 
+          if (isWrapped(ConnectionPrototype.execute)) {
+            this._unwrap(ConnectionPrototype, 'execute');
+          }
+          this._wrap(
+            ConnectionPrototype,
+            'execute',
+            this._patchQuery(moduleExports.format) as any
+          );
+
           return moduleExports;
         },
         (moduleExports: any) => {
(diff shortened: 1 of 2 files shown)


Rauno56 (approved):
Looks good. Thanks, @aptomaKetil, and congrats for the first contribution!
I'll keep the PR open for a day or so more for others to see and react to it.

Please sign the CLA.

blumamir:
> Since `Connection.query` and `Connection.execute` have the same signature

I am not sure that's right. [`execute` accepts calls like this](https://github.com/sidorares/node-mysql2/blob/ab3feb75bb4af1f6d288c0494a99aa2483ac24d5/lib/connection.js#L599):

```js
// execute(options, cb)
```

Where the `sql` value is set in options and not as first function parameter

blumamir (approved):

dyladan (approved):

dyladan:
I thought in some versions of this `query` actually called `execute` which would result in double spans. Are you sure this is not the case?

aptomaKetil:
> I thought in some versions of this query actually called execute which would result in double spans. Are you sure this is not the case?

The tests for execute all have assertions on the number of spans produced by the instrumentation, so this is covered. I'm unaware of any previous discussion around this. I've taken a brief look at the mysql2 source and as far as I can tell, query and execute are on completely different code paths.

Rauno56:
Thanks, @aptomaKetil and congrats for your first contribution!

PR #1036, opened 2022-05-24
fix(instrumentation-mongodb): update mongodb to version without missing types

### LINKED ISSUE #1031: Broken auto-instrumentation dependencies
0.28.0

v18.0.0

### What did you do?
Ran a `npm run build && APP_ENV=tests node_modules/.bin/mocha --recursive --exit dist/test/**/*.spec.js`

dyladan:
Fixes #1031

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mongodb/package.json
+++ b/plugins/node/opentelemetry-instrumentation-mongodb/package.json
@@ -58,7 +58,7 @@
     "@types/node": "16.11.21",
     "gts": "3.1.0",
     "mocha": "7.2.0",
-    "mongodb": "3.6.11",
+    "mongodb": "4.6.0",
     "nyc": "15.1.0",
     "rimraf": "3.0.2",
     "test-all-versions": "5.0.1",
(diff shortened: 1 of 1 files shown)


plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/node": "16.11.21",
     "gts": "3.1.0",
     "mocha": "7.2.0",
-    "mongodb": "3.6.11",
+    "mongodb": "4.6.0",
```
Flarna, line 61: Having mongodb in dev deps only will result in missing types for consumers.
Adding a specific mongodb version as dependency seems not good.

dyladan: I agree. thats why this is a draft. i'm trying to figure out other options. Unfortunately, I think we're going to have to stop trying to type check modules we're patching. causing too many dependency troubles.

Flarna: For the patching part it's quite simple as the instrumentation base classes don't need/use the types at all.
The problem are the hooks offered by some instrumentations as they are APIs to end users. We could use `unknown`/`any` there and just document what is provided and user as to cast. This would even cover differences between versions.

blumamir, line 61: This is deliberately set to version 3 as it is what we run in CI when invoking `npm run test`:

```
    "test": "nyc ts-mocha -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v3.test.ts'",
```

Since we can only run a single version, version 3 was choosen as it is more common

PR #1044, opened 2022-05-31
fix(aws-sdk): avoid repeating MessageAttributeNames in sqs receiveMessage

### LINKED ISSUE #1038: SQS ReceiveMessages crashes with `413 REQUEST ENTITY TOO LARGE`

```bash
❯ grep opentelemetry package.json
    "@opentelemetry/api": "^1.1.0",
    "@opentelemetry/auto-instrumentations-node": "^0.28.0",
    "@opentelemetry/exporter-trace-otlp-grpc": "^0.28.0",
    "@opentelemetry/resources": "^1.2.0",
    "@opentelemetry/semantic-conventions": "^1.0.1",
    "@opentelemetry/sdk-node": "^0.28.0",
    "@opentelemetry/sdk-trace-base": "^1.2.0",
```

14.15.4

#### Example code/test case
https://github.com/marshally/aws_sdk_sqs_message_attributes_test/

#### Text description

We have an sqs consumer process which contains a `while` loop that calls `ReceiveMessages` over and over.

```node
async function sqsConsumer() {
  while (true) {
    const resp = await sqs.receiveMessage(receiveMessageConfig).promise();

    const messages = resp.Messages || [];
    console.log(messages)

    if (messages.length) {
      received_message_ids = messages.map(({ ReceiptHandle }, k) => ({
        Id: k.toString(),
        ReceiptHandle,
      }));

      // do stuff

      const deletionPromise = sqs
            .deleteMessageBatch({
              QueueUrl: receiveMessageConfig.QueueUrl,
              Entries: received_message_ids,
            })
            .promise();

      await deletionPromise;
    }
  }
}
```

I expected my loop to run indefinitely without crashing.

### What did you see instead?

After some hours of running, the process crashes because AWS returns `413 REQUEST ENTITY TOO LARGE`

### Additional context

Digging deeper, we found that `@opentelemetry-instrumentation-aws-sdk` is automatically adding new `MessageAttributes` to each call to `ReceiveMessages`. When those `MessageAttributes` are larger than 265KB, AWS returns 413

```bash
$ node index.js
writing tracing data to NoopSpanProcessor

ReceiveMessage
HTTP BODY DEBUG LOG ------> Action=ReceiveMessage&AttributeName.1=SentTimestamp&MaxNumberOfMessages=1&MessageAttributeName.1=All&QueueUrl=https%3A%2F%2Fsqs.us-east-1.amazonaws.com%2F623766430081%2FMCY-API-SQS&Version=2012-11-05&VisibilityTimeout=20&WaitTimeSeconds=1
Tracing initialized

ReceiveMessage
HTTP BODY DEBUG LOG ------> Action=ReceiveMessage&AttributeName.1=SentTimestamp&MaxNumberOfMessages=1&MessageAttributeName.1=All&MessageAttributeName.2=traceparent&MessageAttributeName.3=tracestate&MessageAttributeName.4=baggage&QueueUrl=https%3A%2F%2Fsqs.us-east-1.amazonaws.com%2F623766430081%2FMCY-API-SQS&Version=2012-11-05&VisibilityTimeout=20&WaitTimeSeconds=1

ReceiveMessage
HTTP BODY DEBUG LOG ------> Action=ReceiveMessage&AttributeName.1=SentTimestamp&MaxNumberOfMessages=1&MessageAttributeName.1=All&MessageAttributeName.2=traceparent&MessageAttributeName.3=tracestate&MessageAttributeName.4=baggage&MessageAttributeName.5=traceparent&MessageAttributeName.6=tracestate&MessageAttributeName.7=baggage&QueueUrl=https%3A%2F%2Fsqs.us-east-1.amazonaws.com%2F623766430081%2FMCY-API-SQS&Version=2012-11-05&VisibilityTimeout=20&WaitTimeSeconds=1
```

JamieDanielson:

If duplicate requests are storing the same value multiple times, this will address the problem by passing the `MessageAttributeNames` through a Set to de-duplicate them.

It's unclear why these duplicates are happening and adding the same value multiple times, so that is still an issue to be looked into. It is not solved by this PR. This PR is also not attempting to measure the size of the attributes to determine whether the size is under the required 256KB.

- closes #1038 

- Use a set to ensure unique values when concatenating propagation fields in the AWS SQS instrumentation.

## Checklist


blumamir (commented):
Thanks for the fix
Can you please add a test for this case?

JamieDanielson:
> Thanks for the fix Can you please add a test for this case?

We're having some trouble figuring out how to test `MessageAttributeNames`, and part of that may be a misunderstanding in how it works.

Our understanding at the moment is that `MessageAttributeNames` is set on `ReceiveMessage` to indicate the specific message attribute names to receive (and thus filter other message attributes). However, we cannot find a place in the code that is actually filtering the attribute names that are omitted from this.

In `sqs.test.ts`, we tried setting the `MessageAttributeNames` on `.receiveMessage`, but still see the message attributes coming through in the response that should be filtered out. We can tell when logging out `MessageAttributeNames` that we are de-duping, but are unclear where to access this in the response so we can test it.

To help us move forward, we'd love some help answering the following questions:

1. Since we can't access `MessageAttributeNames` in the response, what would be a good test to see if `MessageAttributeNames` are being de-duped correctly?
2. Where does `MessageAttributeNames` get used to filter the message attributes?

blumamir:
Thank you for doing the research

> Our understanding at the moment is that `MessageAttributeNames` is set on `ReceiveMessage` to indicate the specific message attribute names to receive (and thus filter other message attributes).

If I understand SQS correctly, this filtering is done in the server itself, so the request contains which message attributes to include in the response and then the server copies them to "receiveMessages" response.

> However, we cannot find a place in the code that is actually filtering the attribute names that are omitted from this.

The tests for this instrumentation use a mock to return responses and validate the request parameters. This is done, for example, [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/3e2f9c56f4a748a5e63f2f054ec67d5db4b646ab/plugins/node/opentelemetry-instrumentation-aws-sdk/test/testing-utils.ts#L40). I think a proper test might be to assert that the request that is "sent" to the mock server does not repeat the propagators fields in `MessageAttributeNames`.



> In `sqs.test.ts`, we tried setting the `MessageAttributeNames` on `.receiveMessage`, but still see the message attributes coming through in the response that should be filtered out. We can tell when logging out `MessageAttributeNames` that we are de-duping, but are unclear where to access this in the response so we can test it.

I don't know how an actual SQS server is implemented to behave if multiple names are sent in the request, it could repeat them in response or just return each one of them once. The original issue if I understood it correctly is due to the request itself being invalid due to the size (These attribute names are just keep piling up forever until the limit is hit). The response in the test is mocked and will not follow the semantics of AWS.



> To help us move forward, we'd love some help answering the following questions:
> 
> 1. Since we can't access `MessageAttributeNames` in the response, what would be a good test to see if `MessageAttributeNames` are being de-duped correctly?
> 2. Where does `MessageAttributeNames` get used to filter the message attributes?

I can think of 2 possible ways to test it:
1. Add assert on the "request" as it [arrives to the mock server](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/3e2f9c56f4a748a5e63f2f054ec67d5db4b646ab/plugins/node/opentelemetry-instrumentation-aws-sdk/test/testing-utils.ts#L44). If you do that, please make sure that a failed "expect" in the callback will catch correctly by mocha. I can see it is already "expecting" in the mock, but we had bugs like this in the past so might be worth making sure.
2. Extract this new functionality into a function in `./src/services/MessageAttributes.ts`. Then you can just invoke the function directly with any interesting input and expect the result to contain each attribute name only once. There are already similar tests [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/3e2f9c56f4a748a5e63f2f054ec67d5db4b646ab/plugins/node/opentelemetry-instrumentation-aws-sdk/test/MessageAttributes.test.ts#L56). We can add a new "describe" for a function like `addPropagationMessageAttributes(origMessageAttributeNames: string[])` which will  implement this logic:
```js
return Array.from(
            new Set([
              ...(origMessageAttributeNames ?? []),
              ...propagation.fields(),
            ])
          );
```

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
```diff
@@ -67,9 +67,12 @@ export class SqsServiceExtension implements ServiceExtension {
...
 
-          request.commandInput.MessageAttributeNames = (
-            request.commandInput.MessageAttributeNames ?? []
-          ).concat(propagation.fields());
+          request.commandInput.MessageAttributeNames = Array.from(
+            new Set([
+              ...(request.commandInput.MessageAttributeNames ?? []),
+              ...propagation.fields(),
+            ])
+          );
```
blumamir, line 75: An optional improvement if you are up to it: since `request.commandInput.MessageAttributeNames` are expected to be unset most of the time, we can move this check to be tested outside of the `Array.from`, and just return  `propagation.fields()` in this case. For these common cases, we won't need to create the interminate arrays and `Set`.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
```diff
@@ -67,9 +68,13 @@ export class SqsServiceExtension implements ServiceExtension {
...
 
-          request.commandInput.MessageAttributeNames = (
-            request.commandInput.MessageAttributeNames ?? []
-          ).concat(propagation.fields());
+          const messageAttributeNames = request.commandInput.MessageAttributeNames ?
+          deduplicateMessageAttributeNames(request.commandInput.MessageAttributeNames) : [];
+
+          request.commandInput.MessageAttributeNames = [
+            ...messageAttributeNames,
+            ...propagation.fields()
```
blumamir, line 76: I think that this array that we build can now contain the `propagation` fields more than once. What do you think about calling `deduplicateMessageAttributeNames` after the arrays are merged?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
```diff
@@ -67,9 +68,13 @@ export class SqsServiceExtension implements ServiceExtension {
...
+          ];
         }
```
blumamir, line 78: Now that we have a new function in `MessageAttributes.ts` to handle this task, what do you think about moving this logic into this function as well? As this code handles message attribute (names) I would expect it to be part of this file

blumamir (commented):
Thank you for adding the test
Added 2 suggestions

marshally:
I just wanted to leave a quick note indicating that we have tested this patch and I can confirm it resolves the issue described in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1038

Thank you @JamieDanielson, we will look forward to this getting this merged and released.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
JamieDanielson: The concern we have about calling `deduplicateMessageAttributeNames` after merging the arrays is that we are then calling deduplicate every time, which it seemed like we were trying to avoid based on your earlier comment.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
JamieDanielson: There is other logic within the `sqs.ts` file that also handles MessageAttributes (e.g. line 110 in sqs.ts), so we were following the pattern set there. We're not sure that reassigning in the other file would improve the logic or readability.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sqs.ts
blumamir: Sorry, I think I wasn't clear before
What I had in mind is something like:
```js
request.commandInput.MessageAttributeNames = 
    request.commandInput.MessageAttributeNames?.length
            ? deduplicateMessageAttributeNames(
                [...request.commandInput.MessageAttributeNames, ...propagation.fields()]
              )
           : propagation.fields()
```

If I'm not mistaken, it will call `deduplicateMessageAttributeNames` only once, only if the `MessageAttributeNames` are set to something, and after merging them with `propagation.fields()` (which guarantees that the final array being sent to SQS server will not contain duplicates)

In the common case, `MessageAttributeNames` are not set and it will fallback to the `:` part and store `propagation.fields()` directly without any intermediate arrays.

This code can be prettified a bit by extracting some of the logic into function/s, but that is a matter of personal preference :) 

What do you think?

pkanal: 👋🏽 I've been working with @JamieDanielson on this fix and we're happy to make this change. The one issue we see with this is that in one case we are de-duping `propagation.fields()` and in the other case we aren't. Would it make more sense to make sure they are either consistent across both cases? So either we don't merge the arrays while de-duping or we don't have the check so it's always de-duped? This also may be a misunderstanding in case we always expect `propagation.fields()` to have unique values in which case this change makes sense! Let us know what you think 🙂

blumamir: Hi @pkanal , thank you for working on this fix.

I think the assumption is that `propagation.fields()` are expected to be distinct.
Fields can be duplicated when we merge `propagation.fields()` into the request`MessageAttributeNames` which can already contain them, so to my understanding we only need to deduplicate when merging. If you want to be extra defensive and deduplicate all the time that is also fine with me :)

But no matter what we decide, I suggest that the final array being sent to SQS server will not contain duplicates by design, which is not currently guaranteed as the arrays are merged **after** calling `deduplicateMessageAttributeNames`.

pkanal: I moved the logic out into the `deduplicateMessageAttributeNames` so that we could unit test it and added some unit tests, let me know what you think!

blumamir (approved):
LGTM, thanks for addressing all comments 💯 
Left a small optional nit about the function name

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/MessageAttributes.ts
```diff
@@ -106,3 +106,12 @@ export const extractPropagationContext = (
   }
   return undefined;
 };
+
+export const deduplicateMessageAttributeNames = (
```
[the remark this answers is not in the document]
pkanal: Makes sense to me! I updated the function name :)

blumamir (approved):

blumamir:
@JamieDanielson @pkanal , I think this is ready to be merged.
I can't merge while "This branch is out-of-date with the base branch", so please rebase this PR or ["allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so we can rebase before merging.

Thanks again for working on this fix.

PR #1046, opened 2022-06-02
allow hapi plugin to register from argument

dmathieu:

HAPI accepts loading a plugin with the `server.register(plugin)` format, in addition to the already supported `server.register({plugin: plugin})`.
See https://hapi.dev/tutorials/plugins/?lang=en_US#loading

The former currently raises exceptions because we try to load the `.plugin` key within the object, which doesn't exist.
This PR provides that support.

## Checklist


dmathieu:
The test failure seems unrelated (and also failed in main).

Rauno56 (approved):

dyladan (approved):

Issue #1030, opened 2022-05-19
improv: ioredis instrumentation are leaking data by default


vmarchaud:
By default we will join all args of the redis command, which might contains sensitive data. Also from our production usage it creates huge GC pressure for applications heavily using ioredis (which generally leads to OOM / high cpu usage)

cc @blumamir since you are currently working on bringing ioredis v4 support

vmarchaud:
> so SET foo secret command should become SET foo in db.statement?

Yeah thats the behavior of the redis instrumentation i think ? 

> Can you share the impact of the on the GC pressure you observed?

It's heavily dependent on my workload so it isn't really representative but the GC was 40% of our cpu and when changing the serializer we did go down to 15%. FYI those were bare-metal AMD Ryzen 9 5900X that clocks at 4.8Ghz  (still inside container but without cpu limit so they were not limited at all)

blumamir:
> Yeah thats the behavior of the redis instrumentation i think ?

redis instrumentation just stores the command name into the statement. no key.
I think the key is not considered a secret and is useful to be recorded into an attribute.

The straightforward approach as mentioned above is to collect the command name and first argument which is `key` for most commands but not all. For example, [`SUBSCRIBE` command](https://redis.io/commands/subscribe/) can be written as `SUBSCRIBE news music`, for which this strategy will yield `SUBSCRIBE news`.

I think we have few options:
1. record only command name by default, no key, and let user collect extra information with hook
2. record command name + first argument, and accept that it will make sense to common commands but will not look best all the time.
3. record command name + first argument and add exceptions for problematic commands in default serializer. Can be harder to maintain I suppose, but will be improved over time.

@vmarchaud - WDYT?

vmarchaud:
I would go for 2) and see how users react

PR #1052, opened 2022-06-08
feat(ioredis): only serialize non sensitive arguments in db statement attribute

aptomaKetil:

The ioredis instrumentation is currently serializing the whole redis command, including potentially sensitive (or simply large) data.

Closes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1030.

Only serialize non-sensative command argument in `db.statement` attribute.

## Checklist


plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
```diff
@@ -563,7 +561,8 @@ describe('ioredis', () => {
 
         const attributes = {
           ...DEFAULT_ATTRIBUTES,
-          [SemanticAttributes.DB_STATEMENT]: `evalsha bfbf458525d6a0b19200bfd6db3af481156b367b 1 ${testKeyName}`,
+          [SemanticAttributes.DB_STATEMENT]:
+            'evalsha bfbf458525d6a0b19200bfd6db3af481156b367b',
```
blumamir, line 565: also here - the new `db.statement` value is not correct. Might be worth mentioning in a comment to communicate that this value is not a bug and ignored intentionally for now.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +34,9 @@ export const endSpan = (
...
-) =>
-  Array.isArray(cmdArgs) && cmdArgs.length
-    ? `${cmdName} ${cmdArgs.join(' ')}`
-    : cmdName;
+) => {
+  if (Array.isArray(cmdArgs) && cmdArgs.length) {
+    return `${cmdName} ${cmdArgs[0]}`;
+  }
+  return cmdName;
+};
```
blumamir, line 42: please also update [the README](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-ioredis#custom-dbstatement-serializer)

aptomaKetil:
I went through the full list at https://redis.io/commands, and not only is there a lot of commands where we can just serialize everything, but the remainder should very often serialize more than just the first argument (as @blumamir pointed out). I've created a map of command names to number of serializable arguments so we can deal with these on a case-by-case basis.

It's not perfect, and several commands will still be missing "safe" data that could be serialized, such as [SET](https://redis.io/commands/set/), which has a ton of optional arguments that come after the value. I've added `[x other arguments]` to the end of the statement string for this purpose, so users can get some idea of the full command.

blumamir:
Awsome @aptomaKetil , that is great work, thank you so much for doing it. I like your idea of adding `[x other arguments]` for the dropped values

I will look at the exact details later.

If you want (optionally), it could be great to extract this logic into another new package maybe `@opentelemetry/instrumentation-redis-utils` under `packages` directory and then share the `defaultDbStatementSerializer` also with `redis` and `redis-4` instrumentations. They currently [implement a serializer](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/125c2f7d0c1f25f350f23d66f1c6b2d3f0bd4cf7/plugins/node/opentelemetry-instrumentation-redis/src/utils.ts#L78) that only captures the command name:
```js
const defaultDbStatementSerializer: DbStatementSerializer = cmdName => cmdName;
```
This PR is a great improvement on its own, so if you don't have the time or will to do this, let me know and I'll open a relevant issue

aptomaKetil:
Better create an issue for it, I won't have time to look at that right now.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -17,6 +17,47 @@
 import { Span, SpanStatusCode } from '@opentelemetry/api';
 import { DbStatementSerializer } from './types';
 
+/**
+ * Map of command names to the number of arguments that should be serialized.
+ * For example, SET should serialize which key it's operating on, but not its value.
+ * Commands not listed will have all their arguments serialized.
```
blumamir, line 23: Ideally, I think we would probably want to list which commands should capture all arguments, and then fall back to collecting just command name for commands not in the white list.

This will cover things like [define custom commands](https://github.com/luin/ioredis#lua-scripting) and will generally be more defensive and secure.

Since I guess writing such a list could be very long and not practical to write, I am good with keeping it as is. Since you did some research, I am curious if you have any thoughts on this

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +75,17 @@ export const endSpan = (
...
-    ? `${cmdName} ${cmdArgs.join(' ')}`
-    : cmdName;
+) => {
+  if (Array.isArray(cmdArgs) && cmdArgs.length) {
+    const argsSubset = SerializationSubsets[cmdName.toUpperCase()];
+    if (argsSubset) {
+      const args = cmdArgs.slice(0, argsSubset);
+      const remainder = cmdArgs.slice(argsSubset);
+      return `${cmdName} ${args.join(' ')} [${
+        remainder.length
```
blumamir, line 85: Should we write "[0 other arguments]" when `remainder.length` is zero?

below, when we serialize all arguments, we write just `${cmdName} ${cmdArgs.join(' ')}` without the "other arguments" part so it make sense to me to drop it here as well for consistency.

blumamir (approved):
LGTM.
Thanks again for fixing this
Left few optional nits

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +75,17 @@ export const endSpan = (
...
-) =>
-  Array.isArray(cmdArgs) && cmdArgs.length
-    ? `${cmdName} ${cmdArgs.join(' ')}`
-    : cmdName;
+) => {
+  if (Array.isArray(cmdArgs) && cmdArgs.length) {
+    const argsSubset = SerializationSubsets[cmdName.toUpperCase()];
+    if (argsSubset) {
+      const args = cmdArgs.slice(0, argsSubset);
+      const remainder = cmdArgs.slice(argsSubset);
```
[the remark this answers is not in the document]
aptomaKetil: Good point. Can do.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
aptomaKetil: Can do.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
aptomaKetil: > Ideally, I think we would probably want to list which commands should capture all arguments, and then fall back to collecting just command name for commands not in the white list.

I think there's definite value in capturing some arguments, even when we're not capturing all. Especially for SET/HSET/LPUSH and the like, we really want to register which keys were modified. For pub/sub, we want to know all the channels that were subscribed to, and so on.

> This will cover things like [define custom commands](https://github.com/luin/ioredis#lua-scripting) and will generally be more defensive and secure.

Perhaps a config option for additional commands to obfuscate would solve the same problem? One consequence of making the default serializer more opinionated is that there's more to replace if you decide to roll your own. This may have gotten to the point where it should be possible to append to the default as well as replace it entirely.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +75,17 @@ export const endSpan = (
...
+      return `${cmdName} ${args.join(' ')} [${
+        remainder.length
+      } other arguments]`;
+    }
+    return `${cmdName} ${cmdArgs.join(' ')}`;
```
vmarchaud, line 88: We must not fallback on joining every args in the result, could you change this to `cmdArgs[0]` ?

aptomaKetil:
Updated based on feedback from @blumamir and @vmarchaud.

Fallback to only serialize first argument, for added safety. Change to a whitelist approach for commands that should serialize a different subset of commands. Use regexes for subset groups to keep things short.

vmarchaud (approved):
lgtm

blumamir (commented):
@aptomaKetil @vmarchaud , What do you think about changing the default behavior to capture no arguments instead of one for all commands not in the white list? I think it will be safer

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -17,6 +17,34 @@
...
 
+/**
+ * List of regexes and the number of arguments that should be serialized for matching commands.
+ * For example, HSET should serialize which key and field it's operating on, but not its value.
+ * Setting the subset to -1 will serialize all arguments.
+ * Commands without a match will have their first argument serialized.
+ *
+ * Refer to https://redis.io/commands/ for the full list.
+ */
+const SerializationSubsets = [
```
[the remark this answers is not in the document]
aptomaKetil: Will fix.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +62,16 @@ export const endSpan = (
...
-) =>
-  Array.isArray(cmdArgs) && cmdArgs.length
-    ? `${cmdName} ${cmdArgs.join(' ')}`
-    : cmdName;
+) => {
+  if (Array.isArray(cmdArgs) && cmdArgs.length) {
+    const argsSubset =
+      SerializationSubsets.find(({ regex }) => {
+        return regex.test(cmdName);
+      })?.args || 1;
```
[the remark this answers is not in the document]
aptomaKetil: Good point. Will fix.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +62,16 @@ export const endSpan = (
...
+    const args = argsSubset > 0 ? cmdArgs.slice(0, argsSubset) : cmdArgs;
```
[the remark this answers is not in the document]
aptomaKetil: Will fix.

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -34,7 +62,16 @@ export const endSpan = (
...
+    return `${cmdName} ${args.join(' ')} [${
+      args.length !== cmdArgs.length ? cmdArgs.length - argsSubset : 0
```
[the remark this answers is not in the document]
aptomaKetil: Will fix.

plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
```diff
@@ -187,19 +187,22 @@ describe('ioredis', () => {
...
         args: [hashKeyName, 'testField', 'testValue'],
+        serializedArgs: [hashKeyName, 'testField', '[1 other arguments]'],
         method: (cb: ioredisTypes.CallbackFunction<number>) =>
           client.hset(hashKeyName, 'testField', 'testValue', cb),
       },
       {
         description: 'get',
         name: 'get',
         args: [testKeyName],
+        serializedArgs: [testKeyName, '[0 other arguments]'],
```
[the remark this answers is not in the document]
aptomaKetil: I've gone back and forth on it myself, but I think maybe I'd prefer dropping it if it's not needed. It does end up being kinda noisy.

vmarchaud:
> What do you think about changing the default behavior to capture no arguments instead of one for all commands not in the white list? I think it will be safer

I don't think there is problem adding the command name since its already on the span name right ?

plugins/node/opentelemetry-instrumentation-ioredis/src/utils.ts
```diff
@@ -17,6 +17,34 @@
...
+ *
+ * Refer to https://redis.io/commands/ for the full list.
+ */
+const serializationSubsets = [
+  {
+    regex: /^ECHO/i,
+    args: 0,
+  },
+  {
+    regex: /^(LPUSH|MSET|SET|PFA|PUBLISH|RPUSH|SADD|SET|SPUBLISH|XADD|ZADD)/i,
```
blumamir, line 34: nit: `SET` is written twice :)

plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
```diff
@@ -951,4 +955,36 @@ describe('ioredis', () => {
       });
     });
   });
+
+  describe('#defaultDbStatementSerializer()', () => {
```
blumamir, line 959: Thanks for adding a dedicated test 😍

blumamir (approved):

blumamir:
> Updated to serialize no arguments by default.

Amazing, thank you for addressing everything. This is great work 🥇 
I'll keep this open for another day and merge it tomorrow if there are no new comments.

PR #1055, opened 2022-06-11
feat(propagation-utils): end pub-sub process span on promise settled

nordfjord:
When promises are returned from a mapping function we now
end the span when that promises resolves rather than doing
so immediately.

This makes sure that the processing span wraps the entirety of the
processing

- The pubsub propagation module would end spans prematurely when promises were involved.

An example of what this looks like:
Screen Shot 2022-06-11 at 2 17 00 PM

- Handle promises from mapping functions

## Checklist


blumamir:
@nordfjord thank you for the contribution :)
can you please run `npm run lint:fix` to revert all the linter changes? It will really help to focus on the logical changes while reviewing and make CI happy.

Also please add relevant unit tests for this change. Either in the `propagation-utils` package or in the `aws-sdk` instrumentation package which is consuming these utils. Thank you!

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
```diff
@@ -22,13 +22,13 @@ import {
   context,
   trace,
   diag,
-  SpanAttributes,
-} from '@opentelemetry/api';
+  Attributes,
```
blumamir, line 25: `Attributes` was added to `@opentelemetry/api` in version `1.1.0` [here](https://github.com/open-telemetry/opentelemetry-js-api/releases/tag/v1.1.0).
The current dependency is on `^1.0.0` which is not compatible. We should either update the dependency or revert to use `SpanAttributes`

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
```diff
@@ -50,15 +50,26 @@ const patchArrayFilter = (
...
 
-  Object.defineProperty(messages, 'filter', {
+  Object.defineProperty(messages, "filter", {
     enumerable: false,
     value: patchedFunc,
   });
 };
 
+function isPromise(value: unknown): value is Promise<unknown> {
+  return value != null && typeof (value as any).then == "function";
```
blumamir, line 60: We prefer to use `===` instead of `==`
optional nit: I think it can be shortened to `typeof (value as any)?.then === 'function'`.

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
```diff
@@ -50,15 +50,26 @@ const patchArrayFilter = (
...
   };
 
-  Object.defineProperty(messages, 'filter', {
+  Object.defineProperty(messages, "filter", {
     enumerable: false,
     value: patchedFunc,
   });
 };
 
+function isPromise(value: unknown): value is Promise<unknown> {
```
blumamir, line 59: Shouldn't the return type be `boolean` instead of `value is Promise<unknown>`

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
```diff
@@ -50,15 +50,26 @@ const patchArrayFilter = (
...
+  return value != null && typeof (value as any).then == "function";
+}
+
+function Promise_finally<T>(p: Promise<T>, onFinally: () => any): Promise<T> {
```
blumamir, line 63: Our convention for function names is camelCase: `promiseFinally`.
Also, could be nice to align the style with the rest of the functions in this package, e.g: 
```js
const promiseFinally<T> = (p: Promise<T>, onFinally: () => any): Promise<T> => {...}
```

blumamir: Is this function being called anywhere?

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
```diff
@@ -77,14 +88,30 @@ const patchArrayFunction = (
...
+                message[END_SPAN_FUNCTION]?.();
+                return x;
+              },
+              (err) => {
+                message[END_SPAN_FUNCTION]?.();
+                throw err;
+              }
+            );
+          }
+          return result;
```
blumamir, line 106: Here we are creating a new promise and returning it instead of the original one.
Since we are only interested in capturing the promise `resolve` / `rejected` event and do not care about modifying the result itself, I think this can be simplified to just hang another callback on the result promise:
```js
    result.then( () => message[END_SPAN_FUNCTION]?.(), () => message[END_SPAN_FUNCTION]?.())
}
return result;
```

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
nordfjord: it depends. the `value is` lets TypeScript know that if this predicate passes it can treat the value as a Promise.

```ts
if (isPromise(value)) {
  // TypeScript will know that inside this if block the value can be considered a promise for typechecking purposes
  value.then(x => ...) 
}
```

If `boolean` is used as a return value TypeScript will not narrow the type definition.

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
nordfjord: Good point. I've changed this.

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
nordfjord: It's not. I've removed it

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
nordfjord: 🤦 , sorry about that. I changed this because my tooling was saying SpanAttributes were deprecated.

nordfjord:
Thanks for taking the time to review this. I think this is ready for another review now.

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
Flarna: I think this might result in unhandled rejections now.
Calling `.then()` returns a `Promise` which is not used at all here. The catch handler throws which results in rejecting this lost promise. In the original code this chained promise was returned and the user catch handler will be called (if present).

try the following with node 14 or newer:
```js
function foo() {
  const p = Promise.reject("catch foo")
  p.then(() => {}, (e) => { throw e })
  return p
}

function bar() {
  const p = Promise.reject("catch bar")
  return p.then(() => {}, (e) => { throw e })
}

(async () => {
  try { await foo() } catch (e) {}
  try { await bar() } catch (e) {}
})()
```

packages/opentelemetry-propagation-utils/src/pubsub-propagation.ts
blumamir: Cool! I learned something new :)

blumamir (approved):
LGTM.
After tests are green this can be merged AFAIC

Thank you for contributing this fix

packages/opentelemetry-propagation-utils/package.json
```diff
@@ -43,8 +45,16 @@
...
+    "@opentelemetry/contrib-test-utils": "^0.31.0",
+    "@types/mocha": "^9.1.1",
     "@types/node": "16.11.21",
+    "@types/sinon": "^10.0.11",
+    "expect": "^28.1.1",
     "gts": "3.1.0",
+    "mocha": "^10.0.0",
+    "nyc": "^15.1.0",
+    "sinon": "^14.0.0",
+    "ts-mocha": "^10.0.0",
```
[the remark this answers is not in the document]
nordfjord: Alrighty, copied some versions over! 👍

nordfjord:
Not quite sure what was going on but I had to remove the `--require @opentelemetry/contrib-test-utils` part from the package.json test script so that the tests would pass.

I've verified that `yarn test` works locally now.

Rauno56 (approved):

Rauno56:
`catch {}` is not supported in node 8. Will merge this after https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1065.

vmarchaud (approved):

PR #1056, opened 2022-06-13
feat(lru-memoizer): add instrumentation for lru-memoizer

blumamir:

- Add instrumentation for the [lru-memoizer](https://www.npmjs.com/package/lru-memoizer).

This package is internally storing callbacks in a queue and then invokes them with a different context from which they were queued. This behavior causes spans to appear in the wrong trace.

I did not generate any spans for this instrumentation (although it can be done as a future enhancement). This PR is just about fixing the context which is currently incorrect because callbacks are used in a manner that is not supported by ContextManager and it creates some really funky traces.

The package itself seems to be unmaintained, and also includes some typescript issues which made the instrumentation testing fail to compile. I had to turn on `skipLibCheck` and to make some `any` casts to make it work. I welcome any tips on how to resolve these issues cleanly


blumamir:
I have few typescript issues to fix and then it will be ready for review

plugins/node/instrumentation-lru-memoizer/README.md
```diff
@@ -0,0 +1,57 @@
...
+npm install --save @opentelemetry/instrumentation-lru-memoizer
+```
+
+## Supported Versions
+
+- `>1.2.0 <3`
+
+## Usage
+
+This instrumentation does not produce any telemetry data. It only bind the caller context to callbacks so downstream operations are recorded with the right context (traceId / parentSpanId / baggadge / etc). This package is a dependency for other popular packages such as [jwks-rsa](https://www.npmjs.com/package/jwks-rsa)
```
habmic, line 24: Typo in baggadge -> baggage

habmic (approved):
LGTM

plugins/node/instrumentation-lru-memoizer/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
Rauno56, line 189:
```suggestion
   Copyright [2022] OpenTelemetry Authors
```

plugins/node/instrumentation-lru-memoizer/src/instrumentation.ts
```diff
@@ -0,0 +1,65 @@
...
+              // last argument is the callback
+              const origCallback = modifiedArguments.pop();
+              modifiedArguments.push(
+                context.bind(context.active(), origCallback)
+              );
+              return origMemoizer.apply(this, modifiedArguments);
+            };
+          };
+
+          // sync function preserve context, but we still need to export it
```
Rauno56, line 56:
```suggestion
          // sync function preserves context, but we still need to export it
```

plugins/node/instrumentation-lru-memoizer/src/instrumentation.ts
```diff
@@ -0,0 +1,65 @@
...
+    return [
+      new InstrumentationNodeModuleDefinition<any>(
+        'lru-memoizer',
+        ['>1.2 <=2'],
+        moduleExports => {
+          this._diag.debug('applying patch for lru-memoizer');
+
+          // moduleExports is a function which receives an options object,
+          // and return a "memoizer" function upon invocation
+          // we want to patch this "memoizer" internal function
```
Rauno56, line 40:
```suggestion
          // moduleExports is a function which receives an options object,
          // and returns a "memoizer" function.
          // We want to patch this "memoizer's" internal function
```

plugins/node/instrumentation-lru-memoizer/src/instrumentation.ts
```diff
@@ -0,0 +1,65 @@
...
+          const asyncMemoizer = function (this: unknown) {
+            // this function is invoked every time the user want to read something from the cache
+            // we replace it with another function which bind the current context to the last argument (callback)
+            const origMemoizer = moduleExports.apply(this, arguments);
+            return function (this: unknown) {
+              const modifiedArguments = [...arguments];
+              // last argument is the callback
+              const origCallback = modifiedArguments.pop();
```
Rauno56, line 48: ~~What if the user just does not pass a callback or passes something invalid like a number or an object? I understand it's a silly usecase, but we should then still just error in the way the underlying package fails in.~~

I guess `context.bind` is a pass-through it the callback is not callable?

plugins/node/instrumentation-lru-memoizer/src/instrumentation.ts
```diff
@@ -0,0 +1,65 @@
...
+        ['>1.2 <=2'],
+        moduleExports => {
+          this._diag.debug('applying patch for lru-memoizer');
+
+          // moduleExports is a function which receives an options object,
+          // and return a "memoizer" function upon invocation
+          // we want to patch this "memoizer" internal function
+          const asyncMemoizer = function (this: unknown) {
+            // this function is invoked every time the user want to read something from the cache
+            // we replace it with another function which bind the current context to the last argument (callback)
```
Rauno56, line 43:
```suggestion
            // This function is invoked every time the user wants to read something from the cache
            // We replace it with another function in which we bind the current context to the last argument (callback)
```

Rauno56:
Few nits and questions, nothing major. :+1:

blumamir:
> Few nits and questions, nothing major. 👍

Thank you 🙏
Addressed all comments

plugins/node/instrumentation-lru-memoizer/src/instrumentation.ts
Rauno56: Aah... right. Well in that case we should test for a function(`typeof origCallback === 'function'`) in the instrumentation itself before binding the context to avoid breaking the user and short-circuit if it doesn't. Just in case.

blumamir: Added 👍

Rauno56:
Perfect! :green_circle: from my end! Thanks!

Issue #978, opened 2022-04-22
ioredis instrumentation: support v5


blumamir:
ioredis instrumentation [currently support versions '>1 <5'](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/cb915cfe018678060b014e1385fcefd0dc73c776/plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts#L52).

[A new version 5 was published at 26/03/2022](https://www.npmjs.com/package/ioredis/v/5.0.0)

This issue is about adding support to the new version.

Renovate tests fails to compile with new version in #957 .
Some examples of the CI output:

```
src/instrumentation.ts:55:39 - error TS2339: Property 'prototype' does not exist on type 'typeof import("/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/ioredis/built/index")'.

55           if (isWrapped(moduleExports.prototype.sendCommand)) {
                                         ~~~~~~~~~

src/instrumentation.ts:56:40 - error TS2339: Property 'prototype' does not exist on type 'typeof import("/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/ioredis/built/index")'.

56             this._unwrap(moduleExports.prototype, 'sendCommand');
                                          ~~~~~~~~~

src/instrumentation.ts:59:27 - error TS2339: Property 'prototype' does not exist on type 'typeof import("/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/ioredis/built/index")'.

59             moduleExports.prototype,
                             ~~~~~~~~~

src/instrumentation.ts:63:39 - error TS2339: Property 'prototype' does not exist on type 'typeof import("/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/ioredis/built/index")'.

63           if (isWrapped(moduleExports.prototype.connect)) {
```

And many more

nazarkulyk:
Can someone port instrument to get it work with ioredis v5 ?

PR #1062, opened 2022-06-18
Update ioredis library from 4.27.7 to 5.0.6

raushi:
- It is not working with ioredis 5.0+ version. It will fix that

- Update ioredis version from 4.27.7 to 5.0.6 

## Checklist


Rauno56:
At minimum we should also add test-all-versions configuration to also test 5.x range.
The file I'm talking about is `.tav.yml`. For now, just addig `^5 || ...` to the range is probably enough.

dyladan:
If the instrumentation doesn't work with 5.x then simply updating the devdependency is unlikely to fix the real issue

Rauno56:
Right. @raushi , would you like to work on adding the support for v5 of `ioredis`?

raushi:
> Right. @raushi , would you like to work on adding the support for v5 of `ioredis`?

Yes.

blumamir:
To add support for `ioredis` for version 5, you should begin with updating the supportedVersions [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e2657232c9007a77ddc873a93e8247f99087b9c1/plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts#L52).

Instrumentation code will probably also need to update to support the new moduleExports. Relevant issue: #978

PR #1066, opened 2022-06-22
fix: avoid type imports of the aws-sdk package in the built assets

Rauno56:

This is a subset of #1017 that touches `aws-sdk` types because of the legal discussion holding merging those back.

See #1017

## Checklist


blumamir (approved):

dyladan:
Just wanted to let you know there is a service desk ticket for this CNCFSD-1369

dyladan:
https://github.com/cncf/foundation/blob/main/copyright-notices.md#what-about-third-party-code

looks like if we don't modify their types we would just preserve their copyright notice and not add our own

Rauno56:
Added 3rd party copyright notice because we do change them slightly.

PR #1068, opened 2022-06-22
chore: lint PR titles

dyladan:
Some PR authors (especially new ones) often don't know that PR titles should be conventional commit messages. This adds a lint check to both let those authors know and also to ensure we dont accidentally merge a PR before it is properly named.


dyladan:
Looks like this when it fails:

image

pichlermarc (approved):

PR #1072, opened 2022-06-30
feat(express): add the ability to set request bodies as root a span attribute

luismiramirez:
Using the new `requestBodyAsAttribute` config option, the automatic
instrumentation will stringify the request bodies and add them as an
attribute of the HTTP root span.

It uses the helpers from RPC metadata to easily access and set
the attribute to the HTTP root span.

With this new config option, you can add a request body as HTTP root span attribute. Having this kind of optional insight is useful for request tracing in APIs and web applications.

- Used the RPC metadata helpers to access the HTTP root span to add the body attributes.

- If the proposal is accepted I'd like to add similar functionality for query params with a new config option and a new attribute.

## Checklist


dyladan (changes requested):
I do not think I would approve this feature. Adding a configuration option implies that we condone the behavior it enables. I think adding request/response bodies to span attributes is a mistake for the following reasons:

1. It is extremely likely to contain PII in many cases
2. The telemetry will become massive
3. A result of (2) is likely to be incurring additional transmission/storage/processing costs
4. This is likely to increase the performance footprint of the express instrumentation significantly on the client from both a memory use perspective and the cost of sending telemetry to your backend.

I would suggest if you think you need this you think very hard about what you actually are trying to solve. There is very likely to be a significantly less impactful option you can take. If you do decide that you need it, I suggest you create an express middleware which sets the span attribute for you.

luismiramirez:
Thanks @vmarchaud and @dyladan for your reviews and for sharing your concerns.

I understand your reasoning and it makes a lot of sense. I'm going to close this pull request and open a new one with what @vmarchaud proposed, adding a hook as the one in the [ioredis](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-ioredis#ioredis-instrumentation-options) instrumentation.

PR #1084, opened 2022-07-08
feat: added instana resource detector

kirrg001:
This adds a vendor specific resource detector. It solves the problem of process and trace correlation.

Added Instana Resource detector and tests,

## Checklist
- [x] Ran npm run test for the new package on the latest commit if applicable.
- [x] discuss usage of service.instance.id


dyladan (approved):
LGTM % readme nit

dyladan (changes requested):
Please also add an entry to the release please manifest

release-please-config.json
```diff
@@ -1,8 +1,6 @@
 {
   "bootstrap-sha": "798fb2d5c585f6fdee10146bfe15e03409bdddfe",
```
vmarchaud, line 2: nit but i think you can rollback changes to this file since its just formating

vmarchaud (approved):

release-please-config.json
vmarchaud: Sorry  i was pointing to the format change in the file, not the L12 that you added. You might auto format file on save with your IDE ?

PS: There is no command to modify release-please configs so you are correct to modify them by hand

kirrg001: Ah sorry thanks. I will correct

kirrg001: Fixed

detectors/node/opentelemetry-resource-detector-instana/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[`gcp-metadata`]: https://www.npmjs.com/package/gcp-metadata
+[GCP Metadata Documentation]: https://cloud.google.com/compute/docs/metadata/overview
+[discussions-url]: https://github.com/open-telemetry/opentelemetry-js/discussions
+[license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/LICENSE
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[npm-url]: https://www.npmjs.com/package/@opentelemetry/resource-detector-gcp
+[npm-img]: https://badge.fury.io/js/%40opentelemetry%2Fresource-detector-gcp.svg
```
blumamir, line 66: links point to `resource-detector-gcp`

detectors/node/opentelemetry-resource-detector-instana/README.md
```diff
@@ -0,0 +1,66 @@
...
+- For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
+- For more about OpenTelemetry JavaScript: <https://github.com/open-telemetry/opentelemetry-js>
+- For help or feedback on this project, join us in [GitHub Discussions][discussions-url]
+
+## License
+
+Apache 2.0 - See [LICENSE][license-url] for more information.
+
+[`gcp-metadata`]: https://www.npmjs.com/package/gcp-metadata
+[GCP Metadata Documentation]: https://cloud.google.com/compute/docs/metadata/overview
```
blumamir, line 61: should be removed

detectors/node/opentelemetry-resource-detector-instana/README.md
```diff
@@ -0,0 +1,66 @@
...
+const sdk = new NodeSDK({
+   autoDetectResources: false,
+   resource: globalResource,
+});
+
+await sdk.detectResources({
+   detectors: [envDetector, processDetector, instanaAgentDetector],
+});
+
+// console.log(sdk['_resource'])
```
blumamir, line 46: I don't think we should suggest users to access the resource like this

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+      process.env.INSTANA_AGENT_PORT || this.INSTANA_AGENT_DEFAULT_PORT
+    );
+
+    const data = await this._retryHandler(host, port, 0);
+
+    return new Resource({
+      [SemanticResourceAttributes.PROCESS_PID]: data.pid,
+      'instana.host.id': data.agentUuid,
+      // NOTE: The agent reads this property to correlate the process with the service
+      'service.instance.id': data.pid.toString(),
```
blumamir, line 42: Please use [SemanticResourceAttributes](https://github.com/open-telemetry/opentelemetry-js/blob/bec6c3c085290fb254c5a27df294fad2a830af52/packages/opentelemetry-semantic-conventions/src/resource/SemanticResourceAttributes.ts#L436)

blumamir, line 42: using the `pid` is not globally unique. [From the spec](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service):

> MUST be unique for each instance of the same service.namespace,service.name pair (in other words service.namespace,service.name,service.instance.id triplet MUST be globally unique). The ID helps to distinguish instances of the same service that exist at the same time (e.g. instances of a horizontally scaled service)

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+    port: number,
+    tries: number
+  ): Promise<any> {
+    const MAX_TRIES = 3;
+
+    try {
+      return await this._fetchAgentData(host, port);
+    } catch (err) {
+      if (tries < MAX_TRIES) {
+        console.log(
```
blumamir, line 61: please don't write to `console` directly. use the [diag api](https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/api/diag.ts)

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+    return new Resource({
+      [SemanticResourceAttributes.PROCESS_PID]: data.pid,
+      'instana.host.id': data.agentUuid,
+      // NOTE: The agent reads this property to correlate the process with the service
+      'service.instance.id': data.pid.toString(),
+    });
+  }
+
+  private timeout(ms: number) {
+    return new Promise(resolve => setTimeout(resolve, ms));
```
blumamir, line 47: Can this timeout block the process from proper shutdown?
In other places we usually [`unref`](https://nodejs.org/api/timers.html#timeoutunref) so opentelemetry will  not hold the event loop active

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    try {
+      return await this._fetchAgentData(host, port);
+    } catch (err) {
+      if (tries < MAX_TRIES) {
+        console.log(
+          `Retrying to connect to the Instana agent on ${host}:${port}....`
+        );
+
+        const retryTimeout = process.env.INSTANA_RETRY_TIMEOUT
```
blumamir, line 65: I don't know if the env variable is legacy or not, but could be nice to add the unit so it is not a mystery. e.g: `INSTANA_RETRY_TIMEOUT_MS`

detectors/node/opentelemetry-resource-detector-instana/README.md
```diff
@@ -0,0 +1,66 @@
...
+```bash
+npm install --save @opentelemetry/resource-detector-instana
+```
+
+## Environment variables
+
+- INSTANA_AGENT_HOST: The Instana agent hostname.
+- INSTANA_AGENT_PORT: The Instana agent port.
+- INSTANA_RETRY_TIMEOUT: The resource detector does three retries to connect to the Instana agent. This is the timeout between the retries.
+- INSTANA_AGENT_TIMEOUT: The client timeout when connecting the Instana agent.
```
blumamir, line 19: WDYT about adding the default values and specifying the value is in ms?

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const payloadStr = JSON.stringify(payload);
+    const contentLength = Buffer.from(payloadStr, 'utf8').length;
+
+    return new Promise((resolve, reject) => {
+      // agent needs more than 1s!
+      const timeoutId = setTimeout(() => {
+        req.destroy();
+        reject(new Error('Instana Agent request timed out.'));
+      }, agentTimeout);
```
blumamir, line 94: can we use the `timeout` option on the `http.request` call instead of implementing it?
https://nodejs.org/api/http.html#httprequestoptions-callback

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+        console.log(
+          `Retrying to connect to the Instana agent on ${host}:${port}....`
+        );
+
+        const retryTimeout = process.env.INSTANA_RETRY_TIMEOUT
+          ? Number(process.env.INSTANA_RETRY_TIMEOUT)
+          : 1000;
+
+        await this.timeout(retryTimeout);
+        return await this._retryHandler(host, port, tries + 1);
```
blumamir, line 70: This will retry at most `MAX_TRIES` so we are probably not expected to run into stack overflow, but I wonder what is the gain from using recursion over a simple loop? 🤔

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+
+      throw err;
+    }
+  }
+
+  private async _fetchAgentData(host: string, port: number): Promise<any> {
```
blumamir, line 77: Can we type the return value from the function, e.g:
```js
Promise<{pid: number, agentUuid: string}>
```
?

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+              res.statusCode < 300
+            ) {
+              try {
+                resolve(JSON.parse(rawData));
+              } catch (e) {
+                reject(e);
+              }
+            } else {
+              return reject(
+                new Error('Failed to load page, status code: ' + res.statusCode)
```
blumamir, line 127: What do you mean by "load page"? I guess the error message can be more informative

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,145 @@
...
+              res.statusCode >= 200 &&
+              res.statusCode < 300
+            ) {
+              try {
+                resolve(JSON.parse(rawData));
+              } catch (e) {
+                reject(e);
+              }
+            } else {
+              return reject(
```
blumamir, line 126: no need for the `return` here

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
kirrg001: This error message is used in all detectors. So I kept the pattern.

detectors/node/opentelemetry-resource-detector-instana/README.md
kirrg001: Its only a hint for debugging purposes. Can be removed, I dont mind

detectors/node/opentelemetry-resource-detector-instana/README.md
```diff
@@ -0,0 +1,66 @@
...
+const globalResource = new Resource({
+   [SemanticResourceAttributes.SERVICE_NAME]: "TestService",
+});
+
+const sdk = new NodeSDK({
+   autoDetectResources: false,
+   resource: globalResource,
+});
+
+await sdk.detectResources({
```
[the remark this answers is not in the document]
kirrg001: Yes, because this is not a full example

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
kirrg001: Sure can also be written in a loop. Just used to recursive fn's.

detectors/node/opentelemetry-resource-detector-instana/README.md
kirrg001: Removed

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
kirrg001: We are discussing the usage of the property currently internally. But it looks like we won't use it at all.

kirrg001: We are discussing the usage of the property currently internally. But it looks like we won't use it at all.

kirrg001: Changed the value of service.instance.id. Used a uuid.

basti1302 (approved):
LGTM

detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -0,0 +1,144 @@
...
+            return reject(
+              new Error(`Instana Agent returned status code ${res.statusCode}`)
+            );
+          }
+
+          try {
+            const data = JSON.parse(rawData);
+
+            if (data.pid && data.agentUuid) {
+              return resolve(JSON.parse(rawData));
```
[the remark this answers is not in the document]
basti1302: fixed

dyladan (approved):

PR #1003, opened 2022-05-04
fix: move @opentelemetry/core to dependencies

Flarna:
## Which problem is this PR solving?

https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1000

## Short description of the changes

Move @opentelemetry/core to dependencies

## Checklist

- [ ] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.

PR #1085, opened 2022-07-09
chore: introduce no-extraneous-dependencies

shikaan:

This should prevent https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1003 from happening again.

The glob is used to specify locations where the rule should not apply, we should there be safe in importing devDependencies in tests.

## Question

It _seems_[^1] to have caught some problems already, though I am not sure if we want to also address the problems in this PR.

[^1]: The error could actually be intended behaviour, as we seem to need only the typings [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/7510757aeeee47a7f0c4bb31de45be3a71bb673e/plugins/node/opentelemetry-instrumentation-express/src/types.ts#L18) for example. I am not confident with the codebase enough to add an exception right off the bat, a couple of more expert eyes on this would be appreciated.

## Checklist

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,7 +1,8 @@
 module.exports = {
   plugins: [
     "@typescript-eslint",
-    "header"
+    "header",
+    "import"
   ],
   extends: [
       "./node_modules/gts",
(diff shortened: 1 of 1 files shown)


shikaan:
Looks like I have to work out legal details with my employer. Feel free to copy paste this code if you want to, closing this PR.

PR #1086, opened 2022-07-10
ci: introduce no extraneous dependencies lint check

dyladan:
Taken with permission from @shikaan in #1085

Original PR description:

> ## Which problem is this PR solving?
> 
> This should prevent https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1003 from happening again.
> 
> ## Short description of the changes
> 
> The glob is used to specify locations where the rule should not apply, we should there be safe in importing devDependencies in tests.
> 
> ## Question
> 
> It _seems_[^1] to have caught some problems already, though I am not sure if we want to also address the problems in this PR.
> 
> [^1]: The error could actually be intended behaviour, as we seem to need only the typings [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/7510757aeeee47a7f0c4bb31de45be3a71bb673e/plugins/node/opentelemetry-instrumentation-express/src/types.ts#L18) for example. I am not confident with the codebase enough to add an exception right off the bat, a couple of more expert eyes on this would be appreciated.
> 
> ## Checklist
> 
> - [x] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.


blumamir (approved):

vmarchaud (approved):

dyladan:
The lint is failing because express is used in types. The lint check wants me to add express to the main deps, but I don't really want to bloat the install more than it already is so vendoring the types is the way to go here i think

legendecas:
> The lint is failing because express is used in types. The lint check wants me to add express to the main deps, but I don't really want to bloat the install more than it already is so vendoring the types is the way to go here i think

Can the lint error be suppressed by updating the import statement to be `import type` statement, as the type `Request` is only used in type declarations? TypeScript stripes `import type` statements in the JS outputs so I suppose eslint will not complain about it anymore.

blumamir:
Thanks @dyladan ,

I think this can merge, right?
Can you please rebase? for some reason I am not able to rebase it myself

vmarchaud:
> Can you please rebase? for some reason I am not able to rebase it myself

We can't rebase when org doesn't allow edit, i believe thats the case of dynatrace-oss one

pichlermarc (approved):

vmarchaud:
@dyladan you will need to rebase to merge this one

PR #1088, opened 2022-07-13
fix(example): fix koa example

umuoy1:

- koa example wasn't to be able to run correctly.

- fixes the status code for spans
- sets the serviceName for the example


examples/koa/client.js
```diff
@@ -13,10 +13,10 @@ function makeRequest() {
   api.context.with(api.trace.setSpan(api.ROOT_CONTEXT, span), async () => {
     try {
       const res = await axios.get('http://localhost:8081/run_test');
-      span.setStatus({ code: api.StatusCode.OK });
+      span.setStatus({ code: api.SpanStatusCode.OK });
```
blumamir, line 16: mmm, I think that we should actually delete this line.
span status code should be `UNSET` which is the default.

[from the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status):
> Generally, Instrumentation Libraries SHOULD NOT set the status code to Ok, unless explicitly configured to do so. Instrumentation Libraries SHOULD leave the status code as Unset unless there is an error, as described above.

examples/koa/tracer.js
```diff
@@ -9,17 +9,23 @@ const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
 const { SimpleSpanProcessor } = require('@opentelemetry/sdk-trace-base');
 const { JaegerExporter } = require('@opentelemetry/exporter-jaeger');
 const { ZipkinExporter } = require('@opentelemetry/exporter-zipkin');
+const { Resource } = require("@opentelemetry/resources")
+const { SemanticResourceAttributes } = require("@opentelemetry/semantic-conventions")
```
blumamir, line 13: The `require`s above use a single quote, can be nice to follow the existing style: `require('@opentelemetry/resources')`

blumamir (approved):

examples/koa/client.js
legendecas: Isn't this example an end-user example?

legendecas (approved):
Thank you for your contribution! LGTM % the style nits mentioned by @blumamir.

examples/koa/tracer.js
umuoy1: done

examples/koa/client.js
legendecas: @blumamir would you mind shedding your light on this again? I believe this is an end-user example so setting the status code here should be fine.

dyladan: @legendecas is correct the end user should set the span status and this example is fine IMO

dyladan (approved):

PR #1091, opened 2022-07-18
feat(express): add requestHook support

luismiramirez:
The `requestHook` support allows custom span handling per request.

- Adds `requestHook` support so users can add custom handling on spans per request.

- It's inspired by how `requestHook` is implemented in other instrumentations
- The `requestHook` user-defined function is called when instrumenting the Express request handler layer if the function exists.

## Checklist


plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -97,4 +107,7 @@ export interface ExpressInstrumentationConfig extends InstrumentationConfig {
   /** Ignore specific layers based on their type */
   ignoreLayersType?: ExpressLayerType[];
   spanNameHook?: SpanNameHook;
+
+  /** Function for adding custom attributes on Express request */
+  requestHook?: ExpressRequestCustomAttributeFunction;
```
vmarchaud, line 112: could you document it on the readme ?

luismiramirez: Doing it right now; somehow I didn't commit the file. Thanks for noticing!

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -244,6 +245,23 @@ export class ExpressInstrumentation extends InstrumentationBase<
         const span = instrumentation.tracer.startSpan(spanName, {
           attributes: Object.assign(attributes, metadata.attributes),
         });
+
+        if (
+          instrumentation.getConfig().requestHook &&
+          metadata.attributes[AttributeNames.EXPRESS_TYPE] ===
```
vmarchaud, line 251: i think we can generalize the hook to all layer type WDYT ? we just need to be sure we export the `ExpressLayerType` so that end user can use the correct name

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file. Do not remo
   * [#622](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/622) docs: add missing fetch instrumentation ([@meteorlxy](https://github.com/meteorlxy))
   * [#553](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/553) feat: Add NestJS instrumentation ([@Rauno56](https://github.com/Rauno56))
   * [#588](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/588) chore: adding instrumentation for connect ([@obecny](https://github.com/obecny))
+  * [#1091](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1091) feat(express): add Express requestHook support ([@luismiramirez](https://github.com/luismiramirez))
```
blumamir, line 22: I think for contrib repo you don't need to update the CHANGELOG.md file. it is generated automatically upon publish.

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -78,6 +79,37 @@ Express instrumentation has few options available to choose from. You can set th
 - `info: ExpressRequestInfo` containing the incoming Express.js request, the current route handler creating a span and `ExpressLayerType` - the type of the handling layer or undefined when renaming the root HTTP instrumentation span.
 - `defaultName: string` - original name proposed by the instrumentation.
 
+#### Using `requestHook`
+
+Instrumentation configuration accepts a custom "hook" function which will be called on every request with the request handler span and the request information. Custom attributes can be set on the span or run any custom logic per request.
```
blumamir, line 84: > will be called on every request with the request handler span and the request information

Each request can generate multiple spans, right? It sounds from the text like this hook will be called once per request, but it will actually be called per middleware?
I suggest making it more clear.

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -88,6 +88,20 @@ export type SpanNameHook = (
...
+ * Function that can be used to add custom attributes to the current span or the root span on
+ * a Express request
+ * @param requestSpan - The Express request handler span
+ * @param expressRequest: - The Express request object
+ */
+export interface ExpressRequestCustomAttributeFunction {
+  (
+    span: Span,
+    layerType: ExpressLayerType,
+    expressRequest: PatchedRequest
```
blumamir, line 101: please store these properties in an info object, so that if more data is added in the future, we will not end up with huge function signatures.

You can see an example in this file for the spanNameHook which stores all the relevant data in `ExpressRequestInfo` (or even reuse it?)

blumamir, line 101: Any reason to use `PatchedRequest` instead of `Request`?

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -78,6 +79,37 @@ Express instrumentation has few options available to choose from. You can set th
...
+const expressInstrumentation = new ExpressInstrumentation({
+  requestHook: function (
+    span: Span,
+    layerType: ExpressLayerType,
+    expressRequest: PatchedRequest
+  ) {
+
+    if (layerType === ExpressLayerType.REQUEST_HANDLER) {
+      currentSpan.setAttribute(
+        'request method',
```
blumamir, line 100: even though attribute names can contain spaces, I suggest using `http.method` and `express.base_url` to make it consistent with semantic conventions style.

formal examples and documentation should promote proper usage of attribute names and semantic conventions where possible.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -244,6 +245,19 @@ export class ExpressInstrumentation extends InstrumentationBase<
         const span = instrumentation.tracer.startSpan(spanName, {
           attributes: Object.assign(attributes, metadata.attributes),
         });
+
+        if (instrumentation.getConfig().requestHook) {
+          safeExecuteInTheMiddle(
+            () => instrumentation.getConfig()?.requestHook!(span, type, req),
```
blumamir, line 251: you have optional chaining after the call to `getConfig()` which is lacking in the same statement on the `if` above it. 
I think `getConfig` will not return falsy values, so no need for this `?`, but if we do use it let's stay consistent.

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -88,6 +88,20 @@ export type SpanNameHook = (
   defaultName: string
 ) => string;
 
+/**
+ * Function that can be used to add custom attributes to the current span or the root span on
+ * a Express request
+ * @param requestSpan - The Express request handler span
+ * @param expressRequest: - The Express request object
+ */
```
blumamir, line 96: This comment seems broken, and the params names and count does not match

blumamir (commented):
Left few comments.
I will be on vacation until August 5th, so feel free to resolve the comments when addressed

plugins/node/opentelemetry-instrumentation-express/src/types.ts
luismiramirez: Now it uses an `ExpressRequestInfo` as suggested

vmarchaud:
since @blumamir is on holiday that would be welcome if someone else can check this @open-telemetry/javascript-approvers

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -88,6 +88,16 @@ export type SpanNameHook = (
   defaultName: string
 ) => string;
 
+/**
+ * Function that can be used to add custom attributes to the current span or the root span on
+ * a Express request
+ * @param span - The Express middleware layer span.
+ * @param info: - An instance of ExpressRequestInfo that contains info about the request such as the route, and the layer type.
```
blumamir, line 95: nit: redundant `:` after `info`

plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
```diff
@@ -169,4 +171,97 @@ describe('ExpressInstrumentation hooks', () => {
...
+      port = httpServer.port;
+    });
+
+    afterEach(() => {
+      server.close();
+    });
+
+    it('should call requestHook when set in config', async () => {
+      const requestHook = sinon.spy((span: Span, info: ExpressRequestInfo) => {
+        span.setAttribute('http.method', info.request.method);
```
blumamir, line 198: nit: prefer to use the enum: `SemanticAttributes.HTTP_METHOD` (here and in other places)

blumamir (approved):

PR #1092, opened 2022-07-18
feat(instrumentation-document-load): add custom root span to document page load instrumentation

gregolsen:
Page load instrumentation creates and ends its own root span. However, for a typical single-page JS app, transaction is not completed after the initial page load is finished. Ideally, single-page app should be able to attach more spans to the root span and have control over when the root span is ended.

This PR extends instrumentation with an optional `rootSpan` config parameter. When `rootSpan` is passed instrumentation attaches all the child spans/attributes to that `rootSpan` and doesn't end the span.

## Checklist

- [N/A] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.


Rauno56:
@gregolsen, updating changelogs is actually done automatically during the release.

Can someone knowledgeable in RUM look at this? @t2t2, @MSNev, @mhennoch

gregolsen:
Thank you @Rauno56 - I got rid of the release update and rebased against the fresh main 🙇🏻

mhennoch:
Will take a look tomorrow

Rauno56:
@mhennoch, have you had a moment?

mhennoch:
Looked OK but wanted to test it but haven't had the time. Maybe also @t2t2 can take a look.

Rauno56:
@t2t2, will yoy have a moment to check it out?

t2t2:
Depends on if you want an answer to this PR wise or a big picture answer

Like yes pr does what it says it does, if user wants to manually manage the document load span like this then yes this is the way

But big picture wise, if I wanted a good set of rum instrumentations, I'd probably come up with a way to have documentLoad span be active from start of instrumentations being loaded until page load event happens. This would probably also decouple html resources (img, link, script tags) from document load instrumentation somewhat allowing it to also work for SPA use, making both documentLoad and userInteraction work similarly:

image

As noted on image this sort of idea wouldn't work for those who would base interaction duration from root span duration (but rather would need post-processing once sure that no more data comes from the trace). For those, quoting PR "Ideally, single-page app should be able to attach more spans to the root span and have control over when the root span is ended.", would be better. So the more automatically possible behavior would probably be better as an alternative choice of instrumentation for those who can use it

gregolsen:
A bit more context about our setup that may help here:
In our tracing setup page load is followed by the Ember.js app loading. Only after the Ember.js view is loaded we complete the traces. User clicking on something is considered a separate interaction and would emit a separate trace.

PR #1094, opened 2022-07-18
fix(hapi): ensure route wrapper starts a new context

aptomaKetil:

The Hapi instrumentation does not start a new context for the route handler. As a result, spans which are created from within the route handler end up as siblings of the route span, instead of as its children. Having the handler wrapper start a new context seems to be the standard, looking at the other framework instrumentation packages in this repo (koa, restify, fastify, express, etc.). The Hapi instrumentation does this correctly for extension methods, but not route handlers.

Wrap the handler invocation in a `context.with`, as is done for the other frameworks. Add a test to verify that spans created in the handler are children of the route span.

## Checklist


dyladan (approved):
Seems reasonable

vmarchaud (approved):

PR #1100, opened 2022-07-27
fix: allow react 18 in peer deps

### LINKED ISSUE #1098: React 18 support
Using `@opentelemetry/plugin-react-load` with React 18 throw an error.

Looks like by "just" extending the range of React versions supported, we could use "as is" the plugin with newer versions of React.

Forcing `--legacy-peer-deps` with npm.

### Additional context

Reproduction:
- Create a new project using React 18
- Add opentelemetry-js and `@opentelemetry/plugin-react-load`

Djiit:

- Fixes #1098 

- Added React 18 as a valid peerDependency version range.

## Checklist

### DIFF
--- a/plugins/web/opentelemetry-plugin-react-load/package.json
+++ b/plugins/web/opentelemetry-plugin-react-load/package.json
@@ -53,7 +53,7 @@
     "@opentelemetry/propagator-b3": "^1.3.1",
     "@types/mocha": "7.0.2",
     "@types/node": "16.11.21",
-    "@types/react": "17.0.16",
+    "@types/react": "18.0.15",
     "@types/react-addons-test-utils": "0.14.26",
     "@types/react-dom": "18.0.2",
     "@types/shimmer": "1.0.2",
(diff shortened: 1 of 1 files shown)


vmarchaud:
@Djiit You should bump it there too: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1100/files#diff-85c89db06090f742f4dca1a42f48eff7b266c253d13d36eb3c6bf39d65a492a9R74 so tests are run with the latest

vmarchaud:
> Sorry I think your link points to my changes grimacing

Github is not highlighting the line since its not in your changes, i though it worked. Anyway, i was refering to this line: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-plugin-react-load/package.json#L74

dyladan:
Is this package tested using `test-all-versions`? If so, can we make sure the new version is being tested?

Djiit:
Nope, AFAIU, it's a web plugin and it's not using `tav`. Looking at the test errors, looks like if we update the root react version, there's a *lot* more work to do :( not sure it's worth it just to get rid of a warning :(

Issue #1102, opened 2022-07-28
Options to set only span key in span name of memcached instrumentation


aadharsh-rengarajan:
In memcached instrumentation, we get span name like

```
 set api:1:value
 get api:1:value
```

which creates a difficulty when we want to group the span names and observe it for metrics purposes.

In our case, we would need to add an option that would let us only get memcached key instead of key:value like this
```

get or memcached:get
set or memcached:set

```
instead of
```

get api:1:value
set api:1:value

```
### Describe the solution you'd like to see

The solution would be adding an option along with enhancedDatabaseReporting. That would let us decide if we need the entire span with key value pair or just the key.

I would need an additional option where we can just ignore the memcached key in the span name and instead have query.type alone as the span name instead of query.type and query.key as mentioned below.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/2229cbe3b503e2b7c54f657e9f29698cc9f8f380/plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts#L37

Here we can add an additional option where we set a key called `cleanSpanName` which defaults to false. We can refer the cleanSpanName below in the place below. 

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/2229cbe3b503e2b7c54f657e9f29698cc9f8f380/plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts#L141

Where we just check this condition: ` const result = ( cleanSpanName ) ? span.updateName(`${query.type}`): span.updateName(`${query.type}  ${query.key}`);
### Describe alternatives you've considered

We can ignore the memcached value (query.key) in the span name as it is redundant and can be taken from span attributes..

RahulGanesh08:
+1 to this , since the telemetry instrumentation uses operation it performs  `get ,set`  and  memcache key` api:1:value` for the span name .  which results in much noise as jager considers every memcache operations unique since the key is present rather than grouping them by the opertions. 

as the solution mentioned above just like an option for `enhanced database reporting` , an option for filtering out the key from the span name would be very useful for many usecases.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts#L141

dyladan:
I would actually argue we should just remove the value from the name entirely. The specification wants low cardinality span names and the value can be extremely high cardinality. This is actually the reason for the spec being the way it is.

[...]

PR #1104, opened 2022-08-01
fix(memcached): low cardinality span name

aadharsh-rengarajan:

- Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1102

We have addressed the problem above by making the changes necessary and we can get the memcached value from the span attributes itself. 

- Removed query.type from the span name as it creates a high cardinality of span names.
- The span name will be memcached get or memcached set. (Open to change this span name)

## Checklist

- [x] Ran `npm test` for the edited package(s) on the latest commit if applicable.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
@@ -138,7 +138,7 @@ export class Instrumentation extends InstrumentationBase<typeof Memcached> {
       const query = original.apply(this, arguments as any);
       const callback = query.callback;
 
-      span.updateName(`${query.type} ${query.key}`);
+      span.updateName(`memcached ${query.type}`);
       span.setAttributes({
         'db.memcached.key': query.key,
         'db.memcached.lifetime': query.lifetime,
--- a/plugins/node/opentelemetry-instrumentation-memcached/test/index.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-memcached/test/index.test.ts
@@ -280,7 +280,7 @@ const assertSpans = (actualSpans: any[], expectedSpans: any[]) => {
       assert.notStrictEqual(span, undefined);
       assert.notStrictEqual(expected, undefined);
       assertMatch(span.name, new RegExp(expected.op));
-      assertMatch(span.name, new RegExp(expected.key));
+      assertMatch(span.name, new RegExp('memcached'));
       assert.strictEqual(span.kind, SpanKind.CLIENT);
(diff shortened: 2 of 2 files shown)


vmarchaud (approved):

plugins/node/opentelemetry-instrumentation-memcached/src/instrumentation.ts
```diff
@@ -138,7 +138,7 @@ export class Instrumentation extends InstrumentationBase<typeof Memcached> {
       const query = original.apply(this, arguments as any);
       const callback = query.callback;
 
-      span.updateName(`${query.type} ${query.key}`);
+      span.updateName(`memcached ${query.type}`);
```
blumamir, line 141: [there are tests](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4c8843be14896d1159a622c07eb3a049401ccba1/plugins/node/opentelemetry-instrumentation-memcached/test/index.test.ts#L283) asserting the key is found in the span name. They should be updated as well

aadharsh-rengarajan: This is done.

dyladan (approved):

plugins/node/opentelemetry-instrumentation-memcached/test/index.test.ts
```diff
@@ -280,7 +280,7 @@ const assertSpans = (actualSpans: any[], expectedSpans: any[]) => {
       assert.notStrictEqual(span, undefined);
       assert.notStrictEqual(expected, undefined);
       assertMatch(span.name, new RegExp(expected.op));
-      assertMatch(span.name, new RegExp(expected.key));
+      assertMatch(span.name, new RegExp('memcached'));
```
dyladan, line 283: This asserts only that the span contains the word `memcached`. It would also fail if the span name still contained the value. Do we know the possible query types so we can build a regex of possible names?

aadharsh-rengarajan: Yes, you are right. It only asserts if it contains the word memcached. Other assertions like expected.op in the line above does operation match like memcached set, get and so on. That was the case even before this case as it was only regex matching the assertion instead of strict matching. 

Did you mean `query types` as in memcached operations like set, get? Also, I don't understand this, _building regex of possible names_, in essence, building `memcached set`, `memcached get` as individual regex instead of matching them separately is what you mean?

dyladan: It's probably fine. I just think the assertion isn't providing a lot but it really wasn't before either.

Issue #1071, opened 2022-06-28
Support `pino@8` in `@opentelemetry/instrumentation-pino`


Rauno56:
`@opentelemetry/instrumentation-pino` doesn't have a component owner, but was contributed by @seemk.
Could you, by any chance, take a look?

zemd:
while I was trying to fix the https://github.com/open-telemetry/opentelemetry-js-contrib/issues/854 and also figuring out how to adopt pino@8 with opentelemetry I ended up with the thought that there is no need to make and maintain instrumentations for js loggers at all. Because each user always creates his own logger instance and shares it within the app, and each user configures this instance for each project, so no need to intercept module loader, that can break things and can be unstable due to issues like https://github.com/open-telemetry/opentelemetry-js-contrib/issues/854

So I think it so much easier to add manually `mixin` for the pino configuration:
```
export const logger = pino({
  // ... 
  mixin: () => {
    const record = {}
    const span = trace.getSpan(context.active());
    if (span) {
      const spanContext = span.spanContext();

      if (isSpanContextValid(spanContext)) {
        Object.assign(record, {
          trace_id: spanContext.traceId,
          span_id: spanContext.spanId,
          trace_flags: `0${spanContext.traceFlags.toString(16)}`,
        });
      }
    }
   const appCustomField = { bar: 'baz' };
   return Object.assign(appCustomField, record)
  }
})
```

and that's it. As improvement strategy `opentelemetry` lib can provide such type of helper function which would build desired object, instead of patching module.

dyladan:
I think we need to start enforcing ownership more forcefully the way the collector does. One way to start would be to add a section to each README of a component without an owner stating that it does not have a maintainer and should be treated as beta

Flarna:
The idea of autoinstrumention like `@opentelemetry/instrumentation-pino` is to have no manual tasks. Add OTel + wanted instrumentations to whatever app and it is traced/log enriched/...
All autoinstrumentations could be replaced by creating spans manually and it would be most likely less risk for the app, less overhead. But each and every dev has to do this in their app/lib.

zemd:
> is to have no manual tasks

This is valid intent! But taking pino-instrumentation as example we can see that the plugin is just replacing some amount manual tasks from one place to another, in the same time hiding issues. 

[...]

Issue #854, opened 2022-01-26
pino-instrumentation not injecting trace_id into logs when used with TypeScript


armandAkop:
### What version of OpenTelemetry are you using?
```
    "@opentelemetry/auto-instrumentations-node": "^0.25.0",
    "@opentelemetry/core": "^1.0.1",
    "@opentelemetry/exporter-collector": "^0.24.0",
    "@opentelemetry/exporter-prometheus": "^0.18.2",
    "@opentelemetry/instrumentation": "^0.24.0",
    "@opentelemetry/instrumentation-pino": "^0.28.1",
    "@opentelemetry/metrics": "^0.18.2",
    "@opentelemetry/node": "^0.24.0",
    "@opentelemetry/propagator-b3": "^1.0.0",
    "@opentelemetry/resources": "^0.18.2",
    "@opentelemetry/sdk-trace-node": "^1.0.1",
    "@opentelemetry/semantic-conventions": "^1.0.0",
    "pino": "^7.6.4",
   "express": "^4.17.1",
  "my-logger": "file: ../my-logger"
```

### What version of Node are you using?
v16.6.1

### What did you do?
**"my-logger"** is a separate typescript project on my machine, here is the simple implementation which instantiates an instance of pino and exports as default.
https://gist.github.com/armandAkop/3e064c2e378cdd8a442b339f3b5e28f6

I am running "tsc" in the repo and outputting the transpiled file to "dist/index.js", which contains
```node
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const pino_1 = require("pino");
const logger = pino_1.pino;
exports.default = (0, pino_1.pino)();
//# sourceMappingURL=index.js.map
```

Here is the script I am testing with auto-instrumentation-node and pino-instrumentation. This script lives in a folder called project-A and has a local reference to "my-logger" project on line 157
https://gist.github.com/armandAkop/d2fb6f806b31b6844262918489bf0679

### What did you expect to see?
I expected my test scripts log entry to contain the trace_id field

### What did you see instead?
trace_id is not present in the log message. When I change the **my-logger** project to be a regular Javascript project with an 
_index.js_ which looks like
```node
module.exports = require('pino')();
```

_AND_
changed _package.json_ to "main" property to "main": "index.js"

The trace_id appears in the log as expected

Flarna:
Not sure if this is the root cause but there is quite a mix/max of older and newer versions of the various otel packages.
[...]

PR #1108, opened 2022-08-02
feat: improve pino instrumentation by patching additional exports of the same function

Ugzuzg:

- Closes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/854
- Closes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1071

- expand supported version range (it's already tested by tav)
- instrument additional exports of pino

## Checklist


plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
```diff
@@ -72,6 +72,11 @@ export class PinoInstrumentation extends InstrumentationBase {
             return pinoModule(...(args as Parameters<Pino>));
           }, pinoModule);
 
+          // @ts-expect-error the same function reexported
+          patchedPino.pino = patchedPino;
```
Flarna, line 76: as far as I remember these two exports are not present in all versions so maybe use an if around like this:
```ts
if (typeof(pinoModule.pino) === "function") {
  pinoModule.pino = patchedPino;
}
```

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
```diff
@@ -75,11 +75,13 @@ describe('PinoInstrumentation', () => {
     return record;
   }
 
-  function init() {
+  function init(importType: 'global' | 'default' | 'pino' = 'global') {
     stream = new Writable();
     stream._write = () => {};
     writeSpy = sinon.spy(stream, 'write');
-    logger = pino(stream);
+    if (importType === 'global') logger = pino(stream);
```
Flarna, line 82: We usually use parenthesis even for single line if.

plugins/node/opentelemetry-instrumentation-pino/package.json
```diff
@@ -58,16 +64,16 @@
     "gts": "3.1.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
+    "pino": ">=8.0.0",
     "rimraf": "3.0.2",
+    "semver": "^7.3.5",
```
Flarna, line 69: usually we pin devDeps (except otel core components) to avoid "auto breaking" CI because of new releases

dyladan (commented):
implementation looks good. I agree with @Flarna's recommendation about the guard. Do you mind explaining why you decided to do the optional peer dep? Does this allow us to get types without bundling a real dependency? We may want to do the same in our other instrumentations.

Flarna:
`peerDependenciesMeta` in available since npm 7.x so node 14 users don't have it on default.

As the pino instrumentation has no hook which requires the Pino types it would be possbile to get rid of the dependency at all like it was done a while ago for some ohter instrumentations in #1017

dyladan:
> `peerDependenciesMeta` in available since npm 7.x so node 14 users don't have it on default.

It isn't the end of the world since if someone installs the instrumentation manually they will have pino anyway. The issue is if you use the metapackage to install all instrumentations npm will complain about missing peer deps for all packages you aren't choosing to use.

Ugzuzg:
> Do you mind explaining why you decided to do the optional peer dep? Does this allow us to get types without bundling a real dependency? We may want to do the same in our other instrumentations.

The main reason was to generate a warning if pino is installed, but the version isn't supported. It will still generate a typescript error if `skipLibCheck` is false.

plugins/node/opentelemetry-instrumentation-pino/package.json
Ugzuzg: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1108/commits/8b6f0b76fd741b577d80078b623caeabe73df26c

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
Ugzuzg: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1108/commits/a8700762c89367abfb706602bcf82cbd84f438f4

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
```diff
@@ -61,17 +60,22 @@ export class PinoInstrumentation extends InstrumentationBase {
                 args.splice(0, 0, {
                   mixin: instrumentation._getMixinFunction(),
                 });
-                return pinoModule(...(args as Parameters<Pino>));
+                return pinoModule(...(args as Parameters<any>));
```
Flarna, line 63:
```suggestion
                return pinoModule(...args);
```
Not sure but I think that this cast is no loner needed after moving to `InstrumentationNodeModuleDefinition<any>`

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
```diff
@@ -135,7 +139,7 @@ export class PinoInstrumentation extends InstrumentationBase {
     };
   }
 
-  private _combineOptions(options?: pino.LoggerOptions) {
+  private _combineOptions(options?: any) {
```
Flarna, line 142: You could actually keep the `pino.LoggerOptions` type here as it is a private function and a .d.ts file includes only the property name of private members but no types.

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
```diff
@@ -243,7 +272,7 @@ describe('PinoInstrumentation', () => {
       instrumentation.enable();
     });
 
-    beforeEach(init);
+    beforeEach(() => init());
```
Flarna, line 275: this change is actually not needed, it just creates an unneded function

Flarna (approved):

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
Ugzuzg: There is a type mismatch. `beforeEach` accepts a function and can pass a callback as an argument to it. `init`'s optional argument is not a function.

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
Ugzuzg: That would invite someone to add types to public functions, I think, as they are already imported. I would do that, at least, if there was no comment. :)

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
Ugzuzg: You're right: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1108/commits/e31c65f950d8ebecad92efcbd07c0da86167394f

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
Flarna: strange, at line 246 it seem to work. Anyhow, it's not important.

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
Flarna: Yeah. A test to verify that types are not part of the exports in the .d.ts files would be nice. But I have no idea how this could be done.

blumamir (approved):
LGTM

vmarchaud:
@Ugzuzg Could you rebase and fix the lint ? thanks

vmarchaud (commented):
Could you update the readme to state that it support pino@8 ? https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-pino#supported-versions

Ugzuzg:
@vmarchaud, done.

vmarchaud (approved):

Ugzuzg:
pino@8 does not support Node 12. I have modified the `.tav.yml` file to only test versions that support the currently used node version. Also, excluded pino instrumentation from Node 12 unit tests.

PR #1119, opened 2022-08-16
fix(koa): ignore generator-based Koa middleware

unflxw:

- Before this change, using a generator-based middleware alongside
the OpenTelemetry instrumentation would break the request, as the
instrumentation does not expect a generator-based middleware.
- Furthermore, the replacement of the middleware with its
generator-based patch would silence the deprecation warning about
the generator-based middleware.

- After this change, generator-based middleware are not
instrumented, and the deprecation warning is shown. Users who wish
for their generator-based middleware to be instrumented can
manually use `koa-convert`.

## Checklist


plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
```diff
@@ -134,6 +134,13 @@ export class KoaInstrumentation extends InstrumentationBase<typeof koa> {
     )
       return middlewareLayer;
     middlewareLayer[kLayerPatched] = true;
+    if (
+      middlewareLayer.constructor.name === 'GeneratorFunction' ||
+      middlewareLayer.constructor.name === 'AsyncGeneratorFunction'
+    ) {
+      api.diag.debug('ignoring generator-based Koa middleware layer');
+      return middlewareLayer;
+    }
```
blumamir, line 143: I think this block should be moved one line up, so that we don't mark 
```js
middlewareLayer[kLayerPatched] = true;
```

if we are not patching the middleware

blumamir, line 143: Can you please add a unit-test for this?

dyladan:
The generator detection is not perfect but it is actually not possible to have perfect detection based on this https://stackoverflow.com/a/19660350/1272149

Might be worth a note in the README but otherwise LGTM

unflxw:
@dyladan While the detection logic is indeed not perfect, it's worth noting that it's [the same detection logic used by the koa-convert library](https://github.com/koajs/convert/blob/bb06e77c8cf45d9afcbe5019c7fa9104631d5207/index.js#L29-L35), which is in turn [used by the Koa 2.x internals](https://github.com/koajs/koa/blob/d36e5f78d8829c7b2605b7c20edf25c8256f2347/lib/application.js#L128) to automatically convert legacy generator-based middleware to async middleware.

Meaning that in Koa 2.x, which is the Koa version we support, the only generator middleware who would ever work (with or without the OpenTelemetry instrumentation) are those that are `GeneratorFunction` or `AsyncGeneratorFunction`, as those are the only ones that will be converted.

It is true that, [in order to decide whether to call `koa-convert` and show the deprecation warning](https://github.com/koajs/koa/blob/v2.x/lib/application.js#L124), Koa itself uses [`is-generator-function`](https://github.com/inspect-js/is-generator-function/blob/main/index.js), which has some more convoluted logic for figuring out whether something is a generator function, and does not check for `AsyncGeneratorFunction`s. Meaning that an `AsyncGeneratorFunction` will not show the warning, will not be converted, and will not work, with or without the instrumentation. If you think using `is-generator-function` here (and pulling the extra dependency) as well is preferable, let me know.

dyladan:
Thanks for the thorough explanation. I think the logic is fine the way it is in this case

unflxw:
@blumamir When adding a unit test that re-uses the existing test middleware, I noticed what I believe is a bug in the logic to avoid re-patching a middleware. I'm not sure, because I'm not sure what the desired behaviour of `kLayerPatched` is. This is not a blocker for this PR, as I can rewrite the test to avoid this issue, I just wanted to make this issue known.

When a middleware is marked as patched, we mark it with `kLayerPatched`, to avoid patching it again if it is used more than once. However, when the same middleware is encountered again, since we already marked it as patched, we return the un-patched middleware, not the same patched middleware. What this means is that a middleware is only patched (and only instrumented) the first time it's given to `app.use`, and future uses are not re-patched, but they're not instrumented either. Consider the following (slightly contrived) example:

```
const commonMiddleware = (_ctx, next) => next()

const firstApp = new koa()
firstApp.use(commonMiddleware)
firstApp.use(commonMiddleware)
# ☝️ this won't be re-patched, but it also won't be instrumented at all!

const secondApp = new koa()
secondApp.use(commonMiddleware)
# ☝️ this won't be re-patched or instrumented either!
```

The second app will not have its middleware instrumented, because the middleware was already marked as patched the first time, and marking it as patched does not mean we return the previously patched version, but the middleware as-is.

So, contrived example aside, where does this happen in real life? Well, it happens in our test suite! Although the functionality it tests does actually work, the "should not create span if there is no parent span" test is currently a false positive of sorts, in the sense that the test passes for the wrong reasons: because it comes after other tests, and it reuses middleware that have already been used by the previous tests, the middleware that are passed to it are marked as patched, so the test uses the un-patched middleware, meaning it's not actually testing the instrumentation at all. Unless it's executed in isolation, the test does not fail when adding a parent span.

From this explanation on, choose your own adventure (I'm happy to implement whatever the desired behaviour is):

- This is *not* a bug, and a middleware that is used several times should only be instrumented on its first usage: in that case, our tests should not reuse middleware.
- This is a bug, and a middleware that is used several times should be instrumented for each use...
  - ... and we don't mind creating several patches: remove all `kLayerPatched` related logic.
  - ... but we still would like to avoid creating several patches for the same middleware: instead of setting `kLayerPatched` to `true`, set it to the patched function. When skipping re-patching a middleware, instead of returning the unpatched middleware, return the patched function stored in `kLayerPatched` (which might be better renamed as `kLayerPatch`).

Either way, there should be a test that checks whatever the expected behaviour is when reusing middleware.

vmarchaud (approved):

PR #1122, opened 2022-08-22
fix(pg): avoid disjoint spans from pg instrumentation

seemk:

* `pg.Client.connect` was uninstrumented. With `net` and `dns` instrumentations in use it meant that when `client.connect()` was called with no active context, it produced 2 disjoint traces, one for `net` and one for `dns`. This now produces a new span called `pg.connect`.
* `pg.Pool.connect` had the same problem where it did not activate the newly created context, causing disjoint spans for `net` and `dns`.

**Downstream spans from `pg` are now linked to `pg` spans correctly.**

Misc: When instrumenting `pg.Client.connect` the connection string contained `jdbc` prefix (which I removed). How did this end up in a Node.js instrumentation? :thinking: 

## Checklist


plugins/node/opentelemetry-instrumentation-pg/package.json
```diff
@@ -36,7 +36,7 @@
   "author": "OpenTelemetry Authors",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=8.12.0"
+    "node": ">=14.0.0"
```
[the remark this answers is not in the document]
seemk: Right - thought OTel JS dropped support for older Node versions, isn't that the case for instrumentations as well?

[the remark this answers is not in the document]
dyladan: We'll talk about this at SIG since I think everyone should be on the same page here. I think actually you're right since the base instrumentation class doesn't support old node versions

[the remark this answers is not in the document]
seemk: Removed the change from this PR so the Node version change can be done for all instrumentations at once.

dyladan (approved):

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -236,3 +240,22 @@ export function patchCallbackPGPool(
     cb.call(this, err, res, done);
   };
 }
+
+export function patchConnectErrorCallback(
```
blumamir, line 244: Is this callback only for errors?

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -107,6 +120,74 @@ export class PgInstrumentation extends InstrumentationBase {
...
+                  code: SpanStatusCode.ERROR,
+                  message: error.message,
+                });
+                span.end();
+                reject(error);
+              });
+            })
+        );
+      };
+    };
```
blumamir, line 188: This block is a duplicate of the code in `_getPoolConnectPatch`.
Could be nice to extract into a function and invoke from both places

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
blumamir, line 244: There is already a function `patchCallbackPGPool` which does almost the same thing. The only difference is the arguments list to the callback which we don't use.
Can we reuse one function for both cases?
Maybe write it with [`arguments` array parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) instead of listing the different arguments for each case

blumamir (approved):
LGTM. left few suggestions

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
seemk: Refactored this into a single function named `handleConnectResult`

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
seemk: Renamed it to `patchClientConnectCallback`

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -236,3 +240,22 @@ export function patchCallbackPGPool(
...
+  span: Span,
+  cb: PgErrorCallback
+): PgErrorCallback {
+  return function patchedConnectErrorCallback(
+    this: pgTypes.Client,
+    err: Error
+  ) {
+    if (err) {
```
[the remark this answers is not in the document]
seemk: Done!

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
seemk: I am not sure about this, think there is some value in keeping descriptive function names. I did rename the function though

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -50,16 +50,20 @@ function getCommandFromText(text?: string): string {
   return words[0].length > 0 ? words[0] : 'unknown';
 }
 
-export function getJDBCString(params: PgClientConnectionParams) {
```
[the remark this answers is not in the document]
seemk: Removed this change and reused the same type albeit with changing the fields to possibly be undefined. In `client.connect` case the client itself is passed to `getConnectionString` and the client has all of these fields as optional. Perhaps the `PgClientConnectionParams` fields should've been optional since the beginning.

seemk:
Seem to be unrelated unit tests failing atm

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -67,11 +68,23 @@ export class PgInstrumentation extends InstrumentationBase {
...
         this._wrap(
           moduleExports.Client.prototype,
           'query',
           this._getClientQueryPatch() as never
         );
+
+        this._wrap(
+          moduleExports.Client.prototype,
+          'connect',
+          this._getClientConnectPatch() as never
```
Rauno56, line 85: Using `as never` to avoid type checker complaining is silly. Fine if you don't want to bloat the PR fixing an unrelated issue, but we shouldn't add more.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -254,40 +310,44 @@ export class PgInstrumentation extends InstrumentationBase {
...
+      )
+      .catch((error: Error) => {
+        return new Promise((_, reject) => {
+          span.setStatus({
+            code: SpanStatusCode.ERROR,
+            message: error.message,
+          });
+          span.end();
+          reject(error);
+        });
```
Rauno56, line 350: No need to create Promises using the constructor here. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1125#discussion_r967104209

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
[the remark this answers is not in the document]
Rauno56: Did you forget to push?

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
seemk: This instrumentation seems to use `as never` everywhere in the code, I changed them to `as any` now like all other instrumentations. The connect function itself is overloaded.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
seemk: Thanks, fixed

Rauno56 (approved):
Awesome! Thanks for taking this on!

PR #1125, opened 2022-08-23
feat: add redis 4 connect span

seemk:

This PR adds a new span `redis.connect` which is produced by `client.connect()` call. This is a small nicety so that downstream `net` and `dns` spans are properly attached to a redis span when initializing the connection, otherwise there would be 2 disjoint single span traces.

## Checklist


plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -332,6 +338,52 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
...
+
+        const attributes = {
+          [SemanticAttributes.DB_SYSTEM]: DbSystemValues.REDIS,
+          [SemanticAttributes.NET_PEER_NAME]: options?.socket?.host,
+          [SemanticAttributes.NET_PEER_PORT]: options?.socket?.port,
+          [SemanticAttributes.DB_CONNECTION_STRING]: options?.url,
+        };
+
+        const span = plugin.tracer.startSpan(
+          `${RedisInstrumentation.COMPONENT}.connect`,
```
blumamir, line 355: Instrumentation currently builds spans names with format:
```js
`${RedisInstrumentation.COMPONENT}-${commandName}`
```

Do we want to stay consistent here and use `-`? I am ok with changing it to dot in all places but it's a bit weird to have a mix from the same instrumentation IMO.

plugins/node/opentelemetry-instrumentation-redis-4/test/redis.test.ts
```diff
@@ -183,6 +185,54 @@ describe('redis@^4.0.0', () => {
     });
   });
 
+  describe('client connect', () => {
+    it('produces a span', async () => {
+      const client = createClient({
```
blumamir, line 190: I suggest giving it another name. `client` is already defined in the global `describe` scope which can be confusing to read.

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -332,6 +338,52 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
...
+    return function connectWrapper(original: Function) {
+      return function patchedConnect(this: any): Promise<void> {
+        const options = this.options;
+
+        const attributes = {
+          [SemanticAttributes.DB_SYSTEM]: DbSystemValues.REDIS,
+          [SemanticAttributes.NET_PEER_NAME]: options?.socket?.host,
+          [SemanticAttributes.NET_PEER_PORT]: options?.socket?.port,
+          [SemanticAttributes.DB_CONNECTION_STRING]: options?.url,
+        };
```
blumamir, line 352: This code is a duplication of the one in [`_traceClientCommand` function](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/11ab701051cedca051f52707673b28a26e965389/plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts#L354).

WDYT about extracting it to `utils` (maybe call it `getClientAttributes`) and calling the function from both places?

blumamir (approved):
Left few optional suggestions

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
seemk: I was doubting whether to use a `.` or `-` here (or something else). The actual redis commands have `-`, so I wanted to differentiate between a redis command and connect operation (which isn't a command). But I agree it looks somewhat inconsistent  :thinking:

blumamir: Actually, I am not even sure if the `redis` prefix is encouraged or not [in semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#semantic-conventions-for-database-client-calls)

> The span name SHOULD be set to a low cardinality value representing the statement executed on the database. It MAY be a stored procedure name (without arguments), DB statement without variable arguments, operation name, etc

[The examples](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#redis) also suggest naming redis spans in like `"HMSET myhash"`.

Anyway, this is not very relevant to this PR, so feel free to do what you think is best. We can revisit this in the future.

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -332,6 +338,52 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
...
+          .catch((error: Error) => {
+            return new Promise((_, reject) => {
+              span.setStatus({
+                code: SpanStatusCode.ERROR,
+                message: error.message,
+              });
+              span.end();
+              reject(error);
+            });
+          });
```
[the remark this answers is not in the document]
blumamir: Actually, when thinking about it again, this might not be equivalent.
If this promise rejects and the user does not catch the error, then without instrumentation the app will crash. If instrumentation adds a `catch` block to the returned promise, then the app will no longer crash which is not desirable.

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
seemk: Good point, moved it to `utils`

plugins/node/opentelemetry-instrumentation-redis-4/test/redis.test.ts
seemk: Renamed the connection test clients

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
seemk: For consistency I changed it to `-`, but indeed this might need another PR

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -332,6 +338,52 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
...
+        return res
+          .then((result: unknown) => {
+            return new Promise(resolve => {
+              span.end();
+              resolve(result);
+            });
+          })
+          .catch((error: Error) => {
+            return new Promise((_, reject) => {
+              span.setStatus({
```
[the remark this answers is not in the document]
seemk: Done

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -332,6 +335,48 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
...
+        const res = context.with(trace.setSpan(context.active(), span), () => {
+          return original.apply(this);
+        });
+
+        return res
+          .then((result: unknown) => {
+            return new Promise(resolve => {
+              span.end();
+              resolve(result);
+            });
```
Rauno56, line 363: This doesn't need the use of the Promise constructor:
```suggestion
            span.end();
            return result;
```

Rauno56 (approved):
Thanks!
Nit: `context.with` calls could be without creating a nested function.

plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
seemk: Fixed :pray:

PR #1126, opened 2022-08-24
feat: add fastly instrumentation lib

parangsjo-ingka:
The intention of the module/library in this PR is to add support for OpenTelemetry tracing for services running on Fastly Compute@Edge.

When running services on highly distributed environments where performance and reliability are key, like Fastly Compute@Edge, observability and distributed tracing is really helpful to understand where time is consumed. We tried to instrument our code in Fastly Compute@Edge with the standard OpenTelemetry JS libs but realized that we needed a module/lib that enables OpenTelemetry tracing in Fastly Compute@Edge with a smaller footprint and overhead. Based on the nature of Fastly Compute@Edge the module/library is written in JavaScript rather than TypeScript.

## Checklist

- [ ] Followed the style guidelines of this project, but made some adjustments to the eslint and gitignore configuration to work with JavaScript and Fastly artifacts
- [ ] Unit tests have been added
- [ ] Documentation has been added


parangsjo-ingka:
Fastly (vendor) has released their version of JavaScript instrumentation in Compute@Edge which makes this lib redundant.

Issue #1132, opened 2022-08-29
TS2430: Interface 'Request' incorrectly extends interface 'IncomingMessage' when using getNodeAutoInstrumentations


pragmaticivan:
### What version of OpenTelemetry are you using?
```
    "@opentelemetry/api": "^1.1.0",
    "@opentelemetry/auto-instrumentations-node": "^0.31.2",
    "@opentelemetry/core": "^1.6.0",
    "@opentelemetry/exporter-jaeger": "^1.6.0",
    "@opentelemetry/exporter-prometheus": "^0.32.0",
    "@opentelemetry/host-metrics": "^0.29.0",
    "@opentelemetry/instrumentation-graphql": "^0.29.0",
    "@opentelemetry/instrumentation-pino": "^0.30.0",
    "@opentelemetry/instrumentation-winston": "^0.29.0",
    "@opentelemetry/propagator-b3": "^1.6.0",
    "@opentelemetry/propagator-jaeger": "^1.6.0",
    "@opentelemetry/sdk-node": "^0.32.0",
```

### What version of Node are you using?
16.x

### What did you do?
Just using `getNodeAutoInstrumentations()`.

### What did you expect to see?
No error.

### What did you see instead?
```
node_modules/@opentelemetry/instrumentation-restify/node_modules/@types/restify/index.d.ts:36:18 - error TS2430: Interface 'Request' incorrectly extends interface 'IncomingMessage'.
  Types of property 'log' are incompatible.
    Type 'Logger' is not assignable to type 'Logger<LoggerOptions>'.
      Property 'silent' is missing in type 'Logger' but required in type 'BaseLogger'.

36 export interface Request extends http.IncomingMessage {
```

jscherer92:
Taking a look

PR #1133, opened 2022-08-30
fix(restify): upgrade package to fix type problem #1132

jscherer92:
#1132 

Updated the Restify package for @opentelemetry/instrumentation-restify

## Checklist

- [ x] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-restify/package.json
+++ b/plugins/node/opentelemetry-instrumentation-restify/package.json
@@ -54,7 +54,7 @@
     "gts": "3.1.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
-    "restify": "4.3.4",
+    "restify": "8.6.1",
     "rimraf": "3.0.2",
     "ts-mocha": "10.0.0",
     "typescript": "4.3.5"
--- a/plugins/node/opentelemetry-instrumentation-restify/package.json
+++ b/plugins/node/opentelemetry-instrumentation-restify/package.json
@@ -54,16 +54,16 @@
     "gts": "3.1.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
-    "restify": "8.6.1",
+    "restify": "4.3.4",
     "rimraf": "3.0.2",
     "ts-mocha": "10.0.0",
-    "typescript": "4.3.5"
+    "typescript": "4.3.5",
+    "@types/restify": "4.3.8"
   },
   "dependencies": {
     "@opentelemetry/core": "^1.0.0",
     "@opentelemetry/instrumentation": "^0.31.0",
-    "@opentelemetry/semantic-conventions": "^1.0.0",
-    "@types/restify": "4.3.8"
+    "@opentelemetry/semantic-conventions": "^1.0.0"
   },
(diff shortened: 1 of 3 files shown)


plugins/node/opentelemetry-instrumentation-restify/package.json
```diff
@@ -54,7 +54,7 @@
     "gts": "3.1.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
-    "restify": "4.3.4",
+    "restify": "8.6.1",
```
blumamir, line 57: Thank you for submitting a fix. I think the original error originated from [`@types/restify` here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/ec9ee131635dc2db88deea4f2efb887ff6f60577/plugins/node/opentelemetry-instrumentation-restify/package.json#L66) which is a regular dependency. Your PR addresses the dev dependency on "restify" packages, which might be updated as well regardless IMO.

This seems reasonable to me, as version `4.3.4` is 4 years old and has only 20K downloads vs 313K for most popular version.

I wonder if the reason that `4.3.4` produced an error implies that the instrumentation no longer supports it's [advertised range of >= 4.0.0](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-restify#supported-versions)? (probably due to some other dependency that broke types?)

This instrumentation has no `test-all-versions` setup in place, so it's difficult to tell, but maybe you investigated this issue and can share insights?

blumamir:
I think the original issue is in part due to the `@types/restify` package even being a regular dependency.
The only reason it's a dependency for end users (instead of dev dependency) is due to the template argument for InstrumentationBase [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/ec9ee131635dc2db88deea4f2efb887ff6f60577/plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts#L37).

What we tend to do is to specify `any` type instead, and move the dependency to be dev-dependency, which can still break our development and CI, but not end users.

As this is related to #1132 , could be a good chance to fix this on the way if you are up for it. If not, just bumping the versions and verifing that v4 is still supported is also a great improvement.

jscherer92:
I'm happy to do that fix. I know what I've done instead for this type of stuff is re export the type but if we just want to go with any that is fine.

Would we want unknown instead so to keep type safety?

plugins/node/opentelemetry-instrumentation-restify/package.json
```diff
@@ -57,13 +57,13 @@
     "restify": "4.3.4",
     "rimraf": "3.0.2",
     "ts-mocha": "10.0.0",
-    "typescript": "4.3.5"
+    "typescript": "4.3.5",
+    "@types/restify": "4.3.8"
```
dyladan, line 61: If you move the types to dev dependencies I believe users have compilation failures. That's why we moved them to regular dependencies to begin with.

jscherer92: I'll take a look and see if that happens locally.

PR #1134, opened 2022-08-30
feat(graphql): allow the ability to ignore certain fields

jscherer92:
The inability to specify fields that you may want to ignore due to high levels processing needed to get through all of those fields.

Add a new option into the config to ignore fields.

## Checklist


jscherer92:
@dyladan sorry that this is so late and it does not have code coverage on the new code, but this is what I was talking about in our one meeting. When you have arrays that are 20,000 objects that each of those objects have 20-40 properties, it leads to a major slowdown. This is one way of fixing the issue since it would allow us to ignore fields that we do not want to put into the tracing.

It also should allow us to hide values from entering into our telemetry platform. If you think this is a good idea, I'll continue with this and add an actual feature and finish this up.

jscherer92:
@Flarna @blumamir or others, can I get a set of eyes on this? Will help me do some more things with this instrumentation from a PII perspective

Flarna:
@jscherer92 I'm not familiar with graphql so can't really comment that much on this change.
The current owner of this instrumenation is @obecny but as far as I know he moved to other topics then OTel a while ago.

@open-telemetry/javascript-maintainers  what is the right approach here? I think we should be happy to have people to contribute but the contribution guideline is not that clear to me how to move ownership.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -296,9 +296,11 @@ export function wrapFields(
   tracer: api.Tracer,
   getConfig: () => GraphQLInstrumentationParsedConfig
 ): void {
+  const config = getConfig();
   if (
     !type ||
     typeof type.getFields !== 'function' ||
+    config.ignoreFields.includes(type.name) ||
```
obecny, line 303: when you instantiate class you could create a hash map of all keys to ingore using `includes` everywhere which is slow.

obecny (commented):
please improve the performance - create hash map at start from config array property

jscherer92:
Would a HashSet be better?

obecny:
> Would a HashSet be better?

Pseude code

```
this._ignorehMap = {};
for (let i = 0, j = ignoreKeys.length; i < j; i++) {
  this._ignorehMap[ignoreKeys[i]] = true;
}


if (
    !type ||
    typeof type.getFields !== 'function' ||
    this._ignorehMap(type.name) ||
...
```

jscherer92:
Oh I agree that we could do it that way. I'm just wondering if the HashSet would be better;

`
this.ignoreFields = new HashSet(this.config.ignoreFields);


if(this.ignoreFields.has(field))
`

obecny:
> Oh I agree that we could do it that way. I'm just wondering if the HashSet would be better;
> 
> ` this.ignoreFields = new HashSet(this.config.ignoreFields);
> 
> if(this.ignoreFields.has(field)) `

hash map is faster the set

jscherer92:
should be switched over now

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -71,6 +74,17 @@ export class GraphQLInstrumentation extends InstrumentationBase {
       VERSION,
       Object.assign({}, DEFAULT_CONFIG, config)
     );
+    const ignoreMap = new Map();
+    config.ignoreFields?.map(item => ignoreMap.set(item, true));
```
obecny, line 78: map is meant to return a new array if you want to iterate please use forEach instead

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -71,6 +74,17 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+    this._setIgnoreMap(ignoreMap);
+  }
+
+  private _getIgnoreMap(): Map<string, boolean> {
```
obecny, line 82: this seems to be over engineering , a private function that returns a private, I think you can get rid of this, unless there is very specific reason for that ?

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -71,6 +74,17 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+    return this._ignoreMap;
+  }
+
+  private _setIgnoreMap(map: Map<string, boolean> = new Map()) {
+    this._ignoreMap = map;
```
obecny, line 87: this whole function can be removed

plugins/node/opentelemetry-instrumentation-graphql/src/types.ts
```diff
@@ -56,7 +56,12 @@ export interface GraphQLInstrumentationConfig extends InstrumentationConfig {
    * @default false
    */
   mergeItems?: boolean;
-
+  /**
+   * List of fields to not add spans to.
```
obecny, line 60:
```suggestion
   * List of fields name / key  to be ignored for span creation.
```

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -363,7 +367,10 @@ export function wrapFieldResolver<TSource = any, TContext = any, TArgs = any>(
 
     let field: any;
     let shouldEndSpan = false;
-    if (config.depth >= 0 && config.depth < depth) {
+    if (
+      (config.depth >= 0 && config.depth < depth) ||
+      config.ignoreFields.includes(path[path.length - 1])
```
obecny, line 372: you created a map but you still didn't use it

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
obecny: you already created a new map, then you create a new new one and then you created a new private function to override the created at the beginning. 
why cant you simply do
```
config.ignoreFields?.forEach(item => this._ignoreMap.set(item, true));
```
?

obecny (changes requested):

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
jscherer92: It is now in two locations for the set behavior, so I kept this but repurposed

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
jscherer92: the reason is due to the wrapField and wrapFieldResolver functions. Since it can on the fly change the config, we need to make sure that those util functions get the updates. There are embedded functions so this will make sure we don't have to rerun that code.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
jscherer92: fixed

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -58,11 +58,14 @@ const DEFAULT_CONFIG: GraphQLInstrumentationConfig = {
   mergeItems: false,
   depth: -1,
   allowValues: false,
+  ignoreFields: [],
 };
 
 const supportedVersions = ['>=14'];
 
 export class GraphQLInstrumentation extends InstrumentationBase {
+  private _ignoreMap = new Map();
```
obecny, line 67: this is always overridden in  `_setIgnoreMap` so I think it should be safe to remove assigning this to new map here ?

obecny (approved):
There is one comment which I hope you can address before merging. Other than than lgtm, thx for changes.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
jscherer92: Yep! You are definitely right, fixed it in the latest change

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -294,11 +294,13 @@ export function getSourceFromLocation(
...
   type: Maybe<graphqlTypes.GraphQLObjectType & OtelPatched>,
   tracer: api.Tracer,
-  getConfig: () => GraphQLInstrumentationParsedConfig
+  getConfig: () => GraphQLInstrumentationParsedConfig,
+  ignoreMap: () => Map<string, boolean>
 ): void {
   if (
     !type ||
     typeof type.getFields !== 'function' ||
+    ignoreMap().has(type.name) ||
```
Rauno56, line 303: Is the indirection required?

Rauno56: OK, saw https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1134#discussion_r963077861 now.
The whole thing about config changing is such an edge case I think, but alright - I guess that's another topic.

It shouldn't be an issue if you never overwrite the variable. Map has a `clear()` method just for that.

Rauno56:
> hash map is faster the set

I cannot understand how that can even be the case. I mean... even if it would, why didn't the smart people behind node think of implementing Set through Map then?

Rauno56:
GraphQL tests seem to be flaky(see #1184). Merging this will be blocked until that's fixed.

@jscherer92, maybe you'd be willing to look into it?

Rauno56:
Nvm, GraphQL tests are now fine.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -363,7 +371,10 @@ export function wrapFieldResolver<TSource = any, TContext = any, TArgs = any>(
 
     let field: any;
     let shouldEndSpan = false;
-    if (config.depth >= 0 && config.depth < depth) {
+    if (
+      (config.depth >= 0 && config.depth < depth) ||
+      ignoreMap().has(path[path.length - 1])
```
blumamir, line 376: Can we use `info.fieldName` instead of `path[path.length - 1]`?
Seems to me more straightforward and robust, than using the output of the `pathToArray`

blumamir, line 376: What is the motivation of running the resolver with the context of parent span, instead of just not instrumenting this call?

If I understand correctly, then for this query
```
{
  foo {
    bar
  }
}
```

if bar is ignored, then downstream spans created under the `bar` resolver will be children of `foo`, right?

Is this what we want?

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -310,12 +312,17 @@ export function wrapFields(
   Object.keys(fields).forEach(key => {
     const field = fields[key];
 
-    if (!field) {
+    if (!field || ignoreMap().has(key)) {
```
blumamir, line 315: If I understand the flow correctly, then this will not only ignore this field, but also all the schema resolver field below it, e.g.
```
{
  foo {
    bar
  }
}
```

Then ignoring only `foo` will terminate this recursion function, preventing it from calling `wrapFields` below on `foo`s field resolver, including `bar`.

This will apply for resolvers on the schema, but for "fieldResolver" invocations, the behavior will be different, as nothing will prevent it from being called on `bar` and thus the span will be recorded, right?

I suggest refactoring the code to have consistency between the 2 resolver options, and adding a test for it.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -58,11 +58,14 @@ const DEFAULT_CONFIG: GraphQLInstrumentationConfig = {
   mergeItems: false,
   depth: -1,
   allowValues: false,
+  ignoreFields: [],
 };
 
 const supportedVersions = ['>=14'];
 
 export class GraphQLInstrumentation extends InstrumentationBase {
+  private _ignoreMap!: Map<string, boolean>;
```
blumamir, line 67: @obecny - you wrote
> hash map is faster the set

I guess you refer to this issue?
https://stackoverflow.com/questions/49841972/javascript-how-come-map-has-is-so-much-faster-than-set-has-and-array-indexof

I don't know the current status of the issue, [but from looking on node source code](https://github.com/nodejs/node/blob/36805e852471cc7a24524a329848da5c8f163864/deps/v8/src/compiler/js-call-reducer.cc#L7557), it seems that `Set.has` is nowdays treated the same as `Map.has` in V8:
```cpp
Reduction JSCallReducer::ReduceMapPrototypeHas(Node* node) {
  return ReduceCollectionPrototypeHas(node, CollectionKind::kMap);
}

Reduction JSCallReducer::ReduceSetPrototypeHas(Node* node) {
  return ReduceCollectionPrototypeHas(node, CollectionKind::kSet);
}
```

I think for our case this "ignoreFields" config list is expected to be very small, so anyhow I see little value in harming instrumentation readability for this optimization.
Adding a `Map` from string to `true` just seems wired to me.

blumamir, line 67: The config option is called "ignoreFields".

How about calling this field "_ignoreFieldsMap" (here and in other places)? Stating "ignore" without any context what is ignored is less readable IMO.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -79,6 +83,16 @@ export class GraphQLInstrumentation extends InstrumentationBase {
 
   override setConfig(config: GraphQLInstrumentationConfig = {}) {
     this._config = Object.assign({}, DEFAULT_CONFIG, config);
+    this._setIgnoreMap(config.ignoreFields);
+  }
+
+  private _setIgnoreMap(vals?: Array<string>) {
+    this._ignoreMap = new Map();
+    vals?.forEach(item => this._ignoreMap.set(item, true));
```
blumamir, line 91: nit: you can do it all in one go with no side effects:
```js
this._ignoreMap = new Map(vals?.map(fieldName => [fieldName, true]));
```

plugins/node/opentelemetry-instrumentation-graphql/test/graphql.test.ts
```diff
@@ -572,6 +572,99 @@ describe('graphql', () => {
     });
   });
 
+  describe('when ignoreFields has a property', () => {
```
blumamir, line 575: Currently, instrumentation tests only with resolvers defined on the schema itself, but there are no tests for the patching of the "fieldResolver" (or defaultFieldResolver).

This PR adds logic that is currently not tested for the "fieldResolver".
Could be amazing to test your logic for this flow as well. Since there are no tests for this even for other features, I am ok if you choose not to do it.

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
obecny: Originally it was using ".includes" so if you optimise something it is best to use the fastest way imho. This might be called thousands of times during usage so I would say it is still better to use less cpu than more. This is a kind of thing you do once and then you probably wont be touching this again for a long time. 
```
ignoreFields" config list is expected to be very small
```
what if not ?, besides that the list can be small but still it can be called many times per second.

blumamir:
@jscherer92 Thanks for the time you put into this. I apologize for taking long to review.
Are you still interested in it or should I close?

jscherer92:
Definitely still interested! I can take a look at the comments and fix up what is there

Issue #1136, opened 2022-08-30
[Bug] DNS Lookup and Promise Lookup test for error NOT_FOUND failing


MSNev:
### What version of OpenTelemetry are you using?
Current Contrib repo

### What version of Node are you using?
16.14.0

### What did you do?
- clean repo
- npm install
- npm run compile
- npm run test

### What did you expect to see?
passing tests

### What did you see instead?
- dns tests failed in local environment due to timeout exceeding 2 seconds

PR #1138, opened 2022-08-30
ci: [Bug] DNS Lookup and Promise Lookup test for error NOT_FOUND failing #1136

MSNev:

[Bug] DNS Lookup and Promise Lookup test for error NOT_FOUND failing #1136 

DNS resolution in local environment is exceeding 2 seconds (due to lots of fallback DNS servers) for invalid URL this is causing these 2 tests to timeout with the default 2 second delay.

Resolution:
Extend the timeout for these 2 tests so for environments where DNS resolution can take longer than expected the tests still pass.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-dns/test/integrations/dns-lookup.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-dns/test/integrations/dns-lookup.test.ts
@@ -98,24 +98,30 @@ describe('dns.lookup()', () => {
       });
     });
 
-    it('should export a valid span with error NOT_FOUND', done => {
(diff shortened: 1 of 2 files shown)


dyladan (approved):
Seems fine as a short term solution but wouldn't it be better to mock the DNS and not rely on a slow external service?

MSNev:
> Seems fine as a short term solution but wouldn't it be better to mock the DNS and not rely on a slow external service?

I suspect most people won't hit this issue so I think mocking could be overkill.

Flarna (approved):

vmarchaud (approved):

blumamir (approved):

PR #1140, opened 2022-08-31
fix(mysql): set proper context for outgoing queries

seemk:
Set the proper context for outgoing mysql queries so `net` and possibly `dns` instrumentation spans are linked correctly.


t2t2 (approved):

mhennoch (approved):

blumamir (approved):

dyladan:
Can you add a test for it?

seemk:
> Can you add a test for it?

Done, not sure if there is a better way to do it instead of stubbing a private method used in the `query` function. Suggestions welcome :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -328,7 +337,9 @@ export class MySQLInstrumentation extends InstrumentationBase<
           });
         }
         span.end();
-        return originalCallback(...arguments);
+        return context.with(parentContext, () =>
```
Rauno56, line 340: :point_up:

seemk: Both cases produce compiler errors, as the original function is of `Function` type. In the first case it's an overloaded function so unless there's a good/simple way to fix the types I'd leave it as is :thinking:

Rauno56: Oh well! Let's not dwell on that too much then!

PR #1141, opened 2022-08-31
fix: mongodb types fails to compile with latest tsc v4.8

osherv:
There were some sub-packages that used the latest typescript version and caused errors against the deprecated @types/mongodb: 3.6.20 dependency.

1. Moved the deprecated MongoDB dependency to a devDependency:
    * The use of `@types/mongodb` in the instrumentation itself was minor. 
       All the places that used the MongoDB types were replaced with `any` as it applied in many other instrumentations,
       in order to avoid the use of deprecated dependencies.

2. Aligned all typescript versions in the project to 4.3.5:
    There were three packages that were used latest typescript version (duo to "^version):
    * "koa-example"
    * express-example
    * @opentelemetry/resource-detector-docker

Also, we have to think about how to restrict those typescript version update changes.

## Checklist

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-docker/package.json
+++ b/detectors/node/opentelemetry-resource-detector-docker/package.json
@@ -48,7 +48,7 @@
     "rimraf": "3.0.2",
     "sinon": "14.0.0",
     "ts-mocha": "10.0.0",
-    "typescript": "^4.5.5"
+    "typescript": "4.3.5"
   },
   "peerDependencies": {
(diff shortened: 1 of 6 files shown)


MSNev (approved):

Flarna (approved):

blumamir (approved):

blumamir:
Thank you @osherv 
Merging since this PR fixes our currently broken CI, and it already has 3 approves.

PR #1065, opened 2022-06-22
chore: remove support for node 8-12, add 18

Rauno56:
## Which problem is this PR solving?

It's a complimentary PR for https://github.com/open-telemetry/opentelemetry-js/pull/3048

## Short description of the changes

Required changes for node@18:

- 1e6bddf481f6e9201a0f77edd4596da987e2c41e: turned off tests for alibaba detector **in node@18**.
- [fceb7b7](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1065/commits/fceb7b708235bd677de04f064962269c09978a9a): turned off tests for fastify instrumentation **in node@18**
- https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1065/commits/0067aecf2cf725387c4dfc0f692cef274ffe21be: error message used in tests changed
- https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1065/commits/ef29bcfdde997232038fe0817e409baf4bd43113: Changed how root context equality is checked. Check the commit message.

## Checklist

- [ ] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.

PR #1148, opened 2022-09-01
test(nestjs-core): fixed nestjs-core tav to work on all node versions

osherv:

test-all-versions for node8 failed on nestjs-core@8 tests.
Here's the error:
https://github.com/open-telemetry/opentelemetry-js-contrib/runs/8134447972?check_suite_focus=true

Pinning the tests to version 8.0.0 fix the problem, and now it works for all node versions.


dyladan:
We're not getting the latest versions of these dependencies in our tests now though right? Isn't it better to test in an environment as similar to what a user would see as possible? It might be better to drop node 8 testing

/cc @open-telemetry/javascript-maintainers

osherv:
@dyladan yes you're right. Maybe it's better to remove the whole node8 support.
Let me know if that's preferred :)

dyladan:
Waiting to hear what other maintainers think but removing node 8 is probably the way to go

osherv:
@dyladan @Rauno56 According to this pr: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1065
I think that you agree that the best this is to remove the support for tav in node8.

blumamir:
Thank you @osherv for fixing this.

I support dropping node8 support for nest-js instrumentation, but it seems that nestjs (at least with latest version 9) [suuports node >= 12.9.0](https://github.com/nestjs/nest/blob/8d66cece7990fa6fd74c75dea87d76def0920cb1/package.json#L176). So it should fail on node10 as well, right? any idea why we see problem for node8 only?

osherv:
@blumamir Thanks :)
I have no idea, i can dig into this, but we can drop this version as well XD

Rauno56:
@osherv, I agree - let's add skip for 10 as well and merge this.

osherv:
@Rauno56 Thanks, added 10 version

Rauno56 (approved):

PR #1161, opened 2022-09-07
feat: add auto-instrumentation for `pg-native`

airhorns:

Add instrumentation for `pg-native`

pg-native is a higher-performance, API compatible postgres client that is a drop in replacement for the pg package. It uses a native extension for performance, but the JS bits quack the same as the pure-JS pg, so the same instrumentation just works!

## Checklist

- [X] Ran `npm run test-all-versions` for the edited package(s) on the latest commit if applicable.


Rauno56:
Would be cool if that just works! Let's wait for the green build.

Rauno56:
From the fact that the last version had a syntax error, I assume you didn't run TAV on the last - could you please do that and confirm?

Current test setup(running tests for `pg` and `pg-native`) is suboptimal for the TAV runs. We need a way to isolate them to test all supported versions of those libs, but still have `npm test` do the "correct" default.

Rauno56 (changes requested):
As per my last comment, I mark it with changes requested to help with PR triage.

blumamir:
@airhorns are you still working on this?

airhorns:
Sorry folks, I dont have time to finish this!

PR #1167, opened 2022-09-12
feat(testing-utils): add metrics support to the test utils package

osherv:

The testing agent didn't support metrics creation and collection.
We are starting to work on the metrics collection in the instrumentations so it would be nice if 
we could add tests more easily.

The test utils package now also supports test metrics. 

## Checklist


packages/opentelemetry-test-utils/README.md
```diff
@@ -49,16 +49,21 @@ Or by using config file / package.json config:
...
+  getTestMetrics,
+  registerInstrumentationTesting,
+} from '@opentelemetry/test-utils';
 
 const instrumentation = registerInstrumentationTesting(new MyAwesomeInstrumentation());
 
 it('some test', () => {
     // your code that generate spans for this test
     const spans: ReadableSpan[] = getTestSpans();
+    const metrics: ResourceMetrics[] = getTestMetrics();
```
legendecas, line 66: IIUC, `getTestMetrics` is an async method.

```suggestion
    const metrics: ResourceMetrics[] = await getTestMetrics();
```

packages/opentelemetry-test-utils/src/instrumentations/otel-provider-api.ts
```diff
@@ -18,27 +18,76 @@ import {
...
+  numberOfExports: number
+): Promise<ResourceMetrics[]> {
+  if (numberOfExports <= 0) {
+    throw new Error('numberOfExports must be greater than or equal to 0');
+  }
+
+  const exporter = getMetricsTestMemoryExporter()!;
+  let totalExports = 0;
+  while (totalExports < numberOfExports) {
+    await new Promise(resolve => setTimeout(resolve, 20));
```
legendecas, line 79: See https://github.com/open-telemetry/opentelemetry-js/pull/3242, we should be able to implement a deterministic MetricReader that performs metric collection every time `getTestMetrics` is called.

legendecas (commented):

packages/opentelemetry-test-utils/src/instrumentations/otel-provider-api.ts
osherv: Wow awesome! Thanks!
Added TODO to use this reader.

legendecas: The reader in the PR I referenced is not going to be exported since it is used in ad-hoc tests. I believe we can create a subclass of MetricReader in this test-utils package instead, to provide basic testing features.

osherv: Oh :/ Should we ask them to move it to the src? similar to the InMemoryExporters.
I can copy paste the code but I think it's better to implement it in one place

legendecas: No, it is not part of the spec so I believe we are not going to expose it in the public interfaces. I think duplicating the work here is fine since the interfaces exposed here are going to be public, whilst the test facilities in that PR are not going to be exposed as public interfaces.

osherv: Oh, that makes sense. I'll implement it here. Thanks for quick response :)

osherv:
Hey @legendecas @pichlermarc :)
Added TestMetricReader, ill be more than happy if you'll take a look :)
Thanks!

packages/opentelemetry-test-utils/package.json
```diff
@@ -51,6 +51,8 @@
     "@opentelemetry/exporter-jaeger": "^1.3.1",
     "@opentelemetry/instrumentation": "^0.32.0",
     "@opentelemetry/sdk-trace-node": "^1.3.1",
+    "@opentelemetry/api-metrics": "^0.32.0",
+    "@opentelemetry/sdk-metrics": "^0.32.0",
```
legendecas, line 55:
```suggestion
    "@opentelemetry/sdk-metrics": "^0.33.0",
```

packages/opentelemetry-test-utils/src/TestMetricReader.ts
```diff
@@ -0,0 +1,60 @@
...
+        if (result.error != null) {
+          reject(result.error);
+        } else {
+          resolve();
+        }
+      });
+    });
+  }
+
+  selectAggregation(instrumentType: InstrumentType): Aggregation {
```
legendecas, line 51: This method is no longer needed to be overridden with this patch: https://github.com/open-telemetry/opentelemetry-js/commit/b73c29dcff65d2c9db9c8f6ccbdaabb14b5622bd (released in v0.33.0)

packages/opentelemetry-test-utils/src/TestMetricReader.ts
```diff
@@ -0,0 +1,60 @@
...
+    return Aggregation.Default();
+  }
+
+  selectAggregationTemporality(
```
legendecas, line 55: Ditto

packages/opentelemetry-test-utils/src/instrumentations/index.ts
```diff
@@ -38,17 +41,27 @@ export const mochaHooks = {
...
+      [SemanticResourceAttributes.SERVICE_NAME]: serviceName,
     });
-    getInstrumentation()?.setTracerProvider(provider);
+
+    const providers = registerInstrumentationTestingProvider(
+      {
+        resource: testResource,
+      },
+      {
+        resource: testResource,
```
legendecas, line 53: This seems superfluous to repeat the identical options. Can we try to merge the trace provider options and the meter provider options? As a reference: https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-sdk-node/src/types.ts#L29.

packages/opentelemetry-test-utils/src/instrumentations/otel-provider-api.ts
```diff
@@ -18,27 +18,88 @@ import {
...
+  return _global[OTEL_METRICS_TESTING_READER];
+};
+
+export const setMetricsTestMemoryExporter = (
+  memoryExporter: InMemoryMetricExporter
+) => {
+  _global[OTEL_METRICS_TESTING_MEMORY_EXPORTER] = memoryExporter;
+};
+
+export const setTracingTestMemoryExporter = (
```
legendecas, line 69: Not sure if we have naming guidance on this, but I'd suggest that we keep the term aligned with the SDK names

```suggestion
export const setTestMemorySpanExporter = (
```

packages/opentelemetry-test-utils/src/instrumentations/otel-provider-api.ts
```diff
@@ -18,27 +18,88 @@ import {
...
+export const getTracingTestMemoryExporter = ():
+  | InMemorySpanExporter
+  | undefined => {
+  return _global[OTEL_TRACING_TESTING_MEMORY_EXPORTER];
 };
-const _global = global as OTelProviderApiGlobal;
 
-export const getTestMemoryExporter = (): InMemorySpanExporter | undefined => {
-  return _global[OTEL_TESTING_MEMORY_EXPORTER];
+export const getMetricsTestMemoryExporter = ():
```
legendecas, line 53: Ditto for naming.

```suggestion
export const getTestMemoryMetricsExporter = ():
```

packages/opentelemetry-test-utils/src/TestMetricReader.ts
osherv: Fixed

packages/opentelemetry-test-utils/src/TestMetricReader.ts
osherv: Fixed

packages/opentelemetry-test-utils/src/instrumentations/index.ts
osherv: Sorry for the delay, fixed that comment :)

legendecas (approved):
Thanks for working on this!

blumamir:
@osherv - this PR is approved. When you have time, please fix the conflicts and the tests so we can merge.
Thanks!

packages/opentelemetry-test-utils/README.md
```diff
@@ -11,10 +11,10 @@ This package exports a mocha [root hook plugin](https://mochajs.org/#root-hook-p
 
 This package:
 
-- Initializes and registers a global trace provider for tests.
-- Registers a global memory exporter which can be referenced in test to access span.
+- Initializes and registers a global trace, metric provider for tests.
+- Registers a global memory exporters which (for both metricsm traces) can be referenced in test to access span.
 - Make sure there is only a single instance of an instrumentation class that is used across different `.spec.ts` files so patching is consistent, deterministic and idiomatic.
-- Reset the memory exporter before each test, so spans do not leak from one test to another.
+- Reset the memory exporters before each test, so spans do not leak from one test to another.
```
pichlermarc, line 17:
```suggestion
- Reset the memory exporters before each test, so spans/metrics do not leak from one test to another.
```

pichlermarc (approved):
Thank you for working on this :slightly_smiling_face: 
It would be really nice to have this added to the test utils, some conflicts still need to be resolved first, though :slightly_smiling_face:

PR #1170, opened 2022-09-14
feat(mongodb): collect mongodb4 metrics

### LINKED ISSUE #1157: Metrics instrumentaion MongoDB
Hey, we want to start to implement metrics collection in the instrumentations like other SDKs already did.
I'll start with MongoDB according to:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md

If you find it right to start implementing metrics collection, please assign this issue to me :)

osherv:

-

Fixes #1157 

Added metrics collection for MongoDB >= 4, according to: 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md
This is the first of many contributions planned under metrics collections at the instrumentation level, so I'll be really happy to get as much as comments as we can on this :)
Thanks!

## Checklist


vmarchaud:
Do you want to wait for https://github.com/open-telemetry/opentelemetry-js/pull/3267 to merge this ?

vmarchaud:
Converting to draft in the mean time :)

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -640,28 +826,56 @@ export class MongoDBInstrumentation extends InstrumentationBase {
...
+          instrumentation._connectionsUsage.add(-1, {
+            state: 'idle',
+            'pool.name': instrumentation._poolName,
+          });
+        }
         return resultHandler.apply(this, args);
       });
     };
   }
+  private setPoolName(options: any) {
```
haddasbronfman, line 867: Comment to myself and others - I think until this pr: https://github.com/open-telemetry/opentelemetry-specification/pull/3050 is not approved, we can't approve the pool name issue. (This pr wants to define the pool.name to be the db.connection_string in case the DB doesn't supply pool name).

haddasbronfman:
@open-telemetry/javascript-approvers Can someone please review?

plugins/node/opentelemetry-instrumentation-mongodb/src/types.ts
```diff
@@ -61,3 +62,33 @@ export type CommandResult = {
...
+  ): void;
+};
+
+// https://github.com/mongodb/node-mongodb-native/blob/v4.2.2/src/cmap/connect.ts
+export type V4Connect = {
+  connect: (options: any, callback: any) => void;
+};
+
+// https://github.com/mongodb/node-mongodb-native/blob/v4.2.2/src/sessions.ts
+export type V4Session = {
```
pichlermarc, line 91: Can this and `V4Connect` be an internal type instead? :thinking:

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -36,23 +36,47 @@ import {
...
     super('@opentelemetry/instrumentation-mongodb', VERSION, _config);
   }
 
+  override _updateMetricInstruments() {
+    this._connectionsUsage = this.meter.createUpDownCounter(
+      'db.client.connections.usage',
+      {
+        description:
+          'The number of connections that are currently in the state referenced by the attribute "state".',
+        unit: '{connections}',
```
pichlermarc, line 65: Not sure if the one in the semconv is a typo or this one here (see https://github.com/open-telemetry/opentelemetry-specification/blob/c210cf5884cf2c06dc1d1a6a34a4ea76c542841f/specification/metrics/semantic_conventions/database-metrics.md?plain=1#L34) :slightly_smiling_face: 
```suggestion
        unit: '{connection}',
```

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -36,23 +36,47 @@ import {
...
   constructor(protected override _config: MongoDBInstrumentationConfig = {}) {
     super('@opentelemetry/instrumentation-mongodb', VERSION, _config);
   }
 
+  override _updateMetricInstruments() {
+    this._connectionsUsage = this.meter.createUpDownCounter(
+      'db.client.connections.usage',
+      {
+        description:
+          'The number of connections that are currently in the state referenced by the attribute "state".',
```
pichlermarc, line 64: We've recently had an issue in the core repo (https://github.com/open-telemetry/opentelemetry-js/pull/3710) where the prometheus collector exporter would complain when the description does not exactly match the semconv and other language implementations are.

It's a bit unfortunate that the description has to be followed exactly even though it is not specified as identifying, but I think this could save us some maintenance later :slightly_smiling_face: 

```suggestion
          'The number of connections that are currently in state described by the state attribute.',
```

pichlermarc (commented):
Thanks for adding this, looking good :slightly_smiling_face: 
Just a few question/suggestions around types.

Looks like the failing tests may be unrelated, it may be happening on main too. :thinking:

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
haddasbronfman: I looked [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-units) and it says: "All non-units that use curly braces to annotate a quantity need to match the grammatical number of the quantity it represent. For example if measuring the number of individual requests to a process the unit would be {request}, not {requests}.".
If I understand correctly we should use {connection} and not {connections}. what do you think?

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
haddasbronfman: two questions about this:
1. When you say 'exactly' it means that 'state' should appear with single quotes? I mean `'state'` and not `state` as in semconv. 
2. This is relevant also for mysql, right? see [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1220#discussion_r1046851794), where we had a long discussion about this :)

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
pichlermarc: Ah yes, that's also what I understand when reading the spec there. I agree, I think we should switch to `{connection}`. :slightly_smiling_face:

haddasbronfman: I should change it in mysql too https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1220

haddasbronfman:
@legendecas @dyladan  Hi. I'm looking for reviewers who are familiar with metrics. Can you please review this PR?

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.metrics.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    );
+
+    assert.strictEqual(exportedMetrics.length, 1);
+    let metrics = exportedMetrics[0].scopeMetrics[0].metrics;
+    assert.strictEqual(metrics.length, 1);
+    assert.strictEqual(metrics[0].dataPointType, DataPointType.SUM);
+
+    assert.strictEqual(
+      metrics[0].descriptor.description,
+      'The number of connections that are currently in the state referenced by the attribute "state".'
```
pichlermarc, line 122: Could it be that this test is not being run? :thinking: 
The description has changed, so this should fail but [the job](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/4853420360/jobs/8649566069) somehow passes. :thinking: 
[TAV fails, though.](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/4853420381/jobs/8649578432?pr=1170)

haddasbronfman: you're right. look at the [tav.yml](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-mongodb/.tav.yml):
```
mongodb:
  jobs:
    - versions: ">=3.3 <4"
      commands: npm run test
    - versions: ">=4 <5"
      commands: npm run test-new-versions
``` 
 version 4 tests will run by the command 'test-new-versions' and version 3 tests will run by the command 'npm run test'. This is why you don't see the v4 test in the unit-test job. 

I'll fix the description issue so TAV will pass. 

But there is something still bothering me: in unit-test job it seems that v3 test didn't run at all:
image
I'll take a look at this.

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.metrics.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  ResourceMetrics,
+} from '@opentelemetry/sdk-metrics';
+
+import * as mongodb from 'mongodb';
+
+const otelTestingMeterProvider = new MeterProvider();
+const inMemoryMetricsExporter = new InMemoryMetricExporter(
+  AggregationTemporality.CUMULATIVE
+);
+const metricReader = new PeriodicExportingMetricReader({
```
legendecas, line 37: Nit: IIUC, a pulling metric reader might be better suited for testing usage since it can pull metrics proactively instead of awaiting timeouts.

legendecas (approved):
I'm not familiar with mongodb. The metrics instrumentation looks good to me.

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.metrics.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    assert.strictEqual(exportedMetrics.length, 1);
+    let metrics = exportedMetrics[0].scopeMetrics[0].metrics;
+    assert.strictEqual(metrics.length, 1);
+    assert.strictEqual(metrics[0].dataPointType, DataPointType.SUM);
+
+    assert.strictEqual(
+      metrics[0].descriptor.description,
+      'The number of connections that are currently in state described by the state attribute.'
+    );
+    assert.strictEqual(metrics[0].descriptor.unit, '{connections}');
```
pichlermarc, line 124: looks like tav is failing here now :thinking: 
```suggestion
    assert.strictEqual(metrics[0].descriptor.unit, '{connection}');
```

haddasbronfman: at least we have TAV. it helps us don't miss anything :)

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.metrics.test.ts
pichlermarc: Marking this as resolved as the tests are also not running on main. Thank you for creating the issue (#1487) :slightly_smiling_face:

pichlermarc:
Looks like TAV is failing on line 137 in the v4 metrics tests, the idle state is now exporting `2`. :thinking:

PR #1172, opened 2022-09-14
fix: handle string ports for Socket.connect

seemk:
`Socket.connect` with a string port (e.g. `socket.connect('4317', 'localhost', () => {})`) was previously creating an IPC span while it should have been creating a TCP span.

For example `cassandra-driver` is a package that uses this form of `Socket.connect`.


Rauno56 (approved):
:+1: Nice catch.

PR #1180, opened 2022-09-18
feat(cassandra-responsehook): added response hook to execute func

### LINKED ISSUE #1162: ResponseHook for cassandra driver
Would love to have a responseHook to a query that being sent to cassandra driver. This will allow me to insert some custom attributes after response is back.

A responseHook function so I can insert custom attributes to the span after query was made.

samimusallam:

Fixes #1162 

- Added Response hook to the execute function in Cassandra driver's instrumentation


plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -45,6 +45,14 @@ export class CassandraDriverInstrumentation extends InstrumentationBase {
...
   }
 
+  private _getConfig(): CassandraDriverInstrumentationConfig {
+    return this._config;
+  }
+
+  override setConfig(config?: CassandraDriverInstrumentationConfig): void {
+    super.setConfig(config);
+  }
+
```
osherv, line 55: Why this is needed?

samimusallam: it's not really needed, but it gives the user another option to configure the instrumentation
eg: new CassandraDriverInstrumentation().setConfig({ })

osherv: Oh cool

osherv (approved):

samimusallam:
@dyladan Would it be possible to review this PR, please?
The original issue #1162 is not assigned to anyone and has no tags assigned to it

plugins/node/opentelemetry-instrumentation-cassandra/src/types.ts
```diff
@@ -15,6 +15,12 @@
  */
 
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+import { Span } from '@opentelemetry/api';
+import type * as CassandraDriver from 'cassandra-driver';
+
+export interface CassandraDriverResponseCustomAttributeFunction {
+  (span: Span, response: CassandraDriver.types.ResultSet): void;
```
blumamir, line 22: We usually construct a hook signature to have `span` object and then an `Info` object which contains all the relevant data for this hook.
In the future, more data could be added to the hook (for example: moduleVersion) and then we end up with long signatures with many arguments which are harder to work with.

You can see an example [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b35277bb5fc66910e8942bc0b64347b68ecffa26/plugins/node/instrumentation-amqplib/src/types.ts#L20)

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -319,6 +335,14 @@ export class CassandraDriverInstrumentation extends InstrumentationBase {
       attributes,
     });
   }
+
+  private _callResponseHook(span: Span, response: any) {
+    safeExecuteInTheMiddle(
+      () => this._getConfig().responseHook!(span, response),
+      () => {},
```
blumamir, line 342: we usually `diag.err` the exception so it's easier to debug. A common pattern is:
```js
            e => {
              if (e) {
                this._diag.error('responseHook error', e);
              }
            },
```

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -228,6 +253,38 @@ describe('CassandraDriverInstrumentation', () => {
...
+    describe('responseHook', () => {
+      const customAttribute = {
+        attributeName: 'custom-attribute',
+        attributeValue: 'response-value',
+      };
+
+      before(() => {
+        const config: CassandraDriverInstrumentationConfig = {
+          responseHook: (span, response) => {
+            span.setAttribute(
```
blumamir, line 266: Could be nice to also add an attribute from `response`, to validate hook populate it as expected

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -228,6 +253,38 @@ describe('CassandraDriverInstrumentation', () => {
         assertSingleSpan('cassandra-driver.execute', query.substr(0, 25));
       });
     });
+
+    describe('responseHook', () => {
```
blumamir, line 257: Can you please also add a test where the hook throws, to assert the exception is handled and not affecting the user?
You can see many examples in other instrumentation tests

plugins/node/opentelemetry-instrumentation-cassandra/src/types.ts
```diff
@@ -29,4 +35,6 @@ export interface CassandraDriverInstrumentationConfig
    * @default 65536
    */
   maxQueryLength?: number;
+  /** Function for adding custom attributes before response is handled */
+  responseHook?: CassandraDriverResponseCustomAttributeFunction;
```
blumamir, line 39: This new instrumentation option should also be documented in the [README](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-cassandra#instrumentation-options) under "Instrumentation Options"

plugins/node/opentelemetry-instrumentation-cassandra/src/types.ts
samimusallam: True, I'll construct an object which will contain the response for now, which will allow for easier data addition in the future

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
samimusallam: Of course! I added a test to make sure hook exceptions do not affect span creation or user flow

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -64,6 +67,13 @@ function assertSpan(
     attributes[SemanticAttributes.DB_STATEMENT] = query;
   }
 
+  if (customAttributes) {
+    customAttributes.forEach(customAttribute => {
```
blumamir, line 71: Instead of the `if` you can write `customAttributes?.forEach` which is a bit cleaner.

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -228,6 +251,64 @@ describe('CassandraDriverInstrumentation', () => {
...
+        assertAttributeInSpan('cassandra-driver.execute', [
+          {
+            attributeName: customAttributeName,
+            attributeValue: customAttributeValue,
+          },
+          {
+            attributeName: responseAttributeName,
+            attributeValue: '2',
+          },
+        ]);
```
blumamir, line 288: There is already an [`Attributes` interface](https://github.com/open-telemetry/opentelemetry-js/blob/afe065772360d3acec8b8fc15e20536fd896c6c6/api/src/common/Attributes.ts#L22) that does exactly this.

Do we need to introduce a new data structure?

An alternative for the attributes array can be:
```js
{[customAttributeName]: customAttributeValue, [responseAttributeName]: '2'}
```

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -228,6 +251,64 @@ describe('CassandraDriverInstrumentation', () => {
...
+
+      it('adds custom attributes to span', async () => {
+        const responseAttributeName = 'response.attribute';
+        const customAttributeName = 'custom.attribute';
+        const customAttributeValue = 'custom attribute value';
+
+        instrumentation.setConfig({
+          responseHook: (span: Span, responseInfo: ResponseInfo) => {
+            const row = responseInfo.response.first();
+            span.setAttribute(responseAttributeName, row.count.toString());
```
blumamir, line 270: nit: why convert the row count to string?
Even though it's allowed, it makes more sense to me to be a number. I think the test should promote idiomatic use of attributes

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -76,6 +86,18 @@ function assertSingleSpan(name: string, query?: string, status?: SpanStatus) {
   assertSpan(span, name, query, status);
 }
 
+function assertAttributeInSpan(
+  name: string,
+  attributes?: { attributeName: string; attributeValue: string }[],
+  query?: string,
+  status?: SpanStatus
```
blumamir, line 93: The 2 only invocations of this function do not set `query` and `status`.
I suggest removing them

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -76,6 +86,18 @@ function assertSingleSpan(name: string, query?: string, status?: SpanStatus) {
   assertSpan(span, name, query, status);
 }
 
+function assertAttributeInSpan(
```
blumamir, line 89: This function also asserts that there is only 1 span, but the name does not explain this.

In the current implementation, it asserts more than attributes (span name, query, status). Can be nice to rename this function to something that better describes what it does

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
Rauno56, line 55: Are these any different from the base implementation? There are `getConfig` and `setConfig` defined on the InstrumentationBase.

plugins/node/opentelemetry-instrumentation-cassandra/src/types.ts
```diff
@@ -15,6 +15,16 @@
  */
 
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+import { Span } from '@opentelemetry/api';
+import type * as CassandraDriver from 'cassandra-driver';
+
+export interface ResponseInfo {
+  response: CassandraDriver.types.ResultSet;
+}
```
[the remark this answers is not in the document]
Rauno56: That's correct. We should not use anything from instrumented package in the public interface.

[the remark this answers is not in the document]
samimusallam: That's true, this will throw as it's in the devDependency. But I'm not sure what the right approach is, should I move it to the regular dependency, or replace the type with `any` or a different object?

plugins/node/opentelemetry-instrumentation-cassandra/README.md
```diff
@@ -39,11 +39,12 @@ await client.execute('select * from foo');
...
+| `responseHook`              | `CassandraDriverResponseCustomAttributeFunction` | `undefined` | Hook for adding custom attributes before response is handled                                                                                                                                                                   |
+| `maxQueryLength`            | `number`                                         | `65536`     | If `enhancedDatabaseReporting` is enabled, limits the attached query strings                                                                                                                                                   |
+| to this length.             |                                                  |             |                                                                                                                                                                                                                                |
```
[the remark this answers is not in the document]
samimusallam: Yes, that extra line was there from a previous PR and I wasn't sure if I should fix it here or not.
What do you think? It should be alright to fix it in this PR, right?

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
samimusallam: Thanks! wasn't aware of that interface. Using it now instead :)

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
samimusallam: The reason for converting the count to string is because its type is Long, which can't be set as an attribute in setAttribute().
I updated the test to parse it to a number that maybe makes more sense for the count :)

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
samimusallam: Not really the only reason for these was the return/parameter types. Let me know what you think of [the changes](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1180/commits/f0e3bfca434ed26bd14f8f56a1870285a0cf7535) :)

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -319,6 +335,21 @@ export class CassandraDriverInstrumentation extends InstrumentationBase {
...
     });
   }
+
+  private _callResponseHook(span: Span, response: any) {
+    safeExecuteInTheMiddle(
+      () => this._getConfig().responseHook!(span, { response }),
+      e => {
+        if (e) {
+          this._diag.error(
+            'cassandra-driver instrumentation: responseHook error',
```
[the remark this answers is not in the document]
samimusallam: Didn't know that! Thanks for letting me know :)

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -336,10 +367,17 @@ function combineQueries(queries: Array<string | { query: string }>) {
     .join('\n');
 }
 
-function wrapPromise<T>(span: Span, promise: Promise<T>): Promise<T> {
+function wrapPromise<T>(
+  span: Span,
+  promise: Promise<T>,
+  successCallback?: (span: Span, result: any) => void
```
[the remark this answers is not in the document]
samimusallam: Yup, missed that 😅

plugins/node/opentelemetry-instrumentation-cassandra/README.md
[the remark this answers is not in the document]
blumamir: Yes of course, thank you :)

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -53,9 +56,10 @@ function assertSpan(
   span: ReadableSpan,
   name: string,
   query?: string,
-  status?: SpanStatus
+  status?: SpanStatus,
+  customAttributes?: Attributes
 ) {
-  const attributes = {
+  const attributes: any = {
```
blumamir, line 62: nit: Instead of `any` you can use `Attributes`. Not sure about it, but I think that the `Attributes` interface was introduce in `@opentelemetry/api@1.1.0`, and since this instrumentation supports 
```json
"peerDependencies": {
},
```

I think we might need to use `SpanAttributes`

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -64,6 +68,12 @@ function assertSpan(
     attributes[SemanticAttributes.DB_STATEMENT] = query;
   }
 
+  if (customAttributes) {
+    for (const [k, v] of Object.entries(customAttributes)) {
+      attributes[k] = v;
+    }
+  }
```
blumamir, line 75: nit: You can use the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and merge this with above definition like this:

```js
const attributes: SpanAttributes = {
    [SemanticAttributes.DB_SYSTEM]: DbSystemValues.CASSANDRA,
    [SemanticAttributes.DB_USER]: 'cassandra',
    ...(customAttributes ?? {}),
}
```

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
```diff
@@ -228,6 +246,60 @@ describe('CassandraDriverInstrumentation', () => {
...
+
+      it('adds custom attributes to span', async () => {
+        const responseAttributeName = 'response.attribute';
+        const customAttributeName = 'custom.attribute';
+        const customAttributeValue = 'custom attribute value';
+
+        const config: CassandraDriverInstrumentationConfig = {
+          responseHook: (span: Span, responseInfo: ResponseInfo) => {
+            const row = responseInfo.response.first();
+            const responseValue = parseInt(row.count);
```
blumamir, line 263: according to the docs, [`parseInt` argument is `string`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt).

If `row.count` is [long](https://www.npmjs.com/package/long), might be more correct to use `row.count.toNumber()`?

plugins/node/opentelemetry-instrumentation-cassandra/src/types.ts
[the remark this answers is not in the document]
blumamir: yeah that's a bummer.

I can think of these 2 non-optimal options:
1. use `any` with a comment that points to the actual type to guide users on how to use this hook.
2. Copy the `ResultSet` type, along with it's dependencies into this file, with a comment that points to the source.

@Rauno56 - Do you have any other ideas?

plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts
```diff
@@ -319,6 +329,18 @@ export class CassandraDriverInstrumentation extends InstrumentationBase {
       attributes,
     });
   }
+
+  private _callResponseHook(span: Span, response: any) {
+    safeExecuteInTheMiddle(
+      () => this._config.responseHook!(span, { response }),
```
blumamir, line 335: optional nit: we use here `this._config.responseHook!` which will throw if undefined. The reason we do it is because before calling the function we check `if (plugin._config.responseHook)`.
I think it might be more robust and increase readability to move the if check into this function, this way the function doesn't need to assume things about it's caller, and someone reading this code can see the 2 related statements in the same scope, close to each other.

plugins/node/opentelemetry-instrumentation-cassandra/test/cassandra-driver.test.ts
samimusallam: It does parse it correctly, but yes this makes more sense :)

plugins/node/opentelemetry-instrumentation-cassandra/src/types.ts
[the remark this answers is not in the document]
samimusallam: If I remember correctly, I've seen the 2nd option in other instrumentations, which I think is better than using `any` since it can be a bit confusing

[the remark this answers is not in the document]
blumamir: I agree, but in this specific case, [the `ResultSet` interface is quite complex and has transitive dependencies on other types](https://github.com/datastax/nodejs-driver/blob/d42176e4baa1cfc3df79699cc3b5d575c86e3cec/lib/types/index.d.ts#L323):
```js
 interface ResultSet extends Iterable<Row>, AsyncIterable<Row> {
    info: {
      queriedHost: string,
      triedHosts: { [key: string]: any; },
      speculativeExecutions: number,
      achievedConsistency: consistencies,
      traceId: Uuid,
      warnings: string[],
      customPayload: any
    };

    columns: Array<{ name: string, type: { code: dataTypes, info: any } }>;
    nextPage: (() => void) | null;
    pageState: string;
    rowLength: number;
    rows: Row[];

    first(): Row;

    wasApplied(): boolean;
  }
```

Maybe not all of them are relevant for the hook and we can just declare an interface with a subset?
I guess you are trying to collect the response payload which only cares about the `rows` property.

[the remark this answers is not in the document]
samimusallam: True it has a lot of dependencies.
I feel like the user will expect the entire result to be in the hook since this is the expected behavior.
But I do agree with you that a subset could be sufficient since the most important part of the result is the `rows` property. Do you think we should go with that?

blumamir (approved):

PR #1181, opened 2022-09-19
feat: container ID detector for cgroup v2

abhee11:
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1173
docker v2 detector
-

- generalized detector for docker v1, v2 and containerd

## Checklist


detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -74,4 +94,4 @@ export class DockerCGroupV1Detector implements Detector {
   }
 }
 
-export const dockerCGroupV1Detector = new DockerCGroupV1Detector();
+export const dockerDetector = new DockerDetector();
```
blumamir, line 97: is this name change a breaking change for existing users?

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -25,10 +25,12 @@ import * as fs from 'fs';
 import * as util from 'util';
 import { diag } from '@opentelemetry/api';
 
-export class DockerCGroupV1Detector implements Detector {
+export class DockerDetector implements Detector {
   readonly CONTAINER_ID_LENGTH = 64;
   readonly DEFAULT_CGROUP_PATH = '/proc/self/cgroup';
```
blumamir, line 30: Should we call it `DEFAULT_CGROUP_V1_PATH` to make it consistent with the new definition?

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -51,21 +53,39 @@ export class DockerCGroupV1Detector implements Detector {
...
-      const splitData = rawData.trim().split('\n');
+      let splitData = rawData.trim().split('\n');
       for (const str of splitData) {
         if (str.length >= this.CONTAINER_ID_LENGTH) {
           return str.substring(str.length - this.CONTAINER_ID_LENGTH);
         }
       }
+
+      // If this code is reached - we then check for the V2_PATH
+      rawData = await DockerDetector.readFileAsync(
```
blumamir, line 68: Suggestion: this function is currently trying to get `containerId` for both V1 and V2.
I think it will improve the readability to split the function into 2: `_getContainerIdV1()` and `_getContainerIdV2()` and then call them from here:
```js
return _getContainerIdV1() ?? _getContainerIdV2();
```

The code is also very similar and you might have opportunities to reuse some of it for both cases

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -51,21 +53,39 @@ export class DockerCGroupV1Detector implements Detector {
...
+        if (str.includes(this.HOSTNAME)) {
+          str = str
+            .split('/')
+            .filter(s => s.length >= this.CONTAINER_ID_LENGTH)[0];
+          if (str.length >= this.CONTAINER_ID_LENGTH) {
+            return str.substring(str.length - this.CONTAINER_ID_LENGTH);
+          }
+        }
+      }
     } catch (e) {
```
blumamir, line 84: Question: When V1 tries to `readFileAsync(this.DEFAULT_CGROUP_PATH, ..)`, if the file is missing I think it will throw "Error: ENOENT: no such file or directory, open '/proc/self/cgroup'" (at least on my setup).

In this case, it will fall into the `catch` and will not attempt to try with DEFAULT_CGROUP_V2_PATH. Am I missing something?

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -51,21 +53,39 @@ export class DockerCGroupV1Detector implements Detector {
...
+      // If this code is reached - we then check for the V2_PATH
+      rawData = await DockerDetector.readFileAsync(
+        this.DEFAULT_CGROUP_V2_PATH,
+        this.UTF8_UNICODE
+      );
+
+      splitData = rawData.trim().split('\n');
+      for (let str of splitData) {
+        if (str.includes(this.HOSTNAME)) {
+          str = str
```
blumamir, line 76: I think it will increase readability if instead of modifying `str` we introduce a new variable with a good name (is `containerId` what this string represents?): 
```js
const containerId = str.split('/').filter(...);
```

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -51,21 +53,39 @@ export class DockerCGroupV1Detector implements Detector {
...
+            .split('/')
+            .filter(s => s.length >= this.CONTAINER_ID_LENGTH)[0];
```
blumamir, line 78: instead of `filter(...)[0]`, you can use `find(s => s.length >= this.CONTAINER_ID_LENGTH)` which does the same :)

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
```diff
@@ -51,21 +53,39 @@ export class DockerCGroupV1Detector implements Detector {
...
+
+      // If this code is reached - we then check for the V2_PATH
+      rawData = await DockerDetector.readFileAsync(
+        this.DEFAULT_CGROUP_V2_PATH,
+        this.UTF8_UNICODE
+      );
+
+      splitData = rawData.trim().split('\n');
+      for (let str of splitData) {
+        if (str.includes(this.HOSTNAME)) {
```
blumamir, line 75: Suggestion: Instead of using `if` inside the for loop which adds another indentation and branching, you can use `filter`:
```js
rawData.trim().split('\n').filter(str => str.includes(this.HOSTNAME))`
```

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: It is - if they update - they will have to change the detector name - but this name is scalable and would not need a future change.

I shall update the Read.me

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: Makes sense

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: I thought about it and it seemed natural to do that but overall it was almost the same code but I agree - it is more readable - will do.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: The assumption is and I have confirmed that the file would exist - it would just not have the id in it. and then it will fall back to the v2 implementation.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: sounds good.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: makes sense.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: makes sense.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
svrnm: why not ContainerDetector if it is compatible with containerd as well?

abhee11: @svrnm makes sense - I shall rename it to that.

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
abhee11: @blumamir I believe I have addressed all commets - please let me know if there is anything else .

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
blumamir: Great, thank you for validating.
Can I please ask you to add this insight as comment, so future developers looking at this code will know that it's intended, and confirmed to behave in the right way?

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -42,30 +44,56 @@ export class DockerCGroupV1Detector implements Detector {
...
+    const rawData = await ContainerDetector.readFileAsync(
+      this.DEFAULT_CGROUP_V2_PATH,
+      this.UTF8_UNICODE
+    );
+    const str = rawData
+      .trim()
+      .split('\n')
+      .find(s => s.includes(this.HOSTNAME));
+    const containerIdStr =
+      str && str.split('/').find(s => s.length >= this.CONTAINER_ID_LENGTH);
```
blumamir, line 78: nit: instead of `str && str.split(...` you can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to get the same effect in a cleaner syntax:
```js
const containerIdStr = str?.split('/').find(s => s.length >= this.CONTAINER_ID_LENGTH);
```

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -42,30 +44,56 @@ export class DockerCGroupV1Detector implements Detector {
...
+    if (containerIdStr) {
+      return containerIdStr.substring(
+        containerIdStr.length - this.CONTAINER_ID_LENGTH
+      );
+    }
+    return undefined;
```
blumamir, line 84: same here, if you choose, you can write this entire section as:
```js
return containerIdStr?.substring(containerIdStr.length - this.CONTAINER_ID_LENGTH);
```

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
```diff
@@ -0,0 +1,139 @@
...
+import {
+  assertContainerResource,
+  assertEmptyResource,
+} from '@opentelemetry/contrib-test-utils';
+
+import { ContainerDetector } from '../src';
+
+describe('ContainerDetector', () => {
+  let readStub;
+  const correctCgroupData =
```
blumamir, line 30: maybe add V1 if we now also have V2?

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklm';
+  const correctCgroupV2Data =
+    'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklm/hostname';
+
+  const wrongCgroupV2Data =
```
blumamir, line 35: nit: comment why this is wrong (contains text "host" where it should say "hostname"). Might be hard to catch up at first glance.

blumamir (approved):
LGTM. Thanks
All open threads are optional. Feel free to resolve or implement

detectors/node/opentelemetry-resource-detector-docker/src/detectors/DockerDetector.ts
svrnm: The question is, do we want end-users to swap out that detector every time they move from one container engine to another? Maybe they are not even aware that things got switched out because that's not the developers decision.

The detection provided here works for at least 2 engines we have tests, we could try out podman (and others) as well.

Actually it would be great to have this fixed via OCI spec (see this issue: https://github.com/opencontainers/runtime-spec/issues/1105), but that's a long standing issue.

blumamir: > The question is, do we want end-users to swap out that detector every time they move from one container engine to another? Maybe they are not even aware that things got switched out because that's not the developers decision.
> 
> The detection provided here works for at least 2 engines we have tests, we could try out podman (and others) as well.
> 
> Actually it would be great to have this fixed via OCI spec (see this issue: [opencontainers/runtime-spec#1105](https://github.com/opencontainers/runtime-spec/issues/1105)), but that's a long standing issue.

I think both options are valid and have pros and cons.
Since it's already implemented for general container, I support keeping it as it is now :)

Thank you for the reference to the OCI spec

abhee11:
@blumamir I addressed the rest of the comments as well - please merge when you get a chance

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
abhee11: done

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
abhee11: done

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
abhee11: done

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
abhee11: done

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
```diff
@@ -0,0 +1,139 @@
...
+} from '@opentelemetry/contrib-test-utils';
+
+import { ContainerDetector } from '../src';
+
+describe('ContainerDetector', () => {
+  let readStub;
+  const correctCgroupV1Data =
+    'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklm';
+  const correctCgroupV2Data =
+    'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklm/hostname';
```
Rauno56, line 33: Do you have any real life examples of full files? I think it would be worth to put them to test as well.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -42,30 +44,59 @@ export class DockerCGroupV1Detector implements Detector {
...
+      this.DEFAULT_CGROUP_V2_PATH,
+      this.UTF8_UNICODE
+    );
+    const str = rawData
+      .trim()
+      .split('\n')
+      .find(s => s.includes(this.HOSTNAME));
+    const containerIdStr = str
+      ?.split('/')
+      .find(s => s.length >= this.CONTAINER_ID_LENGTH);
```
Rauno56, line 79: So we are looking for any string that has the correct length?
I think I've seen examples of `/proc/self/cgroup` that have truncated container ID or that have container id with a prefix(`docker-`) or suffix(`.slice`).

See some pseudo-examples(the length isn't checked in that implementation, and I don't know whether I would even, but you get the idea): https://github.com/signalfx/splunk-otel-js/blob/main/test/resource.test.ts

Rauno56: I think contents for both `cgroup` and `mountinfo` files is a implementation detail of whatever calls the container runtime APIs.

abhee11: @Rauno56 Sorry - was out for a few days - so in terms of changes requested at this time - is to add more test cases around the files itself as in the examples right ?

Current implementation is check for keyword "hostname" and 64 length str - which may need to be updated depends on other containers.

Rauno56: I actually totally missed the fact that you are filtering the rows to find the one with "hostname". It doesn't have to be perfect. Sorry for unproductive feedback.

What about `>=`? will it not be invalid id if the lenght = 67?

Rauno56:
Thanks for working on this. Please add tests with files having multiple lines and we'll be golden!

abhee11:
@Rauno56 Changed existing test to take multiple lines as input - I think we should be covered - thank you for the review - feel free to merge when you can.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
Rauno56: @abhee11, your thoughts on my last comment?

abhee11: @Rauno56 Honestly - I dont have a good answer for this as we breaking down the string with "/" delimeter. 
We just return 64 length substring, it does seem a little crass but this is the best logic we could comeup with.

abhee11: I will keep an eye out in case we could make this logic more robust in the future, but current implementation in dockerv1 and how aws is doing is similar

Rauno56: Sorry, I meant why we do we accept ids that are longer than 64 chars?

Rauno56 (approved):
Thanks for bearing with the long back-and-forth! I don't have any blockers.
Only curious about the length check.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
abhee11: @Rauno56 Valid, I think it is just the case we "consider everything >=64 " ( we do not expect to get anything more than 64 ) but if we do - we just return substr of that -- but i see your point and can change it to "==" check

abhee11: @Rauno56 I have made the "===" change as it made sense. Once it does pass the check please merge it if everything looks good.

abhee11:
@Rauno56 I strongly believe in code reviews like these ones, may be a bit long but it's a good learning and quality check. Thank you for an engaged review.

Issue #1186, opened 2022-09-19
@opentelemetry/auto-instrumentations-web not ES5 compatible


mdbudnick:
### What version of OpenTelemetry are you using?
0.30.0

### What version of Node are you using?
14.20.0

### What did you do?

I've installed "@opentelemetry/auto-instrumentations-web": "^0.30.0",
and in my project use `import { getWebAutoInstrumentations } from '@opentelemetry/auto-instrumentations-web'` with an OTLPExporter.

It works fine except for IE 10 and 11 because the module code in my client js is not ES5-compatible:
```
function(e, t, r) {
            "use strict";
            Object.defineProperty(r, "__esModule", {
                value: !0
            }), r.getWebAutoInstrumentations = void 0;
            const n = e("@opentelemetry/api"),
                o = e("@opentelemetry/instrumentation-document-load"),
                s = e("@opentelemetry/instrumentation-fetch"),
                i = e("@opentelemetry/instrumentation-user-interaction"),
                a = e("@opentelemetry/instrumentation-xml-http-request"),
                l = {
                    "@opentelemetry/instrumentation-document-load": o.DocumentLoadInstrumentation,
                    "@opentelemetry/instrumentation-fetch": s.FetchInstrumentation,
                    "@opentelemetry/instrumentation-user-interaction": i.UserInteractionInstrumentation,
                    "@opentelemetry/instrumentation-xml-http-request": a.XMLHttpRequestInstrumentation
                };
            r.getWebAutoInstrumentations = function(e = {})
}
```
i.e. default values are not allowed for arguments: `function(e = {})`

### What did you expect to see?
ES5 compliant code that works in IE10 and 11. From what I have [found in this project](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/557), ES5 compliance can ([should?](https://github.com/open-telemetry/opentelemetry-js/pull/1368)) be expected.

### What did you see instead?
An error in my run.js: `Expected ')'` after the e in `function(e = {})` because default argument values were introduced in ES6

### Additional context

It is possible I am not doing something obvious, I am a backend engineer setting up OTel tracing in our company.

mdbudnick:
Also this https://github.com/open-telemetry/opentelemetry-js/pull/1368

mdbudnick:
Created https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1187

dyladan:
[...]

PR #1187, opened 2022-09-19
feat: Compile metapackages/auto-instrumentations-web to es5

mdbudnick:
- [autoinstrumentations-web does not work in IE10 or IE11](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1186)

- Change tsconfig compilerOptions.target to es5 

## Checklist

- [x] Ran `npm run test:browser` for the edited package(s) on the latest commit if applicable.

### DIFF
--- a/metapackages/auto-instrumentations-web/tsconfig.json
+++ b/metapackages/auto-instrumentations-web/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "../../tsconfig.base",
+  "extends": "../../tsconfig.es5",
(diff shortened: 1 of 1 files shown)


dyladan:
not all of the dependencies support es5. at least the web document load plugin does not, so adding es5 support to this package likely doesn't solve what you think it does. This package also doesn't really have any code, it just depends on other packages as an installer helper. es5 support may be possible but i think it would be more work and touch a lot more packages.

PR #1191, opened 2022-09-21
test: fix tests for restify >=5

### LINKED ISSUE #1132: TS2430: Interface 'Request' incorrectly extends interface 'IncomingMessage' when using getNodeAutoInstrumentations

Rauno56:

Restify instrumentation required some mainenance.

There was a previous effort by @jscherer92: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1133
I built on top of that PR and left what Justin had done(thanks!) intact so the attribution should be intact in the end.

Fixes #1132
Closes #1133

1. updated the main package version
1. removed type exports
1. fixed tests for newer versions
1. added TAV
1. turned off tests on `node@18`: https://github.com/restify/node-restify/pull/1906

## Checklist


plugins/node/opentelemetry-instrumentation-restify/test/restify.test.ts
```diff
@@ -73,24 +87,29 @@ const getHandler: restify.RequestHandler = (req, res, next) => {
...
 const throwError: restify.RequestHandler = (req, res, next) => {
-  throw new Error('NOK');
+  throw new AppError('NOK');
+};
+const returnError: restify.RequestHandler = (req, res, next) => {
+  next(new AppError('NOK'));
 };
 
 const createServer = async (setupRoutes?: Function) => {
+  const restify = require('restify');
```
blumamir, line 97: This definition overruns the one above:
```js
import type * as restify from 'restify';
```

I wonder if the `restify` type import is used anywhere.
Can we `import * as restify from 'restify'` after we setup the instrumentation?

blumamir (approved):

plugins/node/opentelemetry-instrumentation-restify/test/restify.test.ts
Rauno56: The types are used at the top level: `useHandler`, etc. Import declarations can only be used at the top level so I cannot replace it as is.

The reason I changed it was to make sure the instrumentation is always at the correct state(enabled or disabled) before the require, but I guess I can bring it back to the top level, yes.

Will do.

Rauno56: Done. PTAL, @blumamir

vmarchaud (approved):

Issue #1190, opened 2022-09-20
(instrumentation-mongodb)  - Cannot find name 'Document'.


97amarnathk:
### What version of OpenTelemetry are you using?
@opentelemetry-api 1.2.0

### What version of Node are you using?
14.17.1

### What did you do?
I have a private nodejs service to which I am trying to add openTelemetry
1. install open telemetry and http, grpc instrumentation (this is working fine)
2. installed mongodb instrumentation
```
yarn add @opentelemetry/instrumentation-mongodb                                   
```
3. built my node module
```
yarn build
```

### What did you expect to see?
My module should have been built successfully

### What did you see instead?
Typescript fails to build it, with error - 
```
../../node_modules/@opentelemetry/instrumentation-mongodb/build/src/types.d.ts:86:27 - error TS2304: Cannot find name 'Document'.

86     command(ns: any, cmd: Document, options: undefined | unknown, callback: any): void;
                             ~~~~~~~~


Found 1 error.
```

### Additional context

**Working code**
```typescript
import { Tracer } from '@opentelemetry/api';
import { JaegerExporter } from '@opentelemetry/exporter-jaeger';
import { registerInstrumentations } from '@opentelemetry/instrumentation';
import { GrpcInstrumentation } from '@opentelemetry/instrumentation-grpc';
import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
import { MongoDBInstrumentation } from '@opentelemetry/instrumentation-mongodb';
import { Resource } from '@opentelemetry/resources';
import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base';
import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';

export function initTracer(serviceName: string): Tracer {
  const resource = Resource.default().merge(
    new Resource({
      [SemanticResourceAttributes.SERVICE_NAME]: serviceName,
    })
  );

  const provider = new NodeTracerProvider({
    resource: resource,
  });

  const exporter = new JaegerExporter({ endpoint: 'http://localhost:14268/api/traces' });
  const processor = new BatchSpanProcessor(exporter);

  provider.addSpanProcessor(processor);
  provider.register();

  registerInstrumentations({
    instrumentations: [new HttpInstrumentation(), new GrpcInstrumentation(), new MongoDBInstrumentation()],
  });

[...]

PR #1194, opened 2022-09-21
fix: remove unneeded type exports in mongodb instrumentation

Flarna:

fixes: #1190 

Avoid leaking types from mongodb instrumentation by splitting types.ts into a public and internal file.
Added type Document to avoid that lib dom needs to be added in build.

~~change typescript config to exclude lib dom on default and add it only where needed (web plugins and AWS SDK).~~

## Checklist


Flarna:
reverted the commit to remove lib dom from build. This worked fine in my local environment but in hoisting is done different so lib dom is needed everywhere.

blumamir (approved):

Rauno56 (approved):

vmarchaud (approved):

sytranvn:
I got this error with version 0.36.0
```
npm run build
> tsc

node_modules/@opentelemetry/instrumentation-mongodb/build/src/types.d.ts:49:27 - error TS2304: Cannot find name 'Document'.

49     command(ns: any, cmd: Document, options: undefined | unknown, callback: any): void;
                             ~~~~~~~~


Found 1 error in node_modules/@opentelemetry/instrumentation-mongodb/build/src/types.d.ts:49

```
Screenshot 2023-07-26 at 10 13 21

PR #1199, opened 2022-09-26
feat(pg): add requireParentSpan option

ryan-codaio:

The current pg autoinstrumentation creates spans for any postgres connection or query, even if this is not part of an active context. This generates a large number of extraneous spans that we don't care about, and opentelemetry currently doesn't have good support for filtering these out.

This PR adds a `requireParentTrace` configuration option that will prevent the autoinstrumentation from generating root-level traces. Similar options already exist in [instrumentation-http](https://github.com/open-telemetry/opentelemetry-js/blob/ce7e98fd983a72c0e82450654e2f838e30b4cf95/experimental/packages/opentelemetry-instrumentation-http/src/types.ts#L120-L123), [instrumentation-ioredis](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/d3cb60d28e186b42f0c17ab48df8757555bbe6e0/plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts#L34), [instrumentation-mongoose](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b35277bb5fc66910e8942bc0b64347b68ecffa26/plugins/node/instrumentation-mongoose/src/types.ts#L61), and others.

The implementation is based on the [implementation in opentelemetry-instrumentation-http](https://github.com/open-telemetry/opentelemetry-js/blob/597ea98e58a4f68bcd9aec5fd283852efe444cd6/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L665-L683).

## Checklist


dyladan:
Looks like the CLA isn't signed. Please sign it.

>how can I be useful?

code reviews are always helpful

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -640,4 +640,33 @@ describe('pg', () => {
...
+      assert.strictEqual(spans.length, 0);
+      await connClient.end();
+    });
+
+    it('should not generate traces for client.query(text, callback) when requireParentSpan=true', done => {
+      client.query('SELECT NOW()', (err, res) => {
+        assert.strictEqual(err, null);
+        assert.ok(res);
+        const spans = memoryExporter.getFinishedSpans();
+        console.log(spans);
```
henrinormak, line 666: Probably accidental?

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -640,4 +640,33 @@ describe('pg', () => {
       });
     });
   });
+
+  describe('Instrumentation with requireParentSpan', () => {
```
henrinormak, line 644: For completeness, I would move these tests to #client.connect block and replicate them in #client.query block (ofc with querying the client instead of just connecting).

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -187,7 +181,12 @@ export class PgInstrumentation extends InstrumentationBase {
               params
             );
           } else {
-            span = utils.handleTextQuery.call(this, plugin.tracer, query);
+            span = utils.handleTextQuery.call(
+              this,
+              plugin.tracer,
+              plugin.getConfig() as PgInstrumentationConfig,
```
henrinormak, line 187: As I was going over the code in this instrumentation, I noticed that the config requires casting here, we could remove that need by doing something similar to some of the other instrumentations - https://github.com/open-telemetry/opentelemetry-js-contrib/blob/6684b56b8043f094b95fc3c1ce5e5599e694bad4/plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts#L50-L56

This would remove some annoying type casts and replace it with a singular one. Not super necessary to be done, but would be nice to clean up the codebase a bit?

plugins/node/opentelemetry-instrumentation-pg/src/types.ts
```diff
@@ -40,6 +40,9 @@ export interface PgInstrumentationConfig extends InstrumentationConfig {
    * @default undefined
    */
   responseHook?: PgInstrumentationExecutionResponseHook;
+
+  /** Require that is a parent span to create new spans. Defaults to false. */
+  requireParentSpan?: boolean;
```
henrinormak, line 45:
```suggestion

  /**
   * If true, requires a parent span to create new spans.
   *
   * @default false
   */
  requireParentSpan?: boolean;
```

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -58,19 +62,41 @@ export function getConnectionString(params: PgClientConnectionParams) {
...
-      [SemanticAttributes.DB_NAME]: client.connectionParameters.database, // required
-      [SemanticAttributes.DB_SYSTEM]: DbSystemValues.POSTGRESQL, // required
-      [SemanticAttributes.DB_CONNECTION_STRING]: jdbcString, // required
-      [SemanticAttributes.NET_PEER_NAME]: client.connectionParameters.host, // required
-      [SemanticAttributes.NET_PEER_PORT]: client.connectionParameters.port,
-      [SemanticAttributes.DB_USER]: client.connectionParameters.user,
-    },
+    attributes,
+  });
+}
```
henrinormak, line 83: Seems like some of the exported utils have dedicated tests as well, could also cover this one?

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
ryan-codaio: Yes, thanks for the catch

henrinormak (approved):

Rauno56 (commented):
Other instrumentations call this feature `requireParentSpan` - see dataloader, ioredis or redis for example. If I'm not mistaken, and it is the same feature, we should call it with the same name for ease of use.

We often implement it slightly different as well - short-circuiting into the original function if the option is enabled and there's no parent. Not saying this implementation should be changed, but just interested in your thoughts about whether it would be viable option here and if there's any benefits going with either approach.

henrinormak:
I would also prefer that naming and approach, although I don't have a strong preference, TBH, when I read the PR description, I mistakenly thought it was named like this in the referred other instrumentations as well.

vmarchaud (approved):

ryan-codaio:
> Other instrumentations call this feature requireParentSpan - see dataloader, ioredis or redis for example. If I'm not mistaken, and it is the same feature, we should call it with the same name for ease of use.

I'm a little confused -- it's called `requireParentSpan` here as well, right? Not sure if I mistyped something somewhere

> We often implement it slightly different as well - short-circuiting into the original function if the option is enabled and there's no parent. Not saying this implementation should be changed, but just interested in your thoughts about whether it would be viable option here and if there's any benefits going with either approach.

I'm new to opentelemetry, but my understanding is that the approach I used here (copied from instrumentation-http) is creating a placeholder span with an invalid context so that it is eventually discarded. However, child spans can be attached to this placeholder span. That way, if there are some child spans generated as a result of some `pg` operation when there was no parent span present, they are all discarded. This seems preferable to having some child spans show up for internal operations (e.g. DNS lookups) when we are trying to suppress pg spans with no parent span. However, I might be wrong here, so please correct me if that doesn't seem right.

vmarchaud:
> here as well, right? Not sure if I mistyped something somewhere

I think it was because you named the PR `add requiredParentTrace` (which i renamed afterwards)

> However, child spans can be attached to this placeholder span. That way, if there are some child spans generated as a result of some pg operation when there was no parent span present, they are all discarded

I believed Rauno was refering to the usage of the non-recording span, which is used in the http instrumentation to still propagate context for outgoing requests without creating a span locally, i'm not sure we need this as we don't propagate the context to pg itself but i agree that it doesn't hurt anything to do so

Rauno56 (approved):

PR #1200, opened 2022-09-26
chore: update lerna

Rauno56:

Dropping support for node<14 now let's us upgrade tooling. Not statistically significant, but bootstrapping deps seems to take less time now.

Updated `lerna` and `lerna-changelog`. There should be no relevant breaking changes for us.


legendecas (approved):

pichlermarc (approved):

Flarna (approved):

Flarna:
Is it possible that this update caused that compile error during bootstrap are no longer seen in logs?

see e.g. [here](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/3439875168/jobs/5742172446)

having only
```
> npm run compile

lerna ERR! lifecycle "prepare" errored in "@opentelemetry/resource-detector-instana", exiting 1
```

isn't that helpful to provide a fix.

PR #1202, opened 2022-09-27
chore: add dataloader to release-please manifest and fix mongoose

Rauno56:

We forgot to add instrumentation-dataloader to the manifest and instrumentation-mongoose had a wrong path.


Flarna:
I'm still confused about which files need to be touched to keep release please happy. I thought that `.release-please-manifest.json` is updated by release please and users have to add their module only in `release-please-config.json`?

Rauno56:
I feel you. I'm constantly confused myself as well.

Adding packages to manifest is only important because we don't release `1.0.0` as the first version. In my mind:

- manifest -> release please's state;
- config -> our preferences for release please.

Flarna (approved):

PR #1206, opened 2022-09-27
chore: remove outdated entries from RP manifest

Flarna:

removed wrong dataloader and mongoose entries from release-please manifest. sort entries in release-please config.

Seems I reviewed #1202 not good enough as this ended up with two entries instead one.

fyi @Rauno56


Rauno56:
I thought I removed those. Thanks!

Always having my back, @Flarna!

Rauno56 (approved):

PR #1207, opened 2022-09-27
chore: implement CI job to validate Release Please's config and manifest

Rauno56:

I'm often confused what needs to be added where regarding Release Please.

Added a reminder for myself and others.


scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+
+const errors = [];
+const logErrorIf = (condition, errorMessage) => {
+	if (condition) {
+		errors.push(errorMessage);
+	}
+};
+const PROJECT_ROOT = path.resolve();
+const readJson = (path) => {
+	return JSON.parse(readFileSync(path));
```
Flarna, line 24: hiding the imported `path` here is confusing

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+	makes sure all those lists are in sync. If not, guides you to do the correct thing
+*/
+
+const errors = [];
+const logErrorIf = (condition, errorMessage) => {
+	if (condition) {
+		errors.push(errorMessage);
+	}
+};
+const PROJECT_ROOT = path.resolve();
```
Flarna, line 22:
```suggestion
const PROJECT_ROOT = process.cwd();
```
looks more readable to me

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+
+	- lerna packages,
+	- release please's manifest, and
+	- release please's config, and
+
+	makes sure all those lists are in sync. If not, guides you to do the correct thing
+*/
+
+const errors = [];
+const logErrorIf = (condition, errorMessage) => {
```
Flarna, line 17:
```suggestion
function logErrorIf(condition, errorMessage) {
```
or maybe `recordErrorIf`

There are other places where functions are written as `const foo = () => {`. Is there a reason to prefer this over the function keyword?

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+		return [relativeLocation, { ...pkg, relativeLocation: path.relative(PROJECT_ROOT, pkg.location) }];
+	})
+);
+const manifest = readJson('.release-please-manifest.json');
+const config = readJson('release-please-config.json');
+
+const manifestPackages = new Set(Object.keys(manifest));
+const configPackages = new Set(Object.keys(config.packages));
+
+console.log('lerna packages', );
```
Flarna, line 47: seems second arg is missing here

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+console.log('manifest packages', manifestPackages);
+console.log('config packages', configPackages);
+
+packageList.forEach(({ relativeLocation }) => {
+	logErrorIf(
+		!manifestPackages.has(relativeLocation),
+		`Could not find ${relativeLocation} from manifest. If you are adding a new package. Add following to .release-please-manifest.json
+  "${relativeLocation}": "0.0.1",`);
```
Flarna, line 55: I thought we start usually with `0.1.0`?

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+const getProcessOutput = (cmd, args) => {
+	const result = childProcess.spawnSync(cmd, args);
+	assert(!result.error, result.error);
+	return result.stdout.toString('utf8');
+}
+
+const lernaList = JSON.parse(
+getProcessOutput('lerna', ['list', '--json'])
+);
+const packageList = new Map(
```
blumamir, line 35: It seems we only use the `relativeLocation` for these packages.
Can we use `Set` instead, this will also be consistent with the other variables

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+const configPackages = new Set(Object.keys(config.packages));
+
+console.log('lerna packages', );
+console.log('manifest packages', manifestPackages);
+console.log('config packages', configPackages);
+
+packageList.forEach(({ relativeLocation }) => {
+	logErrorIf(
+		!manifestPackages.has(relativeLocation),
+		`Could not find ${relativeLocation} from manifest. If you are adding a new package. Add following to .release-please-manifest.json
```
blumamir, line 54: Is the wording "from manifest" correct? Shouldn't it be "in manifest"?

scripts/check-release-please.mjs
Rauno56: The first release is then `0.2.0`. See [dataloader's first release](https://github.com/open-telemetry/opentelemetry-js-contrib/releases/tag/instrumentation-dataloader-v0.2.0).

Haven't validated, but if it bumps minor then `0.0.1` -> `0.1.0`.

scripts/check-release-please.mjs
Rauno56: I tend to use arrow-functions whenever I don't need `this`. Especially with pure functions. I wouldn't do this change.

scripts/check-release-please.mjs
Rauno56: Renamed the argument. :+1:

scripts/check-release-please.mjs
Rauno56: :man_facepalming:

scripts/check-release-please.mjs
```diff
@@ -0,0 +1,82 @@
...
+	return result.stdout.toString('utf8');
+}
+
+const lernaList = JSON.parse(
+getProcessOutput('lerna', ['list', '--json'])
+);
+const packageList = new Map(
+	lernaList.map((pkg) => {
+		const relativeLocation = path.relative(PROJECT_ROOT, pkg.location);
+		return [relativeLocation, { ...pkg, relativeLocation: path.relative(PROJECT_ROOT, pkg.location) }];
```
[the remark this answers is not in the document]
Rauno56: True. Wasn't sure what I need from the output while I started the script and later didn't critically check anymore. Thanks!

scripts/check-release-please.mjs
Rauno56: Makes sense. Made everything into Set.

scripts/check-release-please.mjs
Rauno56: I think you're right. Reworded it.

Flarna (approved):

dyladan (approved):

PR #1210, opened 2022-09-28
fix(express): use the same clock for span start and end

### LINKED ISSUE #1193: Express spans showing impossible duration_ms
0.33.0

`16 lts`

### What did you do?
Instrumented node. I use the following to set up my tracing:
```ts
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-proto";
import { PrismaInstrumentation } from "@prisma/instrumentation";

const traceExporter = new OTLPTraceExporter();

const sdk = new NodeSDK({
    traceExporter,
    instrumentations: [
        getNodeAutoInstrumentations(),
        new PrismaInstrumentation(),
    ],
});

// eslint-disable-next-line @typescript-eslint/no-floating-promises
sdk.start().then(() => {
    // eslint-disable-next-line no-console
    console.log(`✅ OTLP ready`);
});
```
Reasonable tracing

### What did you see instead?
Screenshot from 2022-09-21 12-45-04

### LINKED ISSUE #1209: Inconsistent start and end time, startTime > endTime issues with express

```
$ npm list @opentelemetry/instrumentation-express   

test-obs@0.0.0 /Users/alexandre/Dev/Alteia/test-observability/test-obs
└─┬ @opentelemetry/auto-instrumentations-node@0.32.1
  └── @opentelemetry/instrumentation-express@0.31.1
```

```
$ node -v  
v16.17.0
```

Running on a M1 MacBook Pro

### What did you do?

I am running a simple nest app to familiarize with open telemetry instrumentation and tooling.
I have added the auto-instrumentation as mentioned in the docs, and running a custom nest app.
After some times (especially If the Macbook go to sleep) I start to see such warning in the output : 

`Inconsistent start and end time, startTime > endTime [ 1664350641, 619671106 ] [ 1664350641, 619333687 ]`

Jaeger traces looks completely wrong
image

Here is the jaeger traces before the problem is appearing

image

### What did you see instead?

The timing for the express span is getting completely off.

### Additional context

I did a quick debugging session and source code analysis and I am very suspicious about the following code : 
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1c450f5fe31437aa7157ff6bb8602d330dc3b429/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L266

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1c450f5fe31437aa7157ff6bb8602d330dc3b429/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L274

It seems it is using a performance timer for the span end time, which may somehow differ from the wall time (https://github.com/open-telemetry/opentelemetry-js/issues/852 looks related), whereas other instrumentations does not override the span end time and works correctly in my case.

dyladan:
Fixes #1193 
Fixes #1209 

Express instrumentation does not pass a time to span start but does pass a time to span.end, which results in undefined behavior. Previous to https://github.com/open-telemetry/opentelemetry-js/pull/3134 this was not a problem as the span always used the same hrTimer that we expose in core. Now that we anchor clocks, it is possible for these clocks to be out of sync.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
@@ -15,7 +15,6 @@
  */
 
 import {
-  hrTime,
   setRPCMetadata,
   getRPCMetadata,
   RPCType,
--- a/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
@@ -14,11 +14,7 @@
  * limitations under the License.
  */
 
-import {
-  setRPCMetadata,
-  getRPCMetadata,
-  RPCType,
-} from '@opentelemetry/core';
+import { setRPCMetadata, getRPCMetadata, RPCType } from '@opentelemetry/core';
 import { trace, context, diag, SpanAttributes } from '@opentelemetry/api';
 import type * as express from 'express';
 import {
--- a/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
@@ -263,22 +263,19 @@ export class ExpressInstrumentation extends InstrumentationBase<
           );
         }
 
-        const startTime = hrTime();
         let spanHasEnded = false;
-        // If we found anything that isnt a middleware, there no point of measuring
-        // their time since they dont have callback.
         if (
           metadata.attributes[AttributeNames.EXPRESS_TYPE] !==
           ExpressLayerType.MIDDLEWARE
         ) {
-          span.end(startTime);
+          span.end();
           spanHasEnded = true;
         }
         // listener for response.on('finish')
         const onResponseFinish = () => {
           if (spanHasEnded === false) {
             spanHasEnded = true;
-            span.end(startTime);
+            span.end();
           }
         };
         // verify we have a callback


Flarna (approved):

blumamir (approved):

gregolsen:
Hey @dyladan - first off thank you so much for your work on the anchored clock to mitigate the time drift!
We have recently tried to rollout 1.7.0/0.33.0 OTEL instrumentation and faced a similar issue as the ones this PR fixes – negative span durations caused by start time being higher than the end time (negative span duration probably results in out of bound error and is shown as a very long duration in Honeycomb instead):
image
I understand that this PR address the issue with node.js instrumentation only.
Do you think there could be a similar issue but in the web instrumentation?

dyladan:
> Hey @dyladan - first off thank you so much for your work on the anchored clock to mitigate the time drift! We have recently tried to rollout 1.7.0/0.33.0 OTEL instrumentation and faced a similar issue as the ones this PR fixes – negative span durations caused by start time being higher than the end time (negative span duration probably results in out of bound error and is shown as a very long duration in Honeycomb instead): image I understand that this PR address the issue with node.js instrumentation only. Do you think there could be a similar issue but in the web instrumentation?

There definitely could be a similar issue but I'm surprised to see it both in XHR and Fetch. Is that happening on all spans or just some?

gregolsen:
Good question @dyladan , looks like in our case it happened on a significant portion of the spans but not all of them:
image

dyladan:
All of the bad spans are coming from the same instrumentations? Do you have any other instrumentations? Are they reporting good or bad spans?

Flarna:
fetch is setting end time also manually see
https://github.com/open-telemetry/opentelemetry-js/blob/afe065772360d3acec8b8fc15e20536fd896c6c6/experimental/packages/opentelemetry-instrumentation-fetch/src/fetch.ts#L99-L116
https://github.com/open-telemetry/opentelemetry-js/blob/afe065772360d3acec8b8fc15e20536fd896c6c6/experimental/packages/opentelemetry-instrumentation-fetch/src/fetch.ts#L281-L289

gregolsen:
> All of the bad spans are coming from the same instrumentations? Do you have any other instrumentations? Are they reporting good or bad spans?

Majority of the "negative" spans were coming from XHR and Fetch instrumentations. Notable, document load wasn't affected at all as it probably just uses the timings provided by the browser resource timing.

gregolsen:
Relevant code for the XHR instrumentation:
https://github.com/open-telemetry/opentelemetry-js/blob/610808d3b64b9f660f4dd640ad961b8c9f67be66/experimental/packages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts#L434-L441

I think both Fetch and XHR have to store the timestamp and then schedule a timeout to wait for the network events to be ready. There's a comment explaining that:

>      // the timeout is needed as observer doesn't have yet information
>      // when event "load" is called. Also the time may differ depends on
>      // browser and speed of computer

MSNev (approved):

PR #1211, opened 2022-09-28
feat: add requireParentSpan option to pg instrumentation

henrinormak:

pg instrumentation currently lacks the option to require a parent span, which can produce quite a bit of noise, especially in situations where there are things like health-checks being done against DBs etc.

Adds a new option (keeping the default to false so that this change does not alter existing deployments) `requireParentSpan` to the config of pg instrumentation. If set to true, it does not create spans unless there is a parent span that it can attach to


dyladan:
This is a possible duplicate of #1199 can you see if that solves your issue and help with that PR?

dyladan:
The best way to help would be to review the PR

PR #1219, opened 2022-10-07
fix: restrict hashed package.json files

Rauno56:

Seeing the `hashFiles` taking minutes makes me think it searches the `node_modules` dirs as well.
I have experienced and debugged it for test builds as well.
The examples using `action/cache` hash package-lock.json files, which would generally make sense if we checked our package-lock files in.
Why don't we? I've forgotten, but it has been discussed multiple times.

Restrict the globs to only hash package.json.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,7 +35,7 @@ jobs:
             plugins/node/*/node_modules
             plugins/web/*/node_modules
             propagators/*/node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/package.json') }}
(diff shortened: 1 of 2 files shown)


blumamir:
I opened a random CI run [here](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/3252920009/jobs/5339665708)

image

From what I understand, the CI log shows the key (calculated with previous hash as `hashFiles('**/package.json')`), as "Linux-16-6bbd533bd80a63cf309894ff379bc9fb331e6d4bdcc4150df233f13da78e5283" and it took 0s.

Am I missing something?

Issue #1176, opened 2022-09-15
Metrics instrumentation mysql


haddasbronfman:
We want to start to implement metrics collection in the instrumentations like other SDKs already did.
I would like to start with MySQL (according to:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md)

If that is ok with you, please assign this issue to me :)

PR #1220, opened 2022-10-07
feat(mysql): started working on mysql metrics

haddasbronfman:

https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1176

Add the metric `db.client.connections.usage` to mysql according to this: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md#connection-pools

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql/package.json
+++ b/plugins/node/opentelemetry-instrumentation-mysql/package.json
@@ -50,6 +50,7 @@
     "@opentelemetry/context-async-hooks": "^1.3.1",
     "@opentelemetry/contrib-test-utils": "^0.32.0",
     "@opentelemetry/sdk-trace-base": "^1.3.1",
+    "@opentelemetry/sdk-metrics": "^0.32.0",
     "@types/mocha": "7.0.2",
     "@types/node": "16.11.21",
     "@types/sinon": "10.0.13",
@@ -63,6 +64,7 @@
     "typescript": "4.3.5"
   },
   "dependencies": {
+    "@opentelemetry/api-metrics": "^0.32.0",
     "@opentelemetry/instrumentation": "^0.32.0",
     "@opentelemetry/semantic-conventions": "^1.0.0",
     "@types/mysql": "2.15.19"
--- a/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
@@ -36,6 +36,8 @@ import type * as mysqlTypes from 'mysql';
 import { MySQLInstrumentationConfig } from './types';
 import { getConnectionAttributes, getDbStatement, getSpanName } from './utils';
 import { VERSION } from './version';
+import { UpDownCounter } from '@opentelemetry/api-metrics';
+import { MeterProvider } from '@opentelemetry/sdk-metrics';
 
 type formatType = typeof mysqlTypes.format;
 
@@ -50,9 +52,27 @@ export class MySQLInstrumentation extends InstrumentationBase<
   static readonly COMMON_ATTRIBUTES = {
     [SemanticAttributes.DB_SYSTEM]: DbSystemValues.MYSQL,
   };
+  private _connectionsUsage!: UpDownCounter;
 
   constructor(config?: MySQLInstrumentationConfig) {
     super('@opentelemetry/instrumentation-mysql', VERSION, config);
+    this._setMetricInstruments();
+  }
+
+  override setMeterProvider(meterProvider: MeterProvider) {
+    super.setMeterProvider(meterProvider);
+    this._setMetricInstruments();
+  }
+
+  private _setMetricInstruments() {
+    this._connectionsUsage = this.meter.createUpDownCounter(
+      'db.client.connections.usage',
+      {
+        description:
+          'The number of connections that are currently in state described by the state attribute.',
+        unit: '{connections}',
+      }
+    );
   }
 
   protected init() {
@@ -100,10 +120,52 @@ export class MySQLInstrumentation extends InstrumentationBase<
           this._unwrap(moduleExports, 'createConnection');
           this._unwrap(moduleExports, 'createPool');
           this._unwrap(moduleExports, 'createPoolCluster');
-        }
+        },
+        // [
+        //   new InstrumentationNodeModuleFile<any>(
+        //     'mysql/lib/connection.js',
+        //     ['2.*'],
+        //     (moduleExports: any, moduleVersion?: string) => {
+        //       diag.debug(`METRICS: Applying patch for mysql@${moduleVersion}`);
+        //       if (isWrapped(moduleExports.connect)) {
+        //         console.log("METRICS: connect isWrapped. unwrap it");
+        //         this._unwrap(moduleExports, 'connect');
+        //       }
+        //       console.log("METRICS: wrap connect function");
+        //       this._wrap(moduleExports, 'connect', this._getConnectCommand());
+        //       return moduleExports;
+        //     },
+        //     (moduleExports?: any, moduleVersion?: string) => {
+        //       diag.debug(`METRICS: Removing internal patch for mysql@${moduleVersion}`);
+        //       if (moduleExports === undefined) return;
+        //       this._unwrap(moduleExports, 'connect');
+        //     }
+        //     ),
+        // ]
       ),
     ];
   }
+
+  // private _getConnectCommand() {
+  //   const instrumentation = this;
+  //   return (original: MysqlConnect['connect']) => {
+  //     return function patchedConnect(
+  //       this: unknown,
+  //       options: any,
+  //       callback: any
+  //     ) {
+  //       const patchedCallback = function(err: any, conn: any) {
+  //         console.log("METRICS: patchedCallback has been called!!!");
+  //         if(err || !conn) {
+  //           callback(err, conn);
+  //           return;
+  //         }
+  //         callback(err, conn);
+  //       }
+  //       return original.call(this, options, patchedCallback);
+  //     };
+  //   };
+  // }
 
   // global export function
   private _patchCreateConnection(format: formatType) {
@@ -194,6 +256,7 @@ export class MySQLInstrumentation extends InstrumentationBase<
             arg1 as getConnectionCallbackType,
             format
           );
+          thisPlugin._setPoolcallbacks(pool, thisPlugin);
           return originalGetConnection.call(pool, patchFn);
         }
         if (arguments.length === 2 && typeof arg2 === 'function') {
@@ -201,6 +264,7 @@ export class MySQLInstrumentation extends InstrumentationBase<
             arg2 as getConnectionCallbackType,
             format
           );
+          thisPlugin._setPoolcallbacks(pool, thisPlugin);
           return originalGetConnection.call(pool, arg1, patchFn);
         }
         if (arguments.length === 3 && typeof arg3 === 'function') {
@@ -208,6 +272,7 @@ export class MySQLInstrumentation extends InstrumentationBase<
             arg3 as getConnectionCallbackType,
             format
           );
+          thisPlugin._setPoolcallbacks(pool, thisPlugin);
           return originalGetConnection.call(pool, arg1, arg2, patchFn);
         }
 
@@ -342,5 +407,41 @@ export class MySQLInstrumentation extends InstrumentationBase<
         );
       };
     };
+  }
+  private _setPoolcallbacks(pool: mysqlTypes.Pool | mysqlTypes.PoolCluster, thisPlugin: MySQLInstrumentation) {
+    pool.on('connection', function (connection){
+      thisPlugin._connectionsUsage.add(1, {
+        'db.client.connection.usage.state': 'idle',
+        'db.client.connection.usage.name': connection?.threadId,
+      });
+      connection.on('end', () => {
+        thisPlugin._connectionsUsage.add(-1, {
+          'db.client.connection.usage.state': 'idle',
+          'db.client.connection.usage.name': connection?.threadId,
+        });
+      });
+    });
+
+    pool.on('acquire', function (connection){
+      thisPlugin._connectionsUsage.add(-1, {
+        'db.client.connection.usage.state': 'idle',
+        'db.client.connection.usage.name': connection?.threadId,
+      });
+      thisPlugin._connectionsUsage.add(1, {
+        'db.client.connection.usage.state': 'used',
+        'db.client.connection.usage.name': connection?.threadId,
+      });
+    });
+
+    pool.on('release', function (connection) {
+      thisPlugin._connectionsUsage.add(-1, {
+        'db.client.connection.usage.state': 'used',
+        'db.client.connection.usage.name': connection?.threadId,
+      });
+      thisPlugin._connectionsUsage.add(1, {
+        'db.client.connection.usage.state': 'idle',
+        'db.client.connection.usage.name': connection?.threadId,
+      });
+    });
   }
 }
--- a/plugins/node/opentelemetry-instrumentation-mysql/src/types.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql/src/types.ts
@@ -17,3 +17,7 @@
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
 
 export type MySQLInstrumentationConfig = InstrumentationConfig;
+
+export type MysqlConnect = {
+  connect: (options: any, callback: any) => void
+}
--- a/plugins/node/opentelemetry-instrumentation-mysql/test/index.metrics.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql/test/index.metrics.test.ts
@@ -1,0 +1,174 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  AggregationTemporality,
+  DataPointType,
+  InMemoryMetricExporter,
+  MeterProvider,
+  PeriodicExportingMetricReader,
+  ResourceMetrics,
+} from '@opentelemetry/sdk-metrics';
+import * as assert from 'assert';
+import { MySQLInstrumentation } from '../src';
+import { MysqlError, PoolConnection } from "mysql";
+import * as testUtils from '@opentelemetry/contrib-test-utils';
+
+const otelTestingMeterProvider = new MeterProvider();
+const inMemoryMetricsExporter = new InMemoryMetricExporter(
+  AggregationTemporality.CUMULATIVE
+);
+const metricReader = new PeriodicExportingMetricReader({
+  exporter: inMemoryMetricsExporter,
+  exportIntervalMillis: 100,
+  exportTimeoutMillis: 100,
+});
+
+otelTestingMeterProvider.addMetricReader(metricReader);
+
+import { registerInstrumentationTesting } from '@opentelemetry/contrib-test-utils';
+const instrumentation = registerInstrumentationTesting(
+  new MySQLInstrumentation()
+);
+
+instrumentation.setMeterProvider(otelTestingMeterProvider);
+
+async function waitForNumberOfExports(
+  exporter: InMemoryMetricExporter,
+  numberOfExports: number
+): Promise<ResourceMetrics[]> {
+  if (numberOfExports <= 0) {
+    throw new Error('numberOfExports must be greater than or equal to 0');
+  }
+  let totalExports = 0;
+  while (totalExports < numberOfExports) {
+    await new Promise(resolve => setTimeout(resolve, 20));
+    const exportedMetrics = exporter.getMetrics();
+    totalExports = exportedMetrics.length;
+  }
+  return exporter.getMetrics();
+}
+
+import * as mysqlTypes from 'mysql';
+
+describe('mysql@2.x-Metrics', () => {
+
+  let pool: mysqlTypes.Pool;
+  const testMysql = process.env.RUN_MYSQL_TESTS; // For CI: assumes local mysql db is already available
+  const testMysqlLocally = process.env.RUN_MYSQL_TESTS_LOCAL; // For local: spins up local mysql db via docker
+  const shouldTest = testMysql || testMysqlLocally; // Skips these tests if false (default)
+
+  before(function (done) {
+    if (!shouldTest) {
+      // this.skip() workaround
+      // https://github.com/mochajs/mocha/issues/2683#issuecomment-375629901
+      this.test!.parent!.pending = true;
+      console.log('Skipping test-mysql for metrics.');
+      this.skip();
+    }
+
+    if (testMysqlLocally) {
+      testUtils.startDocker('mysql');
+      // wait 15 seconds for docker container to start
+      this.timeout(20000);
+      setTimeout(done, 15000);
+    } else {
+      done();
+    }
+  });
+
+  after(function () {
+    if (testMysqlLocally) {
+      this.timeout(5000);
+      // testUtils.cleanUpDocker('mysql');
+    }
+  });
+
+  beforeEach(() => {
+    inMemoryMetricsExporter.reset();
+    // credentials to connect to pool if you run docker locally using 'npm run docker:start' from 'examples' folder
+    // pool = mysqlTypes.createPool({
+    //   host: 'localhost',
+    //   user: 'root',
+    //   password: 'secret',
+    // });
+    pool = mysqlTypes.createPool({
+      port: 33306,
+      user: 'otel',
+      host: '127.0.0.1',
+      password: 'secret',
+      database: 'test_db'
+    });
+  });
+
+  afterEach(done => {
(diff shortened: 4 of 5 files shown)


plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -50,9 +52,27 @@ export class MySQLInstrumentation extends InstrumentationBase<
   static readonly COMMON_ATTRIBUTES = {
     [SemanticAttributes.DB_SYSTEM]: DbSystemValues.MYSQL,
   };
+  private _connectionsUsage!: UpDownCounter;
```
dyladan, line 55: The `!` assertion could be removed if the metric creation was done in the constructor. Is there a reason it is split into a separate `setMetricInstruments` call?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -50,9 +52,27 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+    super.setMeterProvider(meterProvider);
+    this._setMetricInstruments();
+  }
+
+  private _setMetricInstruments() {
+    this._connectionsUsage = this.meter.createUpDownCounter(
+      'db.client.connections.usage',
+      {
+        description:
+          'The number of connections that are currently in state described by the state attribute.',
```
dyladan, line 72: "in state described by the state attribute" isn't very clear to me. Is this the number of connections in the connection pool? Does it include connections that are currently available only or all connections in the pool?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -343,4 +366,43 @@ export class MySQLInstrumentation extends InstrumentationBase<
       };
     };
   }
+  private _setPoolcallbacks(
+    pool: mysqlTypes.Pool | mysqlTypes.PoolCluster,
+    thisPlugin: MySQLInstrumentation
+  ) {
+    pool.on('connection', connection => {
+      thisPlugin._connectionsUsage.add(1, {
+        'db.client.connection.usage.state': 'idle',
```
dyladan, line 375: is `'db.client.connection.usage.{state,name}'` in semantic conventions?

dyladan (commented):
Implementation looks good to me. I know the semantic conventions package doesn't include metrics semconv yet but maybe we can add a TODO comment for when it is added?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
dyladan: I see state but not name https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md#connection-pools

dyladan (commented):
Actually looking at this closer it looks like it will create a new metric stream for each threadId which I think is not what you were intending. There should be a metric stream for `idle` and one for `used` which describes the count of idle and used connections respectively.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: There could be 2 states: `idle` and `used`, but the metric that describes them is one. I'm attaching the JSON of the metric I collected: the metric description is `"The number of connections that are currently in the state described by the state attribute"` (which means: `The number of connections that are currently in the state idle/used`), and the `dataPoints` field elaborates the metrics we collected, one dataPoint for `idle` and one dataPoint for `used`

(Maybe better to see it in a json format) 
`metrics: [
          {
            descriptor: {
              name: "db.client.connections.usage",
              description: "The number of connections that are currently in state described by the state attribute.",
              type: "UP_DOWN_COUNTER",
              unit: "{connections}",
              valueType: 1,
            },
            aggregationTemporality: 1,
            dataPointType: 3,
            dataPoints: [
              {
                attributes: {
                  "db.client.connection.usage.state": "idle",
                  "db.client.connection.usage.name": 2,
                },
                startTime: [
                  1667130552,
                  959287593,
                ],
                endTime: [
                  1667130553,
                  43613057,
                ],
                value: 1,
              },
              {
                attributes: {
                  "db.client.connection.usage.state": "used",
                  "db.client.connection.usage.name": 2,
                },
                startTime: [
                  1667130552,
                  959622210,
                ],
                endTime: [
                  1667130553,
                  43613057,
                ],
                value: 0,
              },
            ],
            isMonotonic: false,
          },
        ]`

dyladan:
@haddasbronfman does the above comment make sense to you? if not please let me know so i can make sure to clarify

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: You're right. I'll remove `name`

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: regarding the call to `setMetricInstruments` - we call `createUpDownCounter` to create a counter that is related to a `meterProvider`. If we call it in the constructor, and the user after that will call `setMeterProvider` (which he can..), it will override the `meterProvider`.
This is why I put it in a separate function called `_setMetricInstruments`, so it will be called in any case.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
dyladan: ah I see. i'm not sure how to better word it. maybe "in the state referenced by the attribute `db.client.connections.state`" or something. I didn't realize you were referring to another attribute

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
legendecas: See also https://github.com/open-telemetry/opentelemetry-js/pull/3267. The meter provider may be set multiple times. The metric instruments need to be re-created once that happens.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
legendecas, line 375: The spec defines the attribute name to be `state`, without the prefix `db.client.connection.usage.`. Am I interpreting the spec correctly?

pichlermarc (commented):
Happy to see metrics being added to this instrumentation. :slightly_smiling_face: Thank you for working on this :tada:

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: see the spec [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md#connection-pools), it says: _"All `db.client.connections.usage` measurements MUST include the following attribute"_. I understood it as if the field `state` should be a sub-field and include this prefix but maybe I didn't understand correctly... who can determine this issue?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -36,6 +36,8 @@ import type * as mysqlTypes from 'mysql';
 import { MySQLInstrumentationConfig } from './types';
 import { getConnectionAttributes, getDbStatement, getSpanName } from './utils';
 import { VERSION } from './version';
+import { UpDownCounter } from '@opentelemetry/api-metrics';
+import { MeterProvider } from '@opentelemetry/sdk-metrics';
```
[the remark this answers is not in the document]
haddasbronfman: Hi, `UpDownCounter` belongs to `'@opentelemetry/api-metrics'` and `MeterProvider ` belongs to another one - '@opentelemetry/sdk-metrics', so I think I can't unified them.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
legendecas: Kind note: please avoid marking threads as resolved before we settled the final resolution. Thank you!

I don't find any words that these attribute names are abbreviated. Other attribute definition shows that if they need to be scoped, they are cleared spelled out: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md#attributes.

@lmolkova As the author of the spec change, would you mind sharing your insight on this? Should the attribute name be scoped with `db.client.connection.usage`?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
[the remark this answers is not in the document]
legendecas: Instrumentations should not depends on SDK as production dependency directly. `MeterProvider` is exported from `@opentelemetry/api-metrics`: https://github.com/open-telemetry/opentelemetry-js/blob/experimental/v0.33.0/experimental/packages/opentelemetry-api-metrics/src/index.ts#L27. We should be able to use the type `MeterProvider` from `@opentelemetry/api-metrics`.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: @legendecas Should I wait until this issue (#3267) will be merged? When will it happen?

legendecas: I think it is ok to proceed without that patch. It may need to wait for the next release cycle.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: ok. accepted. I'll fix it.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: It seems you're right. (if they want to scope it - they say it explicitly). I'll change it.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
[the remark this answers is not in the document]
haddasbronfman: Thanks for the explanation. I'll fix this.

plugins/node/opentelemetry-instrumentation-mysql/package.json
```diff
@@ -63,6 +63,8 @@
     "typescript": "4.3.5"
   },
   "dependencies": {
+    "@opentelemetry/api-metrics": "^0.32.0",
+    "@opentelemetry/sdk-metrics": "^0.32.0",
```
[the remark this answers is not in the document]
haddasbronfman: thanks. fixed

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: ok. thanks. @dyladan is it ok to keep the `!` assertion? 
I can't initialize it in the constructor because the initialization is internal in `_setMetricInstruments`.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
lmolkova: > @lmolkova As the author of the spec change, would you mind sharing your insight on this? Should the attribute name be scoped with db.client.connection.usage

@legendecas  you probably wanted to summon @mateuszrzeszutek (who wrote DB metrics spec). If I read it correctly. `state` seems to be an attribute name and it doesn't have a namespace.

haddasbronfman:
I opened this issue: https://github.com/open-telemetry/opentelemetry-specification/issues/2958
until it is resolved, I think I can't finish this PR.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -343,4 +384,56 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+      thisPlugin._connectionsUsage.add(1, {
+        state: 'idle',
+        name: poolName,
+      });
+      connection.on('end', () => {
+        thisPlugin._connectionsUsage.add(-1, {
+          state: 'idle',
+          name: poolName,
+        });
+      });
```
dyladan, line 405: Are we sure only an idle connection can end? Is it not possible for this to be a used connection?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -343,4 +384,56 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+  }
+}
+function getPoolName(pool: mysqlTypes.Pool): string {
+  const c = pool.config.connectionConfig;
+  let poolName = '';
+  poolName += c.host ? `host: ${c.host} ` : '';
+  poolName += c.port ? `port: ${c.port} ` : '';
+  poolName += c.database ? `database: ${c.database} ` : '';
+  poolName += c.user ? `user: ${c.user}` : '';
+  return poolName;
```
dyladan, line 438: This might have a trailing space if there is no user. Is the pool name specified in the semconv or is this a scheme you came up with?

dyladan (commented):
Overall LGTM just have a quick question about the pool name and we can get this going.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: I'll trim the string. 
the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/database-metrics.md) doesn't say how the `pool.name` should look like, this is why I opened this issue: https://github.com/open-telemetry/opentelemetry-specification/issues/2958 and you can see there Mateusz's answer. 
If there is a need to put it in the semconv first, please let me know.

plugins/node/opentelemetry-instrumentation-mysql/package.json
```diff
@@ -47,6 +47,7 @@
   },
   "devDependencies": {
     "@opentelemetry/api": "^1.3.0",
+    "@opentelemetry/sdk-metrics": "^0.32.0",
```
pichlermarc, line 50: There is a stable metrics SDK release now :slightly_smiling_face: 
```suggestion
    "@opentelemetry/sdk-metrics": "^1.8.0",
```

plugins/node/opentelemetry-instrumentation-mysql/package.json
```diff
@@ -55,16 +56,17 @@
...
     "mocha": "7.2.0",
-    "mysql": "2.18.1",
     "nyc": "15.1.0",
     "rimraf": "3.0.2",
     "sinon": "14.0.0",
     "ts-mocha": "10.0.0",
     "typescript": "4.3.5"
   },
   "dependencies": {
+    "@opentelemetry/api-metrics": "^0.32.0",
```
pichlermarc, line 66: When using  `"@opentelemetry/sdk-metrics": "^1.8.0"` as suggested above this can be removed. Of course, the instrumentation needs to import metrics from `@opentelemetry/api` instead in that case.
```suggestion
```

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -32,10 +32,11 @@ import {
   DbSystemValues,
   SemanticAttributes,
 } from '@opentelemetry/semantic-conventions';
-import type * as mysqlTypes from 'mysql';
+import * as mysqlTypes from 'mysql';
 import { MySQLInstrumentationConfig } from './types';
 import { getConnectionAttributes, getDbStatement, getSpanName } from './utils';
 import { VERSION } from './version';
+import { UpDownCounter, MeterProvider } from '@opentelemetry/api-metrics';
```
pichlermarc, line 44:
```suggestion
import { UpDownCounter, MeterProvider } from '@opentelemetry/api';
```

NOTE: For this suggestion, we'd also need to remove `@opentelemetry/api-metrics` from the dependencies `package.json` :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -50,9 +51,27 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+  }
+
+  override setMeterProvider(meterProvider: MeterProvider) {
+    super.setMeterProvider(meterProvider);
+    this._setMetricInstruments();
+  }
+
+  private _setMetricInstruments() {
+    this._connectionsUsage = this.meter.createUpDownCounter(
+      'db.client.connections.usage', //TODO:: use semantic convention
```
pichlermarc, line 73: This is already following the semantic conventions, right? :thinking:

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -50,9 +51,27 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+      {
+        description:
+          'The number of connections that are currently in the state referenced by the attribute "state".',
```
pichlermarc, line 76:
```suggestion
          'The number of connections that are currently in state described by the state attribute.',
```
Nit: to match the description in the Spec exactly (not sure if necessary). :slightly_smiling_face:

pichlermarc (commented):
This is looking great already :slightly_smiling_face: 
Just a few nits + comments from my side.

Thank you for taking the time to work on this :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: Oh, you're right! I tried to do: `pool.getConnection` > `conn.query` > `pool.end` - meaning without calling `conn.release` before calling `end`, and the result was adding (-1) to idle.... while it should be added (-1) to `used` (which never happened because user didn't call `release` and add (-1) to `idle`. right?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: I moved 'getPoolName' to the utils, and add a test file named: 'utils.test.ts' that test this function.

plugins/node/opentelemetry-instrumentation-mysql/examples/src/tracer.ts
```diff
@@ -10,30 +10,48 @@ import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
...
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+import {
+  MeterProvider,
+  PeriodicExportingMetricReader,
+} from '@opentelemetry/sdk-metrics';
+const { OTLPMetricExporter } = require('@opentelemetry/exporter-metrics-otlp-grpc');
 
 const EXPORTER = process.env.EXPORTER || '';
 
 export const setupTracing = (serviceName: string) => {
```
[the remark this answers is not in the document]
haddasbronfman: This is the name of the tracer in all the `examples` folders. In my opinion, it is not worth breaking the uniformity.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
dyladan: Does the callback receive the connection? Is there a way to tell if it is idle?

plugins/node/opentelemetry-instrumentation-mysql/examples/src/tracer.ts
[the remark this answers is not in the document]
pichlermarc: Alright. :+1:

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: It's funny that you said that :)  because @dyladan said that if we follow the specification it might be unclear to users:
 https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1220#discussion_r1012954004

plugins/node/opentelemetry-instrumentation-mysql/package.json
haddasbronfman: done

plugins/node/opentelemetry-instrumentation-mysql/package.json
haddasbronfman: done

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: done

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: In [Pool.js](https://github.com/mysqljs/mysql/blob/master/lib/Pool.js#L16) there are arrays containing the connections. 
If I call `pool.getConnection` > `conn.query` > `conn.release` > `pool.end` - then the `end` function will remove the connection from `_freeConnections[]` and from `_allConnections[]` arrays, 
but If I call `pool.getConnection` > `conn.query` > `pool.end` without calling `conn.release` in the middle, then the `end` function will remove the connection only from `_allConnections[]` array,

Is it make sense to patch the `end` function and make our `end` be called before their `end`? maybe this way I can observe the pool on which `end` was called, and see if the `_freeConnections[]` contains the connection or not. 

of course I have an assumption here, that if a connection is in `_freeConnections[]` then `release` was already called and we already added (-1) to `used` and (+1) to `idle`.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
pichlermarc: I agree that your description is the better one of the two, the spec one is somewhat confusing :slightly_smiling_face: 

I'm mainly concerned about the following scenario:

Two apps using a MySQL library, both instrumented with OTel, written in two different languages. One language's instrumentation uses the spec description, the other does not. As far as I recall, descriptions are not part of the metric identity so both metrics get ingested without problems. However, depending on how the metrics backend handles descriptions, the description may be:
- the spec description
- the deviating description

The description in the backend may be based on which one gets ingested first, or keep changing, and that may be confusing to users. Not sure if that is a valid concern. :sweat_smile: 

WDYT @dyladan ?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -142,11 +166,39 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+            state: 'used',
+            name: poolName,
+          });
+        }
+        for (let i = 0; i < nFree; i++) {
+          thisPlugin._connectionsUsage.add(-1, {
+            state: 'idle',
+            name: poolName,
+          });
+        }
```
osherv, line 197: you can straight add and reduce the amount of the connections without these for loops.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
dyladan: That might be the only way to do it if the callback for `end` doesn't receive the connection itself to inspect

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
dyladan: TBH it's not clear how description is even meant to be used. Using the spec description is probably better.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
dyladan: Yeah I didn't catch this before but `thisPlugin._connectionsUsage.add(-nUsed, { state: 'used', name: poolName })` should have the same effect as the top loop. The bottom loop can be similarly replaced.

haddasbronfman: You're both right. my mistake. I got too used to the addition of 1/-1 .....

dyladan (approved):
LGTM. Minor nit on the readme/docs but the impl looks good

plugins/node/opentelemetry-instrumentation-mysql/examples/README.md
```diff
@@ -1,6 +1,6 @@
 # Overview
 
-OpenTelemetry MySQL Instrumentation allows the user to automatically collect trace data and export them to the backend of choice (we can use Zipkin or Jaeger for this example), to give observability to distributed systems.
+OpenTelemetry MySQL Instrumentation allows the user to automatically collect trace data and metrics (currently only the number of connections is supported) and export them to the backend of choice (we can use Zipkin, Jaeger or Grafana for this example), to give observability to distributed systems.
```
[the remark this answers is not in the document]
haddasbronfman: thanks. I'll fix that.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
pichlermarc: Fixed in [3b78a86](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1220/commits/3b78a867c9ba471a069813c21993448cac723786) :slightly_smiling_face:

pichlermarc (approved):
Great work! Thank you for adding this :slightly_smiling_face:

PR #1221, opened 2022-10-07
fix: remove types of the instrumented libs form public apis

### LINKED ISSUE #1217: Mongoose instrumentation imports types from the underlying package
`utils.d.ts` and `mongoose.d.ts` of the build assets import types from `mongoose`:

```
import type [...] from 'mongoose';
```

Which means that the compilation fails, if the user has not installed the `mongoose` package - happens with metapackage users for example.

This needs to be fixed by removing all references of the mongoose types from the public APIs.

### LINKED ISSUE #1218: Dataloader instrumentation imports types from the underlying package
`instrumentation.d.ts` of the build assets import types from `dataloader`:

```
import type [...] from 'dataloader';
```

Which means that the compilation fails, if the user has not installed the `dataloader` package - happens with metapackage users for example.

This needs to be fixed by removing all references to any of the dataloader types from the public APIs.

Rauno56:

Fixes #1217
Fixes #1218

- Cast Dataloaders module definitions to `InstrumentationNodeModuleDefinition<unknown>`
- Turned `patch` and `unpatch` methods on mongoose instrumentation private
- Cast the Collection argument in mongoose untils inside the function as opposed to having the type on the signature

### DIFF
--- a/plugins/node/instrumentation-mongoose/src/utils.ts
+++ b/plugins/node/instrumentation-mongoose/src/utils.ts
@@ -20,10 +20,8 @@ import { safeExecuteInTheMiddle } from '@opentelemetry/instrumentation';
 import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 
 export function getAttributesFromCollection(
-  _collection: unknown
+  collection: Collection
 ): SpanAttributes {
-  // Cast to avoid using mongoose types in the public API
-  const collection = _collection as Collection;
   return {
     [SemanticAttributes.DB_MONGODB_COLLECTION]: collection.name,
     [SemanticAttributes.DB_NAME]: collection.conn.name,
--- a/plugins/node/instrumentation-mongoose/src/utils.ts
+++ b/plugins/node/instrumentation-mongoose/src/utils.ts
@@ -20,8 +20,10 @@ import { safeExecuteInTheMiddle } from '@opentelemetry/instrumentation';
 import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 
 export function getAttributesFromCollection(
-  collection: Collection
+  _collection: unknown
 ): SpanAttributes {
+  // Cast to avoid using mongoose types in the public API
+  const collection = _collection as Collection;
   return {
     [SemanticAttributes.DB_MONGODB_COLLECTION]: collection.name,
     [SemanticAttributes.DB_NAME]: collection.conn.name,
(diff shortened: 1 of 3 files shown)


Rauno56:
Requesting reviews from the component owners: @henrinormak and @blumamir :pray:

henrinormak (approved):
The dataloader part looks good, the mongoose changes syntactically make sense, but I can't speak for their logical need.

MSNev (approved):

vmarchaud (approved):

plugins/node/instrumentation-mongoose/src/utils.ts
```diff
@@ -20,8 +20,10 @@ import { safeExecuteInTheMiddle } from '@opentelemetry/instrumentation';
 import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 
 export function getAttributesFromCollection(
-  collection: Collection
+  _collection: unknown
```
blumamir, line 23: I might be missing something, but why is this function part of the public api? it is not exported in `index.ts`

blumamir (approved):

plugins/node/instrumentation-mongoose/src/utils.ts
Rauno56: It is not exported to `index.ts`, but it's exported from utils. I expected TS to recursively import ts files and compile them. It seems that I was wrong and in the normal case utils is not compiled. Only when 

```
import * as mongooseUtils from '@opentelemetry/instrumentation-mongoose/build/src/utils';
```

is run explicitly, the problem occurs. Is there a compiler option that turns that process recursive?

That said, if I understand the situation correctly, we don't have to remove the type from that signature. Thanks for pointing that out, @blumamir!

dyladan (approved):

plugins/node/instrumentation-mongoose/src/utils.ts
blumamir: > Is there a compiler option that turns that process recursive?

From what I know, typescript is compiling only the single `.d.ts` file that is declared in `package.json` like this:
```js
{
  "types": "index.d.ts"
}
```

This file in turn can `import` other type files (or not), but since `utils.d.ts` is not imported from `index.d.ts`, then typescript never encounters `import * from 'mongoose'` so there are no issues.
This is only my understanding, but I could be missing some important details. typescript is so complicated 😅

PR #1222, opened 2022-10-07
fix(instrumentation-fs): fix `fs.exists` when it's util.promisified

### LINKED ISSUE #1185: fs.exists with util.promisify causes to unhandled rejection
```  
"name": "@opentelemetry/api",
"version": "1.2.0"
```

**Latest installed globally.**

Installed dependencies:
 npm i --unsafe-perm -g @splunk/otel @opentelemetry/api @opentelemetry/instrumentation-http @opentelemetry/instrumentation-express @opentelemetry/instrumentation-pino @opentelemetry/instrumentation-winston @opentelemetry/instrumentation-nestjs-core @opentelemetry/instrumentation-ioredis @opentelemetry/instrumentation-fs opentelemetry-instrumentation-mongoose @opentelemetry/instrumentation-dns @opentelemetry/instrumentation-amqplib 

Note: Also tried to install everything except fs but same behaviour.

v16.17.0

### What did you do?
Reproducing is easy:

test.mjs
```mjs
import * as fs from 'fs';
import * as util from 'util'; 

const result = await util.promisify(fs.exists)('./index.js');
console.log(result);
```

fs.exists should return 'true' as a resolved value and not as a rejected value.

### What did you see instead?

Executing node with `node -r @splunk/otel/instrument test.mjs` causes to this error:

node:internal/process/esm_loader:97
    internalBinding('errors').triggerUncaughtException(
                              ^
true
(Use `node --trace-uncaught ...` to show where the exception was thrown)

Without -r @splunk/otel/instrument it works as expected.

### Additional context
It looks like the promisified fs.exists is handled incorrectly (CB parameters) by opentelemetry and Promise.reject is called with (true) instead of resolve.

Rauno56:

`fs.exists` has an unusual signature with the callback never returning an error.
To get around that when using `util.promisify`, custom promisifing function is defined.
Our instrumentation doesn't take that into account.

Fixes #1185

Added separate patcher for `fs.exists` that is aware of the above and define a new custom promisifying function for our patched function.

### DIFF
--- a/plugins/node/instrumentation-fs/src/instrumentation.ts
+++ b/plugins/node/instrumentation-fs/src/instrumentation.ts
@@ -35,6 +35,7 @@ import type {
   EndHook,
   FsInstrumentationConfig,
 } from './types';
+import { promisify } from 'util';
 
 type FS = typeof fs;
 
@@ -63,6 +64,19 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
             );
           }
           for (const fName of CALLBACK_FUNCTIONS) {
+            if (fName === 'exists') {
+              // handling separately because of the inconsistent cb style:
+              // `exists` doesn't have error as the first argument, but the result
+              if (isWrapped(fs[fName])) {
+                this._unwrap(fs, fName);
+              }
+              this._wrap(
+                fs,
+                fName,
+                <any>this._patchExistsCallbackFunction.bind(this, fName)
+              );
+              continue;
+            }
             if (isWrapped(fs[fName])) {
               this._unwrap(fs, fName);
             }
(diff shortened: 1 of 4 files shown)


vmarchaud (approved):

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -63,6 +64,19 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
             );
           }
           for (const fName of CALLBACK_FUNCTIONS) {
+            if (fName === 'exists') {
+              // handling separately because of the inconsistent cb style:
+              // `exists` doesn't have error as the first argument, but the result
+              if (isWrapped(fs[fName])) {
```
Flarna, line 70: there is no need to duplicate this if. could be moved up before `if (fName === 'exists') {` once for all.

Flarna (approved):

plugins/node/instrumentation-fs/src/instrumentation.ts
Rauno56: Absolutely! Thanks!

PR #1223, opened 2022-10-09
fix(dns): remove lookupPromise polyfill for node8 dns promise tests 

evantorrie:

In #110 the `dns` promise based tests were reinstated for node 8 via a ["polyfill"-style function in the test source](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/110/files#diff-3060c03216ba47af1a62f232e8128a3844ab6e97ca67328f4c9c5ec4b5cfc7fcR36-R50) with
[this comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/110#discussion_r455406397).

It's unclear to me what exactly was being attempted here, since the net result of making all calls  in the test function funnel through this `lookupPromise()` function was that the functionality that was supposed to be tested (i.e. `dns.promises.lookup()` was no longer tested at all.

This may have been missed because code coverage was probably not run as part of CI then, but looking at code coverage now, the code in `instrumentation.ts` that handles [patching promise based calls](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/a6f054de256acc3415deb8137c7ea4bd6926c08d/plugins/node/opentelemetry-instrumentation-dns/src/instrumentation.ts#L133-L157) is [uncovered](https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/commit/55c81e71f316e288d4b5d641eb5bea08002eff6d/plugins/node/opentelemetry-instrumentation-dns/src/instrumentation.ts) by the tests in this file.

* Remove `lookupPromise()` test polyfill for `dns.promise.lookup()`
* Restore original intent of test: directly call `dns.promises.lookup()` function.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-dns/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-dns/src/utils.ts
@@ -107,7 +107,7 @@ export const setLookupAttributes = (
  * @param obj obj to inspect
  * @param pattern Match pattern
  */
-export const satisfiesPattern = <T>(
+export const satisfiesPattern = (
   constant: string,
   pattern: IgnoreMatcher
 ): boolean => {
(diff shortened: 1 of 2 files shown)


Rauno56 (approved):
Thanks!

Flarna (approved):

PR #1224, opened 2022-10-10
fix(instrumentation-pg): pass client to the connect callback

Rauno56:
- fix: only import types
- fix: pass client to the callback

-

-


seemk (approved):
Think it's a bug / misuse on knex's side, but nonetheless we shouldn't break user apps :weary: 

What if instead we'd pass all of the arguments given to `patchedClientConnectCallback` to the original callback? Could get rid of the signature and if pg adds any new arguments (which would be a non-breaking change), it'll still work.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -286,6 +287,6 @@ export function patchClientConnectCallback(
       });
     }
     span.end();
-    cb.call(this, err);
+    cb.call(this, err, client);
```
blumamir, line 290: WDYT about using the [`arguments object`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) so that we don't need to assume anything about the number, order and types of the actual parameters?
 
This practice ensures that we introduce minimal changes to the instrumented package, and just pass whatever arguments are sent to the patch right into the original callback AS IS with no modifications.

We could do something like:
```js
cb.apply(this, arguments);
```

or
```js
return function patchedClientConnectCallback(
  this: pgTypes.Client,
  err: Error,
  ...otherArgs: unknown[]
  ) {
  ...
  cb.call(this, err, ...otherArgs);
}
```

(didn't actually tried to comply these snippets)

weyert:
What could we do help merge this?

blumamir:
fixed in #1395

PR #1226, opened 2022-10-11
feat(koa): add layer type to request hook context

unflxw:

Unlike [the context given to the Express request hook](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-express/src/types.ts#L73-L80), the Koa request hook context does not contain information on whether the layer whose instrumentation is being intercepted by the request hook is a middleware layer or a router layer.

- Add the `layerType` to the context object passed to the `requestHook` interceptor function.
- Add the `layerType` property to the `KoaRequestInfo` context type.
- Add a check on the layer type to the unit tests.


Rauno56 (approved):

unflxw:
The failure in the browser tests seems unrelated to me (different package) -- let me know if there's something that needs to be looked at before merging this.

vmarchaud (approved):

blumamir (approved):

PR #1227, opened 2022-10-13
feat(lambda): add OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION environment variable

dyladan:
Adds an option to configure the `disableAwsContextPropagation` by environment variable


Aneurysm9 (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -77,6 +78,18 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
+
+    if (_config.disableAwsContextPropagation == null) {
+      if (
+        typeof env['OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION'] ===
+          'string' &&
+        env[
+          'OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION'
+        ].toLocaleLowerCase() === 'true'
+      ) {
+        _config.disableAwsContextPropagation = true;
```
blumamir, line 90: I think this will modify the value of the user's config object, e.g.
```js
const config = { foo: 'bar' };
const instrumentation = new AwsLambdaInstrumentation(config);
console.log(config); // {"foo": "bar", "disableAwsContextPropagation": true }
```

What do you think about copying the config object before modifying it, so that we don't introduce any side-effect to user?

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -77,6 +78,18 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
 
   constructor(protected override _config: AwsLambdaInstrumentationConfig = {}) {
     super('@opentelemetry/instrumentation-aws-lambda', VERSION, _config);
```
blumamir, line 80: WDYT about modifying the config object before using it as a paramater in the `super` call? might be a bit more robust IMO

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
```diff
@@ -541,6 +541,29 @@ describe('lambda handler', () => {
       assert.strictEqual(spans.length, 0);
     });
 
+    it('ignores sampled lambda context if env OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION is set to "true"', async () => {
+      process.env['OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION'] = 'true';
+      process.env[traceContextEnvironmentKey] = sampledAwsHeader;
+      initializeHandler('lambda-test/async.handler', {
+        disableAwsContextPropagation: true,
+      });
```
blumamir, line 549: This test sets both env variable and config object to the same value.
Since this PR adds the ability to set the option via env variable, I think it would be more interesting to test with an empty config object.

e.g. we should have a test that invokes 
```js
_config.disableAwsContextPropagation = true;
```

The current test will not enter the first `if` since:
```js
if (_config.disableAwsContextPropagation == null)
```
is not truthy

blumamir: I would even consider adding tests for some of the other new logic:
- config object has precedence over env (`disableAwsContextPropagation` set to false, `OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION` set to "true" => false should be used).
- env variable set to something which is not string should be ignored (trigger the `typeof env['OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION'] === 'string'` code
- env variable set to "True" or "TRUE" should be used.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -77,6 +78,18 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
+
+    if (_config.disableAwsContextPropagation == null) {
+      if (
```
blumamir, line 83: According to [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#boolean-value):
> If any value other than a true value, case-insensitive string "false", empty, or unset is used, a warning SHOULD be logged to inform users about the fallback to false being applied.

I think we should probably add this as a util function [in core repo](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-core/src/platform/node/environment.ts), to both simplify the code, don't repeat in the future, and comply with the spec.

Not a blocker for this PR of course, just a suggestion for future improvement.

codeboten (approved):
@dyladan I was about to submit this same PR! It would be great to have this shipped

JacksonWeber (approved):

mwear (approved):

blumamir:
There are still some open comments in this PR which I think are easy to address.
Can you please take a look and fix/resolve the comments, and then rebase so it can be merged?

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
dyladan: think a shallow copy like `_config = { ..._config };` is sufficient or should we deep copy with JSON stringify/destringify and risk crash if the config object contains something not serializable?

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
dyladan: 7f40ea85

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
dyladan: 7f40ea85

plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
dyladan: e2da529b

blumamir (approved):

blumamir:
Needs to run lint:fix and check why tests are failing.

codeboten (changes requested):
There's been an update toe the spec to change the behaviour and add a span link for the trace available in the environment variable. See https://github.com/open-telemetry/opentelemetry-specification/pull/3166 and here's an example implementation: https://github.com/open-telemetry/opentelemetry-specification/pull/3166

codeboten (approved):

Issue #1229, opened 2022-10-16
migrate circleci docker images to official images


blumamir:
This repo uses circleci docker images for running the unittest for pg and mysql.
Since we migrated to github actions long time ago, it doesn't make sense to use circleci images and we should migrate to the offficial images.

[see more info here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1197#pullrequestreview-1125068213)

PR #1230, opened 2022-10-16
chore: migrate tests docker images from circleci to official

blumamir:

- #1229 

- use official docker images for tests for postgres and mysql
- mysql image is kept as 5.7 (and not upgraded to 8) (renovate PR #1196). This is because mysql v8 changed default authentication from "mysql_native_password" to "caching_sha2_password" which is not compatible with old versions of `mysql` package which the instrumentation supports. See more details [here](https://stackoverflow.com/a/56509065).
- pg image is upgraded to latest "postgres:15-alpine". this was tested to work on CI and `npm test`
- remove "circle_" prefix from database names, and give them a more meaningful name


plugins/node/opentelemetry-instrumentation-mysql2/env
```diff
@@ -1,4 +1,4 @@
-export MYSQL_DATABASE=circle_database
+export MYSQL_DATABASE=otel_mysql_database
 export MYSQL_HOST=mysql
 export MYSQL_PASSWORD=secret
 export MYSQL_PORT=3306
```
osherv, line 4: Worth to check whether this file is still needed

osherv (approved):

plugins/node/opentelemetry-instrumentation-mysql2/env
blumamir: I cannot find how this env file is used.
@Rauno56 seems like you added it, can you tell us if it is used anywhere?

pichlermarc (approved):

PR #1231, opened 2022-10-17
test: reduce number of aws-sdk versions for tav

blumamir:
Once in a while, we need to reduce the number of aws-sdk versions to test with "test-all-versions".
They publish many versions and it takes too much time to test them all


Flarna (approved):

Rauno56 (approved):
Thanks. Hope that's the last time we have to do that. :crossed_fingers:

dyladan (approved):

PR #1246, opened 2022-10-24
chore: add renovate config for OTel core experimental

Flarna:
OTel core experimental packages require a manual update on each release because ^ ranges for 0.x versions only include new patch versions.

Add a separate renovate config for these packages.


Flarna:
Don't know if there is any way to test a renovate configuration. I added the new entry mostly based on guessing.

blumamir (approved):

Flarna:
@open-telemetry/javascript-maintainers How are the current rules regarding merging of PRs like this which is not within of a package having an owner?

blumamir:
> @open-telemetry/javascript-maintainers How are the current rules regarding merging of PRs like this which is not within of a package having an owner?

I don't think we have rules for that.
It's been open long enough for anyone interested in reviewing it to add comments.

Let's merge when CI is green

PR #1251, opened 2022-10-24
fix: separate public and internal types for all instrumentations

### LINKED ISSUE #1120: standerize the types.ts file for instrumentations
following [this PR](#1097), I noticed that some instrumentations has a `types.ts` file that contains both public and internal types mixed together.
I think this is not optimal, and want to suggest the following:

1. agree on a standart for how to structure the types in instrumentations (where to store internal / public types, how the files should be named, how to export them etc)
2. document it in the repo so new contributions / reviewers can use it as reference.
3. align all current instrumentations according to the standart.

If we get to an agreement, I can take the task to add documention markdown file to the repo and align all current instrumentations accordingly.

### Public Types
- Instrumentation config along with interfaces/types for hooks and filters (ignore matchers). [example](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/44f95270cfba2cb66e6eb9f78b871087479721bc/plugins/node/instrumentation-amqplib/src/types.ts#L72)
- Redecalring (copying) of types that we cannot import, because the package where they are declared includes implementation and is very large (`aws-sdk`).

A non backward compatible change in public types is breaking change and will be a problem once instrumentations are declared stable, so extra causion must be taken when adding types to this file. This is also why I think this change should be applied now while instrumentations are still experimental.

### Internal Types
Meant to be consumed by instrumentation itself (`instrumentation.ts`, `utils.ts` etc) and should not be exported to instrumentation users. Changes can be made to these types which are guarenteed to not break users as they are not exported.

- Types that extend some interface of the instrumented library itself. For example, adding new properties on `Request` for instrumentation purposes. [exmaple](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/44f95270cfba2cb66e6eb9f78b871087479721bc/plugins/node/opentelemetry-instrumentation-express/src/types.ts#L41)
- Declare types / interfaces used in the instrumentation logic itself. [example](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/44f95270cfba2cb66e6eb9f78b871087479721bc/plugins/node/opentelemetry-instrumentation-graphql/src/types.ts#L139)

### Two Type Files (my personal preference)
one file for internal types and another file for public types.
Then, in `index.ts` we can `export * from './types';` from the public types file, and in `instrumentaiton.ts` and other internal files we can import from the internal and public types files. This will help to prevent exporting internal types, and will introduce more structure into the codebase.

Suggestion for name: `types.ts` for public types. `internal-types.ts` for internal types. This is how [winston instrumentation is built](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-winston/src).

### Single Types File
Put all types in a single `types.ts` file like we do now, but document that `index.ts` must export only public types and not `*`.

I am happy to get any comments and suggestions regarding this. We can possibly extend in the future to other aspects of instrumentation code structure and styling.

blumamir:

- fixes #1120 See detailed explanation in the issue

- Added `GUIDELINES.md` file to the repo, which documents the applied conventions for future instrumentation implementors. This file can be extended in the future with more guidelines.

For all instrumentations:
- `types.ts` file now contains only public api types (instrumentation config + transitive types)
- all types which are required for the instrumentation and should not be publicly exported are moved to `internal-types.ts` file. They are only imported inside the instrumentation files.
- `export * from 'types'` in `index.ts`. many instrumentations had types that were not exported, making it difficult to use. Now that `types.ts` contains only public types, it's simpler and less error prone to just export them all instead of picking a subset of the relevant ones.
- I also exported `./enums/AttributeNames` for custom attribute names. They are now exported everywhere, making them available for users to consume.
- Added the `type` for relevant imports from the instrumented packaged. e.g. `import type * as foo from 'foo';`. This guarantee that we only use types. Since the instrumented package is not a dependency for the instrumentation, types is all we are allowed to import


blumamir:
@Flarna You investigated the instrumentation types issues quite a lot.
If you have some free cycles to review this change, I would highly appreciate it 🙏

GUIDELINES.md
```diff
@@ -0,0 +1,41 @@
...
+
+All types from `types.ts` file MUST be exported from instrumentation `index.ts` using export statement `export * from './types'`, which guarentee that they publicly available.
+
+#### Breaking Changes
+
+Since these types are publicly exported, a breaking change in this file can cause transpilation issues or require code changes for existing users. Special care and attention should be put when modifiying this file to guarantee backword compatibility or proper documentation of breaking changes.
+
+### Internal Types
+
+All types and constnats that instrumentation needs internally to implement the instrumentation logic. This can include extentions to instrumented package interfaces (for example - when adding data to existing objects), symbols for patches, enums etc.
```
Flarna, line 25:
```suggestion
All types and constants that instrumentation needs internally to implement the instrumentation logic. This can include extensions to instrumented package interfaces (for example - when adding data to existing objects), symbols for patches, enums etc.
```

Flarna, line 25: Maybe add that these types must not be exported

GUIDELINES.md
```diff
@@ -0,0 +1,41 @@
...
+
+#### File Name
+
+It is sometimes convinent to place these declrations in a dedicated file which can then be imported from various insturmentation files such as `instrumentation.ts`, `utils.ts` or test files.
```
Flarna, line 29:
```suggestion
It is sometimes convenient to place these declarations in a dedicated file which can then be imported from various instrumentation files such as `instrumentation.ts`, `utils.ts` or test files.
```

GUIDELINES.md
```diff
@@ -0,0 +1,41 @@
...
+
+Using this file is optional - when a type is used only in a single file, it is ok to declare it and use it in this file **without exporting it**. When a type is expected to be shared between multiple files, it is encourged to be declared in `internal-types.ts` to prevent circular dependencies.
+
+#### Exporting
+
+This file MUST NOT be exported publicly from instrumentation package, not directly (via `index.ts`) and not transitivly via export of other files.
+
+#### Breaking Changes
+
+Since the declarations in this file are not exported in the public instrumentation api, it is allowed to apply any breaking change refactors to this file.
```
Flarna, line 41: It's a bit confusing the name something "breaking changes" but actually it's not breaking at all because it's an internal implementation detail.

GUIDELINES.md
```diff
@@ -0,0 +1,41 @@
+# Instrumentations Implementation Guide
+
+This document captures general guidelines for implementing instrumentations in NodeJS and browser.
```
Flarna, line 3:
```suggestion
This document captures general guidelines for implementing instrumentations.
```
I assume nothing would change for e.g. deno so why limiting?

GUIDELINES.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Types
+
+### Public Types
+
+Public types are meant to be consumed by instrumentation users (OpenTelemetry distribution packages or end users implementing OpenTelemetry in their services). These are mostely instrumentation specific config interface (extending `InstrumentationConfig`) and the transitive types used in the config.
```
Flarna, line 9: Should we mention somewhere that re-exporting (and therefore depending on) types from the instrumented library or it's `@types` package is not allowed/should be avoided?

GUIDELINES.md
blumamir: I was just thinking to clarify that these are not OpenTelemetry general guidelines, but specific guidelines we adopt in this (js) repo.
But if you think it's better without I'll remove it

GUIDELINES.md
blumamir: Yes, we should.
I wanted to keep this PR as short as possible and only address one thing.
After it is merged, I'll open follow-up PRs for other issues such as the one you mentioned.

GUIDELINES.md
blumamir: I updated the text, can you please look again? 👍

Flarna (approved):
I haven't looked into each and every instrumentation but the pattern itself looks good and the files I took a look were fine.

blumamir:
> I haven't looked into each and every instrumentation but the pattern itself looks good and the files I took a look were fine.

Thank you for the review 🙏

osherv:
Great Job!

PR #1253, opened 2022-10-26
fix: allow hapi plugin from array to be registered as argument

dmathieu:

I missed this in #1046.
HAPI plugins can be registered directly as objects, with `server.register(plugin)`. But they can also be registered as an array, with `server.register([plugin])`.

Right now, we only accept `server.register([{plugin: plugin}]);`, which doesn't match the hapi documentation.
See https://hapi.dev/tutorials/plugins/?lang=en_US#loading

Allow registering hapi plugins from an array when the object is passed directly, and they are not nested within a deeper object.


dmathieu:
Failing browser test seems unrelated to the changes in this PR.

weyert (approved):

blumamir (approved):

PR #1255, opened 2022-10-26
feat(fastify): add requestHook support

luismiramirez:
The `requestHook` config option allows custom span handling per request layer.

- Adds `requestHook` support so users can add custom handling on spans.

- It's inspired by how `requestHook` is implemented in web instrumentations
- The `requestHook` user-defined function is called any time `_hookPreHandler()` is executed


plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
```diff
@@ -271,6 +283,19 @@ export class FastifyInstrumentation extends InstrumentationBase {
         spanName,
         spanAttributes
       );
+
+      if (instrumentation.getConfig().requestHook) {
+        safeExecuteInTheMiddle(
+          () => instrumentation.getConfig().requestHook!(span, request),
+          e => {
+            if (e) {
+              diag.error('fastify instrumentation: request hook failed', e);
```
blumamir, line 292: please use `instrumentation._diag.error(...)` instead, which is a component logger that automatically records the component name.
Then you can remove the "fastify instrumentation: " prefix and make this logging consistent with the rest of the instrumentation logging

plugins/node/opentelemetry-instrumentation-fastify/src/types.ts
```diff
@@ -17,9 +17,28 @@
 import { Span } from '@opentelemetry/api';
 import type { FastifyReply } from 'fastify';
 import { spanRequestSymbol } from './constants';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+import type { FastifyRequest } from 'fastify/types/request';
```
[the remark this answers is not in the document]
luismiramirez: Thanks for this feedback. This is really useful and educational.

I'll go with the `any` plus the comment, even though it's not a great option, but I see more possible issues from the other two options. Copying the type or the convenient subset could lead to maintenance issues in the future, and adding the whole fastify package as a dependency is a huge overhead as you stated.

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
```diff
@@ -271,6 +283,19 @@ export class FastifyInstrumentation extends InstrumentationBase {
         spanName,
         spanAttributes
       );
+
+      if (instrumentation.getConfig().requestHook) {
```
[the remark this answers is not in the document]
luismiramirez: I'm going to rephrase the docs to say that it is executed for every request instead. `preHandler` hook is the one that always has the request object available. Also, middlewares are not supported out of the box by Fastify.

plugins/node/opentelemetry-instrumentation-fastify/README.md
```diff
@@ -45,6 +45,28 @@ registerInstrumentations({
 
 See [examples/fastify](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/fastify) for a short example.
 
+## Fastify Instrumentation Options
+
+| Options | Type | Example | Description |
+| `requestHook` | `FastifyCustomAttributeFunction` | `(span, request) => {}` | Function for adding custom attributes to Fastify requests. Receives params: `Span, FastifyRequest`. |
```
blumamir, line 51: update to info here as well: `(span, requestInfo)`, Receives params: `Span, FastifyRequestInfo`

plugins/node/opentelemetry-instrumentation-fastify/src/types.ts
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Span } from '@opentelemetry/api';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+
+export type FastifyRequestInfo = {
```
blumamir, line 20: let's make it an interface instead of type =?
So it's consistent with other instrumentation and the below declarations :)

plugins/node/opentelemetry-instrumentation-fastify/test/instrumentation.test.ts
```diff
@@ -426,5 +428,65 @@ describe('fastify', () => {
         await startServer();
       });
     });
+
+    describe('using requestHook in config', () => {
+      it('calls requestHook provided function when set in config', async () => {
+        const requestHook = (span: Span, info: FastifyRequestInfo) => {
+          span.setAttribute('http.method', info.request.method);
```
blumamir, line 435: Can be more idiomatic to use `SemanticAttributes.HTTP_METHOD`, as this is not just a random attribute, it's from semantic conventions

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
```diff
@@ -271,6 +280,20 @@ export class FastifyInstrumentation extends InstrumentationBase {
...
+        safeExecuteInTheMiddle(
+          () => instrumentation.getConfig().requestHook!(span, { request }),
+          e => {
+            if (e) {
+              diag.error('request hook failed', e);
```
blumamir, line 289: no need for this one if you added `instrumentation._diag.error('request hook failed', e);`

blumamir (commented):
Thanks for addressing everything.
added few nit comments

blumamir (approved):
LGTM 🥇

PR #1256, opened 2022-10-27
feat(graphql): add ignoreTrivialResolveSpans config option

blumamir:

graphql instrumentation is creating a "resolve" span for each field in the query result.
this can be a HUGE amount of spans, out of which most are definitely not interesting.
Objects with many fields or lists can grow to hundreds or thousands spans per query!

Users usually don't write a resolver for each and every field. they have few resolvers to fetch data from database or http endpoint, and then they map the returned object fields using the default field resolver which just look for a property with the same name on the object.
https://graphql.org/learn/execution/#trivial-resolvers

Creating spans for these trivial property resolvers is not very interesting and increases noise and costs.

This PR adds a config option so that for fields that use the custom resolver which just resolve to a property, span creation is skipped.

Example:
```js
const schema = buildSchema(`
    type Query {
        users: [User!]!
    }

    type User {
        name: String
        age: Int
    }
`);

const rootValue = {
    // mock an async call to external data source
    users: Promise.resolve([{ name: 'John', age: 40 }, { name: 'Jane', age: 25 }, { name: 'Bob', age: 31 }]),
};

const source = '{ users { name, age} }';

await graphql({ schema, source, rootValue });
```

This query will create 7 "resolve" spans. One for resolving the users list, and then 3 for resolving each object name and 3 for resolving the age.
When turning on the new config option, only the first one is recorded. The span describing resolving of the query field "name" from the property "name" on "{ name: 'John', age: 40 }" to 'John' is not instrumented.

- Added config option to instrumentation config
- Documented it and other options in README
- updated the resolver function patch to be aware if it's patching the defaultFieldResolver, and if so, to check if it's about to resolve a property (in contrast to function). If so - it will call the original function without any instrumentation.


plugins/node/opentelemetry-instrumentation-graphql/README.md
```diff
@@ -55,8 +56,56 @@ registerInstrumentations({
 |  mergeItems | boolean |     false     |                    Whether to merge list items into a single element. example: `users.*.name` instead of `users.0.name`, `users.1.name`                   |   |
 |    depth    |  number |       -1      |                       The maximum depth of fields/resolvers to instrument. When set to 0 it will not instrument fields and resolvers. When set to -1 it will instrument all fields and resolvers.                      |   |
 | allowValues | boolean |     false     | When set to true it will not remove attributes values from schema source.   By default all values that can be sensitive are removed and replaced with "*" |   |
+| ignoreTrivialResolveSpans | boolean | false | Don't create spans for the execution of the default resolver on object peroperties. |
```
obecny, line 59:
```suggestion
| ignoreTrivialResolveSpans | boolean | false | Don't create spans for the execution of the default resolver on object properties. |
```

plugins/node/opentelemetry-instrumentation-graphql/README.md
```diff
@@ -55,8 +56,56 @@ registerInstrumentations({
...
 | responseHook | GraphQLInstrumentationExecutionResponseHook |     undefined     | Hook that allows adding custom span attributes based on the data returned from "execute" GraphQL action. |   |
 
+## Verbosity
+
+The instrumentation will be default create a span for each invocation of a resolver.
```
obecny, line 64:
```suggestion
The instrumentation by default will create a span for each invocation of a resolver.
```

plugins/node/opentelemetry-instrumentation-graphql/README.md
```diff
@@ -55,8 +56,56 @@ registerInstrumentations({
...
+
+A resolver is run by graphql for each field in the query response, which can be a lot of 
+spans for objects with many properties, or when lists are involved.
+
+There are few config options which can be used to reduce the verbosity of the instrumentations.
+
+They are all OFF by default. User can opt in to any combination of them to contol the amount of spans.
```
obecny, line 70:
```suggestion
They are all disabled by default. User can opt in to any combination of them to contol the amount of spans.
```

plugins/node/opentelemetry-instrumentation-graphql/src/types.ts
```diff
@@ -57,6 +59,18 @@ export interface GraphQLInstrumentationConfig extends InstrumentationConfig {
    */
   mergeItems?: boolean;
 
+  /**
+   * Don't create spans for the execution of the default resolver on object peroperties.
```
obecny, line 63:
```suggestion
   * Don't create spans for the execution of the default resolver on object properties.
```

plugins/node/opentelemetry-instrumentation-graphql/src/types.ts
```diff
@@ -57,6 +59,18 @@ export interface GraphQLInstrumentationConfig extends InstrumentationConfig {
...
+   *
+   * When a resolver function is not defined on the schema for a field, graphql will
+   * use the default resolver which just looks for a property with that name on the object.
+   * If the property is not a function, it's not very interesting to trace.
+   * This option can reduce noise and number of spans created.
+   *
+   * @default false
+   */
+  ignoreTrivialResolveSpans?: boolean;
```
obecny, line 72: NIt, move the param between `depth` and `mergeItems` to keep them in alphabetical order

obecny (commented):
thx for this change, few typos / nitpicks

obecny (approved):
lgtm

blumamir:
Thank you @obecny for the review. Very much appreciate it.

plugins/node/opentelemetry-instrumentation-graphql/README.md
unflxw: (also s/contol/control/)

Issue #1243, opened 2022-10-23
mongodb: instrumentation captures "net.host.name" instead of "net.peer.name"


blumamir:
### What version of OpenTelemetry are you using?
"@opentelemetry/instrumentation-mongodb": "^0.32.1",
"mongodb": "^4.11.0",

### What version of Node are you using?
16

### What did you do?
execute any mongodb command

### What did you expect to see?
[According to the spec]("@opentelemetry/instrumentation-mongodb": "^0.32.1",
"mongodb": "^4.11.0",), I expected instrumentation to record the mongodb network address information into `net.peer.*`

Attribute | Type | Description | Examples | Requirement Level
-- | -- | -- | -- | --
net.peer.name | string | Name of the database host. [1] | example.com | Conditionally Required: See alternative attributes below.
net.peer.port | int | Logical remote port number | 80; 8080; 443 | Conditionally Required: [2]
net.sock.family | string | Protocol address family which is used for communication. | inet6; bluetooth | Conditionally Required: [3]
net.sock.peer.addr | string | Remote socket peer address: IPv4 or IPv6 for internet protocols, path for local communication, etc. | 127.0.0.1; /tmp/mysql.sock | See below
net.sock.peer.port | int | Remote socket peer port. | 16456 | Recommended: [4]

### What did you see instead?
[Instrumentation records the mongodb network address info into `net.host.name` and `net.host.port`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/23589d65a1e167da375def9a532f58b1f0969d24/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts#L572).

```js
    if (host && port) {
      span.setAttributes({
        [SemanticAttributes.NET_HOST_NAME]: host,
        [SemanticAttributes.NET_HOST_PORT]: port,
      });
    }
```

[These values represent the local host address:](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#network-transport-attributes)

Attribute | Type | Description | Examples | Requirement Level
-- | -- | -- | -- | --
net.host.name | string | Logical local hostname or similar, see note below. | localhost | Recommended
net.host.port | int | Logical local port number, preferably the one that the peer used to connect | 8080 | Recommended

Also [from the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#nethostname):
[...]

PR #1257, opened 2022-10-27
fix(mongodb): use net.peer namespace for mongo host and port

blumamir:

- #1243 

- Changed the attributes "net.host.name" and "net.host.port" to the correct ones according to spec "net.peer.name" and "net.peer.port".

[From the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#netpeername):
> Sometimes host name is only available to instrumentation as a string which may contain DNS name or IP address. net.peer.name SHOULD be set to the available known hostname (e.g., "127.0.0.1" if connecting to an URL https://127.0.0.1.com/foo).

Thank you @lmolkova for the help


osherv (approved):

blumamir:
approved by the component owner. merging

blumamir:
@osherv is the component owner, but I need another approve to merge:

> Merging is blocked
> Merging can be performed automatically with 1 approving review.

vmarchaud (approved):

Issue #1248, opened 2022-10-24
[CI] browser-test are broken


Flarna:
see e.g.
https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/3312465704/jobs/5469167967
https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/3312081248/jobs/5468319602

```
> nyc karma start --single-run

ℹ ｢wdm｣: Hash: 9d2c943b68425fd58dd0
Version: webpack 4.46.0
Time: 35ms
Built at: 10/24/2022 11:26:46 AM
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |       0 |        0 |       0 |       0 |                   
----------|---------|----------|---------|---------|-------------------
lerna ERR! npm run test:browser stderr:

<--- Last few GCs --->

[7536:0x4a892d0]    38085 ms: Scavenge (reduce) 2043.9 (2048.3) -> 2043.4 (2049.5) MB, 19.4 / 0.0 ms  (average mu = 0.271, current mu = 0.240) allocation failure 
[7536:0x4a892d0]    38110 ms: Scavenge (reduce) 2044.1 (2048.5) -> 2043.5 (2049.8) MB, 24.4 / 0.0 ms  (average mu = 0.271, current mu = 0.240) allocation failure 
[7536:0x4a892d0]    38125 ms: Scavenge (reduce) 2044.4 (2051.8) -> 2043.8 (2052.0) MB, 14.9 / 0.0 ms  (average mu = 0.271, current mu = 0.240) allocation failure 


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xa3ac10 node::Abort() [node]
 2: 0x970199 node::FatalError(char const*, char const*) [node]
 3: 0xbba5ce v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xbba947 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xd76b65  [node]
 6: 0xd776ef  [node]
[...]

PR #1264, opened 2022-10-31
fix: address webpack memory issue for browser tests

MSNev:

- Contrib Browser tests are failing with out of memory issue #1248

- Make sure the same versions of webpack and karma-webpack are used for all packages

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -165,6 +165,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NPM_CONFIG_UNSAFE_PERM: true
+      NODE_OPTIONS: --max-old-space-size=4096
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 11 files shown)


.github/workflows/unit-test.yml
```diff
@@ -165,6 +165,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NPM_CONFIG_UNSAFE_PERM: true
+      NODE_OPTIONS: --max-old-space-size=4096
```
blumamir, line 168: On my machine, tests for `user-interactions` are failing with 2G of memory, but tests for `long-task` which has very similar dependencies are passing even with `256MB`.

I don't mind increasing the memory, but I think there is a real issue with the `user-interactions` code/deps/config which we are missing.

blumamir (approved):
Thanks!

.github/workflows/unit-test.yml
blumamir: As I mentioned in the issue, this is probably related to the code coverage plugin for webpack which consumes more memory due to 
https://www.npmjs.com/package/convert-source-map/v/1.9.0

I don't think we have a lot to do besides increasing the memory, but at least we know that it is not related to our code.

dyladan (approved):

Issue #1266, opened 2022-11-02
Missing type definitions for GraphQL and Fastify instrumentations when used through auto instrumentation


dotboris:
### What version of OpenTelemetry are you using?

<details>
<summary>Full versions listing</summary>

```sh
$ yarn list --pattern @opentelemetry
yarn list v1.22.19
├─ @opentelemetry/api-metrics@0.33.0
├─ @opentelemetry/api@1.2.0
├─ @opentelemetry/auto-instrumentations-node@0.34.0
├─ @opentelemetry/context-async-hooks@1.7.0
├─ @opentelemetry/core@1.7.0
├─ @opentelemetry/instrumentation-amqplib@0.31.0
├─ @opentelemetry/instrumentation-aws-lambda@0.33.1
├─ @opentelemetry/instrumentation-aws-sdk@0.9.3
├─ @opentelemetry/instrumentation-bunyan@0.30.1
├─ @opentelemetry/instrumentation-cassandra-driver@0.31.0
├─ @opentelemetry/instrumentation-connect@0.30.1
├─ @opentelemetry/instrumentation-dataloader@0.2.1
├─ @opentelemetry/instrumentation-dns@0.30.1
├─ @opentelemetry/instrumentation-express@0.31.3
├─ @opentelemetry/instrumentation-fastify@0.30.1
├─ @opentelemetry/instrumentation-fs@0.5.1
├─ @opentelemetry/instrumentation-generic-pool@0.30.0
├─ @opentelemetry/instrumentation-graphql@0.32.0
├─ @opentelemetry/instrumentation-grpc@0.32.0
│  ├─ @opentelemetry/api-metrics@0.32.0
│  └─ @opentelemetry/semantic-conventions@1.6.0
├─ @opentelemetry/instrumentation-hapi@0.30.1
├─ @opentelemetry/instrumentation-http@0.32.0
│  ├─ @opentelemetry/core@1.6.0
│  └─ @opentelemetry/semantic-conventions@1.6.0
├─ @opentelemetry/instrumentation-ioredis@0.32.2
├─ @opentelemetry/instrumentation-knex@0.30.1
├─ @opentelemetry/instrumentation-koa@0.33.0
├─ @opentelemetry/instrumentation-lru-memoizer@0.31.0
├─ @opentelemetry/instrumentation-memcached@0.30.1
├─ @opentelemetry/instrumentation-mongodb@0.32.2
[...]

PR #1267, opened 2022-11-03
fix(instrumentation-fastify): stop using fastify types in public api

dotboris:

Related to #1266 

Prior to this fix, typescript users that have the fastify instrumentation installed but do not have the `fastify` package installed would get compilation errors because typescript was expecting the `fastify` package to be present. This issue is bound to happen for anyone using the auto instrumentation feature.

- Mark all methods that use types from `fastify` as private in `instrumentation.ts`.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
@@ -209,7 +209,7 @@ export class FastifyInstrumentation extends InstrumentationBase {
     return fastify;
   }
 
-  public _patchSend() {
+  private _patchSend() {
     const instrumentation = this;
     this._diag.debug('Patching fastify reply.send function');
 
@@ -238,7 +238,7 @@ export class FastifyInstrumentation extends InstrumentationBase {
     };
   }
 
-  public _hookPreHandler() {
+  private _hookPreHandler() {
     const instrumentation = this;
     this._diag.debug('Patching fastify preHandler function');
 


plugins/node/opentelemetry-instrumentation-fastify/src/internal-types.ts
```diff
@@ -15,11 +15,10 @@
  */
 
 import { Span } from '@opentelemetry/api';
-import type { FastifyReply } from 'fastify';
 import { spanRequestSymbol } from './constants';
 
 export type HandlerOriginal = (() => Promise<unknown>) & (() => void);
 
-export type PluginFastifyReply = FastifyReply & {
+export type WithOptionalSpans<T> = T & {
```
blumamir, line 22: Can you please explain the motivation for this change?
Is this type exported publicly somewhere?

dotboris: With the previous code, this would add an `import {...} from 'fastify'` in `build/src/internal-types.d.ts`. Any code that imports this file directly or indirectly would get compilation errors.

Now that I think about it, I'm not quite sure if that was a real problem after all. I'm not sure if this file actually gets imported when someone uses the fastify integration.

What's kind-of confusing to me is that `internal-types.ts` and `utils.ts` are mostly private but their type definitions are still produced and made available in the package.

blumamir: This is very confusing. As far as I understand, public API is only what's exported from `index.d.ts`, as this is the file that typescript is importing when a user `import * from '@opentelemetry/instrumentation-fastify`'.
`index.d.ts` can in turn import other files (like `instrumentation.d.ts`) and they need to be transpiled successfully as well (must not contain import from fastify).

This is why we split the types into `types.ts` and `internal-types.ts` - so we can `import type {} from 'fastify'` without this leaking into public API.

> I feel that this new type makes it less likely that someone will expose types from fastify publicly on accident so I believe that it'd be worthwhile to keep it but it's clearly not required. I'll leave this up to you. Would you like me to revert this type?

This is a great goal, but I'm afraid that the codebase is full of these traps, and I don't think it's feasible to clean each and every instrumentation file that we use (those that are not imported from `index.ts`).
So I support preferring readability and simplicity :)
e.g. making the instrumentation functions private so their types are removed from the `instrumentation.d.ts` artifact.

dotboris: I have reverted this type.

blumamir (approved):
Thanks for submitting a fix

dyladan (approved):

PR #1268, opened 2022-11-03
fix(instrumentation-graphql): stop using types from `graphql` library in public api

dotboris:

Related to: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1266

When a typescript user installs the auto instrumentation package, they will get typescript compilation errors like so:

```
node_modules/@opentelemetry/instrumentation-graphql/build/src/instrumentation.d.ts:2:36 - error TS2307: Cannot find module 'graphql' or its corresponding type declarations.

2 import type * as graphqlTypes from 'graphql';
                                     ~~~~~~~~~

node_modules/@opentelemetry/instrumentation-graphql/build/src/types.d.ts:2:36 - error TS2307: Cannot find module 'graphql' or its corresponding type declarations.

2 import type * as graphqlTypes from 'graphql';
                                     ~~~~~~~~~
```

This PR solves this problem by ensuring that the public API of the GraphQL instrumentation doesn't reference or import types from the `graphql` library.

- Change `InstrumentationNodeModuleDefinition<typeof graphqlTypes>` to `InstrumentationNodeModuleDefinition<any>`

- Replace `ExecutionResult` with `any` in `GraphQLInstrumentationExecutionResponseHook`

    This includes documentation that points the user to the actual definition of the types in the `graphql` repository. I opted for this approach over copying the types from `graphql` because I didn't want to resolve the differences between the different versions of that type. This should make maintenance of this package easier.

- Remove `graphql` as a `peerDependency`

    Since the public API doesn't expose any `graphql` types anymore, this `peerDependency` didn't make much sense.

    Also, this change makes it so that users who have strict peer dependency requirements such as `pnpm` users don't have to install `graphql` even if they don't use it. This can happen when using auto instrumentation.


blumamir (approved):
LGTM

Thanks again for fixing this

osherv (approved):

Issue #1106, opened 2022-08-02
Runtime instrumentation


morigs:
### Is your instrumentation request related to a problem? Please describe

It's very useful to have metrics about platform runtime - heap status, GC metrics, event loop blocks etc...
OpenTelemetry Go provides [runtime instrumentation](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/runtime) for this.

For example, if application code behaves strangely, these metrics can understand if this because of GC lock or because of event loop block.

This instrumentation can move OTel forward to full-featured APM solution.

### Is it applicable for Node or Browser or both?

Mostly Node

### Do you expect this instrumentation to be commonly used?
I believe, it's very useful. prom-client [includes](https://github.com/siimon/prom-client#default-metrics) these metrics and they're commonly used

tomerfriedman:
Going to start working on this (first contribution). Any objections?

I already implemented something like that on my own. And was thinking to put it under "packages" here.

vmarchaud:
>  Any objections?

None ! I will assign the issue to you !

> And was thinking to put it under "packages" here. (same as host-metrics)

Yeah i think its the most logical place

dyladan:
Maybe this should go in the core repo? Similar to http and grpc instrumentation I think it is quite important and it is likely to get more maintainer attention there

dyladan:
Host metrics too maybe

daniel-white:
would love to see the GC metrics, something along the lines of https://github.com/SimenB/node-prometheus-gc-stats

LiranBri:
+1
seeing event-loop delays is super required to be able to monitor the health of a nodejs service

pikalovArtemN:
Add some prom-client metrics
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2136

david-luna:
Hi 👋 !

thanks to @maryliag we have some metrics in semantic conventions :)
https://github.com/open-telemetry/semantic-conventions/pull/991

I was about to start adding these into runtime metrics instrumentation but I found this issue and PR. I wonder if it makes sense to have a dedicated PR for the already added metrics (could be a partial of the ongoing #2136) or it's preferable to update & continue #2136 although it adds some metrics not defined yet in semconv

pikalovArtemN:
> Hi 👋 !
> 
[...]

PR #1272, opened 2022-11-06
feat(opentelemetry-instrumentation-runtime): initial version

mentos1386:
Prepares a module for runtime instrumentation, runtime metrics.
- #1106 

I have created a new module for runtime instrumentation and added initial example of "event loop lag" metrics.
I would refrain from adding more metrics in this PR to not block it for too long and add new metrics in subsequent PR.

The "event loop lag" metrics are inspired by https://github.com/marcbachmann/opentelemetry-node-metrics which is inspired by https://github.com/siimon/prom-client.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-runtime/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-runtime/src/instrumentation.ts
@@ -1,0 +1,33 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InstrumentationBase } from "@opentelemetry/instrumentation";
+import { createEventLoopLagMetrics } from "./metrics/event-loop-lag";
+import { RuntimeInstrumentationConfig } from "./types";
+import { VERSION } from "./version";
+
+/**
+ * Runtime instrumentation for Opentelemetry
+ */
+export class RuntimeInstrumentation extends InstrumentationBase {
+  constructor(_config: RuntimeInstrumentationConfig = {}) {
+    super("@opentelemetry/instrumentation-runtime", VERSION, _config);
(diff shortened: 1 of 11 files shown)


plugins/node/opentelemetry-instrumentation-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export class RuntimeInstrumentation extends InstrumentationBase {
+  constructor(_config: RuntimeInstrumentationConfig = {}) {
+    super("@opentelemetry/instrumentation-runtime", VERSION, _config);
+  }
+
+  init() {
+    createEventLoopLagMetrics(this.meter, this._config);
+  }
+}
```
mentos1386, line 33: I am not sure if this is the approach for "metric only" instrumentation. The `host-metrics` module is not using this approach[1]. 

[1] https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/opentelemetry-host-metrics

plugins/node/opentelemetry-instrumentation-runtime/src/metrics/event-loop-lag.ts
```diff
@@ -0,0 +1,99 @@
...
+  meter: Meter,
+  histogram: IntervalHistogram
+) => {
+  meter
+    .createObservableGauge("process.runtime.nodejs.eventloop.lag", {
+      description: "Lag of event loop.",
+      unit: "seconds",
+    })
+    .addCallback(async (observable) => {
+      const startTime = process.hrtime();
```
Ugzuzg, line 30: Since the lowest version of Node is 14, `process.hrtime.bigint()` can be used instead.

plugins/node/opentelemetry-instrumentation-runtime/src/instrumentation.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  init() {
+    const histogram = monitorEventLoopDelay({
+      resolution: this._config.monitorEventLoopDelayResolution,
+    });
+
+    histogram.enable();
+
+    createEventLoopLagMetrics(this.meter, histogram);
```
Ugzuzg, line 42: A meter can change while the application is running. The metrics need to be recreated on the new meter.
More context on the issue: https://github.com/open-telemetry/opentelemetry-js/issues/3249

mentos1386:
@Ugzuzg thanks for the review! Im a bit busy these days, but i'll try to make some time and fix the mentioned issues.

Are there any comments on the overall structure that i should address?

tomerfriedman:
great! We really really need this! :) Is there an ETA as to when it'll be available?

stoberov:
+1

This feature will be greatly appreciated.

blumamir:
@mentos1386 any update on this PR? are you still interested in finishing it?

Issue #1273, opened 2022-11-07
Non-spec / AWS Lambda Detector / Add escape hatch for synchronous resource detection


NPellet:
### Is your feature request related to a problem? Please describe

- We want to get telemetry running for our AWS Lambdas.
- If resource detection is asynchronous, then:
  - So should the SDK initialization
  - Can't run the initialization with node -r (otherwise setup might not be done before the lambda has actually run)
  - Has to change application code (I have hundreds and hundreds of lambdas...)
  - Need to create one new handler file for each lambda (very impractical)

### Describe the solution you'd like to see
Luckily, the lambda detector is actually synchronous. As it turns out, in my case, it's the only asynchronous resource detection I'd like to run.
So until the opentelemetry-js core developers figure out how they want to handle asynchronous resources, I'd love to get a simple escape-hatch for the lambda resource detector, that would be out-of-spec, but would allow synchronous detection of the resource.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetector.ts

It could look something like:

```typescript

export class AwsLambdaDetector implements Detector {
  detectSync(_config?: ResourceDetectionConfig): Resource {
    
    // same implementation as now
  }

  async detect( _config?: ResourceDetectionConfig): Promise<Resource> {
    return this.detect(_config);
  }
}
```

### Describe alternatives you've considered
When node 19+ can be used as runtime in AWS Lambda, then we can use the `--import` cli option and a top-level await.

I'll happily open a PR if you'd consider it.

Flarna:
As alternative you could use `@opentelemetry/sdk-trace-node` instead `@opentelemetry/sdk-node`.

It's a lower layer part of SDK for traces only used internally by NodeSDK and requires a bit more lines to write to set it up. For example you have to provide `Resource` as config. But this has exactly the advantage that you can do your sync resource detection and then create the NodeTracerProvider,...

Flarna:
There was some discussion about sync resource detection: https://github.com/open-telemetry/opentelemetry-js/issues/3055
And also a PR which looked quite promising as far as I remember: https://github.com/open-telemetry/opentelemetry-js/pull/2933

but both were closed by the stale bot...

NPellet:
Well it's what I do, e.g. for EKS

```typescript
    const resource = Resource.default()
      .merge(
        new Resource({
[...]

PR #1274, opened 2022-11-07
feat(resource-detector-aws): allow synchronous lambda detection

NPellet:

See #1273
When using asynchronous SDK init, the lambda might have executed before the SDK has terminated its init (as no top-level await is possible for any of the AWS node runtimes)

Our kubernetes services are usually long-lived, so if we miss a few traces at startup, it's not a problem. For lambdas it's a bit of a different story. I need to find a way to ensure that the init is synchronous.

As it turns out, the detector is actually synchronous, but just returns a Promise. If we can bypass that by exposing a synchronous method (and ditching the detectResources method at the same time at the application level), then everything becomes peachy :)

Added a synchronous method for lambda resource detection.

(PS: Sorry for the quote replacement, maybe we could add a .prettierrc somewhere)

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetector.ts
+++ b/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetector.ts
@@ -18,12 +18,12 @@ import {
   Detector,
   Resource,
   ResourceDetectionConfig,
-} from '@opentelemetry/resources';
+} from "@opentelemetry/resources";
 import {
   CloudProviderValues,
   CloudPlatformValues,
   SemanticResourceAttributes,
-} from '@opentelemetry/semantic-conventions';
+} from "@opentelemetry/semantic-conventions";
 
 /**
  * The AwsLambdaDetector can be used to detect if a process is running in AWS Lambda
@@ -31,7 +31,7 @@ import {
  * Returns an empty Resource if detection fails.
  */
 export class AwsLambdaDetector implements Detector {
-  async detect(_config?: ResourceDetectionConfig): Promise<Resource> {
+  detectSync(_config?: ResourceDetectionConfig): Resource {
     const functionName = process.env.AWS_LAMBDA_FUNCTION_NAME;
     if (!functionName) {
       return Resource.empty();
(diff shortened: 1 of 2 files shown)


detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetector.ts
```diff
@@ -18,20 +18,20 @@ import {
   Detector,
   Resource,
   ResourceDetectionConfig,
-} from '@opentelemetry/resources';
+} from "@opentelemetry/resources";
```
Flarna, line 21: seems you should tell your editor to leave ' untouched :o)

NPellet: Seems to me it should be set at the repo level :D. But fair, will do

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetector.ts
```diff
@@ -18,20 +18,20 @@ import {
...
+} from "@opentelemetry/semantic-conventions";
 
 /**
  * The AwsLambdaDetector can be used to detect if a process is running in AWS Lambda
  * and return a {@link Resource} populated with data about the environment.
  * Returns an empty Resource if detection fails.
  */
 export class AwsLambdaDetector implements Detector {
-  async detect(_config?: ResourceDetectionConfig): Promise<Resource> {
+  detectSync(_config?: ResourceDetectionConfig): Resource {
```
blumamir, line 34: If we add this option for sync detection, we should probably also document it in the README so users are aware of it and how to leverage it.

blumamir (commented):
Thank you for the PR and sorry for the late review.
Would [this PR](https://github.com/open-telemetry/opentelemetry-js/pull/3460) solve the issue once merged?

Adding a `detectSync` function will allow the lambda detector to be used in this case, but people that need to use other detectors in a lambda will still run into the same issue. I wonder if adding a new function to each non-async detector is the way to go here.

Once stable, this function will be part of the public API and we will not be able to change it, so I suggest we are careful with extending the detector API for specific packages

blumamir:
I think https://github.com/open-telemetry/opentelemetry-js/pull/3460 will resolve this usecase. If not, please feel free to reopen

PR #1276, opened 2022-11-09
fix(opentelemetry-instrumentation-user-interaction): donot patch empt…

jinphic:
improve opentelemetry-js-contrib/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts/_patchAddEventListener 

- when listener is null value，plugin.addPatchedListener will throw error because WeakMap key is not valid.
`new WeakMap<AnyFunction | EventListenerObject, Map<string, Map<HTMLElement, AnyFunction>>>();`

- check listener before addPatchedListener

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
@@ -320,7 +320,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
             return plugin._invokeListener(listener, this, args);
           }
         };
-        if (plugin.addPatchedListener(this, type, listener, patchedListener)) {
+        if (listener && plugin.addPatchedListener(this, type, listener, patchedListener)) {
           return original.call(this, type, patchedListener, useCapture);
         }
       };


jinphic:
11

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -320,7 +320,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase<unknown>
             return plugin._invokeListener(listener, this, args);
           }
         };
-        if (plugin.addPatchedListener(this, type, listener, patchedListener)) {
+        if (listener && plugin.addPatchedListener(this, type, listener, patchedListener)) {
```
blumamir, line 323: Wouldn't this be taken care of [few lines above](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/f8cabf306faa3e3eb4b9ce38ccdde842abdb2b82/plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts#L288)? seems the functions starts with 

```js
        // Forward calls with listener = null
        if (!listener) {
          return original.call(this, type, listener, useCapture);
        }
``` 

So I would expect that when hitting the current line `listener` is tested to be truthy

blumamir:
@jinphic I commented above on your PR, can you please take a look?

PR #1278, opened 2022-11-10
feat: update experimental deps to `^0.34.0`, core deps to `^1.8.0`, api to `^1.3.0`

mhassan1:
## Which problem is this PR solving?

This PR updates experimental dependencies to `^0.34.0` across all packages (except `@opentelemetry/instrumentation-aws-lambda`, which is not yet compatible with `@opentelemetry/instrumentation@^0.34.0`, see https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1285). It also upgrades `@opentelemetry/api` peer dependencies to `^1.3.0` in order to satisfy `^0.34.0`.

## Short description of the changes

This PR updates `package.json` files across the repository for the following dependencies (to `^0.34.0`):
* `@opentelemetry/instrumentation`
* `@opentelemetry/instrumentation-fetch`
* `@opentelemetry/instrumentation-xml-http-request`
* `@opentelemetry/instrumentation-http`
* `@opentelemetry/instrumentation-grpc`
* `@opentelemetry/sdk-node`
* `@opentelemetry/exporter-trace-otlp-http`

It also updates those `package.json` files with upgraded `@opentelemetry/api` peer dependencies (to `^1.3.0`) to satisfy the experimental dependency upgrades ([#3340](https://github.com/open-telemetry/opentelemetry-js/pull/3340/files#diff-40bc84a8c7f98466ec888091ddb23a6f4f4eb64ccf6d869e279c0913e5f9fb44R76)) , which led to upgrades of the following dependencies (to `^1.8.0`, see [#3340](https://github.com/open-telemetry/opentelemetry-js/pull/3340/files#diff-53a9ab71a29dccd334c7b4c57fb78107e15872fa60ba99836271e3fd1dfed678L61)):
* `@opentelemetry/core`
* `@opentelemetry/context-async-hooks`
* `@opentelemetry/context-zone-peer-dep`
* `@opentelemetry/resources`
* `@opentelemetry/sdk-trace-base`
* `@opentelemetry/sdk-trace-node`
* `@opentelemetry/sdk-trace-web`

PR #1281, opened 2022-11-11
feat: use GA version of metrics

Flarna:
Update host-metrics to use the GA version of metrics SDK and no longer use api-metrics.

Remove metrics packages from experimental OTel renovate config.

### DIFF
--- a/packages/opentelemetry-host-metrics/package.json
+++ b/packages/opentelemetry-host-metrics/package.json
@@ -59,7 +59,6 @@
     "typescript": "4.3.5"
   },
   "dependencies": {
-    "@opentelemetry/core": "^1.8.0",
     "@opentelemetry/sdk-metrics": "^1.8.0",
     "systeminformation": "^5.0.0"
   },
--- a/packages/opentelemetry-host-metrics/package.json
+++ b/packages/opentelemetry-host-metrics/package.json
@@ -43,10 +43,10 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@opentelemetry/api": "^1.0.0"
+    "@opentelemetry/api": "^1.3.0"
   },
   "devDependencies": {
-    "@opentelemetry/api": "^1.0.0",
+    "@opentelemetry/api": "^1.3.0",
     "@types/mocha": "8.2.3",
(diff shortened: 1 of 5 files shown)


Flarna:
seems this requires #1278 or #1279 to be merged first because of lerna hoisting.

packages/opentelemetry-host-metrics/test/metric.test.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 const SI = require('systeminformation');
-import { MetricAttributes } from '@opentelemetry/api-metrics';
+import { MetricAttributes } from '@opentelemetry/api';
```
blumamir, line 18: Should it use [`Attributes` instead of `MetricAttributes`](https://github.com/open-telemetry/opentelemetry-js/blob/d1621d2c8bb21484e72706834b53ad0c7da73ca6/api/src/metrics/Metric.ts#L86)?

Since this package now peer depends on `@opentelemetry/api@^1.3.0`, there is no reason to use the old type.

packages/opentelemetry-host-metrics/package.json
```diff
@@ -59,9 +59,8 @@
     "typescript": "4.3.5"
   },
   "dependencies": {
-    "@opentelemetry/api-metrics": "^0.32.0",
-    "@opentelemetry/core": "^1.0.0",
-    "@opentelemetry/sdk-metrics": "^0.32.0",
+    "@opentelemetry/core": "^1.8.0",
```
blumamir, line 62: While checking where it is actually being used, I found that it is listed as a dependency but not imported anywhere in the package.
I think it should be removed (not sure if it should be done as part of this PR, but could be nice while we are at it)

blumamir (approved):

legendecas (approved):

packages/opentelemetry-host-metrics/package.json
Flarna: done

packages/opentelemetry-host-metrics/test/metric.test.ts
Flarna: done

PR #1286, opened 2022-11-14
feat: add sqlcommenter comment with trace context to queries in pg instrumentation

henrinormak:
In order to get end-to-end telemetry, Google proposed a standard for sending trace context to DB engines - [sqlcommenter](https://google.github.io/sqlcommenter/). This is now part of the [OTel family of packages](https://github.com/open-telemetry/opentelemetry-sqlcommenter) (the Node variants have not been published under @opentelemetry scope AFAIK).

With this in mind, it makes sense to bring this capability directly to the instrumentation, as that way users without ORMs can still benefit from the ability to propagate a tracing context to the DB engine.

Added a new config flag to the pg instrumentation `addSqlCommenterCommentToQueries` (name up for discussion as it might end up being used by other DB driver instrumentations as well). If this is set to true, the instrumentation will modify the query (according to the sqlcommenter spec) before sending it onwards to the underlying library.

My end goal is to test this out with Postgres running in CloudSQL, which should allow me to get much deeper detail into the DB related Spans - https://cloud.google.com/blog/products/databases/sqlcommenter-merges-with-opentelemetry

Currently this PR is a draft, as I feel it is missing tests to validate that the query is actually correctly augmented with the tracing comment - the utility function is tested, but would be nice to have an integration test for this. I might be able to figure something out based on DB logs, but not 100% sure yet.


plugins/node/opentelemetry-instrumentation-pg/README.md
```diff
@@ -48,6 +48,8 @@ PostgreSQL instrumentation has few options available to choose from. You can set
 | ------- | ---- | ----------- |
 | [`enhancedDatabaseReporting`](./src/types.ts#L30) | `boolean` | If true, additional information about query parameters and results will be attached (as `attributes`) to spans representing database operations |
 | `responseHook` | `PgInstrumentationExecutionResponseHook` (function) | Function for adding custom attributes from db response |
+| `requireParentSpan` | `boolean` | If true, requires a parent span to create new spans (default false) |
```
henrinormak, line 51: This was previously added, but README was out of date.

weyert:
Awesome, great PR. Thank you. I had a similar query on the sqlcommenter repo many moons ago:
https://github.com/open-telemetry/opentelemetry-sqlcommenter/issues/9

I think it would be nice if this gets integrated directly in `otel-js`

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +290,68 @@ export function patchClientConnectCallback(
...
+    return query;
+  }
+
+  // As per sqlcommenter spec we shall not add a comment if there already is a comment
+  // in the query
+  if (hasValidSqlComment(query)) {
+    return query;
+  }
+
+  const headers: { [key: string]: string } = {
```
Flarna, line 331: Is there a reason why you built the headers by yourself instead of using `W3CTraceContextPropagator`?

henrinormak: I was unsure whether I should add a dependency against opentelemetry/core (where that propagator comes from), felt like it would be a bit too big of a dep? If that was not an issue, could definitely do that.

Flarna: core is quite small so I think it should be ok. It's likely installed anyway because the SDK depends on it.
The implementation there considers a bit more like the suppress propagation entry in context. This is likely a good thing but should not be repeated here and all other sql instrumentations.

henrinormak: Refactored to use the propagator, as you mention that core is relatively small, I added it as a direct dependency instead of peer-dep.

henrinormak:
@Flarna @weyert any further comments? Could we get this merged? Looking forward to trying it out with CloudSQL and pg, thanks a bunch! 🙇

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +292,60 @@ export function patchClientConnectCallback(
...
+  if (indexOpeningSlashComment < 0) {
+    return false;
+  }
+
+  const indexClosingSlashComment = query.indexOf('*/');
+  return indexOpeningDashDashComment < indexClosingSlashComment;
+}
+
+export function addSqlCommenterComment(span: Span, query: string): string {
+  if (query.length === 0) {
```
Flarna, line 312: Maybe add a `typeof(query) === "string"` check here as `queryConfig.text` is passed directly after a cast.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +292,60 @@ export function patchClientConnectCallback(
...
+  if (hasValidSqlComment(query)) {
+    return query;
+  }
+
+  const propagator = new W3CTraceContextPropagator();
+  const headers: { [key: string]: string } = {};
+  propagator.inject(
+    trace.setSpan(ROOT_CONTEXT, span),
+    headers,
+    defaultTextMapSetter
```
Flarna, line 327: This could be omitted as it should be the default anyway.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +292,60 @@ export function patchClientConnectCallback(
...
+  // sqlcommenter spec requires keys in the comment to be sorted lexicographically
+  const sortedKeys = Object.keys(headers).sort();
+
+  if (sortedKeys.length === 0) {
+    return query;
+  }
+
+  const commentString = sortedKeys
+    .map(key => {
+      const uriEncodedKey = encodeURIComponent(key);
```
Flarna, line 339: Is this needed? I would expect W3C header names don't require this.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -649,6 +665,108 @@ describe('pg', () => {
...
+
+      const span = tracer.startSpan('test span');
+      await context.with(trace.setSpan(context.active(), span), async () => {
+        try {
+          const query = 'SELECT NOW()';
+          const resPromise = await client.query(query);
+          assert.ok(resPromise);
+
+          const [span] = memoryExporter.getFinishedSpans();
+          const commentedQuery = addSqlCommenterComment(
```
Flarna, line 729: I guess if `addSqlCommenterComment` would be a noop function this test would pass.
Maybe assert that query contains the strings `traceparent` and `span.spancontext.traceid` or similar?

Flarna (approved):
left some minor comments. Actual merging is up to component owner/repo maintainers.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
henrinormak: Good point, the values are most likely also already safe?

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
henrinormak: There is an assert just a few lines below `assert.notEqual(query, commentedQuery)` that should achieve this check.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
henrinormak: W3CTraceContextPropagator is typed in a way where this is a required argument, otherwise would omit it.

Flarna: ah ok right. the default is only used in the inject method on Otel APi.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -309,7 +309,7 @@ function hasValidSqlComment(query: string): boolean {
 }
 
 export function addSqlCommenterComment(span: Span, query: string): string {
-  if (query.length === 0) {
+  if (query.length === 0 || typeof query !== 'string') {
```
Flarna, line 312:
```suggestion
  if (typeof query !== 'string' || query.length === 0) {
```
in case `query` is undefined you would not reach the typeof check

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -141,8 +143,21 @@ describe('pg', () => {
...
+    // that the sqlcommenter logic works
+    const privateClient: {
+      queryQueue: (pg.Query & { text?: string })[];
+    } = client as any;
+    const originalPush = privateClient.queryQueue.push;
+    privateClient.queryQueue.push = (...items) => {
+      executedQueries.push(...items);
+      return originalPush.apply(privateClient.queryQueue, items);
+    };
+
```
blumamir, line 160: nit: A possible alternative to this, which is a bit cleaner and idiomatic, is to use [sinon spy](https://sinonjs.org/releases/v14/spies/)

I think you should be able to do something like:
```js
  beforeEach(() => {
    // other code
    sandbox.spy((client as any).queryQueue, 'push');
  });
```

and then in your assertions, do something like
```js
const executedQueries = (client as any).queryQueue.push.getCall(0).args
```

You will need to also
```js
  afterEach(() => {
    sandbox.restore();
  });
```

This is optional

blumamir (commented):
Thank you for your contribution
Added few comments

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +292,56 @@ export function patchClientConnectCallback(
...
+  // sqlcommenter spec requires keys in the comment to be sorted lexicographically
+  const sortedKeys = Object.keys(headers).sort();
+
+  if (sortedKeys.length === 0) {
+    return query;
+  }
+
+  const commentString = sortedKeys
+    .map((key) => {
+      const escapedValue = headers[key].replace(/([^\\])(')/g, (_, prefix) => {
```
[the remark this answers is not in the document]
henrinormak: Good catch, I will add a proper test for this as well as refactor the logic.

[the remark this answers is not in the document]
henrinormak: I opted to go with the approach that other implementations of sqlcommenter seem to take, which is to just use percent encoding for everything. In Node it does mean we have to have special handling for quotes as well as some other characters (as `encodeURIComponent` is too lenient), but this approach is far more bulletproof than the regex based one.

aabmass:
@sjs994 can you take a look at this PR? 

- did we decide if we are migrating SQLCommenter instrumentation into the individual contrib repos as an overall strategy?
- is there some code we should be re-using from the SQLCommenter repo here?

weyert:
> @sjs994 can you take a look at this PR?
> 
> * did we decide if we are migrating SQLCommenter instrumentation into the individual contrib repos as an overall strategy?
> * is there some code we should be re-using from the SQLCommenter repo here?

Originally, the idea was to wait for SQLCommenter to be specced out before this. But I this could be a nice test case for it. The final dream state was to move them to instrumentations instead of a separate package.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -32,6 +32,7 @@ import {
 } from '@opentelemetry/sdk-trace-base';
 import * as assert from 'assert';
 import type * as pg from 'pg';
+import * as Sinon from 'sinon';
```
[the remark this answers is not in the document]
henrinormak: Will change, although copy-pasted this from other instrumentation.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +292,65 @@ export function patchClientConnectCallback(
...
+  if (indexOpeningSlashComment < 0) {
+    return false;
+  }
+
+  const indexClosingSlashComment = query.indexOf('*/');
+  return indexOpeningDashDashComment < indexClosingSlashComment;
+}
+
+// sqlcommenter specification expects us to URL encode all reserved
+// characters, but encodeURIComponent does not handle ! ' ( ) *,
```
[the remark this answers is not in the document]
henrinormak: Amended the comment, the sqlcommenter spec follows RFC spec for URL encoding, added reference to both. The current code does not do any special handling for SQL reserved characters (not sure if there is a need for it?)

blumamir (approved):
Thanks for addressing everything.
LGTM

We talked about it yesterday at the SIG. @aabmass and @sjs994 are working on this at google and offered to review it as well

sjs994 (commented):
Have one comment (this may be in general to all SQLCommenter library). Otherwise LGTM.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -289,3 +292,65 @@ export function patchClientConnectCallback(
     cb.call(this, err);
   };
 }
+
+function hasValidSqlComment(query: string): boolean {
+  const indexOpeningDashDashComment = query.indexOf('--');
```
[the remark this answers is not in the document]
henrinormak: Does the same also apply for the block comment? i.e if it is inside of a string literal, it is not a comment? I took the current code almost verbatim from the existing sqlcommenter Node implementation (the [knex one for example](https://github.com/open-telemetry/opentelemetry-sqlcommenter/blob/2f8841add68358069ebf1c0ee560ab3e98a59aa9/nodejs/sqlcommenter-nodejs/packages/sqlcommenter-knex/util.js#L35-L55)), so the fault is present there as well.

Adding a parser would not really be a problem, I think, although not sure what the overhead would be for each query execution, perhaps a comment + a OpenTelemetry diagnostic log line for queries that are ignored?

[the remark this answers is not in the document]
sjs994: Yes, it might affect a block comment also. I agree the issue originates in the original SQLCommenter libraries. That's why i am inclined towards adding a comment regarding this issue here (and if possible callout in docs). 

Adding a parser might affect the performance as well as will be an additional dependency in this library (which databases are anyway doing). 

We are thinking if the specs regarding "do not add a comment if there is already a comment" could be modified. Should it just be "add a new comment at the end of the statement" ?. 
But, for that we need to understand whether different databases (at-least the supported ones Postgres, MySQL, SQLite, etc) have any constraints on comments and whether any popular "hint"(Ex [pg_hint_plan](https://pghintplan.osdn.jp/pg_hint_plan.html), [MySQL query hints](https://dev.mysql.com/doc/refman/5.7/en/optimizer-hints.html)) features will get affected if multiple comment sections are present.

So, maybe till that is finalised, we can add a comment and tech debt maybe ?

sjs994 (approved):

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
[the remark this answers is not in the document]
henrinormak: Added a comment and a short note to the README for the flag.

vmarchaud (approved):

joelmukuthu:
Hi @henrinormak, thanks for working on this feature! Do you have thoughts on how other/custom attributes could be added to the SQL comments e.g. `action`, `controller` etc?

On another note, I currently use the following code to get the trace context (inspired by https://github.com/open-telemetry/opentelemetry-sqlcommenter/tree/2f8841add68358069ebf1c0ee560ab3e98a59aa9/nodejs/sqlcommenter-nodejs):

```ts
import { context, propagation } from '@opentelemetry/api';

type TraceContext = {
  traceparent?: string;
  tracestate?: string;
};

export const getActiveTraceContext = (): TraceContext => {
  const traceContext: TraceContext = {};

  propagation.inject(context.active(), traceContext);

  return traceContext;
};
```

I use this function in some other code where all SQL goes through before being passed on the node-postgres module. 
 
It works, somewhat (I never get `tracestate`), but it's quite different from what you did here and I'm fairly certain I'm doing something wrong. Would you be able to suggest the best way to get the trace context in my case? For full context, I'm not able to take advantage of `addSqlCommenterCommentToQueries` because I need to add `action` and `controller` to my SQL comments. Thanks in advance!

henrinormak:
Regarding the `action`, `controller`, `framework`. I suppose the correct approach would be to allow the instrumentation to be configured somehow to extract these from the context that is active when the query is being made. Query + the active context should be enough to know those values, right? I think that addition might be simple enough, but not sure whether I'll have time for this in the near future (PRs are welcome though as always :) )

As for the issue with the code snippet, I don't quite understand what you suspect being wrong, it does seem like one way of getting the context to be injected into the query (assuming you have configured the W3C propagator). It of course does not cover the actual injection of the context into the query as a comment, but I assume that is also handled. As for why the instrumentation implementation may differ - simply put, as the instrumentation does not know which of the propagators you have configured on OTel API, we use the propagator directly. Or did you mean some other difference?

joelmukuthu:
>  I suppose the correct approach would be to allow the instrumentation to be configured somehow to extract these from the context that is active when the query is being made

Sounds good. I can find time to work on it in the coming weeks; I'll probably start with an issue to verify the approach and ask for pointers on where to get started in the code.

As for my `getActiveTraceContext`, you've answered the question! I looked at the implementation here and thought that I should also be creating a propagator and maybe getting the active context some other way. Thanks for responding!

withnale:
@joelmukuthu - have you got an example of your code that is injecting the route and others into the comments. We're stuck with the same problem

joelmukuthu:
Hi @withnale. My implementation boils down to something like this: 

```ts
import { context, propagation } from '@opentelemetry/api';
import { Pool, type QueryConfig } from 'pg';

const getActiveTraceContext = (): { traceparent?: string; tracestate?: string; } => {
  const traceContext: TraceContext = {};
  propagation.inject(context.active(), traceContext);
  return traceContext;
};

// This is my own implementation of the spec. You can also borrow the one implemented
// in this PR.
const formatSqlComments = (comments: Record<string, string>): string => {
  // Based on this spec: https://google.github.io/sqlcommenter/spec/
  const keyValuePairs = Object.entries(comments).map(([key, value]) => {
    key = encodeURIComponent(key);
    value = encodeURIComponent((value || '').replace(/'/g, "\\'"));
    return `${key}='${value}'`;
  });

  if (!keyValuePairs.length) {
    return '';
  }

  return `/* ${keyValuePairs.sort().join(', ')} */`;
};

const pool = new Pool();

const runQuery = async (sql: string | QueryConfig): ReturnType<typeof pool.query> => {
  // According to the spec, you should first check that the query has no comments before 
  // appending any. You can borrow the implementation of `hasComments` from this PR.
  if (hasComments(sql)) {
    return pool.query(sql);
  }

  const activeTraceContext = getActiveTraceContext();
  const comments = {
    ...activeTraceContext,
    // Your custom attributes. These need to come from some kind of context that is unique to a request
    action: 'foo', // or something like `req.action`
    controller: 'bar',
    route: 'baz',
    // etc ...
  };

  const formattedComments = formatSqlComments(comments);

  if (typeof sql === 'string') {
    sql = `${options.sql} ${formattedComments}`;
  } else {
    sql.text = `${options.sql.text} ${formattedComments}`;
  }

  return pool.query(sql);
};
```

There's nothing special here, the main thing you'd need to make sure of is that all your database calls go through a central point e.g. the `runQuery` function above. Also this is not the best place to offer support so if you need more help, see here for more info: https://github.com/open-telemetry/opentelemetry-js-contrib#useful-links

But thanks for asking, I'll take a stab at https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1353 which would ideally solve your issue

withnale:
Ah. Thanks. I wrote something very similar. Initially I thought you'd worked out a way to patch instrumentation-pg, since that generates additional spans at a lower level. In the end, I decided the same as you that the sql comments were more important than these autogenerated spans.

Issue #999, opened 2022-05-03
@opentelemetry/instrumentation-aws-lambda should read API gateway parameters by default


lizthegrey:
### Is your feature request related to a problem? Please describe
Invokes create a root span, but populate only details about the lambda itself (e.g. /usr/bin/node, lambda runtime version, function name) rather than about the request being performed. When the lambda is being called directly from API gateway, these details should automatically be populated, rather than relying upon telemetry from API gateway to have been also plumbed up and amazon trace header used for correlation.

### Describe the solution you'd like to see
the following `event.input` (but _not_ `event.body`) values should be automatically extracted and put into attributes (essentially as a default requestHook); example from https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html
```
  "input":{
    "resource":"/helloworld",
    "path":"/helloworld",
    "httpMethod":"POST",
    "headers": {
      "Accept":"*/*",
      "content-type":"application/json",
      "Host":"r275xc9bmd.execute-api.us-east-1.amazonaws.com",
      "User-Agent":"curl/7.64.0",
      "X-Amzn-Trace-Id":"Root=1-1a2b3c4d-a1b2c3d4e5f6a1b2c3d4e5f6",
      "X-Forwarded-For":"72.21.198.64",
      "X-Forwarded-Port":"443",
      "X-Forwarded-Proto":"https"
    },
    "queryStringParameters":{
      "city":"Seattle",
      "name":"John"
    },
  },
```

### Describe alternatives you've considered
These attributes could be left on an API gateway span, except there's no easy way to produce OTel formatted spans out of API gateway logs, and even then, API gateway only speaks amazon trace header, not w3c tracecontext, so you'd need to use amazon propagator.

### Additional context
tagging component owners @willarmiros @NathanielRN and tagging Honeycomb PM @cartermp
tagging user who encountered this problem: @longility

lizthegrey:
still is a problem/footgun.

NathanielRN:
Tagging @vasireddy99 and @bryan-aguilar to triage this for investigation.

dyladan:
This is not stale. Tagging component owners @willarmiros @NathanielRN

cartermp:
Still not stale. Happy to contribute a fix myself, just unsure where I'd access the kind of information that we should be attaching to the created span.

willarmiros:
Hi @cartermp,

[...]

PR #1290, opened 2022-11-17
feat(opentelemetry-instrumentation-aws-lambda): Handle event when coming from an API Gateway, see #999

NPellet:

Enhancement proposal as discussed in #999 

When we determine that the Lambda has been triggered from the API Gateway, we create a new span (wrapping the lambda span itself), giving it the name of the path

Given https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html (See "Output format of a Lambda function for proxy integration") we analyze the statusCode replied by the lambda, and fail the span if that error is < 200 or >= 400.

Reacting to thrown exceptions doesn't fail the span, but the exception gets recorded.

I think creating a wrapping span makes more sense than annotating the existing one. It allows a clear separation in the tracing UI between lambdas that are called through SQS, the event bridge, etc. and the ones triggered through the Gateway.


NPellet:
By the way, this is how it would look in Jaeger (juste tested on a Lambda):

image

NPellet:
I guess this doesn't respect the specification: https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#status

Because I set the status to errored when replying 400 or 500.
But to be honest I don't understand the motivation behind this part of the spec. If the caller is not instrumented then we can raise no alert due to a failed Gateway API request.

NPellet:
Edited to add the query parameters, the path parameters and the headers as span attributes:

image

NPellet:
Added configuration to select whether returning error codes should set the span status to failed or not:

Configuration
```typescript
 detectApiGateway?: {
    enable: true,
    errorCodes?: Array<RegExp | number>
  }
```

When using `errorCodes: [ 400 ]`, only the error code 400 will set the span status to error, otherwise it will be set to ok

When using `errorCodes: [/^4[0-9]{2}$/]`, codes 400-499 will set the span status to error, otherwise it will be set to ok

When using `errorCodes: undefined`, the span status will be left unset

PR #1295, opened 2022-11-21
Eventbus propagation

NPellet:

Adds context propagation through the EventBridge `putEvents` methods

Uses the TraceHeader available from the SDK.
Sadly this only works with the X-Ray propagator, because it seems like AWS won't accept another type of propagator in the TraceHeader field.
The good part is that the AWS Lambda instrument already works without any change


NPellet:
Sorry, I screwed up. Reopened there: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1297

PR #1297, opened 2022-11-21
feat(instrumentation-aws-sdk): Add propagation through the event bridge using TraceHeader

NPellet:

Adds context propagation through the EventBridge `putEvents` methods

Uses the TraceHeader available from the SDK.
Sadly this only works with the X-Ray propagator, because it seems like AWS won't accept another type of propagator in the TraceHeader field.
The good part is that the AWS Lambda instrument already works without any change


plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/eventbridge.ts
```diff
@@ -0,0 +1,103 @@
...
+  NormalizedRequest,
+  NormalizedResponse,
+} from '../types';
+import {
+  MessagingDestinationKindValues,
+  SemanticAttributes,
+} from '@opentelemetry/semantic-conventions';
+import { EventBridge } from 'aws-sdk';
+
+class ContextSetter
```
NPellet, line 37: Note here a custom ContextSetter, unlike SQS and SNS

NPellet:
image

Proof that it works :)

NPellet:
Hmm it's a bit upsetting. The TraceHeader is not propagated to the EventBridge ApiDestination request headers.

So I think I should expose two modes: use TraceHeader or do like SQS and add the propagation as a message parameter

PR #1308, opened 2022-11-26
feat(dns): record looked-up hostname as attribute

ethanresnick:

The span for a DNS lookup should contain the hostname that was being looked up. This stores it on the span as an attribute.


vmarchaud:
This was already proposed with https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1178 with much discussion

ethanresnick:
@vmarchaud Ah, thanks, missed that somehow. I'll close this in favor of the other PR.

PR #1309, opened 2022-11-28
feat(instrumentation-aws-lambda): Propagate context when called from SQS

NPellet:

When an AWS Lambda is used as the target of an SQS Event, propagate context from the event producer to the consumer

Note on the following:

- In my tests, when using open telemetry, the trace header is actually found both in the MessageAttributes (that's the AWS-SDK SQS Service extension) as well as in Attributes > AWSTraceHeader (I'm not sure who sets this tbh). The context in AWSTraceHeader refers to the HTTP Post span, while the MessageAttribute refers to the `SendMessage` call.

- In this implementation, we first look for the header in Attributes.AWSTraceHeader and falls back to the MessageAttributes when not found.

- Because the link to the producer span is more important than the wrapping lambda imo, I added a flag to consider the producer span as the contextual span, and the parent span (the lambda invocation) as the linked span

- The code change is still a bit rough, but I'd like to get an idea if this has any chance to go through, in which case I could pour more efforts into this :).

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
@@ -51,6 +51,7 @@
     "@opentelemetry/sdk-trace-node": "^1.8.0",
     "@types/mocha": "7.0.2",
     "@types/node": "18.11.7",
+    "aws-sdk": "^2.1262.0",
     "gts": "3.1.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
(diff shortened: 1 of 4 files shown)


plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
```diff
@@ -60,6 +61,7 @@
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.32.0",
```
osherv, line 63: Update to 0.34

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -56,6 +58,72 @@ import {
...
+  : T; //[ keyof T ]
+
+type V = LowerCase<SQS.Message>;
+declare const a: V;
+
+class ContextGetter
+  implements TextMapGetter<LowerCase<SQS.MessageBodyAttributeMap>>
+{
+  keys(carrier: LowerCase<SQS.MessageBodyAttributeMap>): string[] {
+    return Object.keys(carrier);
```
osherv, line 81: I think better to move this logic to new modules

Issue #1285, opened 2022-11-14
`@opentelemetry/instrumentation-aws-lambda` is not compatible with `@opentelemetry/instrumentation@^0.34.0`


mhassan1:
- [X] This only affects the JavaScript OpenTelemetry library
- [ ] This may affect other libraries, but I would like to get opinions here first

In https://github.com/open-telemetry/opentelemetry-js/pull/3161, we started passing `null` to `require-in-the-middle` (RITM), which allows us to intercept all `require` calls with a single RITM instance; however, when `null` is passed, RITM no longer intercepts `require` calls to absolute paths. Since `@opentelemetry/instrumentation-aws-lambda` needs an absolute path to be intercepted, we cannot upgrade to `@opentelemetry/instrumentation@^0.34.0`. As far as I can tell, `@opentelemetry/instrumentation-aws-lambda` is the only instrumentation plugin that uses an absolute path.

We will need to resolve this one of these ways:
1. In `require-in-the-middle`, add support for absolute paths when `null` is passed
2. In `@opentelemetry/instrumentation`, handle this scenario by creating a new instance of `require-in-the-middle`, passing the absolute path
3. In `@opentelemetry/instrumentation-aws-lambda`, instrument code using another method, e.g. use `require-in-the-middle` directly, or patch the code directly

dyladan:
We knew this when we updated the instrumentation. The lambda instrumentation will probably have to switch to using require-in-the-middle directly

Flarna:
fyi @willarmiros

Flarna:
Should we rate this as bug? If instrumentations require different versions of `@opentelemetry/instrumentation` every user of `@opentelemetry/auto-instrumentations-node` gets two instances of it.

icco:
This is blocking our upgrade to 1.8.0 and 0.34. Do we have a sense of the amount of work this would take?

Flarna:
I changed the label from discussion to bug.

We might consider to remove `@opentelemetry/instrumentation-aws-lambda` from `@opentelemetry/auto-instrumentations-node` to unblock users to upgrade.

mhassan1:
[...]

PR #1314, opened 2022-11-29
feat(instrumentation-aws-lambda): use `require-in-the-middle` directly

mhassan1:

`@opentelemetry/instrumentation-aws-lambda` is incompatible with `@opentelemetry/instrumentation@^0.34.0` and higher, since it no longer supports instrumentation of absolute paths. This PR switches to using `require-in-the-middle` directly.

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1285

This PR replaces usage of `InstrumentationNodeModuleDefinition` and `InstrumentationNodeModuleFile` with direct invocation of `require-in-the-middle`. Any instrumentation plugins that instrument absolute paths will need to use this strategy; currently, this is the only one.

One downside of this approach is that it does not use the normal instrumentation patching infrastructure, so some features (e.g. `instrumentation.disable`) may not work.

In https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1285, we discussed a few options for resolving the need to instrument absolute paths. This PR represents option 3:
> In `@opentelemetry/instrumentation-aws-lambda`, instrument code using another method, e.g. use `require-in-the-middle` directly, or patch the code directly

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json
@@ -63,7 +63,8 @@
     "@opentelemetry/propagator-aws-xray": "^1.1.1",
     "@opentelemetry/resources": "^1.8.0",
     "@opentelemetry/semantic-conventions": "^1.0.0",
-    "@types/aws-lambda": "8.10.81"
+    "@types/aws-lambda": "8.10.81",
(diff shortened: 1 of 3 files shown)


mhassan1:
Superseded by https://github.com/open-telemetry/opentelemetry-js/pull/3457.

PR #1316, opened 2022-12-01
feat(pulsar-client): implement initial pulsar support

jaysonsantos:
- This adds support of pulsar-client's library, a Kafka like system

- It adds a new package to the plugins workspace with support for producing and consuming messages and linking them with spans.
- We had trouble monkeypatching pulsar's functions, which are native so, we ended up by creating proxy classes that implement the same interface as the original ones.


plugins/node/opentelemetry-instrumentation-pulsar-client/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InstrumentationConfig as BaseInstrumentationConfig } from '@opentelemetry/instrumentation';
+
+export type InstrumentationConfig = BaseInstrumentationConfig;
```
osherv, line 19: If you don't add any additional config, I think that this file is redundant.

plugins/node/opentelemetry-instrumentation-pulsar-client/src/instrumentation.ts
```diff
@@ -0,0 +1,98 @@
...
+    super(
+      '@opentelemetry/instrumentation-pulsar-client',
+      VERSION,
+      Object.assign({}, config)
+    );
+  }
+
+  override setConfig(config: InstrumentationConfig = {}) {
+    this._config = Object.assign({}, config);
+  }
```
osherv, line 46: If you use the Instrumentation config, I don't think you need to override this method

g-sponda: True, I removed and did some tests, everything looks fine without overriding this method

g-sponda:
Hey @blumamir 
I was also working with @jaysonsantos. I believe this PR is still relevant for us. 

I already did the rebase as requested.

plugins/node/opentelemetry-instrumentation-pulsar-client/src/proxies/consumer.ts
```diff
@@ -0,0 +1,211 @@
...
+  };
+}
+
+async function callback(
+  original: ConsumerListener,
+  message: Pulsar.Message,
+  consumer: Pulsar.Consumer
+) {
+  // Deal with both sync and asynchronous functions
+  const rv = original(message, consumer);
```
blumamir, line 207: I think that if you invoke `original` like this, it will not have the proper `this` value of the original call. This can possibly introduce a change for users after installing the instrumentation. 

If you search the repo for `original.apply` you can see examples of how the `this` parameter should be propagated to the patched function. Sometimes `original.call` is also used.

plugins/node/opentelemetry-instrumentation-pulsar-client/src/proxies/consumer.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+
+  acknowledge(message: Pulsar.Message): Promise<null> {
+    this.closePreviousSpan();
+    return this.consumer.acknowledge(message);
+  }
+
+  acknowledgeId(messageId: Pulsar.MessageId): Promise<null> {
+    this.closePreviousSpan();
```
blumamir, line 90: We have the id of the message being acknowledged, and we know the id of the message in `this._lastSpan` we will be closing in `closePreviousSpan`. I guess it can lead to using the wrong signal to end irrelevant spans, right? thus making the span duration not very useful or even misleading.

plugins/node/opentelemetry-instrumentation-pulsar-client/src/proxies/consumer.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  acknowledgeCumulative(message: Pulsar.Message): Promise<null> {
+    this.closePreviousSpan();
+    return this.consumer.acknowledgeCumulative(message);
+  }
+
+  acknowledgeCumulativeId(messageId: Pulsar.MessageId): Promise<null> {
+    this.closePreviousSpan();
+    return this.consumer.acknowledgeCumulativeId(messageId);
+  }
```
blumamir, line 112: There is some useful information here, like if the acknowledge is positive or negative (which indicates an error in processing the message that the operator might find useful). There is also the messageId which is not recorded and I guess can be handy for debugging and root causes investigations. The acknowledge is possibly cumulative or for a single message.

This can also be addressed by later PRs, just drawing attention to a possible improvement for usability.

blumamir (commented):
Thank you for your patient. 
I reviewed the code and added some comments. Some of them are nice-to-haves and some are important for stability and conventions of this repo.

Please feel free to comment on the ones you do not wish to address and we can reject them.

A few other suggestions:
- Add yourself to the component owner file [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/component_owners.yml) and take ownership of reviewing PRs and maintaining this instrumentation later on.
- Implement [`test-all-versions`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4503d3efe98c0b440582101df69a6df49a6cdb97/plugins/node/instrumentation-amqplib/package.json#L40) script to verify that tests are green for the range of advertised supported version.
- Add the instrumentation to the [`auto-instrumentation-node`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/auto-instrumentations-node) package, which includes all instrumentations in this repo.

Thank you again for your contribution 🙏

PR #1318, opened 2022-12-04
Fix/add aws oredering detectors

osherv:

-

-


samimusallam (approved):

nemoshlag (approved):

detectors/node/opentelemetry-resource-detector-aws/README.md
```diff
@@ -37,6 +37,20 @@ const tracerProvider = new NodeTracerProvider({ resource });
 - `awsLambdaDetector`: Populates `faas` and `cloud` for functions running on [AWS Lambda](https://aws.amazon.com/lambda/)
   - `faas.id` is currently not populated as it is not provided by the runtime at startup
 
+## Ordering
+
+Note that if multiple detectors are inserting the same attribute name, the first detector to insert wins.
+For example, if you had `detectors: [AwsEksDetector, AwsEc2Detector]` then `cloud.platform` will be `eks` instead of `ec2`.
+The below ordering is recommended.
+
+### AWS Detectors
```
blumamir, line 46: I think this header is redundant, as all the detectors in this package are AWS.
The [reference](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#ordering) in collector contrib lists AWS and GCP, but here we can just enumerate them directly.

detectors/node/opentelemetry-resource-detector-aws/README.md
```diff
@@ -37,6 +37,20 @@ const tracerProvider = new NodeTracerProvider({ resource });
 - `awsLambdaDetector`: Populates `faas` and `cloud` for functions running on [AWS Lambda](https://aws.amazon.com/lambda/)
   - `faas.id` is currently not populated as it is not provided by the runtime at startup
 
+## Ordering
+
+Note that if multiple detectors are inserting the same attribute name, the first detector to insert wins.
+For example, if you had `detectors: [AwsEksDetector, AwsEc2Detector]` then `cloud.platform` will be `eks` instead of `ec2`.
```
blumamir, line 43: This text is copied from collector-contrib component. I am pretty sure this is not how it behaves in js.
I assume this part `detectors: [AwsEksDetector, AwsEc2Detector]` is meant to be used in this context as described in the example above:
```js
const resource = await detectResources({
   detectors: [AwsEksDetector, AwsEc2Detector],
})
```

Which uses `merge` on all the resources [here](https://github.com/open-telemetry/opentelemetry-js/blob/1c3af6ce28969d3f42b014b765976b205690e8d2/packages/opentelemetry-resources/src/platform/node/detect-resources.ts#L50) like this:
```js
  return resources.reduce(
    (acc, resource) => acc.merge(resource),
    Resource.empty()
  );
```

The merge function should prefer the later attribute as [stated in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#merge):

> If a key exists on both the old and updating resource, the value of the updating resource MUST be picked (even if the updated value is empty).

Anyhow, this is not actually related to the aws detectors themselves, but rather to the node `detectResources` function specific implementation which is external to this package. 

I suggest verifying the behavior and refining the text to describe the full context of the usage to the user.

blumamir (changes requested):
@osherv added a concern. would be great to verify this text is tested to behave correctly. Maybe even add a unittest to assert this claim.

Thanks for working on it and sorry for the late review 🙏

detectors/node/opentelemetry-resource-detector-aws/README.md
osherv: fixed

detectors/node/opentelemetry-resource-detector-aws/README.md
osherv: So in [this](https://github.com/open-telemetry/opentelemetry-js/pull/3460) PR, the behaviour is tested, it should get merged soon. About the refining, refine to the specs readme?

blumamir: The current text "Note that if multiple detectors are inserting the same attribute name, the first detector to insert wins." is wrong. It should be

```suggestion
Note that if multiple detectors are inserting the same attribute name, the last detector to insert wins.
For example, if you had `detectors: [AwsEc2Detector, AwsEksDetector]` then `cloud.platform` will be `eks` instead of `ec2`.
```

Also in the list below.

> About the refining, refine to the specs readme?

I looked in spec and couldn't find any reference on how multiple resources should be merged together in case of attribute collisions (if the first occurrence wins or the last one). If you want to align it in the spec that would be great, but just updating the text here is also fine.

detectors/node/opentelemetry-resource-detector-aws/README.md
blumamir: Can you please push it? :)

PR #1321, opened 2022-12-06
chore(mongo): add DB_OPERATION attribute

### LINKED ISSUE #1242: mongodb: add db.operation attribute
mongodb instrumentation does not record [`db.operation` attribute](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes).
This attribute is available to instrumentation to capture, and is already used in the span name.

From the spec:

db.operation | string | The name of the operation being executed, e.g. the MongoDB command name such as findAndModify, or the SQL keyword. [4] | findAndModify; HMSET; SELECT | Conditionally Required: If db.statement is not applicable.
-- | -- | -- | -- | --

I am not sure how to interpret the "Requirement Level" for this attribute in this case, but since it's useful and easy to capture, I suggest adding it to the instrumentation recorded attributes.

haddasbronfman:

- fixes #1242

- add db.operation attribute to span

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
@@ -210,7 +210,8 @@ export class MongoDBInstrumentation extends InstrumentationBase {
           ns,
           server,
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          ops[0] as any
+          ops[0] as any,
+          operationName
         );
         const patchedCallback = instrumentation._patchEnd(span, resultHandler);
         // handle when options is the callback to send the correct number of args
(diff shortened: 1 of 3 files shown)


plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -255,7 +256,7 @@ export class MongoDBInstrumentation extends InstrumentationBase {
         const span = instrumentation.tracer.startSpan(`mongodb.${type}`, {
           kind: SpanKind.CLIENT,
         });
-        instrumentation._populateV3Attributes(span, ns, server, cmd);
+        instrumentation._populateV3Attributes(span, ns, server, cmd, type);
```
blumamir, line 259: if `commandType === MongodbCommandType.UNKNOWN` maybe it's better to leave this attribute empty instead of writing `command`

blumamir (commented):
thanks, LGTM
Please add/update the tests for this new attribute :)

blumamir (approved):

PR #1324, opened 2022-12-08
fix(redis-4): add support to new version of redis

osherv:
In the new version of @redis/client (1.4.2) they changed the function name of the function that we wrap (multi-->MULTI)

Link for the [CI](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/3646626475/jobs/6157912674).

Link for the new [redis](https://github.com/redis/node-redis/compare/client%401.4.1...client%401.4.2).

### DIFF
--- a/.github/component_owners.yml
+++ b/.github/component_owners.yml
@@ -14,6 +14,8 @@ components:
     - legendecas
   packages/opentelemetry-id-generator-aws-xray:
     - willarmiros
+  packages/opentelemetry-redis-common:
+    - haddasbronfman
   plugins/node/instrumentation-amqplib:
     - blumamir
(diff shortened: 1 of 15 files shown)


blumamir:
Have you taken a look at why the `net.peer.name` is not captured?
removing the entire support for instrumentation just because of this single attribute is not ideal. Maybe it's an easy fix?

osherv:
@blumamir after digging into the issue, found the problem and fixed :)

blumamir (approved):
Thanks 🥇 
Can you please also add it to the unpatch section?

osherv:
@blumamir oh right, sure!

blumamir (approved):

PR #1328, opened 2022-12-12
feat(propagator/aws-xray): Extract X-Ray header in a case-insensitive fashion

NPellet:

We observed that we receive API Gateway triggers where the X-Ray header is in camel case (despite having injected() with the same propagator). 
This PR allows the X-Ray propagator to extract from a object with a case-insensitive key.

1. Extract the keys from the carrier
2. Match the lower case key with the lower case reference key and extract the first match
3. Carry on with that key instead of the reference lower-case key


propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -28,26 +28,26 @@ import {
   INVALID_TRACEID,
   INVALID_SPANID,
   INVALID_SPAN_CONTEXT,
-} from '@opentelemetry/api';
+} from "@opentelemetry/api";
```
Flarna, line 31: Please remove these formatting changes.
They are unrelated and besides that it's likely that lint will complain.

NPellet: There we go, sorry about that. I had my prettierrc on another branch

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
```diff
@@ -92,9 +92,18 @@ export class AWSXRayPropagator implements TextMapPropagator {
...
+      return e.toLowerCase() === AWSXRAY_TRACE_ID_HEADER;
+    });
+    if (!relevantHeaderKey) {
       return INVALID_SPAN_CONTEXT;
+    }
+    const traceHeader = getter.get(carrier, relevantHeaderKey);
+
+    if (!traceHeader || typeof traceHeader !== 'string') {
+      return INVALID_SPAN_CONTEXT;
+    }
```
osherv, line 106: Maybe it's better to implement that logic in a new proprietary TextMapGetter

willarmiros (approved):
LGTM - the header should be case-insensitive

propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts
NPellet: @osherv I don't think that's advisable.
If, as a platform engineer, I use the SDK to register an AWS XRay propagator, or a B3 propagator, that's my problem. It should be invisible to the API consumer.

The developer who then uses the API to extract or inject the context should only concern himself with the serialisation/deserialisation of the context into/from the payload. It should not be concerned whether or not the propagator is expecting a lowercase string or not.

If you stand your point, then the AWS Lambda instrumentation has to be modified to use case-independent header keys, because at the moment, using both appears to be incompatible.

blumamir (approved):

blumamir:
@NPellet can you please fix the lint issue so CI is green and we can merge? Thanks

PR #1329, opened 2022-12-12
feat: update docker-related scripts and document docker engine prerequisite

patrickocoffeyo:

- Compose is now part of the `docker` cli, and can be used by running `docker compose` instead of `docker-compose`.

- Updates fastify example docker-related scripts so they use `docker compose` instead of `docker-compose`. 
- Update README so Docker Engine is listed as a prerequisite.


osherv (approved):

blumamir (commented):
Thank you for the fix.
I wonder if using `docker-compose` fails for the new docker clis, as this change might now confuse users with old cli versions that not yet support the new syntax.

WDYT?

patrickocoffeyo:
@blumamir You make a good point, I updated the readme with instructions to install docker, perhaps I should add a note about ensuring docker is up to date?

blumamir:
> @blumamir You make a good point, I updated the readme with instructions to install docker, perhaps I should add a note about ensuring docker is up to date?

Yes. If it is possible to find the minimum supported docker cli version for this syntax and add it to the `Prerequisites` that would be best

pichlermarc:
Thanks for your contribution :slightly_smiling_face: 

> > @blumamir You make a good point, I updated the readme with instructions to install docker, perhaps I should add a note about ensuring docker is up to date?
> 
> Yes. If it is possible to find the minimum supported docker cli version for this syntax and add it to the `Prerequisites` that would be best

Looks like we could add
 - a minimum Docker Desktop version ([4.12.0 maybe?](https://docs.docker.com/desktop/release-notes/#4120))
 - a note to install `docker-compose-plugin` for Docker Engine users with an existing installation :slightly_smiling_face: 

I had an existing (and up-to date) installation of Docker Engine for Linux, turns out it's the plugin that I had to install. :slightly_smiling_face:

blumamir:
@patrickocoffeyo are you still working on this? should we keep it open?

patrickocoffeyo:
@blumamir my apologies, this dropped off my radar. I'll update this later today, thank you for the reminder!

blumamir (approved):

PR #1332, opened 2022-12-17
fix(instrumentation-fs): allow realpath.native and realpathSync.native

### LINKED ISSUE #1315: @opentelemetry/instrumentation-fs: Node service crashes when 'native' function property is used
API | Stable Packages | Experimental Packages
-- | -- | --
1.3.x | 1.8.x | 0.34.x

16.18.1

### What did you do?
Updated OTel dependencies this included the addition of instrumentation-fs. The service instrumented is using node auto instrumentation with all enabled. 
The service imports [fs-extra@10.0.1 ](https://github.com/jprichardson/node-fs-extra), which internally extends fs evaluating `fs.realpath.native.name` in the process.
Starting the service caused a crash.

Re-tested by calling `fs.realpath.native(..args)` directly instead of using **fs-extra** and this also caused a crash.

Service up and running being instrumented with all automatic instrumentation applied.

### What did you see instead?
Service crashed with following exception: 
`{
  "error": {
    "message": "Cannot read properties of undefined (reading 'name')", 
    "name": "TypeError"
  },
  "level": "error",
  "message": "Uncaught exception",
  "name": "root",
  "stack": "TypeError: Cannot read properties of undefined (reading 'name')
    at exports.fromCallback (/services/node_modules/universalify/index.js:15:26)
    at Object.<anonymous> (/services/node_modules/fs-extra/lib/fs/index.js:57:27)
    at Module._compile (node:internal/modules/cjs/loader:1155:14)
....
}
`
If called directly the message would be along the lines of : `fs.realpath.native is not a function`

### Additional context
The issue seems to be that `shimmer` used for patching does not seem to include function properties like `native` when patching. It exists on the `__original` property but not in the root. This might be known as it was previously discussed in other issue: [remove shimmer, implement new one #618](https://github.com/open-telemetry/opentelemetry-js/issues/618).

I worked around this issue on my end first by **disabling** `instrumentation-fs` but then noticed that `fs-extra@10.1.0` has a workaround for this as well.

JakubKoralewski:
Hi! First PR here 🎅 Let me know if I did something wrong.

Fixes #1315 

Maybe I overengineered it a little by adding ['realpath', 'native'], and ['realpathSync', 'native'] to `CALLBACK_FUNCTIONS` and `SYNC_FUNCTIONS` respectively, instead of just adding if statements, but I guess it's future-proof. The crux of the matter is the call to `Object.assign(rv, original);` that assigns properties from the original function onto the new one, the `native` property gets assigned, but the function specific stuff is read-only so the function itself doesn't get overwritten.


weyert:
Will this also fix the issue with using `fs/promises?

JakubKoralewski:
No, sorry

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -172,14 +195,18 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
         span.end();
       }
     };
+    Object.assign(rv, original);
+
+    return rv;
   }
 
   protected _patchCallbackFunction<T extends (...args: any[]) => ReturnType<T>>(
```
blumamir, line 203: I see you didn't modify `_patchPromiseFunction` accordingly. Is it because the two 2 level functions don't have a promise signature `fs.promises.realpath.native`?

Can you please add a comment to describe this behavior, or modify the code so it's consistent in all cases?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -127,7 +149,8 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
     original: T
   ): T {
     const instrumentation = this;
-    return <any>function (this: any, ...args: any[]) {
+    functionName = memberToDisplayName(functionName) as FMember;
+    const rv = <any>function (this: any, ...args: any[]) {
```
blumamir, line 153: What is `rv` in this context? can we give it a meaningful name?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -253,6 +280,8 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
         return original.apply(this, args);
       }
     };
+    Object.assign(rv, original);
```
blumamir, line 283: If I understand correctly, this `assign` is required so that 1st-level functions like `realpath` will also have 2nd-level functions (`native`) copied to the patch function object, which is then pathed separately in a future iteration of the `wrap` function, right?

Can you please add a comment explaining the purpose of this line for future readers?

blumamir (commented):
Thanks for working on this fix.
Added a few comments, mainly about style and types.

plugins/node/instrumentation-fs/src/instrumentation.ts
JakubKoralewski: Short for returned value 😅 Is`patchedFunction` ok?

plugins/node/instrumentation-fs/src/instrumentation.ts
JakubKoralewski: You understand correctly.

Do you think it would make sense to create a wrapper function for Object.assign(rv, original) with a name like 'patchedFunctionWithOriginalProperties(function: F, original: O): O & F`, since Object.assign is used in two places and the comments could get desynchronized? I think then the function name speaks for itself without comments?

plugins/node/instrumentation-fs/src/types.ts
```diff
@@ -23,7 +23,19 @@ export type FunctionPropertyNames<T> = {
...
+  [K in keyof T]: {
+    [L in keyof T[K]]: L extends string
+      ? T[K][L] extends Function
+        ? [K, L]
+        : never
+      : never;
+  }[keyof T[K]];
+}[keyof T];
+
+export type FMember =
```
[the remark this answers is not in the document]
JakubKoralewski: My thinking was that you won't get property y of property x with 'x.y' so you'd need to check for the dot and split it to do anything useful, so I wanted to make the difference more obvious. However, I haven't used hooks, so I can't say what the requirements and use for them is. If it's just  for display then with the dot may be easier to understand?

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -54,32 +71,35 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
         (fs: FS) => {
           this._diag.debug('Applying patch for fs');
           for (const fName of SYNC_FUNCTIONS) {
-            if (isWrapped(fs[fName])) {
-              this._unwrap(fs, fName);
+            const { fs: fsToIndex, fName: fNameToIndex } = indexFs(fs, fName);
```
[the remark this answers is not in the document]
JakubKoralewski: I meant something like "this is the new FS object to be indexed" mainly because I didn't want to overwrite the 'fs' variable.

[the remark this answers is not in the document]
blumamir: but what are we indexing here? maybe `toPatch`?

plugins/node/instrumentation-fs/src/types.ts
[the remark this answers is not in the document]
blumamir: I don't use it myself in fs instrumentation, but I think people generally will use it for 2 purposes:
1. store the value into a custom attribute - even though it's in the span name, recording data into attributes is usually more desirable as it is simpler for processors to handle.
2. performing some switch-case to derive decisions.

In both of these cases, I think having the full, dot-separated name is more useful for users. Note that currently the hook is typed with `FMember` which is a close-list of just the fs function names. If we make this change, the hook will get a general `string` like `open` or `realpath.native` which is not part of `FunctionPropertyNames<typeof fs>`.

plugins/node/instrumentation-fs/src/instrumentation.ts
blumamir: `patchedFunction` is great, thanks :)

plugins/node/instrumentation-fs/src/instrumentation.ts
blumamir: I like the idea of extracting it into a function. 

The name `patchedFunctionWithOriginalProperties` describes the action it performs `Object.assign(rv, original)`, but not the intent. I think future readers would appreciate us stating it is important for 2-level functions like `realpath.native` to retain the 2nd-level when patching the 1st-level, then it is obvious why it is used and what function it serves in the code.

plugins/node/instrumentation-fs/src/instrumentation.ts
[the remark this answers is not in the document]
JakubKoralewski: We're indexing fsToIndex

JakubKoralewski:
@blumamir sorry for the delay, I tried to address all your suggestions let me know if I missed anything

blumamir:
Hi @JakubKoralewski , I apologize for the this PR is open.
Thank you so much for addressing the comments.

Would it be possible for you to jump on a quick zoom call to discuss this? I searched for your name in CNCF slack but couldn't find you there.

plugins/node/instrumentation-fs/src/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+export function memberToDisplayName(member: FMember): string {
+  if (Array.isArray(member)) {
+    const [K, L] = member;
+    return `${K}.${L}`;
+  } else {
+    return member;
+  }
+}
```
blumamir, line 62: This can move to `test/utils.ts` since its only used in tests

plugins/node/instrumentation-fs/src/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw Error(`Invalid member function name ${functionName}`);
+    return {
+      twoLevel: memberParts as [FunctionPropertyNames<FSObject>, string],
+    };
+  } else {
+    return {
+      oneLevel: functionName as FunctionPropertyNames<FSObject>,
+    };
+  }
+}
```
blumamir, line 37: We can return an array of the 1/2 levels

blumamir (approved):

plugins/node/instrumentation-fs/src/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    return {
+      objectToPatch: fs[functionName1],
+      functionNameToPatch: functionName2,
+    };
+  } else {
+    return { objectToPatch: fs, functionNameToPatch: member };
+  }
+}
+export function memberToDisplayName(member: FMember): string {
+  if (Array.isArray(member)) {
```
blumamir, line 56: Another question in mind - looks like `Fmember` is never an array now if I am not mistaking:
```js
export type FMember =
  | FunctionPropertyNames<typeof fs>
  | FunctionPropertyNamesTwoLevels<typeof fs>;
```

So I wonder if this function is really needed?

plugins/node/instrumentation-fs/src/utils.ts
JakubKoralewski: Removed.

plugins/node/instrumentation-fs/src/utils.ts
JakubKoralewski: Yeah, simplified the returned object to be array instead.

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -172,14 +188,15 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
...
   protected _patchCallbackFunction<T extends (...args: any[]) => ReturnType<T>>(
     functionName: FMember,
     original: T
   ): T {
     const instrumentation = this;
-    return <any>function (this: any, ...args: any[]) {
+    const patchedFunction = <any>function (this: any, ...args: any[]) {
       const activeContext = api.context.active();
 
       if (!instrumentation._shouldTrace(activeContext)) {
```
JakubKoralewski, line 202: this was where the merge conflict were (same thing in above and below functions), I hope it's fine

blumamir (approved):

blumamir:
@JakubKoralewski please fix the CI and I'll merge

blumamir:
@JakubKoralewski - I pushed a few commits to your branch to attempt to fix the build and lint issues. Please take a look and approve if you are ok with them

PR #1335, opened 2022-12-27
feat(instrumentation-fs): require parent span

unflxw:

- The fs instrumentation, unlike other instrumentations, does not have a `requireParentSpan` option, in order to only trace calls to the fs module that happen in the context of a broader action. Applications might, for example, read some files during boot, which would rarely make for useful traces.

- Add a `requireParentSpan` option, which defaults to `false`. If set to `true`, calls to the fs module will not be instrumented unless the current active context has a span, so the fs instrumentation will never create a trace of its own.

### DIFF
--- a/plugins/node/instrumentation-fs/src/instrumentation.ts
+++ b/plugins/node/instrumentation-fs/src/instrumentation.ts
@@ -128,9 +128,9 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
   ): T {
     const instrumentation = this;
     return <any>function (this: any, ...args: any[]) {
-      if (isTracingSuppressed(api.context.active())) {
-        // Performance optimization. Avoid creating additional contexts and spans
-        // if we already know that the tracing is being suppressed.
+      const activeContext = api.context.active();
+
+      if (!instrumentation._shouldTrace(activeContext)) {
         return original.apply(this, args);
       }
       if (
@@ -139,7 +139,7 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
         }) === false
       ) {
         return api.context.with(
-          suppressTracing(api.context.active()),
+          suppressTracing(activeContext),
           original,
           this,
           ...args
@@ -153,7 +153,7 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
       try {
         // Suppress tracing for internal fs calls
         const res = api.context.with(
-          suppressTracing(api.trace.setSpan(api.context.active(), span)),
+          suppressTracing(api.trace.setSpan(activeContext, span)),
           original,
           this,
           ...args
@@ -180,9 +180,9 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
   ): T {
     const instrumentation = this;
     return <any>function (this: any, ...args: any[]) {
-      if (isTracingSuppressed(api.context.active())) {
-        // Performance optimization. Avoid creating additional contexts and spans
-        // if we already know that the tracing is being suppressed.
+      const activeContext = api.context.active();
+
+      if (!instrumentation._shouldTrace(activeContext)) {
         return original.apply(this, args);
       }
       if (
@@ -191,7 +191,7 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
         }) === false
       ) {
         return api.context.with(
-          suppressTracing(api.context.active()),
+          suppressTracing(activeContext),
           original,
           this,
           ...args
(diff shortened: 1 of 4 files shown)


plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -415,4 +407,23 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
       }
     }
   }
+
+  protected _shouldTrace(): boolean {
+    const activeContext = api.context.active();
```
blumamir, line 412: suggestion: you can move the `activeContext = api.context.active()` to be a function argument. This has 2 potential benefits:
1. It communicates that the function depends on an external value `context` instead of reading it from the global scope as side effect. This makes the function dependencies explicit instead of implicit. If the value is hoisted to the function signature, it depends only on `this` and function arguments which might be simpler to maintain and understand.
2. It is more consistent with other places in the project where the context is passed as a function argument.

This is completely optional and not blocker

plugins/node/instrumentation-fs/test/parent.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    const contextManager = new AsyncHooksContextManager();
+    api.context.setGlobalContextManager(contextManager.enable());
+  });
+
+  afterEach(() => {
+    plugin.disable();
+    memoryExporter.reset();
+  });
+
+  const createSpanTests = (shouldCreateSpan: boolean) => {
```
blumamir, line 62: when reading this, I expected the function to create a span itself which is a bit confusing. How about naming it `expectFsSpan`?

plugins/node/instrumentation-fs/test/parent.test.ts
```diff
@@ -0,0 +1,164 @@
...
+const provider = new BasicTracerProvider();
+const memoryExporter = new InMemorySpanExporter();
+provider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
+
+const tracer = provider.getTracer('default');
+
+describe('fs instrumentation: requireParentSpan', () => {
+  let plugin: Instrumentation;
+  let fs: typeof FSType;
+  let context: api.Context;
```
blumamir, line 39: The name `context` is a bit general and doesn't reveal a lot about its role. How about `testAmbientContext`?

plugins/node/instrumentation-fs/test/parent.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  let endRootSpan: () => void;
+  let additionalSpans: number;
```
blumamir, line 41: also here, a better name could help a lot with reading this code. I suggest `testAmbeintSpansCount`

plugins/node/instrumentation-fs/test/parent.test.ts
blumamir, line 62: Maybe instead of `createSpanTests` it can be called `generateTestsForApiVariants`?

plugins/node/instrumentation-fs/test/parent.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    beforeEach(() => {
+      initializePlugin({ requireParentSpan: false });
+    });
+
+    withRootSpan(() => {
+      createSpanTests(true);
+    });
+
+    withoutRootSpan(() => {
+      createSpanTests(true);
```
blumamir, line 161: optional: When reading this, one has to jump to the function definition to understand the meaning of `true` and `false` for the `createSpanTests` invocation. A possible alternative is to use `createSpanTests({expectFsSpan: true})` which can be considered more readable, similar to how `initializePlugin({ requireParentSpan: false });` communicates the intention directly

blumamir (approved):
Thanks for working on this 
I added a few comments about code style. They are all non-blocker and optional, so feel free to resolve them if you think otherwise.

unflxw:
Thank you for the comments @blumamir! They make a lot of sense to me -- the naming around the test generator function is very wonky. I'll address these the coming week.

plugins/node/instrumentation-fs/test/parent.test.ts
unflxw: I went with `expectedAmbientSpanCount`, as it is added to `expectedSpans` (which I renamed to `expectedSpanCount`)

plugins/node/instrumentation-fs/test/parent.test.ts
unflxw: I went with `ambientContext`.

unflxw:
Done. Feel free to merge if the tests pass 🚀

blumamir (approved):

PR #1337, opened 2022-12-29
fix(mysql): add enhancedDatabaseReporting to mysql

### LINKED ISSUE #1237: mysql instrumentation: query parameters are recorded by default
"@opentelemetry/instrumentation-mysql": "0.31.1",

16

### What did you do?
```js
import { promisify } from 'util';
import mysql from 'mysql';
const connection = mysql.createConnection({});
const asyncQuery = promisify(connection.query).bind(connection);
asyncQuery('SELECT * FROM users WHERE id = ?', [1]);
```

I expected `db.statement` attribute to not record the parameter value by default, as it might contain sensitive values.

### What did you see instead?
`db.statement` contains the value `"SELECT * FROM users WHERE id = 1"`

### LINKED ISSUE #1336: MySQL: only serialize non-sensitive data in db.statement attribute
When I use mysql instrumentation, I see the full query in the span: `db.statement: SELECT a FROM b WHERE password='ppp'`.
I think we shouldn't display sensitive data by default. 

In mongo/mongoose, and in redis/ioredis we have a config object that has a field called: `dbStatementSerializer` which is a **function** that determines how the statement will be parsed. I think we need to add this ability to mysql also. 

I thought to add a boolean parameter that enables/disables the `db.statement`.
But I think this solution is too "narrow". With a function, the user can decide **how** the `db.statement` will look like and not just decide if to present it or not.

haddasbronfman:

fixes #1237 

 I added a config object to `MySQLInstrumentation` with a field called `dbStatementSerializer`. This is a function which the user can set and decide how to parse his db.statement.


plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -252,11 +256,11 @@ export class MySQLInstrumentation extends InstrumentationBase<
       const thisPlugin = this;
       diag.debug('MySQLInstrumentation: patched mysql query');
 
-      return function query(
+      return (
```
blumamir, line 259: function and arrow functions in js are not identical. In arrow function `=>` the `this` parameter will now point to the instrumentation object and not to the `mysql` object on which the patched function is called.

You can read more about it [here](https://blog.bitsrc.io/arrow-functions-vs-regular-functions-in-javascript-458ccd863bc1)

I believe it is important to retain the `function` syntax to keep the instrumentation functioning correctly

haddasbronfman: will be fixed. thanks!

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -278,10 +282,15 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
-        span.setAttribute(
-          SemanticAttributes.DB_STATEMENT,
-          getDbStatement(query, format, values)
-        );
+        const dbStatementSerializer =
+          typeof thisPlugin._config.dbStatementSerializer === 'function'
+            ? thisPlugin._config.dbStatementSerializer
+            : thisPlugin._defaultDbStatementSerializer.bind(thisPlugin);
+
+        let dbStatement = getDbStatement(query, format, values);
```
blumamir, line 290: The `getDbStatement` function actually [makes an effort](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/28ad01e040f44a7b9ff1f76c4ba486074e8faf9c/plugins/node/opentelemetry-instrumentation-mysql/src/utils.ts#L89) to insert these values which are not there in the first place:
```js
  if (typeof query === 'string') {
    return values ? format(query, values) : query;
  } else {
    return values || query.values
      ? format(query.sql, values || query.values)
      : query.sql;
  }
```

I wonder if we can just skip the `format` by default, and apply it only if we want to collect full payload. This will make it simpler and more efficient

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -323,6 +332,11 @@ export class MySQLInstrumentation extends InstrumentationBase<
     };
   }
 
+  private _defaultDbStatementSerializer(dbStatement: string) {
+    //replace all quoted words with a question mark (' and ". not `)
+    return dbStatement.replace(/['"]([^'"]*)['"]/g, '?');
+  }
```
blumamir, line 338: a better place for it is in `utils.ts` IMO.

This function belongs to the class but it is not using `this` and is more a utility than instrumentation object method

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -323,6 +332,11 @@ export class MySQLInstrumentation extends InstrumentationBase<
     };
   }
 
+  private _defaultDbStatementSerializer(dbStatement: string) {
+    //replace all quoted words with a question mark (' and ". not `)
```
blumamir, line 336: If I understand correctly, then this sql query:
```SQL
SELECT * FROM "customers"
```

Will turn into 
```SQL
SELECT * FROM ?
```
?

plugins/node/opentelemetry-instrumentation-mysql/src/types.ts
```diff
@@ -16,4 +16,17 @@
...
+ *
+ * @returns serialized string that will be used as the db.statement attribute.
+ */
+export type DbStatementSerializer = (statement: string) => string;
+
+export interface MySQLInstrumentationConfig extends InstrumentationConfig {
+  /**
+   * Custom serializer function for the db.statement tag
+   */
+  dbStatementSerializer?: DbStatementSerializer;
```
blumamir, line 31: `dbStatementSerializer` is used in redis and mongo instrumentations where the db.statement needs to be constructed from object to text (hence the "Serializer").

For SQL instrumentation, the statement is already present as an SQL string, and the convention is to use `enhancedDatabaseReporting` which is a boolean that indicates whether to collect the SQL (potentially sensitive) "values" or not (not sure it's consistent but that is the general attitude)

I suggest considering aligning with these conventions

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -278,10 +282,15 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
-          SemanticAttributes.DB_STATEMENT,
-          getDbStatement(query, format, values)
-        );
+        const dbStatementSerializer =
+          typeof thisPlugin._config.dbStatementSerializer === 'function'
+            ? thisPlugin._config.dbStatementSerializer
+            : thisPlugin._defaultDbStatementSerializer.bind(thisPlugin);
+
+        let dbStatement = getDbStatement(query, format, values);
+        dbStatement = dbStatementSerializer(dbStatement);
```
blumamir, line 291: if you choose to keep `dbStatementSerializer` as a configuration function (and not boolean), then it should be wrapped in try-catch to protect instrumentation from any bad configuration that can throw

blumamir, line 291: Might be more readable to separate the 2 variables instead of overriding a `let` variable, e.g:
```js
const fullDbStatement = getDbStatement(query, format, values);
const sanitizedDbStatement = dbStatementSerializer(dbStatement);
```

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
```diff
@@ -152,6 +152,66 @@ describe('mysql@2.x', () => {
...
+          const spans = memoryExporter.getFinishedSpans();
+          assert.strictEqual(spans[0].name, sql);
+          assert.strictEqual(
+            spans[0].attributes[SemanticAttributes.DB_STATEMENT],
+            'SELECT * FROM `books` WHERE `author` = ?'
+          );
+          done();
+        });
+      });
+    });
```
blumamir, line 171: Maybe move this test below to the `describe('dbStatementSerializer config'` part? I think it's more appropriate there as it tests the default config serializer

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
```diff
@@ -152,6 +152,66 @@ describe('mysql@2.x', () => {
         });
       });
     });
+
+    it('db.statement should have ? instead of quoted words when using default serializer ', done => {
```
blumamir, line 156: This test checks only the query `object` invocation. Below you added tests for both `object` and `string`. Should we also add a test for the default config with string parameter for symmetry?

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
blumamir: You can even consider capturing the values as another attributes like [in postgres](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/28ad01e040f44a7b9ff1f76c4ba486074e8faf9c/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts#L166):

```js
  if (
    instrumentationConfig.enhancedDatabaseReporting &&
    Array.isArray(queryConfig.values)
  ) {
    span.setAttribute(
      AttributeNames.PG_VALUES,
      arrayStringifyHelper(queryConfig.values)
    );
  }
```

Might make it easier to process downstream in the spans pipeline, as the values are formatted and not embedded in the statement text

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -278,10 +287,20 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+        if (typeof thisPlugin._config.dbStatementSerializer === 'function') {
+          dbStatement = getDbStatement(query, format, values);
+          dbStatement = thisPlugin._config.dbStatementSerializer(dbStatement);
+        } else {
+          if (typeof query === 'string') {
+            dbStatement = getDbStatement(query, format, values);
+            dbStatement = defaultDbStatementSerializer(dbStatement);
+          } else {
+            dbStatement = query.sql;
+          }
```
blumamir, line 300: This code makes it so if I invoke the driver with an object
```js
connection.query({ sql: 'SELECT * FROM "customers"'});
```

It will record `db.statement: 'SELECT * FROM "customers"'`

But the same query with a string will produce a different result:
```js
connection.query('SELECT * FROM "customers"');
```
`db.statement: 'SELECT * FROM ?'`

It will also behave differently when the parameters are embedded into the query string and not supplied as values:
```
connection.query({ sql: 'SELECT * FROM customers WHERE password="1234"'}); // db.statement: SELECT * FROM customers WHERE password="1234"

connection.query('SELECT * FROM customers WHERE password="1234"'); // db.statement: SELECT * FROM customers WHERE password=?

```

I think it is preferable to have consistency with how the instrumentation operates by default.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
haddasbronfman: right

blumamir:
Thanks @haddasbronfman , Maybe I am missing something, but wouldn't all the potentially sensitive values still be written to the `db.statement`?

haddasbronfman:
You're right. 
(This is why I thought to add the dbStatementSerializer in the first place).
But what can we do now? 
Serializer is not an option because we should discuss it at the SIG, and enhancedDatabaseReporting is also not an option because it makes no sense to capture the values when they already appear in the `db.statement`.

blumamir:
> You're right. (This is why I thought to add the dbStatementSerializer in the first place). But what can we do now? Serializer is not an option because we should discuss it at the SIG, and enhancedDatabaseReporting is also not an option because it makes no sense to capture the values when they already appear in the `db.statement`.

The instrumentation actually receives the plain SQL statement without values, and it inject the values [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/8b8bfebdd6b4f43a8df540979874a6c01c999957/plugins/node/opentelemetry-instrumentation-mysql/src/utils.ts#L90) by calling the `format` function:
```js
  if (typeof query === 'string') {
    return values ? format(query, values) : query;
  } else {
    return values || query.values
      ? format(query.sql, values || query.values)
      : query.sql;
  }
```

This transforms a query such as `SELECT * FROM users WHERE password = ?` into `SELECT * FROM users WHERE password = "secret"`.

I think that if we remove the `format`, we can record the `query` directly into the values attribute as desired without capturing sensitive data.

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -353,9 +355,19 @@ export class MySQLInstrumentation extends InstrumentationBase<
 
         span.setAttribute(
           SemanticAttributes.DB_STATEMENT,
-          getDbStatement(query, format, values)
```
blumamir, line 356: Since we don't use the `format` anymore, should we remove it from the function parameters as well?

plugins/node/opentelemetry-instrumentation-mysql/src/utils.ts
```diff
@@ -78,22 +78,25 @@ function getJDBCString(
  */
 export function getDbStatement(
   query: string | Query | QueryOptions,
-  format: (
-    sql: string,
-    values: any[],
-    stringifyObjects?: boolean,
-    timeZone?: string
-  ) => string,
   values?: any[]
```
blumamir, line 81: We no longer use `values` can be removed

plugins/node/opentelemetry-instrumentation-mysql/src/types.ts
```diff
@@ -16,4 +16,10 @@
 
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
 
-export type MySQLInstrumentationConfig = InstrumentationConfig;
+export interface MySQLInstrumentationConfig extends InstrumentationConfig {
+  /**
+   * If true, an attribute containing the query's parameters will be attached
+   * the spans generated to represent the query.
+   */
+  enhancedDatabaseReporting?: boolean;
```
blumamir, line 24: please also add to README.md

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
```diff
@@ -154,6 +155,89 @@ describe('mysql@2.x-Tracing', () => {
...
+      context.with(trace.setSpan(context.active(), span), () => {
+        const query = connection.query(
+          'SELECT * FROM `books` WHERE `author` = "David"'
+        );
+        query.on('end', () => {
+          const spans = memoryExporter.getFinishedSpans();
+          assert.strictEqual(
+            spans[0].attributes[AttributeNames.MYSQL_VALUES],
+            ''
+          );
```
blumamir, line 185: If there are no values, should we record this attribute?

plugins/node/opentelemetry-instrumentation-mysql/src/utils.ts
```diff
@@ -110,6 +113,11 @@ export function getSpanName(query: string | Query | QueryOptions): string {
   return query.split(' ')[0];
 }
 
+export function arrayStringifyHelper(arr: Array<unknown> | undefined): string {
+  if (arr) return '[' + arr.toString() + ']';
```
blumamir, line 117: nit: preferable to use template literals:
```js
return `[${arr.toString()}]`
```

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
haddasbronfman: good question that I asked myself too. To my opinion, the answer is yes, because otherwise, if we don't present this, the user might think there is a problem (because he doesn't see the parameter at all). What do you think?

plugins/node/opentelemetry-instrumentation-mysql/src/types.ts
```diff
@@ -16,4 +16,10 @@
 
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
 
-export type MySQLInstrumentationConfig = InstrumentationConfig;
+export interface MySQLInstrumentationConfig extends InstrumentationConfig {
+  /**
+   * If true, an attribute containing the query's parameters will be attached
+   * the spans generated to represent the query.
```
blumamir, line 22: nit: could be great to align this text with the one in the README :)

plugins/node/opentelemetry-instrumentation-mysql/src/instrumentation.ts
```diff
@@ -353,9 +341,19 @@ export class MySQLInstrumentation extends InstrumentationBase<
...
+          getDbStatement(query)
         );
 
+        const instrumentationConfig: MySQLInstrumentationConfig =
+          thisPlugin.getConfig();
+
+        if (instrumentationConfig.enhancedDatabaseReporting) {
+          span.setAttribute(
```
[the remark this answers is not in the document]
haddasbronfman: good point. done

plugins/node/opentelemetry-instrumentation-mysql/src/types.ts
haddasbronfman: done

blumamir (approved):

PR #1342, opened 2023-01-03
fix(instrumentation-net): make tls span parent of net span

blumamir:

- In `@opentelemetry/instrumentation-net`, when tls is used, 2 spans are produced: one for tls and one for network. They are currently siblings to each other and the trace structure does not indicate anything about their relationship, even though they relate to the same tls transaction. In the lack of ambient context (HTTP request span / db driver span) they belong to 2 different traces, as the `net` span starts in an empty ambient context, thus becomes a root span.

- When starting a new tls span, wrap the creation of net span with the tls context so that they are parent-child. Updated the existing tests to assert this behavior


plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts
```diff
@@ -112,7 +112,12 @@ export class NetInstrumentation extends InstrumentationBase<Net> {
   ) {
     const tlsSpan = this.tracer.startSpan('tls.connect');
 
-    const netSpan = this._startSpan(options, socket);
+    const netSpan = context.with(
+      trace.setSpan(context.active(), tlsSpan),
+      () => {
```
Flarna, line 117: you could avoid this closure by passing the context directly down to `startSpan`. But maybe that's not worth the effort.

I wonder also if the net or tls span should be set as active when calling the original to ensure that e.g. a dns request done during results in a correct nested span.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts
blumamir: > you could avoid this closure by passing the context directly down to `startSpan`. But maybe that's not worth the effort.

I agree but that would require propagating it down to `_startSpan` and then to `_startGenericSpan`, `_startIpcSpan`, `_startTcpSpan` and I thought it would be cleaner this way

> I wonder also if the net or tls span should be set as active when calling the original to ensure that e.g. a dns request done during results in a correct nested span.

That also needs to be fixed (in many instrumentations actually, not only in net). I wrote it but had a hard time implementing unit test to assert the behavior, so decided to start with the low-hanging fruits :)

Flarna: That's fine. followup is ok.

PR #1343, opened 2023-01-05
fix(pg): update requireParentSpan to skip instrumentation when parent not present

ryan-codaio:

I previously added a `requireParentSpan` config flag that can be used to suppress instrumentation if there is no active trace (see #1199). The implementation created a no-op span, imitating the [implementation for the requireParent flags in instrumentation-http](https://github.com/open-telemetry/opentelemetry-js/blob/597ea98e58a4f68bcd9aec5fd283852efe444cd6/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L665-L683), then continued running instrumentation code as usual.

This doesn't seem to be working well in practice. There are cases where the pg instrumentation creates a (no-op) span, then does something else that tries to create another span. The second invocation sees that a parent span exists, so it creates a real span. But this span is created as a _sibling_ of the no-op span, not as a child, so then it gets emitted as a real trace. For example, I saw cases where BoundPool.connect created a (no-op) span and then called Client.connect, which saw an existing span in the context and subsequently created a (real) span.

I considered modifying `startSpan` to create real spans as children of whatever `currentSpan` was returned by `trace.getSpan(context.active())`; that way, the spans would get dropped as children of a no-op span. But it doesn't look like other opentelemetry instrumentation libraries do this often. I'm not very familiar with opentelemetry so I'm not sure if this is the right thing to do. It seems like a context can have multiple sibling spans, so it seems like making the span a child of `trace.getSpan(context.active())` could end up creating a span as a child of the wrong parent span?

I changed the `requireParentSpan` implementation to avoid running instrumentation on calls when a parent span is required but doesn't exist. This way, no span is created at all. This matches the implementation in [instrumentation-ioredis](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/8b8bfebdd6b4f43a8df540979874a6c01c999957/plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts#L111-L113).

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
@@ -269,6 +269,24 @@ describe('pg-pool', () => {
         assert.strictEqual(resNoPromise, undefined, 'No promise is returned');
       });
     });
+
+    it('should not generate traces when requireParentSpan=true is specified', async () => {
+      // The pool gets shared between tests. We need to create a separate one
+      // to test cold start, which can cause nested spans
+      const newPool = new pgPool(CONFIG);
+      create({
+        requireParentSpan: true,
+      });
+      memoryExporter.reset();
+      const client = await newPool.connect();
+      try {
+        await client.query('SELECT NOW()');
+      } finally {
+        client.release();
+      }
+      const spans = memoryExporter.getFinishedSpans();
+      assert.strictEqual(spans.length, 0);
+    });
   });
 
   describe('#pool.query()', () => {
--- a/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
@@ -20,7 +20,14 @@ import {
   safeExecuteInTheMiddle,
 } from '@opentelemetry/instrumentation';
 
-import { context, diag, trace, Span, SpanStatusCode } from '@opentelemetry/api';
+import {
+  context,
+  diag,
+  trace,
+  Span,
+  SpanStatusCode,
+  SpanKind,
+} from '@opentelemetry/api';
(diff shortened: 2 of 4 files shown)


ryan-codaio:
Hi, would I be able to get a review on this PR? Thanks in advance!

ryan-codaio:
Hi @dyladan, sorry for the ping but it seems like @Rauno56 may unavailable, and it looks like you're the most active maintainer. Do you know of someone else who might be able to review this PR? It's been sitting for a month and I'm hoping to be able to get it merged 🙏

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -269,6 +269,24 @@ describe('pg-pool', () => {
...
     });
+
+    it('should not generate traces when requireParentSpan=true is specified', async () => {
+      // The pool gets shared between tests. We need to create a separate one
+      // to test cold start, which can cause nested spans
+      const newPool = new pgPool(CONFIG);
+      create({
+        requireParentSpan: true,
+      });
+      memoryExporter.reset();
```
blumamir, line 280: is it needed?
looks like already happens at [`afterEach`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/ff109b77928cc9a139a21c63d6b54399bb017fa4/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts#L171)

blumamir (approved):
LGTM

sorry about the time it took to review

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
ryan-codaio: You're right, thanks for the catch. I just removed this.

ryan-codaio:
Thank you so much for the review!

PR #1345, opened 2023-01-05
feat(instrumentation-dataloader): add dataloader name to span names

henrinormak:
After some weeks of using the dataloader instrumentation, it is clear that the spans it produces are currently too vague and not clearly identifiable, which in turn leads to confusion.

Added a possibility to include a generated name for each dataloader in the spans that it creates. Unfortunately there is no identifier clearly present on the dataloaders, so this is a compromise (i.e give the application developer a way to name the dataloaders themselves).

This solution should be flexible enough to allow something as simple as returning a generated UUID (as the name is cached, it would then at least allow separating dataloader spans by which dataloader they are coming from), to solutions more complex where the dataloader object can be used to determine a better name (for example by looking up some global context where said reference might also be present).

Open to discussion whether there is a better approach to this, but couldn't immediately figure one out.


SimenB:
https://github.com/graphql/dataloader/pull/326 has been merged (but not released ATM), so might be better to adjust this to look at that? Instead of requiring a function here

henrinormak:
> [graphql/dataloader#326](https://github.com/graphql/dataloader/pull/326) has been merged (but not released ATM), so might be better to adjust this to look at that? Instead of requiring a function here

Awesome, was not aware of that, will refactor to leverage that

henrinormak:
Refactored to use the new approach, pending Dataloader release for the types update, but functionally should be complete.

henrinormak:
Filed an issue with dataloader, waiting for the release there to proceed with this - https://github.com/graphql/dataloader/issues/328

henrinormak:
This should now be properly up to date and ready for merging, cc @SimenB 🙇

plugins/node/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -89,40 +90,56 @@ export class DataloaderInstrumentation extends InstrumentationBase {
     return hasParentSpan || !config.requireParentSpan;
   }
 
+  private getSpanName(
+    dataloader: DataloaderInternal,
+    operation: 'load' | 'loadMany' | 'batch'
+  ): string {
+    const dataloaderName = dataloader._name;
```
blumamir, line 97: I might be missing something obvious, but in the linked PR it seems that it should be `name` (with no underscores):

```js
  it('allows giving the loader a name', () => {
    expect(new DataLoader(() => []).name).toBeNull();
    expect(new DataLoader(() => [], {}).name).toBeNull();

    expect(new DataLoader(() => [], { name: 'Some name' }).name).toBe(
      'Some name'
    );
  });
```

```js
    /**
     * The name given to this `DataLoader` instance. Useful for APM tools.
     *
     * Is `null` if not set in the constructor.
     */
    name: string | null;
```

Is it supposed to be `_name` in the instrumentation?

henrinormak: Hmm, I thought it was only added to the config and not to the actual instances (based on the types in https://github.com/graphql/dataloader/blob/main/src/index.d.ts#L18), but your code snippet suggests that the types are just not correct.

I'll open a PR against dataloader to augment that, which will allow avoiding the underscored one in this instrumentation here.

blumamir: Cool thanks, but how come the tests passes with `_name`? how is it able to extract the name from the `DataLoader ` instance and make tests such as this one pass?
```js
    it('correctly uses dataloader name', async () => {
      const namedDataloader = new Dataloader(
        async keys => keys.map((_, idx) => idx),
        { name: 'test-name' }
      );
      assert.strictEqual(await namedDataloader.load('test'), 0);

      // We should have exactly two spans (one for .load and one for the following batch)
      assert.strictEqual(memoryExporter.getFinishedSpans().length, 2);
      const [batchSpan, loadSpan] = memoryExporter.getFinishedSpans();

      assert.strictEqual(loadSpan.name, 'dataloader.load test-name');
      assert.strictEqual(batchSpan.name, 'dataloader.batch test-name');
    });
```

blumamir: It seems to be storing the value in `this.name` [here](https://github.com/graphql/dataloader/blob/0a09c563d40f027ecf875d72c84c1c4099377bf6/src/index.js#L60). So what will `dataloader._name` return?

henrinormak: Opened PR there - https://github.com/graphql/dataloader/pull/334

henrinormak: So the instrumentation currently wraps the constructor and in there stores the name into the underscored variant - https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1345/files#diff-db5fdab7eeb21bef0d4988932a6f1dbfb063ab156a3a9977301c2e529b530e58R121, basically duplicating the logic in the actual constructor

blumamir: > > It seems to be storing the value in `this.name` [here](https://github.com/graphql/dataloader/blob/0a09c563d40f027ecf875d72c84c1c4099377bf6/src/index.js#L60). So what will `dataloader._name` return?
> 
> So the instrumentation currently wraps the constructor and in there stores the name into the underscored variant - https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1345/files#diff-db5fdab7eeb21bef0d4988932a6f1dbfb063ab156a3a9977301c2e529b530e58R121, basically duplicating the logic in the actual constructor

I see, makes sense :)
So if I understand correctly, we can already remove the `_name` handling and read it directly from `.name`, right?
And https://github.com/graphql/dataloader/pull/334 will allow us to also remove the type from `DataloaderInternal`?

henrinormak: Yes, but I'd rather wait for that small typing fix to get released and not have to declare the type in the instrumentation, or do you reckon we should proceed here and once the fix is released in dataloader do the cleanup?

henrinormak: I changed it so this could be merged before, but I'm also OK with waiting as hopefully the dataloader PR gets merged soon-ish.

blumamir: depends on you, but I do think that even after the type is fixed in the `dataloader` package, `test-all-versions` might still run into versions where this property is not present, and we might have to include it for those cases. WDYT?

henrinormak:
Let's go ahead and merge, not entirely sure how much time Dataloader repo gets from maintainers at the moment, once it gets merged there, I will come back to this and get rid of the TODO

plugins/node/instrumentation-dataloader/src/instrumentation.ts
henrinormak: True, our internal type is indeed more "backwards compatible" as it also includes undefined (i.e case of older package), so I guess let's go ahead and merge and if there is a way to clean it up a bit in future, I'll return.

plugins/node/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -89,40 +92,53 @@ export class DataloaderInstrumentation extends InstrumentationBase {
     return hasParentSpan || !config.requireParentSpan;
   }
 
+  private getSpanName(
+    dataloader: DataloaderInternal,
+    operation: 'load' | 'loadMany' | 'batch'
+  ): string {
+    const dataloaderName = dataloader.name;
+    if (dataloaderName === undefined || dataloaderName === null) {
```
blumamir, line 100: This can be written as shortcut 
```suggestion
    if (dataloaderName == null) {
```

or even
```suggestion
    if (!dataloaderName) {
```
to also catch empty strings

blumamir (approved):

PR #1355, opened 2023-01-15
ci: fix lint

blumamir:
CI currently fails on lint workflow for a few days, and we can't merge approved PRs. I guess some transitive dependency was updated and change the prettier config/functioning.

I do not have time to understand where it comes from, but run `lint:fix` on the repo to make CI green again.


dyladan (approved):
I agree with these changes anyway

blumamir:
Looks like it's related to [this PR](https://github.com/prettier/prettier/pull/14042) in prettier, which now requires parentheses around `typeof x`. This was published as patch version `2.8.2` which was picked up by `gts` that has a caret range dependency on prettier.

I guess it should have been considered a breaking change and published on a major version, but maybe prettier uses a different workflow. Anyway it is fixed now in this repo

PR #1358, opened 2023-01-17
fix(deps): update otel core experimental to ^0.35.1

Flarna:
Take changes from https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1367 as base

Add a commit to adapt fastify tests to fit http instrumentation.

http instrumentation ends server spans later which results in exporting them later.
Adapt the tests to fit the new sequence. and additionally adapt fastify tests 

Refs: https://github.com/open-telemetry/opentelemetry-js/pull/3407

fyi: @legendecas


legendecas (approved):
Thanks for the fix!

Flarna:
seems the browser test job fails during compilation because of missing grpc types. Log output in CI is missing since a while for failures like this. Don't know why.

Had no time yet to look into details why node 14 unittest job works browser doesn't work. Likely some lerna hoisting issue.

SimenB:
FWIW I get a type error in our project as well due to missing grpc types when attempting to upgrade `@opentelemetry/instrumentation-grpc` to `0.35`.

Flarna:
grpc instrumentation doesn't has @grpc/grpc-js as dependency but it seems the types are required.
Likely @grpc/grpc-js was present in node_modules in the past because of some other package depending on it.

This type leaks/deps are a constant problem...

SimenB:
Upstream is fixed - rebase this?

Flarna:
rebased and CI is green now (besides the usual, unrelated coverage failures).

@SimenB any more details why it fails in your setup? maybe you miss lib Dom in your build setup?

SimenB:
Yeah, this is a node module not for the browser. 1.9.0/0.35.1 does not have the same issue

SimenB:
I opened https://github.com/open-telemetry/opentelemetry-js/issues/3580

dyladan (approved):

dyladan:
Looks good. The release will for sure fail because of so many packages (rate limiting) but i'll manually do it after my next meeting

PR #1364, opened 2023-01-27
feat(detector-gcp): collect hostname resource attribute from GCP Metadata API

mwear:

This PR is mostly for convenience. When working on [collecting host-id](https://github.com/open-telemetry/opentelemetry-js/pull/3575) for non-cloud environments, I was looking at the AWS EC2 Detector and the GCP Detector. I noticed that the EC2 Detector collects both host.name and `host.id`, and the GCP Detector collects only `host.id`. I updated the GCP Detector to collect host.name for congruence with the EC2 Detector and also so that GCP users will not have to use the Host Detector, or worry about Detector ordering when collecing `host.name` and `host.id`.

- Collect `host.name` resource attribute in GCP Detector


detectors/node/opentelemetry-resource-detector-gcp/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### Features
+
+* feat: collect `host.name` resource attribute from GCP Metadata API (#[1364](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1364))
+
```
blumamir, line 6: I think this is auto-generated in contrib.

blumamir (approved):

naseemkullah (approved):

PR #1366, opened 2023-01-30
chore(user-interactions): remove component and user agent span attributes

blumamir:
Removed 2 attributes from user-interactions web instrumentation:

1) `component` - this attribute is not speced, and the relevant data is already present in the scope name.
2) `http.user_agent` - this attribute is recording `navigator.userAgent` which is static to the application and thus should be a resource attribute. The [BrowserDetector](https://github.com/open-telemetry/opentelemetry-js/blob/279458e7ddf16f7ddca5fe60c78672e05fafce66/packages/opentelemetry-resources/src/detectors/BrowserDetector.ts#L34) already record this value into `process.runtime.version`, and it is thus duplicated on the span


blumamir:
related: https://github.com/open-telemetry/opentelemetry-specification/pull/447

haddasbronfman (approved):

PR #1370, opened 2023-02-01
feat(instrumentation-lambda): Flush MeterProvider at end of handler

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

The Lambda instrumentation wrapper does not flush metrics at the end of handler invocation.

Flushes the `MeterProvider` at the end of handler invocation.


pichlermarc (approved):
Thanks for adding this :slightly_smiling_face: Looks good to me. :slightly_smiling_face: 
Failing tests look unrelated to these changes, I think a rebase should fix those. :)

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -294,6 +314,17 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
       );
       callback();
     }
+    if (this._metricForceFlusher) {
+      this._metricForceFlusher().then(
+        () => callback(),
```
legendecas, line 319: Would you mind adding a test to verify that this callback is only invoked once when the tracer and meter are both registered?

Aneurysm9: good catch!  I've added a test and it's failing.  Working on a fix now.

```
1) force flush
       should callback once after force flush providers:

2 !== 1

      + expected - actual

      -2
      +1
      
      at Context.<anonymous> (test/integrations/lambda-handler.force-flush.test.ts:236:12)
```

pichlermarc: Good catch, thanks for following up on this!

Aneurysm9: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1384 should fix this.

PR #1374, opened 2023-02-06
chore: fix contrib build

kobi-co:
contrib CI and build is broken for several days due to 2 dependencies that were released lately and contained breaking changes:

## MongoDB
MongoDB instrumentation supports both v3 and v4. for v3 we depend on the `@types/mongodb@3` package, which in turn depends on `@types/bson@*` (This is a bug in definitely typed!). MongoDB released v5 last week which contains breaking changes (removing the `ObjectID` export). Since the types dependency does not limit a version range, the new version is used and since it's not compatible with v3 typescript fails.

See [related issue](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/64241) in defiantly typed.

To fix this, we have to depend on the real version of `bson` and not the incompatible latest.

Things we tried and didn't work:
- remove instrumentation support for mongodb@3 - it has 1M downloads in npm, more than v4
- fix the wrong dependency range in definitely typed - since v4 is distributed with types, the package is deprecated and removed from the code base. This fix will need to be applied to v3 of `@types/mongodb` and we couldn't find any process for updating and releasing types packages for older major versions.
- use [npm "overrides"](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) - introduced in npm@7 and we support node14 which uses npm@6. Thus not feasible at the moment.
- copy the types declarations into the package and drop the dependency on the buggy `@types/mongodb` altogether - this failed to compile and we were not knowledgable enough in ts to figure out how to make it work.

The fix that resolved the issue was to introduce our own dependency on `@types/bson` which links to the correct version. 

## Winston
Winston depends on [`logform@^2.4.0`](https://github.com/winstonjs/winston/blob/bf14c3bbb172ff36fa940703b699e408c21b5194/package.json#L31) which published a new version yesterday [`v2.5.0`](https://www.npmjs.com/package/logform/v/2.5.0). 

This version introduced types in the public API that are not a dependency of the package, thus breaking everyone using typescript. See more discussion in [this issue](https://github.com/winstonjs/logform/issues/242) and a fix (not yet merged) in [this PR](https://github.com/winstonjs/logform/pull/243). Since the PR is not yet merged and the fix is not published, we need to fix this in our codebase to make our build work. In the future, when winston fixes the issue, we can remove the dependency on `@types/triple-beam`.

The other issue is that the new `logform` version introduced [this change](https://github.com/winstonjs/logform/pull/240) which is only implemented in typescript in [this PR](https://github.com/microsoft/TypeScript/pull/44512) which is not part of our current dependency `4.3.5`. To fix this, we had to bump the dependency version (to `4.5.5`) which also meant that all the `catch` clauses were now `unknown` instead of `any` [introduced in `typescript@4.4`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables). So for the code to transpile correctly, we had to type all those exceptions as `any` which is not great, but it's the best we can do for now. This can be fixed in a followup PR if someone is up for it.

CI is now green which will allow all other PRs to get merged.


package.json
```diff
@@ -58,6 +58,9 @@
     "lerna-changelog": "2.2.0",
     "typescript": "4.3.5"
   },
+  "overrides":{
+    "@types/bson": "4.0.5"
+  },
```
blumamir, line 63: This needs to also be present in the `package.json` for the mongo instrumentation, so if lerna hoisting is not applied the correct version is still used

blumamir (approved):

dyladan:
Looks like this isn't supported in node 14. Another possibility is to drop mongo v3 right?

blumamir:
> Looks like this isn't supported in node 14. Another possibility is to drop mongo v3 right?

Drop the support from the instrumentation or just not run the test?

Looks like mongo latest 3 - `3.7.3` is [downloaded in npm](https://www.npmjs.com/package/mongodb?activeTab=versions) 1M times a week (more than v4)

kobi-co:
> Looks like this isn't supported in node 14. Another possibility is to drop mongo v3 right?

The only other option I see besides that is to patch `@types/mongodb` somehow and change the bson dependency there to a valid version instead of `*`. 
Which seems quite difficult as this module already been deleted from DefinitelyTyped

blumamir (approved):

blumamir:
@open-telemetry/javascript-approvers @open-telemetry/javascript-maintainers  since our build is currently broken, please have a look so we can merge fast. Thanks!

dyladan:
Please note that this will raise the minimum required TS version to 4.4 because of a change in TS 4.4 which affects the output that causes older versions to fail to read it. The SDK and API already have 4.4.4 so this should be no impact to end users.

In a future SIG meeting we will discuss a more general policy for upgrading typescript (not this week or next due to maintainer vacations).

wbt:
The reference PR has been merged into logform.

PR #1375, opened 2023-02-06
fix(instrumentation-fs): fix instrumentation of `fs/promises`

mhassan1:

This PR fixes `@opentelemetry/instrumentation-fs` when used with `fs/promises`.

This PR adds logic to `@opentelemetry/instrumentation-fs` that instruments `fs/promises`. This will properly instrument `fs` methods when they come from `fs/promises`; currently, that instrumentation is not happening.


blumamir:
I just merged #1332 and it seems to create conflicts in this PR, sorry about that

plugins/node/instrumentation-fs/test/fsPromises.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      });
+
+      selection.forEach(([name, args, result]) => {
+        promiseTest(name as FPMember, args, result, []);
+      });
+    });
+  });
+}
+
+const assertSpans = (spans: ReadableSpan[], expected: any) => {
```
haddasbronfman, line 154: This function is the same as in fs.test.ts file. maybe it is better to put both of them in a 'utils.ts' file.

plugins/node/instrumentation-fs/test/fsPromises.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      .map((s: any) => `"${s.name}"`)
+      .join(', ')})`
+  );
+
+  spans.forEach((span, i) => {
+    assertSpan(span, expected[i]);
+  });
+};
+
+const assertSpan = (span: ReadableSpan, expected: any) => {
```
haddasbronfman, line 168: same as 'assertSpans' (better to put in utiles.ts file)

plugins/node/instrumentation-fs/test/fsPromises.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+
+    const promiseTest: TestCreator = (
+      name: FPMember,
+      args,
+      { error, result, resultAsError = null },
+      spans
+    ) => {
+      const rootSpanName = `${name} test span`;
+      it(`promises.${name} ${error ? 'error' : 'success'}`, async () => {
```
haddasbronfman, line 77: Maybe I'm missing something, but isn't this test already exist in the file 'fs.test.ts' ?

mhassan1: This is testing `fs/promises`, whereas `fs.test.ts` is testing `fs.promises`.

plugins/node/instrumentation-fs/src/instrumentation.ts
```diff
@@ -55,7 +56,10 @@ export default class FsInstrumentation extends InstrumentationBase<FS> {
     super('@opentelemetry/instrumentation-fs', VERSION, config);
   }
 
-  init(): InstrumentationNodeModuleDefinition<FS>[] {
+  init(): (
+    | InstrumentationNodeModuleDefinition<FS>
+    | InstrumentationNodeModuleDefinition<FSPromises>
+  )[] {
```
dyladan, line 62: nit: i think this type is more explicit

```suggestion
  init(): [InstrumentationNodeModuleDefinition<FS>, InstrumentationNodeModuleDefinition<FSPromises>] {
```

dyladan (commented):
Looks like this is just applying the existing patch from `fs.promises` and adding tests. Looks good to me

dyladan (approved):

plugins/node/instrumentation-fs/test/fsPromises.test.ts
```diff
@@ -0,0 +1,153 @@
...
+import * as assert from 'assert';
+import Instrumentation from '../src';
+import * as sinon from 'sinon';
+import type * as FSPromisesType from 'fs/promises';
+import tests, { FsFunction, TestCase, TestCreator } from './definitions';
+import type { FPMember, EndHook } from '../src/types';
+import { assertSpans, makeRootSpanName } from './utils';
+
+const supportsPromises =
+  parseInt(process.versions.node.split('.')[0], 10) >= 14;
```
blumamir, line 32: The repo supports node >= 14 anyway, so I guess this is not needed?

haddasbronfman:
Sorry, I merged this PR but now I see there is still a comment about it: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1375#discussion_r1118094721. If you want to address this comment, I suggest to do it on another PR.

plugins/node/instrumentation-fs/test/fsPromises.test.ts
mhassan1: Resolved by https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1413

PR #1377, opened 2023-02-06
fix(instrumentation-dns): fix instrumentation of `dns/promises`

mhassan1:

This PR fixes `@opentelemetry/instrumentation-dns` when used with `dns/promises`.

This PR adds logic to `@opentelemetry/instrumentation-dns` that differentiates between `dns` and `dns/promises`. This will properly instrument `lookup` when it comes from `dns/promises`; currently, that instrumentation is not happening.


plugins/node/opentelemetry-instrumentation-dns/test/integrations/dns-slash-promises-lookup.test.ts
```diff
@@ -0,0 +1,223 @@
...
+import * as utils from '../utils/utils';
+import { assertSpan } from '../utils/assertSpan';
+import { SpanStatusCode } from '@opentelemetry/api';
+
+const memoryExporter = new InMemorySpanExporter();
+const provider = new NodeTracerProvider();
+provider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
+
+const supportsPromises =
+  parseInt(process.versions.node.split('.')[0], 10) >= 15;
```
blumamir, line 34: why >=15?
Is it not [added in `10.6.0`?](https://nodejs.org/docs/latest-v14.x/api/dns.html#dns_dnspromises_lookup_hostname_options)

mhassan1: `dns.promises` was added in 10.6.0, but `dns/promises` was [added in 15.0.0](https://nodejs.org/docs/latest-v15.x/api/dns.html#dns_dns_promises_api).

blumamir: right! i missed that :)

blumamir (approved):

blumamir:
BTW, the PR description mentions that currently, we get a "no original function lookup to wrap" warning. Why is that?

PR #1378, opened 2023-02-06
docs: Remove duplication in auto-instrumentations-node readme

passcod:
Just a silly little readme change that removes duplicates in the `@opentelemetry/auto-instrumentations-node` Supported instrumentations list.


metapackages/auto-instrumentations-node/README.md
```diff
@@ -76,17 +76,14 @@ registerInstrumentations({
 - [@opentelemetry/instrumentation-redis](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-redis)
 - [@opentelemetry/instrumentation-pg](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-pg)
 - [@opentelemetry/instrumentation-knex](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-knex)
-- [@opentelemetry/instrumentation-koa](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-koa)
```
MSNev, line 79: Looks like the `[@opentelemetry/instrumentation-http](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-instrumentation-http)` is also duplicated -- would you mind also removing that one.

Suggestion: Is it possible to remove the duplicates on 76, 77 rather than 87, 89 -- that way is still more in sorted order (unless there is a reason for the order)

passcod: I could instead sort the entire list? I didn't know if there was a reason it was sorted oddly.

MSNev (approved):

Flarna (approved):

blumamir (approved):

PR #1380, opened 2023-02-08
chore: move release-please action run

SimenB:

Looking at https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/4118780372/jobs/7111714212, the comments were posted to the PR almost 40 minutes before the packages were even built, not to mention when they were actually published

Moves the action posting comments to after the packages are built, in order for there to be less time between a comment saying "the update is published" and it actually being true.


dyladan (approved):
There will still be a time delay as the publish step itself takes quite some time, but overall I think this is an improvement

PR #1384, opened 2023-02-08
fix(instrumentation/aws-lambda): Ensure callback is only called once

Aneurysm9:
…regardless how many providers are flushed

Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

Defect identified in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1370#pullrequestreview-1289551665.

Combine all provider flush invocations to a single `Promise.all()` that can be resolved once.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
@@ -319,8 +319,7 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
       );
     }
 
-    Promise.all(flushers)
-      .then(callback, callback);
+    Promise.all(flushers).then(callback, callback);
   }
 
   private _applyResponseHook(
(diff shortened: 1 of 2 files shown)


pichlermarc (approved):
Thanks for following up on it - I should've caught that when reviewing the previous PR.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -303,28 +303,26 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
       diag.error(
         'Metrics may not be exported for the lambda function because we are not force flushing before callback.'
       );
-      callback();
     }
+
+    Promise.all(flushers).then(
+      () => callback(),
+      () => callback()
+    );
```
dyladan, line 325: Should the `err` be provided to the callback on line 324?

```suggestion
    Promise.all(flushers).then(
      () => callback(),
      (err) => callback(err)
    );
```

It would probably also work to do:
```suggestion
    Promise.all(flushers)
      .then(callback, callback);
```

dyladan (approved):
Looks good. Minor clarification nit

legendecas (approved):

PR #1390, opened 2023-02-12
chore(nestjs-core): remove types/vinyl-fs dependency

samimusallam:

Renovate is trying to bump the types/vinyl-fs dependency in [this PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1387) which is not used.

Ran `test-all-versions` to make sure it is not needed by older versions.


blumamir (approved):
LGTM
Thanks for checking

PR #1395, opened 2023-02-14
fix(postgres): pass 'arguments' to the connect callback

### LINKED ISSUE #991: instrumentation-pg used with pg-query-stream causes user application crash
I use this versions:
"pg": "^8.7.3",
"pg-cursor": "^2.7.3",
"pg-query-stream": "^4.2.3"
"@opentelemetry/api": "^1.0.4",
"@opentelemetry/exporter-jaeger": "^1.0.1",
"@opentelemetry/instrumentation-amqplib": "^0.28.0",
"@opentelemetry/instrumentation-hapi": "^0.27.1",
"@opentelemetry/instrumentation-http": "^0.27.0",
"@opentelemetry/instrumentation-pg": "^0.28.0",

16.10

### What did you do?
i have configured opentelemetry with instrumentation-pg. It seems to work perfectly but not on streamed queries.

No errors on console, i exect results of streamed query

### What did you see instead?
When i perform a streamed query, i've got a 500 in my hapi handler.

The console output something like this:
`{
  traceId: '74b485fc07b9463d77990009eaa33d35',
  parentId: 'e7798cf16dc3587e',
  name: 'pg.query:unknown',
  id: 'd7a5c8567e99de95',
  kind: 2,
  timestamp: 1651240871135298,
  duration: 5962,
  attributes: {
    'db.name': 'measurements',
    'db.system': 'postgresql',
    'db.connection_string': 'jdbc:postgresql://persistence:5432/measurements',
    'net.peer.name': 'persistence',
    'net.peer.port': 5432,
    'db.user': 'postgres'
  },
  status: {
    code: 2,
    message: 'A query must have either text or a name. Supplying neither is unsupported.'
  },
  events: []
}`

and this stack trace

`Debug: internal, implementation, error
TypeError: Cannot read properties of undefined (reading 'on')
at Database.queryStream (/main/node_modules/@a2asc/common-persistence/src/database.js:115:12)
at processTicksAndRejections (node:internal/process/task_queues:96:5)
at async Repository.findInDatesRangeByDevicesStream (/main/app/measurements/persistence/repository.js:110:12)
at async handler (/main/app/measurements/web/controller.js:178:36)
at async exports.Manager.execute (/main/node_modules/@hapi/hapi/lib/toolkit.js:60:28)
at async Object.internals.handler (/main/node_modules/@hapi/hapi/lib/handler.js:46:20)
at async exports.execute (/main/node_modules/@hapi/hapi/lib/handler.js:31:20)
at async Request._lifecycle (/main/node_modules/@hapi/hapi/lib/request.js:371:32)
at async Request._execute (/main/node_modules/@hapi/hapi/lib/request.js:281:9)`

If i disable instrumentation-pg it works perfectly

haddasbronfman:
This PR comes instead of this one: #1224 

Turns out that pg actually passes the client to connect callback. That is undocumented but relied upon by knex.
Fixes #991

Added the 'arguments' object to the callback's signature and pass it through to the patched callback.


plugins/node/opentelemetry-instrumentation-pg/src/internal-types.ts
```diff
@@ -46,7 +46,10 @@ export type PgPoolCallback = (
   done: (release?: any) => void
 ) => void;
 
-export type PgErrorCallback = (err: Error) => void;
+// Even though the documented signature for connect() callback is `(err) => void`
+// `pg` actually also passes the client if the connection was successful and some
+// packages(`knex`) might rely on that
+export type PgErrorCallback = (err: Error, client?: pgTypes.Client) => void;
```
blumamir, line 52: I wonder if we actually need this type in the instrumentation. Is it used for anything meaningful? Can we replace where it is used with `Function` which signal that we don't care about the arguments of the function?

haddasbronfman: good point, but on the other hand, why shouldn't we always use `Function`? It still gives some readability. Anyway, I removed it in this case since we don't want to be committed to the callback signature.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -259,6 +259,16 @@ describe('pg', () => {
       assert.strictEqual(res, undefined, 'No promise is returned');
     });
 
+    it('should pass the client connection object in the callback function', done => {
+      connClient.connect(function (err: Error, conn?: pg.Client) {
+        //the 'connect' function passes the client object as an argument
+        //in the callback function: https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/client.js#L282
+        assert(arguments[1] === connClient);
+        assert(arguments[1] === conn);
```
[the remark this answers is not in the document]
haddasbronfman: Thanks, I thought about this - making sure that connClient === conn is like verifying postgres works. because they send the "this" (aka the client) as a parameter to the function. So I left only the first assertion.

plugins/node/opentelemetry-instrumentation-pg/src/internal-types.ts
blumamir: I think we should declare a function type when it has a meaning in the application, e.g. when we depend on the actual parameters or return value type in the source code. Generic `Function`, as I understand it, means that we get a value that is a function and we don't care about its parameters

blumamir: Maybe we could declare a function where the first argument is `err: Error` and then [rest parameters typed as `any`](https://www.typescriptlang.org/docs/handbook/2/functions.html#rest-parameters-and-arguments):

```ts
export type PgErrorCallback = (err: Error, ...otherArguments: any[]) => void;
```

I think this might be the correct typing for the callback function, but feel free to choose whichever syntax you prefer.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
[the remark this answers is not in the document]
haddasbronfman: If I assert this: `assert(conn === connClient)` this is like asserting what postgres does (here: https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/client.js#L290). What I want to check, is if **our patched function** passes the client also. This is why I check `arguments[1] === connClient`.

blumamir (approved):

PR #1399, opened 2023-02-16
fix: remove component attribute from instrumentations

blumamir:

- The "component" attribute [was removed from semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/pull/447) back in Feb 2020, but some instrumentations were written before this change and still report this attribute.

- remove the "component" attribute from "@opentelemetry/instrumentation-document-load" and "@opentelemetry/instrumentation-long-task" as well as clean up some related dead code in other places

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 export enum AttributeNames {
-  COMPONENT = 'graphql',
   SOURCE = 'graphql.source',
   FIELD_NAME = 'graphql.field.name',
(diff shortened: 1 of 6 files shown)


Flarna (approved):

haddasbronfman (approved):

martinkuba (commented):
Before merging this I want to double check that there is another way of identifying the document load spans. I don't think there are any semantic conventions specifically for the browser instrumentations (unlike the database one). Users might be relying on the component to be there for these spans.

Flarna:
There is instrumentation scope on each span which identifies the tracer. Each instrumentation has a tracer so this should be an exact replacement

martinkuba:
@Flarna That's a good point. I think it still might break some queries for some users, but this are experimental instrumentations, and as long as there is some way to identify I am 👍

martinkuba (approved):

PR #1414, opened 2023-02-28
Add custom attributes document load

Abinet18:

Adding custom attributes to the Document load instrumentation spans (documentLoad, documentFetch, resourceFetch)

This adds a configuration to the document load instrumentation which takes a map of span type to a function that can access the corresponding span and add some attributes to it. By providing the function as a config while registering the DocumentLoad instrumentation will allow for such addition of some vendor specific attributes to the spans.


MSNev (approved):

martinkuba:
Could this not be accomplished by using a custom processor?

Abinet18:
I see that this can be achieved with custom processors. The difference is that with custom processor , all spans will have to be intercepted and modified accordingly , filtering based on some span properties. This will only apply to document load instrumentation spans. There is a similar implementation for fetch and xhr to add custom attributes. If using custom processors is the choice for all scenarios, we may have to deprecate the fetch and xhr apply custom attributes functionality too.

martinkuba:
I see, makes sense since there is already a precedence for this.

All the [other instrumentations](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js%20applyCustomAttributesOnSpan%3F&type=code) use the name `applyCustomAttributesOnSpan`. I would recommend that we use the same name for consistency.

martinkuba (approved):

Abinet18:
Can this be merged ?

PR #1416, opened 2023-03-01
ci: increase memory for browser tests

blumamir:
Browser test are currently failing because of "JavaScript heap out of memory".

This PR increases the memory for the node process in CI browser-test job. I don't know if anything changed that now consumes more memory, or if it makes sense that browser tests consume so much.


blumamir:
Apparently, the problem has to do with source map generation and merging. Increasing the memory did not help with the CI run

PR #1417, opened 2023-03-01
ci: run browser tests with node18

blumamir:
Check if browser tests works with node 18. They are currently fails for node14 due to "out of memory". Might be related to [this issue](https://github.com/nodejs/node/issues/38300) in old node versions


pichlermarc (approved):
I think even with the original reason gone, we can still go ahead with PR - updating it to run on node 18 won't hurt. WDYT @blumamir ? :slightly_smiling_face:

rauno56 (approved):

blumamir:
Closing so we can address this after code coverage stabilizes.
Opened an issue to track: #1634

PR #1421, opened 2023-03-04
fix(pg): record correct db.connection_string when pg.Pool is configured by connectionString

### LINKED ISSUE #1415: instrumentation-pg records wrong db.connection_string when pg.Pool is configured by connectionString

```
├── @opentelemetry/api@1.4.0
├── @opentelemetry/exporter-jaeger@1.9.1
├── @opentelemetry/exporter-zipkin@1.9.1
├── @opentelemetry/instrumentation-http@0.35.1
├── @opentelemetry/instrumentation-pg@0.34.1
├── @opentelemetry/instrumentation@0.35.1
├── @opentelemetry/sdk-trace-base@1.9.1
├── @opentelemetry/sdk-trace-node@1.9.1
```

v16.19.0

### What did you do?

I set up auto instrumentation of an application using [pg.Pool](https://node-postgres.com/apis/pool). In the application, pg.Pool was configured by `connectionString` like this:

```js
const pool = new Pool({
  connectionString: "postgresql://postgres:postgres@localhost:54320/postgres"
})
```

Then I launched the application and found a trace having wrong value of `db.connection_string`.

Repro application is available at: https://github.com/orangain/opentelemetry-js-contrib/tree/spike-address-pg-pool-issue/examples/postgres

#### Complete recipe to reproduce

```sh
git clone -b spike-address-pg-pool-issue https://github.com/orangain/opentelemetry-js-contrib
cd opentelemetry-js-contrib/examples/postgres
npm install
# Launch PostgreSQL
npm run docker:start
# Launch Jaeger
docker run -d --name jaeger \
  -e COLLECTOR_ZIPKIN_HOST_PORT=:9411 \
  -e COLLECTOR_OTLP_ENABLED=true \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  -p 14250:14250 \
  -p 14268:14268 \
  -p 14269:14269 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.42
# Launch Server
npm run jaeger:server
```

After launching the server, open http://localhost:16686 in a browser, and you will see the Jaeger UI. Then select `unknown_service:node` from "Service" dropdown and click "Find Traces" button.

The span `pg-pool.connect` contains a tag `db.connection_string` and its value is `postgresql://localhost:54320/postgres`.

### What did you see instead?

The span `pg-pool.connect` contains a tag `db.connection_string` and its value is `postgresql://localhost:5432/`. The port number is wrong and the database name does not exist.

image

### Additional context

Note that the child span `pg.connect` contains a tag `db.connection_string` and its value is `postgresql://localhost:54320/postgres` as expected.

image

https://github.com/open-telemetry/opentelemetry-js-contrib/issues/782 is a similar issue but it seems to be a different situation.

orangain:

- Fixes #1415 

- When pg.Pool is configured by connectionString, parse and use it as attributes.
  - The parsing logic is much simpler than [the actual one](https://github.com/brianc/node-postgres/blob/pg-connection-string%402.5.0/packages/pg-connection-string/index.js#L11), but I believe this logic covers major use cases.


plugins/node/opentelemetry-instrumentation-pg/src/internal-types.ts
```diff
@@ -47,10 +47,11 @@ export type PgPoolCallback = (
...
 
 export interface PgPoolOptionsParams {
-  database: string;
-  host: string;
-  port: number;
-  user: string;
+  database?: string;
+  host?: string;
+  port?: number;
+  user?: string;
```
orangain, line 53: Actually, there four properties are not mandatory, so I changed them optional.

See:
- https://github.com/brianc/node-postgres/blob/pg-pool%403.5.2/packages/pg-pool/index.js#L64
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg-pool/index.d.ts#L12
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg/index.d.ts#L17-L21

PR #1424, opened 2023-03-09
fix(instrumentation-redis): do not use `Callback` from `@types/redis`

### LINKED ISSUE #1323: tsc compiling both redis and redis-4 when importing getNodeAutoInstrumentations
## Steps to Reproduce
- redis@4.0.2 or above installed in project
- `tsc --project tsconfig.json`

tsconfig file:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es6",
    "rootDir": "./src",
    "outDir": "./build",
    "esModuleInterop": true,
    "inlineSourceMap": true,
    "inlineSources": true
  }
}
```

## Expected Result
No errors

## Actual Result
```
> xxx1.0.0 build
> tsc --project ./tsconfig.prod.json && tsc --project ./tsconfig-prisma.json

../node_modules/@opentelemetry/instrumentation-redis/build/src/types.d.ts:8:26 - error TS2694: Namespace '"/home/xxx/xxx/node_modules/redis/dist/index"' has no exported member 'Callback'.

8     callback: redisTypes.Callback<unknown>;
                           ~~~~~~~~
```

### OpenTelemetry Setup Code

```
import { HoneycombSDK } from "@honeycombio/opentelemetry-node";

export const honeycombSdk: NodeSDK = new HoneycombSDK({
  instrumentations: [getNodeAutoInstrumentations()]
});
```

pichlermarc:

Having both @opentelemetry/instrumentation-redis and redis@^4.0.2 (common when using `@opentelemetry/autoinstrumentations-node`) installed causes typescript compilation errors. 

Fixes #1323

- This PR fixes the above problem by not using the removed `Callback` type in the public API.
- Also updates the link to command.js in the comment.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-redis/src/types.ts
+++ b/plugins/node/opentelemetry-instrumentation-redis/src/types.ts
@@ -16,15 +16,14 @@
 
 import { Span } from '@opentelemetry/api';
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
-import type * as redisTypes from 'redis';
 
 // exported from
-// https://github.com/NodeRedis/node_redis/blob/master/lib/command.js
+// https://github.com/redis/node-redis/blob/v3.1.2/lib/command.js
 export interface RedisCommand {
   command: string;
   args: string[];
   buffer_args: boolean;
-  callback: redisTypes.Callback<unknown>;
+  callback: (err: Error | null, reply: unknown) => void;
   call_on_write: boolean;
 }
 


Flarna:
The change itself is ok but it doesn't solve the problem that the redis instrumentation exposes the redis types.
The use of  `typeof redisTypes`  in a [protected method](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/278ba99c4c82259813be4089b2cdb36fd4f98899/plugins/node/opentelemetry-instrumentation-redis/src/instrumentation.ts#L51) and on the [base class](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/278ba99c4c82259813be4089b2cdb36fd4f98899/plugins/node/opentelemetry-instrumentation-redis/src/instrumentation.ts#L37) still causes that `@types/redis` is required to be a dependency instead a devDependency.

Could you replace `typeof redisTypes` at these places by `any` or similar and update package.json?

pichlermarc:
> The change itself is ok but it doesn't solve the problem that the redis instrumentation exposes the redis types. The use of `typeof redisTypes` in a [protected method](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/278ba99c4c82259813be4089b2cdb36fd4f98899/plugins/node/opentelemetry-instrumentation-redis/src/instrumentation.ts#L51) and on the [base class](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/278ba99c4c82259813be4089b2cdb36fd4f98899/plugins/node/opentelemetry-instrumentation-redis/src/instrumentation.ts#L37) still causes that `@types/redis` is required to be a dependency instead a devDependency.
> 
> Could you replace `typeof redisTypes` at these places by `any` or similar and update package.json?

Thanks for the feedback, I updated the PR with these changes ([b86dc5e](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1424/commits/b86dc5e100d2f4d46f626fe2cd5b44ee52658b5e)) :slightly_smiling_face: 

I feel like I still lack some understanding of the subject, though:
 - not using `Callback` from `redis` fixed the issue in my local testing setup
   - As I understand this fixes the problem as `Callback` does not exist in redis 4, so not using it in the public API works with the types shipped with `redis@^4.0.0`.
   - As we're now not using anything from the `@types/redis` package, I was under the impression that it should be safe to use it in this way as type-arguments.

We're doing something similar in the express instrumentation (depending on `@types/express` using it in the instrumentation class and in one of its protected methods) - would that need to be subject to the same changes? I'd like to provoke some of the issues caused by using types in this manner in a local app so I can understand the problem a bit better :slightly_smiling_face:

Flarna:
Using the types in the user facing API is always a time bomb. We had this in several instrumentations in the past. You may fix one problem by removing `Callback` but other problems remain.

If it is in the user facing API it needs to be in the dependencies otherwise users using typescript see issues because of missing types.
If it is in the dependencies it's a specific version which is likely not that one used in actual application. Might again cause all sorts of issues.
If types are included in the package itself it's even worse as instrumentation has to depend on a specific version of the module it instruments.

The only safe variant is to keep the user facing API free of types of 3rd party modules. see also [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md).

Flarna (approved):
You might want to update description a bit as it's now a bit more then do not use `Callback`.

pichlermarc:
@Flarna got it, thanks! :slightly_smiling_face:

sfred-ox (approved):

rauno56 (approved):

PR #1426, opened 2023-03-13
feat(instrumentation-document-load): Baggage Support

paychex-ssmithrand:
I am opening this PR to present the added feature of supporting Baggage extraction in the Document Load Instrumentation.

Initial testing reveals baggage extraction works as expected and is a fairly un-obtrusive change. The only issue is creating proper unit tests to be able to capture active baggage at the right time. You'll see console log statements in both code and the test suite to log out the active baggage. The test data is accurately logged out in code, proving the extraction went as expected. However, due to the asynchronous nature of the tests (they are retroactively checking finished Spans for correct trace context values), I am struggling cook up a way to accurately unit test the baggage extraction.

Output from a test run:
```
LOG LOG: 'From code: ', [['key1', Object{value: ...}], ['key2', Object{value: ...}]]
LOG: 'From code: ', [['key1', Object{value: ...}], ['key2', Object{value: ...}]]
LOG LOG: 'From Test: ', undefined
LOG: 'From Test: ', undefined
```

Would love some ideas on how this can be achieved.

Note: CLA will be signed once my Senior Management becomes a CLA Manager with EasyCLA.

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
+++ b/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
@@ -49,6 +49,9 @@ export class DocumentLoadInstrumentation extends InstrumentationBase<unknown> {
   readonly version: string = '1';
   moduleName = this.component;
 
+  //TODO: Update W3CBaggagePropagator to export Baggage Header constant.
+  BAGGAGE_HEADER = 'baggage';
+
   /**
    *
    * @param config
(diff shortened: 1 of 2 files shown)


paychex-ssmithrand:
@dyladan @vmarchaud @obecny @rauno56 @legendecas  any suggestions/thoughts/feedback for this change?

dyladan:
The relevant people already get notifications when the PR opens. Please allow some time for review before pinging.

On a technical side I do have some concerns about this feature. First, it is not specified either by the W3C or by OTel which makes me wary to implement it. Second, baggage fundamentally "flows down." I'm not sure what the usecase is to send baggage from the server to the client that can't be better achieved with other pre-existing means. Since the document load is the head of a transaction, how are there already baggage properties? To me this seems like either a misunderstanding of what baggage is meant to be used for or a usecase that was not considered when baggage was designed.

Can you help me understand what you're trying to achieve by shipping baggage from the server to the client?

Issue #1431, opened 2023-03-16
instrumentation-pg: `Submittable` query parameter is not returned


unflxw:
### What version of OpenTelemetry are you using?

`@opentelemetry/instrumentation-pg@0.35.0`

### What version of Node are you using?

`v16.13.1`

### What did you do?

Performed a query against a Postgres client (`client.query()`) with the argument to the query function being a
`Submittable` object (such as those from the `pg-copy-streams` library)

### What did you expect to see?

The `Submittable` object passed to `client.query()` is returned by it.

### What did you see instead?

The return value is `undefined`.

### Additional context

Original issue reported as appsignal/appsignal-nodejs#871. Example code to run can be found in the issue and in https://github.com/brianc/node-pg-copy-streams/.

Looking at the `pg` code, it decides whether it's a `Submittable` by checking whether `typeof config.submit === 'function'` (where `config` is the first argument given to `client.query()`) and setting it as the return value (`result`). This logic can probably be replicated in the instrumentation's patch for the query function.

Instrumenting these sorts of queries is out-of-scope; just not breaking expected library behaviour would be enough to close this issue.

PR #1432, opened 2023-03-17
fix(pg): Do not replace argument with plain object

unflxw:
If the object given as a first argument contained a `.callback` property, the callback would be patch and the object in the argument array would be replaced with a plain object. If the object was a class, the prototype would be lost.

Instead, this implementation replaces the `.callback` property on the existing object, which is otherwise left as-is.

This fixes issue #1431, where `pg-copy-streams` provides a `Submittable` object with `.callback` and `.submit` methods defined in the class prototype. Because `.callback` exists, the object in the argument array is replaced with a plain object, containing the patched `.callback`, but no `.submit`.

Because `.submit` was no longer present, `pg` would not treat it as a submittable, causing it to attempt to execute the query by its `.text` property, and return `undefined` instead of the submittable.

### Notes

I dropped the logic to make a copy of the arguments because I'm not sure what the purpose of not modifying the original argument containing the callback is. If it's necessary to keep that, we might want to copy the prototype over to the reconstructed object in some other way.


rauno56:
Thanks for the ping. Sorry for my absence. I'll take a look tomorrow.

unflxw:
@rauno56 No hurry! Thanks for looking into it.

rauno56 (approved):
I'm reluctant of merging without passing tests, but the case you made is clear and makes sense.

unflxw:
It seems that the test failures are unrelated, right?

pichlermarc:
> It seems that the test failures are unrelated, right?

Yep they were unrelated, some lint issues remain that look related, though :slightly_smiling_face:

unflxw:
@pichlermarc Whoops, fixed! Apologies for the delay :)

unflxw:
I have rebased this against `main`. Can it be merged @pichlermarc @rauno56 @dyladan? 🥺🙏

unflxw:
Thank you @dyladan!

Issue #802, opened 2021-12-29
`@opentelemetry/instrumentation-graphql` Version `0.27.1` Missing `@types/graphql` dependency


asafpelegcodes:
### What version of OpenTelemetry are you using?
1.0.4

### What version of Node are you using?
v16.11.0

### Please provide the code you used to setup the OpenTelemetry SDK
```ts
import { GraphQLInstrumentation } from "@opentelemetry/instrumentation-graphql";
```

### What did you do?
In a fresh folder
```bash
yarn add @opentelemetry/api 
yarn add @opentelemetry/instrumentation-graphql
yarn add -D typescript
```
Add the build script to your `package.json`
```js
{
  "scripts": {
    "build": "tsc otel.ts"
  },
  "devDependencies": {
    "typescript": "^4.5.4"
  },
  "dependencies": {
    "@opentelemetry/api": "^1.0.4",
    "@opentelemetry/instrumentation-graphql": "^0.27.1"
  }
}
```
Attempt to compile the typescript
```bash
yarn build
```

### What did you expect to see?
I expected the file to compile successfully

### What did you see instead?
The file failed to compile because the graphql dependency was not installed
```
$ tsc otel.ts
node_modules/@opentelemetry/instrumentation-graphql/build/src/instrumentation.d.ts:2:36 - error TS2307: Cannot find module 'graphql' or its corresponding type declarations.

2 import type * as graphqlTypes from 'graphql';
```

### Additional context
In Version `0.26.1` it contains the dependency `"@types/graphql": "14.5.0"` which has somehow been dropped in the latest version

vmarchaud:
It has been dropped in this PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/754, we need to find a long term solution for this
cc @nata7che

Flarna:
Please note that `@types/graphql` is deprecated because `graphql` itself contains the types now. Therefore it would be needed to add `graphql` itself as dependency which most likely results in two different graphql versions installed.

Other instrumentations may have similar issues.

As far as I remember the easiest workaround is to set `skipLibCheck: true` in your tsconfig file.

dyladan:
As a long-term solution that doesn't require our users to skip lib check we could hand-roll our own types (or copy official ones)

asafpelegcodes:
Isn't copying the types more error prone than simply making `graphql` a dependency? 

[...]

PR #1436, opened 2023-03-23
Move @types/ packages to devDependencies

lpsinger:
TypeScript type packages are normally dev-only dependencies. You generally only need types for development, not at run time.

- Excessive size of `node_modules` directory when using these plugins.

- Move all `@types/` packages from `dependencies` to `devDependencies`.
- Update `GUIDELINES.md`.

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -99,7 +99,7 @@ To support this use case, you can choose one of the following options:
     // package.json
     {
         ...
-        "dependencies": {
+        "devDependencies": {
             "@types/foo": "1.2.3"
         },
         ...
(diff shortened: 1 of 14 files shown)


Flarna:
Have you ensured that these types are not used in any pubic API?
Otherwise consumers end up with typescript issues because of missing types.

The issue tracker holds a few such issues and there are a few PRs moving types from/to dev deps. e.a. https://github.com/open-telemetry/opentelemetry-js-contrib/issues/802

https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1320 shows how to remove types dependencies in API surface to move them into dev deps without causing problems.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md holds also some more details.

lpsinger:
Oh dear. This is more complicated than I at first realized.

PR #1439, opened 2023-03-27
fix(eslint-config): replace gts with prettier and eslint

### LINKED ISSUE #1433: Using gts instead of eslint
I've been having some issues with eslint's configuration as it is configured as part of the gts module. Is there a specific reason for using gts instead of eslint, prettier, etc just like in the core repo?

samimusallam:
Fixes #1433 

- Replaced the gts module with Prettier.
- Also using the following configuration for eslint: `eslint:recommended, plugin:@typescript-eslint/recommended, plugin:prettier/recommended` to match the core repo's configuration.


eslint.config.js
```diff
@@ -1,52 +1,61 @@
...
+    project: './tsconfig.json',
   },
   rules: {
-    "quotes": [2, "single", { "avoidEscape": true }],
-    "@typescript-eslint/no-this-alias": "off",
-    "eqeqeq": "off",
-    "prefer-rest-params": "off",
-    "@typescript-eslint/naming-convention": [
-        "error",
+    quotes: [2, 'single', { avoidEscape: true }],
```
Flarna, line 13: Some rules here differ from core repo. Is this intended to keep the changeset small?

samimusallam: True, but I wasn't sure if there were rules (eg: the import rule) that were added for a specific reason in this repo.
I would prefer to have no differences between the two repos.

samimusallam: I've updated the eslint.config file to match the core repo's configuration as close as possible.
The following is the difference between this and core to prevent eslint errors in existing code:

- `eqeqeq` set to `off` instead of `["error", "smart"]`
- removed `"no-console": "error"`
- removed `"@typescript-eslint/no-floating-promises": "error"`
- added `"@typescript-eslint/no-var-requires": "off"`
- removed `"@typescript-eslint/no-unused-vars": ["error", {"argsIgnorePattern": "^_", "args": "after-used"}]`

Let me know if it makes sense to re-add/remove these rules which will require some code changes in multiple packages in the repo.

Flarna: I think that's fine as a first step but we should aim to get in sync soon via some followup PRs. Best one by one to keep changesets small and easy/fast reviewed.
In special disabling `eqeqeq` and `no-console` allows code patterns we usually don't want.

Flarna (approved):

Issue #1440, opened 2023-03-28
opentelemetry-instrumentation-aws-lambda silently fails to instrument files with the extension.cjs


Ankcorn:
### What version of OpenTelemetry are you using?
latest

### What version of Node are you using?
v16.x

### What did you do?

https://github.com/baselime/lambda-node-opentelemetry-serverless-stack-v2
Here is a repo reproducing the issue 

### What did you expect to see?
Traces sent

### What did you see instead?
lambda handler is not instrumented

### Additional context
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/96055281717d54305ff5c7b02b7a9d16c38dcea7/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L103 issue is this line of code does not handle all possible options supported by lambda node runtime

dyladan:
@willarmiros can you please look into this and the linked PR?

Ankcorn:
Thanks for getting this published!

I just saw this https://github.com/open-telemetry/opentelemetry-js/pull/3698 pr was closed!

I realize that makes my fix a little half baked. sorry .mjs files...

Should I make a new issue to track that @dyladan

dyladan:
Yes

PR #1442, opened 2023-03-28
fix: .cjs extension support for lambda functions

Ankcorn:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1440

- checks files in lambda path and selects the match with the same priority as the lambda runtime


Flarna:
There are a lot formatting changes which cause lint to complain. Please revert them. Likely you editor has some different config then that one used in this repo.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -99,8 +100,25 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
     // Lambda loads user function using an absolute path.
     let filename = path.resolve(taskRoot, moduleRoot, module);
     if (!filename.endsWith('.js')) {
```
willarmiros, line 102: I wonder, could we do a quick optimization and also check if the filename ends with`.cjs` here? In that case we can also just carry on without needing to do all the expensive operations below.

willarmiros (commented):
A few comments, in principle looks good though

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -99,8 +100,25 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
...
+
+      // if we find a .js file that will be loaded first, otherwise .cjs will be loaded by the lambda runtime
+      // fallsback to {module}.js incase of issue.
+      const name =
+        matchedFiles.find(file => path.parse(file).ext === '.js') ||
+        matchedFiles.find(file => path.parse(file).ext === '.cjs') ||
+        `${module}.js`;
+
       // Patching infrastructure currently requires a filename when requiring with an absolute path.
       filename += '.js';
```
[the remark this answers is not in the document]
Ankcorn: Thanks!

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
Ankcorn: I ran a test and we might be able to do something with fs.stat here

```
import fs from 'node:fs/promises';

console.time('stat');
await fs.stat('./test.js').catch(e=>{});
console.timeEnd('stat')


console.time('listdir');
await fs.readdir('./ping');
console.timeEnd('listdir')
```

```
stat: 0.111ms
listdir: 0.416ms
```
Obviously that performance hit is pretty significant but I could check for a .js file then fallback to .cjs if its not found.

I suspect this will also need some work once the otel.js library supports esm

willarmiros: Yeah I like this approach the best, it's simple, not perfect, but an improvement over what we have today. I'd be good with this change.

Ankcorn: Just realized the numbers are .11ms not 111ms :sweat_smile:

Ankcorn:
@willarmiros Those changes you request have been made, let me know if you want me to make any other updates.

Thanks for all your help on this

willarmiros (approved):
lgtm!

dyladan (approved):
LGTM and approved by code owner. merging

Ankcorn:
@dyladan 
can't figure out where that checkbox is but have updated from main :) let me know if I need to do it again.
Thank you!

dyladan:
> @dyladan
> can't figure out where that checkbox is but have updated from main :) let me know if I need to do it again.
> Thank you!

Right side bar 
image

Ankcorn:
Hmm I don't see. I will merge again and lets hope it all goes okay :)

pichlermarc:
> Hmm I don't see. I will merge again and lets hope it all goes okay :)

Looks like this does not work because the fork was created by an organization. Anyway, it's up to date now, so I'll merge this. Thanks! :slightly_smiling_face:

Ankcorn:
Thanks @pichlermarc

jindong-zhang-git:
Hi friends, does this package support `.mjs` files after this fix?

Ankcorn:
I never got that working. We started manually instrumenting handlers because it was much more reliable. 

Feel free to reach out for a chat about node.js otel on lambda 😊

PR #1444, opened 2023-03-29
fix(gql)!: conform GraphQL span name to spec

naseemkullah:

Rename graphql "execute" spans to "<operation.type> <operation.name>" to conform to the [spec](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/instrumentation/graphql/).

In addition to conforming to the spec, this change can provide additional value for monitoring purposes. In cases where o11y providers generate metrics from pre-ingested traces (without all the attributes), the new naming convention makes it easier to group and filter metrics by operation type and name, since it is now reflected in the span name. This can be particularly useful for setting up monitors or alerts.

When operation name and type are determined within the execute span, `span.updateName()` is called.

The query steps other than "execute"'s spans are out of scope as they
do not need to be renamed.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
@@ -411,14 +411,15 @@ export class GraphQLInstrumentation extends InstrumentationBase {
 
       const operationName = nameNode?.value;
 
-      if (operationName) {
-        span.setAttribute(AttributeNames.OPERATION_NAME, operationName);
-      }
-
       // https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/instrumentation/graphql/
       // > The span name MUST be of the format <graphql.operation.type> <graphql.operation.name> provided that graphql.operation.type and graphql.operation.name are available.
       // > If graphql.operation.name is not available, the span SHOULD be named <graphql.operation.type>.
-      span.updateName(`${operationType} ${operationName ?? ''}`.trim());
+      if (operationName) {
+        span.setAttribute(AttributeNames.OPERATION_NAME, operationName);
+        span.updateName(`${operationType} ${operationName}`);
+      } else {
+        span.updateName(operationType);
+      }
     } else {
       let operationName = ' ';
       if (processedArgs.operationName) {
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
@@ -404,19 +404,21 @@ export class GraphQLInstrumentation extends InstrumentationBase {
 
     const span = this.tracer.startSpan(SpanNames.EXECUTE, {});
     if (operation) {
-      const operationDefinition =
+      const { operation: operationType, name: nameNode } =
         operation as graphqlTypes.OperationDefinitionNode;
-      span.setAttribute(
-        AttributeNames.OPERATION_TYPE,
-        operationDefinition.operation
-      );
 
-      if (operationDefinition.name) {
-        span.setAttribute(
-          AttributeNames.OPERATION_NAME,
-          operationDefinition.name.value
-        );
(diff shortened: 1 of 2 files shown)


plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -402,21 +402,22 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+      span.setAttribute(AttributeNames.OPERATION_TYPE, operationType);
+
+      if (operationName) {
+        span.setAttribute(AttributeNames.OPERATION_NAME, operationName);
       }
+
+      // https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/instrumentation/graphql/
+      // > The span name MUST be of the format <graphql.operation.type> <graphql.operation.name> provided that graphql.operation.type and graphql.operation.name are available.
+      // > If graphql.operation.name is not available, the span SHOULD be named <graphql.operation.type>.
+      span.updateName(`${operationType} ${operationName ?? ''}`.trim());
```
naseemkullah, line 420: this is the only non-refactor change

naseemkullah:
I think something is up with the unit test checks - tests pass locally  and i see this issue in other PRs 
image

naseemkullah:
i am actually going to revert renaming the enum, i think aside from the execute step, the other steps will always be named as such. im looking at what vendor specific agent calls those and its the same as ours, only execute is renamed

naseemkullah:
pardon the force pushing, i reverted alot of stylistic changes to keep the review as easy as possible

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -404,19 +404,21 @@ export class GraphQLInstrumentation extends InstrumentationBase {
...
+      const operationName = nameNode?.value;
+
+      if (operationName) {
+        span.setAttribute(AttributeNames.OPERATION_NAME, operationName);
       }
+
+      // https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/instrumentation/graphql/
+      // > The span name MUST be of the format <graphql.operation.type> <graphql.operation.name> provided that graphql.operation.type and graphql.operation.name are available.
+      // > If graphql.operation.name is not available, the span SHOULD be named <graphql.operation.type>.
+      span.updateName(`${operationType} ${operationName ?? ''}`.trim());
```
dyladan, line 421:
```suggestion
      // https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/instrumentation/graphql/
      // > The span name MUST be of the format <graphql.operation.type> <graphql.operation.name> provided that graphql.operation.type and graphql.operation.name are available.
      // > If graphql.operation.name is not available, the span SHOULD be named <graphql.operation.type>.
      if (operationName) {
        span.setAttribute(AttributeNames.OPERATION_NAME, operationName);
        span.updateName(`${operationType} ${operationName}`);
      } else {
        span.updateName(operationType);
      }
```

dyladan (approved):
Overall LGTM

mwear (approved):
LGTM, but I have one question about the changelog. I see there's a changelog for each package, and for contrib as a whole. Are they generated automatically? Is there something we can do to make it completely clear to users that the span name will change after upgrading?

naseemkullah:
> > +1 Changelogs are created automatically on [release PRs](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/947).
> > Other than emphasizing it in the changelog and bumping major(which we cannot do, because `0.x`), I don't know how to make it more clear, especially since we're fixing spec noncompliance.
> 
> I also think there's nothing we can do to make it clearer. I think it should be categorized correctly as a breaking change in the changelog when titling the PR as `fix(gql)!: conform GraphQL span name to spec`, so maybe changing the title could help.

Thanks, I was unsure whether this was considered breaking or not since we are adhering to spec. `!` added.

pichlermarc (approved):
Looks good :slightly_smiling_face:

PR #1447, opened 2023-03-29
fix(@types/koa): update @types/koa version to latest

haddasbronfman:
## Which problem is this PR solving?

fixes #1446 

## Short description of the changes

change the @types/koa version from 2.13.4 to 2.13.6

PR #1449, opened 2023-03-29
feat(host-metrics): Add process metrics

hectorhdzg:

- Add Process Metrics defined in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process

### DIFF
--- a/packages/opentelemetry-host-metrics/src/enum.ts
+++ b/packages/opentelemetry-host-metrics/src/enum.ts
@@ -22,6 +22,9 @@ export enum METRIC_NAMES {
   NETWORK_DROPPED = 'system.network.dropped',
   NETWORK_ERRORS = 'system.network.errors',
   NETWORK_IO = 'system.network.io',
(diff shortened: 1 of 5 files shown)


hectorhdzg:
Build is currently failing because of following error:
lerna ERR! lifecycle "prepare" errored in "@opentelemetry/instrumentation-koa", exiting 1

Maybe related to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1447

haddasbronfman:
I believe yes. it is related to #1447. this is the same error (and happened around the same time)

legendecas (approved):
LGTM. Thank you for working on this.

When reviewing this work, I realized that we can already adopt batch callbacks in the host metrics: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1450.

Anyway, this can be landed first and I can update that one respectively.

MSNev (approved):

JacksonWeber (approved):

PR #1451, opened 2023-04-02
feat(mongodb): support v5

GuillaumeDeconinck:

- `@opentelemetry/instrumentation-mongodb`: Handle version 5 of MongoDB NodeJS Driver

- feat(mongodb): Extend range of versions allowed for v4 patches to include v5
- test(mongodb) Add test cases for v5
  - Duplicate of v4 with just callbacks removed (as it has been removed in v5)
- chore(mongodb): Add command and configuration for v5 tests (`package.json` and `.tav.yml`)
- chore(mongodb): Renamed package.json command `test-new-versions` to `test-v4` (and adapted `.tav.yml` for it)

## Notes
- Unit tests of v4 are apparently running fine on MongoDB v5 when callbacks are adapted to Promises
- Tested quickly on an app with jaeger, otel collector and signoz, it seems to work fine too
- I'm not sure if there's any breaking change impacting the tracing in the V5. From what I understand, it's not the case

Feel free to comment/adapt/suggest things. I'm still learning on this kind of stuff 😉


GuillaumeDeconinck:
Actually we tested a bit more today and noticed that the details of the queries are not present (e.g. the `find` parameters). I'll try to look at it soon.

Eji4h:
I'm waiting for it.

GuillaumeDeconinck:
For what I said before:

> Actually we tested a bit more today and noticed that the details of the queries are not present (e.g. the find parameters). I'll try to look at it soon.

Nevermind, I thought it automatically included all the parameters of `find()`, but the `enhancedDatabaseReporting` option is actually used for that. By default, only the first level is included, like before.

I think the PR can be reviewed.

karanssj4:
@osherv any update on this? v5 is working similar to v4 except for dropping callback support

icco:
I'm not an approver, but this looks good to me (there are currently merge conflicts though). We'd love to see this merged.

GuillaumeDeconinck:
I quickly checked but it seems there are some tests failing now after fixing the conflicts.
I'll likely be able to fix these in the coming days.
Once done, I'll ping another maintainer as the (randomly?) assigned one doesn't seem to be active for the moment.

GuillaumeDeconinck:
I've adapted the code based on the changes on main. The unit tests are all passing locally.

I had to use the same database name for all the tests, as right now the instrumentation [only allows one poolName](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts#L53) (and not one per connection). That's likely something to fix in another issue/PR.
  
@haddasbronfman Hello, sorry to ping you like that. I see you checked/worked on another PR for MongoDB. Would it possible to review this one when you have some time ? Thanks !

icco:
@legendecas or @dyladan could we get a review here?

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -7,8 +7,9 @@
   "repository": "open-telemetry/opentelemetry-js-contrib",
   "scripts": {
     "docker:start": "docker run -e MONGODB_DB=opentelemetry-tests -e MONGODB_PORT=27017 -e MONGODB_HOST=127.0.0.1 -p 27017:27017 --rm mongo",
-    "test": "nyc ts-mocha -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v3.test.ts'",
-    "test-new-versions": "nyc ts-mocha -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/mongodb-v4**.test.ts'",
+    "test-v3": "nyc ts-mocha -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v3.test.ts'",
```
pichlermarc, line 10: I think this script should be kept as `test` as that is the one that will be executed by default. :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4-v5.metrics.test.ts
```diff
@@ -83,31 +94,31 @@ describe('MongoDBInstrumentation-Metrics', () => {
 
   const HOST = process.env.MONGODB_HOST || DEFAULT_MONGO_HOST;
   const PORT = process.env.MONGODB_PORT || '27017';
-  const DB_NAME = process.env.MONGODB_DB || 'opentelemetry-tests-metrics';
+  const DB_NAME = process.env.MONGODB_DB || 'opentelemetry-tests';
```
pichlermarc, line 97: why change this? :thinking:

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4-v5.metrics.test.ts
```diff
@@ -28,7 +28,27 @@ import {
...
+let instrumentation: MongoDBInstrumentation;
+{
+  const instance: MongoDBInstrumentation | undefined = getInstrumentation();
+  if (!instance) {
+    instrumentation = new MongoDBInstrumentation();
+    registerInstrumentationTesting(instrumentation);
+  } else {
+    instrumentation = instance;
+  }
+}
```
pichlermarc, line 48: Hmm, what's the reason to do it like that vs. what how it was before? Was `registerInstrumentationTesting()` not working as intended? :thinking:

pichlermarc (commented):
Just a few questions from my side. :slightly_smiling_face:  

@haddasbronfman I know you've been contributing to this instrumentation lately, would you mind taking a look? :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-mongodb/package.json
GuillaumeDeconinck: Ah yeah makes sense. By default should it run only the test for `v3`, or should it run all versions ? 👀

pichlermarc: By default, we had it run only for v3, so I think keeping just `v3` is fine. The test-all-versions script will take care of running the other tests (and that will run if the `pkg:instrumentation-mongodb` label is applied to the PR) :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4-v5.metrics.test.ts
GuillaumeDeconinck: Following your comment, I have adapted to make it less complex, but in short I had some issues with the singleton and the fact that re-calling [registerInstrumentationTesting](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/opentelemetry-test-utils/src/instrumentations/instrumentation-singelton.ts#L37) disables the new instrumentation and returns the old one.

It works fine in the tests now because

- `test/mongodb-v4-v5.metrics.test.ts` creates a new instrumentation and registers it - it never disables it
- `test/mongodb-v4.test.ts` creates a new instrumentation and tries to register it
  - But actually the new instrumentation is enabled (constructor) and then directly disabled by registerInstrumentationTesting. The old one, still active, is returned
- `test/mongodb-v4.test.ts` disables the instrumentation at the end of the tests

I only noticed that recently, that the order of test files is important :/
The other way around doesn't work because the traces' tests disable the instrumentation at the end, making the metrics tests run/fail without any instrumentation.

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4-v5.metrics.test.ts
GuillaumeDeconinck: I had some issues with the pool names, but you're right, I shouldn't have changed it. Reverted

GuillaumeDeconinck:
@pichlermarc Thanks for the review and for maintaining the branch up-to-date with master 🙂 
I've updated the PR based on your comments

haddasbronfman (commented):
LGTM. thank you very much for handling this !!
 only one comment from my side.

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -51,16 +53,16 @@
...
     "@opentelemetry/sdk-trace-node": "^1.8.0",
+    "@types/bson": "4.0.5",
     "@types/mocha": "7.0.2",
+    "@types/mongodb": "3.6.20",
     "@types/node": "18.11.7",
     "mocha": "7.2.0",
-    "mongodb": "3.6.11",
-    "@types/mongodb": "3.6.20",
-    "@types/bson": "4.0.5",
+    "mongodb": "4",
```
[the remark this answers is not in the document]
GuillaumeDeconinck: Good point, I forgot to revert it after my tests 😅 Changing it back now

pichlermarc (approved):
Looks great, thanks :slightly_smiling_face: 
Waiting for one more CI run before merging :slightly_smiling_face:

pichlermarc:
Looks like there's a problem with our workflows - we'll have to take care of #1541 first to make CI pass. Sorry for the delay. :disappointed:

PR #1452, opened 2023-04-02
Pg enhancedDatabaseReporting

### LINKED ISSUE #1235: pg instrumentation: enhancedDatabaseReporting value is not parsable
 "@opentelemetry/instrumentation-pg": "0.31.1"

16

### What did you do?
Executed the following pg query with instrumentation config `enhancedDatabaseReporting: true`
```js
    const res = await client.query('SELECT $1::text as message1, $2::text as message2', ['Hello,World', 'foo,bar']);
```

I expected to be able to tell what the query values are for this operation.

### What did you see instead?
I got this value in the `db.postgresql.values` attribute: `"[Hello,World,foo,bar]"`
This text is not reversible. all of these arrays will yield the same attribute value and there is no way to differentiate them in a deterministic way:
```js
['Hello,World', 'foo,bar']
['Hello', 'World,foo,bar']
['Hello,World,foo', 'bar']
```

### Additional context
The reason is that [this data is serialized](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/682d6107b7285829bb72e592936c585ccdfab16a/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts#L48) in a non-standard, non-reversible way:
```js
function arrayStringifyHelper(arr: Array<unknown>): string {
}
```

I suggest using `JSON.stringify`

haddasbronfman:

Fixes #1235 

use `JSON.stringify()` instead of `arrayStringifyHelper`


haddasbronfman:
By mistake, the last commit [13b57df](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1452/commits/13b57dfd057f105bd8cb69f2fd31bd64995b83c1) was pushed to here. 
There should be pr #1170  for mongo
and #1453 for pg

this pr is a combination of both by mistake.

PR #1453, opened 2023-04-02
fix(pg-values): values should be parsable when enhancedDatabaseRepoting:true

### LINKED ISSUE #1235: pg instrumentation: enhancedDatabaseReporting value is not parsable

haddasbronfman:

Fixes #1235 

use `JSON.stringify()` instead of `arrayStringifyHelper`


plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -158,7 +155,7 @@ export function handleConfigQuery(
   ) {
     span.setAttribute(
       AttributeNames.PG_VALUES,
-      arrayStringifyHelper(queryConfig.values)
+      JSON.stringify(queryConfig.values)
```
blumamir, line 158: I recommend wrapping this function with try-catch, as `JSON.stringify` can throw for cyclic objects, and `queryConfig.values` is a user supplied argument for which we cannot assume anything.

blumamir (commented):
Too bad https://github.com/open-telemetry/opentelemetry-specification/pull/3092
Didn't get through. 

Thanks for fixing this. Left just one small note about `JSON.stringify` can throw exception

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -156,10 +152,14 @@ export function handleConfigQuery(
...
     Array.isArray(queryConfig.values)
   ) {
-    span.setAttribute(
-      AttributeNames.PG_VALUES,
-      arrayStringifyHelper(queryConfig.values)
-    );
+    try {
+      span.setAttribute(
+        AttributeNames.PG_VALUES,
+        JSON.stringify(queryConfig.values)
```
Flarna, line 158: values may hold objects or buffers and [PG has some handling](https://node-postgres.com/features/queries#parameterized-query) how this is converted.
Not sure if `JSON.stringify` fits well in that cases.

e.g. `JSON.stringify([Buffer.from("abc")])` results in `'[{"type":"Buffer","data":[97,98,99]}]'`

It differs also from the proposal done in spec repo

haddasbronfman: thanks for this comment.
Do you have any idea what can be used instead of 'stringify'?

Flarna: I think your initial idea to use an array of strings should be fine. Likely some conversion code is needed to handle `null`/`undefined`, buffers, and objects similar as pg does.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -153,10 +153,19 @@ export function handleConfigQuery(
     Array.isArray(queryConfig.values)
   ) {
     try {
-      span.setAttribute(
-        AttributeNames.PG_VALUES,
-        JSON.stringify(queryConfig.values)
-      );
+      const convertedValues = queryConfig.values.map(value => {
+        if (!value) {
```
Flarna, line 157:
```suggestion
        if (value == null) {
```
only null/undefined should fit here, not 0 or empty strings

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -153,10 +153,19 @@ export function handleConfigQuery(
...
+          return 'null';
+        } else if (value instanceof Buffer) {
+          return value.toString();
```
Flarna, line 160: maybe use "hex" formatting to get a consistent format.
e.g.
`Buffer.from([65, 0, 1, 2]).toString()` => `'A\x00\x01\x02'`
`Buffer.from([65, 0, 1, 2]).toString("hex")` => `'41000102'`

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -153,10 +153,19 @@ export function handleConfigQuery(
...
   ) {
     try {
-      span.setAttribute(
-        AttributeNames.PG_VALUES,
-        JSON.stringify(queryConfig.values)
-      );
+      const convertedValues = queryConfig.values.map(value => {
+        if (!value) {
+          return 'null';
+        } else if (value instanceof Buffer) {
```
Flarna, line 159:
```suggestion
        } else if (Buffer.isBuffer(value)) {
```

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -153,10 +153,19 @@ export function handleConfigQuery(
...
+          return value.toString();
+        } else if (typeof value === 'object') {
+          return JSON.stringify(value);
```
Flarna, line 162: Should we try to use a potential existing `toPostgres` function like PG does?

In general a try/catch around this conversions is likely a good idea as `JSON.stringify` could throw.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
haddasbronfman: good catch. thanks!

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
haddasbronfman: Why prefer `isBuffer()` over `instanceof`? it seems in the code that it's the same: 
https://github.com/nodejs/node/blob/main/lib/buffer.js#L556

Flarna: ah ok. Feel free to keep it as is. I was assuming that the existence of `Buffer.isBuffer()` indicates that there is some more magic included there.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
haddasbronfman: Ok. added 'toPostgres' condition.

We already have a try/catch for all this conversion, but I added a printout of the error itself now.

Flarna (approved):
Maybe tune tests a bit to add e.g. a `toPostgres()`, empty string, 0, NaN cases if you find some time.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
haddasbronfman: if we use "hex", then the value of `Buffer.from('abc').toString('hex')` is `616263` rather than `abc`. is that what we want to present to the user?

Flarna: not sure what is best here. If buffers actually contain strings using hex encoding is not nice. But if buffers hold binary data it's quite nice. I have no strong opinion here.

By the way, should be also consider some len limit here or just rely on SDK built in limits?

haddasbronfman: 1. I left toString() as is. without 'hex'.
2. I added a test for 'toPostgres'.

haddasbronfman:
@Flarna I added a test. do you want to take a look or is it ok to merge?

Flarna (approved):

Flarna:
fine for me.
as of now @rauno56 is the owner of this instrumentation but not sure if component_owners.yml is up to date.

Issue #1353, opened 2023-01-12
Add custom attributes to SQLCommenter comments in pg instrumentation


joelmukuthu:
### Is your feature request related to a problem? Please describe

Support for attaching [SQLComment comments](https://google.github.io/sqlcommenter/spec/#meta-characters) to SQL statements handled by the `pg` module was recently added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1286. This issue proposes adding support for custom attributes to the SQL comments, besides `tracestate` and `traceparent`, which are now automatically generated by pg instrumentation (if the `addSqlCommenterCommentToQueries` config is enabled).

Examples of the custom attributes (as suggested by the spec) are `action`, `controller`, `framework` and `route`.  

### Describe the solution you'd like to see

As suggested in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1286#issuecomment-1375196941, one approach would be to allow the instrumentation to extract the custom attributes from the active context when a query is run (perhaps controlled by another config option). With this approach, one question that comes to mind is whether or not the attributes should be changeable from span to span.

### Describe alternatives you've considered

Not using pg instrumentation to handle SQLCommenter comments, and instead appending the comments myself, where I can add whichever attributes I'd like. 

### Additional context

I'm creating this issue to get some thoughts on:

1. whether or not this makes sense for pg instrumentation
2. if 1, what approach would be best for implementing this
3. if 1 and 2, pointers on where to get started so I or someone else can get started on a PR

joelmukuthu:
Don't close.

alphashuro:
@joelmukuthu did you ever find a solution to this?

PR #1454, opened 2023-04-03
feat(instrumentation-pg): support custom SQL Commenter attributes

joelmukuthu:

- Closes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1353

- This PR is a follow-up to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1286 that adds support for custom attributes in the SQL commenter comments. Since the pg instrumentation has no way (that I know of) of getting the attributes itself, this PR uses the active context to get the attributes that'd have to be set by the user. If this approach is approved, I'll add docs for it.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
@@ -215,18 +215,28 @@ export class PgInstrumentation extends InstrumentationBase {
         // Modify query text w/ a tracing comment before invoking original for
         // tracing, but only if args[0] has one of our expected shapes.
         //
-        // TODO: remove the `as ...` casts below when the TS version is upgraded.
+        // TODO: remove the `as ...` casts on `arg0` below when the TS version is upgraded.
         // Newer TS versions will use the result of firstArgIsQueryObjectWithText
         // to properly narrow arg0, but TS 4.3.5 does not.
         if (instrumentationConfig.addSqlCommenterCommentToQueries) {
+          const customSqlCommenterAttributes = context
+            .active()
+            .getValue(
+              Symbol('customSqlCommenterAttributes')
+            ) as utils.SqlCommenterAttributes;
           args[0] = firstArgIsString
-            ? utils.addSqlCommenterComment(span, arg0 as string)
+            ? utils.addSqlCommenterComment(
+                span,
+                arg0 as string,
+                customSqlCommenterAttributes
+              )
             : firstArgIsQueryObjectWithText
             ? {
                 ...(arg0 as utils.ObjectWithText),
                 text: utils.addSqlCommenterComment(
                   span,
-                  (arg0 as utils.ObjectWithText).text
+                  (arg0 as utils.ObjectWithText).text,
+                  customSqlCommenterAttributes
                 ),
               }
             : args[0];
--- a/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
@@ -47,7 +47,7 @@ function arrayStringifyHelper(arr: Array<unknown>): string {
   return '[' + arr.toString() + ']';
 }
 
-type SqlCommenterAttributes = Record<string, string>;
+export type SqlCommenterAttributes = Record<string, string>;
 
 /**
  * Helper function to get a low cardinality span name from whatever info we have
--- a/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
@@ -47,6 +47,8 @@ function arrayStringifyHelper(arr: Array<unknown>): string {
   return '[' + arr.toString() + ']';
 }
 
+type SqlCommenterAttributes = Record<string, string>;
+
 /**
  * Helper function to get a low cardinality span name from whatever info we have
  * about the query.
@@ -284,7 +286,11 @@ function fixedEncodeURIComponent(str: string) {
   );
 }
 
-export function addSqlCommenterComment(span: Span, query: string): string {
+export function addSqlCommenterComment(
+  span: Span,
+  query: string,
+  customAttributes: SqlCommenterAttributes = {}
+): string {
   if (typeof query !== 'string' || query.length === 0) {
     return query;
   }
@@ -296,7 +302,7 @@ export function addSqlCommenterComment(span: Span, query: string): string {
   }
 
   const propagator = new W3CTraceContextPropagator();
-  const headers: { [key: string]: string } = {};
+  const headers: SqlCommenterAttributes = customAttributes;
   propagator.inject(
     trace.setSpan(ROOT_CONTEXT, span),
     headers,
@@ -311,9 +317,10 @@ export function addSqlCommenterComment(span: Span, query: string): string {
   }
 
   const commentString = sortedKeys
-    .map(key => {
+    .map((key) => {
+      const encodedKey = fixedEncodeURIComponent(key);
       const encodedValue = fixedEncodeURIComponent(headers[key]);
-      return `${key}='${encodedValue}'`;
+      return `${encodedKey}='${encodedValue}'`;
     })
     .join(',');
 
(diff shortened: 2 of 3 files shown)


plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -47,6 +47,8 @@ function arrayStringifyHelper(arr: Array<unknown>): string {
   return '[' + arr.toString() + ']';
 }
 
+export type SqlCommenterAttributes = Record<string, string>;
```
joelmukuthu, line 50: Would it be preferred that I list all the known attributes instead of the index signature? Or both?

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -215,18 +215,28 @@ export class PgInstrumentation extends InstrumentationBase {
...
         //
-        // TODO: remove the `as ...` casts below when the TS version is upgraded.
+        // TODO: remove the `as ...` casts on `arg0` below when the TS version is upgraded.
         // Newer TS versions will use the result of firstArgIsQueryObjectWithText
         // to properly narrow arg0, but TS 4.3.5 does not.
         if (instrumentationConfig.addSqlCommenterCommentToQueries) {
+          const customSqlCommenterAttributes = context
+            .active()
+            .getValue(
+              Symbol('customSqlCommenterAttributes')
```
joelmukuthu, line 225: I'm not sure about this bit, any suggestions are welcomed

pichlermarc:
Hi @joelmukuthu, thank you for opening this PR :slightly_smiling_face: 
In order for us to proceed with the PR, the CLA needs to be signed first (see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1454#issuecomment-1493942431 above). :slightly_smiling_face:

joelmukuthu:
@pichlermarc my bad. thanks for the info, now signed!

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
dyladan: This won't work. This names the symbol but it is still unique each time it is called. You would need to use `Symbol.for` in order to get the same symbol each time, but I think this approach loses some of the advantage of using a symbol.

Using context seems like a reasonable way to go about this but I haven't seen any other instrumentations do this in any language I think. I think the idea of using context to inject things into instrumentations has merit, but maybe the mechanism can be generalized to other instrumentations as well.

@open-telemetry/javascript-maintainers what would you think?

joelmukuthu: > You would need to use Symbol.for in order to get the same symbol each time, but I think this approach loses some of the advantage of using a symbol.

You're quite right, I was forced to use a `Symbol` here because that's what the API expects. I think if we came up with a standard way of using the context to pass data to instrumentations, then there'd be a different API here. 

I'm not sure I can offer much insight into what the API could look like but I'm keen to hear what the others think and very happy to implement what is agreed upon :raised_hands:

joelmukuthu:
No worries @blumamir, I understand that this is a slow cooker. I'm happy to rebase it, but the main blocker here is that we don't yet have an agreed-upon way of passing custom attributes into the instrumentation. I suggested doing it via the context and as there hasn't been a lot of opposition to that, I'll give it a go as soon as I can (early September)

blumamir:
> No worries @blumamir, I understand that this is a slow cooker. I'm happy to rebase it, but the main blocker here is that we don't yet have an agreed-upon way of passing custom attributes into the instrumentation. I suggested doing it via the context and as there hasn't been a lot of opposition to that, I'll give it a go as soon as I can (early September) at which point this will be ready for review.

From reading your issue, I understand that you plan to use this feature to inject these attributes?

> action, controller, framework and route.

If I understand the motivation correctly, the reason you choose to use context is that, for example, `nestjs` instrumentation can inject these values into the context to be picked up later on by downstream pg instrumentation?

Or do you imagine some middleware that the user will need to hook into the framework? or will it be completely manual from user code with no automatic support from OTEL packages?

I think in this case it does make sense to inject these values via context. Since the values are not static, and change from operation to operation (for example, the route), I don't see any other way which is more elegant. To my understanding, `context` api was designed for these cases exactly to propagate useful information downstream.

There are a few existing examples of the usage of context in the project:
1. [rpc-metadata](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-core/src/trace/rpc-metadata.ts)
2. [suppress-tracing](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-core/src/trace/suppress-tracing.ts)
3. [baggage](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/baggage/context-helpers.ts) which is an opentelemetry speced API.

What they all have in common is that they define the type and getters/setters in a neutral package which can later be consumed by: users or other instrumentation that want to set these values upstream, users and other instrumentations that want to get/read/use these values downstream.

The one thing that I am not comfortable with is that the way you currently access the context key is via `"customSqlCommenterAttributes"` which makes it specific to sqlCommenter. This would mean that we need to inject this value in many places and if other instrumentation will also need access to `"framework"` or `"controller"`, they would have to read it from the SqlCommenter key which is not great, or will have to define there own type and key which means duplicating the data many times for each consumer. 
I would even consider if you can extend the [type RPCMetadata](https://github.com/open-telemetry/opentelemetry-js/blob/8ac369f87e792702045f9847c7ab4ac72aa48ce2/packages/opentelemetry-core/src/trace/rpc-metadata.ts#L27), which already claims to track the `route`, and add more info there? so that each instrumentation can additionally describe it's framework, controller etc and pg instrumentation will just read the values from there and inject them as SQL comments?

blumamir:
And one more question, what are these values use for?
I can understand how the `traceparent` and `tracestate` are used to emit telemetry such as new spans, but what happens with the other attributes such as `route`? is it only for logging the queries and gaining more context when examining a single statement in pg logs? or is it picked up by pg server somehow and indexed to be useful?

Since we already have the `traceparent`, the user can pick up any span and follow the complete context upstream to see the routes, frameworks, and the full story on what executed the query, and not just hand-picked attributes, no?

joelmukuthu:
@blumamir thank you so much for sharing the info you did! It's great to know that this kind of thing is what the context is designed for. I'll look into the examples you shared and suggest an approach (on this PR) to solving the problem the PR is trying to solve.

Wrt what problem this PR is actually trying to solve, yes, AFAIK these values are used to provide insights into query performance or just logged by other tools downstream. For instance, GCP's Cloud SQL captures the values and aggregates query performance:

image

[Ref](https://cloud.google.com/blog/topics/developers-practitioners/enable-query-tagging-sqlcommenter-understand-application-impact-database-performance).

As for how I imagined this feature being used, it's exactly as you anticipated: the values can be populated by other instrumentation or manually by the user e.g. in some kind of middleware (e.g. express middleware). 

> The one thing that I am not comfortable with is that the way you currently access the context key is via "customSqlCommenterAttributes" which makes it specific to sqlCommenter.

Great feedback! My choice with the wording here was to try and minimise the impact or not ruffle too many feathers. But yes, there's nothing about these values (at least at the time of this writing) that is specific to SQL.  I agree that a more generalised name is better. 

It looks like extending `RPCMetadata` could solve the need here quite well. On that note, what do you think about using `RPCMetadata` itself? Else, if we go with extending it instead, do you have any suggestions on what name we could give it? I'm struggling to come up with a name, perhaps because I'm too close to the "SQL Commenter" problem.

JamieDanielson:
Thank you for working on this and sorry for the delay! We would need to do more research on this to be more confident in how this will work generally and be maintained. We're concerned that there may not be a common enough use case in the community for this feature, and we're hesitant to take it on right now. To effectively set expectations, we'd like to hold off on implementing this PR for now and if there is enough interest we can revisit in the future for a more generic option.

joelmukuthu:
@JamieDanielson thanks for the response. Just so I understand better, is it the "SQL Commenter" use-case that's unpopular or is there not enough justification to access attributes stored in the context from within instrumentations?

JamieDanielson:
> @JamieDanielson thanks for the response. Just so I understand better, is it the "SQL Commenter" use-case that's unpopular or is there not enough justification to access attributes stored in the context from within instrumentations?

At the moment we maintainers have a limited bandwidth and as we are not well-versed in this feature, there is a lot more research we have to do to be comfortable adding and maintaining this change. We are currently trying to limit scope to focus on bug fixes and stabilization efforts before adding too many features to instrumentation libraries. I hope this explanation helps.

joelmukuthu:
Yes that helps, thanks for responding and for all the work you and your team put in!

PR #1460, opened 2023-04-05
chore: exclude examples from renovate

Flarna:
Exclude examples from renovate runs because there is no CI to verify that they still work after the update.

As they are terrible outdated already any update is likely breaking them because some still use components not released anymore and therefore no compatibility to actual release without code change.

Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1459

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npx --no-install commitlint --edit $1
--- a/renovate.json
+++ b/renovate.json
@@ -26,7 +26,7 @@
     }
   ],
   "ignoreDeps": ["lerna", "lerna-changelog"],
-  "ignorePaths": ["archive/**"],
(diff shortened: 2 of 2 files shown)


dyladan (approved):

trentm:
@Flarna Was it intentional that the .husky/... file was deleted as part of this change.
I got here because https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1788 recently updated the `@commitlint/cli` dep. I started looking for a usage of the `commitlint` CLI in this repo and couldn't find it.

Flarna:
Sorry, can't remember for sure. Long time ago husky was used to avoid malformed commits already in user repo via husky. That was terrible annoying and therefore removed. e.g. there is anyway no need to have validated commit message on each commit because we squash during integration and commits done in browser skipped the check.

I think in core repo a CI lint step checks the commit message meanwhile but contrib repo tends to be behind regarding tooling. I guess there are some leftovers.

Refs: https://github.com/open-telemetry/opentelemetry-js/issues/2078

PR #1468, opened 2023-04-17
fix(deps): update otel core experimental to ^0.38.0

pichlermarc:

- Tests are currently failing over at #1459, we cannot update to 0.38.0 until tests pass there.

- I'll be fixing all the problems here in this PR.


pichlermarc:
Not exactly sure why the coverage step is failing. Investigating...

Flarna (approved):

Flarna:
Is there any chance to get this merged soon?

My local build files with following error:
```
test/InstanaAgentDetectorIntegrationTest.test.ts:62:7 - error TS2322: Type 'import("c:/workspaces/GitHubForks/opentelemetry-js-contrib/node_modules/@opentelemetry/resources/build/src/Resource").Resource' is not assignable to type 'import("c:/workspaces/GitHubForks/opentelemetry-js-contrib/node_modules/@opentelemetry/sdk-node/node_modules/@opentelemetry/resources/build/src/Resource").Resource'.
  Types have separate declarations of a private property '_syncAttributes'.

62       resource: globalResource,
```

I assume the build in #1471 fails for the same reason (no useful logs in CI for whatever reason).

Above typescript error indicates that more then one version of `@opentelemetry/resources` is there which is likely because instana detector moved to 0.37 of `@opentelemetry/sdk-node`  which in turn pulls in 0.37 of other experimental packages and maybe other version of GA core packages.
No idea why main broke without a PR. Maybe because OTel core was released and GA packages are selected via ^here.

pichlermarc:
> Maybe because OTel core was released and GA packages are selected via ^here.

I think this is most likely the issue.

I pushed the last commit again to re-run the `codecov` step, it does not seem to update the comment/check-status even when re-running the pipeline to upload new reports. Looks like the previous failing step was a false positive, now it's reporting the expected coverage.

There is a flaky test in the `instrumentation-nestjs-core` that is triggered by the limited compute resources in CI, which makes the test run into the timeout. I'll create an issue to follow up on this but I don't think this should prevent this PR from merging.

pichlermarc:
I was not able to reproduce the coverage problem locally. This seems to be a problem with codecov not merging the coverage correctly on `main`, where it only takes `web` package coverage into account - which seems to have a higher percentage of covered lines than the `node` packages:

https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/tree/main
https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/tree/renovate%2Fotel-core-experimental

I believe this should be okay to merge even with the `codecov` step failing. 
It's out of scope for this PR to fix it but I'll create an issue to follow up on this.

**This branch's coverage report (detecting 4706 lines and including detectors)**
image

**`main` branch's coverage report (detecting only 906 lines)**
image

dyladan (approved):

PR #1470, opened 2023-04-19
chore(mysql2): support version 3

### LINKED ISSUE #1372: Support for mysql2 version 3
Current `mysql2` instrumentation has [support for version range `['>= 1.4.2 < 3.0']`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/ff109b77928cc9a139a21c63d6b54399bb017fa4/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts#LL53C9-L53C27).

The package released a new major version [3.0.0](https://www.npmjs.com/package/mysql2/v/3.0.0) at 12/01/2023, and we should update the instrumentation accordingly:
- supported version for patch
- README
- unit tests
- test-all-versions config

haddasbronfman:

- Fixes #1372 

- update README, instrumentation patch, and tav.yml


pichlermarc (approved):
Thanks for adding support for v3 :slightly_smiling_face: 
The failing codecov check looks to be related to #1473, so I think we can ignore the failing check here.

Issue #1465, opened 2023-04-14
Host-Metrics is no more compatible with old NodeJS versions because of  process.memoryUsage.rss


JeremyLafontaine:
### What version of OpenTelemetry are you using?
    "@opentelemetry/api": "^1.4.1",
    "@opentelemetry/auto-instrumentations-node": "^0.36.5",
    "@opentelemetry/exporter-trace-otlp-http": "^0.37.0",
    "@opentelemetry/host-metrics": "^0.32.0",
    "@opentelemetry/sdk-node": "^0.37.0",

### What version of Node are you using?
v14.17.3

### What did you do?

I tried to use @opentelemetry/host-metrics

### What did you expect to see?
System metrics being exported

### What did you see instead?

> TypeError: process.memoryUsage.rss is not a function
>     at getProcessMemoryData (C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\host-metrics\build\src\stats\common.js:103:32)
>     at C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\host-metrics\build\src\metric.js:172:75
>     at C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\sdk-metrics\build\src\state\ObservableRegistry.js:101:47
>     at Array.map (<anonymous>)
>     at ObservableRegistry._observeBatchCallbacks (C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\sdk-metrics\build\src\state\ObservableRegistry.js:99:37)
>     at ObservableRegistry.observe (C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\sdk-metrics\build\src\state\ObservableRegistry.js:75:43)
>     at MeterSharedState.collect (C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\sdk-metrics\build\src\state\MeterSharedState.js:61:54)
>     at C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\sdk-metrics\build\src\state\MetricCollector.js:33:139
>     at Array.map (<anonymous>)
>     at MetricCollector.collect (C:\EkoProject\Zombie3\Code\WebServer\node_modules\@opentelemetry\sdk-metrics\build\src\state\MetricCollector.js:33:98)

pichlermarc:
[...]

PR #1471, opened 2023-04-19
fix(host-metrics): fallback to process.memoryUsage()

legendecas:

- Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1465

- Fallback to `process.memoryUsage()` when `process.memoryUsage.rss()` is not available.


Flarna (approved):

pichlermarc (approved):
Thanks for taking care of this :slightly_smiling_face:

PR #1480, opened 2023-04-27
fix(mysql): metrics - align metrics description and unit to semconv

haddasbronfman:

While reviewing this PR: #1170, there were to comments that were relevant to mysql metrics too:
1. regarding the [metric description](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1170/#discussion_r1172510542) 
2. regarding the [metric unit](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1170/#discussion_r1172506395)

1. the description was replaced to 'The number of connections that are currently in state described by the state attribute.'
2. the unit was replaced to {connection}


pichlermarc (approved):
Thank you for following up on this :slightly_smiling_face:

PR #1490, opened 2023-05-07
fix(mysql2-tav): add 'pretest' command to tav.yml for mysql2 version 3

haddasbronfman:

- tav is failing on the main branch because of [this problem](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/4906701688/jobs/8761324293#step:9:20761).
I think it happens because mysql2 version3 doesn't have a 'pretest' command in the tav.yml (my fault because I added this version) and this is why it complains `Cannot find module '@opentelemetry/contrib-test-utils'`

- I added a 'pretest' command for version3 test like the other version (v1 & v2) have. 

adding a screenshot of the error in case the link won't work:
image


pichlermarc (approved):
Looks good, thank you for taking care of this. :)

Issue #937, opened 2022-03-07
instrumentation-aws-sdk error handling incorrect for aws.error leading to Invalid attribute value set for key


jreichhold:
### What version of OpenTelemetry are you using?
"aws-sdk": "^2.675.0",
"@opentelemetry/instrumentation-aws-sdk": "^0.5.0",

### What version of Node are you using?
Node 12.22.9

### What did you do?
When I turned on tracing in my production env I started seeing lots of
```Invalid attribute value set for key: aws.error```

### What did you expect to see?
No error for errors in response and instead to record the error

### What did you see instead?
Lots of errors

### Additional context
https://github.com/open-telemetry/opentelemetry-js-contrib/blame/09b8555007c3c05ad046dd67925f3640a7b35fbe/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts#L321

has 
```span.setAttribute(AttributeNames.AWS_ERROR, response.error);```

Which calls https://github.com/open-telemetry/opentelemetry-js/blob/315abae2fb197259515d0b5139af5cc8ced25594/packages/opentelemetry-sdk-trace-base/src/Span.ts

Which has the following check:
```
if (!isAttributeValue(value)) { api.diag.warn(Invalid attribute value set for key: ${key}); return this; }
```

Which calls isAttributeValue in:
https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-core/src/common/attributes.ts

which has the following checks:
```
export function isAttributeValue(val: unknown): val is SpanAttributeValue {
  if (val == null) {
    return true;
  }

  if (Array.isArray(val)) {
    return isHomogeneousAttributeValueArray(val);
  }

  return isValidPrimitiveAttributeValue(val);
}
```

My suspicion is instead of calling Span.setAttribute on an Error one should do one of the following
- likely call Span.recordException
- Stringify the exception
- Use the error.message only

Seems this should fail for everyone as request.error is an object and not a string.

**Comment by jreichhold** [comment]
Digging into the AWS response object I see the following: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Response.html#error-property

```
error ⇒ Error (readonly)
Note: This attribute is only filled if a service or networking error occurs.
An structure containing information about a service or networking error.

Returns:

(Error) —
code [String] a unique short code representing the error that was emitted.
message [String] a longer human readable error message
retryable [Boolean] whether the error message is retryable.
[...]

PR #1495, opened 2023-05-15
fix(aws-sdk): correct setting error in attributes

thib3113:

This PR solve #937 .

to summarize : 
 - response.error is not a valid attribute value (an exception)
 - produce a warning and don't set the information

replace setting the attribute by a string summarizing the error + recording the exception

I test it manually by deploying a lambda, and throwing error


blumamir (commented):
Thanks for the fix.

Few questions please:
1. is `response.error.code` guaranteed to exist? What value is placed in it? is it the HTTP status code? What would be the text in the attribute if the `error.code` is missing?

2. I support recording the error with `span.recordException`, but then what value do we get from also recording it as span attribute? Since the `aws.error` attribute is not in spec, I would consider recording it only once and removing the attribute.

Please also add/update the tests for this fix.

Thanks

blumamir:
[The implementation code for `recordException`](https://github.com/open-telemetry/opentelemetry-js/blob/758c7af0c183b36bbce0b0dfbda3dd43bdc0ca8d/packages/opentelemetry-sdk-trace-base/src/Span.ts#L275) should already take care of extracting both the `code` and `message` into span event attributes, so I think you are covered in this area. 

```js
    const attributes: SpanAttributes = {};
    if (typeof exception === 'string') {
      attributes[SemanticAttributes.EXCEPTION_MESSAGE] = exception;
    } else if (exception) {
      if (exception.code) {
        attributes[SemanticAttributes.EXCEPTION_TYPE] =
          exception.code.toString();
      } else if (exception.name) {
        attributes[SemanticAttributes.EXCEPTION_TYPE] = exception.name;
      }
      if (exception.message) {
        attributes[SemanticAttributes.EXCEPTION_MESSAGE] = exception.message;
      }
      if (exception.stack) {
        attributes[SemanticAttributes.EXCEPTION_STACKTRACE] = exception.stack;
      }
    }
```

I personally would remove the `aws.error` attribute, since it's not in spec, and to my understanding the proper place for recording errors is as a span event. Keeping it would cause duplication of the values. Instrumentations are experimental so we don't need to worry about doing such changes if they align things to the spec. 

If you do choose to keep the span attribute `aws.error` (which is fine by me), I suggest verifying that the generated text will make sense in the (possibly) absence of `code` and `message`, so we don't end up with texts like `[undefined] undefined`

thib3113:
So, I remove it .

I also remove it from the readme (maybe need to write the "breaking change" somewhere ?) . 

Also, I'm not a big fan of my tests about : 
 - https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1495/files#diff-8a5d80a6ebde5c7cb5914f94763057dfa7e09b171a2263e6021fd6ecc7ee1c60R37 => need to import an un-imported from `@opentelemetry/sdk-trace-base` . So if the name change, the tests will pass . but, it can be a problem if the path change
  - https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1495/files#diff-8a5d80a6ebde5c7cb5914f94763057dfa7e09b171a2263e6021fd6ecc7ee1c60R280-R284 => search in event with the name of an exception, and continue only if found one exception ... it will allow futur events to be included, but not another exception

plugins/node/opentelemetry-instrumentation-aws-sdk/test/aws-sdk-v2.test.ts
```diff
@@ -33,6 +33,8 @@ import { AttributeNames } from '../src/enums';
 import { mockV2AwsSend } from './testing-utils';
 import { expect } from 'expect';
 import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
+//just to match the name, even if it changes. if path change this will crash
+import { ExceptionEventName } from '@opentelemetry/sdk-trace-base/build/src/enums';
```
blumamir, line 37: looks like other places in the repo just use the constant string 'exception', examples: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js-contrib%20%27exception%27&type=code

I would recommend doing the same here too so that we don't import from an internal file as you mentioned.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/aws-sdk-v2.test.ts
```diff
@@ -275,9 +277,21 @@ describe('instrumentation-aws-sdk-v2', () => {
         const awsSpans = getAwsSpans();
         expect(awsSpans.length).toBe(1);
         const [spanCreateBucket] = awsSpans;
-        expect(spanCreateBucket.attributes[AttributeNames.AWS_ERROR]).toBe(
-          responseMockWithError.error
+        const exceptionEvent = spanCreateBucket.events.filter(
+          event => event.name === ExceptionEventName
         );
+        //need to handle test differently if more than one exception
```
blumamir, line 283: since the current implementation can only record a single error, I would consider removing this comment.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/aws-sdk-v2.test.ts
```diff
@@ -275,9 +277,21 @@ describe('instrumentation-aws-sdk-v2', () => {
...
+        expect(exceptionEvent.length).toBe(1);
+
+        expect(exceptionEvent[0]).toStrictEqual(
+          expect.objectContaining({
+            name: 'exception',
+            attributes: {
+              'exception.message': 'something went wrong',
```
blumamir, line 290: This PR is about the error being an object with message and code, and not a string (which was the original motivation for the change). This test uses `responseMockWithError` which is

```js
  const responseMockWithError = {
    requestId: '0000000000000',
    error: 'something went wrong',
    httpResponse: {
      statusCode: 400,
    },
  };
```

Should we add a test where the error is something like: 
```js
{
  code: 1234,
  message: 'test error as object'
}
```

and also assert the error code in the test?

blumamir (approved):

PR #1530, opened 2023-06-09
attempt to instrument esm-lambda

Ankcorn:

- Adds support for instrumenting esm lambda functions

- For a lambda to use esm in needs the .esm extension. This resolves that correctly
- Adds a test to verify esm is being handled (its not :confused: )


blumamir:
@Ankcorn 
Thanks for working on this 

Is this draft PR still being used? we can keep it open if you need it, or close it if you moved to something else

Ankcorn:
Please close it, we took a different track because we couldn't get Esm lambdas instrumented automatically.

Thanks

PR #1534, opened 2023-06-14
fix(connect): Skip update HTTP's span name and update RpcMetadata's route instead

### LINKED ISSUE #1381: Setting HTTP_ROUTE inconsistent accross HTTP server frameworks
`HTTP_ROUTE` propagation from HTTP server frameworks to HTTP span is inconsistent:

* connect updates only the span name
* express, koa and restify update the span name and return the route via `rpcData` where Http instrumentation pickes it up
* hapi and fastify update the span name and additionally set `HTTP_ROUTE` directly on HTTP span

I think all frameworks should avoid modifications of the HTTP span directly instead return the `route` info via `rpcData` and leave it up to the HTTP instrumentation to build the final span name and set the `HTTP_ROUTE` attribute.

Setting `HTTP_ROUTE` directly breaks for example the assumption done in this [PR](https://github.com/open-telemetry/opentelemetry-js/pull/3581).

chigia001:

This PR fixes the above issue for [connect](https://github.com/senchalabs/connect)

- Skip modifying HTTP's Span name directly in instrumentation-connect code


plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -120,16 +120,16 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
...
+      const [resArgIdx, nextArgIdx] = isErrorMiddleware ? [2, 3] : [1, 2];
       const res = arguments[resArgIdx] as ServerResponse;
       const next = arguments[nextArgIdx] as NextFunction;
 
       const rpcMetadata = getRPCMetadata(context.active());
       if (routeName && rpcMetadata?.type === RPCType.HTTP) {
-        rpcMetadata.span.updateName(`${req.method} ${routeName || '/'}`);
+        setRPCMetadata(
+          context.active(),
+          Object.assign(rpcMetadata, { route: routeName || '/' })
```
chigia001, line 131: I'm super new to connect so I'm not sure how this will behave for a nested route.

should we expand the RpcMetadata's route instead?

Maybe another PR?

Flarna, line 131: Why not simply set the route instead of using `Object.assign`? e.g.
```ts
if (routeName && rpcMetadata?.type === RPCType.HTTP) {
  rpcMetadata.route = routeName;
}
```
There should be also no need to default to `/` as the if ensures that it is no empty.

Flarna: I think we should report the final route (e.g. `/foo/bar` in above case) but I would move this to a separate PR.

chigia001: update, I follow this pattern from some of the other frameworks.

Flarna (approved):

chigia001:
@Flarna seem like there something wrong with last run, it fail at docker pull step.

Flarna:
@open-telemetry/javascript-maintainers Could someone please take a look on this as connect instrumentation lacks a maintainer but I think this change is useful.

pichlermarc (approved):
Thanks for taking care of this - happy to see this behavior aligned with the other instrumentations :tada:

PR #1536, opened 2023-06-14
feat(instrumentation-pg): Span Names Matching Semantic Conventions

xoscar:

- Currently, the span names generated by the instrumentation library are not matching what is defined in the database spans [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/specification/trace/semantic_conventions/database.md).
Which calls out that names for db spans should be composed by: `<db.operation> <db.name>.<db.sql.table>` or `<db.operation> <db.name>`  if the table is not available.

- Removes the Component prefix from all span names
- Uses the command as operation plus the database to name the span
- Updates the connect spans to not include the component name

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
@@ -46,12 +46,10 @@ import {
 } from '@opentelemetry/semantic-conventions';
 import { VERSION } from './version';
 
-const PG_POOL_COMPONENT = 'pg-pool';
-
 export class PgInstrumentation extends InstrumentationBase {
   static readonly COMPONENT = 'pg';
 
-  static readonly BASE_SPAN_NAME = PgInstrumentation.COMPONENT + '.query';
+  static readonly BASE_SPAN_NAME = 'query';
 
   constructor(config: PgInstrumentationConfig = {}) {
     super(
@@ -138,7 +136,7 @@ export class PgInstrumentation extends InstrumentationBase {
         }
 
         const span = plugin.tracer.startSpan(
-          `${PgInstrumentation.COMPONENT}.connect`,
+          `connect${this.database ? ` ${this.database}` : ''}`,
           {
             kind: SpanKind.CLIENT,
             attributes: {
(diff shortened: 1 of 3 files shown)


plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -46,12 +46,10 @@ import {
 } from '@opentelemetry/semantic-conventions';
 import { VERSION } from './version';
 
-const PG_POOL_COMPONENT = 'pg-pool';
```
xoscar, line 49: I think we should consider using an attribute to keep track of the `pg-pool` vs `pg` component instead of having it at the name level

haddasbronfman: If we remove PG_POOL_COMPONENT, we'll lose this information. We can't know if the user called pg.connect() of pg-pool.connect(). I would like to see what other @open-telemetry/javascript-approvers think.

haddasbronfman (commented):
Thanks for handling this and helping improve the code :)
Overall LGTM (just one thought about the PG_POOL_COMPONENT)

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
Flarna: I think the correct way to reflect this would be the instrumentation scope. So 1 tracer for pg and one for pg-pool.
But not sure if it is important at all to have this info available.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -138,7 +136,7 @@ export class PgInstrumentation extends InstrumentationBase {
         }
 
         const span = plugin.tracer.startSpan(
-          `${PgInstrumentation.COMPONENT}.connect`,
+          `connect${this.database ? ` ${this.database}` : ''}`,
```
blumamir, line 139: Here we are doing 
```js
`connect${this.database ? ` ${this.database}` : ''}`
```

and below we have 
```js
`connect ${this.options.database}`,
```

and in utils.ts we have:
```js
`${command}${dbName ? ` ${dbName}` : ''}`
```

Perhaps we can extract this logic into a function in utils that receives command and dbName and return a db span name from these values? and then use it in all 3 places?

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -46,12 +46,10 @@ import {
 } from '@opentelemetry/semantic-conventions';
 import { VERSION } from './version';
 
-const PG_POOL_COMPONENT = 'pg-pool';
-
 export class PgInstrumentation extends InstrumentationBase {
   static readonly COMPONENT = 'pg';
 
-  static readonly BASE_SPAN_NAME = PgInstrumentation.COMPONENT + '.query';
+  static readonly BASE_SPAN_NAME = 'query';
```
blumamir, line 52: As this variable is only used in `utils.ts`, and now only contains the value 'query' without any component part, I suggest that we move it to `utils.ts` as const, and give it a better name, like dbOperationQuery which communicate it's role more clearly

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
blumamir: I would love to see it captured as a scope attribute.

@xoscar if you want to include it in this one, please go ahead and add it.

Since currently it is only being published as part of the span name which isn't very useful anyway, I think maybe we can postpone the introduction of a new attribute to a follow-up PR. We can open an issue for that and merge the current PR.

@haddasbronfman @Flarna WDYT?

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -46,12 +46,10 @@ import {
 } from '@opentelemetry/semantic-conventions';
 import { VERSION } from './version';
 
-const PG_POOL_COMPONENT = 'pg-pool';
-
 export class PgInstrumentation extends InstrumentationBase {
   static readonly COMPONENT = 'pg';
```
blumamir, line 50: optional: I think it's a good time to get rid of this `COMPONENT` var which, after your PR, will only be used in one place for log.
If you have time and energy, please fell free to remove it as well :)

blumamir:
@xoscar , are you still interested in this PR?

PR #1543, opened 2023-06-20
feat(aws-ecs): add cloud resource attributes for fargate

davidweterings:

ECS on Fargate containers don't include Cloud resource attributes like account id/region and availability zone (if available). This PR determines these values based on the ECS Metadata query. I've looked at the Go implementation (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5fff1040a2844d4d435359b8108202317762593d/processor/resourcedetectionprocessor/internal/aws/ecs/ecs.go#L69) which works in a similar fashion.

- Set AWS Cloud resource attributes (account id, region, availability zone) based on ECS Fargate metadata


haddasbronfman (approved):
Thanks for adding that :)

PR #1545, opened 2023-06-20
feat(minification): Add importHelpers and tslib as a dependency

MSNev:

To better support minification of the generated code we need to change the default settings by adding the following to the tsconfg.base.json.

Fixes #3913

Adds `importHelpers` to `tsconfig.base.json`
Updates all `package.json` to include `tslib` as a dependency with "^2.3.1" as this version is the first version that supported TypeScript 4.4, this should allow all v2.x versions of `tslib` to be used.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)

This is not expected to be a code breaking change

## How Has This Been Tested?

Compiled and built locally using `npm run compile` and `npm test`

## Checklist:

- [X] Followed the style guidelines of this project
- [ ] Unit tests have been added
- [ ] Documentation has been updated


detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
```diff
@@ -60,7 +60,8 @@
   },
   "dependencies": {
     "@opentelemetry/resources": "^1.0.0",
-    "@opentelemetry/semantic-conventions": "^1.0.0"
+    "@opentelemetry/semantic-conventions": "^1.0.0",
+    "tslib": "^2.3.1"
```
kirrg001, line 64: Wondering why this is a prod dependency.

pichlermarc (approved):

legendecas (approved):

PR #1546, opened 2023-06-21
fix(ci): use npx to run lerna

pichlermarc:

- previously we had the problem that `lerna` was not pinned on installation and `lerna@7.0.2` was used #1541 
- now we have the problem that `lerna@7.0.0` is also [included in the GitHub runner `ubuntu-latest` image](https://github.com/actions/runner-images/blob/ubuntu22/20230611.1/images/linux/Ubuntu2204-Readme.md#project-management) and that one is used by our scripts. 

- use `npx` for lerna everywhere to prefer the version we install over the one from the image


pichlermarc:
(applied contrib test utils label to test the tav workflow)

pichlermarc:
(applying labels to see if the tav script works)

martinkuba (approved):

Issue #1547, opened 2023-06-22
Express instrumentation does not normalize paths (removing double slashes)


rrva:
### What version of OpenTelemetry are you using?

auto-instrumentations-node 0.37.1
instrumentation-express 0.32.4

### What version of Node are you using?

18

### What did you do?

define some routes like this:

```
server.use(`/foo/`,
router.use('/v5/',
router.use(
    '/bar/',
router.get('/refresh',
```

### What did you expect to see?

http.route attribute in traces should contain:

```
/foo/v5/bar/refresh
```

Instrumentation should normalise the path in the same way that Express does (Express knows that declaring /v5/ inside /foo/ means /foo/v5/ not /foo//v5/ )

### What did you see instead?

```
/foo//v5//bar//refresh
```

### Additional context

express 4.18.2

unflxw:
Thanks for reporting this @rrva! Adding a link to the CNCF Slack conversation for additional context: https://cloud-native.slack.com/archives/C01NL1GRPQR/p1687429185406239

pkanal:
I took a closer look at this and I think [`http.route` actually gets set in the HTTP instrumentation](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-http/src/utils.ts#L545).  The value for route is [set by `rpcMetadata`](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-http/src/utils.ts#L533) which comes from the trace package. So I'm wondering whether the normalization would need to happen either in the trace package or the HTTP instrumentation package?

pichlermarc:
Hmm, I think `rpcMetadata.route` should be pre-formatted in the server framework's instrumentation (somewhere before this line in the express instrumentation), as the http-instrumentation can't know how the server framework usually formats the route string. :thinking: 

The [semconv spec](https://github.com/open-telemetry/semantic-conventions/blob/main/specification/http/http-spans.md) mentions that we should use 

> The matched route (path template in the format used by the respective server framework).

While unlikely, if `//` as a substring of `route` is intended in some other server framwork for some reason, we'd modify the intended value of `http.route` in the http instrumentation if we were to normalize there.

unflxw:
[...]

PR #1555, opened 2023-07-04
fix(connect): fix wrong rpcMetada.route value not handle nested route

chigia001:

- connect allow having nested connect instance, which can be used to setup nested route. When using this setup, the current instrumentation code will alway override with the sub-route and not take into account the route of parent connect instance 
- This is discover when implement #1534 ([discusion](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1534#discussion_r1229283320))
- Also need to take into the account of trailing splash in route middlewere when construct nested route. See #1547 

- Need another layer of patching for connect's `handle` method. Whenever `handle` is call, we add another layer in the stack. If `handle` also call with an `out` method, we must pop the stack before out is execute. (this is opposite with current patched next where we call next first before finish span)
- On route's middlewere, each time connect call the route's handle, update the latest value in the stack with middlewere's route value, use current stack value to calculate current rpcMetadata.route value.
- This setup allow capture the last route middlewere executed with correct nested app.


plugins/node/opentelemetry-instrumentation-connect/test/instrumentation.test.ts
```diff
@@ -243,5 +243,88 @@ describe('connect', () => {
...
+      });
+
+      app.use('/foo/bar/test', (req, res, next) => {
+        next();
+      });
+
+      await httpRequest.get(`http://localhost:${PORT}/foo/bar/test`);
+      rootSpan.end();
+
+      assert.strictEqual(rpcMetadata.route, '/foo/bar/test');
```
chigia001, line 327: if we don't pop the stack before call out this will become `/foo/foo/bar/test`

pichlermarc:
Broken CI will be fixed once #1556 merged, sorry for the inconvenice. :disappointed:

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -180,4 +212,33 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
...
+      addNewStackLayer(req);
+
+      function completeStack() {
+        req[_LAYERS_STORE_PROPERTY].pop();
+      }
+      const done = out
+        ? instrumentation._patchOut(out as NextFunction, completeStack)
+        : undefined;
+
+      return original.apply(this, [req, res, done as Function]);
```
Flarna, line 233: not sure if it is relevant but this may result in an visible side effect.
inside the original function `arguments.length` is now always 3, it was potentially 2 before.
Also any extra argument after `out` is lost.

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -180,4 +212,33 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
...
+      req: PatchedRequest,
+      res: ServerResponse,
+      out?: Function
+    ): ReturnType<Server['handle']> {
+      addNewStackLayer(req);
+
+      function completeStack() {
+        req[_LAYERS_STORE_PROPERTY].pop();
+      }
+      const done = out
```
Flarna, line 229: Not sure if it can happen that `out` is not falsy but not a function.

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -180,4 +212,33 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
...
+        ? instrumentation._patchOut(out as NextFunction, completeStack)
+        : undefined;
+
+      return original.apply(this, [req, res, done as Function]);
+    };
+  }
+
+  public _patchOut(out: NextFunction, completeStack: () => void): NextFunction {
+    return function nextFunction(this: NextFunction, err?: any): void {
```
Flarna, line 238:
```suggestion
    return function nextFunction(this: NextFunction, ...args: any[]): void {
```
and below you could use `Reflect.apply(out, this, args);`
or even use `Reflect.apply(out, this, arguments);`

plugins/node/opentelemetry-instrumentation-connect/src/internal-types.ts
```diff
@@ -14,9 +14,14 @@
  * limitations under the License.
  */
 
-import type { HandleFunction, Server } from 'connect';
+import type { HandleFunction, IncomingMessage, Server } from 'connect';
+
+export const _LAYERS_STORE_PROPERTY = '__ot_middlewares';
```
Flarna, line 19: If you use a `symbol` it would not show up as enumerable property and also no clash is possible.

Flarna (approved):

chigia001:
I think this PR also have some GH action's cache issue like my PR for express. Not sure how we can resolve those issue.

chigia001:
can we merge this or add pkg:instrumentation-connect label so we can test tav for this?

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -180,4 +212,33 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
...
+    const instrumentation = this;
+    return function (this: Server): ReturnType<Server['handle']> {
+      const [reqIdx, outIdx] = [0, 2];
+      const req = arguments[reqIdx] as PatchedRequest;
+      const out = arguments[outIdx];
+      addNewStackLayer(req);
+
+      function completeStack() {
+        req[_LAYERS_STORE_PROPERTY].pop();
+      }
```
blumamir, line 226: since above we have 2 function to manipulate the `req[_LAYERS_STORE_PROPERTY]` property, it might be readable to also add this function there, something like:
```js
const addNewStackLayer = (request: PatchedRequest) => {
...
}

const replaceCurrentStackRoute = (request: PatchedRequest, value?: string) => {
...
}

const popLastRoute = (request: PatchedRequest) => {
...
}

```

plugins/node/opentelemetry-instrumentation-connect/test/instrumentation.test.ts
```diff
@@ -243,5 +243,88 @@ describe('connect', () => {
...
+
+      app.use('/foo/', nestedApp);
+      nestedApp.use('/bar/', (req, res, next) => {
+        next();
+      });
+
+      await httpRequest.get(`http://localhost:${PORT}/foo/bar`);
+      rootSpan.end();
+
+      assert.strictEqual(rpcMetadata.route, '/foo/bar/');
```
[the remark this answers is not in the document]
chigia001: we only remove double slash betwen route, not trailing slash

Because the deepest route in that match define with `/bar/`, the rpcMetadata.route still have trailing slash.

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -120,13 +145,20 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
...
       const res = arguments[resArgIdx] as ServerResponse;
       const next = arguments[nextArgIdx] as NextFunction;
 
+      replaceCurrentStackRoute(req, routeName);
+
       const rpcMetadata = getRPCMetadata(context.active());
       if (routeName && rpcMetadata?.type === RPCType.HTTP) {
-        rpcMetadata.route = routeName;
+        rpcMetadata.route = req[_LAYERS_STORE_PROPERTY].reduce(
+          (acc, sub) => acc.replace(/\/+$/, '') + sub
```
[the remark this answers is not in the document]
chigia001: I move this part to utils.ts with comment + unit test.

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
chigia001: instead of making popLastRoute a dedicated method, I make it a return function from addNewStackLayer.

That way we can ensure we can only pop the correct stack's item

blumamir (approved):
LGTM.

Thanks for addressing the comments 🙏 and for these contributions.

Please fix the lint so I can merge

chigia001:
@blumamir fixed

PR #1556, opened 2023-07-04
fix(instrumentation-fastify): fix fastify typescript compilation issue

pichlermarc:
- Looks like recent changes over at fastify desupported older versions of typescript, which breaks CI at the moment
- The PR pins the devDependency to `4.18.0` until https://github.com/fastify/fastify/issues/4873 is resolved
  - if the resolution to this issue is that support will be restored, I'll unpin it again
  - if the resolution to this issue is that support will **not** be restored, I'll look into alternative ways of testing newer versions of fastify (possibly via TAV).


pichlermarc:
@open-telemetry/javascript-approvers PTAL, this is currently blocking all PRs in contrib.

Flarna (approved):

PR #1557, opened 2023-07-05
feat(express): Skip update HTTP's span name and update RpcMetadata's route instead

chigia001:

This PR fixes the above issue for [express](https://github.com/expressjs/express)

- Skip modifying HTTP's Span name directly in instrumentation-connect code
  - This will be also a breaking change as @opentelemetry/instrumentation-express's `spanNameHook`. This config previously support overide HTTP route span's name, which is allow in [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/33113489fb5a1b6da563abb4ffa541447b87f515/specification/trace/semantic_conventions/http.md?plain=1#L70) => this might be something we can introduce back to @opentelemetry/instrumentation-http
  - Due to instrumentation-express alway working in conjunction with instrumentation-http, althrough instrumentation-express can update HTTP span's name, instrumentation-http will overide this name later when `rpcMetadata.route` is set. So there should be no visible change to end-user.
- Refactor some test due to these tests are not properly clean up (`server.close()` is not called) when the test is failing => this cause running test with `npm run test -- --scope @opentelemetry/instrumentation-express` seem stuck without any indicator


chigia001:
PS: I also encounter issue with `npm run test -- --scope @opentelemetry/instrumentation-express` when initate with `npm install`.

The package include a conflict `eslint` (eslint@6.8.0) version that incompatiple with `eslint-plugin-plugin` in root dir. The dependency chain of that version:

```
eslint@6.8.0 extraneous
node_modules/standard/node_modules/eslint
  eslint@"~6.8.0" from standard@14.3.4
  node_modules/standard
    standard@"^14.3.4" from parse-env-string@1.0.0
    node_modules/parse-env-string
      parse-env-string@"^1.0.0" from test-all-versions@5.0.1
      node_modules/test-all-versions
```

plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -27,7 +27,7 @@ export type ExpressRequestInfo = {
   /**
    * If layerType is undefined, SpanNameHook is being invoked to rename the original root HTTP span.
    */
```
chigia001, line 29: I already remove this comment on fork, not sure why GH not picking up that commit 😕

pichlermarc: There's currently an incident at GitHub, i've seen some changes are not accurately reflected across the Github UI at the moment. Guess we'll have to wait it out :confused: 

https://www.githubstatus.com/

Flarna (approved):

JamieDanielson (commented):
Thank you for this, it looks great!

I think the only concern I have on this is related to the breaking change for using the spanNameHook to override the HTTP span name. I'm not sure how common it is to use or if we need a deprecation period. We can still override Express span names with the hook, which I would think is the true purpose of this hook anyway but I know folks like to have options to rename auto-generated span names 🤔 .

> This will be also a breaking change as @opentelemetry/instrumentation-express's spanNameHook. This config previously support overide HTTP route span's name, which is allow in [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/33113489fb5a1b6da563abb4ffa541447b87f515/specification/trace/semantic_conventions/http.md?plain=1#L70) => this might be something we can introduce back to @opentelemetry/instrumentation-http

chigia001:
> Thank you for this, it looks great!
> 
> I think the only concern I have on this is related to the breaking change for using the spanNameHook to override the HTTP span name. I'm not sure how common it is to use or if we need a deprecation period. We can still override Express span names with the hook, which I would think is the true purpose of this hook anyway but I know folks like to have options to rename auto-generated span names 🤔 .
> 
> > This will be also a breaking change as @opentelemetry/instrumentation-express's spanNameHook. This config previously support overide HTTP route span's name, which is allow in [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/33113489fb5a1b6da563abb4ffa541447b87f515/specification/trace/semantic_conventions/http.md?plain=1#L70) => this might be something we can introduce back to @opentelemetry/instrumentation-http

Due to how `instrumentation-http` is working right now, once `rpcMetadata.route` is set, `instrumentation-http` will update the HTTP span name anyway, so the custom name hook for HTTP span in `instrumentation-express` don't have any effect.
If we don't set rpcMetadata.route, then we lost more functionality (missing HTTP span's attribute + metric's attribute)

Another option is to delay this PR after `instrumentation-http` introduce a custom span name hook and instructs the user to use that method instead.

Flarna:
The `applyCustomAttributesOnSpan` hooks in http instrumentation is called after the span name update based on http.route and it gets the span as argument.

So one could actually rename the span as of now. Missing piece for real world is likely that this hook doesn't know the http.route and `span` has no API to read attributes nor name.

In general I think the current solution that express instrumenation is not touching the parent span is the way to go. Having the rpc data as a path to transfer data in a define way is fine.
But an instrumentation should not tinker with foreign spans. If several instrumentations start doing this we may end up in race who offers the "latest" hook.

If we need span name customization for http spans it's should be offered by http instrumentation.

JamieDanielson (approved):
>If we need span name customization for http spans it's should be offered by http instrumentation.

Agreed 💯 .

Thank you both for clarification! This looks good to me.

chigia001:
Don't know wy this PR fail unit test, let me investigate a little bit before merge.

pkanal (approved):

Issue #1473, opened 2023-04-24
[codecov] coverage not reported correctly on `main`


pichlermarc:
### What version of OpenTelemetry are you using?

`codecov` is not reporting coverage correctly on main and most PRs. It's only detecting `906` lines, and looks like it's only using the browser test reports, whereas some PRs detect more lines (`4706`), including the `node` test reports as well, which is the intended amount.

### What did you do?

Created a PR (#1468), had the pipeline run there - the `codecov` step failed with many indirect coverage changes.

### What did you expect to see?

`codecov` detecting all lines in the report on `main`, not only the ones hit in `browser` tests.

### What did you see instead?

`codecov` reports only include files used in `browser` tests on `main`.

### Additional context

[**renovate/otel-core-experimental branch's coverage report (detecting 4706 lines and including detectors)**](https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/tree/renovate%2Fotel-core-experimental)
![image](https://user-images.githubusercontent.com/22105064/233953634-8c240896-81d0-4cd6-b6bf-cd2dcef1c8bb.png)

[**main branch's coverage report (detecting only 906 lines)**](https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/tree/main)
![image](https://user-images.githubusercontent.com/22105064/233954036-783173ed-e7a2-4d99-bdb6-5a4adb2143ae.png)

chigia001:
@pichlermarc @blumamir I double check 2 latest commit on main and seem like there are still some issue as codecov randomly unable to upload our code coverage file. 
```
[2023-08-10T13:39:53.315Z] ['verbose'] https://codecov.io/upload/v4?package=github-action-3.1.4-uploader-0.6.2&branch=main&build=5821384904&build_url=https%3A%2F%2Fgithub.com%2Fopen-telemetry%2Fopentelemetry-js-contrib%2Factions%2Fruns%2F5821384904&commit=4f1124524aee565c3cfbf3975aa5d3d039377621&job=Unit+Tests&pr=&service=github-actions&slug=open-telemetry%2Fopentelemetry-js-contrib&name=&tag=&flags=&parent=
        Content-Type: 'text/plain'
        Content-Encoding: 'gzip'
        X-Reduced-Redundancy: 'false'
[...]

PR #1561, opened 2023-07-06
chore: update lerna to v7 and migrate to npm workspace

### LINKED ISSUE #904: Local build differs from CI build
Bootstrapping differs between local builds using `npm install` in root folder and CI builds.

Local builds execute [`lerna bootstrap --no-ci`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b9a13817b21f6be804479f24619c765b33ec0add/package.json#L25)

CI builds execute [`lerna bootstrap --no-ci --hoist --nohoist='zone.js' --nohoist='mocha' --nohoist='ts-mocha'`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b9a13817b21f6be804479f24619c765b33ec0add/.github/workflows/unit-test.yml#L144)

I can't tell which variant is better/more correct but I think there should be no difference.

Refs.: #900 (fixes an issue which happens in CI but not local)

### LINKED ISSUE #1542: Update dependency lerna to `v7`
`The "bootstrap" command was removed by default in v7, and is no longer maintained.`, which means we'd have to change the way we link packages, possibly by using npm workspaces. 

Further information:

- [ ] #1541, we will have to also update the pinned version here to v7

See also:

Same issue in the otel-js core repo https://github.com/open-telemetry/opentelemetry-js/issues/3894

chigia001:

- As lerna@7 drop support for `lerna bootstrap`, `lerna link`, and `lerna add`, we must depend on the package's manager native solution
Fixes #904 
Fixes #1542

- Add "workspaces" definition to root package.json
- Remove the "prepare" script in each workspace's package.json
- Clean up `gts` as it does not use anymore; I just happened to see this because its part of the `lerna bootstrap` arguments
- Fix any hoist issue that occurred with the new setup:
  - `@types/sinon<10.0.14` depend on `@sinonjs/fake-timers` which install a very old version of sinon => this cause some compile issue
  - `zone.js` => instrumentation-user-interaction depends on zone.js's type, but the way the type is defined in zone.js is generated, it can't be used with the import statement. We must use `/// <reference` syntax so that we don't need to fix the path to `zone.js/dist/zone.js.d.ts`
- Update GH's action cache using setup-node
  - This will cache the ~/.npm instead of node_modules folder and won't cache any generated package-lock.json 
    - this previous setup might create a cache lock file that we unable to replicate on local machine
    - as we also don't use `yarn bootstrap` anymore, the previous [best practice for node lerna cache](https://github.com/actions/cache/blob/main/examples.md#node---lerna) not apply for this case
- Remove lerna:link in .tav.yml file

There might be some hoist issues in the future, and the current npmcli script doesn’t provide enough support for us for more complex hoisting like what `lerna` used to provide.

## Next Step

As lerna@7 is just provide some additional feature on top of nx, ee might consider to include nx.json to increase caching between task and apply for nx-cloud. Nx-cloud should be free for OSS project like opentelemetry-js-contrib. Those change might also fix issue with #1473


chigia001:
the default node-14 env is bundle with npm@6's which seem not support npm workspace.
I'm looking into node-18's unit test issue.

chigia001:
nx@16.5.0 are using nodejs's global performance object, which introduced in node@16. this cause nx fail on node@14
https://github.com/nrwl/nx/issues/17986

chigia001:
I will need some more time to test for tav.

DanielLoyAugmedix:
@chigia001 why did this get closed and not merged in?

chigia001:
@DanielLoyAugmedix  I have a new pr for this but still WIP

PR #1563, opened 2023-07-10
fix: Move @types/generic-pool from `dependencies` to `devDependencies`

secondfry:

`@opentelemetry/auto-instrumentations-node` is unusable with TypeScript 5 due to following error:
```
> tsc && tsc-alias && touch .ok

error TS2688: Cannot find type definition file for 'generic-pool'.
  The file is in the program because:
    Entry point for implicit type library 'generic-pool'


Found 1 error.
```

Honesly, this fix is untested, but makes the most sense.

Move @types/generic-pool from `dependencies` to `devDependencies`.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-generic-pool/package.json
+++ b/plugins/node/opentelemetry-instrumentation-generic-pool/package.json
@@ -49,6 +49,7 @@
     "@opentelemetry/context-async-hooks": "^1.8.0",
     "@opentelemetry/sdk-trace-base": "^1.8.0",
     "@opentelemetry/sdk-trace-node": "^1.8.0",
+    "@types/generic-pool": "^3.1.9",
     "@types/mocha": "7.0.2",
     "@types/node": "18.11.7",
     "@types/semver": "7.3.8",
@@ -63,7 +64,6 @@
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.40.0",
     "@opentelemetry/semantic-conventions": "^1.0.0",
-    "@types/generic-pool": "^3.1.9",
     "tslib": "^2.3.1"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-generic-pool#readme"


plugins/node/opentelemetry-instrumentation-generic-pool/package.json
```diff
@@ -63,7 +64,6 @@
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.40.0",
     "@opentelemetry/semantic-conventions": "^1.0.0",
-    "@types/generic-pool": "^3.1.9",
```
blumamir, line 66: Currently this package actually has to be a regular dependency as a type from this package is exported publicly from the package `index.d.ts`.

For end users installing this instrumentation in there application, typescript will attempt to import types from `index.d.ts`, as specified in [`package.json`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/ea865837f88121973ac30d9420f21be262b0a669/plugins/node/opentelemetry-instrumentation-generic-pool/package.json#L6) ):

```
```

It will find this content in `index.d.ts`:
```ts
import GenericPoolInstrumentation from './instrumentation';
export { GenericPoolInstrumentation };
export default GenericPoolInstrumentation;
//# sourceMappingURL=index.d.ts.map
```

And will follow `import GenericPoolInstrumentation from './instrumentation';` and find the following content in `instrumentation.d.ts`:
```ts
import { InstrumentationBase, InstrumentationConfig, InstrumentationNodeModuleDefinition } from '@opentelemetry/instrumentation';
import type * as genericPool from 'generic-pool';
export default class Instrumentation extends InstrumentationBase<typeof genericPool> {
    private _isDisabled;
    constructor(config?: InstrumentationConfig);
    init(): (InstrumentationNodeModuleDefinition<typeof genericPool> | InstrumentationNodeModuleDefinition<any>)[];
    private _acquirePatcher;
    private _poolWrapper;
    private _acquireWithCallbacksPatcher;
}
//# sourceMappingURL=instrumentation.d.ts.map
```

Once it hit the line 
```ts
import type * as genericPool from 'generic-pool';
```

Typescript will emit an error as the types are missing.

You can follow [this documentation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#dependencies) to fix the issue. Basically what you need to do is to remove `import type * as genericPool from 'generic-pool';` from `instrumentation.ts` and replace `typeof genericPool` in code with `any`.

blumamir (changes requested):

blumamir:
@secondfry - are you still interested in this?
Should we keep the PR open?

blumamir:
fixed in #1637

Issue #1434, opened 2023-03-21
Renovate bot doesn't create PRs


Flarna:
It seems renovate bot doesn't work anymore. Otel core repo was released more then a week ago but still no PR.

pichlermarc:
Update: We're having issues troubleshooting as we can't access logs, I'm talking to the renovate folks to see what we can do about it.

pichlermarc:
Getting a hold of the logs took more work than expected. The fix I came up with looks unrelated, but it looks like a docker image that's been removed from Docker Hub is making the bot fail. Opened #1458 to update it to the latest collector.

pichlermarc:
Closing this as the PRs were updated when #1458 merged.
Looks like it's intentional that the bot does not continue when the images go missing from docker hub: https://github.com/renovatebot/renovate/discussions/21248#discussioncomment-5522649

PR #1566, opened 2023-07-11
fix(deps): update otel core experimental to ^0.41.0

nwalters512:

I'm using the latest version of the experimental `@opentelemetry/*` packages in my project. However, because the instrumentations in this repo don't use that latest version, I end up with many copies of `@opentelemetry/instrumentation` installed. This, in turn, results in one instance of `require-in-the-middle` per instrumentation, which severely slows down the loading time of my application.

I patterned this diff off of #1527. I used find/replace to actually bump all the versions.

It seems like renovate should be doing these bumps automatically, but I don't see any recent PRs from it that would have performed these updates.


trentm:
Possibly related: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1434 was a recent-ish fix for renovate for updating deps to ^0.40.0.

The renovate schedule is: `"before 3am on Friday"`
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/renovate.json#L31-L33

The 0.41 release (https://github.com/open-telemetry/opentelemetry-js/pull/3963) was merged last *Thursday*: 

```
commit 06e919d6c909e8cc8e28b6624d9843f401d9b059 (tag: v1.15.0, tag: experimental/v0.41.0)
Date:   2023-07-06T13:06:26+02:00 (5 days ago)

    chore: prepare release 1.15 / 0.41 (#3963)
```

and `@opentelemetry/instrumentation@0.41.0` was published to npm on Thursday:

```
$ npm info @opentelemetry/instrumentation time
...
'0.41.0': '2023-07-06T11:27:10.524Z'
```

I have no experience with Renovate, but I would have expected it to attempt to update on Friday morning.

(I'm here because of the linked require-in-the-middle issue. :)

pichlermarc:
> The renovate schedule is: `"before 3am on Friday"` [`main`/renovate.json#L31-L33](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/renovate.json?rgh-link-date=2023-07-11T20%3A41%3A33Z#L31-L33)
> ...
> I have no experience with Renovate, but I would have expected it to attempt to update on Friday morning.


It should have run on friday morning, but the dependency dashbaord tells me that the PR is awaiting schedule. I think for these dependencies it may be worth to have that run daily - could'nt see any errors in the renovate job log, but they only go back one day, so something may have gone wrong somewhere - all it takes is some timeout while the bot is running, then it takes a week for it to open the PR again. :thinking: 

I forced the creation of the PR now to ensure it's working. I'll also look into bumping the schedule to daily, we typically don't always release on the same day of the week so it would make sense to react more quickly to core updates. :slightly_smiling_face:

pichlermarc (approved):
Thanks for taking care of this, I looks like one was missing but I took the liberty to push a commit that also adds the last one :slightly_smiling_face: 

In the meantime I forced a renovate run, and it seems it's working again. I'll open a PR regardless to bump the schedule for otel core updates to run daily.

PR #1567, opened 2023-07-12
feat(koa): Skip update HTTP's span name and update RpcMetadata's route instead

chigia001:

This PR fixes the above issue for [koa](https://github.com/koajs/koa) when integrate with [@koa/router](https://github.com/koajs/router) as koa are not support routing natively.

- Skip modifying HTTP's Span name directly in instrumentation-koa code


Flarna (approved):

chigia001:
can we merge this or add pkg:instrumentation-koa label so we can test tav for this?

blumamir (approved):

blumamir:
This one is ready to merge, but the build seems to have an issue

> lerna ERR! lifecycle "prepare" errored in "@opentelemetry/instrumentation-koa", exiting 1

Can you please take a look?

chigia001:
My bad, forgot to remove unused import.

blumamir (approved):

PR #1568, opened 2023-07-12
chore: re-enable instrumentation-fastify unit test on node@18

chigia001:

- I test multiple time on my local node@18 and instrumentation-fastify seem working fine with node@18, so we might want to enable this test again.

- remove instrumentation-fastify from the ignore list of node@18 unit test


pichlermarc (approved):
Thanks :slightly_smiling_face:

PR #1570, opened 2023-07-12
feat(hapi): Skip update HTTP's span name and update RpcMetadata's route instead

chigia001:

This PR fixes the above issue for [hapi](https://github.com/hapijs/hapi)

- Skip modifying HTTP's Span name directly in instrumentation-hapi code


Flarna (approved):

chigia001:
can we merge this or add `pkg:instrumentation-hapi` label so we can test tav for this?

Flarna:
Hapi instrumentation has unfortunately no active maintainer.
Could someone from @open-telemetry/javascript-maintainers please take a look. I think this PR looks good (like the similar ones before).

blumamir (approved):

PR #1572, opened 2023-07-12
chore(renovate): split patch and minor rules

haddasbronfman:

Currently, renovate bot groups all minor and patch version upgrades into one PR, which makes it difficult to handle problems when they occur. This PR separates minor and patch into 2 different rules as we have in [core](https://github.com/open-telemetry/opentelemetry-js/blob/main/renovate.json). 

* Create a rule for bumping all patch versions in one PR, excluding "prettier" (same as in [core](https://github.com/open-telemetry/opentelemetry-js/blob/main/renovate.json#L3C20-L3C20))
* Create a rule for all minor version upgrades, and requires #637 for minor version bumps

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,9 +4,15 @@
   ],
   "packageRules": [
     {
-      "groupName": "all non-major dependencies",
-      "updateTypes": ["patch", "minor"],
-      "groupSlug": "all-minor-patch"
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
Thank you for taking care of this :slightly_smiling_face:

PR #1581, opened 2023-07-13
chore: upgrade nock@13.3.1 and re-enable test in node@18 for resource-detector-alibaba-cloud

chigia001:

- Nock@13.3.1 fix issue with their socket implementation, this previously block resource-detector-alibaba-cloud to be test on node@18

- upgrade nock@13.3.1
- re-enable unit test in node@18 for @opentelemetry/resource-detector-alibaba-cloud

not sure if we should merge this instead? #1549


Flarna (approved):

blumamir (approved):
LGTM
I think renovate PR #1549 will be updated once we merge this one.

Wondering why `@opentelemetry/resource-detector-alibaba-cloud` was added to `test-all-versions.yaml` ignore list to begin with as this package is not running tav

PR #1584, opened 2023-07-13
fix(instrumentation-mysql2): sql-common should be a dependency

raphael-theriault-swi:

- `sql-common` was accidentally listed as a dev dependency

- Moves it to dependencies


blumamir (approved):
That sounds right, as it is imported in `instrumentation.ts` file
Doubled checked in other packages that consumes `@opentelemetry/sql-common` and it seems to be correct in `@opentelemetry/instrumentation-pg`.

Thanks for submitting a fix 🥇

raphael-theriault-swi:
Probably warrants a patch release given this makes the instrumentation unusable unless the `pg` instrumentation is also in the tree (it's also on me for putting it in the wrong place to begin with, woops)

PR #1603, opened 2023-07-19
fix(renovate): change renovate patch schedule

haddasbronfman:

- Renovate was scheduled (#1572) to bump all patch versions every day, which turns out to be too noisy. For example: while we fix the issues raised from #1593, Renovate opened #1597 the next day, and #1602 on the day after...

- Change renovate to bump all patch versions `"before 3am on Monday"` - which means at the beginning of the week. 
@open-telemetry/javascript-approvers - please share your thoughts about this schedule.


blumamir (approved):
LGTM
Thanks!

pichlermarc (approved):
I think this is a good schedule. It's more noisy here than in core as there are so many more depdendencies in contrib. Thanks for taking care of this. :)

PR #1611, opened 2023-07-25
fix: Revert "feat(minification): Add importHelpers and tslib as a dependency (#1545)"

dyladan:
This reverts commit 65f612e35c4d67b9935dc3a9155588b35d915482.

see https://github.com/open-telemetry/opentelemetry-js/issues/3976


pichlermarc (approved):
Thanks for fixing this! :slightly_smiling_face:

PR #1614, opened 2023-07-26
fix(express):  fix wrong rpcMetadata.route when router.use no  layer…

biuboombiuboom:

    opentelemetry: 1.4.0
    node: 16.14.2

### What did you do?
        const app = express();
        app.use('/users', router);
        router.get('/:userId', (req, res) => {
          res.status(200).end('user-'+req.params.userId);
        });
        router.use(customMiddleware);
        const bookRouter = express.Router({ mergeParams: true });
        router.use('/:userId/books', bookRouter);
        bookRouter.get('/:bookId', (req, res) => {
          setImmediate(() => {
            res.status(200).end('book-' + req.params.bookId);
          });
        });       

    request http://localhost:8080/proxy/v1/bar/1

    rpcMetadata.route='/users/:userId/books/:bookId'

### What did you see instead?

    rpcMetadata.route='/:userId/books/:bookId'


biuboombiuboom:
amend auth

PR #1616, opened 2023-07-27
chore: bump typescripts to 4.4.4 to fix compile issue with @types/koa

chigia001:

- tsc in instrumentation-cucumber and instrumentation-aws-sdk is older than other package. When run `tsc -p .`for instrumentation-aws-sdk `@types/koa` is type check but the  new version of [@types/koa](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65976) required at least [typescript@4.4](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#symbol-and-template-string-pattern-index-signatures) to work

- bump typescript for instrumentation-cucumber and instrumentation-aws-sdk to 4.4.4

EDIT: enable explainFiles in tsconfig show this message:
```
Entry point for implicit type library 'koa' with packageId '@types/koa/index.d.ts@2.13.7'
```
So clearly tsc is loading all `@types/*` package in node_modules


plugins/node/instrumentation-cucumber/package.json
```diff
@@ -63,7 +63,7 @@
     "sinon": "14.0.0",
     "test-all-versions": "5.0.1",
     "ts-mocha": "10.0.0",
-    "typescript": "4.3.5"
```
chigia001, line 66: Seem like 4.3.5 already support `Symbol Index Signatures`, the main branch have no issue with intrumentation-cucumber but fail with instrumentation-aws-sdk

pichlermarc (approved):
Thanks for opening this PR. I think since `instrumentation-aws-sdk` is experimental, and `instrumentation-cucumber` is not yet released, it's fine to bump typescript; we'll need to fix this to get #1611 out the door.

PTAL @Ugzuzg (component owner for cucumber), @blumamir @carolabadeer (component owner for aws-sdk)

Ugzuzg (approved):

blumamir (approved):
I always favor aligning the dependencies in the monorepo to use the same version for all packages. 

For cucumber, I guess the version is older since the PR was opened long time ago and I think it should be bumped.

For aws-sdk, I am not aware of why it was not bumped in #1374 but that seems to me like it was unintentionally missed and it make sense to me to bump it as well.

One thing I don't yet understand is why tsc complains about koa when the dependency for the koa instrumentation is aligned correctly. Is it related to dependency hoisting?

chigia001:
> One thing I don't yet understand is why tsc complains about koa when the dependency for the koa instrumentation is aligned correctly. Is it related to dependency hoisting?

Yes, we hoist on `@types/*` to root directory and tsc will check all `@types/*` in both your current node_modules folder and root node_modules folder.

Issue #1618, opened 2023-07-31
Express instrumentation 0.33.0 breaks handling http_route metric attribute


mkrudele:
### What version of OpenTelemetry are you using?

```
    "@opentelemetry/api": "1.4.1",
    "@opentelemetry/exporter-prometheus": "0.41.1",
    "@opentelemetry/instrumentation-express": "0.33.0",
    "@opentelemetry/instrumentation-http": "0.41.1",
    "@opentelemetry/sdk-node": "0.41.1",
```

### What version of Node are you using?

```
v18.16.0
```

### What did you do?

I'm using instrumentation-express to have the `http_route` attribute added to metrics. Not using traces, just metrics with Prometheus exporter.  
It is working with `"@opentelemetry/instrumentation-express": "0.32.4"`, getting this output when scraping metrics:  

```
# HELP target_info Target metadata
# TYPE target_info gauge
target_info{service_name="unknown_service:node",telemetry_sdk_language="nodejs",telemetry_sdk_name="opentelemetry",telemetry_sdk_version="1.15.1",process_pid="48486",process_executable_name="node",process_executable_path="/Users/mcrudele/.nvm/versions/node/v18.16.0/bin/node",process_command_args="[\"/Users/mcrudele/.nvm/versions/node/v18.16.0/bin/node\",\"/Users/mcrudele/projects/michele-crudele/echo-server-instrumented/simple-express.js\"]",process_runtime_version="18.16.0",process_runtime_name="nodejs",process_runtime_description="Node.js",process_command="/Users/mcrudele/projects/michele-crudele/echo-server-instrumented/simple-express.js",process_owner="mcrudele"} 1
# HELP http_server_duration Measures the duration of inbound HTTP requests.
# UNIT http_server_duration ms
# TYPE http_server_duration histogram
http_server_duration_count{http_scheme="http",http_method="GET",net_host_name="localhost",http_flavor="1.1",http_status_code="200",net_host_port="6900",http_route="/hello"} 1
http_server_duration_sum{http_scheme="http",http_method="GET",net_host_name="localhost",http_flavor="1.1",http_status_code="200",net_host_port="6900",http_route="/hello"} 12
http_server_duration_bucket{http_scheme="http",http_method="GET",net_host_name="localhost",http_flavor="1.1",http_status_code="200",net_host_port="6900",http_route="/hello",le="0"} 0
[...]

PR #1620, opened 2023-08-01
fix(express): make rpcMetadata.route capture the last layer even when if the last layer is not REQUEST_HANDLER

chigia001:

- fixes: #1618
- The old implementation have 2 different condition for update rpcMetadata.span.name and rpcMetadata.route and we pick the stricter condition.

- Update the condition to match the previous implementation for rpcMetadata.route
- Move this block bellow all the ignore flow, which make it the same as old implementation.


plugins/node/opentelemetry-instrumentation-express/test/ignore-all.test.ts
```diff
@@ -127,14 +127,14 @@ describe('ExpressInstrumentation', () => {
...
-    it('rpcMetadata.route should be modified to /todo/:id', async () => {
+    it('rpcMetadata.route should be undefined', async () => {
       assert.strictEqual(memoryExporter.getFinishedSpans().length, 0);
       await context.with(
         trace.setSpan(context.active(), rootSpan),
         async () => {
           await httpRequest.get(`http://localhost:${port}/toto/tata`);
           rootSpan.end();
-          assert.strictEqual(rpcMetadata.route, '/toto/:id');
+          assert.strictEqual(rpcMetadata.route, undefined);
```
chigia001, line 137: as all the layer is ignore, we don't have anywhere to update rpcMetadata.route

Flarna: hmm, maybe my hint to move this down after the ignore check was not the best idea.
Route detection and span creation are actual two distinct topics so maybe better to detect route independent of layer span creation.
For irrelevant requests (e.g. health) the ignore must anyway start in http layer.

chigia001: Then should we also remove this logic here?

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/8fefd23877a847f0906cf9c582afef9dd55d395b/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L200

When the layer is ignore the route layer also pop.

Flarna (approved):

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -275,6 +275,35 @@ describe('ExpressInstrumentation', () => {
...
+      const rootSpan = tracer.startSpan('rootSpan');
+      let rpcMetadata: RPCMetadata | undefined;
+      const httpServer = await serverWithMiddleware(tracer, rootSpan, app => {
+        app.use(express.json());
+        app.use((req, res, next) => {
+          rpcMetadata = getRPCMetadata(context.active());
+          next();
+        });
+
+        app.use('/bare_middleware', (req, res) => {
```
blumamir, line 289: This test should assert that the route is still captured when 
```js
type === ExpressLayerType.REQUEST_HANDLER
```

This type will be set when `layer.name === 'bound dispatch'`
```js
  } else if (layer.name === 'bound dispatch') {
    return {
      attributes: {
        [AttributeNames.EXPRESS_NAME]: layerPath ?? 'request handler',
        [AttributeNames.EXPRESS_TYPE]: ExpressLayerType.REQUEST_HANDLER,
      },
      name: `request handler${layer.path ? ` - ${layerPath}` : ''}`,
    };
```

Is using the `app.use` triggers a layer with name 'bound dispatch'?

blumamir (approved):
LGTM. Thanks again 🥇 

Added a question regarding the test for this change

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
chigia001:
```

  router.use("/router_use", () => {}); // #3
  router.get('/tada/, () => {}); // #2

  app.use('/toto', router); // #1
  app.use('/bare_use', () => {}); // #3

  app.get('/bare_get', () => {}); // #2
```

"#1", the middleware that bound app.use/router.use with Router object is 'router' layer

"#2" is `bound dispatch`, ie, route handler with HTTP method name. This is for both Express's App and Express's Router object. This is something our unit test cover alot.

"#3" is the last case, IE middleware layer, where app.use/router.use have generic handler. It can be with or without the route's parameter

Most of our test with app.use/router.use is for common's middleware use case, that don't have layername at the begining.

This new test for specificly app.use with route layer.

chigia001: I add additional test for these 2 case, as I don't see we have specific test for them
```
  router.use("/router_use", () => {}); // #3
  app.get('/bare_get', () => {}); // #2
```
internally I think they are the same with case that we already cover but just to make sure.

blumamir (approved):

Issue #1519, opened 2023-05-25
instrumentation-fastify fails to initialise if "type" is set to "module" in package.json


mgcrea:
### What version of OpenTelemetry are you using?

Latest versions available:

```
    "@opentelemetry/api": "^1.4.1",
    "@opentelemetry/auto-instrumentations-node": "^0.37.0",
    "@opentelemetry/exporter-metrics-otlp-proto": "^0.39.1",
    "@opentelemetry/exporter-prometheus": "^0.39.1",
    "@opentelemetry/exporter-trace-otlp-proto": "^0.39.1",
    "@opentelemetry/instrumentation": "^0.39.1",
    "@opentelemetry/instrumentation-fastify": "^0.31.3",
    "@opentelemetry/instrumentation-http": "^0.39.1",
    "@opentelemetry/resources": "^1.13.0",
    "@opentelemetry/sdk-metrics": "^1.13.0",
    "@opentelemetry/sdk-node": "^0.39.1",
    "@opentelemetry/sdk-trace-base": "^1.13.0",
    "@opentelemetry/sdk-trace-node": "^1.13.0",
    "@opentelemetry/semantic-conventions": "^1.13.0",
```

### What version of Node are you using?

`v18.16.0`

### What did you do?

Tried to use `@opentelemetry/instrumentation-fastify` on a new project did not understand why the instrumentation did not work.

### What did you expect to see?

`@opentelemetry/instrumentation-fastify Applying patch for fastify@4.17.0` 

### What did you see instead?

nothing

### Additional context

instrumentation-http works in both modes:

```
@opentelemetry/instrumentation-http Applying patch for http@18.16.0
@opentelemetry/instrumentation-http Applying patch for https@18.16.0
```

Flarna:
support for instrumenting ESM was recently merged (see https://github.com/open-telemetry/opentelemetry-js/pull/3698) but it's not yet in a release.

Please retest once a new release is out.

dyladan:
New release is out. Can you please confirm if this is fixed?

pichlermarc:
Closing this as it should work as of the latest release.
Feel free to comment if we should re-open the issue. :slightly_smiling_face:

nlindley:
This appears to still be a problem. We converted the Fastify example to ESM and see the HTTP modules being patched, but not fastify. https://github.com/nlindley/opentelemetry-js-contrib/compare/main...fastify-esm

nlindley:
[...]

PR #1624, opened 2023-08-07
fix(fastify): Make sure consturctor patching works with esm

mydea:

instrumentation-fastify does not work in ESM mode. This PR fixes this so that fastify can be instrumented when using esm.

The underlying problem is that in ESM mode, the `moduleExports` is not === the fastify constructor. Instead, this can _only_ be found under the `default` key.

This changes the wrapping behavior to always use the named `fastify` export, which should always exist.

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1519


Flarna:
Is there an impact to commonjs users using fastify like this (copied from fastify docs)?
```js
const fastify = require('fastify')({
  logger: true
})
```

mydea:
> Is there an impact to commonjs users using fastify like this (copied from fastify docs)?
> 
> ```js
> const fastify = require('fastify')({
>   logger: true
> })
> ```

I've tried it in a small test app of mine where both syntaxes continue to work for CJS, but would appreciate if somebody else could also give it a go (due to the "mysterious" ways that these things sometimes work 😅 )

pichlermarc:
@mydea thank you for your contribution. Would you mind signing the CLA? Then we can proceed with the PR. :slightly_smiling_face:

mydea:
> @mydea thank you for your contribution. Would you mind signing the CLA? Then we can proceed with the PR. 🙂

Done!

pichlermarc (approved):
> > Is there an impact to commonjs users using fastify like this (copied from fastify docs)?
> > ```js
> > const fastify = require('fastify')({
> >   logger: true
> > })
> > ```
> 
> I've tried it in a small test app of mine where both syntaxes continue to work for CJS, but would appreciate if somebody else could also give it a go (due to the "mysterious" ways that these things sometimes work 😅 )

Looks good, thanks! :slightly_smiling_face: I also checked and it seems to work for me too.

It seems to need some adjustments so that `prettier` succeeds, though (see [here](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6001135065/job/16274673430?pr=1624#step:8:240), you can also run `npm run lint` from the `plugins/node/opentelemetry-instrumentation-fastify/` to check locally), then this should be good to go :slightly_smiling_face:

mydea:
> > > Is there an impact to commonjs users using fastify like this (copied from fastify docs)?
> > > ```js
> > > const fastify = require('fastify')({
> > >   logger: true
> > > })
> > > ```
> > 
> > 
> > I've tried it in a small test app of mine where both syntaxes continue to work for CJS, but would appreciate if somebody else could also give it a go (due to the "mysterious" ways that these things sometimes work 😅 )
> 
> Looks good, thanks! 🙂 I also checked and it seems to work for me too.
> 
> It seems to need some adjustments so that `prettier` succeeds, though (see [here](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6001135065/job/16274673430?pr=1624#step:8:240), you can also run `npm run lint` from the `plugins/node/opentelemetry-instrumentation-fastify/` to check locally), then this should be good to go 🙂

ah, right, I ran `npm run lint:fix` and pushed the fix!

PR #1631, opened 2023-08-11
WIP: chore(deps): update mysql docker tag to v8

blumamir:
This is an attempt to apply the changes in this renovate PR: #1406 along with some necessary changes:

- also update the image tag in test-utils and package.json that reference the old version

- As described in details in [this StackOverflow post](https://stackoverflow.com/a/56509065):

> MySQL 8 has supports pluggable authentication methods. By default, one of them named caching_sha2_password is used rather than our good old mysql_native_password

Our tests uses native passwords, so obviously they couldn't connect. The new authentication method is not supported by `mysql` package, and old widely used versions of `mysql2`. I found that mysql8 can be run with `mysql_native_password` by using the following flag when running docker: `--default-authentication-plugin=mysql_native_password`.

This PR is for applying relevant fixes to the renovate PR to use the new image tag.
The current image 5.7 has no docker image for Apple M2 chip on mac, therefore I am not even able to run current tests on my Mac.


blumamir:
the service in github actions is defined like so:
```
      mysql:
        image: mysql:8.0
        env:
          MYSQL_USER: otel
          MYSQL_PASSWORD: secret
          MYSQL_DATABASE: otel_mysql_database
          MYSQL_ROOT_PASSWORD: rootpw
        ports:
        - 3306:3306
        options: >-
          --health-cmd="mysqladmin ping"
          --health-interval 10s
          --health-timeout 5s
          --health-retries 5
```

And I need a way to pass the `--default-authentication-plugin=mysql_native_password` flag to the container. Looking at [gh docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idservices), I don't see a way to do that. 

Another option is to mount a MySQL config file, as suggested [here](https://hub.docker.com/_/mysql) under "Using a custom MySQL configuration file":
```
-v /my/custom:/etc/mysql/conf.d
```

This will require us to add a config file next to github workflows which isn't great.

PR #1637, opened 2023-08-14
fix(generic-pool): remove deps on types package for ts5 compatibility

### LINKED ISSUE #1632: Cannot find type definition file for 'generic-pool'
### What version of `OpenTelemetry` are you using?
@opentelemetry/auto-instrumentations-node@0.38.0

### What version of `Node` are you using?
20.5.0

### What version of `Typescript` are you using?
5.1.6

### What did you do?
- Installed & setup opentelemetry node-sdk + auto-instrumentation-node  
- build with `npm run tsc`  

Build without error

### What did you see instead?
```
error TS2688: Cannot find type definition file for 'generic-pool'.
  The file is in the program because:
    Entry point for implicit type library 'generic-pool'
```

### Additional context
tsc build is working fine on version 4.8 of typescript but fails on 5.1

The error is likely caused by the `@types/generic-pool` on which depends the `@opentelemetry/instrumentation-generic-pool` package. 
Latest version `@types/generic-pool`  does not exports any typings anymore as the `generic-pool` package provides them directly.

Also, `@opentelemetry/instrumentation-generic-pool`  [imports @types/generic-pool as dependency](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-generic-pool/package.json#L66) and not `devDependency` which looks like a mistake. 

This is probably causing the issue with tsc on consumer projects using  `@opentelemetry/instrumentation-generic-pool` or the auto instrumentation

### Potential Solutions

1) Use typings from generic-pool package directly and remove the dependency to `@types/generic-pool`
2) Move the `@types/generic-pool` to devDependencies

blumamir:

fix #1632 

Replace: #1563 . Since the original PR seems inactive, and the fix is relatively easy, I opened this one.

removed the types from the instrumentation public generic functions, so they will not show up in `instrumentation.d.ts`. Now that the package is not part of the public API, it can safely be moved to dev dependency. This practice is common for instrumentation in the repo, as the type of the instrumentation's `init` function gives very little value (if any) and causes many problems.

After the change, `instrumentation.d.ts` which is imported from `index.d.ts` looks like this:
```js
import { InstrumentationBase, InstrumentationConfig, InstrumentationNodeModuleDefinition } from '@opentelemetry/instrumentation';
export default class Instrumentation extends InstrumentationBase<any> {
    private _isDisabled;
    constructor(config?: InstrumentationConfig);
    init(): InstrumentationNodeModuleDefinition<any>[];
    private _acquirePatcher;
    private _poolWrapper;
    private _acquireWithCallbacksPatcher;
}
//# sourceMappingURL=instrumentation.d.ts.map
```

The `generic-pool` is not imported in types files.
More info [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#types-public-api)


blumamir:
@Flarna I know you have a lot of background in these kinds of issues.
Will appreciate your review on this PR when you have some cycles 🙏

Thank you

pichlermarc (approved):
Looks good, thanks for fixing this :slightly_smiling_face:

PR #1638, opened 2023-08-15
chore(tedious): support new tedious versions 15, 16

blumamir:
replaces: #1521 

When the tedious instrumentations was written, the latest version was 14.
Since then, `tedious` package released versions 15 and 16, for which renovate opened to following PRs: #1403  #1521 .
So it turns out that tedious is already compatible and patches new versions. This PR is only about aligning the supported versions and update the tests to use new tedious versions.

Prior to this PR, the instrumentation was [patching version range](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b3d30afe7376760d211b20c4fb45988ac58c1d85/plugins/node/instrumentation-tedious/src/instrumentation.ts#L77) `*`, which isn't very safe as new major versions can introduce breaking changes which the instrumentation might not be ready for.

This PR replaces the unmergable renovate PR, by introducing the following changes:
1. Cap the maximum supported version to latest (16) instead of `*`. This algins the instrumentation to the practice which is common in this repo and safer.
2. Update the README to advertise the supported versions range accordingly
3. Add the new versions to TAV.
4. Do not run tedious `tav` for node 14, as the new versions dropped support for it which obviously fails to install in node 14.
5. Upgrade the tedious version for `npm run test` to use the most downloaded version which is not 16 (which dropped support for node 14) - `15.1.3`,  and the `"@types/tedious"` to latest - `^4.0.9`

As I am using a mac with apple chip, for which the `mcr.microsoft.com/mssql/server:2017-latest` docker image has no compatible distribution, I am not able to run the unit-tests locally, so uploading it here to test on CI. If you are aware of a better workflow, please let me know.


blumamir:
I wasn't able to make tests green for tedious >=16 and node 14.

Documented the task in #1656 and fixed the repo code to cap support at the last testable version which is 15.

PR #1653, opened 2023-08-28
chore: align @types/node version to core and disable it in renovate

blumamir:
This PR replaces: #1617.

It pins the version of `"@types/node"` across the repo to use a single version - `18.6.5`, the one we are currently using in core repo, and disables renovate from creating PRs to bump the version.

We don't want to depend on newer features of node by accident. See also: https://github.com/open-telemetry/opentelemetry-js/pull/4091


pichlermarc (approved):
Looks good; thanks for taking care of this. :slightly_smiling_face:

PR #1646, opened 2023-08-23
chore(deps): update dependency semver to v7.5.4

AasthaMehtaTech:
## Which problem is this PR solving?
- This PR updates semver version in [opentelemetry-instrumentation-dns](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-dns/package.json#L67) in order to fix https://security.snyk.io/vuln/SNYK-JS-SEMVER-3247795

## Short description of the changes
- After this PR, the version for semver will be upgraded & aligned with other opentelemetry packages in this monorepo.

PR #1654, opened 2023-08-29
fix(instrumentation-dns): use caret range for semver regular dependency

blumamir:
#1646 fixed the version of "semver" in the repo, but I accidentally merged it without noticing that it pinned a regular dependency (not dev) to a specific version which does not follow our conventions as documented [here](https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/GUIDELINES.md#third-party-library-dependencies).

This PR is to fix this. After it is merged, the "semver" versions in the repo should be aligned correctly


pichlermarc (approved):
Thanks! :slightly_smiling_face:

PR #1655, opened 2023-08-29
chore(tedious): align supported versions in code tests and docs

blumamir:
Tedious instrumentation has some inconsistencies with it's supported range of versions.

- It claims to support all `tedious` versions (`*`) which is something we want to avoid, as new major versions might introduce breaking changes for which the instrumentation is not prepared and might potentially harm end users' app stability
- It does not include new versions in `tav` which means the instrumentation will attempt patching for versions for which we never run a test to verify compatibility.
- The supported versions range in docs is not the same as what's in the code.

This PR fixes those 3 issues.

As for August 2023, tedious have a new major version 16, which dropped support for node 14, thus we are unable to run our CI tests properly on it. I attempted to fix this in #1638 , but since mssql does not publish docker version for my mac apple chip, I am not able to run the test locally.

I will document this in an issue and wait for us to drop node14 support or for someone with motivation to pick this up and fix the tests so the CI is green.


pichlermarc (approved):
Thanks :slightly_smiling_face:

PR #1670, opened 2023-09-07
chore(deps): upgrade @types/koa__router dependency

pauloprestes:

- The KOA router type dependency is clashing with the dependencies on the API

- ~Moving type dependencies to devDependencies~
- upgrade @types/koa__router dependency 
- Moving KoaMiddleware from `types` to `internal-types`


Flarna:
The koa types are used in the public API of the instrumenation. Therefore moving it into dev dependency requires consumers to install it.

See https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md for more details.

pauloprestes:
> The koa types are used in the public API of the instrumenation. Therefore moving it into dev dependency requires consumers to install it otherwise typescript build would be broken.
> 
> See https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md for more details.

Thanks for you comment and reviewing this PR. 
Given the `koa-router` in devdependencies are in `12.0.0`, I believe it is safe to upgrade the `@types/koa__router` to the same version too. I've updated this PR to upgrade it.

Could you let me know if this would cause issues?

Flarna (approved):
I think this is fine. 

Unfortunately the koa instrumenation has no maintainer anymore. Let's see what the repo maintainers decide in this regard.

pauloprestes:
Thanks for your comment @blumamir .

I've tried to move `export type KoaMiddleware` to `internal-type.ts`. To do that, I had to move other types as well, to avoid cycle references.
Let me know if the changes that I did looks correct to you.

plugins/node/opentelemetry-instrumentation-koa/src/types.ts
```diff
@@ -13,28 +13,9 @@
...
-export type KoaMiddleware = Middleware<DefaultState, KoaContext> & {
-  router?: Router;
-};
-
-export type KoaRequestInfo = {
-  context: KoaContext;
-  middlewareLayer: Middleware<DefaultState, KoaContext>;
-  layerType: KoaLayerType;
-};
+import { KoaLayerType, KoaRequestInfo } from './internal-types';
```
blumamir, line 18: `types.ts` cannot import from `internal-types.ts`, as this would include all private types in this file into the public api for the package. The other way is ok.

the `types.ts` should include everything public, which is the `KoaInstrumentationConfig` and all the tansitive types it references.

So `KoaLayerType`, `KoaRequestInfo` and `KoaMiddleware` are all part of the public API thus they must remain in `types.ts`. `KoaContext` is not used for anything public, thus it should be moved to `internal-types.ts`.

blumamir (commented):
Thanks @pauloprestes :)

You are almost there, the PR still needs a little separation between public and private types.[here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md) you can find guidelines on what should go into `types.ts` and what should go into `internal-types.ts`

blumamir:
> To do that, I had to move other types as well, to avoid cycle references.

Can you elaborate on what cycle references you encountered?
The rule is that `types` CANNOT import `internal-types` but `internal-types` CAN import `types`.

pauloprestes:
> > To do that, I had to move other types as well, to avoid cycle references.
> 
> Can you elaborate on what cycle references you encountered? The rule is that `types` CANNOT import `internal-types` but `internal-types` CAN import `types`.

The issue is:
- `KoaContext` is used in the `types.ts` for `KoaRequestInfo`.
- `KoaPatchedMiddleware` declared in the `internal-types.ts` is equal to `KoaMiddleware` (declared in types.ts).

So, if we move `KoaContext` to `internal-types`, we would need to import it into `types` for `KoaRequestInfo`, that would create a cyrcle reference.

However, reading the guidelines and after knowing the rule you shared, I don't think we can move `KoaContext` to `internal-types`, as this would break the rule.

Hope it all make sense for you.

plugins/node/opentelemetry-instrumentation-koa/src/types.ts
pauloprestes: `KoaRequestInfo` uses `KoaContext`, so if I understood correct, it makes `KoaContext` used in the public API, so it can't be moved to the `internal-types`.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-koa/src/types.ts#L34

pauloprestes:
Given the comments above, I just reverted all my last changes.

pauloprestes:
given the comments above, I've moved only `KoaMiddleware` to `internal-types`. As it seems it is the only type that doesn't need to be part of the public API.

plugins/node/opentelemetry-instrumentation-koa/src/types.ts
blumamir: You are right!
I meant to write before that only `KoaMiddleware` should be moved to `internal-types.ts` and not `KoaContext`, which is what you implemented in the latest version. Thanks again 🥇

blumamir (approved):
LGTM

Thanks again

Seems CI is broken by other package, so I will rebase and merge once CI is fixed.

pauloprestes:
> LGTM
> 
> Thanks again
> 
> Seems CI is broken by other package, so I will rebase and merge once CI is fixed.

Yeah, I notice CI was broken by other changes.
Thanks again for reviewer the PR and help with directions.

Issue #1275, opened 2022-11-08
FastifyDeprecation: Request#context property access is deprecated. Please use "Request#routeConfig" or "Request#routeSchema" instead for accessing Route settings. The "Request#context" will be removed in `fastify@5`.


stefee:
### What version of OpenTelemetry are you using?

`@opentelemetry/instrumentation-fastify@0.30.1`

### What version of Node are you using?

`v16.14.2`

### What did you do?

Node auto instrumentation on a Fastify application

### What did you expect to see?

No deprecation notices

### What did you see instead?

```
FastifyDeprecation: Request#context property access is deprecated. Please use "Request#routeConfig" or "Request#routeSchema" instead for accessing Route settings. The "Request#context" will be removed in `fastify@5`.
```

### Additional context

Request `.context` is used here:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/aff84bba1d391ec2061b8d0121ac8dd36fc1980c/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts#L262

stefee:
Bump

Bruno-DaSilva:
bump :(

Uzlopak:
To fix the issue in opentelemetry, you would need to patch 

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/154b30bc98ff0b848348faa18197eb7f11dcbdc4/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts#L263

There I think you would need to change it to:
```js
      const handlerName = (((request as any).routeConfig || {})?.name || '').substr(6);
```

Uzlopak:
Or maybe even simpler:

```js
      const handlerName = (request as any)?.routeConfig?.name?.substr(6) || '';
```

I did not test it.

freakyfelt:
Alas that one won't work as the routeConfig does not expose the handler, so we're waiting for https://github.com/fastify/fastify/pull/4470

Uzlopak:
I closed the remaining issues in the fastify PR and requested review by @eomm

drewcorlin1:
Wouldn't this be the most type safe way to fix the warning?

```typescript
const handlerName = `${request.routeOptions.method} ${request.routeOptions.url}`;
```

freakyfelt:
> Wouldn't this be the most type safe way to fix the warning?

[...]

PR #1679, opened 2023-09-12
fix: Removed deprecated properties usage in Fastify instrumentation

seidelmartin:

Instrumentation has been using deprecated properties producing errors like `FastifyDeprecation: Request#context property access is deprecated. Please use "Request#routeConfig" or "Request#routeSchema" instead for accessing Route settings. The "Request#context" will be removed in fastify@5.`
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1275

I've added getting deprecated properties from newly introduced `routeOptions` object in fastify request. It supports also old versions that doesn't have this object exposed. It has been introduced in fastify https://github.com/fastify/fastify/releases/tag/v4.23.0

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-fastify/package.json
+++ b/plugins/node/opentelemetry-instrumentation-fastify/package.json
@@ -53,13 +53,13 @@
     "@opentelemetry/sdk-trace-node": "^1.8.0",
     "@types/express": "4.17.18",
     "@types/mocha": "7.0.2",
-    "@types/node": "18.6.5",
-    "fastify": "4.18.0",
+    "@types/node": "18.15.3",
+    "fastify": "4.23.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
     "rimraf": "5.0.5",
     "ts-mocha": "10.0.0",
-    "typescript": "4.4.4"
+    "typescript": "5.2.2"
   },
   "dependencies": {
     "@opentelemetry/core": "^1.8.0",
--- a/plugins/node/opentelemetry-instrumentation-fastify/package.json
+++ b/plugins/node/opentelemetry-instrumentation-fastify/package.json
@@ -59,7 +59,7 @@
     "nyc": "15.1.0",
     "rimraf": "5.0.5",
     "ts-mocha": "10.0.0",
-    "typescript": "5.2.2"
+    "typescript": "4.7.4"
   },
   "dependencies": {
     "@opentelemetry/core": "^1.8.0",
--- a/plugins/node/opentelemetry-instrumentation-fastify/package.json
+++ b/plugins/node/opentelemetry-instrumentation-fastify/package.json
@@ -54,12 +54,12 @@
     "@types/express": "4.17.18",
     "@types/mocha": "7.0.2",
     "@types/node": "18.15.3",
-    "fastify": "4.23.0",
+    "fastify": "4.18.0",
     "mocha": "7.2.0",
     "nyc": "15.1.0",
     "rimraf": "5.0.5",
     "ts-mocha": "10.0.0",
-    "typescript": "4.7.4"
+    "typescript": "4.4.4"
   },
   "dependencies": {
     "@opentelemetry/core": "^1.8.0",
--- a/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
@@ -93,10 +93,11 @@ export class FastifyInstrumentation extends InstrumentationBase {
       }
       instrumentation._wrap(reply, 'send', instrumentation._patchSend());
 
-      const anyRequest = (request as any);
+      const anyRequest = request as any;
 
       const rpcMetadata = getRPCMetadata(context.active());
-      const routeName = anyRequest.routeOptions?.config?.url || request.routerPath;
+      const routeName =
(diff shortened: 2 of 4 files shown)


seidelmartin:
Hi @pichlermarc. I'm confused by the GH workflow failure. Is it something that fails globally or just in my PR? I've tried `main` version locally and `lerna boostrap` fails for me as well...

pichlermarc:
@seidelmartin I'm looking into it, I think it may be both. Trying it on this branch makes it fail on the fastify intrumentation's `prepare` script, but trying it on `main` makes it fail for the AWS SDK instrumentation for me. :thinking: I'm not sure yet what the cause for it is, yet; I'll keep you posted.

pichlermarc:
@seidelmartin looks like I had an old revision checked out and `main` is actually compiling fine when re-installing dependencies. I think the problem here could be that the types from `@types/mongodb` hoisted to the project root no longer compile with typescript 5 (we align typescript for this reason). Would you mind trying with the lowest possible typescript version that still compiles with fastify? :thinking: Otherwise we'll have to add `--nohoist='@types/mongodb'` when running `lerna bootstrap`.

seidelmartin:
@pichlermarc I'll try that. 👍🏻

seidelmartin:
@pichlermarc good news, after downgrading typescript to 4.7.4 all checks pass.

seidelmartin:
Alright, all 🟢.

seidelmartin:
Hi @pichlermarc, what will be the next step here to publish it as new version?

pichlermarc (commented):
Looks good, thanks @seidelmartin :slightly_smiling_face: 
Once this is merged, a release PR is created automatically. 

@blumamir do we have any specific limitations on bumping typescript for a single package in this repo? :thinking: Since this package is not 1.0 yet, we can make a breaking change like bumping it for `@opentelemetry/instrumentation-fastify` (it's needed for new versions of the package anyway). Thinking about it closer though, it'll likely affect `@opentelemetry/auto-instrumentations-node` too by possibly requiring a later version for consumers.

blumamir:
> Looks good, thanks @seidelmartin 🙂 Once this is merged, a release PR is created automatically.
> 
> @blumamir do we have any specific limitations on bumping typescript for a single package in this repo? 🤔 Since this package is not 1.0 yet, we can make a breaking change like bumping it for `@opentelemetry/instrumentation-fastify` (it's needed for new versions of the package anyway). Thinking about it closer though, it'll likely affect `@opentelemetry/auto-instrumentations-node` too by possibly requiring a later version for consumers.

To my latest understanding, if we use a newer typescript, we can break existing users with incompatible versions. But @Flarna will probably know best

Flarna:
Newer typescript versions may generated .d.ts files which are incompatible with an older typescript version.
I don't have in mind which ts version includes such problematic changes but it's not bound to major versions only. I guess a search in the issue tracker of [core repo](https://github.com/open-telemetry/opentelemetry-js) has more details.

While typescript brings a lot of benefits it adds also some compatibility burden...

An update of typescript will for sure happen with SDK 2.0.

blumamir:
Why do we need to bump the typescript version in the first place? is it so we can use `"fastify": "4.23.0"` instead of `4.18.0` to verify the tests assert this fix?

Perhaps we can keep all current dependencies as they are and use `test-all-versions` script to cover all versions instrumentation supports?

seidelmartin:
> Why do we need to bump the typescript version in the first place? is it so we can use `"fastify": "4.23.0"` instead of `4.18.0` to verify the tests assert this fix?
> 
> Perhaps we can keep all current dependencies as they are and use `test-all-versions` script to cover all versions instrumentation supports?

I've updated fastify to v4.23.0 because it exposes new properties that instrumentation uses. Otherwise it still produces a lot of deprecation warnings. I can still downgrade it but then I wouldn't be able to test my changes.

Fernandomr88:
Guys, this is generating tons of errors in production, what's the ETA of this PR?

pichlermarc:
Thanks to everyone for the feedback on the PR here. 

I've created a PR to add a `test-all-versions` script that will allow us to keep `fastify@4.18.0`/`typescript@4.4.4`, while at the same time also allowing us to test against `fastify@4.23.2`/`typescript@4.7.4` (#1710). This way we won't run into compatibility issues with users of `typescript@4.4.4`.

@seidelmartin would you mind changing the dependencies back to `fastify@4.18.0`/`typescript@4.4.4`? Once #1710 is merged we'll then automatically test against  `fastify@4.23.2` when the `pkg:instrumentation-fastify` label is applied - and once that passes we can merge this PR. :slightly_smiling_face:

seidelmartin:
Yes, I'll do that.

markrzen:
@seidelmartin You may want to rebase to main as it appears you are out-of-date.

seidelmartin:
@markrzen I've rebased to latest version. Can we proceed with it? What will be next steps?

markrzen:
@seidelmartin I don't have access to approve, I am just another contributor like yourself.

@pichlermarc Anything else to get this PR approved?

markrzen:
@open-telemetry/javascript-approvers Anyone able to review this PR? It's a fairly simple update, impactful and getting long in the tooth.

Fernandomr88:
awesome! When will this be available @ Npm?

pichlermarc:
@Fernandomr88 we released it earlier today: https://github.com/open-telemetry/opentelemetry-js-contrib/releases/tag/instrumentation-fastify-v0.32.3

PR #1680, opened 2023-09-12
fix(fastify): Use plugin name for middleware span name

mydea:
I noticed that I was getting a bunch `middleware - anonymous` spans from my fastify instrumentation. 

I think it would be more useful to see e.g. `middleware - router` or `middleware - fastify -> @fastify/multipart -> @fastify/helmet` there?


blumamir (approved):
LGTM. Thank you for the contribution

Issue #1677, opened 2023-09-12
"container.id" resource attribute is wrong


liurui-1:
### What version of OpenTelemetry are you using?
```
"telemetry.sdk.language": "nodejs",
"telemetry.sdk.version": "1.10.1",
```

```
  "dependencies": {
    "@grpc/grpc-js": "1.6.7",
    "@opentelemetry/api": "1.4.1",
    "@opentelemetry/auto-instrumentations-node": "0.36.4",
    "@opentelemetry/auto-instrumentations-web": "0.31.2",
    "@opentelemetry/context-zone": "1.10.1",
    "@opentelemetry/core": "1.10.1",
    "@opentelemetry/exporter-metrics-otlp-grpc": "0.36.1",
    "@opentelemetry/exporter-trace-otlp-grpc": "0.36.1",
    "@opentelemetry/exporter-trace-otlp-http": "0.36.1",
    "@opentelemetry/instrumentation": "0.36.1",
    "@opentelemetry/resources": "1.10.1",
    "@opentelemetry/sdk-metrics": "1.10.1",
    "@opentelemetry/sdk-node": "0.36.1",
    "@opentelemetry/sdk-trace-base": "1.10.1",
    "@opentelemetry/sdk-trace-node": "1.10.1",
    "@opentelemetry/sdk-trace-web": "1.10.1",
    "@opentelemetry/resource-detector-alibaba-cloud": "0.27.4",
    "@opentelemetry/resource-detector-aws": "1.2.2",
    "@opentelemetry/resource-detector-container": "0.2.2",
    "@opentelemetry/resource-detector-gcp": "0.28.0",
    ...
```

### What version of Node are you using?
v18.16.0

### What did you do?
We are using OpenTelemetry official demo: 
https://github.com/open-telemetry/opentelemetry-demo/blob/main/kubernetes/opentelemetry-demo.yaml

We find the container.id resource attribute of `my-otel-demo-frontend` deployment (written by Node.JS) is wrong.
The expected value is: `00601ecc8806271163e6f8e8456d1614817bc5e6d7fa19025fea6fb0b373e7bb`
The returned value is: `cc8806271163e6f8e8456d1614817bc5e6d7fa19025fea6fb0b373e7bb.scope`
Note: you can see the wrong value lose the first 6 characters and has a wrong suffix of "`.scope`"

The problem is similar to the issue reported by https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1408

[...]

PR #1682, opened 2023-09-14
fix: Update defect of of wrong resource attribute of "container.id"

liurui-1:

As described in "https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1677", the way to get resource attribute of "container.id" is wrong.

I made the changes following the code of OTel Java SDK which works well.
Here is the code I followed:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/ContainerResource.java#L54-L60

I have tested the code working well in my Kubernetes environments.


pichlermarc:
Thanks for fixing this! :slightly_smiling_face:
Looks like the tests still need some attention as their current test data is not very realistic, so the new logic drops them. 

Feel free to take the test-data from https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/0f1199620db05a3739cd3bae7e0c408a9bc0d65b/instrumentation/resources/library/src/test/java/io/opentelemetry/instrumentation/resources/CgroupV1ContainerIdExtractorTest.java#L82-L113 and adapt the tests here to use those. :slightly_smiling_face:

liurui-1:
Made some slight change. Please review again.

liurui-1:
Hello, I updated the UT code. Please review again.

liurui-1:
Hello, I updated the UT code. Please review again.

liurui-1:
@pichlermarc, I updated code according to the lint message. It seems the lint result does not show up automatically. Please help review.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -57,9 +57,29 @@ export class ContainerDetector implements Detector {
...
     );
     const splitData = rawData.trim().split('\n');
-    for (const str of splitData) {
-      if (str.length >= this.CONTAINER_ID_LENGTH) {
-        return str.substring(str.length - this.CONTAINER_ID_LENGTH);
+    for (const line of splitData) {
+      const lastSlashIdx = line.lastIndexOf('/');
+      if (lastSlashIdx === -1){
+        continue;
+      }
```
pichlermarc, line 64: looks like there was just some missing whitespace ` `
```suggestion
      if (lastSlashIdx === -1) {
        continue;
      }
```

Issue #1693, opened 2023-09-21
pg: cannot instrument an ESM-imported pg


mjsalinger:
### What version of OpenTelemetry are you using?

honeycomb/opentelemetry-node 0.5.0
opentelemetry/auto-instrumentations-node 0.39.2
opentelemetry/api 1.4.1

### What version of Node are you using?
16.20.2

### What did you do?
Started telemetry with auto-instrumentation enabled
Started node with

```
node --experimental-loader=@opentelemetry/instrumentation/hook.mjs ./src/index.js",
```

### What did you expect to see?
Node start and instrumentation produced for pg module

### What did you see instead?

```
/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/@opentelemetry/instrumentation-pg/build/src/instrumentation.js:33
            if ((0, instrumentation_1.isWrapped)(moduleExports.Client.prototype.query)) {
                                                                      ^

TypeError: Cannot read properties of undefined (reading 'prototype')
    at InstrumentationNodeModuleDefinition.patch (/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/@opentelemetry/instrumentation-pg/build/src/instrumentation.js:33:71)
    at PgInstrumentation._onRequire (/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:153:39)
    at hookFn (/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:194:29)
    at callHookFn (/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/import-in-the-middle/index.js:28:22)
    at Hook._iitmHook (/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/import-in-the-middle/index.js:76:11)
    at /Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/import-in-the-middle/lib/register.js:29:31
    at Array.forEach (<anonymous>)
    at register (/Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/import-in-the-middle/lib/register.js:29:15)
    at file:///Users/mjsalinger/Development/platform/api/backend-api-service/node_modules/pg/lib/index.js?iitm=true:13:1
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
```

### Additional context
[...]

PR #1701, opened 2023-09-27
fix(pg): fix instrumentation of ESM-imported pg

trentm:
The ESM imported top-level object is a Module Namespace Object per ECMA262 spec that is, apparently, incompatible with the 'PG' class object that 'pg' returns. The "<module>.defaults" holds the equivalent of `require('pg')`.

Fixes: #1693 

---

This is a similar fix to #1694. See the discussion there on the use of `module[Symbol.toStringTag] === 'Module'` to sniff out ESM.

A dump of `moduleExports` *before* this change, when loading `pg` as ESM is as follows. This shows how the `PG` instance export is replaced with the `[Module: null prototype]` and how `.default` holds the top-level CommonJS exports.

```
[Module: null prototype] {
  default: PG {
    defaults: {
      host: 'localhost',
      user: 'trentm',
      database: undefined,
      password: null,
      connectionString: undefined,
      port: 5432,
      rows: 0,
      binary: false,
      max: 10,
      idleTimeoutMillis: 30000,
      client_encoding: '',
      ssl: false,
      application_name: undefined,
      fallback_application_name: undefined,
      options: undefined,
      parseInputDatesAsUTC: false,
      statement_timeout: false,
      idle_in_transaction_session_timeout: false,
      query_timeout: false,
      connect_timeout: 0,
      keepalives: 1,
      keepalives_idle: 0,
      parseInt8: [Setter]
    },
    Client: [class Client extends EventEmitter] {
      Query: [class Query extends EventEmitter]
    },
    Query: [class Query extends EventEmitter],
    Pool: [class BoundPool extends Pool],
    _pools: [],
    Connection: [class Connection extends EventEmitter],
    types: {
      getTypeParser: [Function: getTypeParser],
      setTypeParser: [Function: setTypeParser],
      arrayParser: [Object],
      builtins: [Object]
    },
    DatabaseError: [class DatabaseError extends Error]
  }
}
```

# Checklist

- [ ] Decide whether and how to add a test case for this. I could add a separate `npm run test:esm`. The downside of that is that the `npm run test:local` version of this will stop and start a Postgres instance once for CommonJS tests (`*.test.js`) and then once again for ESM tests (`*.test.mjs`).
- [x] Deal with TypeScript-related lint errors.


trentm:
> * [ ]  Deal with TypeScript-related lint errors.

Oh, lint passes. There are a number of *warnings* from `npm run lint`, but I guess those are okay. Most of those warnings existed before my change.

pichlermarc:
>  Decide whether and how to add a test case for this. I could add a separate npm run test:esm. The downside of that is that the npm run test:local version of this will stop and start a Postgres instance once for CommonJS tests (*.test.js) and then once again for ESM tests (*.test.mjs).

I would consider this an acceptable trade-off for testing this. :+1: 
We could also add more scripts for local testing: `test:local:cjs`, `test:local:esm` that are both run by `test:local` -  this way one can test both individually when working on the package. I think the important part would be to have the ESM tests run in the unit-testing and TAV workflows.

pichlermarc (approved):
Thanks for taking care of this! :slightly_smiling_face:

PR #1706, opened 2023-10-01
feat: add support for graphql-tools-executor

YutaUra:

- Some GraphQL server libraries, such as graphql-yoga, use @graphql-tools/executor instead of graphql.js built-in executor.

- Instrument @graphql-tools/executor

I have used the existing graphql.js excutor implementation as a reference, but if there is anything missing or incomplete in the testing or source code, I would be glad to hear about it.


JamieDanielson:
@YutaUra Sorry we have not had a chance to review this and it has been a while. Is this something you are still interested in, and would you be willing to be a maintainer of the instrumentation? This will be important to know before we can give this a thorough review. Thanks!

phal0r:
Facing the same problem, it makes sense to have this instrumented aswell. Since these instrumentations are rather low-level it took hours to figure out, why the instrumentation did not work as expected.

JamieDanielson:
> @YutaUra Sorry we have not had a chance to review this and it has been a while. Is this something you are still interested in, and would you be willing to be a maintainer of the instrumentation? This will be important to know before we can give this a thorough review. Thanks!

If it helps to understand why I ask, here is a [link to some new guidance](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#new-instrumentation) we've written about new instrumentations getting added to this repo.

PR #1709, opened 2023-10-02
fix: reconfigure stalebot to not close bugs

pichlermarc:

Stalebot seems to apply `stale` labels to `bug` issues and closing them. This PR addresses that by adding `bug` to the `exempt-issue-labels`.


JamieDanielson (approved):

JamieDanielson:
@pichlermarc I guess one idea to try (based on convo just had in SIG) - instead of having "exempt" labels to exempt bugs and instrumentation requests, we can have "any-of" labels, which is something we do on Honeycomb repos. This way we don't have to specifically remove a bug label if something is perhaps a misunderstanding but maybe a bug and needing info from the author.
So for example, our stalebot looks at `any-of-labels: 'status: info needed,status: revision needed'`

pichlermarc:
> @pichlermarc I guess one idea to try (based on convo just had in SIG) - instead of having "exempt" labels to exempt bugs and instrumentation requests, we can have "any-of" labels, which is something we do on Honeycomb repos. This way we don't have to specifically remove a bug label if something is perhaps a misunderstanding but maybe a bug and needing info from the author. So for example, our stalebot looks at `any-of-labels: 'status: info needed,status: revision needed'`

Yep I think that's a good idea. :slightly_smiling_face:  I've been meaning to introduce a few more that may lead to an issue being automatically closed for some time now as part of a new issue triage worflow. I think we can start with `information-requested` and gradually expand that as we add more labels. 

Would you like to open a PR @JamieDanielson? Otherwise, I'll get around to it tomorrow. :slightly_smiling_face:

Issue #1477, opened 2023-04-25
Tests failing for `@aws-sdk/client-sqs >=3.316` 


pichlermarc:
### What version of OpenTelemetry are you using?

Current revision on `main`

### What version of Node are you using?

Node 16, 18

### What did you do?

`npm run test-all-versions` in `@opentelemetry/instrumentation-aws-sdk`

### What did you expect to see?

Tests pass.

### What did you see instead?

Tests fail for `@aws-sdk/client-sqs` `>=3.316`

### Additional context

Run for Node 16: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/4787435108/jobs/8512667684
Run for Node 18: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/4787435108/jobs/8512667928

pichlermarc:
cc @carolabadeer

carolabadeer:
Thanks for raising this @pichlermarc! I will take a look

dyladan:
labeling p4 as this does not affect users

pichlermarc:
@carolabadeer any update?

carolabadeer:
Currently working with the AWS SDK for JS team to identify possible issues in the changes between @aws-sdk/client-sqs v3.315 and v3.316 that could be causing this test failure. Will post updates here on any findings and progress

carolabadeer:
While waiting I also looked into the behavior in the unit tests and found that there are two spans being used (not sure if they are meant to be the same) - a [test span](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/418b6f6d471b4b38507e8c6ef2916bee31ed2001/plugins/node/opentelemetry-instrumentation-aws-sdk/test/aws-sdk-v3.test.ts#L402) and a [callback span](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/418b6f6d471b4b38507e8c6ef2916bee31ed2001/plugins/node/opentelemetry-instrumentation-aws-sdk/test/aws-sdk-v3.test.ts#L416). 

The callback span is the one whose attributes are being checked in the unit test that is failing. In that span, the test is checking for a `SemanticAttributes.MESSAGING_OPERATION` attribute but cannot find one. This span seems to be wrapping an S3 `CreateBucket` call. These are it's attributes: 
```
{
  'aws.operation': 'createBucket',
  'aws.signature.version': 's3',
  'aws.service.api': 'S3',
[...]

PR #1711, opened 2023-10-02
fix(instrumentation-aws-sdk): temporarily disable failing sqs test 

pichlermarc:

See #1477. This test constantly failing reduces the visibility of failing tests in other packages. Skipping this test for now until #1477 is resolved. I'll add a comment in #1477 to remove the `skip()`.


pichlermarc:
Looks like actually all 3 tests in `custom service behavior` failing.

trentm (approved):

david-luna (approved):

pichlermarc:
Hmm, looks like now the tests are failing with `Error: Cannot find module '@smithy/hash-stream-node'` - I wonder if this is as similar issue to the one we're seeing in #1828

trentm:
> Hmm, looks like now the tests are failing with `Error: Cannot find module '@smithy/hash-stream-node'` - I wonder if this is a similar issue to the one we're seeing in #1828

Explained here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1828#issuecomment-1834276719
tl;dr: `@aws-sdk/client-s3@3.377.0` was just a bad release.  #1828 will resolve it, because I've reduced the aws-sdk/client-s3 versions tested in the .tav.yml file, and explicitly skipped that broken version.

trentm:
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1847 is an alternative change that reduces the number of tests skipped.

PR #1712, opened 2023-10-03
feat(fastify): Add `onError` configuration

mydea:
This makes it very easy to provide an error handler for the instrumentation. This is helpful for libraries because users may add their own error handlers, which may swallow errors. For example, at Sentry, we want to have an `onError` handler in order to ensure we capture all errors that happen, even if users add their own `onError` to e.g. render a custom 400 or 500 response.

Usage:

```js
new FastifyInstrumentation({
  onError: (request, reply, error) => {
    // do something with the error
  }
})
```

For e.g. libraries wrapping this under the hood, this is extremely helpful as it allows to provide an error handler without the user needing to do anything.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
@@ -102,6 +102,22 @@ export class FastifyInstrumentation extends InstrumentationBase {
     };
   }
 
+  private _hookOnError() {
+    const instrumentation = this;
+
+    return (request: any, reply: any, error: Error, done: () => void) => {
+      if (!instrumentation.isEnabled()) {
+        return done();
+      }
+
+      const onError = instrumentation.getConfig().onError;
+      if (onError) {
+        onError(request, reply, error);
+      }
+      done();
+    };
+  }
+
   private _wrapHandler(
     pluginName: string,
     hookName: string,
@@ -206,6 +222,7 @@ export class FastifyInstrumentation extends InstrumentationBase {
       const app: FastifyInstance = moduleExports.fastify.apply(this, args);
       app.addHook('onRequest', instrumentation._hookOnRequest());
       app.addHook('preHandler', instrumentation._hookPreHandler());
+      app.addHook('onError', instrumentation._hookOnError());
 
       instrumentation._wrap(app, 'addHook', instrumentation._wrapAddHook());
 
--- a/plugins/node/opentelemetry-instrumentation-fastify/src/types.ts
+++ b/plugins/node/opentelemetry-instrumentation-fastify/src/types.ts
@@ -30,10 +30,16 @@ export interface FastifyCustomAttributeFunction {
   (span: Span, info: FastifyRequestInfo): void;
 }
 
+export interface FastifyOnErrorFunction {
+  (request: any, reply: any, error: Error): void;
+}
+
 /**
  * Options available for the Fastify Instrumentation
  */
 export interface FastifyInstrumentationConfig extends InstrumentationConfig {
   /** Function for adding custom attributes to each handler span */
   requestHook?: FastifyCustomAttributeFunction;
+  /** Function to call when an error is captured by Fastify. */
+  onError?: FastifyOnErrorFunction;
 }
--- a/plugins/node/opentelemetry-instrumentation-fastify/test/instrumentation.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-fastify/test/instrumentation.test.ts
@@ -498,5 +498,30 @@ describe('fastify', () => {
         });
       });
     });
+
+    describe('using onError in config', () => {
+      it('calls onError provided function when set in config', async () => {
+        const errors: Error[] = [];
+
+        const onError = (request: unknown, res: unknown, error: Error) => {
+          console.log('on error');
+          errors.push(error);
+        };
+
+        instrumentation.setConfig({
+          ...instrumentation.getConfig(),
+          onError,
+        });
+
+        app.get('/test', (req, res) => {
+          throw new Error('test error here');
+        });
+
+        await startServer();
+        await httpRequest.get(`http://localhost:${PORT}/test`);
+
+        assert.deepEqual(errors, [new Error('test error here')]);
+      });
+    });
   });
 });


plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
```diff
@@ -102,6 +102,22 @@ export class FastifyInstrumentation extends InstrumentationBase {
...
+        return done();
+      }
+
+      const onError = instrumentation.getConfig().onError;
+      if (onError) {
+        onError(request, reply, error);
+      }
+      done();
+    };
+  }
```
pichlermarc, line 119: Does not re-throwing the error here change fastify behvaior? ("swallowing"  error for other handlers?)

pichlermarc (commented):
I think I may not fully understad the reasoning behind this change - it allows users to add a hook to their fastify instance and does not really add instrumentation to it. Wouldn't it be possible for them to add their own error hook and record telemetry there?

The request hook is there for users that want to add additional attributes - what is the purpose of the error handler in this case? If the goal is to add the error information to the span, maybe we could add that functionality behind a flag and save the user some steps.

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
mydea: good point, I will double check how this is supposed to behave/work!

mydea:
> I think I may not fully understad the reasoning behind this change - it allows users to add a hook to their fastify instance and does not really add instrumentation to it. Wouldn't it be possible for them to add their own error hook and record telemetry there?
> 
> The request hook is there for users that want to add additional attributes - what is the purpose of the error handler in this case? If the goal is to add the error information to the span, maybe we could add that functionality behind a flag and save the user some steps.

So the idea of this is that this can be used by libraries wrapping this, not necessarily by "end users". 

At Sentry, we would love to use this instrumentation to automatically capture performance data through OpenTelemetry, but also attach some error handlers for capturing errors etc. Without this change, this requires us to double wrap fastify (once through OTEL, and once ourselves), which is a bit tricky. Having a hook like this would drastically streamline this for end users, who would not have to configure anything manually.

I hope that explains the reasoning better!

pichlermarc:
Thank you for the response @mydea! I can see that this can be very useful for vendors that want to redistribute the packages with their spin on them. 

However, it goes against my understanding of what OpenTelemetry-hosted instrumentation packages should include. In my opinion, we've been developing these packages with the expressed purpose of providing vendor-neutral instrumentations. Maybe I'm still misunderstanding, but adding a hook that only aims to support a single vendor's use case seems to go against vendor neutrality. From my experience, similar existing hooks in this repository primarily aim to benefit the end-user. An example would be helping them add missing attributes that the default instrumentation can't include as they may be specific to the user's app. Another example would be ignoring metrics attributes that increase cardinality and, therefore, incur memory, ingest, or storage costs. 

But here, the end-user could very well register their hook and add manual instrumentation, so the value added looks minimal. What I am missing is the benefit other OpenTelemetry users would have from such an option. Under an increase in value here, I'd understand if the registered default hook writes some information to the span, writes some metrics, or writes some logs - but it being empty seems to like a hard-to-justify increase in API surface and complexity compared to the value it brings. 🤔

Is there maybe some change that we could make to the instrumentation that would make it easier to double-wrap and achieve the objectives? Or is there some way to add the functionality you're trying to add in your library here in this instrumentation (in pure OpenTelemetry) instead?

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
mydea: So i verified that just calling `done()` will bubble the error up to the next hook:

```js
fastify.addHook('onError', (request, reply, error, done) => {
  console.log('hook 1');
  done();
});

fastify.addHook('onError', (request, reply, error, done) => {
  console.log('hook 2');
  done();
});

app.setErrorHandler(function (error, request, reply) {
 console.log('error handler');
  reply.status(409).send({ ok: false })
});
```

When an error happens, results in:

* `hook 1`
* `hook 2`
* `error handler`

I also verified that as soon as a user adds an error handler via `setErrorHandler()`, errors are not bubbled up anymore to global error handlers.

mydea:
Thank you for your thought out response, I totally understand your points. I understand that we don't want to add anything that is vendor-specific, I am trying to find a way to make this work as vendor neutral as possible!

The challenging part is that there is no (fully realized, at least) concept for error capturing on Open telemetry (as far as I've seen). the closest I could find was to add a timed event to a span with exception data & stack. However, this is not enough here as since we cannot attach the actual error itself to that (because we can only attach primitives to a timed event), and because of that it would become hard/impossible to make sure we do not capture an error twice (think a user _does not_ define an error handler, so the error _also_ bubbles up to a global handler and is reported again - without access to the error instance it will be very unreliable to identify this error was already captured somewhere else).

So I think without a change to expose either an error hook, or some other similar thing, it will not be possible to streamline this for users of wrapping SDKs. (note that this is not Sentry specific but would be the same for any vendor that wants to capture/handle errors somehow in addition to performance data).

I see three possible ways to make this happen with minimal user requirements:

1. The current implementation of this PR, expose an `onError` or similar hook that is called on errors.
2. Instead add a different hook that is called after instrumentation happened, e.g. `afterInstrument((fastify) => {})`, which would receive the fastify instance and could be used to add stuff generically to it.
3. Double wrap it, which may work but is also subject to break because e.g. any check in the instrumentation if the method is already wrapped (think `isWrapped(xxx)`) would "incorrectly" return true for the other wrapping, and may be removed/unwrapped by instrumentation. This would kind of make this behave on internal semantics (e.g. if the instrumentation is refactored to check for prior wrapping and unwrap, like e.g. the express instrumentation already does today).

All of these have up and downsides and I'll defer to you what you think makes most sense. If you do not want to add any error-related things at all, I also accept this of course, then our users will simply have to manually add some error handler functionality - nothing we can do about this then! Or if there is any other way how errors could be captured in a more generic way, I'd also be very happy to hear other ideas 👍

Flarna:
It's interesting that the proposed error hook seems to be not needed/used to actually set the span status.
So it seems this may end up in having some captured errors but none of them is visible on the span and span status is ok.
Or we have a span telling error but this error was maybe not reported via the error callback.

My feeling here is that the usecase for capturing errors independent of spans/logs/metrics is something not covered at all by OpenTelemetry as of now. Is this maybe a topic for OTel spec to have a generic error reporting signal?

mydea:
> It's interesting that the proposed error hook seems to be not needed/used to actually set the span status. So it seems this may end up in having some captured errors but none of them is visible on the span and span status is ok. Or we have a span telling error but this error was maybe not reported via the error callback.

FWIW we could maybe extend this to _also_ set the span status to be not OK (but that would only touch the ability to _also_ capture errors tangientally). But _if_ we decide to go through with this that would make sense for sure! 

> My feeling here is that the usecase for capturing errors independent of spans/logs/metrics is something not covered at all by OpenTelemetry as of now. Is this maybe a topic for OTel spec to have a generic error reporting signal?

Yeah, I guess that may make sense (but is obv. a long way off for actual standardisation to happen there 😢 )

pichlermarc:
>> My feeling here is that the usecase for capturing errors independent of spans/logs/metrics is something not covered at all by OpenTelemetry as of now. Is this maybe a topic for OTel spec to have a generic error reporting signal?
>
>Yeah, I guess that may make sense (but is obv. a long way off for actual standardisation to happen there 😢 )

@mydea I have to admit im not very knowledgable about error reporting works in different backends so what I'm saying might make little sense, but wonder if it would make sense to emit errors in a similar fashion to what's described here: https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/ 

In that case a specification for it may not be as far out as one would think (no need to specifiy a full new signal+api+sdk, *just* some semantic conventions, which arguably can take some time to stabilize but I'd definetly be vendor neutral then). Right now we actually do even provide an experimental logging API, SDK and exporters in the core repo already that can be used to write log records like this.

mydea:
> @mydea I have to admit im not very knowledgable about error reporting works in different backends so what I'm saying might make little sense, but wonder if it would make sense to emit errors in a similar fashion to what's described here: https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/

Yeah, the problem with this is mainly that since these records only contain the message (and other primitive) fields, it is very hard to properly ensure you are not capturing the same error multiple times.

Think of the following (not unrealistic) scenario:

We add a fastify error handler to ensure we can get access to _all_ errors. In there, we add a LogRecord for the exception (pseudo code coming):

```js
fastify.addHook('onError', function(error) {
  opentelemetry.addLogRecord({
    'exception.message': error.message,
    'exception.stacktrace': error.stack
  });
});
```

This error is then sent to some backend, so far so good!

But if you _also_ have global error handlers (again, in order to get access to _all_ errors), you may capture the same error _again_, because based on the message & stack alone you cannot reliably identify that this error was already captured somewhere else:

```js
globalThis.addEventListener('error', function(error) {
  // we don't know if error was previously also handled by the fastify error hook :(
});
```

Now, some ideas for what we could do - I am more brainstorming here, so bear with me :D

1. We could add a non enumerable property to the error with an ID, and also add this to the LogRecord. Something like this:
  ```js
  fastify.addHook('onError', function(error) {
    const id = uuid();
    opentelemetry.addLogRecord({
      'exception.message': error.message,
      'exception.stacktrace': error.stack,
      'exception.id': id
    });
   Object.defineProperty(error, '__opentelemetryId', id);
  });
  ```
  While this would probably work, it would also be a bit specific, and probably hard to standardize :(

2. We can try to identify already sent errors via the message/stack only. This may work reasonably enough 90% of the time, but would def. lead to some false duplicates or similar.

pichlermarc:
> Now, some ideas for what we could do - I am more brainstorming here, so bear with me :D
> 
>  1. We could add a non enumerable property to the error with an ID, and also add this to the LogRecord. Something like this:
> 
> 
> ```js
> fastify.addHook('onError', function(error) {
>   const id = uuid();
>   opentelemetry.addLogRecord({
>     'exception.message': error.message,
>     'exception.stacktrace': error.stack,
>     'exception.id': id
>   });
>  Object.defineProperty(error, '__opentelemetryId', id);
> });
> ```
> 
> While this would probably work, it would also be a bit specific, and probably hard to standardize :(

I think this is an interesting idea. This approach is something that may be hard to do in some languages, but I can see how an `exception.id` attribute could be useful to distinguish the same errors that are reported multiple times. I don't think it is far-fetched that the folks over at https://github.com/open-telemetry/semantic-conventions may be interested in something like this. I encourage you to open an issue there and make that case - I think the example you gave here is sound and would be a great starting point for a discussion.

Having this in the semantic conventions is something that I'd recommend if you want to rely it for error de-duplication in production (otherwise the attribute may get a different meaning in the semantic-conventions later and both OTel authors and vendors will have to handle it; this tends to be a lot of work for all people that are involved). 

The way forward for this would be:
- open an issue in the semantic-conventions repository (which may lead to a PR)
- implement the feature here (we're happy to accept prototypes if that is something that's needed to get a possible semantic-conventions PR merged)

> 2. We can try to identify already sent errors via the message/stack only. This may work reasonably enough 90% of the time, but would def. lead to some false duplicates or similar.

Yes I can see this approach not being optimal.

pichlermarc:
@mydea we will not move forward with this as it's IMO out of scope for an OTel instrumentation at the moment. 
If semconv or the specification changes with regards to how we should handle errors we can re-consider adding a feature like this.

PR #1721, opened 2023-10-06
fix: fix context loss when cursor are accesed concurrently

### LINKED ISSUE #1688: [mongodb] instrumentation loosing context

@opentelemetry/api: 1.4.1 
@opentelemetry/auto-instrumentations-node: 0.39.2
@opentelemetry/sdk-metrics: 1.15.2
@opentelemetry/sdk-node: 0.41.2

and

mongodb: 5.8.1

All versions supported by `mongodb@5`
\>=14.20.1

### What did you do?
Created an app which does queries to MongoDB and instrumented with Otel to get tracing of HTTP spans and their child database spans.

Repo with the code
https://github.com/david-luna/node-mongodb-native-async-resource

When requesting the app concurrently I expect each HTTP span to have its child database span.

### What did you see instead?
Database spans are grouped as children of few HTTP spans and the rest of HTTP spans have no child.

### Additional context
I've done some research and found the issue is happening since `mongodb` driver uses a pool for connections queueing all command until a connection is available. Queued callbacks loose the execution context therefore mongodb spans are linked to the wrong parent HTTP span.

I've made a small gist to show how a pool affects async context
https://gist.github.com/david-luna/8a66fc1308301f79af8bd2eb65bcbc07

Also I've created a repo with an app that reproduces the issue which is affecting not only Otel but other tracers like `elastic-apm-node`.
https://github.com/david-luna/node-mongodb-native-async-resource

This could be fixed in `mongodb` itself if they do bind the execution context to the queued callbacks. I've opened an issue and created a PR to fix it but it's waiting for mongodb's team response.
issue: https://jira.mongodb.org/browse/NODE-5639
PR: https://github.com/mongodb/node-mongodb-native/pull/3871 

Even if they accept the fix we probably will need to support earlier versions for customers that can't update to the ones with the fix.

david-luna:

This PR is solving the problem of `mongodb` instrumentation loosing the context when cursors are accessed concurrently.

Add a new instrumentation on `ConnectionPool` class to patch the method that produces the context loss. This class is present from v4 and up so the fix works also for v5

Closes #1688


plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -288,6 +324,15 @@ export class MongoDBInstrumentation extends InstrumentationBase {
     };
   }
 
+  private _getV4ConnectionPoolCheckOut() {
+    return (original: V4ConnectionPool['checkOut']) => {
+      return function patchedCheckout(this: unknown, callback: any) {
+        const patchedCallback = AsyncResource.bind(callback);
```
trentm, line 330: It might be helpful for future maintenance to have a comment here that this patch will be unnecessary for future versions of mongodb that include the https://github.com/mongodb/node-mongodb-native/pull/3871 fix.

david-luna: done!

david-luna:
@osherv could you please add your feedback and approval if applies?

plugins/node/opentelemetry-instrumentation-mongodb/src/internal-types.ts
```diff
@@ -184,6 +184,13 @@ export type V4Connection = {
   ): void;
 };
 
+// https://github.com/mongodb/node-mongodb-native/blob/v4.2.2/src/cmap/connection_pool.ts
+export type V4ConnectionPool = {
+  // Instrumentation jsut cares about carrying the async context so
```
pichlermarc, line 189:
```suggestion
  // Instrumentation just cares about carrying the async context so
```

pichlermarc (approved):
Looks great; thanks for fixing this. :slightly_smiling_face: 
Edit: looks like this still needs some lint fixes but then this should be good to merge :slightly_smiling_face:

trentm (approved):
LGTM, modulo the linting and typo that Marc mentioned.

david-luna:
Thanks you both for the review :)

 @pichlermarc there are some workflows awaiting approval. Are you able to approve them?

PR #1728, opened 2023-10-11
feat(instrumentation-mongodb): support aggregation commands and support nested statements

Fox32:

At Serrala, we have a lot of micro services build with different languages that we observe with OpenTelemetry. We noticed that the mongo instrumentation of Node lacks some details compared to the Java instrumentation:

- Aggregations are not supported (they are just displayed as `mongo.command`).
- `db.statement` is sanitized to a degree that it doesn't contain enough details to pin down what the query is doing.

This PR adds support for aggregation command. As aggregation pipelines typically have a nested structure to describe the statement, the current way of sanitizing only leaves a low amount of hints what the command is about:

For example:

```ts
collection
  .aggregate([
    { $match: { key: 'value' } },
    { $group: { _id: '$a', count: { $sum: 1 } } },
  ])
  .toArray()
```

Is sanitized to:

```js
{
    aggregate: "?",
    pipeline: "?",
    cursor: "?",
}
```

The [Java instrumentation is handling statement sanitizing differently](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/mongo/mongo-3.1/library/src/main/java/io/opentelemetry/instrumentation/mongo/v3_1/MongoDbAttributesGetter.java#L173C26-L213). Instead of replacing all values of the top level object, it traverses into nested objects and array, keeping the existing structure. This PR implements the same behavior for the Node instrumentation. Therefore `db.statement` is in this case:

```js
{
    aggregate: "?",
    pipeline: [
      { $match: { key: "?" } },
      { $group: { _id: "?", count: { $sum: "?" } } },
    ],
    cursor: {},
}
```

While it is possible to configure a customer sanitizer for commands, I think it would better to add support for this upstream for better compatibility to the Java instrumentation.

I fixed some small typos I stumbled upon while reading through the contribution guidelines - I hope it is fine to include them here.


Fox32:
@pichlermarc I'm not sure whether the failing tests are related to my changes at all, looks more like a flaky test?

david-luna:
Sorry I can't trigger the CI

cc: @open-telemetry/javascript-maintainers

david-luna (approved):
Thanks again @Fox32 for your contribution :)

Semantic conventions define the `db.statement` as a string value ([link](https://github.com/open-telemetry/semantic-conventions/blob/ed7d3e0f939cbdcd6c0b6093d157d3f5e8bbc9bd/docs/database/database-spans.md?plain=1#L193)). Also the specific section for `mongodb` does not mention anything about statements ([link](https://github.com/open-telemetry/semantic-conventions/blob/ed7d3e0f939cbdcd6c0b6093d157d3f5e8bbc9bd/docs/database/mongodb.md?plain=1#L5))

However Java instrumentation and even this one also send an object in some cases (haven't checked other agents). I guess this is something that should be updated in semantic-conventions and we should check if there is an issue for that (or crate one if not).

Saying that this PR looks good :)

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v3.test.ts
```diff
@@ -309,6 +335,48 @@ describe('MongoDBInstrumentation-Tracing-v3', () => {
...
+              operationName,
+              SpanKind.CLIENT,
+              'aggregate',
+              undefined,
+              false,
+              false
+            );
+            const mongoSpan = spans.find(s => s.name === operationName);
+            const dbStatement = JSON.parse(
+              mongoSpan!.attributes[SemanticAttributes.DB_STATEMENT] as string
```
JamieDanielson, line 363:
```suggestion
              mongoSpan!.attributes[SEMATTRS_DB_STATEMENT] as string
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.test.ts
```diff
@@ -344,6 +370,48 @@ describe('MongoDBInstrumentation-Tracing-v4', () => {
...
+              operationName,
+              SpanKind.CLIENT,
+              'aggregate',
+              undefined,
+              false,
+              false
+            );
+            const mongoSpan = spans.find(s => s.name === operationName);
+            const dbStatement = JSON.parse(
+              mongoSpan!.attributes[SemanticAttributes.DB_STATEMENT] as string
```
JamieDanielson, line 398:
```suggestion
              mongoSpan!.attributes[SEMATTRS_DB_STATEMENT] as string
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v5-v6.test.ts
```diff
@@ -347,6 +373,48 @@ describe('MongoDBInstrumentation-Tracing-v5', () => {
...
+              operationName,
+              SpanKind.CLIENT,
+              'aggregate',
+              undefined,
+              false,
+              false
+            );
+            const mongoSpan = spans.find(s => s.name === operationName);
+            const dbStatement = JSON.parse(
+              mongoSpan!.attributes[SemanticAttributes.DB_STATEMENT] as string
```
JamieDanielson, line 401:
```suggestion
              mongoSpan!.attributes[SEMATTRS_DB_STATEMENT] as string
```

JamieDanielson:
Looks like the merge from main made the change for `SemanticAttributes.*` changing to `SEMATTRS_*`, which is what caused the tests to fail. I've just commited those changes are will see how the tests do this time around 😄

Fox32:
@JamieDanielson thanks for taking care!

Issue #1708, opened 2023-10-02
Redis-4 instrumentation causes application to crash when multi command soft-fails


FredrikAugust:
### What version of OpenTelemetry are you using?

Using the image `otel/opentelemetry-collector-contrib:0.83.0`

### What version of Node are you using?

Node 18.17.1

### What did you do?

We use the auto-instrumentation to instrument our application, and this correctly injects redis-4. We use the `multi` command throughout our application. 

### What did you expect to see?

When a watched key changes within a `multi` command we expect it to behave the same way as it would _without_ the instrumentation.

### What did you see instead?

If we enable the instrumentation, we see a message 

```
no original function multi to wrap
```

Printed to the console when we start the application, and the application will crash when a watched key changes in a `multi` command. This does not happen if we disable the instrumentation.

```
/app/node_modules/@redis/client/dist/lib/multi-command.js:60
            throw new errors_1.WatchError();
                  ^

WatchError: One (or more) of the watched keys has been changed
    at RedisMultiCommand.handleExecReplies (/app/node_modules/@redis/client/dist/lib/multi-command.js:60:19)
    at Commander.exec (/app/node_modules/@redis/client/dist/lib/client/multi-command.js:81:82)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### Additional context

https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1672

This issue is related as it requests the ability to disable a single instrumentation, and it is where we first identified the issue.

drob:
Seeing the same behavior on node 16 as well.

FredrikAugust:
The only two places we use `.watch()` in our backend, the code looks roughly like this:

```ts
try {
  await getClient().executeIsolated(async (client) => {
    await client.watch(`lock`);
  
    const owner = await client.get(`lock`);

    return client.multi().set(`lock`, ..., { PXAT: ... });
  });
} catch (err) {
  if (err instanceof WatchError) throw new LockInUseError();
}
```

But our application crashes because of an uncaught `WatchError` error. I can't see any places in our application where this could occur, and sadly the stacktrace doesn't reveal where it's thrown from.

FredrikAugust:
I managed to patch this in a fork, but I'm unsure if it's the right way to proceed. 

[...]

PR #1729, opened 2023-10-12
fix(instrumentation-redis-4): avoid shimmer warning by only wrapping multi/MULTI if they exist

trentm:
This fixes the "no original function multi to wrap" warning from shimmer when wrapping recent versions of `@redis/client`.

Refs: #1708

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
@@ -166,23 +166,30 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
         this._diag.debug('Patching redis client');
         const redisClientPrototype = moduleExports?.default?.prototype;
 
-        if (isWrapped(redisClientPrototype?.multi)) {
-          this._unwrap(redisClientPrototype, 'multi');
+        // In some @redis/client versions 'multi' is a method. In later
+        // versions, as of https://github.com/redis/node-redis/pull/2324,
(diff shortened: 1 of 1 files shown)


plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts
```diff
@@ -166,23 +166,30 @@ export class RedisInstrumentation extends InstrumentationBase<any> {
         this._diag.debug('Patching redis client');
         const redisClientPrototype = moduleExports?.default?.prototype;
 
-        if (isWrapped(redisClientPrototype?.multi)) {
-          this._unwrap(redisClientPrototype, 'multi');
+        // In some @redis/client versions 'multi' is a method. In later
+        // versions, as of https://github.com/redis/node-redis/pull/2324,
+        // 'MULTI' is a method and 'multi' is a property defined in the
+        // constructor that points to 'MULTI'.
```
pichlermarc, line 172: Thanks for adding the comment here to point out the reason. One suggestion (non-blocking): to me the wording in the issue that explained that it's not part of the prototype helped in understanding the changes in this PR, maybe we should also add it here :) 
```suggestion
        // In some @redis/client versions 'multi' is a method. In later
        // versions, as of https://github.com/redis/node-redis/pull/2324,
        // 'MULTI' is a method and 'multi' is a property defined in the
        // constructor that points to 'MULTI', and therefore it will not
        // be defined on the prototype.
```

pichlermarc (approved):
Thank you for investigating and fixing this :slightly_smiling_face:

PR #1730, opened 2023-10-12
fix(instrumentation-redis-4): fix unhandledRejection in client.multi(...) handling

trentm:
The instrumentation was not handling a rejection of the promise from client.multi(), resulting in unended spans and an unhandleRejection event.

Fixes: #1708

---

This is based on work by @FredrikAugust in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1717.


pichlermarc (approved):
Looks good, thank you for taking care of this so quickly - and thanks for the detailed explaination on the original issue. :slightly_smiling_face:

PR #1732, opened 2023-10-12
chore: specify label for stale, exempt never-stale

JamieDanielson:

- We recently in (#1709 ) exempted bug labels from stalebot; we then realized an issue may be labeled as a bug that should still go stale if no response, and the only way to avoid it going stale is to remove the bug label, which doesn't seem right either. We should be intentional about requesting information and letting issues and PRs go stale if necessary, without needing to remove labels to allow stalebot to come through.

- add [`any-of-labels`](https://github.com/actions/stale#any-of-labels) to the workflow, so any issue or PR labeled as "information-requested" is eligible for stalebot. This is comma-separated allow-list of labels. If we end up having different labels, we can add these here, and we can also get more granular later if needed with `any-of-issue-labels` and `any-of-pr-labels`.
- remove instrumentation-request and bug from exempted labels, to allow stalebot to come through if it's appropriately labeled as "information-requested" (otherwise we'd need to remove the bug or instr-request label, or manually close).
- intentionaly keep the never-stale labels for issue and PR. We may request more information but still want it to be never-stale, so keeping the exempt label first should allow for an early return.


pichlermarc (approved):
Oh nooo, I forgot about this PR :disappointed: 
Sorry about that. Thank you for taking care of this @JamieDanielson

pichlermarc:
> This is weird - I tried clicking Update Branch and I get this error:
> 
> > Update branch attempt failed
> > Couldn't update "jamie.update-stalebot": Changes must be made through a pull request. Required status check "EasyCLA" is expected. You're not authorized to push to this branch. Visit https://docs.github.com/articles/about-protected-branches/ for more info
> 
> Am I missing something obvious?

Ah looks like some branch protection rule was preventing that. I updated them to meet the guidelines set by the organization a while ago - looks like it was possible for approves to create and push to a branch before, now it is not possible anymore.

PR #1733, opened 2023-10-12
feat: allow disabling auto-instrumentations using env variable

maxkomarychev:
# Rationale

While I am pretty happy with provided default instrumentations I would like to be able to disable some of them based on environment variables alone without getting into business of configuring integration manually

### DIFF
--- a/metapackages/auto-instrumentations-node/src/register.ts
+++ b/metapackages/auto-instrumentations-node/src/register.ts
@@ -25,8 +25,17 @@ diag.setLogger(
   opentelemetry.core.getEnv().OTEL_LOG_LEVEL
 );
 
+const disabledAutoinstrumentations = (
+  process.env.OTEL_DISABLED_AUTOINSTRUMENTATIONS ?? ''
+).split(',');
+
 const sdk = new opentelemetry.NodeSDK({
-  instrumentations: getNodeAutoInstrumentations(),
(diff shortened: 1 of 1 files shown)


metapackages/auto-instrumentations-node/src/register.ts
```diff
@@ -25,8 +25,17 @@ diag.setLogger(
   opentelemetry.core.getEnv().OTEL_LOG_LEVEL
 );
 
+const disabledAutoinstrumentations = (
+  process.env.OTEL_DISABLED_AUTOINSTRUMENTATIONS ?? ''
```
blumamir, line 29: This environment variable is not speced. There is the [`OTEL_SDK_DISABLED`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration) environment variable from spec which I think might be what you are looking for. 

However, I think OTEL should be disabled in the SDK level and not in the instrumentations level. Or are you interested in installing an SDK and only having auto instrumentations disabled while allowing manual instrumentation via trace/metrics/logs providers?

What distro are you currently using to instrument your code?

blumamir:
Hi @maxkomarychev 

Please let us know if you are still interested in this feature, or if the PR can be closed. Thanks! 🙏

maxkomarychev:
@blumamir I'll close this for now. 👍

Issue #1731, opened 2023-10-12
Introduce ESM testing strategy


pichlermarc:
### Is your feature request related to a problem? Please describe

In the `@opentelemetry/instrumentation` package, we added experimental ESM support a while ago (https://github.com/open-telemetry/opentelemetry-js/pull/3698). As more and more users started using this, the lack of ESM-releated testing became apparent. A high-priority bug that's present in a large number of instrumentations causes applications to crash as wrapping ES modules fails.

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1693
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1588
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1587
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1519
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1692

There have been PRs opened to address this; however, as we don't test for ESM, it is hard to verify fixes and catch regressions.

### Describe the solution you'd like to see

We should start testing if instrumentations also work with ESM. This issue tracks 
- [ ] documenting an ESM testing strategy for existing packages
- [ ] implementing the suggested tests in at least two instrumentation packages
- [ ] adding a step in the unit tests workflow to run the ESM tests

pichlermarc:
I've been experimenting a bit, and we may be able to re-use the existing `mocha` tests in ESM mode by using the `ts-node/esm` loader and chaining it with `import-in-the-middle` to get full coverage.

At the moment one thing preventing us from doing so is that once I add `@opentelemetry/instrumentation/hook.mjs` as a loader, the `require.main === module` check in `./node_modules/mocha/lib/cli/cli.js` fails and mocha exits without running tests. I think this may be related to how `import-in-the-middle` works, so I opened https://github.com/DataDog/import-in-the-middle/issues/34.

PR #1735, opened 2023-10-13
feat(test-utils): runTestFixture utility for running out-of-process tests

trentm:
Running scripts out-of-process for testing allows running tests with different node options and with isolation. The former is useful for ESM testing. This change includes adding an ESM test for ioredis. This depends on #1694 to pass.

Closes: #1731

- ESM testing. See https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1731 and some discussion on ESM testing options here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1694#issuecomment-1759782571

(Note: This is one option for #1731. An alternative is being discussed over there.)

- This adds a `runTestFixture(...)` test utility function that takes a script to run out-of-process. It is expected that the script uses OTel tracing using `NodeSDK` from `@opentelemetry/sdk-node`.  The utility first starts a `TestCollector` HTTP server that can collect OTLP (using the HTTP/JSON flavour). Then it execs the script with `OTEL_` env vars set so that `NodeSDK` will export to that collector. When the script finishes, a couple given callbacks, `checkResult` and `checkCollector`, are called so that the mocha test code can assert expectations about the process result and the collected spans. The `TestCollector` provides a `collector.sortedSpans` that returns the exported spans sorted by start-time.
- As a first user of this utility, I've added a "fixtures/use-ioredis.mjs" that uses ioredis from ESM code. It does a simple `client.set(...)` and `client.get(...)`. An added case in "ioredis.test.ts" runs that and asserts that the script exited successfully and exported the expected spans.
- The "fixtures/use-ioredis.mjs" script can be run independent of the test suite, when playing with developing these. By default it will export spans to the console and connect to a redis on the default port. For example:

```
% docker run --rm -p 6379:6379 -d redis
74cfbbf1d636aaaaacc2669d0d4b5fa10d0467feea530c8d3c98813e308b0c62

% node --experimental-loader=@opentelemetry/instrumentation/hook.mjs test/fixtures/use-ioredis.mjs
(node:29410) ExperimentalWarning: Custom ESM Loaders is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
{
  traceId: '4df8b32419a6252d4995a23bfe70b15c',
  parentId: '1d319fd2c2dea2dd',
  traceState: undefined,
  name: 'set',
  id: '9f207451efc4391d',
  kind: 2,
...
```

@pichlermarc and others interested, I would love some feedback on this. 

## Checklist

- [x] wait for https://github.com/open-telemetry/opentelemetry-js/pull/4220 fix and a release with that, then update this PR to use the fixed `startTimeUnixNano` serialization


trentm:
The first commit does *not* include the fix from https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1694 for ioredis ESM support, so the test fails:

```
% npm run test:local

> @opentelemetry/instrumentation-ioredis@0.35.2 test:local
> cross-env RUN_REDIS_TESTS_LOCAL=true npm run test


> @opentelemetry/instrumentation-ioredis@0.35.2 test



  ioredis
    ✓ should have correct module name
    ✓ should work with ESM usage (450ms)
    1) should work with ESM usage


  2 passing (1s)
  1 failing

  1) ioredis
       should work with ESM usage:
     Uncaught ifError got unwanted exception: Command failed: /Users/trentm/.nvm/versions/node/v16.20.2/bin/node fixtures/use-ioredis.mjs redis://localhost:63790
/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/build/src/instrumentation.js:131
                if ((0, instrumentation_1.isWrapped)(moduleExports.prototype.sendCommand)) {
                                                                             ^

TypeError: Cannot read properties of undefined (reading 'sendCommand')
    at InstrumentationNodeModuleDefinition.patch (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/build/src/instrumentation.js:131:78)
    at IORedisInstrumentation._onRequire (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:153:39)
    at hookFn (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:194:29)
    at callHookFn (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/import-in-the-middle/index.js:28:22)
    at Hook._iitmHook (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/import-in-the-middle/index.js:76:11)
    at /Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/import-in-the-middle/index.js:17:41
    at Array.forEach (<anonymous>)
    at addHook (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/import-in-the-middle/index.js:17:10)
    at new Hook (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/import-in-the-middle/index.js:84:3)
    at IORedisInstrumentation.enable (/Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:206:29)

  AssertionError [ERR_ASSERTION]: ifError got unwanted exception: Command failed: /Users/trentm/.nvm/versions/node/v16.20.2/bin/node fixtures/use-ioredis.mjs redis://localhost:63790
  /Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/build/src/instrumentation.js:131
                  if ((0, instrumentation_1.isWrapped)(moduleExports.prototype.sendCommand)) {
                                                                               ^

  TypeError: Cannot read properties of undefined (reading 'sendCommand')
      at InstrumentationNodeModuleDefinition.patch (build/src/instrumentation.js:131:78)
      at IORedisInstrumentation._onRequire (node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:153:39)
      at hookFn (node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:194:29)
      at callHookFn (node_modules/import-in-the-middle/index.js:28:22)
      at Hook._iitmHook (node_modules/import-in-the-middle/index.js:76:11)
      at /Users/trentm/tm/opentelemetry-js-contrib3/plugins/node/opentelemetry-instrumentation-ioredis/node_modules/import-in-the-middle/index.js:17:41
      at Array.forEach (<anonymous>)
      at addHook (node_modules/import-in-the-middle/index.js:17:10)
      at new Hook (node_modules/import-in-the-middle/index.js:84:3)
      at IORedisInstrumentation.enable (node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:206:29)

      at ChildProcess.exithandler (node:child_process:402:12)
      at ChildProcess.emit (node:events:513:28)
      at maybeClose (node:internal/child_process:1100:16)
      at Process.ChildProcess._handle.onexit (node:internal/child_process:304:5)
      at Process.callbackTrampoline (node:internal/async_hooks:130:17)



--------------------|---------|----------|---------|---------|-----------------------------
File                | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
--------------------|---------|----------|---------|---------|-----------------------------
All files           |   34.88 |     5.26 |   47.36 |   34.52 |
 instrumentation.ts |   34.17 |     5.45 |      50 |   34.61 | 55,63,73-76,100-178,187-213
 utils.ts           |   42.85 |        0 |       0 |   33.33 | 23-30
--------------------|---------|----------|---------|---------|-----------------------------
```

trentm:
The second commits adds the fix from #1694 to show that the test now passes:

```
% npm run test:local
...

  ioredis
    ✓ should have correct module name
    ✓ should work with ESM usage (541ms)
...
```

pichlermarc:
Thanks for working on this; having your help on the topic is very appreciated. :rocket: 

> I don't know if there is a good way with lerna to develop in "packages/opentelemetry-test-utils" and have that used in, for example, "plugins/node/opentelemetry-instrumentation-ioredis". I could `npm install ../../../packages/opentelemetry-test-utils` for local dev, but I'm not sure what can be done before merging.

It should be possible to add ` "@opentelemetry/contrib-test-utils": "^0.34.2",` to `devDependencies`, `lerna` should link the version from the repo when running `npm install` from the root of the repository. When you re-install from the package directory, you can run `npm run --prefix ../../../ lerna:link` to restore the links. :slightly_smiling_face: Moving everything to the intended packages in this PR should then be possible.

> * This adds a `runTestFixture(...)` test utility function that takes a script to run out-of-process. It is expected that the script uses OTel tracing using `NodeSDK` from `@opentelemetry/sdk-node`.  

I wonder if we should just do a simple manual tracing setup without `@opentelemetry/sdk-node`. As there'd be less moving parts, I'd expect such a setup to be more stable. :thinking: 

> The utility first starts a `TestCollector` HTTP server that can collect OTLP (using the HTTP/JSON flavour). Then it execs the script with `OTEL_` env vars set so that `NodeSDK` will export to that collector. When the script finishes, a couple given callbacks, `checkResult` and `checkCollector`, are called so that the mocha test code can assert expectations about the process result and the collected spans. The `TestCollector` provides a `collector.sortedSpans` that returns the exported spans sorted by start-time.

This is amazing; I can see this possibly being very useful for integration testing. I'm wondering though if we may be able to get away with something a bit simpler. Over at #1731 I added a comment that I tried running the existing tests in ESM tests, and was almost successful (% some trouble with iitm+mocha, which may be resolvable). This would basically take care of the telemetry testing with the same coverage as CJS, while out-of-process testing could catch some regressions that would be considered `priority:p1` (anything that crashes the app). While I would prefer to do what I mentioned above, I'm unsure if it will pan out (or if there's concerns with going that way). Therefore let's keep the approach of this PR then iterate depending on the outcome of the comment I added to #1731.

trentm:
> you can run `npm run --prefix ../../../ lerna:link` to restore the links

That was the part I was missing, thanks.



> I wonder if we should just do a simple manual tracing setup without `@opentelemetry/sdk-node`. As there'd be less moving parts, I'd expect such a setup to be more stable. 🤔

I don't have strong opinions here yet. Some reasons basing on `NodeSDK` was nice:

- IIUC, using NodeSDK is the usage that users of OTel JS will be encouraged to take, so the fixture scripts will be using OTel setup code that is likely more typical of real usage.  (One example I noticed is that some test modules that need a ContextManager tend to use the AsyncHooksContextManager -- presumably because it is the lowest common denominator -- rather than using the AsyncLocalStorageContextManager for recent-enough Node.js versions. However, I don't have reason to think that'd cause any test issues.)
- The NodeSDK support for the `OTEL_*` envvars for configuring, for example for the OTLP endpoint, was very convenient. This same convenience would apply for testing metrics and logs output sent via OTLP.

I definitely agree that fewer moving parts should be more helpful. And if your other idea pans out, then testing ESM **in-process** means that using an `InMemoryExporter` for collecting test data is a lot lighter (faster, smaller deps) than having OTLP involved.


> Therefore let's keep the approach of this PR then iterate depending on the outcome of the comment I added to #1731.

Sounds good.

trentm:
@pichlermarc I've moved the test utility code over the test-utils package now, so it should be clearer what the actual usage would look like.  I'm still totally fine waiting on the outcome of your other option on #1731.

plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
```diff
@@ -51,13 +51,13 @@ const CONFIG = {
   port: parseInt(process.env.OPENTELEMETRY_REDIS_PORT || '63790', 10),
 };
 
-const URL = `redis://${CONFIG.host}:${CONFIG.port}`;
+const REDIS_URL = `redis://${CONFIG.host}:${CONFIG.port}`;
```
trentm, line 54: Reviewer note: I could revert this variable name change. I had changed it originally because at one point I had been using the [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) object.

trentm:
We discussed today in the SIG going ahead with this approach. Note that I opened a draft attempt of the other approach at https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1752 with some questions/concerns.


> > I wonder if we should just do a simple manual tracing setup without `@opentelemetry/sdk-node`. As there'd be less moving parts, I'd expect such a setup to be more stable. 🤔
> 
> I don't have strong opinions here yet. Some reasons basing on `NodeSDK` was nice: [...]

@pichlermarc Let me know if you'd prefer me to update this PR to have it avoid using `NodeSDK`.

trentm:
The build is failing because it needs an update to all otel core deps to 0.45.0 (equivalent to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1725 for the new release).  I'm not sure which "schedule" entry in "renovate.json" is relevant.

packages/opentelemetry-test-utils/src/test-fixtures.ts
```diff
@@ -0,0 +1,251 @@
...
+  resource: IResource;
+  instrumentationScope: IInstrumentationScope;
+};
+
+/**
+ * A minimal HTTP server that can act as an OTLP HTTP/JSON protocol collector.
+ * It stores the received data for later test assertions.
+ *
+ * Limitations: This only supports traces so far, no metrics or logs.
+ * There is little error checking here; we are assuming valid OTLP.
```
pichlermarc, line 71: I'd say this is likely fine. We can expand this later if necessary.
Little error checking also sounds okay to me, since it's for testing.

packages/opentelemetry-test-utils/src/test-fixtures.ts
```diff
@@ -0,0 +1,251 @@
...
+      if (req.method === 'POST' && parsedUrl.pathname === '/v1/traces') {
+        if (req.headers['content-type'] !== 'application/json') {
+          resStatusCode = 415;
+          resBody = JSON.stringify({ message: 'invalid content-type' });
+        } else {
+          this._ingestTraces(body);
+          resStatusCode = 200;
+          // TODO: I'm not sure what the ExportTraceServiceResponse form should be.
+          // https://github.com/open-telemetry/opentelemetry-proto/blob/v1.0.0/opentelemetry/proto/collector/trace/v1/trace_service.proto
+          resBody = '{}';
```
pichlermarc, line 143: Empty like this is fine. 🙂 

partialSuccess should only be set if there's certain spans set if some of the spans are rejected by the remote, we always accept everything here, so this is fine.

plugins/node/opentelemetry-instrumentation-ioredis/test/ioredis.test.ts
pichlermarc: It's fine to keep the change imo. 🙂

pichlermarc (approved):
Apologies for taking so long to review. I think keeping NodeSDK is fine, I'll be closer to what people are using as well. 

Thank you for working on this. 🚀

packages/opentelemetry-test-utils/src/test-fixtures.ts
trentm: Thanks for the confirmation. I removed the "TODO" in commit e0be467f.

MSNev (approved):

PR #1736, opened 2023-10-13
fix(koa): fix instrumentation of ESM-imported koa

chentsulin:

- ESM-imported koa is not patched by instrumentation.

- Use the same approach introduced in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1694


david-luna (approved):

pichlermarc:
Hi, thanks for your contribution and sorry for the late review. #

#1735 introduced a simple way to test if everything is working as intended with ESM instrumentations. Would you mind adding this to this PR as well?

chentsulin:
Hi @pichlermarc, I failed to create a proper test case based on the #1735 PR. Is there anyone can help?

trentm:
I can help add a test case.

trentm (approved):

JamieDanielson (approved):
Thanks for this!

trentm:
Thanks for the reviews. I turned on the TAV tests here and I see that they fail, so I may need to dig into that. It might be a separate issue, though.

trentm:
Thanks @chentsulin!

Issue #1718, opened 2023-10-05
bug: host-metrics reporting wrong system and process CPU utilization


david-luna:
### What version of OpenTelemetry are you using?
"dependencies": {
    "@opentelemetry/api": "^1.4.1",
    "@opentelemetry/auto-instrumentations-node": "^0.39.2",
    "@opentelemetry/host-metrics": "^0.33.1",
    "@opentelemetry/sdk-metrics": "^1.15.2",
    "@opentelemetry/sdk-node": "^0.41.2"
  }

### What version of Node are you using?
v18.16.0

### What did you do?

Instrument an application with console metrics exporter and watch the CPU utilisation values. There is a repo with the app

https://github.com/david-luna/otel-metrics-cpu-utilization

### What did you expect to see?
According to [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemcpuutilization) I expect to see CPU utilization values between 0 and 1.

### What did you see instead?
Values are way above 1 like the sample provided in the repo I get some metric readings like this one.

```js
{
  descriptor: {
    name: 'system.cpu.utilization',
    type: 'OBSERVABLE_GAUGE',
    description: 'Cpu usage time 0-1',
    unit: '',
    valueType: 1
  },
  dataPointType: 2,
  dataPoints: [
    {
      attributes: [Object],
      startTime: [Array],
      endTime: [Array],
      value: 157.82086149302881
    },
    {
      attributes: [Object],
      startTime: [Array],
      endTime: [Array],
      value: 143.03498303049165
    },
    {
      attributes: [Object],
      startTime: [Array],
      endTime: [Array],
      value: 1083.5302883959548
    },
    {
      attributes: [Object],
      startTime: [Array],
      endTime: [Array],
      value: 1362.709911550165
    },
    {
      attributes: [Object],
      startTime: [Array],
      endTime: [Array],
      value: 0
    }
    // ...
  ]
}
```

### Additional context

N/A

david-luna:
@legendecas since you're the module owner could you please have a look? Thanks :)

legendecas:
Right, and the semantic convention the package used is outdated as well. We should update the semantic convention all together.

david-luna:
@legendecas thanks for the confirmation :)

I'm planning on crate a PR which:
- will refer to the updated semantic convention
- will change the calculation to make diffs from the latest measurement as semantic convention specifies

[...]

PR #1741, opened 2023-10-16
fix!: host-metrics `system.cpu.utilization` calculation fix

david-luna:

this PR is fixing the calculation of `system.cpu.utilization` metrics which is currently returning values that are out of the range [0,1] specified in semantic conventions. See #1718 

the calculation now is made based on the previous values returned from the `os.cpus` NodeJS API and the time it was called. Also `Date.now` static function is used to avoid unnecessary object creation (date objects) on each measurement.


packages/opentelemetry-host-metrics/test/metric.test.ts
```diff
@@ -156,31 +163,31 @@ describe('Host Metrics', () => {
...
       ensureValue(metric, { state: 'nice', cpu: '1' }, 0);
     });
 
     it('should export CPU utilization metrics', async () => {
       const metric = await getRecords(reader, 'system.cpu.utilization');
 
-      ensureValue(metric, { state: 'user', cpu: '0' }, 30247.935978659552);
-      ensureValue(metric, { state: 'system', cpu: '0' }, 21071.23374458153);
-      ensureValue(metric, { state: 'idle', cpu: '0' }, 124998.56618872957);
+      ensureValue(metric, { state: 'user', cpu: '0' }, 0.7);
```
dyladan, line 182: Is the value rounded or is this just a coincidence that they're all 1 significant figure?

packages/opentelemetry-host-metrics/src/stats/common.ts
```diff
@@ -20,30 +20,56 @@ import { CpuUsageData, MemoryData, ProcessCpuUsageData } from '../types';
...
 
 /**
- * It returns cpu load delta from last time - to be used with SumObservers.
- * When called first time it will return 0 and then delta will be calculated
+ * For each CPU returned by `os.cpus()` it returns
+ * - the CPU times in each state (user, sys, ...) in seconds
+ * - the % of time the CPU was in each state since last measurement
+ *
+ * The first time will return 0 for % masurements since there is not enough
+ * data to calculate it
```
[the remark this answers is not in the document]
david-luna: it would make testing harder but it is possible. I'll give it a try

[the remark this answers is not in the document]
david-luna: Luckily the test is setup in a way that is not affected by this data preload. The changes are pushed :)

david-luna:
@legendecas @dyladan could you give another look please?

packages/opentelemetry-host-metrics/src/enum.ts
```diff
@@ -27,6 +27,11 @@ export enum METRIC_NAMES {
   PROCESS_MEMORY_USAGE = 'process.memory.usage',
 }
 
+export enum METRIC_ATTRIBUTES {
```
david-luna, line 30: I've added this enum to properly report the attributes with their full name as set on the semantic convettions

See: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemcputime

packages/opentelemetry-host-metrics/test/metric.test.ts
legendecas: It's not a coincidence. The data is mocked with `packages/opentelemetry-host-metrics/test/mocks/cpu.json`.

legendecas (approved):

legendecas:
Thank you for working on this.

david-luna:
Thanks @legendecas :)

I'll prepare another PR for `process.cpu.*` metrics once this one is merged.

PR #1749, opened 2023-10-20
add event listeners for each event on init

pkanal:

-

-


martinkuba:
The behavior is intended to generate spans only for events that have handlers in the application code. The handlers in application code will generally invoke other work (e.g. HTTP requests), which then make the interactions more significant (longer duration, part of a bigger trace, applicable to application logic) than just tracking all events that happen on the page. If there is a use case to track all events, I would suggest we add a new instrumentation that does not rely on Zone.js.

PR #1753, opened 2023-10-26
fix: Use real fastify request type

asakusuma:
Replace the `any` with the real type. I tested this locally in a node app.

The request object coming from the `requestHook` is of type `any`, so you don't get autocompletion of fields on the request object.

Import the type from fastify package.


Flarna:
This change is problematic.
Using fastify types in public interface requires that end users (using typescript) have fastify types installed. Fastify types are part of fastify itself so this would require that all users have fastify installed.
In special users for users of `@opentelemetry/auto-instrumentations-node` this is often not the case.

So it would be needed to add fastify as a dependency here which is not wanted because the instrumentation should not depend on the module to instrument.

See [GUIDELINES](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md) for more details.

blumamir (changes requested):
Thanks for the PR @asakusuma 

I assume you tested it with a setup that includes `fastify` in the node_modules.

The `types.ts` file is part of the package public API, since it is exported from `index.ts`. 
Thus, any typescript program that will import the instrumentation will hit this line and will attempt to 

```js
import type { FastifyRequest } from 'fastify';
```

If your setup does not include the `fastify` package, typescript will fail transpilation, which is the reason why this object is typed with `any` in the first place.

You can read more info [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md).

blumamir:
This PR cannot be merged as is due to the issues mentioned above.
If you think this is a mistake, please feel free to re-open.

Thanks!

PR #1754, opened 2023-10-26
fix(instrumentation-koa): handle koa routes being of type RegExp

ramesius:

Koa route paths can be `string | RegExp`. `RegExp` [is not a primitive type](https://github.com/open-telemetry/opentelemetry-js/blob/1c7d7a354281c720e7229d3d5498bbec8d3b2b8d/packages/opentelemetry-core/src/common/attributes.ts#L87-L96) and thus results in the following logs from Otel when using the koa instrumentation:
- `Invalid attribute value set for key: http.route`
- `Invalid attribute value set for key: koa.name`

This change updates the typings package `@types/koa__router` that in-turn has been updated with the correct union type `string | RegExp` (See PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67194). The Koa instrumentation now properly handles the case when routes are of type `RegExp`.

I have also replaced the use of the deprecated `SpanAttributes` with `Attributes`.


ramesius:
@david-luna Is this something that you could look into?

david-luna:
> @david-luna Is this something that you could look into?

I can give you feedback on this PR for now but nothing more since I don't have the required role/access for actioning on PRs.

So PR LGTM but I was wondering if its necessary to send `KOA_TYPE` and `HTTP_ROUTE` attributes if `layerPath` is not defined (not passed in the function call)

david-luna (approved):

blumamir:
@ramesius Thank you for the contribution. the fix LGTM.
Can you please add a test for this case?

ramesius:
@blumamir Sure, just pushed up a test for the functionality 👍🏻

ramesius:
@david-luna Probably makes sense. I can raise a seperate PR with that feedback if you don't mind.

blumamir (approved):
Thanks! 🥇

ramesius:
CI was failing due to linting, just fixed that up 🤞🏻

Issue #1757, opened 2023-10-29
FastifyDeprecation warning for accessing `request.routerPath` when accessing a non-existing page


omBratteng:
Separating the issue reported here https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1275#issuecomment-1777456234

### What version of OpenTelemetry are you using?

`@opentelemetry/instrumentation-fastify@032.3`

### What version of Node are you using?

`v18`

### What did you do?

Instrumented node with fastify instrumentation, did a request to a non-existing page.

Can be reproduced by starting the [fastify example](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/fastify) and doing a request to any non-existent page

```
curl http://localhost:8080/i_do_not_exist
```

### What did you expect to see?

No deprecation notices

### What did you see instead?

```
(node:9) [FSTDEP017] FastifyDeprecation: You are accessing the deprecated "request.routerPath" property. Use "request.routeOptions.url" instead. Property "req.routerPath" will be removed in `fastify@5`.
```

### Additional context

We are using the latest version of the intrumentation

https://github.com/dailydotdev/daily-api/blob/dc3bbecd7c04984a60a224b19ae9752102caf09e/package.json#L48

and version 4.24.2 of fastify

https://github.com/dailydotdev/daily-api/blob/dc3bbecd7c04984a60a224b19ae9752102caf09e/package.json#L71

So digging a bit further into it, I think it might when the requests are to not found pages. but then both `request.routeOptions.url` and `request.routerPath` are undefined.

Wazbat:
I'm also getting this when accessing existing pages, is that perhaps an unrelated issue?

omBratteng:
That might be related to #1275 which should have been fixed in #1679

dyladan:
I'll cut a release that includes #1679 and we can see if this is fixed. Seems to me that it shoudl be.

10xLaCroixDrinker:
@dyladan That says it was released 3 weeks ago. Was it not?

dyladan:
Sorry I didn't confirm before typing that. @Wazbat @10xLaCroixDrinker  can you confirm your issue is fixed?

Seems we didn't catch all uses of deprecated properties then

trentm:
@dyladan I can take this. I was just looking at it.

omBratteng:
[...]

PR #1710, opened 2023-10-02
fix(instrumentation-fastify): add tav script

pichlermarc:
## Which problem is this PR solving?

We currently don't test against `fastify` versions after `4.18.0` because `fasitify` has changed their minimum supported typescript version, this PR adds a `test-all-versions` script to the `@opentelemetry/instrumentation-fastify` package. 

## Short description of the changes

Adds TAV script to test against later versions of `fastify`

PR #1763, opened 2023-11-01
fix(instrumentation-fastify): fix span attributes and avoid FSTDEP017 FastifyDeprecation warning for 404 request

trentm:
For a 404, `request.routeOptions.url` is undefined. Since fastify@4.10.0
when routeOptions was added, we shouldn't fallback to the deprecated
request.routerPath.

This also corrects the assumption that the handler name is "bound ..."
in all cases. E.g. for a 404 it is Fastify's core "basic404" internal
function.

Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1757


trentm:
@pichlermarc I have a side question about the .tav.yml file for instrumentation-fastify (added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1710).  Is it the intention to only test with a single explicit version of fastify ("4.23.2")? 

Oh, or do I need to *not* update the fastify version in "package.json" in this PR, **because** versions after 4.18.0 require a typescript greater than 4.4.4?   I'll see if CI fails and then move it back to `"fastify": "4.18.0"` if needed.

If we want `test-all-versions` to test with recent releases of fastify as they come out, we could have a .tav.yml like this:

```diff
--- a/plugins/node/opentelemetry-instrumentation-fastify/.tav.yml
+++ b/plugins/node/opentelemetry-instrumentation-fastify/.tav.yml
@@ -1,5 +1,5 @@
-"fastify":
-  - versions: "4.23.2"
+fastify:
+  - versions: "4.0.0 || >4.24.3 <5"
     commands: npm run test
 "typescript":
   - versions: "4.7.4"
```

This would sanity test with the first 4.0.0 release and then with any releases at or after the current latest. This *does* require occasionally bump that `>4.24.3` if/when there are so many fastify releases that it becomes slow/expensive to test them all.

trentm:
> Oh, or do I need to _not_ update the fastify version in "package.json" in this PR, **because** versions after 4.18.0 require a typescript greater than 4.4.4? I'll see if CI fails and then move it back to `"fastify": "4.18.0"` if needed.

Indeed it did fail the `npm run compile`. I've reverted the fastify version back to 4.18.0.

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
```diff
@@ -96,8 +96,9 @@ export class FastifyInstrumentation extends InstrumentationBase {
       const anyRequest = request as any;
 
       const rpcMetadata = getRPCMetadata(context.active());
-      const routeName =
-        anyRequest.routeOptions?.config?.url || request.routerPath;
+      const routeName = anyRequest.routeOptions
+        ? anyRequest.routeOptions.url
+        : request.routerPath;
```
trentm, line 101: Regarding the codecov warning on this line and below: TAV tests *could* cover line 101, but currently they won't because it doesn't go back to an early-enough fastify version. Shall I add an `// istanbul ignore next` directive here?

pichlermarc:
> > Oh, or do I need to _not_ update the fastify version in "package.json" in this PR, **because** versions after 4.18.0 require a typescript greater than 4.4.4? I'll see if CI fails and then move it back to `"fastify": "4.18.0"` if needed.
> 
> Indeed it did fail the `npm run compile`. I've reverted the fastify version back to 4.18.0.

Sorry, I was out of office so I wasn't able to respond. This is exactly the reason. We can't bump beyond 4.4.4 because the fastify instrumentation is included by `@opentelemetry/auto-instrumentations-node` which may break if we go beyond 4.4.4 with one of the packages, so this is the workaround for now.

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
pichlermarc: Interesting; AFAIK codecov uses the reports generated by istanbul, so adding the directive *should* help. Looks like the uncovered lines above are actually not tested at the moment ([see full file on the codecov report](https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/pull/1763/blob/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts) - not sure if you have access). We could also add an earlier fastify version to TAV if that fixes the issue. 

I've just recently added TAV to this package, having more versions tested would likely be a good idea anyway.

trentm: I've added testing of fastify@4.0.0 plus the current latest and any subsequent fastify 4.x releases to the .tav.yml.
I don't know if coverage data from TAV runs get included in the codecov summary.

trentm:
@pichlermarc The `TAV for PR / tav / Run test-all-versions (14)` job is failing for this PR currently because of some funkiness with the npm install layout:

```
@opentelemetry/instrumentation-fastify: -- installing ["fastify@4.24.3"]
@opentelemetry/instrumentation-fastify: -- running test "npm run test" with fastify (env: {})
@opentelemetry/instrumentation-fastify: > @opentelemetry/instrumentation-fastify@0.32.3 test /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-fastify
@opentelemetry/instrumentation-fastify: > nyc ts-mocha -p tsconfig.json 'test/**/*.test.ts'
@opentelemetry/instrumentation-fastify: Error: Cannot find module '@opentelemetry/contrib-test-utils'
@opentelemetry/instrumentation-fastify: Require stack:
@opentelemetry/instrumentation-fastify: - /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-fastify/test/instrumentation.test.ts
```

I was inclined to wait for https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1771 to go in first to get some improvements to the npm install handling. When that is merged, I'll update this PR and dig into the test failure if it still exists.
(I can make this a draft PR, if that helps.)

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
pichlermarc: Ah right, it looks like reports from TAV are not included. 

IMO just having the version tested is good enough for now, we may look into adding coverage from TAV in the future, but we've been having problems with codecov for a while now, so I think this will likely be a larger task.

pichlermarc (approved):
Looks good, thanks for fixing this :+1:

Issue #1762, opened 2023-10-30
[@opentelemetry/instrumentation-fastify] Breaks use of preClose with fastify


SeanReece:
### What version of OpenTelemetry are you using?
@opentelemetry/sdk-node@0.44.0
@opentelemetry/instrumentation-fastify@0.32.3

### What version of Node are you using?
v20.9.0

### What did you do?

When adding the hook `preClose` hook to fastify the application crashes when the `preClose` hook is called with this error:

```
TypeError: Cannot read properties of undefined (reading 'Symbol(opentelemetry.instrumentation.fastify.request_active_span)')
    at startSpan (/Users/me/src/test/node_modules/@opentelemetry/instrumentation-fastify/build/src/utils.js:30:24)
    at Object.<anonymous> (/Users/me/src/test/node_modules/@opentelemetry/instrumentation-fastify/build/src/instrumentation.js:74:48)
    at /Users/me/src/test/node_modules/fastify/lib/hooks.js:166:24
    at _encapsulateThreeParam (/Users/me/src/test/node_modules/avvio/boot.js:562:7)
    at Boot.timeoutCall (/Users/me/src/test/node_modules/avvio/boot.js:458:5)
    at Boot.callWithCbOrNextTick (/Users/me/src/test/node_modules/avvio/boot.js:440:19)
    at Object.push (/Users/me/src/test/node_modules/fastq/queue.js:110:14)
    at Boot.ready (/Users/me/src/test/node_modules/avvio/boot.js:336:18)
    at server.<computed> (/Users/me/src/test/node_modules/avvio/boot.js:59:21)
    at next (/Users/me/src/test/node_modules/fastify/lib/hooks.js:145:5)
```

`onClose` works as expected, but it appears that `preClose` needs to be added to the list of applicationHooks here: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1b0caa61972d969e3baea6a7db365e66dafe0c5d/plugins/node/opentelemetry-instrumentation-fastify/src/constants.ts#L21

### Reproducible Example

```js
const { NodeSDK } = require('@opentelemetry/sdk-node')
const { HttpInstrumentation } = require('@opentelemetry/instrumentation-http')
const { FastifyInstrumentation } = require('@opentelemetry/instrumentation-fastify')

const sdk = new NodeSDK({
  serviceName: 'testotel',
  instrumentations: [new HttpInstrumentation(), new FastifyInstrumentation()],
})

[...]

PR #1764, opened 2023-11-02
fix(instrumentation-fastify): do not wrap preClose and onRequestAbort hooks

trentm:
Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1762

This changes to an allow-list of hook names to wrap -- which means that if a new hook is added, then we won't create spans for it, but we also won't accidentally crash. :)


SeanReece (approved):
LGTM

david-luna (approved):

pichlermarc (approved):
This looks great, thank you for taking care of this. 👍 
I like the change to make this an allow list instead - makes it much safer in case new hooks are added in fastify.

PR #1768, opened 2023-11-02
fix(auto-instrumentations-web)!: Add zone.js as a peerDependency

### LINKED ISSUE #1767: zone.js should be listed as a peerDependency of auto-instrumentations-web
0.33.2

18.18.2

### What did you do?
I set up a reproduction repro here https://github.com/dbasilio/opentelemetry-peer-dep-repro to show that if you use strict peers in yarn, you cannot install auto-instrumentations-web. It also includes possible solutions to the problem.

You do not get any install errors when using yarn with strict peers

### What did you see instead?
Yarn prints an error and won't let your builds proceed

dbasilio:
I set up a Repro Repo to demonstrate the problem https://github.com/dbasilio/opentelemetry-peer-dep-repro

`auto-instrumentations-web` lists `instrumentation-user-interaction` as a dependency. `instrumentation-user-interaction` lists `zone.js@0.11.4` as a peerDependency, but `auto-instrumentations-web` doesn't satisfy that peerDepedency at all. It needs to be listed as a peerDependency or a dependency (or both) to satisfy the peerDependency from instrumentation-user-interaction`

- Fixes #1767

-  Adds zone.js@0.11.4 as a peerDependency to `auto-instrumentations-web`


blumamir (approved):
That makes sense to me 👍 
I wonder why the instrumentation's peer dependency in pinned and not caret 🤔, but that's a different topic

pichlermarc (approved):

PR #1771, opened 2023-11-03
chore: use npm workspaces and a package-lock.json

trentm:
- use npm workspaces instead of lerna for monorepo install handling (requires at least npm@7, so node v14 users will have to update their npm)
- updates to lerna@6
- add package-lock.json for more reliable dev and testing

Refs: https://github.com/open-telemetry/opentelemetry-js/pull/4238

# clean start in your git clones

The change from lerna to npm workspaces means that most dependencies will be
installed in the top-level `node_modules` dir. To avoid surprising conflicts
from `node_modules` and possible package-lock files in subdirs from local dev,
you should start fresh:

Either do a full `git clean -ffdx` ([it is the only way to be sure](https://www.youtube.com/watch?v=aCbfMkh940Q)) -- which will delete *all* non-git files -- or do this:

```
rm -rf node_modules \
  packages/*/{package-lock.json,node_modules} \
  plugins/node/*/{package-lock.json,node_modules} \
  plugins/node/*/examples/{package-lock.json,node_modules} \
  plugins/web/*/{package-lock.json,node_modules} \
  plugins/web/*/examples/{package-lock.json,node_modules} \
  propagators/*/{package-lock.json,node_modules} \
  detectors/node/*/{package-lock.json,node_modules} \
  metapackages/*/{package-lock.json,node_modules}
```

Then install and build, test and lint:

```
npm ci             # or 'npm install'
npm run compile

npm test
npm run lint
```

# smaller install, faster CI builds

One benefit is that size of a git clone after `npm ci` is down 10X -- from
~8.6GB to ~800MB. As a result, clean builds should be faster.

- Compare a recent "unit-test" CI run against "main": https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6670946547/usage
  ```
  browser-test (14) 35m 14s
  unit-test (14) 43m 49s
  unit-test (16) 16m 42s
  unit-test (18) 18m 57s
  ```
- To a recent build of the PR: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6791291162/usage
  ```
  browser-test (14) 14m 22s
  unit-test (14) 9m 7s
  unit-test (16) 9m 7s
  unit-test (18) 11m 35s
  ```

# reviewer notes

- s/prepare/prepublishOnly/g is to avoid running the compile step in packages before `npm install` for all packages has completed. It resulted in various build errors. Using `prepublishOnly` is recommended (https://docs.npmjs.com/cli/v10/using-npm/scripts#prepare-and-prepublish) and is what the core repo was already doing.
- I've updated lint.yml and release-please.yml to use node 18 (was node 14), so it has an npm which supports workspaces. I assume that is okay.
- I don't know how the release-please PRs are created. I haven't tested anything around that.


.github/workflows/lint.yml
```diff
@@ -21,26 +21,7 @@ jobs:
...
-            plugins/web/*/node_modules
-            propagators/*/node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/package.json') }}
-
-      - name: Bootstrap
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: |
-          npm install --only=dev --ignore-scripts
-          npx lerna bootstrap --no-ci --hoist --nohoist='zone.js' --nohoist='gts' --ignore-scripts -- --only=dev
+      # XXX consider 'npm install --only=dev' instead
```
david-luna, line 24: we should give it a try since it will improve in time and resources

david-luna (commented):
I guess you removed the caches to test the CI and not mess with the existing ones. I think we can prefix it so there is no collision with the existing ones.

trentm:
> I guess you removed the caches to test the CI and not mess with the existing ones.

No. `npm ci` deletes node_modules and does a clean re-install, so the caching doesn't help. I'm copying what the equiv PR for the core repo did. See cache discussion there: https://github.com/open-telemetry/opentelemetry-js/pull/4238#discussion_r1378753441

t2t2 *did* mention on that thread that we could enable npm cache on the `setup-node` action:

> You can however [enable npm cache in actions/setup-node](https://github.com/actions/setup-node#caching-global-packages-data) which will cache the local packages cache ([npm-cache](https://docs.npmjs.com/cli/v10/commands/npm-cache))

So that is worth trying out.

RichiCoder1:
Howdy! Saw this PR while browsing.

Unless you have a good reason not too, dropping `container.image` and using setup-node w/ `cache: npm` should net you a decent perf win. There's a non-zero overhead to using container images, and setup-node natively caches both node installs and npm caches now.

trentm:
> Unless you have a good reason not too, dropping `container.image` and using setup-node w/ `cache: npm` should net you a decent perf win.

@RichiCoder1 Thanks. I'll try that out on the .github/workflows/lint.yml a little later and compare job run times.
First I need to figure out this `unit-test (18)` test failure.

trentm:
The unit tests finally passed with node v18. This took longer than it should have.

tl;dr: The following assert was failing in Node v18 because the output from the ConsoleSpanExporter was marked-up with ANSI color escape codes, which broke the `stdout.includes("name: 'GET'")` check.
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1b0caa61972d969e3baea6a7db365e66dafe0c5d/metapackages/auto-instrumentations-node/test/register.test.ts#L38-L39
The output was colored because nx is by default adding the `FORCE_COLOR=true` environment variable.
Only Node v18 was affected because in v18.17.0, `FORCE_COLOR` started being honoured for non-TTY streams (stdout is not a TTY when run piped into anything, as in CI).
Setting `FORCE_COLOR=0` (or `false`), is the *only* way to prevent that with envvars.

Notes:
- It wasn't about adding Node adding color support for CI systems in https://github.com/nodejs/node/pull/40385. That is in recent node v16 versions as well. The Node change allowing `FORCE_COLOR` to ignore `isatty` is https://github.com/nodejs/node/pull/48034
- Setting node's `NODE_DISABLE_COLORS=1` or `NO_COLOR=1` doesn't work, because `FORCE_COLOR` wins.
- The nx code that is defaulting `FORCE_COLOR` is here: https://github.com/nrwl/nx/blob/15.9.7/packages/nx/src/tasks-runner/forked-process-task-runner.ts#L130-L140
- This failure doesn't show in any previous runs is because we were using lerna@5, which didn't use nx.

.github/workflows/lint.yml
trentm: Nope, can't use this:

```
% npm install --only=dev
npm WARN invalid config only="dev" set in command line options
npm WARN invalid config Must be one of: null, prod, production
...
```

Same for `npm ci --only=dev`.  I don't know if the pre-existing usage of `npm install --only=dev` was hopeful usage, or if it did work in earlier versions of npm.

trentm:
Could a maintainer please add one of the "pkg:..." labels, perhaps `pkg:instrumentation-ioredis`, to this PR, so I can sanity check that the "test-all-versions"-related GH Actions workflows are working? Perhaps @legendecas?

metapackages/auto-instrumentations-node/test/register.test.ts
trentm, line 1: Review note: I rewrote how this test spawns while debugging a breakage with node v18. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1771#issuecomment-1793248399 for details.

packages/opentelemetry-test-utils/src/test-utils.ts
```diff
@@ -141,11 +141,22 @@ export interface TimedEvent {
   name: string;
   /** The attributes of the event. */
   attributes?: SpanAttributes;
+  /** Count of attributes of the event that were dropped due to collection limits */
+  droppedAttributesCount?: number;
```
trentm, line 145: Reviewer note: This was added to fix test-all-versions tests of ioredis@4 -- a code path not tested with `npm test`. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1771/commits/78e64f1cf18ba38946eddc2ce8b5862d6b384b18

plugins/web/opentelemetry-instrumentation-long-task/tsconfig.esm.json
```diff
@@ -5,7 +5,7 @@
     "outDir": "build/esm",
     "tsBuildInfoFile": "build/esm/tsconfig.esm.tsbuildinfo"
   },
-  "files": [ "node_modules/zone.js/dist/zone.js.d.ts"],
+  "files": [ "../../../node_modules/zone.js/dist/zone.js.d.ts"],
```
legendecas, line 8: For reference: https://github.com/open-telemetry/opentelemetry-js/pull/4257. We can avoid hard-coding the node_modules path by type references.

lerna.json
```diff
@@ -1,15 +1,6 @@
 {
   "npmClient": "npm",
-  "packages": [
```
legendecas, line 3: I was not aware that we can remove `packages` in `lerna.json` when using workspaces. Would you mind removing it in https://github.com/open-telemetry/opentelemetry-js/ as well?

legendecas (commented):
Overall LGTM. Thank you for working on this!

plugins/web/opentelemetry-instrumentation-long-task/tsconfig.esm.json
trentm: Done in commit faed2143. Thanks.

trentm:
I may need to wait for updates from https://github.com/open-telemetry/opentelemetry-js/pull/4261 to settle. The package-lock.json generation just now got mixed 0.45.0 and 0.45.1 versions for various packages. Also, then the update to the new releases in opentelemetry-js-contrib will require a package-lock.json update here.

trentm:
# lockfileVersion

My understanding of lockfileVersion values in package-lock.json files.

https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#lockfileversion
> - 1: The lockfile version used by npm v5 and v6.
> - 2: The lockfile version used by npm v7 and v8. Backwards compatible to v1 lockfiles.
> - 3: The lockfile version used by npm v9. Backwards compatible to npm v7.

With the switch to npm workspaces we require at least npm v7. (This means node v14 users doing development in this repo need to update the default npm@6 that comes with that version of Node.js.) Requiring at least npm v7 means that we *could* switch to using `lockfileVersion: 3` in our package-lock.json file. Doing so would mean that the package-lock.json file is smaller by about 50% (the ["dependencies"](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#dependencies) block is removed).

We might want to be cautious, as `lockfileVersion: 3` is relatively new.
Renovate *does* support this version since v35.113.0 (https://github.com/renovatebot/renovate/pull/22281).

The core repo is currently using `lockfileVersion: 2`. Also, given that we support
development in this repo with Node.js 16 and its npm v8 (a version of npm that does not yet know about `lockfileVersion: 3`), my inclination is to stick with lockfileVersion 2 for now.
This is explicitly done via `lockfile-version=2` in ".npmrc".

trentm:
@legendecas or someone else: Could you please *remove* the `pkg:instrumentation-aws-sdk` label? This is resulting in all the TAV tests failing due to the already known https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1477 issue. I might look at that issue, but not in this PR.

Other than that, I think everything is basically working here now. I think we could merge this now, or wait for the renovate PR to update to the 0.45.1/1.18.1 release first.

pichlermarc (approved):
Wow, this must've been a chunk of work. Thank you so much for working on this @trentm. 
We've desperately needed this change for a while now, so I'm very happy to see this PR!

It's a huge change but it looks good to me, very high quality overall. Tested locally and everything seems to work for me. @dyladan would also appreciate your review as you're more familiar with the release automation in this repo. 🙂

plugins/node/opentelemetry-instrumentation-mongodb/examples/package.json
```diff
@@ -38,7 +38,7 @@
     "@opentelemetry/instrumentation-mongodb": "^0.32.0",
     "@opentelemetry/sdk-trace-node": "^1.0.0",
     "@opentelemetry/sdk-trace-base": "^1.0.0",
-    "mongodb": "^3.7.3"
+    "mongodb": "^3.6.11"
```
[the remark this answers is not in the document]
trentm: Yes, this was intentional at the time I made that commit at least. It was so that this range overlapped with the `"mongodb": "3.6.11",` in "plugins/node/opentelemetry-instrumentation-mongodb/package.json" one directory up.  At the time (earlier in the development of this PR) I had hit some npm install error.  I don't have the details of that install error anymore, sorry.

I could try to remove that change and re-test. (I might not get to that today.)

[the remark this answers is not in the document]
pichlermarc: No need to re-test, just tested myself as I was curious and it installs fine now with the previous range. 🙂 
I think we can keep it at what you have now. 👍

pichlermarc:
Waiting for one more CI run to finish. If there are no more comments until after the SIG meeting today (18:00 GMT) I'll merge this.

pichlermarc:
Ah, I think this still needs one more branch update and update to its package-lock.json as the most recently merged PR modified dependencies - then we can merge this.

trentm:
> I think this still needs one more branch update and update to its package-lock.json

I can do that now.

trentm:
Ready to merge now.

PR #1772, opened 2023-11-04
feat(auto-instrumentations-node): Expose getting resource detectors

### LINKED ISSUE #1531: Alternative to `auto-instrumentations-node/register` when needing to configure specific instrumentations

I'm currently using the `auto-instrumentations-node/register` module as it's quite convenient:

```sh
node -r @opentelemetry/auto-instrumentations-node/register dist/index
```

Now I'm in a situation where I need to customize the config for one of the instrumentations (in this case it's `@opentelemetry/instrumentation-fs`).

What's the best way to achieve this?

One solution that comes to mind is to take the [source code of `auto-instrumentations-node/register`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4e18a46396eb2f06e86790dbbd68075c4c2dc83b/metapackages/auto-instrumentations-node/src/register.ts) and customize the `getNodeAutoInstrumentations` call to my liking, like so:

```ts
import { DiagConsoleLogger, diag } from "@opentelemetry/api"
import { getNodeAutoInstrumentations } from "@opentelemetry/auto-instrumentations-node"
import { getResourceDetectorsFromEnv } from "@opentelemetry/auto-instrumentations-node/build/src/utils"
import * as opentelemetry from "@opentelemetry/sdk-node"

diag.setLogger(
  new DiagConsoleLogger(),
  opentelemetry.core.getEnv().OTEL_LOG_LEVEL
)

  instrumentations: getNodeAutoInstrumentations({
    "@opentelemetry/instrumentation-fs": { enabled: false },
  }),
  resourceDetectors: getResourceDetectorsFromEnv(),
})

try {
  sdk.start()
  diag.info("OpenTelemetry automatic instrumentation started successfully")
} catch (error) {
  diag.error(
    "Error initializing OpenTelemetry SDK. Your application is not instrumented and will not produce telemetry",
    error
  )
}

process.on("SIGTERM", () => {
  sdk
    .shutdown()
    .then(() => diag.debug("OpenTelemetry SDK terminated"))
    .catch((error) => diag.error("Error terminating OpenTelemetry SDK", error))
})
```

Here's the significant diff:

```diff
<   instrumentations: getNodeAutoInstrumentations(),
---
>   instrumentations: getNodeAutoInstrumentations({
>     "@opentelemetry/instrumentation-fs": { enabled: false },
>   }),
```

Unfortunately, this code is using the `getResourceDetectorsFromEnv` function, which is internal and not exported. I get the following error when running the code above:

```sh
$ node -r ./dist/tracing dist/index

node:internal/modules/cjs/loader:535
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './build/src/utils' is not defined by "exports" in /Users/user/project/dist/node_modules/@opentelemetry/auto-instrumentations-node/package.json
    at new NodeError (node:internal/errors:393:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:358:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:668:3)
    at resolveExports (node:internal/modules/cjs/loader:529:36)
    at Module._findPath (node:internal/modules/cjs/loader:569:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:981:27)
    at Module._load (node:internal/modules/cjs/loader:841:27)
    at Module.require (node:internal/modules/cjs/loader:1061:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/user/project/dist/tracing.js:28:17) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v18.12.1
```

After having a glance at the implementation of `getResourceDetectorsFromEnv`, it seems quite useful, so I wouldn't want to skip it.

Does it make sense to export `getResourceDetectorsFromEnv`? If yes, I can probably submit a PR with this change.

Zirak:
The auto instrumentation package provides users with a very easy way to
instrument their application, either from the commandline or
programatically. Users wishing the configure these instrumentations can also
call `getNodeAutoInstrumentations`.

This commit also exposes `getResourceDetectorsFromEnv`, so that users
configuring or somehow wishing to change the easy to use autoinstrumentation
entrypoint can also use the preconfigured resource detectors. Since the
autoinstrumentation package sets its exports, this function was previously
inaccessible.

Usage can look like this:

```js
const { getNodeAutoInstrumentations, getResourceDetectorsFromEnv } = require('@opentelemetry/auto-instrumentations-node');
const opentelemetry = require('@opentelemetry/sdk-node');

  instrumentations: getNodeAutoInstrumentations({
    '@opentelemetry/instrumentation-fs': { enabled: false },
  }),
  resourceDetectors: getResourceDetectorsFromEnv(),
});

// use the sdk
```

Workarounds can be found [elsewhere in the opentelemetry repos](https://github.com/open-telemetry/opentelemetry-demo/blob/855bde3588b0fe85e500bec185dc2f311b15f98a/src/paymentservice/opentelemetry.js#L28-L38).
These can be fixed once a release is made.

Resolves #1531.


blumamir (approved):

blumamir:
@Zirak 
Can you please fix the lint errors so I can merge the PR?

```
/__w/opentelemetry-js-contrib/opentelemetry-js-contrib/metapackages/auto-instrumentations-node/src/index.ts
  18:3  error  Delete `··`  prettier/prettier
  19:1  error  Delete `··`  prettier/prettier
  20:3  error  Delete `··`  prettier/prettier
  21:1  error  Delete `··`  prettier/prettier
```

You can run `npm run lint:fix` in the package and commit the changes.
Thanks!

Flarna:
Likely to late now....

But the exposed function doesn't have a nice name. The `FromEnv` indicates that it holds only the resource detectors specified via environment variables.
Once we add other config options like files,... we either have to
* export more APIs (`FromFile`, `FromXXX`,... variants)
* rename the export to `getResourceDetectors()` if it should cover the final list

Zirak:
That's a good point. It doesn't seem like a release has been done so it's still fixable, I like the `getResourceDetectors` since it seems reasonable for `getNodeAutoInstrumentations` to in the future also respect environment variables, like [the java agent does](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#enable-only-specific-instrumentation). That'll make the two in line both in functionality and semantics.

@blumamir I'll make a follow-up PR, feel free to close it if it feels irrelevant.

Issue #1756, opened 2023-10-27
`@opentelemetry/instrumentation-document-load` plugin does not work, if `@opentelemetry/instrumentation` esnext used


maximelkin:
### What version of OpenTelemetry are you using?
```
"@opentelemetry/instrumentation-document-load": "0.33.2"
"@opentelemetry/instrumentation": "0.43.0"
```

### What version of Node are you using?
v18.12.1

### What did you do?
1) override `@openetelemetry/instrumentation` to use esnext folder (we do it via webpack, it's annoying to list all packages, but it's working). `@opentelemetry/instrumentation-document-load` not have esnext folder, do nothing (use esm folder). 
2) 
```
import { DocumentLoadInstrumentation } from '@opentelemetry/instrumentation-document-load';

new DocumentLoadInstrumentation()
```

### What did you expect to see?
no error

### What did you see instead?
`DocumentLoadInstrumentation` constructor fails at 
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts#L62
Because `DocumentLoadInstrumentation` is transpiled, super constructor call transpiled to function call, but it does fail, because `InstrumentationBase` class not transpiled, so it only can be called via new

dyladan:
@pkanal and @martinkuba are the code owners here. Can you please take a look and see what is going on here?

@maximelkin do you have an idea how this can be resolved?

maximelkin:
For me, adding `build/esnext/`, as for packages in https://github.com/open-telemetry/opentelemetry-js will work.

PR #1776, opened 2023-11-07
feat: add esnext target for web instrumentations

martinkuba:
Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1756

This adds `esnext` build target for all web instrumentations. Currently, only `esm` is published, which is incompatible with the `esnext` code from core libraries when subclassing (see the linked issue above). This should also reduce bundle size when targeting newer browsers (no polyfills).


legendecas (approved):

trentm (approved):

pichlermarc (approved):
Looks good, thank you for working on this 🙂

PR #1777, opened 2023-11-07
feat: additional instrumentation debug diagnostics

seemk:

A few instrumentations were missing debug messages when patching was done, making it difficult to debug these in the wild.

Additionally I think `graphql` instrumentation was checking for the wrong method names when running unwrap (`validate` and `parse`), changed these to the correct ones.


blumamir:
I think in instrumentations we are trying to use `this._diag` where possible as it is a component logger that includes the instrumentation library name.

seemk:
> I think in instrumentations we are trying to use `this._diag` where possible as it is a component logger that includes the instrumentation library name.

Thanks, replaced the calls with `this._diag`.

blumamir (approved):

PR #1786, opened 2023-11-09
chore: cleanup files in package json

### LINKED ISSUE #820: Set `files` field in all of our `package.json` files
`files` field defines which files ought to be included in the final published package.

- [ ] set `files` for all the `package.json` files. Don't include [files always included](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files), remove non-existing `doc`, TS source files. Basically only `.js`, `.js.map` and `.d.ts` files in the build folder need to be included,
- [ ] remove `.npmignore` files since it doubles `files` as an blacklist.

#### Motivation:

- Users may experience issues with `tsconfig.json` being included: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/733
- Package sizes could be a lot smaller - excluding `build/tsconfig.tsbuildinfo` would halve the package size for `@opentelemetry/auto-instrumentations-node`

macno:

Fixes #820 

* remove useless entries from `files` in `package.json`
* add `files` in `package.json` where missing
* remove `.npmignore`
* fix typo in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ You will receive the following error :
 ✖   type must be one of [ci, feat, fix, docs, style, refactor, perf, test, revert, chore] [type-enum]
 ```
 
-Here an exemple that will pass the verification: `git commit -s -am "chore(opentelemetry-core): update deps"`
+Here an example that will pass the verification: `git commit -s -am "chore(opentelemetry-core): update deps"`
 
 ### Fork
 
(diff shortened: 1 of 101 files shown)


blumamir (approved):
Thanks for this cleanup! 💯

macno:
> Why remove the `.npmignore` this stops unnecessary files getting included into the npm packages which just fill up the consumers disk space for no reason.

Hi @MSNev, `.npmignore` is not necessary since we whitelist what to include using the `files` property in the `package.json`

PR #1796, opened 2023-11-13
chore(graphql): update span name to show the resolver name

### LINKED ISSUE #1791: Add path to span name of `graphq.resolve` spans created by graphql instrumentation

Currently, the `graphql` instrumentation packages adds spans for all resolvers where span name equals to [`graphql.resolve`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/2698bb1a5a4de5a5c6272643d6e50180db874d64/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts#L132). Here's how it looks in datadog for example:

Screenshot 2023-11-10 at 13 29 46

However, to have improved DX, the span name could contain the actual path.

 Something like `graphl.resolve - <path>` as the span name would make this much better:

- Should this be the default behavior?
- Should this be under an option?
- Should the instrumentation offer a hook for every span added, so we could change it, including changing span name or adding attributes?

The most obvious workound is to add a processor that changes the span name:

```js
if (span.name === 'graphql.resolve' && span.attributes['graphql.field.path']) {
  span.updateName(`${span.name} ${span.attributes['graphql.field.path']}`);
}
```

diogotorres97:
Fixes #1791 
-

This PR updates (opentelemetry-instrumentation-graphql) the resolvers' span name. 
It will show something like: `graphql-resolve hello` instead of `graphql-resolve`. This is very useful to easily understand at a top level which field is being resolved.

E.g.,
```
query ExampleQuery {
  me {
    name 
    }
}
```

Screenshot 2023-11-13 at 11 25 18

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
@@ -129,7 +129,7 @@ function createResolverSpan(
   };
 
   const span = tracer.startSpan(
-    SpanNames.RESOLVE,
+    `${SpanNames.RESOLVE} ${attributes[AttributeNames.FIELD_PATH]}`,
     {
       attributes,
     },
--- a/plugins/node/opentelemetry-instrumentation-graphql/test/graphql.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-graphql/test/graphql.test.ts
@@ -613,7 +613,7 @@ describe('graphql', () => {
       await graphql({ schema: simpleSchemaWithResolver, source: '{ hello }' });
       const resovleSpans = exporter
         .getFinishedSpans()
-        .filter(span => span.name === SpanNames.RESOLVE);
+        .filter(span => span.name === `${SpanNames.RESOLVE} hello`);
       assert.deepStrictEqual(resovleSpans.length, 1);
       const resolveSpan = resovleSpans[0];
       assert(resolveSpan.attributes[AttributeNames.FIELD_PATH] === 'hello');
@@ -633,7 +633,7 @@ describe('graphql', () => {
       await graphql({ schema, source: '{ hello }', rootValue });
       const resovleSpans = exporter
         .getFinishedSpans()
-        .filter(span => span.name === SpanNames.RESOLVE);
+        .filter(span => span.name === `${SpanNames.RESOLVE} hello`);
       assert.deepStrictEqual(resovleSpans.length, 1);
       const resolveSpan = resovleSpans[0];
       assert(resolveSpan.attributes[AttributeNames.FIELD_PATH] === 'hello');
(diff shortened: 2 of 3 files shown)


satazor:
Not sure who to ping here for visibility, cc @blumamir

diogotorres97:
Rebased and fixed linter issues

dyladan:
Looks like there's no code owner for these files. @open-telemetry/javascript-approvers please take a look if you have time

Flarna (changes requested):
To my understanding this doesn't fit the [graphql semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/graphql.md).

I guess first step here should be a PR on the semantic-conventions repo to get consensus if this is wanted or not.

diogotorres97:
> To my understanding this doesn't fit the [graphql semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/graphql.md).
> 
> I guess first step here should be a PR on the semantic-conventions repo to get consensus if this is wanted or not.

I didn't understand why it doesn't fit, here we are updating the resolver part (inside of a query or mutation). That link refers to `graphql.operation.type`: query, mutation and subscription. But this is something more internal and not a graphql operation.

dyladan:
It looks like there is no resolver in the semconv. It might be worth opening a semconv issue or PR to cover this area to ensure all languages are handling it the same way. Do we know if other languages have instrumentations for graphql and how they're handling this?

Flarna:
Ah ok sorry. Seems I lack graphql knowledge.
As long as the names fit to the [general span requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span) it should be fine I think.

But I agree with @dyladan that a spec improvement might make sense if this is something other languages have also to keep consistency.

Flarna (approved):

dyladan (approved):
>As long as the names fit to the [general span requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span) it should be fine I think.

I agree. I'm not a graphql expert but I _think_ this fits the requirement to be limited cardinality and should be fine. In any case it can't be higher cardinality than the attribute which should also be limited.

jcarres-mdsol:
IMHO it does not fit the requirement of limited cardinality.
This graph is showing the cardinality of the operation name, the orange is our graphql in `develop` where we updated this library 
Screenshot 2024-03-07 at 11 48 46 AM

Moving this to production would make this much worse. I think any organization with a lot going on with graphql will suffer from this.

satazor:
@jcarres-mdsol I have to disagree. The document you linked lists examples of good and bad values in regards to cardinality. With this PR, the span name is no longer a generic name, increasing cardinality but it doesn’t include any dynamic identifier, such as a user id, which would violate the rules set in the document 

Now span names in graphql resolvers are identifiable and we can make aggregations on which ones are slower etc, mulch like you have different span names for different functions etc.

If this raises concerns about performance or other issues, perhaps this could be under an option.

jcarres-mdsol:
We have solved this with a regex in the collector.

The original operation names are something like x.y.z.a.b , it is true it is not infinite like if they were UUIDs or something but still 10s/100s of 1000s. So we have regex this to x.y this is enough in our case to identify it enough, also the whole path is keep in an attribute anyways.

Issue #1770, opened 2023-11-03
Add ability to add custom span attributes on user interaction traces


adamlyka:
The `UserInteractionInstrumentationConfig` does not allow you to add custom span attributes on events as the `instrumentation-fetch` and `instrumentation-document-load` do. 
It would be good to have the same `applyCustomAttributesOnSpan` added to the configuration of the `instrumentation-user-interaction` package. I am currently adding a unique id we assign per user to all traces to have a complete view of their user journeys across the frontend, and right now the auto instrumentation only produces a series of unassociated user events (some will show the fetch events if the user clicks on a button which triggers an api call)

pkanal:
👋🏽 Hey there, So this is actually currently possible but probably needs to be documented better. You can do it via the [`shouldPreventSpanCreation` option](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-instrumentation-user-interaction/src/types.ts#L39) if you pass it a function. I can open a README PR to document this better.

pkanal:
> I think using this callback for modifying spans is counter-intuitive, especially since the callback `applyCustomAttributesOnSpan` is now a pattern in multiple packages. I personally would prefer to add `applyCustomAttributesOnSpan` callback to this instrumentation, and use `shouldPreventSpanCreation` only for what its name suggests.
> 
> With that said, this has been discussed [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/653#discussion_r702030818), and maybe we do not want to reverse that decision at this point. Curious to hear what others think.

I think following the pattern of `applyCustomAttributesOnSpan` would be my preference since I think `shouldPreventSpanCreation` is pretty confusing for folks.

ccschmitz:
Another issue with using `shouldPreventSpanCreation` is that you can't access the original event. We want to append `clientX` and `clientY` coordinates to click events but aren't able to because we only get the event name and not the actual event. Would be great to have an option for applying custom attributes that gave you access to the original event.

PR #1803, opened 2023-11-16
feat(@opentelemetry/instrumentation-user-interaction): Add custom span attributes

adamlyka:
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1770

Added the ability to add custom attributes on a span
I have set it up to be configured with a `UserInteractionCustomAttributeFunction` which is in the same fashion as the other plugins for consistency.


pichlermarc:
Closing this as per @pkanal 's comment. Please let us know in a new issue if the option documented in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1840 does not work for you.

Issue #1787, opened 2023-11-09
Can the runtime dependency on types be moved to devDependencies


lilwilsond2:
### What version of OpenTelemetry are you using?
1.6.0

### What version of Node are you using?
18.17.0

### What did you do?

The inclusion of @types/express as a runtime dependency causes some typing issues for us in mocked tests.
I found [this discussion](https://github.com/lightstep/otel-launcher-node/pull/42) about why the types were moved originally, but i was wondering if the types could be moved back at this point

### What did you expect to see?
type defs as a dev dependency

### What did you see instead?
type defs as a runtime dependency

### Additional context

fwiw, i added this to my package json and things still seem okay
```json
"resolutions": {
    "@types/express": "link:./node_modules/.cache/null"
  },
```

Flarna:
The problem is that `ExpressInstrumentation` has types from express in it's public interface.
e.g. the `requestHook` uses type `Request`.

If some typescript user installs the instrumentation (maybe indirect via autoinstrumentations meta package) but not the express types typescript complains.

Therefore to move it into dev-deps it's additionally needed to cleanup the public API surface as described [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md).

david-luna:
@Flarna 

The description of `requestHook` says

```
Function for adding custom attributes on Express request.
```

Could we limit the `request` type to a subset of props (the ones necessary to add attributes) and allowing the user to pass the full type if needed?

```typescript
export interface SimpleExpressRequest {
  // some props here
}

export type ExpressRequestInfo<T = SimpleExpressRequest> = {
  request: T;
  route: string;
  layerType: ExpressLayerType;
};

```

Flarna:
As long as the dependency to express is broken it should be fine. The instrumentation itself doesn't need the hooks therefore I think no need for a `SimpleExpressRequest` types. We could default to `unknown` or `any`.

But for sure the docs need to mention what is actually given to the hook.

david-luna:
@Flarna @pichlermarc  I can prepare a PR. you can assign it to me

PR #1804, opened 2023-11-16
fix(instrumentation-express)!: remove `@types/express` from dependencies

david-luna:

`@types/express` is set as a dependency of express instrumentation therefore it is installed in any project using this package. This is causing compatibility issues.

- moved `@types/express` to devDependencies
- made `ExpressRequestInfo` generic so it can be typed by consumers
- update README & references to config options
- move from deprecated `SpanAttributes` type to `Attributes`

Closes: #1787


plugins/node/opentelemetry-instrumentation-express/src/types.ts
```diff
@@ -14,15 +14,15 @@
...
-import type { Request } from 'express';
 import { Span } from '@opentelemetry/api';
 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
 import { ExpressLayerType } from './enums/ExpressLayerType';
 
 export type IgnoreMatcher = string | RegExp | ((name: string) => boolean);
 
-export type ExpressRequestInfo = {
-  request: Request;
+export type ExpressRequestInfo<T = any> = {
```
david-luna, line 23: Note for reviewer: type argument `T` defaults to `any` so anyone already using `requestHook` and `spanNameHook` will still work at the type level. The `unknown` type was considered but it will trigger typing errors within these hooks.

david-luna:
@pkanal @JamieDanielson could you have a look please?

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -14,7 +14,22 @@
  * limitations under the License.
  */
 
-import { SpanAttributes } from '@opentelemetry/api';
+/*
```
trentm, line 17: license block added twice?

plugins/node/opentelemetry-instrumentation-express/README.md
```diff
@@ -79,6 +79,13 @@ Express instrumentation has few options available to choose from. You can set th
 - `info: ExpressRequestInfo` containing the incoming Express.js request, the current route handler creating a span and `ExpressLayerType` - the type of the handling layer.
 - `defaultName: string` - original name proposed by the instrumentation.
 
+`requestHook` is invoked with 2 arguments:
+
+- `span: Span` - the span associated with the express request.
+- `info: ExpressRequestInfo` containing the incoming Express.js request, the current route handler creating a span and `ExpressLayerType` - the type of the handling layer.
+
+NOTE: `ExpressRequestInfo.request` is typed as `any`. If you want type support make sure you have `@types/express` installed then you can use `ExpressRequestInfo<express.Request>`
```
blumamir, line 87: Just to make sure I got it right - so a user looking for hook typing, can use:

```js
import type { Request } from 'express';
import { ExpressRequestInfo } from '@opentelemetry/instrumentation-express';

const spanNameHook = (info: ExpressRequestInfo<express.Request>, defaultName: string): string => {
//....
}
```

Right?

blumamir (approved):
This looks like a great workaround for an annoying problem. I guess it can be adopted in other instrumentations as well?

If you have some free time, it could be awesome to also document this in the [guidelines](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#adding-types-in-public-api)

plugins/node/opentelemetry-instrumentation-express/README.md
david-luna: that's correct @blumamir :)

by doing so devs will have the full type with the current version of `@types/express` they have installed in their app. Otherwise the property is typed as `any` which will not harm TS compilation.

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
david-luna: yes, removing it now

JamieDanielson (commented):
Thanks for this! I've left a question to ensure I'm understanding the change properly, but it seems like the right move here and probably other packages as Amir mentioned as well.

I'll want to update the example (as I'm using it I'm really becoming aware of how outdated it is 😅) to show these changes, but can do as a separate PR once this lands).

plugins/node/opentelemetry-instrumentation-express/src/types.ts
david-luna: Theree is a small detail for existing TS users. The type of request property will become of type any. Therefore TS will not provide autocompletion and type checks for it and its sub-properties. This may not be nice to existing users.

To mitigate that we may create our own default type. I've already tried to copy express' request type but there are many links to other types making it a big chunk of code just for typing.

Maybe we could agree on a subset of the request api but this come with the risk of breaking the build of existing TS users of they access a property not defined on this default type.

david-luna:
> This looks like a great workaround for an annoying problem. I guess it can be adopted in other instrumentations as well?
> 
> If you have some free time, it could be awesome to also document this in the [guidelines](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#adding-types-in-public-api)

We can bring this to the next SIG and if there is consensus I can start right away :)

JamieDanielson:
🤔 I'm still having a tough time with using `any` here, and also see what you mean about the problems we run into with trying our own subset of the Request type. 

[This is an older thread](https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051345) but I see it referenced in various places and agree with the sentiment:

>Given that breaking consumers is a worse problem than slightly-larger packages, we've made `--save` the default in our documentation.

Much of this conversation (and linked issues/PRs in other libraries) is around the idea that although types packages are often dev dependencies, these types are explicitly used to interact with this library and so need to be available to consumers of the library.

What if we tried moving this to an optional peer dependency ([`peerDependenciesMeta`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta)) for `@types/express`? TypeScript users do need this package anyway, so this makes it more explicit.

Flarna:
> TypeScript users do need this package anyway

That's only true if one uses express. The instrumentation is part of the auto-instrument package therefore users get it even they don't use express. It would be strange to force them to install `@types/express`.

Regarding a general guide:
For `express` it's a dedicated `@types` package as of now which is quite small on install, no runtime size and usually no further dependencies. For other packages (e.g. `@grpc/grpc-js`, `fastify`) types are included in the package itself. In this case it would be needed to depend on the package itself. In special if several major versions are supported by the instrumentation it's hard to find a single one which offers the types needed across the whole range.

Therefore the usual rule is to avoid such dependencies by not using any types of the instrumentee in the instrumentation.

Flarna:
auto instrumentation package follow the typical APM approach to "just install the tool" and it takes care of everything. on default the only config needed should be exporter details. If defaults of all instrumentations and SDK are choosen reasonable (maybe even dependent on environment like AWS lambda vs normal host,...) only a few power user should have the need to use hooks.
so far the ideal world :o)

> I don't know that I agree with making an already bloated package slightly smaller for users who don't benefit from these types anyway, while making manual instrumentation/setup harder for users who do benefit from these types.

It's not that much about size (at least not in the case where it is about a types only package).
It's not the instrumentation which should select the @types version - it's the end user along with the express version they use.

We could keep it in dependencies and use `*` as version to match anything and choose latest on default. But that cries for autobreaking of the CI here. Maybe not that a hot potato for express as API is quite stable but I doubt `*` is a good idea as general rule.

I guess once express decides to include types it's clear that we don't want it as dependency. So removing types could be also seen as preparation for this.

> The ideal solution for the auto-instrument package is to make it install only those instrumentations that are necessary

I don't think this is supported by NPM.

JamieDanielson:
I have now read through so many issues and PRs about this (mostly linked from that [types-publisher issue](https://github.com/microsoft/types-publisher/issues/81)) and... phew 😅 . This is a hotly debated topic, and it doesn't seem like there's a definitive right answer that really works for everyone. 

> It's not the instrumentation which should select the types version - it's the end user along with the express version they use.

I totally agree with this. 

> We could keep it in dependencies and use * as version to match anything and choose latest on default. But that cries for autobreaking of the CI here. Maybe not that a hot potato for express as API is quite stable but I doubt * is a good idea as general rule.

I've also considered a larger range for version matching, though yeah `*` does seem excessive. 

🤔  So as I understand it, the general guidance seems to be to include the types as a dependency only if it's exported, which `ExpressRequestInfo` is. This is why we had to make the `ExpressRequestInfo` generic, and consumers of this package that directly use the `ExpressRequestInfo` can type their own, as Amir showed in a [previous comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1804#pullrequestreview-1755846788). Those without types (not installed or using JS) will lose autocompletion and IDE help for this specific functionality, but that is not as much due to the instrumentation as it is due to express not having built-in types.

This may be the right move. I'm not sure how easily it will be implemented in other instrumentations as I'm not sure how intertwined those types are. But if this helps reduce fragility that seems good too. Also PS I'm sorry I'm seeing this on the PR instead of helping with the discussion earlier on the issue.

JamieDanielson (approved):

trentm (approved):

david-luna:
> > > This looks like a great workaround for an annoying problem. I guess it can be adopted in other instrumentations as well?
> > > If you have some free time, it could be awesome to also document this in the [guidelines](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#adding-types-in-public-api)
> > 
> > 
> > We can bring this to the next SIG and if there is consensus I can start right away :)
> 
> @david-luna General consensus in the SIG meeting today was that this should be documented and attempted to be adopted in other instrumentations 👍 cc @blumamir

Thanks @JamieDanielson I'll prepare a PR to add this approach to the docs :)

Issue #1806, opened 2023-11-16
handling package-lock.json updates


trentm:
@pichlermarc I'm not sure what the right thing to do for package-lock.json updates is with the following.

#1771 went in, adding the package-lock.json file.

Then, the https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1723 renovate PR was just merged. Looking at just a part of that change:

```diff
diff --git a/package-lock.json b/package-lock.json
index 70e9e172..4464f673 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -30606,23 +30607,23 @@
     "node_modules/systeminformation": {
-      "version": "5.21.15",
-      "resolved": "https://registry.npmjs.org/systeminformation/-/systeminformation-5.21.15.tgz",
-      "integrity": "sha512-vMLwsGgJZW6GvoBXVWNZuRQG0MPxlfQnIIIY9ZxoogWftUpJ9C33qD+32e1meFlXuWpN0moNApPFLpbsSi4OaQ==",
+      "version": "5.21.17",
+      "resolved": "https://registry.npmjs.org/systeminformation/-/systeminformation-5.21.17.tgz",
+      "integrity": "sha512-JZYRCbIjk3WuBV59A9/rTla2rROX+aAJ9uo2Z1dI+bjieORcukClN8rlM1zE9NYKpULSbaGc+KKct/870lO0DA==",

@@ -34355,21 +34356,21 @@
     "packages/opentelemetry-host-metrics": {
       "name": "@opentelemetry/host-metrics",
       "version": "0.33.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@opentelemetry/sdk-metrics": "^1.8.0",
-        "systeminformation": "^5.0.0"
+        "systeminformation": "^5.21.17"
       },
       "devDependencies": {
...
```

1. It updates the actually installed version of `systeminformation` from `5.21.15` to `5.21.17`. That's well and good.
2. The second hunk, however, looks potentially problematic. It updates the `dependencies["systeminformation"]` value for what I had understood was meant to be a cache of the value in "packages/opentelemetry-host-metrics/package.json".  However the value in those two files are no longer the same thing. (My understanding comes from the somewhat loose docs of the package-lock.json format here -- https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#packages -- and from the following commands.)

If one gets the latest and tries to regenerate the package-lock.json file:

```
[...]

PR #1807, opened 2023-11-16
chore: sync package-lock file to package.json files

trentm:
This PR was generated from running `npm install --package-lock-only`. It undoes part of the change by renovate in #1723.
See discussion at https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1806


pichlermarc (approved):
Thanks for opening this PR (feel free to put it out of draft anytime to get it merged) - looks like we may actually need a non-existent (?) renovate feature that would allow us to group in lock file maintainance with update PRs.

I'll put some more thoughts on the issue #1806

Issue #1609, opened 2023-07-21
AWS S3 behavior changes when using the auto instrumentation


zdelagrange:
### What version of OpenTelemetry are you using?
```   
"@opentelemetry/auto-instrumentations-node": "^0.37.1",
"@opentelemetry/sdk-node": "^0.40.0",
```

### What version of Node are you using?
v16.20.0

### What did you do?
When using the recommended nodejs autoinstrumentation set up found here: https://opentelemetry.io/docs/instrumentation/js/getting-started/nodejs/#setup, While having a global S3 configuration, the autoinstrumentation causes an error to be thrown when the AWS_REGION is set to a region that is not us-east-1 or aws-global. The requests will succeed when the instrumentation.js is not required.

Reproduction here: https://github.com/zdelagrange/otel-nodejs-aws-bug

### What did you expect to see?
I expected the S3 functionality to not change when using the auto instrumentation

### What did you see instead?
S3 functionality changed when using the auto instrumentation causing a 403/failure to be returned when a success returns without using the auto instrumentation:

```
SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.
    at Request.extractError (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/services/s3.js:711:35)
    at Request.callListeners (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/zdelagrange/src/otel-nodejs-aws-bug/node_modules/aws-sdk/lib/request.js:38:9)
[...]

PR #1813, opened 2023-11-18
fix(instrumentation-aws-sdk): Add traceparent to AWS SDK unsignableHeaders

martinkuba:

This is an attempt to address the issue of a signature mismatch when a request fails and is retried. This happens because the instrumentation adds the `traceparent` header to outgoing headers. When a request fails, the value of the header changes, but the signature does not; and this results in an error.

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1609 and https://github.com/open-telemetry/opentelemetry-js/issues/3922

This change adds the `traceparent` header to the list of unsignable headers, so that it is excluded from creating the signature.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
@@ -222,6 +222,12 @@ export class AwsInstrumentation extends InstrumentationBase<any> {
       this._getRequestPromisePatch.bind(this, moduleVersion)
     );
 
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
Looks good! I like this solution.
@blumamir would you mind also taking a look to ensure I'm not missing anything :slightly_smiling_face:

trentm:
I mentioned a possible alternative fix at https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1609#issuecomment-1826167348

trentm:
I think this can be closed now that #1609 is in.

PR #1798, opened 2023-11-13
chore: release main

dyladan:
:robot: I have created a release *beep* *boop*
---

<details><summary>@opentelemetry/auto-instrumentations-node: 0.40.1</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/instrumentation-amqplib bumped from ^0.33.3 to ^0.33.4
    * @opentelemetry/instrumentation-aws-sdk bumped from ^0.36.2 to ^0.37.0
    * @opentelemetry/instrumentation-cassandra-driver bumped from ^0.34.0 to ^0.34.1
    * @opentelemetry/instrumentation-ioredis bumped from ^0.35.3 to ^0.36.0
    * @opentelemetry/instrumentation-lru-memoizer bumped from ^0.33.3 to ^0.33.4
    * @opentelemetry/instrumentation-memcached bumped from ^0.32.3 to ^0.32.4
    * @opentelemetry/instrumentation-mongodb bumped from ^0.37.2 to ^0.37.3
    * @opentelemetry/instrumentation-mongoose bumped from ^0.33.3 to ^0.33.4
    * @opentelemetry/instrumentation-mysql bumped from ^0.34.3 to ^0.34.4
    * @opentelemetry/instrumentation-mysql2 bumped from ^0.34.3 to ^0.34.4
    * @opentelemetry/instrumentation-pg bumped from ^0.37.0 to ^0.37.1
    * @opentelemetry/instrumentation-redis bumped from ^0.35.3 to ^0.35.4
    * @opentelemetry/instrumentation-redis-4 bumped from ^0.35.4 to ^0.35.5
    * @opentelemetry/instrumentation-socket.io bumped from ^0.34.3 to ^0.34.4
    * @opentelemetry/instrumentation-tedious bumped from ^0.6.3 to ^0.6.4
    * @opentelemetry/resource-detector-alibaba-cloud bumped from ^0.28.3 to ^0.28.4
    * @opentelemetry/resource-detector-aws bumped from ^1.3.3 to ^1.3.4
    * @opentelemetry/resource-detector-container bumped from ^0.3.3 to ^0.3.4
    * @opentelemetry/resource-detector-gcp bumped from ^0.29.3 to ^0.29.4
</details>

<details><summary>contrib-test-utils: 0.35.0</summary>

### Features

[...]

PR #1817, opened 2023-11-21
fix(ci): use opentelemetrybot to sync package-lock.json on release PRs

pichlermarc:

- see #1798, release-please is not updating package-lock.json with the new versions, so currently we'd need to do this manually - however release please force-pushes on each merge to main, so this is tedious.
- we currently do not use the opentelemetrybot account to create release PRs
  - [link to community repo issue to get the required access to @opentelemetrybot](https://github.com/open-telemetry/community/issues/1814)

- uses opentelemetrybot credentials to create release PRs via release please
- adds a commit as opentelemetrybot, which syncs the package-lock.json (@opentelemetrybot has signed the CLA)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -20,17 +20,51 @@ jobs:
           node-version: 18
           registry-url: 'https://registry.npmjs.org'
 
-      - name: Build Packages
+      - name: Install packages
         run: |
           npm ci
-          npm run compile
 
       - uses: google-github-actions/release-please-action@v3
         id: release
         with:
           command: manifest
-          token: ${{secrets.RELEASE_PR_TOKEN}}
+          token: ${{secrets.OPENTELEMETRYBOT_GITHUB_TOKEN}}
           default-branch: main
+
+      # get release PR as we're currently on main
+      - name: Checkout release PR
+        # only update if a PR has been created, otherwise this will fail
+        if: ${{ steps.release.outputs.pr }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -20,18 +20,52 @@ jobs:
...
+        run: |
+          npm install --ignore-scripts --package-lock-only
+          git add package-lock.json
+          git config user.name opentelemetrybot
+          git config user.email 107717825+opentelemetrybot@users.noreply.github.com
+          git commit -m "chore: sync package-lock.json"
+          git push
+
+      # get main again
+      - name: Checkout Repository
```
david-luna, line 59: Suggestion: if you want to speed up the process you could avoid checking out again by cleaning up the workspace just after pushing you could add the following commands

```bash
git clean -xdf
npm ci
npm run compile
```

david-luna (approved):

.github/workflows/release-please.yml
pichlermarc: Ah, yes we'd also have to check out main though as we might be on another branch at this point in the workflow.

It's also relatively quick (about 2s) compared to the rest of the workflow so I'd opt to keep it as-is (here's an example run https://github.com/pichlermarc-sample-organization/test-release-please/actions/runs/6941681633/job/18883019651). 🙂

trentm (approved):

PR #1819, opened 2023-11-23
fix(instrumentation-net): Don't operate on closed span

### LINKED ISSUE #1775: tls instrumentation in `@opentelemetry/instrumentation-net` tries to operate on closed span

- `@opentelemetry/auto-instrumentations-node@0.39.4`
- `@opentelemetry/sdk-node@0.44.0`

Reproduced on:
- 18.18.2
- 20.9.0

### What did you do?

When instrumented code uses `fetch` in a non-idiomatic manner, the tls instrumentation attempts to do an action on a closed `tls.connect` span.

For example, instrumenting the following code:

```js
fetch('https://example.com').then(() => {
    // A
    console.log('got a response');
});

// let the fetch timeout, hitting B
setTimeout(() => {}, 10_000);
```

The `fetch` timeouts since the response is not properly handled.

Point `A` is called when the `CONNECT` event is emitted, handled inside otel here: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/de6156aea1db7a7a018ad34f08cfc9f7ff7752b8/plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts#L122-L143

Point `B` is called when the `ERROR` event is emitted, handled here: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/de6156aea1db7a7a018ad34f08cfc9f7ff7752b8/plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts#L145-L151

First `A` is emitted, setting some attributes and `end`ing the span. Then `B` is hit, which attempts to set a status and close the span. That's invalid, since the span has already ended. With `OTEL_LOG_LEVEL` set to `info`, it prints something like the following:

```
Can not execute the operation on ended Span {traceId: 05b3d14d94bff06cd612b28e3df51afe, spanId: 0474f505a765a41f}
Can not execute the operation on ended Span {traceId: 05b3d14d94bff06cd612b28e3df51afe, spanId: 0474f505a765a41f}
tls.connect 05b3d14d94bff06cd612b28e3df51afe-0474f505a765a41f - You can only call end() on a span once.
```

The tls instrumentation handles these scenarios without stepping on its own toes.

### What did you see instead?

When a `fetch` timeouts, the tls instrumentation tries doing operations on an ended span

### Possible solutions

A couple of options (of course, more are possible):

1. Clear the event listeners inside the `connect` handler
2. Wrap the entire tls connection in a span of its own, where `tls.connect` is a child span. Subsequent errors hit the longer parent span

I'm more than willing to create a followup PR for this

mhennoch:
Fixes #1775

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts
@@ -162,7 +162,7 @@ export class NetInstrumentation extends InstrumentationBase<Net> {
       }
     };
 
-    for (const event of [SocketEvent.CLOSE, SocketEvent.ERROR]) {
+    for (const event of [SocketEvent.CLOSE, SocketEvent.ERROR, SocketEvent.SECURE_CONNECT]) {
       socket.once(event, otelTlsRemoveListeners);
     }
 
--- a/plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-net/src/instrumentation.ts
@@ -162,7 +162,11 @@ export class NetInstrumentation extends InstrumentationBase<Net> {
       }
     };
 
-    for (const event of [SocketEvent.CLOSE, SocketEvent.ERROR, SocketEvent.SECURE_CONNECT]) {
+    for (const event of [
+      SocketEvent.CLOSE,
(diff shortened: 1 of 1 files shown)


seemk (approved):

trentm (approved):
LGTM

pichlermarc (approved):

Issue #1587, opened 2023-07-14
Pino instrumentation fails when using ES modules


nlindley:
### What version of OpenTelemetry are you using?

- @opentelemetry/api: 1.4.1
- @opentelemetry/instrumentation: 0.41.0
- @opentelemetry/instrumentation-pino: 0.34.0
- @opentelemetry/sdk-node: 0.41.0
- @opentelemetry/sdk-trace-base: 1.15.0

### What version of Node are you using?

18.16.1

### What did you do?

Attempted to instrument pino using either default or named imports.

### What did you expect to see?

Trace information in the logs.

### What did you see instead?

When using default imports, the application crashes when `pino()` is called. When using named imports, trace information is not added to the logs.

### Additional context

We have minimal reproductions of the issues:

- Default import: https://github.com/nlindley/otel-esm-examples/tree/master/pino-default
- Named import: https://github.com/nlindley/otel-esm-examples/tree/master/pino-named

dyladan:
I think this is the same as https://github.com/open-telemetry/opentelemetry-js/issues/3976

For now you may be able to just use the previous version until we can provide a fix.

klippx:
They reverted https://github.com/open-telemetry/opentelemetry-js/pull/4011 maybe you should do the same?

jacobwood091:
This appears to still be an issue. We may want to look towards a solution such as that proposed by https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1694, as this seems to be a resolution difference within the [ECMAScript Module Namespace Object](https://nodejs.org/api/esm.html#commonjs-namespaces) versus that of the CommonJS resolved Namespace. [This](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1692#issuecomment-1728464734) comment, especially the Aside, provides an excellent analysis relevant to the underlying issue.

mbrevda:
are we sure this was fixed? It doesn't look like Pino is being instrumented, even with [0.34.4](https://github.com/open-telemetry/opentelemetry-js-contrib/compare/instrumentation-pino-v0.34.3...instrumentation-pino-v0.34.4).

<details><summary>otel config</summary>

```js
import {
  CompositePropagator,
  W3CTraceContextPropagator,
} from '@opentelemetry/core';
import {SimpleSpanProcessor} from '@opentelemetry/sdk-trace-base';
import {NodeTracerProvider} from '@opentelemetry/sdk-trace-node';
[...]

PR #1831, opened 2023-11-27
fix(instrumentation-pino): instrument pino used in ESM

trentm:
This also reduces the number of pino versions tested with
test-all-versions test (from 42 to 13, currently).

Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1587


plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
```diff
@@ -61,21 +63,25 @@ export class PinoInstrumentation extends InstrumentationBase {
...
           if (typeof patchedPino.pino === 'function') {
             patchedPino.pino = patchedPino;
           }
           if (typeof patchedPino.default === 'function') {
             patchedPino.default = patchedPino;
           }
+          if (isESM) {
+            module.pino = patchedPino;
+            module.default = patchedPino;
+          }
```
trentm, line 84: > Codecov / codecov/patch
> plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts#L82-L83
> Added lines #L82 - L83 were not covered by tests

These lines actually *are* covered by the test, but I guess the out of process `runTestFixture` doesn't catch that.

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
```diff
@@ -303,4 +307,39 @@ describe('PinoInstrumentation', () => {
       });
     });
   });
+
+  it('should work with ESM usage', async function () {
+    if (semver.lt(pino.version, '6.0.0')) {
+      // Pino v5 did not support named exports.
```
JamieDanielson, line 313: Thank you for commenting this!
Do we need another branch in this logic for default import for Pino v5 then?

trentm: You are right, though, that I could split the ESM testing into two separate tests: one that does default import and one that does named imports.  I started with both of them together... then discovered the pino@5 limitation and arguably got lazy. :)

trentm:
The pino 5.x part of the TAV tests now look like this:

```
...
-- required packages ["pino@5.14.0"]
-- installing ["pino@5.14.0"]
-- running test "npm run test" with pino (env: {})

> @opentelemetry/instrumentation-pino@0.34.3 test

  PinoInstrumentation
    ✓ should work with ESM default import (513ms)
    - should work with ESM named import
    enabled instrumentation
...
```

It was good to do this extra work, because it helped me find that the actual pino version where the named-ESM-exports thing changed was v6.8.0, not right at v6.0.0.

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
trentm: Done in commit b10ec11.

trentm:
@JamieDanielson This is ready for re-review if/when you have a chance. Thanks!

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
JamieDanielson: Noting for posterity:

> I don't know if OTel has a specific policy for support of instrumentations of really old package versions.

As we chatted about in SIG meeting, we don't have a specific policy in place right now for package versions. For now we're going with: If it's a relatively small lift to test older versions, let's keep doing it. Once it starts causing more significant pain, we should revisit and handle those on a case-by-case basis. It's difficult to have a standard policy across all packages since they are all so different, but we'll see how things shape up in the future.

Issue #1828, opened 2023-11-27
[ci] test-all-versions failing for `@cucumber/cucumber@9.1.2` and `@aws-sdk/client-s3@3.377.0`


pichlermarc:
### What happened?

The TAV script currently fails in CI only (tested locally but could not reproduce) for `@opentelemetry/instrumentation-cucumber` with `@cucumber/cucumber@9.1.2`.

It looks like it cannot resolve the package `@cucumber/messages` for some reason - when I run `npm run test-all-versions` locally, everything works fine.

### Logs:
- Node.js v14 (https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6960028459/job/19050349816)
- Node.js v16 (https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6960028459/job/19050350110)
- Node.js v18 (https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/6960028459/job/19050350393)

cc @Ugzuzg (component-owner)

Ugzuzg:
Same. Works locally.

pichlermarc:
Looks like this affects more than just the cucumber instrumentation, having a similar problem over at an instrumentation-aws-sdk PR. I'll re-label the issue accordingly.

Sorry for the noise

trentm:
# Repro for the cucumber TAV (test-all-versions) failure.

tl;dr:

```
npm ci
npx lerna run --scope @opentelemetry/instrumentation-cucumber test-all-versions
```

With details:

```
$ cd .../opentelemetry-js-contrib  # top dir of your clone

$ npm ci   # start with a clean layout of node_modules/... dirs

# Observe the current install location of all the '@cucumber/...' packages. These aren't all 
# the deps of '@cucumber/cucumber', but enough to see the issue.
$ ls **/node_modules/@cucumber/*/package.json
node_modules/@cucumber/ci-environment/package.json
node_modules/@cucumber/cucumber-expressions/package.json
node_modules/@cucumber/cucumber/package.json
node_modules/@cucumber/gherkin-streams/package.json
node_modules/@cucumber/gherkin-utils/node_modules/@cucumber/gherkin/package.json
node_modules/@cucumber/gherkin-utils/node_modules/@cucumber/messages/package.json
node_modules/@cucumber/gherkin-utils/package.json
node_modules/@cucumber/gherkin/package.json
node_modules/@cucumber/html-formatter/package.json
[...]

PR #1838, opened 2023-11-30
test: fix TAV tests for @cucumber/cucumber and @aws-sdk/client-s3

trentm:
- Switch to 'npm run --ws test-all-versions ...' for running TAV tests,
  instead of 'lerna run test-all-versions ...' because nx sets
  'npm_config_legacy_peer_deps=true' which breaks
  '@cucumber/cucumber@9.1.2' install and could break other installs by
  ignoring 'peerDependencies'.
- Skip the bad '@aws-sdk/client-s3@3.377.0' release in TAV tests.

Also:
- Reduce the number of versions of 'aws-sdk' and '@aws-sdk/*' packages
  test in TAV tests from 249, 143, and 132 versions currently, to 7
  each.
- Add a top-level `npm run test-all-versions` script to run that script
  in all packages that have one. This is the equivalent of the
  "test-all-versions.yml" CI workflow.

Fixes: #1828

---

**Update for reviewers**:  See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1838#issuecomment-1834359339 below for why I think this PR should go in, even though TAV tests are still failing.


trentm:
Note to reviews: The TAV tests **failed, but for a different reason, so I think the PR is good to go**.

1. The TAV tests for cucumber passed. Good.
2. The TAV tests for `@aws-sdk/client-sqs@3.462.0` failed (Bad, but a separate issue, see below), before even getting to the `@aws-sdk/client-s3` tests that are being fixed here.


# the client-sqs failure

```
> @opentelemetry/instrumentation-aws-sdk@0.37.0 test-all-versions
> tav

-- required packages ["@aws-sdk/client-sqs@3.462.0"]
-- installing ["@aws-sdk/client-sqs@3.462.0"]
-- running test "npm run test" with @aws-sdk/client-sqs (env: {})
...
  3 failing

  1) instrumentation-aws-sdk-v3
       custom service behavior
         SQS
           sqs send add messaging attributes:
     SyntaxError: Unexpected token < in JSON at position 0
  Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object.
      at JSON.parse (<anonymous>)
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@aws-sdk/client-sqs/dist-cjs/protocols/Aws_json1_0.js:1980:21
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at de_SendMessageCommand (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@aws-sdk/client-sqs/dist-cjs/protocols/Aws_json1_0.js:1013:18)
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@smithy/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@aws-sdk/client-sqs/node_modules/@aws-sdk/middleware-signing/dist-cjs/awsAuthMiddleware.js:30:20
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@smithy/middleware-retry/dist-cjs/retryMiddleware.js:27:46
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@aws-sdk/middleware-sdk-sqs/dist-cjs/send-message.js:7:18
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@aws-sdk/client-sqs/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:7:26
...
```

While was have https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1477 already for client-sqs TAV test failures... this failure is actually a *different* failure reason than the one currently being discussed there. I can update that issue (or create a separate one).

It would help if this PR went in, because the .tav.yml change means we are testing fewer versions of `client-sqs` so we should hit the test failures quicker in TAV test runs.

.github/workflows/test-all-versions.pr.yml
```diff
@@ -16,22 +16,22 @@ jobs:
...
     outputs:
-      args: ${{ steps.lerna-args.outputs.args }}
+      args: ${{ steps.npm-workspace-args.outputs.args }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         # Need lerna to list all packages
       - name: Install lerna
         run: npm install -g lerna@6.6.2
       - name: Parse labels into lerna scope arguments
```
pichlermarc, line 26:
```suggestion
      - name: Parse labels into npm workspace arguments
```

.github/workflows/test-all-versions.pr.yml
```diff
@@ -16,22 +16,22 @@ jobs:
...
       PR_LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
     outputs:
-      args: ${{ steps.lerna-args.outputs.args }}
+      args: ${{ steps.npm-workspace-args.outputs.args }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         # Need lerna to list all packages
       - name: Install lerna
         run: npm install -g lerna@6.6.2
```
pichlermarc, line 25: do we still need lerna? :thinking: 
Looks like the script is not using it anymore :thinking:

.github/workflows/test-all-versions.pr.yml
trentm: Thanks. Fixed in commit 93ae7b0f.

.github/workflows/test-all-versions.pr.yml
trentm: You are right. I missed that. Thanks. Fixed in commit 93ae7b0f.

pichlermarc (approved):
Looks great, thanks for taking care of this :slightly_smiling_face:

PR #1840, opened 2023-12-01
docs(user-interaction): update docs to include examples of config options

### LINKED ISSUE #1770: Add ability to add custom span attributes on user interaction traces

pkanal:
Documents config options for this instrumentation.

Closes #1770

- Add README docs for `eventNames` option
- Add README docs for `shouldPreventSpanCreation` option

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206095275723756

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-user-interaction/README.md
+++ b/plugins/web/opentelemetry-instrumentation-user-interaction/README.md
@@ -20,6 +20,9 @@ npm install --save @opentelemetry/instrumentation-user-interaction
 
 ## Usage
 
+
+### Initialize
+
 ```js
 import { ConsoleSpanExporter, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base';
 import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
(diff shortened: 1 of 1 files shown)


plugins/web/opentelemetry-instrumentation-user-interaction/README.md
```diff
@@ -82,6 +84,69 @@ function getData(url) {
 
 ```
 
+### Send spans for different events
+
+By default, only `click` events are automatically instrumented. To automatically instrument other events, specify the events that should be captured for telemetry. Most [browser events](https://developer.mozilla.org/en-US/docs/Web/Events) are supported.
```
martinkuba, line 89: I would clarify that this applies only to events of the HTMLElement interface and its implementations.

plugins/web/opentelemetry-instrumentation-user-interaction/README.md
```diff
@@ -82,6 +84,69 @@ function getData(url) {
...
+import { registerInstrumentations } from '@opentelemetry/instrumentation';
+
+
+// ...general opentelemetry configuration
+
+registerInstrumentations({
+  instrumentations: [
+    new UserInteractionInstrumentation({
+      eventNames: ['submit', 'click', 'keypress'],
+      shouldPreventSpanCreation: true,
```
martinkuba, line 120: I think this can only be a function

martinkuba (changes requested):

plugins/web/opentelemetry-instrumentation-user-interaction/README.md
```diff
@@ -82,6 +84,67 @@ function getData(url) {
...
+import { UserInteractionInstrumentation } from '@opentelemetryinstrumentation-user-interaction';
+import { registerInstrumentations } from '@opentelemetry/instrumentation';
+
+// ...general opentelemetry configuration
+
+registerInstrumentations({
+  instrumentations: [
+    new UserInteractionInstrumentation({
+      shouldPreventSpanCreation: (event, element, span) => {
+        span.setAttribute('target.id', element.id)
```
martinkuba, line 140:
```suggestion
        span.setAttribute('target.id', element.id);
```

martinkuba (approved):

PR #1846, opened 2023-12-04
fix: Update Azure Functions Detector to Only Detect When Function Version is Defined

JacksonWeber:
Azure Functions Detector should only detect Azure Functions when the function version is set. Currently it is erroneously detecting functions when in app service.


trentm:
Notes to self: 

> Currently it is erroneously detecting functions when in app service.

"Azure App Service" is a base service for a number of  Azure services, of which Azure Functions is one. So the presence of the `WEBSITE_SITE_NAME` envvar is not sufficient to identify this as being an Azure Function.

Docs on envvars in App Service here: https://learn.microsoft.com/en-us/azure/app-service/reference-app-settings
And the subset of envvars in Azure Functions here: https://learn.microsoft.com/en-us/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#functions

trentm (approved):

hectorhdzg (approved):
LGTM

PR #1850, opened 2023-12-07
feat(user-interaction): adding traceparent context from meta tag

xoscar:

- Each user interaction creates a new trace ID instead of using the one from the `traceparent` if exists

- Adds a function to retrieve the context based on the `traceparent` header meta tag
- Useful to generate a single trace for the entire user session


JamieDanielson:
Sorry for the delay on this @xoscar , we are catching up on PRs. I am wondering if the existing behavior is this way for a specific reason, and also curious about whether this change would benefit from being a configuration option. This also may change with the new instrumentation written in events, so it's not clear whether this is useful to implement now or consider something like this for the upcoming instrumentation change. Either way it would be good to get some input from some folks who work more on the web instrumentation. @martinkuba @MSNev @pkanal do you have any thoughts on this?

xoscar:
Hello @JamieDanielson sorry for the late response.

I'm currently working at Tracetest and we have provided a library that extends the current user interaction behavior using a single trace.
This is particularly useful when running synthetic tests from the browser side (playwright/cypress/etc) as a test can be encapsulated in a single trace that propagates the context from the browser side to the backend services.

Allowing users to create assertions against user-triggered events to backend spans.

I understand that the official implementation my diverge from what we currently have but we still think this has a valuable use case.

Thanks!

JamieDanielson:
> I'm currently working at Tracetest and we have provided a library that extends the current user interaction behavior using a single trace.
> This is particularly useful when running synthetic tests from the browser side (playwright/cypress/etc) as a test can be encapsulated in a single trace that propagates the context from the browser side to the backend services.

Oh, interesting use case, thank you for the details. I do wonder if it should be a configuration option, but don't want to ask you to do that work if this isn't something desired here. Pinging @martinkuba @MSNev @pkanal again for thoughts 🙏

MSNev:
https://github.com/open-telemetry/opentelemetry-specification/issues/3811

JamieDanielson:
We discussed this a bit during the JS SIG meeting today. Generally there was hesitation in taking this on, because it seems the instrumentation is actually working as expected right now with the user interaction being the root span, and this change would be breaking. With the expected changes / new instrumentation based on events coming up, we're not sure about having two major changes here. Also Nev linked a spec issue above where we are exploring options to standardize on how this may be done more holistically, and that work/discussion is still underway.

pichlermarc:
Thank you for your contribution @xoscar. A lack of reviews indicates that component-owners are not currently interested in adding this feature. It also seems like this is blocked on https://github.com/open-telemetry/opentelemetry-specification/issues/3811, so I'll close this PR for now. 

Please re-open a new PR if this is still a desired - though I think the way to go forward may be to pick up work on https://github.com/open-telemetry/opentelemetry-specification/issues/3811 first so that we have specification to go off of.

Issue #1758, opened 2023-10-30
Does mysql2 instrumentation record query parameters in db.statement span attribute by default?


dyllandry:
- [x] This only affects the JavaScript OpenTelemetry library
- [ ] This may affect other libraries, but I would like to get opinions here first

I'm not confident this is a bug, and I'm unsure what behaviour I should expect by default, so I'm asking this as an open question/discussion first to clarify.

I've just setup `@opentelemetry/auto-instrumentations-node` and can view traces created the mysql2 auto-instrumentation. I've also noticed it's included the entire database statement in every statement. Is this intended default behaviour for the mysql2 instrumentation?

We are using sequelize as an orm, we aren't writing any of our own sql statements. I know [static statements are not sanitized](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1744), but I assume sequelize is using parameterized statements too.

I saw in the mysql (not mysql2) docs that there's an [`enhancedDatabaseReporting`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-mysql#mysql-instrumentation-options) configuration option that is supposed to add query parameters to `db.statement`. This makes me think that the parameters shouldn't be there by default, but that is a different mysql module after all. That option doesn't exist for the mysql2 instrumentation, so maybe this module does add them by default.

macno:
Even if I see the utility to have the full statement available in the span, I think it should not be enable by default, for at least a couple of reasons:

1. values can be sensitive
2. values size

beside how the statement is generated (by an ORM or by hand) I'd expect to see the placeholders and not the values.

i.e. executing the following script, `db.statement` should be `insert into test_blob (data) values (?)`

```
const sdkNode = require('@opentelemetry/sdk-node');
const exporterOtel = require('@opentelemetry/exporter-trace-otlp-grpc');
const resources = require('@opentelemetry/resources');
const instrumentationMysql2 = require('@opentelemetry/instrumentation-mysql2');
const api = require('@opentelemetry/api');
const fs = require('fs/promises');
const sdk = new sdkNode.NodeSDK({
  traceExporter: new exporterOtel.OTLPTraceExporter(),
  instrumentations: [new instrumentationMysql2.MySQL2Instrumentation()],
  resourceDetectors: [resources.envDetector]
});

sdk.start();

const mysql = require('mysql2/promise');

const tracer = api.trace.getTracer('mysql2-test');

let connection;
[...]

Issue #1552, opened 2023-06-28
Change `db.statement` to only be collected if there is sanitization in mysql and redis


makeavish:
Currently `db.statement` is collected by default and there's no option to disable it. According to [open telemetry spec](https://github.com/open-telemetry/opentelemetry-specification/pull/3127), `db.statement` has to be disabled by default.

ramesius:
DynamoDB can be added to the list too.

ramesius:
@ithompson-gp I have addressed DynamoDB in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1748

PR #1857, opened 2023-12-11
fix(instrumentation-mysql2)!: instrumentation should not include values in db.statement

macno:

This PR addresses the bug in #1758 also following what emerged in https://github.com/open-telemetry/semantic-conventions/issues/436

*This is a breaking change*

-  The DB statement will be included in the span only when placeholders are used. The rationale behind this choice is that if there are no placeholders, it is assumed that the query SQL may include sensitive data, and therefore, no DB statement is included in the telemetry.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
@@ -39,8 +39,6 @@ import {
 /** @knipignore */
 import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 
-type formatType = typeof mysqlTypes.format;
-
 export class MySQL2Instrumentation extends InstrumentationBase<MySQL2InstrumentationConfig> {
   static readonly COMMON_ATTRIBUTES = {
     [SEMATTRS_DB_SYSTEM]: DBSYSTEMVALUES_MYSQL,
@@ -64,7 +62,7 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
           this._wrap(
             ConnectionPrototype,
             'query',
-            this._patchQuery(moduleExports.format, false) as any
+            this._patchQuery(false) as any
           );
 
           if (isWrapped(ConnectionPrototype.execute)) {
@@ -73,7 +71,7 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
           this._wrap(
             ConnectionPrototype,
             'execute',
-            this._patchQuery(moduleExports.format, true) as any
+            this._patchQuery(true) as any
           );
 
           return moduleExports;
@@ -82,14 +80,18 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
           if (moduleExports === undefined) return;
           const ConnectionPrototype: mysqlTypes.Connection =
             moduleExports.Connection.prototype;
-          this._unwrap(ConnectionPrototype, 'query');
-          this._unwrap(ConnectionPrototype, 'execute');
+          if (isWrapped(ConnectionPrototype.query)) {
+            this._unwrap(ConnectionPrototype, 'query');
+          }
+          if (isWrapped(ConnectionPrototype.execute)) {
+            this._unwrap(ConnectionPrototype, 'execute');
+          }
         }
       ),
     ];
   }
 
-  private _patchQuery(format: formatType, isPrepared: boolean) {
+  private _patchQuery(isPrepared: boolean) {
     return (originalQuery: Function): Function => {
       const thisPlugin = this;
       return function query(
@@ -110,7 +112,7 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
           attributes: {
             ...MySQL2Instrumentation.COMMON_ATTRIBUTES,
             ...getConnectionAttributes(this.config),
-            [SEMATTRS_DB_STATEMENT]: getDbStatement(query, format, values),
+            [SEMATTRS_DB_STATEMENT]: getDbStatement(query, values),
           },
         });
 
@@ -150,11 +152,10 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
               );
             }
           }
-
           span.end();
         });
 
-        if (arguments.length === 1) {
+        if (typeof arguments[arguments.length - 1] !== 'function') {
           if (typeof (query as any).onResult === 'function') {
             thisPlugin._wrap(
               query as any,
@@ -178,22 +179,13 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
             });
 
           return streamableQuery;
-        }
-
-        if (typeof arguments[1] === 'function') {
+        } else {
           thisPlugin._wrap(
             arguments,
-            1,
-            thisPlugin._patchCallbackQuery(endSpan)
-          );
-        } else if (typeof arguments[2] === 'function') {
-          thisPlugin._wrap(
-            arguments,
-            2,
+            arguments.length - 1,
             thisPlugin._patchCallbackQuery(endSpan)
           );
         }
-
         return originalQuery.apply(this, arguments);
       };
     };
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
@@ -154,8 +154,16 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
           }
           span.end();
         });
-
-        if (typeof arguments[arguments.length - 1] !== 'function') {
+        function findCallback(_arguments: IArguments) {
+          for (let i = 0; i < _arguments.length; i++) {
+            if (typeof _arguments[i] === 'function') {
+              return i;
+            }
+          }
+          return -1;
+        }
+        const cb = findCallback(arguments);
+        if (cb < 0) {
           if (typeof (query as any).onResult === 'function') {
             thisPlugin._wrap(
               query as any,
@@ -182,7 +190,7 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
         } else {
           thisPlugin._wrap(
             arguments,
-            arguments.length - 1,
+            cb,
             thisPlugin._patchCallbackQuery(endSpan)
           );
         }
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
@@ -36,10 +36,6 @@ interface QueryOptions {
   values?: any | any[] | { [param: string]: any };
 }
 
-interface Query {
-  sql: string;
-}
-
 interface Config {
   host?: string;
   port?: number;
@@ -47,6 +43,7 @@ interface Config {
   user?: string;
   connectionConfig?: Config;
 }
+
 /**
  * Get an Attributes map from a mysql connection config object
  *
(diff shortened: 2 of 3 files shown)


plugins/node/opentelemetry-instrumentation-mysql2/test/mysql2.test.ts
```diff
@@ -0,0 +1,237 @@
+/*
```
blumamir, line 1: The name of the test file is `mysql2.test.ts` but it's not clear why there are 2 files and what is the difference between them.

The instrumented package itself is called `mysql2` so that adds to the confusion. I would recommend giving the files better names that communicate their scope.

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -49,7 +50,7 @@ interface Result extends mysqlTypes.RowDataPacket {
   solution: number;
 }
 
-describe('mysql2@2.x', () => {
+describe('mysql2@2.x includeValuesInDbStatement=true (old behaviour)', () => {
```
blumamir, line 53: The text "old behaviour" is relevant now, but it will probably stay in the code base for a long time and will be less clear without the present context.

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -108,9 +120,13 @@ describe('mysql2@2.x', () => {
     rootConnection.end(() => {
       if (testMysqlLocally) {
         this.timeout(5000);
-        testUtils.cleanUpDocker('mysql');
+        setTimeout(() => {
+          testUtils.cleanUpDocker(testMysqlLocallyImage);
+          done();
+        }, 1000);
```
blumamir, line 126: What is the motivation to wait 1000ms in this case?

blumamir (approved):
Thanks for the fix!
Added few nit optional comments

plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
```diff
@@ -93,23 +94,31 @@ function getJDBCString(
...
   } else {
     // According to https://github.com/mysqljs/mysql#performing-queries
     // The values argument will override the values in the option object.
-    return values || (query as QueryOptions).values
-      ? format(query.sql, values || (query as QueryOptions).values)
+    statement = format
+      ? values || (query as QueryOptions).values
+        ? format(query.sql, values || (query as QueryOptions).values)
+        : query.sql
       : query.sql;
```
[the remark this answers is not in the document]
macno: WDYT?
```
    if (format) {
      const _values = values || (query as QueryOptions).values;
      statement = _values ? format(query.sql, _values) : query.sql;
    } else {
      statement = query.sql;
    }
```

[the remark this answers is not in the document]
blumamir: Much better, thank you 🙏

tolgap:
I just found this PR. Its quite egregious to include values in logs. Opens up a lot of auditing issues. Would be great to have this! Seems like development halted on this though 😞

JamieDanielson:
@macno is this something you are still working on? It seems this type of change will be desired, and we are wondering whether you are able to keep moving forward with this or if we should keep it as an open issue to pick up.

This would be part of updating/fixing #1552

macno:
Sorry, I was working on the tests when I had to interrupt.
The code is OK, if some someone else want to pick it up from here that's great, I'm not sure if I can work on this in the near future

macno:
Hi @JamieDanielson I'll be at kubecon next week, and I'm considering to work on this during the Contribfest, WDYT?

maryliag:
For whoever decides to work on this, a few considerations:
- we do have current semantic convention saying we should not collect the value if is not sanitized, but the prior versions didn't say that. I'm not sure which version was used when this was created, but we do have a [warning](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md) that if was using 1.24 or prior and you make a change, you need to use the env variable `OTEL_SEMCONV_STABILITY_OPT_IN` for the new changes
- I think this can still be treated as a bug fix
- This will be a temporary solution. We have a WG focusing on database metrics, and one of the things we're focusing now is that we want to have values not sanitized as not being set by default, and give the option to set the value anyway (similar to what will be done here), but we want to do that we the config file currently being developed. This way all databases will be consistent. So this can still be worked on, but it will be temporary and eventually replaced.

trentm:
Docker compose snippet:

```yaml
  mysql:
    image: mysql:5.7
    # No ARM64 image layer. See https://stackoverflow.com/a/65592942
    platform: linux/x86_64
    environment:
      MYSQL_ALLOW_EMPTY_PASSWORD: 1
    ports:
      - "3306:3306"
    volumes:
      - nodemysqldata:/var/lib/mysql
    healthcheck:
      test: ["CMD", "mysql" ,"-h", "mysql", "-P", "3306", "-u", "root", "-e", "SELECT 1"]
      interval: 1s
      timeout: 10s
      retries: 30

```

macno:
to run the tests on arm64, spin up a container manually
```
services:
  mysql:
    image: mysql:5.7
    platform: linux/amd64
    environment:
      MYSQL_ROOT_PASSWORD: rootpw
      MYSQL_USER: otel
      MYSQL_PASSWORD: secret
      MYSQL_DATABASE: test_db
    ports:
      - "33306:3306"
```

trentm:
I'll sponsor this to avoid the autoclose for unmaintained packages.

plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
```diff
@@ -149,11 +147,13 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
...
           }
-
           span.end();
         });
 
-        if (arguments.length === 1) {
+        if (
+          arguments.length === 1 ||
+          (arguments.length === 2 && typeof arguments[1] !== 'function')
+        ) {
```
trentm, line 156: This is an interesting fix that could be mentioned in the PR description. IIUC, before this change, the following usage would result in an unended span:

```ts
        const sql = 'SELECT 1+? as solution';
        const query = connection.query(sql, [2]);
```

trentm (changes requested):
- The plan for this PR has changed. Please update the PR description to describe the current plan.
- Given this is a breaking change, please change the PR title to be `fix(instrumentation-mysql2)!: instrumentation should not include values in db.statement`. The exclamation point indicates a breaking change for the release process.
- Note: instrumentation-mysql dealt with this same issue (including values in 'db.statement') in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1337 In that PR a `enhancedDatabaseReporting` boolean config option was added to get the old behaviour back.  I think this PR should *not* add this config option and instead just make the breaking change and wait for common DB handling to be specified by the DB SIG (as [Marylia mentioned above](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1857#issuecomment-2471937734)).

trentm:
@macno This looks great, just a few small requested changes.

macno:
> @macno This looks great, just a few small requested changes.

thank you @trentm ! I'm going to take care of your suggestions, much appreciated

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -67,7 +67,7 @@ describe('mysql2', () => {
   const memoryExporter = new InMemorySpanExporter();
 
   const getLastQueries = (count: number) =>
-    new Promise<string[]>(res => {
+    new Promise<string[]>((res, reject) => {
```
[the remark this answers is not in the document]
macno: I made sure we use it

plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
macno: not sure it _was_ a bug, I had to add it since some existing tests were failing. Anyway, looking at it after your comment, I rewrote the statement to simplify it.

trentm:
@macno a side-node: If you are able to avoid *force*-pushing to a feature branch after there are reviews, that would be helpful.  A *force* push removes the earlier commit shas, so I am not able to see the changes since my last review, for example.

It is okay to have a bunch of commits on a feature branch. They will all be squashed into a single commit when the PR is merged.

plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
```diff
@@ -81,14 +79,18 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
...
           const ConnectionPrototype: mysqlTypes.Connection =
             moduleExports.Connection.prototype;
-          this._unwrap(ConnectionPrototype, 'query');
-          this._unwrap(ConnectionPrototype, 'execute');
+          if (isWrapped(ConnectionPrototype.query)) {
+            this._unwrap(ConnectionPrototype, 'query');
+          }
+          if (isWrapped(ConnectionPrototype.execute)) {
+            this._unwrap(ConnectionPrototype, 'execute');
+          }
```
trentm, line 87: I'm curious, was this added because of your added "initialization" test below? Was that resulting in `logger('no original to unwrap to -- has ' + name + ' already been unwrapped?')` console errors?

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -156,33 +163,76 @@ describe('mysql2', () => {
...
           (poolCluster as any).end(() => {
-            done();
+            done(err);
           });
         }
       });
     });
   });
 
+  describe('initialization', () => {
```
trentm, line 182: Why the addition of this test? Did you hit some specific issue?

1. I think this is not testing what you think it is testing. 

With both `instrumentation` and `instrumentation2` being enabled: the first one wraps, then the second one unwrap and wraps. Then when `.disable()` is called for both, the first one unwraps, and the second one is a no-op.

When running just this "initialization" test and with some added logging to show the value of `isWrapped(mysqlTypes.Connection.prototype.query)` before and after each enable/disable call, it looks like this:

```
% RUN_MYSQL_TESTS=1 npm test

> @opentelemetry/instrumentation-mysql2@0.43.0 test
> nyc mocha 'test/**/*.test.ts'



  mysql2
    initialization
XXX instr2.enable() isWrapped? false
XXX instr2.enabled isWrapped? true
XXX instr.enable() isWrapped? true
XXX instr.enabled isWrapped? true
      ✔ should not wrap more than once
XXX instr.disable() isWrapped? true
XXX instr.disabled isWrapped? false
XXX instr2.disable() isWrapped? false
XXX instr2.disabled isWrapped? false


  1 passing (17ms)
```

My diff for that test run is here: https://gist.github.com/trentm/d85145848dce0f45b8c440a6f655a078


2. We are actually considering *dropping* `Instrumentation#disable()` functionality altogether. See https://github.com/open-telemetry/opentelemetry-js/issues/5154 if you are curious.

My preference would be to not add this test.

trentm (approved):
Looking good. I have some suggestions, but they don't need to block.

plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
macno: yes, correct

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
macno: Honestly, I added it to reach 100% functions and lines coverage and see the branches not covered.
At the moment the only branch not covered is this one https://github.com/macno/opentelemetry-js-contrib/blob/issue-1758-mysql2-instrumentations-values-in-db.statement/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts#L80 which I have no idea how to test..

plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
```diff
@@ -149,11 +151,10 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
               );
             }
           }
-
           span.end();
         });
 
-        if (arguments.length === 1) {
+        if (typeof arguments[arguments.length - 1] !== 'function') {
```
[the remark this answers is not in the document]
macno: I added a fixup that should address the point you raised

morley92:
@macno is this something you're still working on or could somebody take over? Keen to take advantage of this to remove PII from our traces

gflachs:
Thank you very much for the PR and the effort that has been put into it so far.

However, I have a suggestion for a change:

The assumption in the PR:  
*"The rationale behind this choice is that if there are no placeholders, it is assumed that the query SQL may include sensitive data, and therefore, no DB statement is included in the telemetry."*  
does not seem correct. 

For example, in our use cases, we generate static queries (e.g., `SELECT * FROM TABLE`) that logically do not set any values but should still be included as attributes.

### My proposal:

We extend the **MySQL2InstrumentationConfig** with two variables:  
1. **maskStatementHook** → This hook is used to escape SQL statements (optional).  
2. **maskStatement** → Boolean (default: `true`), determines whether the **maskStatementHook** is applied at all.  

For **maskStatementHook**, I suggest a default function that is applied when no custom function is set:

```js
return query.replace(/\b\d+\b/g, '?').replace(/(["'])(?:(?=(\\?))\2.)*?\1/g, '?')
```

This allows all queries to still be tracked.

Since this PR has been in progress for quite some time, I am happy to take over and submit a new PR with my proposed solution if there is interest. I would implement it quickly, as we need this feature in our company but still want to use auto-instrumentation (which is currently not possible, and even with the fix in this PR, it would not be, as explained above).

Issue #1739, opened 2023-10-13
`@opentelemetry/instrumentation-graphql` increases event loop latency by ~10x


HunterLarco:
### What version of OpenTelemetry are you using?

1.17.0

### What version of Node are you using?

17.9.1

### What did you do?

Installed `@opentelemetry/instrumentation-graphql` with default options and `depth: 0`.

### What did you expect to see?

Minimal impact to latency.

### What did you see instead?

While debugging a graphql query using ApolloServer which returns 340KB of data (11,494 keys in the response JSON) I noticed that our business logic was executing just under 50ms while the request was taking ~1.2s. Using the tracing data collected by `@opentelemetry/instrumentation-graphql` I was able to observe that all of our resolvers took ~86ms CPU time but 1.1s wall time. I suspected something was wrong with resolver scheduling so ran `clinic doctor -- node server.ts` to gain insight and noticed during the span of the query, the event loop experienced ~250ms of latency. Removing `@opentelemetry/instrumentation-graphql` decreased this down to ~20ms of latency and reduced the overall request timing to 187ms.

TL;DR even with `depth: 0`, for large payloads `@opentelemetry/instrumentation-graphql` increased event loop latency by an order of magnitude.

pichlermarc:
This is possibly related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1686. Apollo sets every resolver to a custom one, even if it's trivial, which causes extremely verbose traces.

sdacunha:
If it helps, I currently disable the graphql integration for Apollo and setup a plugin and instrument (in my case I instrument in otel via sentry v8) manually via an Apollo plugin. It's possible to detect the trivial resolver case using this: https://community.apollographql.com/t/is-there-a-way-to-know-if-a-graphql-field-will-be-resolved-using-the-default-resolver/2573

Blacksmoke16:
Should this be closed given https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1858 was merged? Or is it not fully resolved?

PR #1858, opened 2023-12-11
feat(instrumentation-graphql): add option to ignore resolver spans

seemk:

#1739 

Add the option to ignore resolver spans.

Creating a span for each resolve can cause way too many spans to be created (e.g. 100k elements in the response equals at least 100k spans, most of them will be dropped). This causes huge latency increases - my test query with an in memory resolver went from 1.2s to 8-9s. With `ignoreResolveSpans: true` the response time went back to ~1.2s.

For reference, the Java implementation of GraphQL instrumentation ignores resolvers [by default](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/graphql-java-12.0/testing/src/main/java/io/opentelemetry/instrumentation/graphql/AbstractGraphqlTest.java) and there isn't even a possibility to enable creating spans from resolvers.

I think ignoring resolvers should be the default behaviour in the Node.js instrumentation as well, but for backwards compatibility I set the default value to `false` in this PR. Up for discussion whether to ignore resolvers by default or not.


blumamir (approved):

seemk:
@pichlermarc Are there any additional approves this PR needs?

satazor:
Just noticed that this PR missed the documentation update on the README.

PR #1859, opened 2023-12-11
wip: fix dual publish express

JamieDanielson:

-

-

WIP: attempting fix for dual-publishing CJS/ESM builds


blumamir:
@JamieDanielson Wondering if this is still relevant and should be kept open.

JamieDanielson:
> @JamieDanielson Wondering if this is still relevant and should be kept open.

I think this will be relevant and might be needed in all the packages... eventually. I guess the main question is to confirm whether this definitely works as expected and also determine the timeline to get these done. I've set it aside for now and will close as this can still be used as a guide even if closed.

Issue #1860, opened 2023-12-12
get TAV tests passing again


trentm:
This is a working issue to attempt to get test-all-versions (TAV) tests passing again. They had been failing for a long time due to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1477 and other issues.

# current status

https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/7174025293/job/19534639796
is a run after some aws-sdk and cucumber TAV fixes were merged.
The Node 16 run actually passed! The jobs for Node.js 14 and 18 failed. I don't yet know why.
They all took ~1h40m.

**Update (2023-12-20):**
- Node 14, 16, and 18 TAV tests are failing for instrumentation-redis-4 due to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1874. This is a change in behaviour in the latest `redis@4` release, so is exactly the kind of issue that TAV is being used for.
- Node 18 TAV tests are failing due to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1872 (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1873 will workaround that)

# timings from logs

I downloaded the logs archive (122MB) and unzipped to "~/Downloads/tav-pass".
This grep roughly shows the runtimes and failures for each TAV module:

```
[~/Downloads/tav-pass/tav  Run test-all-versions (18)]
% rg -N test-all-versions "9_Run test-all-versions.txt"
2023-12-11T22:24:50.3355392Z ##[group]Run npm run --if-present --workspaces test-all-versions
2023-12-11T22:24:50.3355999Z npm run --if-present --workspaces test-all-versions
2023-12-11T22:24:50.6565829Z > @opentelemetry/instrumentation-amqplib@0.33.4 test-all-versions
2023-12-11T22:26:25.7835648Z > @opentelemetry/instrumentation-cucumber@0.2.0 test-all-versions
2023-12-11T22:31:47.1645628Z > @opentelemetry/instrumentation-dataloader@0.5.3 test-all-versions
2023-12-11T22:32:25.7075321Z > @opentelemetry/instrumentation-lru-memoizer@0.33.4 test-all-versions
2023-12-11T22:35:12.7549151Z > @opentelemetry/instrumentation-mongoose@0.33.4 test-all-versions
2023-12-11T23:05:02.9052204Z > @opentelemetry/instrumentation-socket.io@0.34.4 test-all-versions
[...]

PR #1861, opened 2023-12-12
test: reduce number of mongoose versions (TAV) tested from 204 to 19

trentm:
This tests just the latest in each minor version.

Refs: #1860


pichlermarc (approved):
I think just testing the latest in each minor version is a reasonable approach. Maybe we should add that to the `CONTRIBUTING.md` or `GUIDELINES.md`? :thinking:

PR #1862, opened 2023-12-13
test: fix redis-4 TAV tests with Node.js v14

trentm:
redis@4.6.9 was a bad release that accidentally broken v14 support. This updates the versions test to skip that release, and also reduces the number of versions tested from 27 to 8 by selecting just the current latest minor releases.

Also, drop the pretest config that (a) was never used because of the wrong indentation and (b) is no longer valid since the move to npm workspaces.

Refs: #1860

---

**Note**: The redis-4 TAV tests still fail with Node v18 and v16 (sometimes, I think). That is a separate issue to what this one is fixing.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-redis-4/.tav.yml
+++ b/plugins/node/opentelemetry-instrumentation-redis-4/.tav.yml
@@ -1,8 +1,7 @@
 redis:
-
   jobs:
-    - versions: "^4.0.0"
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
Thank you for working on this :slightly_smiling_face: 
(Link to the issue about `4.6.9` and Node 14 for future reference: https://github.com/redis/node-redis/issues/2619)

PR #1582, opened 2023-07-13
fix(aws-xray): align aws xray to use contrib-test-utils v0.34.0

haddasbronfman:
align aws xray to use contrib-test-utils v0.34.0 as all other packages.

PR #1866, opened 2023-12-15
fix(deps): update otel core experimental

pichlermarc:
This is the renovate PR #1865 merged with #1864 and a newly generated `package-lock.json` - for some reason renovate did leave the package-lock.json in a weird state where some old packages were still used leading to compilation failures due to differing types.

This PR also updates the `@opentelemetry/contrib-test-utils` for `@opentelemetry/sampler-aws-xray`. 

Note: we've actually had this problem before that it's left behind for some reason. (#1582) I previously assumed it was because an x-ray sampler PR was merged with old dependencies after the utils were already updated on main, but now I'm not exactly sure why this is happening - the only thing that's different to my eye is that it's a private package.


pichlermarc:
Converting back to draft as it does seem to add some other issues. Looks like it's pulling in some very old version of instrumentation-http. I wonder if that has something to do with the outdated examples in this repo.

pichlermarc:
Ah, now I get it. I think this depends on the fix included in `import-in-the-middle@1.7.2`. I'll try it out locally. We should probably consider changing the version in `@opentelemetry/instrumentation` to allow for `1.7.x`.

pichlermarc:
Ah, it looks like this is actually failing as IITM is swallowing the (cjs) exports from instrumentation-http

pichlermarc:
Tests are all green now, marking as ready-for-review. Some changes I made:
- delete `package-lock.json` and generate a new one
- removed `useUnifiedTopology` from the `mongodb-example` as this is the default in previous versions and would not compile anymore with the updated one that was installed with a fresh `npm install` (bf19f7bbf5bd0a933b6d1080f2db247595ffcc0a)
- bump the `rxjs` dev dependency of `instrumentation-nextjs-core` as it was conflicting with types from another dev dependency used in other parts of the repo ([7845a6b](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1866/commits/7845a6b2075a4f79e52cf43e2517daed6cbfab58))

trentm (approved):
@pichlermarc Thanks very much for working through this!

Issue #1664, opened 2023-09-04
Pino: Allow logging trace metadata under different keys


drewcorlin1:
### Is your feature request related to a problem? Please describe

The current pino instrumentation from `@opentelemetry/instrumentation-pino` injects 3 fields into each log, `trace_id`, `span_id` and `trace_flags`. These fields are useful, but the snake case naming is slightly problematic in my case, where my company has strict logging standards and we'd like to use camelCase.

### Describe the solution you'd like to see
Allowing a different set of keys to be used in the config (this is assuming you are configuring the instrumentation via `getNodeAutoInstrumentations` from `@opentelemetry/auto-instrumentations-node`).

```typescript
'@opentelemetry/instrumentation-pino': {
  camelCase: true
}
```

### Describe alternatives you've considered
Reformatting the log manually, via the `logHook` config.

```typescript
'@opentelemetry/instrumentation-pino': {
    logHook: (span, record) => {
      // Reformat the injected log fields to use camelCase, eg. trace_id -> traceId
      const context = span.spanContext();
      record.traceId = context.traceId;
      record.spanId = context.spanId;
      record.traceFlags = context.traceFlags;

      // Be extra safe, in case someone manually logged these fields, even though they shouldn't
      if (record.trace_id === context.traceId) delete record.trace_id;
      if (record.span_id === context.spanId) delete record.span_id;
      if (record.trace_flags === context.traceFlags) delete record.trace_flags;
    },
  },
```

Flarna:
I think it would be more flexible to allow to override the actual key names instead of adding just one more fixed set. e.g
```
'@opentelemetry/instrumentation-pino': {
  traceIdKey: 'traceId', // default is 'trace_id
  spanIdKey: 'span.id', // default is 'span_id'
  trace_flags: 'my-trace-flags'  // defaults to 'trace_flags'
}
```

drewcorlin1:
Not stale

mbrevda:
This could also be useful for specific log destinations. For example. GCP requires the following format: 

```js
   const trace = `projects/${projectId}/traces/${traceId}`;
      res['logging.googleapis.com/trace'] = trace;
    }
    res['logging.googleapis.com/trace_sampled'] = traceFlags;
    res['logging.googleapis.com/spanId'] = spanId;

trentm:
@drewcorlin1 This can be closed now, right (it was handled by #1867)?

drewcorlin1:
[...]

PR #1867, opened 2023-12-17
feat: Allow configuring pino to log with different keys

drewcorlin1:
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1664

Allow configuring the keys that the pino instrumentation adds to logs


plugins/node/opentelemetry-instrumentation-pino/README.md
```diff
@@ -30,7 +30,14 @@ registerInstrumentations({
     new PinoInstrumentation({
       // Optional hook to insert additional context to log object.
       logHook: (span, record, level) => {
-        record['resource.service.name'] = provider.resource.attributes['service.name'];
+        record['resource.service.name'] =
+          provider.resource.attributes['service.name'];
+      },
+      // Log span context under custom keys
```
Flarna, line 36: Please make it more clear that this is optional.
Besides that the text below which describes the added fields should be improved to mention that below values are the defaults.

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
```diff
@@ -87,10 +106,16 @@ describe('PinoInstrumentation', () => {
...
+  function setup() {
+    instrumentation = new PinoInstrumentation({
+      logKeys,
+    });
     instrumentation.enable();
     pino = require('pino');
+  }
+
+  before(() => {
+    setup();
```
Flarna, line 118: If I understand this correct this causes that all tests below use the new keys, not just the new `injects span context to records with custom keys` test.

drewcorlin1: That's right. I was having trouble configuring just one test independently of the others with the current setup. I can spend a bit more time on it

plugins/node/opentelemetry-instrumentation-pino/README.md
drewcorlin1: Updated, let me know if you think that's clear or needs to be updated still

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
drewcorlin1: Cleaned them up but the tests are interfering with each other causing the "disabled instrumentation" block of tests to fail. If I comment out all the other `describe` blocks then they pass. I'm stumped at the moment but will revisit this in the next couple days

Flarna: Seems like the pino tests were written in a way to instrument just once and then run all tests.
Don't know if there was a reason. Theoretically `.disable()` should cause to de-instrument and therefore should allow side effect free tests.
But there might be cases where de-instrument doesn't work, e.g. if pino caches some stuff internal.

Flarna:
@seemk Could you as owner please take a look?

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
drewcorlin1: I definitely don't have a perfect understanding of this patching, I tried creating a new instrumentation + explicitly calling `enable()` on it as well as clearing the require cache via `delete require.cache[require.resolve('pino')]` but it made no difference. Would love to hear alternatives to try as welll

drewcorlin1: I broke the enabled + disabled tests out into their own modules to skirt the problem. Let me know what you think

plugins/node/opentelemetry-instrumentation-pino/test/common.ts
```diff
@@ -0,0 +1,100 @@
...
+  sinon.assert.calledOnce(writeSpy);
+  const record = JSON.parse(writeSpy.firstCall.args[0].toString());
+  assert.strictEqual(record['trace_id'], undefined);
+  assert.strictEqual(record['span_id'], undefined);
+  assert.strictEqual(record['trace_flags'], undefined);
+  assert.strictEqual(kMessage, record['msg']);
+  return record;
+}
+
+export function init(importType: 'global' | 'default' | 'pino' = 'global') {
```
seemk, line 78: Perhaps instead of using global this should return something like a test context object containing the currently global fields?

seemk (approved):
Looks good, just a subjective comment regarding globals.

drewcorlin1:
@seemk @Flarna could you run/approve the required workflows?

blumamir (approved):

drewcorlin1:
Thank you @blumamir!

Could someone with access merge this whenever you get a chance, thanks!

PR #1879, opened 2023-12-22
ci: automatically map component labels to PRs

pichlermarc:

Currently approvers/maintainers need to add labels manually to PRs to get TAV to run. This PR adds a workflow that adds these labels based on the files that were changed.

### DIFF
--- a/.github/component-label-map.yml
+++ b/.github/component-label-map.yml
@@ -1,0 +1,174 @@
+pkg:resource-detector-alibaba-cloud:
+  - changed-files:
+      - any-glob-to-any-file: detectors/node/opentelemetry-resource-detector-alibaba-cloud/**
+pkg:resource-detector-aws:
+  - changed-files:
+      - any-glob-to-any-file: detectors/node/opentelemetry-resource-detector-aws/**
+pkg:resource-detector-azure:
+  - changed-files:
+      - any-glob-to-any-file: detectors/node/opentelemetry-resource-detector-azure/**
+pkg:resource-detector-container:
+  - changed-files:
+      - any-glob-to-any-file: detectors/node/opentelemetry-resource-detector-container/**
+pkg:resource-detector-gcp:
+  - changed-files:
+      - any-glob-to-any-file: detectors/node/opentelemetry-resource-detector-gcp/**
+pkg:resource-detector-github:
+  - changed-files:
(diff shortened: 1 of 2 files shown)


.github/component-label-map.yml
```diff
@@ -0,0 +1,174 @@
...
+      - any-glob-to-any-file: packages/opentelemetry-id-generator-aws-xray/**
+pkg:sampler-aws-xray:
+  - changed-files:
+      - any-glob-to-any-file: packages/opentelemetry-sampler-aws-xray/**
+pkg:propagation-utils:
+  - changed-files:
+      - any-glob-to-any-file: packages/opentelemetry-propagation-utils/**
+pkg:redis-common:
+  - changed-files:
+      - any-glob-to-any-file: packages/opentelemetry-redis-common/**
```
trentm, line 42: There isn't currently a `pkg:redis-common` label, nor are there likely to be TAV tests for that. However the instr-ioredis, instr-redis, and instr-redis-4 packages depend on this one, so what do you think about adding `- any-glob-to-any-file: packages/opentelemetry-redis-common/**` to the blocks for `pkg:instrumentation-ioredis`, `pkg:instrumentation-redis`, and `pkg:instrumentation-redis-4`?

If so, then we could also add `- any-glob-to-any-file: packages/opentelemetry-sql-common/**` to `pkg:instrumentation-mysql2` and `pkg:instrumentation-pg`.

I know that means encoding more "smarts" in this file, which means it can get out of date.  (I suppose there could eventually be a script that looks for local package dependencies in all packages that have a .tav.yml file and updates this config file.)

trentm (approved):
This looks great. Approving with a question/suggestion.

.github/component-label-map.yml
blumamir: > I know that means encoding more "smarts" in this file, which means it can get out of date. (I suppose there could eventually be a script that looks for local package dependencies in all packages that have a .tav.yml file and updates this config file.)

I think lerna has some features for resolving the dependency graph and applying smart transitive logic which might be a good fit for these kind of tasks. But maintaining the config file manually is also ok as there are rarely any changes on dependencies.

blumamir (approved):
I support @trentm suggestion.
Other than that, LGTM 🥇

trentm (approved):

JamieDanielson (approved):
Love having the robots do more work for us! 🤖

PR #1881, opened 2023-12-27
docs: update typing guidelines

david-luna:

This PR showcases the use of [generic types]((https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-types)) when exporting types to the public API that depend on the instrumented package.

The generic types approach seeks to avoid adding a types package as a dependency of the instrumentation because it may bring some problems to consumers like https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1787

This approach has been added as a replacement of the `any` type approach since its the fallback type when using a generic

- update guildelines doc


david-luna:
@blumamir this PR comes from your comment https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1804#pullrequestreview-1755846788 but Im not sure if you meant exactly this change.

Could you have a look ?

david-luna:
dont know how the other commits got it. Going to recreate the PR

blumamir:
I think it's a great workaround for the breaking type issue and a welcome addition to the guidelines. 
would love to approve when you reopen

PR #1882, opened 2023-12-27
chore: update guidelines

david-luna:

This PR showcases the use of [generic types]((https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-types)) when exporting types to the public API that depend on the instrumented package.

The generic types approach seeks to avoid adding a types package as a dependency of the instrumentation because it may bring some problems to consumers like https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1787

This approach has been added as a replacement of the `any` type approach since its the fallback type when using a generic

- update guildelines doc

cc: @blumamir

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -127,8 +127,40 @@ To support this use case, you can choose one of the following options:
 
     It is recommended to use this option when the types involved are simple and short.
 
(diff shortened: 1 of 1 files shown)


blumamir (approved):
💯

PR #1884, opened 2024-01-03
fix: Fix Azure Fns Detector When Running with App Service Detector

JacksonWeber:

When running inside of an Azure Function and in the case where both the Azure Functions and Azure App Services detector were being run in the same application, because the Azure App Service check was not exclusive of the Azure Functions check, the application would incorrectly be identified as running inside of Azure App Services.

This PR also adds tests for both Azure Functions and Azure App Services detectors to ensure that they function as expected when running in the same application.

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-azure/test/detectors/AzureFunctionsDetector.test.ts
+++ b/detectors/node/opentelemetry-resource-detector-azure/test/detectors/AzureFunctionsDetector.test.ts
@@ -16,7 +16,9 @@
 
 import * as assert from 'assert';
 import { azureFunctionsDetector } from '../../src/detectors/AzureFunctionsDetector';
+import { azureAppServiceDetector } from '../../src/detectors/AzureAppServiceDetector';
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+import { detectResourcesSync } from '@opentelemetry/resources';
 
 describe('AzureFunctionsDetector', () => {
   let originalEnv: NodeJS.ProcessEnv;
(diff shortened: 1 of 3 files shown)


detectors/node/opentelemetry-resource-detector-azure/test/detectors/AzureFunctionsDetector.test.ts
```diff
@@ -35,7 +37,51 @@ describe('AzureFunctionsDetector', () => {
...
+  it('should not detect azure app service values', () => {
+    process.env.WEBSITE_SITE_NAME = 'test-function';
+    process.env.REGION_NAME = 'test-region';
+    process.env.WEBSITE_INSTANCE_ID = 'test-instance-id';
+    process.env.FUNCTIONS_EXTENSION_VERSION = '~4';
+    process.env.WEBSITE_MEMORY_LIMIT_MB = '1000';
+
+    const resource = detectResourcesSync({
+      detectors: [azureFunctionsDetector, azureAppServiceDetector],
+    });
```
trentm, line 84: Now the two test cases ('should test functions values' and 'should not detect azure app service values') are identical, so I'm not sure of the point.

Perhaps just have the one test case and add an assertion that one of the Azure App Service-specific attributes (e.g. `SemanticResourceAttributes.SERVICE_INSTANCE_ID`) is *not* defined?

JacksonWeber: That makes sense. I've unified the two tests and now check for the absence of a few Azure App Service specific attributes.

hectorhdzg (approved):
LGTM

trentm (approved):

dyladan (approved):

Issue #1888, opened 2024-01-04
Error reading configuration for prometheus in fastify example


elvus:
### What version of OpenTelemetry are you using?
Latest

### What version of Node are you using?
v20.9.0

### What did you do?
I try to run the fastify example and it's giving me this error
![image](https://github.com/open-telemetry/opentelemetry-js-contrib/assets/24416449/1c69ca3a-f2d6-4e0e-bab8-edbbe636b5ee)

### What did you expect to see?
I expected the Fastify example to run successfully without any errors.

### What did you see instead?
The application did not start as expected.

### Additional context
System: MacOS Sonoma 14.2.1
Processor: M1

### Configuration files:
collector-config.yaml
```yml
receivers:
  otlp:
    protocols:
      grpc:
      http:
        cors:
          allowed_origins:
            - http://*
            - https://*
  zipkin:
    endpoint: "http://localhost:9411/api/v2/spans"
  prometheus:
    endpoint: "0.0.0.0:9464"

processors:
  batch:

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [zipkin]
      processors: [batch]
    metrics:
      receivers: [otlp]
      exporters: [prometheus]
      processors: [batch]
```
prometheus.yml
```yml
global:
  scrape_interval: 15s # Default is every 1 minute.

scrape_configs:
  - job_name: 'collector'
    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.
    static_configs:
      - targets: ['collector:9464']
```
docker-compose.yml
```yml
version: "3"
services:
  # Collector
  collector:
    #image: otel/opentelemetry-collector:0.38.0
    image: otel/opentelemetry-collector
    command: ["--config=/conf/collector-config.yaml"]
    volumes:
      - ./collector-config.yaml:/conf/collector-config.yaml
    ports:
      - "9464:9464"
      - "4317:4317"
      - "4318:4318"
    depends_on:
      - zipkin-all-in-one
  # Zipkin
  zipkin-all-in-one:
    image: openzipkin/zipkin
    ports:
      - "9411:9411"
  # Prometheus
  prometheus:
    container_name: prometheus
    image: prom/prometheus:latest
    volumes:
      - ./prometheus.yml:/etc/prometheus/prometheus.yml
    ports:
      - "9090:9090"
```

pichlermarc:
[...]

PR #1899, opened 2024-01-15
fix(example-fastify): update example to use latest versions

pichlermarc:

The fastify example was outdated and did not work anymore.

Fixes: #1888

Update everything to latest versions, get rid of some outdated practices.


examples/fastify/package.json
```diff
@@ -36,23 +36,23 @@
...
+    "@opentelemetry/sdk-node": "^0.46.0",
+    "@opentelemetry/exporter-trace-otlp-proto": "^0.46.0",
+    "@opentelemetry/exporter-metrics-otlp-proto": "^0.46.0",
+    "@opentelemetry/instrumentation": "^0.46.0",
+    "@opentelemetry/instrumentation-http": "^0.46.0",
+    "@opentelemetry/instrumentation-fastify": "^0.32.6",
+    "@opentelemetry/sdk-trace-node": "~1.19.0",
+    "@opentelemetry/resources": "~1.19.0",
+    "@opentelemetry/semantic-conventions": "~1.19.0",
+    "@opentelemetry/sdk-trace-base": "~1.19.0",
```
trentm, line 49: The tilde ranges were intentional, right? E.g. this limits `@opentelemetry/sdk-trace-base` to `>=1.19.0 <1.20.0`.

trentm (approved):
LGTM. A couple notes:

This could possibly update the zipkin screenshot (images/trace1.png). Here is one from my using this:

Screenshot 2024-01-18 at 8 57 23 AM

Also, just for interest, I noticed this warning when runnign the client:

```
% npm run client

> example-fastify@0.26.0 client
> cross-env OTEL_SERVICE_NAME=example-fastify-client node --require ./opentelemetry.js ./client.js

starting
status: OK
shutdown may only be called once per MeterProvider
```

Apparently `beforeExit` is called twice, so `sdk.shutdown()` gets called twice. I don't think this is something that needs to be solved for this PR.  My guess is that the `sdk.shutdown()` starts another async task (the requests made as part of flushing), which puts Node.js through its `beforeExit` handling again. Then the second `sdk.shutdown()` just warns and does nothing.

examples/fastify/package.json
pichlermarc: They were, yes. This is to prevent any problems with mixing experimental and stable packages. The example is not part of the workspace, so it should not hold any packages back for the rest of the repo :slightly_smiling_face:

JamieDanielson (approved):
Thanks for updating this!

pichlermarc:
> Apparently beforeExit is called twice, so sdk.shutdown() gets called twice. I don't think this is something that needs to be solved for this PR. My guess is that the sdk.shutdown() starts another async task (the requests made as part of flushing), which puts Node.js through its beforeExit handling again. Then the second sdk.shutdown() just warns and does nothing.

Yes, thanks for bringing that up - I'll look into it in a follow up. :slightly_smiling_face:

PR #1902, opened 2024-01-16
feat(perf_hooks): add node instrumentation perf_hooks

d4nyll:

#1106 - currently there is no Otel way to expose metrics provided by the Node.js runtime

(...continuation from #1272 and #1689)

## Description of changes

This PR adds a `@opentelemetry/instrumentation-perf-hooks` package that exposes the `utilization` measurement from Node.js Performance measurement APIs' [`performance.eventLoopUtilization`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2) method.

The exposed metric will look something like this:

```
# HELP nodejs_performance_event_loop_utilization Event loop utilization
# UNIT nodejs_performance_event_loop_utilization 1
# TYPE nodejs_performance_event_loop_utilization gauge
nodejs_performance_event_loop_utilization 0.010140079547955264
```

## Notes / Questions

- I've named this package `@opentelemetry/instrumentation-perf-hooks` as this initial version only exposes one metric. Happy to use `@opentelemetry/instrumentation-runtime` but that implies a bigger scope than what this PR does.
- The exposed metric is called `nodejs.performance.event_loop.utilization`, but perhaps `nodejs.event_loop.utilization` is sufficient?


pichlermarc (approved):
Thanks for working on this :slightly_smiling_face:

d4nyll:
Updated to fix the failed `release-please-config` and `peer-api-checks` workflows

d4nyll:
I believe the [unit-test (18.18.2)](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/7694860673/job/20968512920) failed due to the test being flakey (c0ef756ac716fe477b6b750db8aa704dba2dfd83 should have fixed it)

However, I am struggling to see why [unit-test (14)](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/7694860673/job/20968512466) hangs and I don't see any test output. (would 928f8ac9ef9759725d8c27ad39e79dfe3b066b38 have fixed it? It's currently passing on my machine using Node.js v14.21.3)

@pichlermarc can we run the tests again?

d4nyll:
@pichlermarc Hey, just want to double check if there are still changes that needs addressing, or we are just waiting for this to be included in the next release. Should I keep merging/resolving conflicts (which will cause the tests to be rerun) or just leave it be? Thanks

pichlermarc:
> @pichlermarc Hey, just want to double check if there are still changes that needs addressing, or we are just waiting for this to be included in the next release. Should I keep merging/resolving conflicts (which will cause the tests to be rerun) or just leave it be? Thanks

Sorry for the delay. I should've merged it long ago. Running tests again and then this should be good to merge.

trentm:
(Sorry this Q comes after this is reviewed and merged.)

Is this something that needs to be implemented as an *instrumentation*? Could it not be something like host-metrics (https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/opentelemetry-host-metrics/README.md#usage) where one separately creates and starts it after a MeterProvider has been registered (via using NodeSDK or whatever).   IIUC, this isn't actually instrumenting `perf_hooks`, but just using it.

d4nyll:
> IIUC, this isn't actually instrumenting `perf_hooks`, but just using it.

Yes, this is only _using_ `perf_hooks` but not _instrumenting_ the `perf_hooks` Node.js API.

> Is this something that needs to be implemented as an instrumentation?

The purpose of the package is to export metrics exposed by the Node.js runtime. I see similar packages in other languages (e.g. [Go](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/runtime)) exposing the language runtime internals implemented as instrumentation packages.

If it is a matter of naming, I would be happy for it to be renamed to `@opentelemetry/instrumentation-runtime`.

If it is a matter of whether it should be a non-instrumentation package, I would need more guidance from the maintainers as to the distinction between an 'instrumentation' package and other types of packages. From my perspective, yes, it could work as a package similar to [`@opentelemetry/host-metrics`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/opentelemetry-host-metrics/)

trentm:
@d4nyll Thanks. I moved my Q to a new issue (https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1956) so this doesn't get lost on a merged PR.

Issue #1778, opened 2023-11-08
Update and document instrumentation semantic conventions


dyladan:
Many packages in OTel JS Contrib export telemetry with out of date semantic conventions. Unfortunately, we have not clearly documented the version of semantic conventions implemented or which attributes are exported for instrumentations. For each package, we need to document the telemetry exported and if possible update it to the latest semantic conventions following the following template:

```md
This package implements the <xxxx> semantic convention version X.Y: <link to semantic convention>

Attributes collected:

| Attribute | Short Description | Notes |
| --------- | ----------------- | ----- |
|           |                   |       |
```

Please limit PRs to a single package in order to make them quickly and easily reviewable.
If you are working on one of the below packages, please comment so others know the issue is being worked on.

- [ ] `@opentelemetry/instrumentation-amqplib`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-cucumber`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-dataloader`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-fs`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-lru-memoizer`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-mongoose`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-socket.io`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-tedious`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-aws-lambda`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-aws-sdk`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
- [ ] `@opentelemetry/instrumentation-bunyan`
    - [ ] Semantic conventions up to date
    - [ ] README documents exported telemetry
[...]

PR #1907, opened 2024-01-18
chore(instr-express): add semconv to readme

JamieDanielson:

- Updates #1778 

- In README, add version of Semantic Conventions used in express instrumentation. See more details re: version in the issue
- In README, add semantic convention attributes collected


JamieDanielson:
As mentioned on the issue, I'm hoping to get feedback on this to see if it satisfies the goals of #1778 or if there are pieces to this that are missing. For example, I'm not sure if attributes listed are intended to just be those from semantic conventions, or if all attributes on an instrumented Express span should be listed here.

MSNev (approved):

pichlermarc (approved):

pichlermarc:
> As mentioned on the issue, I'm hoping to get feedback on this to see if it satisfies the goals of #1778 or if there are pieces to this that are missing. For example, I'm not sure if attributes listed are intended to just be those from semantic conventions, or if all attributes on an instrumented Express span should be listed here.

I think the goal there is to just to the actual semconv ones, so this PR satisfies the goals of that issue :slightly_smiling_face:

Issue #781, opened 2021-12-08
[postgres, mysql, mysql2] Update database connection semantic conventions


dyladan:
It appears it is just drastically outdated. The spec at the time of writing the postgres instrumentation was this: https://github.com/open-telemetry/opentelemetry-specification/blob/c8109895a989480028b282fa796ed87ad1421b8a/specification/data-database.md which specified the JDBC substring.

The recent spec is quite different and requires:

- `db.system` - `postgresql `
- One of `net.peer.name` OR `net.peer.ip` - hostname or ip respectively
- `net.peer.port` only if using nonstandard port
- `net.transport` recommended in general, required only if using in-process - tcp or inproc

`db.connection_string` is not required but seems to be the replacement for the jdbc string.

This most likely also affects the mysql and mysql2 instrumentations

_Originally posted by @dyladan in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/473#issuecomment-989205732_

Just-Sieb:
I am willing to take it on for mysql2. I just made another PR for the packages and this was another improvement I was looking at making for work. Although because it is on the unmaintained list there is not anyone to approve it. I am willing to be a maintainer to help out the package as we use it heavily at my work. Although that is not as beneficial if I am the one making the change and reviewing it.

maryliag:
closing in favour of https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2953

PR #1916, opened 2024-01-28
fix(net.peer.port): net.peer.port needs to be a number not a string

esara:
I noticed that a number of contrib instrumentation populated the net.peer.port as a string parsed out of the connection string. This not only violates the semantic convensations, but in same cases causes and exception the the span attributes are parsed. https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/database.md#connection-level-attributes

This is related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/781
More specifically, there is an inconsistency, where the port variable is a string in multiple instrumentation (memcached, mongodb, mysql/mysql2, mongodb) for example
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts#L802
compared to other instrumentation, where is correctly defined as an int (cassandra and others)
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-cassandra/src/instrumentation.ts#L179

Making sure that the net.peer.port attribute is a number instead of a string

BEGIN_COMMIT_OVERRIDE
fix(instrumentation-memcached): parse attribute value for net.peer.port to a number
fix(instrumentation-mongodb): parse attribute value for net.peer.port to a number
fix(instrumentation-mysql): parse attribute value for net.peer.port to a number
fix(instrumentation-mysql2): parse attribute value for net.peer.port to a number
END_COMMIT_OVERRIDE


br-kearns (approved):

pichlermarc (approved):
Looks good! :slightly_smiling_face: 
Thanks for taking care of this. :slightly_smiling_face:

pichlermarc (changes requested):
Ah looks like the tests also need to be updated (they assert that the values are `string`s, instead of `number`)

esara:
> Ah looks like the tests also need to be updated (they assert that the values are `string`s, instead of `number`)

@pichlermarc let me know if the test look better now

pichlermarc:
@esara, please don't force-push. It makes it hard to review the PR because the history disappears every time I check back :sweat_smile: 

(also makes it hard to see if the tests were passing as I have to dig into old workflow runs to find the run that was associated with the commit before the force-push)

pichlermarc (approved):

Issue #1917, opened 2024-01-31
updating `@opentelemetry/*` deps is currently problematic


trentm:
Updating `@opentelemetry/*` deps in the contrib repo is currently problematic.
There are two kinds of updates here:

1. Updating `@opentelemetry/*` deps that come from the core-repo (opentelemetry-js.git). This is typically done in a PR created after a release of packages from the core repo. In the past this was done via renovatebot, but since the switch to npm workspaces this has been problematic.
2. Updating `@opentelemetry/*` deps that live in this same contrib-repo (internal deps). For example some packages depend on `@opentelemetry/contrib-test-utils` for testing; `@opentelemetry/auto-instrumentations-node` depends on all the `instrumentation-*` deps that live in this repo. This update is typically done by a release PR (handled via the release-please tool).

# problems

The initial motivating problem was that attempts by renovatebot to do an update of OTel deps from the core repo was failing to update everything. For example, an attempt to update deps to versions 0.47.0/1.20.0 would leave parts of the contrib repo with deps at 1.19.0. Mixed 1.19.0 and 1.20.0 deps would cause breakages, e.g. type mismatches. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1905 and then the follow-up https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1906.

However, there are other issues, that I think stem from related causes.
Here are two examples of current silly/problematic behaviour.

First, updating the `@opentelemetry/instrumentation-bunyan` dep in its "examples/" package **crashes npm**.
In a clone of the contrib repo at the current "main" (commit 497a3c3c8c7d72f009c68c27d424f0d039a03df4):

```
[17:44:12 trentm@pink:~/src/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-bunyan/examples (git:main)]
% npm outdated
Package                                Current  Wanted  Latest  Location                                                                                              Depended by
@opentelemetry/instrumentation-bunyan   0.34.1  0.34.1  0.35.0  plugins/node/opentelemetry-instrumentation-bunyan/node_modules/@opentelemetry/instrumentation-bunyan  examples@npm:bunyan-example@0.45.1

[17:44:18 trentm@pink:~/src/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-bunyan/examples (git:main rv:1)]
% npm install @opentelemetry/instrumentation-bunyan@0.35.0
[...]

PR #1928, opened 2024-02-07
chore: add unreleasing packages to release-please config so their local deps get bumped

trentm:
This is option 1 from #1917.

Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1917


scripts/check-release-please.mjs
```diff
@@ -29,35 +29,48 @@ const getProcessOutput = (cmd, args) => {
...
-console.log('config packages', configPackages);
-
-lernaPackages.forEach((relativeLocation) => {
-	logErrorIf(
-		!manifestPackages.has(relativeLocation),
-		`Could not find ${relativeLocation} in .release-please-manifest.json. If you are adding a new package. Add following
-  "${relativeLocation}": "0.0.1",`);
+// console.log('lerna list', lernaList);
+// console.log('manifest packages', manifestPackages);
+// console.log('config packages', configPackages);
```
dyladan, line 49: Did you mean to check in commented code?

dyladan (commented):
I think you might also need to add it to the manifest file

scripts/check-release-please.mjs
trentm: Kinda/sorta?  These *had* been outputs of the script before my changes here. I'll remove them.

trentm:
> I think you might also need to add it to the manifest file

My initial naive guess was that with `skip-github-release: true` for these packages that we would not need to add these to the manifest. That was from @pichlermarc's comments here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1917#issuecomment-1927338695
and the docs on `skip-github-release` here: https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#configfile

FWIW, here is a `--dry-run` run of `release-please ... release-pr`: https://gist.github.com/trentm/b1a08cdc264b20e054e282d58f719825
I'm not sure from the copious output whether that is a successful test. It didn't blow up on not having the new packages in the manifest at least. However, I ran it against *my fork*, so it may be of limited value.

pichlermarc (approved):
Looks good, thanks for taking care of this :+1:

pichlermarc:
> > I think you might also need to add it to the manifest file
> 
> My initial naive guess was that with `skip-github-release: true` for these packages that we would not need to add these to the manifest. That was from @pichlermarc's comments here: [#1917 (comment)](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1917#issuecomment-1927338695) and the docs on `skip-github-release` here: https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#configfile

I also think that should be fine if we don't add these to the manifest file. Let's try without it, adding it later should be an easy fix if it's needed.

Issue #1820, opened 2023-11-23
[opentelemetry/auto-instrumentations-web 0.34.0] Update zone.js peer dependency


ebussieres:
I'm wondering why zone.js peer dependency is hardcoded to version 0.11.4. Why caret was not used ? Change was introduced in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1768.

I have an angular 17 app and i'm using zone.js 0.14.x version. This is preventing a lot of app to use the latest @opentelemetry/auto-instrumentations-web version.

dyladan:
Added the `enhancement` label and modified the PR title. This seems to be a request to support other versions of zone.js. There is a similar issue for the zone context manager in the core repo.

starushykart:
Hi, guys!
Many thanks for your work!
Is there an understanding of when the issue will be resolved?

pichlermarc:
@starushykart no ETA at the moment as no one is actively working on it. PRs from volunteers are welcome :slightly_smiling_face:

PR #1926, opened 2024-02-06
[opentelemetry/auto-instrumentations-web 0.34.0] Update zone.js peer dependency

HenryCorn:
## Which problem is this PR solving?

- Solves: [Update zone.js peer dependency](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1820)

## Short description of the changes

- Updated the version of zone.js in auto-instrumentations-web and dependencies

PR #1930, opened 2024-02-08
chore(deps): Updated zone.js version to support latest Angular

AndrewAKovalenko:

- Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1820

- Update Zone.js version to support Angular 17

### DIFF
--- a/metapackages/auto-instrumentations-web/package.json
+++ b/metapackages/auto-instrumentations-web/package.json
@@ -32,7 +32,7 @@
   },
   "peerDependencies": {
     "@opentelemetry/api": "^1.3.0",
-    "zone.js": "0.11.4"
+    "zone.js": "^0.11.4"
   },
   "devDependencies": {
     "@babel/core": "7.22.17",
--- a/metapackages/auto-instrumentations-web/package.json
+++ b/metapackages/auto-instrumentations-web/package.json
@@ -32,7 +32,7 @@
   },
   "peerDependencies": {
     "@opentelemetry/api": "^1.3.0",
-    "zone.js": "^0.11.4"
+    "zone.js": "^0.11.4 || ^0.13.0 || ^0.14.0"
   },
   "devDependencies": {
     "@babel/core": "7.22.17",
--- a/package-lock.json
+++ b/package-lock.json
@@ -386,7 +386,7 @@
       },
       "peerDependencies": {
         "@opentelemetry/api": "^1.3.0",
-        "zone.js": "0.11.4"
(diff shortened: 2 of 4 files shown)


AndrewAKovalenko:
This PR is duplicate of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1926, but my PR prevents from major `zone.js` version update, by using `^`

Also this PR is valuable for our team, so i'm always ready to make updates :)

metapackages/auto-instrumentations-web/package.json
```diff
@@ -32,7 +32,7 @@
   },
   "peerDependencies": {
     "@opentelemetry/api": "^1.3.0",
-    "zone.js": "0.11.4"
+    "zone.js": "^0.11.4"
```
pichlermarc, line 35: This will allow for `zone.js` at `>=0.11.4` and `<0.12.0` - is that intended?
According to SemVer minor versions in a `0.x` are considered breaking, therefore the caret does not do the same as with `1.x`.

If you're trying to go to `^0.14.0` (Angular 17), please open a PR against the core repo first (related issue https://github.com/open-telemetry/opentelemetry-js/issues/4445) - the dependency on `zone.js` here is transitive from there.

pichlermarc (commented):
Hi @AndrewAKovalenko, thanks for opening the PR. Some notes on the version range below. :slightly_smiling_face: 

Also, please sign the CLA, otherwise we won't be able to act on the PR.

metapackages/auto-instrumentations-web/package.json
AndrewAKovalenko: I see that you've already created PR https://github.com/open-telemetry/opentelemetry-js/pull/4469
Thanks

I'll fix my PR to use same versions, as you do `"zone.js": "^0.11.0 || ^0.13.0 || ^0.14.0"`

pichlermarc: Yes, that sounds good. 

Keep in mind, though, that we'll have to release the core repo first, then update the dependencies here, and once that's done we can merge this PR here. I'll take a few days.

AndrewAKovalenko: Sure, not a problem

AndrewAKovalenko:
@pichlermarc  i've created similar PR with correct CLA verification. I accidentally used wrong git.config to sign it :)

https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1934

Please, check it 

Closing this one

PR #1932, opened 2024-02-10
fix(knex): connection attrs missing if connectionString used

edosrecki:

- Connection related attributes (hostname, port, user, database) are missing if users use `connectionString` property when initializing `knex` (instead of specifying individual fields):

```js
const pg = require('knex')({
  client: 'pg',
  connection: {
    connectionString: config.DATABASE_URL,
  },
});
```

- Span name is also affected, for example here `database = undefined`:

Screenshot 2024-02-10 at 14 28 28

- If `connectionString` is provided in `knex` options, extract attributes from it (hostname, port, user, database).
- If **both** `connectionString` and individual fields are used, `connectionString` takes precedence:

Screenshot 2024-02-10 at 15 43 55


plugins/node/opentelemetry-instrumentation-knex/src/utils.ts
```diff
@@ -88,3 +88,16 @@ export const extractTableName = (builder: any): string => {
...
 };
+
+export const parseConnectionString = (connectionString: string | undefined) => {
+  if (!connectionString) {
+    return undefined;
+  }
+  const url = new URL(connectionString);
+  return {
+    host: url.hostname,
+    port: parseInt(url.port || '5432', 10),
```
scottgreenup, line 99: I'm interested in this pull request getting merged, so I thought I'd add some notes.

It looks like `connectionString` is for both PostgreSQL and RedShift. It looks like the default port for RedShift is 5439.

I think you can check the URL protocol which should be `postgres:` or `redshift:` and set it to 5432/5439?

```
> (new URL("postgres://user:password@myhost/mydb")).protocol
'postgres:'
```

edosrecki: Thanks for the comment. Yes I can do that. However we cannot then cover the case of CockroachDB which uses `postgres` scheme, but has a different default port of 26257.

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -131,13 +131,19 @@ export class KnexInstrumentation extends InstrumentationBase<any> {
     return function wrapQuery(original: () => any) {
       return function wrapped_logging_method(this: any, query: any) {
         const config = this.client.config;
+        const connection = utils.parseConnectionString(
```
deejay1, line 134: Maybe we could pass the entire config.connection here and return the resulting values instead of having undefined checks in the attribute assignment further down?

trentm:
There is also https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2286
I haven't looked closely yet, but it is patching for a similar thing.
@deejay1 and others that use Knex: preferences for other PR?

deejay1:
I think this settles having a custom function returning the connection attributes, when we now have three different ways to configure them, what do you think @edosrecki ?

I think it would be nice rolling this into this PR, I'd wait for @edosrecki to respond, if I'll make both changes it into a separate PR.

maryliag:
Hi @edosrecki , thank you for your contribution. Recently the Database Semantic Conventions was marked as stabled, and with this a few changes came, for example, there are no longer attributes for Connection String, DB User, DB name, etc. And new attributes were added, such as DB namespace (that should include the removed attributes). 
[Here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md) you can find a details doc about migrating to the stable convention.

Because of those changes, we don't want to add or update attributes that are no longer stable and should be removed instead. For this reason, I'll close this PR.

You can still open a new PR focusing only on fixes related to the stable convention, and you can also help out with the migration (which would be very much appreciate it and I would be happy to review), but I can see that a lot has already been done [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2671), so you can check if any attributes are missing. [Here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2881/files) is an example of how I'm doing this migration for the PG package.

Feel free to reach out to me if you have any questions!

PR #1946, opened 2024-02-16
feat: Host Telemetry: Option to filter Host/Process metrics

marcsst:

- Running host-telemetry always adds 5 kinds of metrics (HostCPU, HostRAM, Network, ProcessCPU, ProcessRAM). Depending on your hardware this can add a lot of metrics you don't necessarily need.

-  This PR adds an optional meterScope setting that can be set to only return Host (HostCPU, HostRAM, Network) or Process (ProcessCPU, ProcessRAM) metrics as desired. To avoid breaking changes this setting can be left as undefined and will still return all.


pichlermarc:
@marcsst these metrics can be dropped using the `View` feature of the SDK. Is that what you're looking for?

See https://opentelemetry.io/docs/languages/js/instrumentation/#configure-metric-views

PR #1949, opened 2024-02-21
docs(instr-knex): Fix knex instrumentation example

rtpg:
The instrumentation config is just a type, not a class


trentm (approved):

rtpg:
@trentm this is now done!

PR #1938, opened 2024-02-14
chore: remove the non-publishing packages from the npm workspace

trentm:
Because they cause difficulties with updating `@opentelemetry/*` dependencies.
This is "Option 3" from #1917

Refs: #1917

# summary of changes

- move all `plugins/node/opentelemetry-instrumentation-FOO/examples/` to `examples/FOO` and fix up some details to get them working.
- move `packages/opentelemetry-sampler-aws-xray` to `incubator/opentelemetry-sampler-aws-xray/`
- These packages are now NOT included in the monorepo workspace.
- Removed `npm run compile:examples`.
- Removed entries for these packages from release-please config and manifest.

# checklist (done)

- [x] Undo "option 1" work from #1928
- [x] Q: is "incubator/" dir a reasonable name? Or put sampler-aws-xray in "archive/"?
- [x] move plugins/node/instr-FOO/examples to top-level examples/FOO dir. (Note that this *reverses* the intent from an effort 1-2y ago to move from examples/ to the instrumentation dirs.)
- [x] Try to confirm that my 'update-otel-deps.js' script now works better at updating deps in this repo.
- [x] manually test each of those examples
- [x] clean up intent in examples/README.md
- [ ] ~start an issue for working through modernizing or dropping examples~
- [ ] ~document expectation/procedure for maintaining examples~

PR #1955, opened 2024-02-23
fix(instr-perf-hooks): avoid 'prepare' npm script because that runs on 'npm install'

trentm:
Having a prepare script in the monorepo means that it runs during 'npm ci' which can break if some needed deps for that script haven't yet been installed.

# details

This is the same change that was part of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1771
From that PR's description:

> s/prepare/prepublishOnly/g is to avoid running the compile step in packages before npm install for all packages has completed. It resulted in various build errors. Using prepublishOnly is recommended (https://docs.npmjs.com/cli/v10/using-npm/scripts#prepare-and-prepublish) and is what the core repo was already doing.

I hit a breakage in `npm ci` in a separate PR (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1938). A few of the checks started failing:

Screenshot 2024-02-23 at 11 44 43 AM

E.g.:  https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8023208091/job/21919224572?pr=1938 shows that `instrumentation-perf-hooks`'s prepare and compile scripts were running during `npm ci`:

```
npm ERR! > @opentelemetry/instrumentation-perf-hooks@0.1.0 compile
npm ERR! > tsc -p .
npm ERR! 
npm ERR! node_modules/@types/node/ts4.8/globals.d.ts:6:76 - error TS2307: Cannot find module 'undici-types' or its corresponding type declarations.
npm ERR! 
npm ERR! 6 type _Request = typeof globalThis extends { onmessage: any } ? {} : import("undici-types").Request;
npm ERR!                                                                              ~~~~~~~~~~~~~~
npm ERR! 
...
```

I don't fully understand the type issue here. I think a contributing factor is that the instrumentation-perf-hooks package has a dependency on a *newer* version of `@types/node` than is used in all other packages in this repo:

```
...
plugins/node/opentelemetry-instrumentation-graphql/package.json
51:    "@types/node": "18.6.5",

plugins/node/instrumentation-perf-hooks/package.json
48:    "@types/node": "^20.11.2",

plugins/node/opentelemetry-instrumentation-redis/package.json
58:    "@types/node": "18.6.5",
...
```

@d4nyll Do you happen to know if the newer `@types/node` is required for the types used in instrumentation-perf-hooks?

### DIFF
--- a/plugins/node/instrumentation-perf-hooks/package.json
+++ b/plugins/node/instrumentation-perf-hooks/package.json
@@ -13,7 +13,7 @@
     "lint:fix": "eslint . --ext .ts --fix",
     "precompile": "tsc --version && lerna run version:update --scope @opentelemetry/instrumentation-perf-hooks --include-dependencies",
     "prewatch": "npm run precompile",
-    "prepare": "npm run compile",
+    "prepublishOnly": "npm run compile",
     "test": "nyc ts-mocha -p tsconfig.json 'test/**/*.test.ts'",
     "version:update": "node ../../../scripts/version-update.js"
   },


trentm:
> I don't fully understand the type issue here. I think a contributing factor is that the instrumentation-perf-hooks package has a dependency on a _newer_ version of `@types/node` than is used in all other packages in this repo:

Actually, I'm not sure the usage of a separate `@types/node` version is at all related to build failure for which I made this PR. So the question about the needed `@types/node` version is an unrelated question.

d4nyll:
I mostly copied files from existing packages for `@opentelemetry/instrumentation-perf-hooks`.

> s/prepare/prepublishOnly/g

I think I must have copied an older version of another package that used `prepare`, there's no reason for `prepare` over `prepublishOnly`

> `@types/node`

All of the dependencies in `devDependencies` (`@types/mocha`, `nyc`, etc.) can be aligned with the rest of the repo. I just didn't think it mattered. The only version that matters is the Node.js version specified in `engines`.

So this change LGTM.

Flarna (approved):

trentm:
> All of the dependencies in `devDependencies` (`@types/mocha`, `nyc`, etc.) can be aligned with the rest of the repo. I just didn't think it mattered. The only version that matters is the Node.js version specified in `engines`.

Thanks. I'll merge this now, and leave aligning the versions to a separate PR.

trentm:
> and leave aligning the versions to a separate PR.

Done in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1967, FYI.

PR #1957, opened 2024-02-25
chore(instrumentation-hapi): setup tav

nlochschmidt:

As mentioned in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1491#issuecomment-1547968125, the hapi plugin isn't setup for testing all supported versions of hapi. 

adds tav config for testing against the latest of each currently supported major version of `hapi`


plugins/node/opentelemetry-instrumentation-hapi/.tav.yml
```diff
@@ -0,0 +1,4 @@
+"@hapi/hapi":
+  - versions: "^20.3.0 || ^19.2.0 || ^18.4.1 || ^17.9.0"
+    node: ">=14"
```
trentm, line 3: I'm not sure if we'll hit it here (I haven't looked at the current instrumentation-hapi tests), but my understanding is that `@hapi/hapi` versions 17 and 18 and 19 will crash when handling an HTTP POST with node version v16 or later.  IOW, these older hapi versions only support up to `node <=16.0.0`.

So *potentially* we'd want something like this (untested):

```
"@hapi/hapi":
  - versions: "^19.2.0 || ^18.4.1 || ^17.9.0"
    node: ">=14 <16"
    commands: npm test
  - versions: "^20.3.0"
    node: ">=14"
    commands: npm test
```

nlochschmidt: Good catch. I've only started using hapi in v20.

Tests have been passing in node 16 on my machine. I assume this is because the tests are using `server.inject` instead of going through an actual http request, and they only do `GET` requests at the moment.

Should we merge with the change you proposed or would you want the tests to be changed to real http requests?

trentm: For future reference, I *think* the relevant hapi issue is https://github.com/hapijs/hapi/pull/4225 (this from my notes on earlier work on hapi tests here: https://github.com/elastic/apm-agent-nodejs/blame/main/.tav.yml#L286-L318

> Should we merge with the change you proposed or would you want the tests to be changed to real http requests?

It looks like you have take my suggestion. (If you can, please avoid *force* pushes to PR branches that have any reviews on them. It makes it harder to see the incremental changes.)

I think that's good for now. I don't know instrumentation-hapi well enough to have a strong opinion on changing the tests.

trentm (approved):

plugins/node/opentelemetry-instrumentation-hapi/.tav.yml
nlochschmidt: Great!

Sorry about the force-push. I am just very used to amending fixes 😅. I will not do this for future PRs to this repo 🙏

Issue #1960, opened 2024-02-26
flaky instrumentation-runtime-node test


trentm:
I believe the 'should export event loop utilization metrics after eventLoopUtilizationMeasurementInterval' test case in https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4483a326bff972271f3116cc78e52ab76cf7a727/plugins/node/instrumentation-perf-hooks/test/event_loop_utilization.test.ts#L94 may be flaky.

In the following unit-test run for PR #1938, that test assert failed:

```js
    assert.strictEqual(metrics[0].dataPoints[0].value < 1, true);
```

https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8026238361/job/21928362946?pr=1938

```
> @opentelemetry/instrumentation-perf-hooks@0.1.0 test
> nyc ts-mocha -p tsconfig.json 'test/**/*.test.ts'



  nodejs.event_loop.utilization
    ✓ should stop exporting metrics when disabled
    ✓ should not export immediately after enable
    ✓ can use default eventLoopUtilizationMeasurementInterval
    1) should export event loop utilization metrics after eventLoopUtilizationMeasurementInterval


  3 passing (84ms)
  1 failing

  1) nodejs.event_loop.utilization
       should export event loop utilization metrics after eventLoopUtilizationMeasurementInterval:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

false !== true

      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/event_loop_utilization.test.ts:94:12)
      at runNextTicks (internal/process/task_queues.js:60:5)
      at processTimers (internal/timers.js:497:9)
```

So I wonder if the new nodejs.event_loop.utilization metric can unexpectedly be higher than 1.
I have not been able to reproduce this failing locally.

I have only seen the failure that one time in CI. Mostly I'm just noting it for now. I'll have a PR to add some more output in those asserts in case that might help; but otherwise, I think we can wait and see if the test fails at all frequently before spending a lot of time on it.

/cc @d4nyll

pichlermarc:
I'll look into it.

PR #1961, opened 2024-02-26
chore(instr-perf-tools): add debugging output to some test asserts

trentm:
This is to give more detail if those asserts fail. I suspect the
'val < 1' test is flaky.

Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1960


pichlermarc (approved):
Thanks!

Sorry about that - I should've noticed that when reviewing the PR that added the perf-tools instrumentation. 
I'll look into if it's possible to refactor the tests using a `MetricReader` that's not the `PeriodicExportingMetricReader`. That may be part of why the test is flaky.

PR #1963, opened 2024-02-26
feat(instrumentation-fs): error hook

raphael-theriault-swi:

This adds a new error hook to the `fs` instrumentation to let users filter some errors from being reported as span events. This prevents trace data being polluted by irrelevant error events and failed spans for idiomatic filesystem access patterns where an error is expected, for instance opening a file without checking whether it exists then dealing with that case in the error handler.

- Adds an optional `errorHook` option to the config that is called with the function name and error before reporting the error event. If the hook is present and returns `false`, the error event is not reported.
- The error is still always propagated

### DIFF
--- a/plugins/node/instrumentation-fs/README.md
+++ b/plugins/node/instrumentation-fs/README.md
@@ -40,11 +40,12 @@ registerInstrumentations({
 
 You can set the following:
 
-| Options             | Type                                                                                                | Description                                                                                                     |
-| ------------------- | --------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
-| `createHook`        | `(functionName: FMember \| FPMember, info: { args: ArrayLike<unknown> }) => boolean`                | Hook called before creating the span. If `false` is returned this and all the sibling calls will not be traced. |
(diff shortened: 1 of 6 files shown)


raphael-theriault-swi:
@open-telemetry/javascript-approvers would it be possible for anyone to take a look ?

pichlermarc:
@raphael-theriault-swi - it looks like this component is unmaintained, so this PR has been sitting for a while. Since it is an unmaintained component, this feature would need a sponsor. Is this still something you're interested in working on? 🤔 

(side-note: IMO this instrumentation is on course for removal from `@opentelemetry/auto-instrumentations-node` and then the repo as it's unmaintained and too verbose by default)

pichlermarc:
(removed Rauno from the assignees as he does not work on this project anymore)

raphael-theriault-swi:
@pichlermarc my bad thought I had replied to this. I need to gauge whether it's worth it for us (as in SW) for me to own the instrumentation as opposed to just leaving it to its fate given most people leave it disabled since its super verbose as you mentioned. Will bring it up during a SIG meeting.

PR #1965, opened 2024-02-27
fix(instrumentation-runtime-node): use TestMetricReader, unref timer

### LINKED ISSUE #1960: flaky instrumentation-runtime-node test

pichlermarc:

- tests were failing as the utilization ended up being `1` (#1955) . This might have happened as the collection happened at a time where event loop was fully utilized. This PR attempts to fix this by doing two things:
  - changing the `PeriodicExportingMetricReader` to avoid having to wait for collection, getting rid of one variable that might break tests 
  - allowing `<=1` as a value as it's a valid value for a fully utilized event loop (which can happen in some cases as collection does not need to wait for anything that's not doing work on the event loop).
- for me the process kept running as `disable()` was never called when the test failed. I can see this happening in apps too as it's rather uncommon to `disable()` instrumentations before shutting down. Therefore in `unref`ed the timer when it's created.

Fixes #1960


plugins/node/instrumentation-perf-hooks/test/event_loop_utilization.test.ts
```diff
@@ -102,6 +127,18 @@ describe('nodejs.event_loop.utilization', () => {
...
+      'expected default unit'
+    );
+    assert.strictEqual(
+      metrics[0].dataPoints.length,
+      1,
+      'expected one data point'
+    );
+    const val = metrics[0].dataPoints[0].value;
+    assert.strictEqual(val > 0, true, `val (${val}) > 0`);
+    assert.strictEqual(val < 1, true, `val (${val}) < 1`);
```
d4nyll, line 142: > alternately, we can also allow <=1 as a value (because it's a valid value)

Having read the documentation again, I do think the original reason why the tests were flakey was because of the `<`. So I think it doesn't hurt to changed this to `<=`

d4nyll:
> replacing interval-based collection with on-demand collection

@pichlermarc thanks for cleaning up these tests, they look much sturdier now

> `unref`

Good call 👍

plugins/node/instrumentation-perf-hooks/test/event_loop_utilization.test.ts
pichlermarc: Done in [5a40557](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1965/commits/5a40557618cee84e27414f0e76e73b95a1347e55) :slightly_smiling_face:

plugins/node/instrumentation-runtime-node/test/event_loop_utilization.test.ts
```diff
@@ -14,85 +14,110 @@
...
+      eventLoopUtilizationMeasurementInterval: MEASUREMENT_INTERVAL,
+      enabled: false,
+    });
+    instrumentation.setMeterProvider(meterProvider);
+
+    // act
+    await new Promise(resolve => setTimeout(resolve, MEASUREMENT_INTERVAL * 5));
+    const { resourceMetrics, errors } = await metricReader.collect();
+
+    // assert
```
trentm, line 59: TIL: the arrange-act-assert pattern name (http://wiki.c2.com/?ArrangeActAssert)

trentm (approved):

Issue #1942, opened 2024-02-15
coordinate first pass at ESM support for all instrumentations


trentm:
This is a meta issue to coordinate doing a first pass at ECMAScript module (ESM) support and a test for every node.js instrumentation. (This is about working with the `import-in-the-middle`-based hook so is not relevant to browser instrumentations.)  I'm not necessarily volunteering to *do* all these :), but I'll certainly help.

Hopefully this can also be useful for support/help requests for why instrumentation might not be working for ESM-using code. If the module/instrumentation in question is not checked off here, then ESM instrumentation is not expected to work (it might be luck).

- [ ] instrumentation-grpc (core repo)
- [x] instrumentation-http (core repo, [support + test](https://github.com/open-telemetry/opentelemetry-js/pull/4866))
- [ ] instrumentation-amqplib
- [ ] instrumentation-cucumber
- [ ] instrumentation-dataloader ([support](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2813))
- [ ] instrumentation-fs
- [ ] instrumentation-lru-memoizer
- [x] instrumentation-mongoose (#2793)
- [ ] instrumentation-socket.io
- [ ] instrumentation-tedious
- [ ] instrumentation-aws-lambda (support: https://github.com/open-telemetry/opentelemetry-js/pull/5094 and https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2508, missing test)
- [ ] instrumentation-aws-sdk
- [ ] instrumentation-bunyan
- [ ] instrumentation-cassandra
- [ ] instrumentation-connect
- [ ] instrumentation-dns
- [x] instrumentation-express ([test](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1966))
- [ ] instrumentation-fastify ([test](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1763); there is an ESM limitation, see below)
- [ ] instrumentation-generic-pool
- [ ] instrumentation-graphql
- [x] instrumentation-hapi ([support + test](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2184))
- [x] instrumentation-ioredis ([support](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1694), [test](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1735))
- [ ] instrumentation-knex
[...]

PR #1966, opened 2024-02-27
test(instr-express): add test for ESM

JamieDanielson:

- updates #1942 

- add test for ESM support in express instrumentation


pichlermarc (approved):
Looks good, thank you for working on this. :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -493,4 +494,44 @@ describe('ExpressInstrumentation', () => {
...
+        assert.strictEqual(spans[1].name, 'middleware - query');
+        assert.strictEqual(spans[1].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[1].parentSpanId, spans[0].spanId);
+        assert.strictEqual(spans[2].name, 'middleware - expressInit');
+        assert.strictEqual(spans[2].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[2].parentSpanId, spans[0].spanId);
+        assert.strictEqual(spans[3].name, 'middleware - simpleMiddleware');
+        assert.strictEqual(spans[3].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[3].parentSpanId, spans[0].spanId);
+        // assert.strictEqual(spans[4].name, 'router - /post/:id'); // TODO bug?
```
trentm, line 530: Yah, I think this is a bug, or limitation. I don't think the current express instrumentation handles usage of a `express.Router()`.   Perhaps change `use-express.mjs` for now to not use a Router and we add a separate issue about handling `Router` in instr-express?

plugins/node/opentelemetry-instrumentation-express/test/fixtures/use-express.mjs
```diff
@@ -0,0 +1,69 @@
...
+  // can be flaky.
+  await promisify(setTimeout)(10);
+  next();
+});
+
+const router = express.Router();
+router.get('/post/:id', (req, res) => {
+  res.send(`Post id: ${req.params.id}`);
+});
+app.use(router);
```
trentm, line 53: As mentioned above, perhaps change this to:

```suggestion
app.get('/post/:id', (req, res) => {
  res.send(`Post id: ${req.params.id}`);
});
```

trentm (approved):
Thanks! LGTM, with a possible suggestion to avoid the current Router issue/limitation.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
JamieDanielson: I agree, will update

JamieDanielson: Updated this test and created new issue for tracking #1993

PR #1967, opened 2024-02-27
chore(instr-perf-tools): update deps to similar versions as other contrib packages

trentm:
- '@opentelemetry/instrumentation' dep should match others in this repo
- drop the '@types/{mocha,node}' devDeps because I'm not sure they are needed
- reduce 'mocha' dep to v7 used by all other packages
- bump /api peer dep to 1.3.0 (the version when api-metrics was merged into api)

* * *

> - '@opentelemetry/instrumentation' dep should match others in this repo

I actually not sure if this is cargo-culting.

> - drop the '@types/{mocha,node}' devDeps because I'm not sure they are needed

Is this wrong? Basically all contrib packages include these two packages as devDeps. Are those `@types/...` packages needed for IDE usage or something? My "I'm not sure they are needed" is based on running `npm run compile && npm test`.


pichlermarc:
> I actually not sure if this is cargo-culting.
> 
> > * drop the '@types/{mocha,node}' devDeps because I'm not sure they are needed
> 
> Is this wrong? Basically all contrib packages include these two packages as devDeps. Are those `@types/...` packages needed for IDE usage or something? My "I'm not sure they are needed" is based on running `npm run compile && npm test`.

`@types/node` is needed for `assert`, and `@types/mocha` is needed for `describe`, `beforeEach`, ...

For me, at least, my IDE (JetBrains WebStorm) does not complain, but I think that's because it's being hoisted by other packages in the workspace. Removing the package from the workspace, adding lerna, and `npm install`-ing it causes it not to compile afterward. :thinking: 

We usually include all devDependencies explicitly in each package as we try to avoid depending on another package's hoisted dependencies as `devDependencies`. Another goal was to keep single packages `npm install`-able, speeding up the workflow when working on only one package. I'm not sure if all arguments still hold (I know working on a single package would require at least `lerna` to be installed which is not listed as a `devDependency` in packages, but I think that's the only exception to the rule at the moment).

That being said: the decision to do this was made a long time ago - happy to see that revised it if there's a better alternative. Doing sow would probably better suited for another PR, though. :slightly_smiling_face:

pichlermarc:
> bump /api peer dep to 1.3.0 (the version when api-metrics was merged into api)

:+1: 

> reduce 'mocha' dep to v7 used by all other packages

:+1: 

> '@opentelemetry/instrumentation' dep should match others in this repo

:+1:

trentm:
Replace by https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1994

PR #1969, opened 2024-02-28
docs(instrumentation-graphql): gql supported versions

overbit:

- Fix outdated documentation about supported graphql deps versions

- README update supported versions by the instrumentation (PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/998 should allow to also use GQL version 16, but doc still mentioned only <16).


pichlermarc (approved):
Thanks for updating this :+1: 

Code where this is defined, for reference: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/2c861e07c4861295401b0999ed705191947134fa/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts#L65

Issue #1956, opened 2024-02-23
should instrumentation-perf-hooks move to packages/node-runtime-metrics or similar?


trentm:
(I'm opening a separate issue so my Q on a *merged PR* doesn't get lost.
Some original discussion is here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1902#issuecomment-1961764089)

A new instrumentation was recently merged that uses [`perf_hooks.performance.eventLoopUtilization`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2) to provide a `nodejs.event_loop.utilization` metric.
I wonder if this could *not* be an instrumentation, but instead a package like "packages/opentelemetry-host-metrics".

Perhaps a name like `@opentelemetry/node-runtime-metrics` (see https://opentelemetry.io/docs/specs/semconv/runtime/#metrics)?

/cc @d4nyll, @pichlermarc 

If we think this would make more sense, it would be nice to do this before a first release of `@opentelemetry/instrumentation-perf-hooks`.

trentm:
> Perhaps a name like `@opentelemetry/node-runtime-metrics`

Using this name somewhat puts a stake in the ground for this package possibly growing support for reporting things like Node.js runtime GC, heap, etc. metrics.

pichlermarc:
> (I'm opening a separate issue so my Q on a _merged PR_ doesn't get lost. Some original discussion is here: [#1902 (comment)](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1902#issuecomment-1961764089))

Thanks for raising the issue. :+1: 
 
> A new instrumentation was recently merged that uses [`perf_hooks.performance.eventLoopUtilization`](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2) to provide a `nodejs.event_loop.utilization` metric. I wonder if this could _not_ be an instrumentation, but instead a package like "packages/opentelemetry-host-metrics".

Hmm, you're right, I think we have some inconsistency here. In my opinion having it be an instrumentation makes the most sense:
- it allows it to be enabled like an instrumentation (with `registerInstrumentations`) and users don't have to get used to another interface to start getting telemetry. 
- an instrumentation using diagnostics channels would usually also not wrap things, but I'd still be considered an instrumentation. I see a similar case can be made for metrics that can be produced without having to wrap anything.

[...]

PR #1970, opened 2024-02-28
refactor(instrumentation-perf-hooks): rename to instrumentation-runtime-node

pichlermarc:

This is my proposal for the name change, see #1956 

- renames the entire package instrumentation-perf-hooks to instrumentation-runtime-node
- this is not a breaking change as it has not been released yet
- adds an entry in the component label map so that labels are applied


plugins/node/instrumentation-runtime-node/README.md
```diff
@@ -1,4 +1,4 @@
-# OpenTelemetry Node.js Performance Hooks Instrumentation
+# OpenTelemetry Node.js Runtime Instrumentation
```
trentm, line 1: Would we want the title to make clear this is about runtime *metrics*?

trentm (approved):

plugins/node/instrumentation-runtime-node/README.md
pichlermarc: Good idea, added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1970/commits/dabe063d1cfe479ba1d8b219e34041c972493175

I also added a note that it may produce additional signals future (though I can really on think of metrics really making sense right now :thinking:)

trentm:
@pichlermarc Looking at this again, I suspect we want to change `PerfHooksInstrumentation` to `RuntimeNodeInstrumentation` or something.

pichlermarc:
> @pichlermarc Looking at this again, I suspect we want to change `PerfHooksInstrumentation` to `RuntimeNodeInstrumentation` or something.

@trentm opened another PR #1970

PR #1972, opened 2024-02-28
feat: Add the ability to use span links when consuming a message amqp plugin

McSick:

- Per the [spec](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans) on consumer spans, they should link to the message's creation context.  The current behavior is instead continuing the trace where the message was produced.  This can cause traces to be quite large for apps that produce a lot of messages. This PR offer's a config option to follow the spec more closely.

- A config option was added that when set to true, it follows the spec behavior to link to the produce context instead of continuing the context. The default behavior will be as it is today, so one has to opt-in to this change via the config setting.


McSick:
> I assume this was made a configuration because it is a breaking change. We will probably try to go to this as the default behavior in a future version when the messaging semconv stabilizes. For now, I think this is a decent compromise.
> 
> Can you please add tests for the new behavior?

I added in tests for the new option. And yeah agree with you in the future, likely we will switch over the default. This gives us an easy way to do that as well.  Let me know if you need anything else! I ran the linter and there are quite a few warnings that for code I did not touch.  If you need me to correct those, let me know.

McSick:
> > I ran the linter and there are quite a few warnings that for code I did not touch. If you need me to correct those, let me know.
> 
> Warnings are typically fine, especially if they're on code you didn't introduce. There are many hundreds across the codebase and requiring them to be fixed is a losing battle at this point. Mostly the linter just makes the code style consistent enough to make it easier to review.

Perfect! the code is ready to be reviewed then.

blumamir:
This PR touches on 2 things: 
- parent-child relationship between producer and consumer
- storing link to the producer on the consumer.

I wonder if they are mutually exclusive. 

Always recording a link to the producer makes sense to me, and is not a breaking change.

regarding the parent-child relationship - I can see how some users would want it one way and others will want it the other way (regardless of the default value). perhaps the new config option can address this property instead? something like `startNewTraceOnConsume`?

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -409,9 +410,18 @@ export class AmqplibInstrumentation extends InstrumentationBase {
         const headers = msg.properties.headers ?? {};
         const parentContext = propagation.extract(ROOT_CONTEXT, headers);
         const exchange = msg.fields?.exchange;
-        const span = self.tracer.startSpan(
-          `${queue} process`,
-          {
+        let span: Span;
+        if (self._config.useLinksForConsume) {
```
blumamir, line 414: the 2 cases in the if-else contains some duplication.

WDYT about preparing the `links` array only once (with 0 or 1 elements) and the `parent` based on the instrumentation configuration, then just invoking `self.tracer.startSpan` once with these values? or at least calculate the `attributes` once, outside the if-else

dyladan:
> This PR touches on 2 things:
> 
> * parent-child relationship between producer and consumer
> * storing link to the producer on the consumer.
> 
> I wonder if they are mutually exclusive.

I don't think they're exclusive. If you remove the parent-child link you lose all relationship without the link. If you add the link, it is a bit overlapping with the parent-child relationship and potentially confusing. Because this is opt-in, I think it is probably fine.

dyladan:
> > This PR touches on 2 things:
> > 
> > * parent-child relationship between producer and consumer
> > * storing link to the producer on the consumer.
> > 
> > I wonder if they are mutually exclusive.
> 
> I don't think they're exclusive. If you remove the parent-child link you lose all relationship without the link. If you add the link, it is a bit overlapping with the parent-child relationship and potentially confusing. Because this is opt-in, I think it is probably fine.

edit: Also, this is just implementing the currently specified behavior. It could be argued we don't have a choice.

plugins/node/instrumentation-amqplib/src/amqplib.ts
McSick: Would a null parent be fine in this instance?  Because with the config option, the parentcontext should not exists.  If that's fine, then I can make those changes.

dyladan: `undefined` is fine. It's the default anyway

dyladan: I'd do something like this

```typescript
        const links: Link[] = [];
        if (self._config.useLinksForConsume) {
          const parentSpanContext = trace.getSpan(parentContext)?.spanContext();
          if (parentSpanContext) {
            links.push({ context: parentSpanContext });
          }
        }

        const span = self.tracer.startSpan(`${queue} process`, {
          kind: SpanKind.CONSUMER,
          attributes: {
            ...channel?.connection?.[CONNECTION_ATTRIBUTES],
            [SemanticAttributes.MESSAGING_DESTINATION]: exchange,
            [SemanticAttributes.MESSAGING_DESTINATION_KIND]:
              MessagingDestinationKindValues.TOPIC,
            [SemanticAttributes.MESSAGING_RABBITMQ_ROUTING_KEY]:
              msg.fields?.routingKey,
            [SemanticAttributes.MESSAGING_OPERATION]:
              MessagingOperationValues.PROCESS,
            [SemanticAttributes.MESSAGING_MESSAGE_ID]:
              msg?.properties.messageId,
            [SemanticAttributes.MESSAGING_CONVERSATION_ID]:
              msg?.properties.correlationId,
          },
          links,
        },
        self._config.useLinksForConsume ? undefined : parentContext,
        );
```

dyladan (commented):
Left a suggestion to remove some duplication. Didn't test it (typed in browser) so it might need some tweaks

blumamir:
@McSick Thanks again for working on this.
Are you still interested in finalizing it? should we keep the PR open?

McSick:
@dyladan @blumamir , I went ahead and updated the code based on some feedback, tests also updated.  Let me know if it looks good.

plugins/node/instrumentation-amqplib/src/types.ts
```diff
@@ -96,10 +96,14 @@ export interface AmqplibInstrumentationConfig extends InstrumentationConfig {
    * Default is 1 minute
    */
   consumeTimeoutMs?: number;
+
+  /** Used to use a span link for the consume message instead of continuing a trace */
```
JamieDanielson, line 100:
```suggestion
  /** option to use a span link for the consume message instead of continuing a trace */
```
nit

JamieDanielson (approved):
Thanks for working on this! Generally this looks good to me. Left a few notes though non-blocking from me. Also sorry for the churn but the readme needs to be updated based on some recent changes

plugins/node/instrumentation-amqplib/src/amqplib.ts
McSick: I had to change it a little as parentContext was used deeper in the code so I wanted to explicitly make it undefined in this instance so the trace was not continued.

McSick:
@JamieDanielson , went ahead and updated based on your suggestions and updated the readme to resolve the conflict.

plugins/node/instrumentation-amqplib/README.md
```diff
@@ -69,6 +71,34 @@ If timeout is not big enough, span might be closed with 'InstrumentationTimeout'
...
+This instrumentation was originally published under the name `"opentelemetry-instrumentation-amqplib"` in [this repo](https://github.com/aspecto-io/opentelemetry-ext-js). Few breaking changes were made during porting to the contrib repo to align with conventions:
+
+### Hook Info
+
+The instrumentation's config `publishHook`, `publishConfirmHook`, `consumeHook` and `consumeEndHook` functions signature changed, so the second function parameter is info object, containing the relevant hook data.
+
+### `moduleVersionAttributeName` config option
+
+The `moduleVersionAttributeName` config option is removed. To add the amqplib package version to spans, use the `moduleVersion` attribute in hook info for `publishHook` and `consumeHook` functions.
+
```
JamieDanielson, line 91:
```suggestion
```

I think something went wrong in a merge from the remote branch, and this should be removed as it's no longer on main.

JamieDanielson (commented):
@McSick Hmm I think something got weird with a merge from a remote branch somewhere along the way here... maybe [this commit](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1972/commits/e2e0160660ef43cf1d1827e6e7a9494c4723b66a)? I found one section of the README that should be removed, suggestion highlighted. I don't think there is anything in the other files but I need to take another look through.

JamieDanielson:
> @McSick Hmm I think something got weird with a merge from a remote branch somewhere along the way here... maybe [this commit](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1972/commits/e2e0160660ef43cf1d1827e6e7a9494c4723b66a)? I found one section of the README that should be removed, suggestion highlighted. I don't think there is anything in the other files but I need to take another look through.

I went through and did some branch cleanup to check things out and it seems like this README change is the only change (aside from whitespace)... so I committed that and the tests are running. I included the git diff for [my branch](https://github.com/JamieDanielson/opentelemetry-js-contrib/tree/jamie.mcsick-feature/config-links-amqp-1972-merge-fixes) and yours below.

<details>
<summary>
git diff
</summary>

```diff
diff --git a/plugins/node/instrumentation-amqplib/README.md b/plugins/node/instrumentation-amqplib/README.md
index dcfac7fb..dd211d3a 100644
--- a/plugins/node/instrumentation-amqplib/README.md
+++ b/plugins/node/instrumentation-amqplib/README.md
@@ -51,7 +51,7 @@ registerInstrumentations({
 amqplib instrumentation has few options available to choose from. You can set the following:

 | Options              | Type                                           | Description                                                                         |
-| --------------------------------- | ----------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| -------------------- | ---------------------------------------------- | ----------------------------------------------------------------------------------- |
 | `publishHook`        | `AmqplibPublishCustomAttributeFunction`        | hook for adding custom attributes before publish message is sent.                   |
 | `publishConfirmHook` | `AmqplibPublishConfirmCustomAttributeFunction` | hook for adding custom attributes after publish message is confirmed by the broker. |
 | `consumeHook`        | `AmqplibConsumeCustomAttributeFunction`        | hook for adding custom attributes before consumer message is processed.             |
@@ -77,18 +77,6 @@ By default, consume spans continue the trace where a message was produced. Howev

 Default is false

-## Migration From opentelemetry-instrumentation-amqplib
-
-This instrumentation was originally published under the name `"opentelemetry-instrumentation-amqplib"` in [this repo](https://github.com/aspecto-io/opentelemetry-ext-js). Few breaking changes were made during porting to the contrib repo to align with conventions:
-
-### Hook Info
-
-The instrumentation's config `publishHook`, `publishConfirmHook`, `consumeHook` and `consumeEndHook` functions signature changed, so the second function parameter is info object, containing the relevant hook data.
-
-### `moduleVersionAttributeName` config option
-
-The `moduleVersionAttributeName` config option is removed. To add the amqplib package version to spans, use the `moduleVersion` attribute in hook info for `publishHook` and `consumeHook` functions.
-
 ## Running Tests Locally

 To run the tests locally, you need to have a RabbitMQ server running.  You can use the following command to start a RabbitMQ server using Docker: 
```

</details>

JamieDanielson:
@dyladan @blumamir do you have any remaining concerns on this PR that haven't been addressed? If not I think this is good to go.

blumamir:
> @dyladan @blumamir do you have any remaining concerns on this PR that haven't been addressed? If not I think this is good to go.

Thank you @JamieDanielson , sorry for the delay, busy days at work 🥵
If you think it's good to go I am good with merging 👍

JamieDanielson:
The failed test is for browser and unrelated, looks like a flaky timeout issue. Rerunning just in case

PR #1982, opened 2024-03-01
fix(instrumentation-pg): prevent net.peer.port from being NaN

### LINKED ISSUE #1980: instrumentation-pg: Invalid trace output if port=null
API 1.8.0 / Stable 1.22.0 / Experimental 0.49.0
instrumentation-pg 0.38.0

v18.19

### What did you do?
Due to a configuration issue, our app was passing a port of `null` to pg-pool. This is handled "correctly" (defaulting to 5432) by pg.

Valid traces

### What did you see instead?
`net.peer.port` is set to `NaN`, which is invalid, breaking the entire trace.

pichlermarc:

- when using `null` as a parameter for `port`, the `port` can end up being `NaN`
  - see https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg/lib/connection-parameters.js#L66

Fixes #1980

- when there's a non-number port is in the parameters object, we fall back to `undefined`


carlpett (approved):
:+1:

plugins/node/opentelemetry-instrumentation-pg/test/utils.test.ts
```diff
@@ -193,4 +194,48 @@ describe('utils.ts', () => {
...
+      );
+      assert.strictEqual(
+        utils.getSemanticAttributesFromConnection({
+          port: 1.234,
+        })[SemanticAttributes.NET_PEER_PORT],
+        undefined
+      );
+    });
+
+    it('should set port attribute to undefined when port is an integer', () => {
```
pichlermarc, line 226:
```suggestion
    it('should set port attribute when port is an integer', () => {
```

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -94,14 +94,25 @@ export function getConnectionString(params: PgParsedConnectionParams) {
   return `postgresql://${host}:${port}/${database}`;
 }
 
+function getPort(port: number | undefined): number | undefined {
+  // Port may be NaN as parseInt() is used on the value, passing null will result in NaN being parsed.
+  // https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg/lib/connection-parameters.js#L66
+  if (port != null && Number.isInteger(port)) {
```
trentm, line 100: nit: I suppose the `port != null &&` could be dropped because `Number.isInteger(null) === false`.

trentm (approved):

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
pichlermarc, line 100:
```suggestion
  if (Number.isInteger(port)) {
```

pichlermarc: ah, yes good point :slightly_smiling_face:

pichlermarc: applied in [da7b54d](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1982/commits/da7b54d0e3bce706eb396997f52cb199c46207b2)

PR #1989, opened 2024-03-04
fix(instrumentaiton-runtime-node)!: rename exported instrumentation to match package name

pichlermarc:
See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1970#issuecomment-1974965699


pichlermarc:
> LGTM. Should we consider this a breaking change?

Yep, it's technically a breaking change, I changed the title. Though that package is not released yet so whatever we do with it is fine :slightly_smiling_face:

trentm (approved):

trentm:
Thanks!

JamieDanielson (approved):

PR #1994, opened 2024-03-05
chore(instr-runtime-node): update some deps to same versions as other contrib packages

trentm:
- reduce 'mocha' dep to v7 used by all other packages
- bump /api peer dep to 1.3.0 (the version when api-metrics was merged into api)

* * *

This replaces https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1967. Because this package has moved to a new name (used to be instrumentation-perf-tools), it was easier to start with a fresh PR. This also takes into consideration Marc's feedback at https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1967#issuecomment-1968578311

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -126,6 +126,9 @@
       },
       "engines": {
         "node": ">=14"
+      },
+      "peerDependencies": {
+        "@opentelemetry/api": "^1.0.0"
       }
     },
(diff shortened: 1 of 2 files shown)


pichlermarc (approved):
Thanks :+1:

PR #1995, opened 2024-03-05
fix(instr-express): normalize paths with double slashes

### LINKED ISSUE #1547: Express instrumentation does not normalize paths (removing double slashes)

JamieDanielson:

- closes #1547 

- Add replace to remove duplicate slashes that may exist in defined routes
- Add unit test confirming route does not contain extra slash
- Add route to util `serverWithMiddleware` that contains double slashes, for use in test

Test results before the change:

```
+ expected - actual
-/double-slashes//:id
+/double-slashes/:id
```


plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -455,6 +455,43 @@ describe('ExpressInstrumentation', () => {
...
+      const rootSpan = tracer.startSpan('rootSpan');
+      let finishListenerCount: number | undefined;
+      let rpcMetadata: RPCMetadata | undefined;
+      const httpServer = await serverWithMiddleware(tracer, rootSpan, app => {
+        app.use(express.json());
+        app.use((req, res, next) => {
+          rpcMetadata = getRPCMetadata(context.active());
+          res.on('finish', () => {
+            finishListenerCount = res.listenerCount('finish');
+          });
```
pichlermarc, line 468: Hmm, is there a specific goal with this and the associated assertion below? :thinking: 
To my untrained eye it seems somewhat unrelated to what's being tested here.

pichlermarc (commented):
Looks good already, thanks for working on this :slightly_smiling_face: 
Just one question about the added test

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
JamieDanielson: Oh good point, looks like extra copy-pasta leftover from another test. Removed

pichlermarc (approved):
Looks good, thanks for working on this :slightly_smiling_face:

PR #2000, opened 2024-03-06
feat(instrumentation-aws-lambda): support ESM handlers and other less common handler patterns

raphael-theriault-swi:

The Lambda instrumentation doesn't currently support instrumenting ESM handlers, which are now the default, and other less common handler patterns supported by Lambda such as extensionless handler modules and nested handler functions.

The previous logic to determine the handler to patch is replaced with an implementation based on [the one present in the Lambda runtime](https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/v3.1.0/src/UserFunction.js). I tried to keep the function names and structure as close to possible to the original to make it easier to cross-reference.

The tests also now use code copied from the Lambda runtime to load handlers, as opposed to just calling `require` which doesn't properly reflect a real-world scenario and doesn't work at all for ESM.

I also updated the notes about absolute path support in the patching infrastructure. The instrumentation keeps using the same hack but it would probably be a good idea to fix the issue upstream, which is that while [absolute paths are taken into account while registering hooks](https://github.com/open-telemetry/opentelemetry-js/blob/experimental/v0.52.1/experimental/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L304-L309), they are not when matching against the module name within the hook function, [which ignores the `baseDir`](https://github.com/open-telemetry/opentelemetry-js/blob/experimental/v0.52.1/experimental/packages/opentelemetry-instrumentation/src/platform/node/instrumentation.ts#L209).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,8 +31,8 @@
         "lerna-changelog": "2.2.0",
         "markdownlint-cli2": "0.13.0",
         "minimatch": "^9.0.3",
-        "nx": "15.9.7",
         "mocha": "^10.7.3",
+        "nx": "15.9.7",
         "prettier": "2.8.8",
         "process": "0.11.10",
         "semver": "^7.6.0",
@@ -95,6 +95,8 @@
       "devDependencies": {
         "@opentelemetry/api": "^1.0.0",
         "@opentelemetry/contrib-test-utils": "^0.42.0",
+        "@opentelemetry/instrumentation-fs": "^0.16.0",
+        "@opentelemetry/instrumentation-http": "^0.54.0",
         "@types/mocha": "8.2.3",
         "@types/node": "18.18.14",
         "@types/sinon": "10.0.20",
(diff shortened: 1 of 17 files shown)


raphael-theriault-swi:
Missing a test at the moment, still need to figure out what the best way to go about that would be since the other Lambda tests do not use the fixture utility and all share the same structure.

raphael-theriault-swi:
@JamieDanielson I think you've been pushing a lot for ESM support as a whole, if I rebased this do you think you'd be able to take a look ?

raphael-theriault-swi:
@jj22ee a review would be much appreciated if when have some time

jj22ee:
Hi, I see that there have been merged changes recently to support ESM handlers while not changing the implementation in `instrumentation-aws-lambda`. There are fixes related to the issue of `baseDir`.
- https://github.com/open-telemetry/opentelemetry-lambda/issues/1557#issuecomment-2443807330

It looks like that ESM handlers will be supported once all these changes are released in the latest OTel JS version.
Would your changes still be needed after this?

pragmaticivan:
cc @tylerbenson @serkan-ozal @mchiaramonte

Issue #1983, opened 2024-03-01
[instrumentation-mongodb] test failing for `mongodb@6.4.0` 


pichlermarc:
**Description:**
`mongodb@6.4.0` released with a refactor that changed how the internal `Connection.command()` works, so no spans are generated and test fail.

In the instrumentation, we heavily rely on `command`, see https://github.com/open-telemetry/opentelemetry-js-contrib/blob/2e9fce4d15bc15fd3cb0e98ad36af6f7d4cc3c38/plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts#L495-L506

Since the signature changed we return `original.call()` instead of instrumenting. `resultHandler` is `undefined` and therefore not a function. also the other args have changed, see below. Right now I think this is not causing problems in end-user apps but if it is, we should temporarily disable `mongodb@6.4.0` instrumentation until a fix is ready.

**Resources:**
 - [`connection.ts` at `mongodb@6.3.0`](https://github.com/mongodb/node-mongodb-native/blob/v6.3.0/src/cmap/connection.ts#L493-L497) - works
 - [`connection.ts` at `mongodb@6.4.0`](https://github.com/mongodb/node-mongodb-native/blob/9ac2e38f59b673d831a2ac56e223ca8b1a49c713/src/cmap/connection.ts#L569-L579) - fails

**See failing test on `main`**
- https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8093744773/job/22167022286
 - I restarted the test for `18.18.2` on `main` after seeing failures in #1978

trentm:
Yah, I was poking David as well ;)
I took a look, and in v6.4.0 the signature of `ConnectionPool.checkOut` changed:

```diff
-  checkOut(callback: Callback<Connection>): void {
+  async checkOut(): Promise<Connection> {
```

This breaks this patch in instr-mongodb/src/instrumentation.ts

```ts
  // This patch will become unnecessary once
  // https://jira.mongodb.org/browse/NODE-5639 is done.
  private _getV4ConnectionPoolCheckOut() {
    return (original: V4ConnectionPool['checkOut']) => {
      return function patchedCheckout(this: unknown, callback: any) {
        const patchedCallback = context.bind(context.active(), callback);
        return original.call(this, patchedCallback);
      };
    };
  }
```

david-luna:
Hi @trentm @pichlermarc 

[...]

PR #2001, opened 2024-03-07
fix(mongodb): `mongodb` v6.4.0 & 6.5.0 support

david-luna:

the new version of mongodb `v6.4.0` comes with some internal changes that break the instrumentation. Drtails of the changes are described in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1983#issuecomment-1979740812

Closes: #1983 

patches of the changed methods have been refactored to accept both implementations
- handling callbacks for `<6.4.0`
- handling returned promises for `>=6.4.0`


plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -353,7 +353,12 @@ export class MongoDBInstrumentation extends InstrumentationBase {
           instrumentation.setPoolName(options);
           callback(err, conn);
         };
-        return original.call(this, options, patchedCallback);
+
+        const result = original.call(this, options, patchedCallback);
```
trentm, line 357: It feels slightly messy to be passing in this `patchedCallback` that isn't going to be used.
Thoughts on a separate patch function for mongodb >=6.4.0 -- this one would just do the promise thing.

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/mongodb": "3.6.20",
     "@types/node": "18.6.5",
     "mocha": "7.2.0",
-    "mongodb": "3.6.11",
+    "mongodb": "^6.4.0",
```
trentm, line 61: If this is done, then the default `"test"` script will need to change above, right?

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -527,15 +532,23 @@ export class MongoDBInstrumentation extends InstrumentationBase {
...
             span,
             resultHandler,
             this.id,
             commandType
           );
-
-          return original.call(this, ns, cmd, options, patchedCallback);
         }
+
+        const result = original.call(this, ns, cmd, options, patchedCallback);
```
trentm, line 543: Similar Q to above: would separate patch functions (one that uses the callback, one that uses the promise) be clearer? or would they just be a pain?

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
trentm, line 1: This will need to sync the #1984 change and then revert those changes *except* for the limit on the patching of 'mongodb/lib/cmap/connection_pool.js' right?   My understanding was that the ConnectionPool.checkOut patch is no longer necessary for async-context tracking.  We'd need to verify with the tests here though.

david-luna: indeed we need to sync with #1984. Running the tests I found the patch of `COnnectionPool.checkOut` to be innocuous for the newer version but I agree that if not needed it's better to avoid it.

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
david-luna: my intention was to not split the patch logic of this function to 2 separate patches. I do agree is not optimal to crate a function that may not be used. I'll try to come with a different approach

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
david-luna: here the `patchedCallback` can be reused because basically the API change has been to "promisify" the function. A separate patch is not complicated, it's almost an identical copy.

david-luna:
@trentm I did the split you suggested but partially. I think it was overkill to split the logic of the `connect` patch.

Now I'm hitting a problem in `test-all-versions` script. It is failing for `mongodb@4.17.1` in the beforeEach method of the tests which is quite odd.

if I log the version under test...
```diff
--- a/plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v4.test.ts
@@ -56,6 +56,9 @@ describe('MongoDBInstrumentation-Tracing-v4', () => {
   const COLLECTION_NAME = 'test-traces';
   const URL = `mongodb://${HOST}:${PORT}/${DB_NAME}`;

+  const version = require('mongodb/package.json').version;
+  console.log(`testing mongodb@${version}`);
+
   let client: mongodb.MongoClient;
   let collection: mongodb.Collection;
```

it turns out we get a different version for testing
```bash
> @opentelemetry/instrumentation-mongodb@0.41.0 test-v4
> nyc ts-mocha -p tsconfig.json --require '@opentelemetry/contrib-test-utils' 'test/mongodb-v4-v5-v6.metrics.test.ts' 'test/**/mongodb-v4.test.ts'

testing mongodb@6.4.0

  MongoDBInstrumentation-Metrics
    ✓ Should add connection usage metrics (101ms)
    ✓ Should add disconnection usage metrics (213ms)

  MongoDBInstrumentation-Tracing-v4
      1) "before each" hook: mongoBeforeEach for "should create a child span for insert"
      2) "after each" hook for "should create a child span for insert"


  2 passing (4s)
  2 failing

....
```

@trentm I think you got into a similar situation but I can't find the issue or PR. Do you recall having a similar issue? 🤔

trentm:
> @trentm I think you got into a similar situation but I can't find the issue or PR. Do you recall having a similar issue? 🤔

Yup, it was described here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1938#issuecomment-1960183358
At the time the conflict (also for version mongodb@4.17.1) was -- I think -- because of a conflict with the "plugins/node/opentelemetry-instrumentation/mongodb/examples" package, which we've since removed. However, *now* we get the same because of a pinned "mongodb" transitive dep from instrumentation-mongoose:

```
[~/tm/opentelemetry-js-contrib2/plugins/node/opentelemetry-instrumentation-mongodb (git:1983-mongodb-6-4-support)]
% npm ls mongodb
opentelemetry-contrib@0.1.0 /Users/trentm/tm/opentelemetry-js-contrib2
└─┬ @opentelemetry/instrumentation-mongodb@0.41.0 -> ./plugins/node/opentelemetry-instrumentation-mongodb
  └── mongodb@6.4.0

[~/tm/opentelemetry-js-contrib2/plugins/node/opentelemetry-instrumentation-mongodb (git:1983-mongodb-6-4-support)]
% npm install --no-save mongodb@4.17.1
...

[~/tm/opentelemetry-js-contrib2/plugins/node/opentelemetry-instrumentation-mongodb (git:1983-mongodb-6-4-support)]
% npm ls mongodb
opentelemetry-contrib@0.1.0 /Users/trentm/tm/opentelemetry-js-contrib2
└─┬ @opentelemetry/instrumentation-mongodb@0.41.0 -> ./plugins/node/opentelemetry-instrumentation-mongodb
  └── mongodb@6.4.0
```

The looking for any npm install felt it didn't need to install v4.17.1 locally:

```
[~/tm/opentelemetry-js-contrib2 (git:1983-mongodb-6-4-support)]
% npm ls mongodb
opentelemetry-contrib@0.1.0 /Users/trentm/tm/opentelemetry-js-contrib2
├─┬ @opentelemetry/instrumentation-mongodb@0.41.0 -> ./plugins/node/opentelemetry-instrumentation-mongodb
│ └── mongodb@6.4.0
└─┬ @opentelemetry/instrumentation-mongoose@0.36.0 -> ./plugins/node/instrumentation-mongoose
  └─┬ mongoose@6.12.3
    └── mongodb@4.17.1
```

Ew. This has to be an npm bug. I'm not sure how to cope with this in general.
We discussed (in DMs) explicitly skipping v4.17.1 in the .tav.yml file. That's a hack, but I can't think of anything better.

.github/workflows/unit-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       matrix:
         node: ["14", "16", "18.18.2"]
         include:
-            - node: 14
+            - node: 16
```
trentm, line 14: Note to reviewers: This was discussion in the OTel JS SIG this week as a fix for the code-coverage check failures. Testing coverage only with node v14 artificially results in coverage looking poor for a unit-test run that skips all or most tests for the (now old) node v14.

plugins/node/opentelemetry-instrumentation-mongodb/.tav.yml
```diff
@@ -1,11 +1,17 @@
...
+      # This version has been excluded because `mongoose` instrumentation has it
+      # pined as a transitive dependency.
+      # When `test-all-versions` tries to install it npm actually does nothing and
+      # we end up running tests for the wrong version (v6.5.0)
+      # Ref: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2001#issuecomment-1992202942
+      exclude: "4.17.1"
+    commands: npm run test-v4
+  - versions: ">=5 <6"
+    commands: npm run test-v5-v6
+  - versions: ">=6 <6.4"
```
trentm, line 15:
```suggestion
  - versions: ">=6 <7"
```

trentm (approved):
Thanks for humouring me on the splitting the callback vs. promise patching. That looks like it a was a fair amount of extra work.

LGTM, though I think the .tav.yml update should be added before merge.

plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/mongodb": "3.6.20",
     "@types/node": "18.6.5",
     "mocha": "7.2.0",
-    "mongodb": "3.6.11",
+    "mongodb": "^6.5.0",
```
pichlermarc, line 61: We usually pin devDependencies. :slightly_smiling_face: 
```suggestion
    "mongodb": "6.5.0",
```

pichlermarc (approved):
Thanks for taking care of this :slightly_smiling_face:

david-luna:
So the split decreased the coverage enough to block the PR. Will add more tests to increase it back shortly

david-luna:
By looking at [the docs](https://docs.codecov.com/docs/quick-start) there is no way to tell `codecov` that a specific code path is expected to not to be covered. I guess the solution would be to make use of [merge reports](https://docs.codecov.com/docs/merging-reports) but IMO this is out of scope of this PR

@pichlermarc do you think is okay to move forward and have a follow-up PR about merging coverage reports from unit testing?

trentm:
I agree dealing with merging of coverage reports is out of scope for this PR.

PR #2008, opened 2024-03-13
feat(instrumentation-express): Support non-string routes

onurtemizkan:

This came up while migrating `@sentry/node` to OpenTelemetry. (Related issue: https://github.com/getsentry/sentry-javascript/issues/10168).

This PR adds support for:
 - Routes that are defined as regular expressions. 
 - Multiple routes provided as an array.

The current behaviour falls back to `/` as the `layerPath` when the Express route argument is not a string, which does not cover all uses ([Express docs](https://expressjs.com/en/4x/api.html#app.use)).

`@sentry/node` already has a test suite to cover these uses, so this PR migrates them to OpenTelemetry for validation.

- Adds `getLayerPath` utility to extract multiple route arguments, including regular expressions.
- Adds new ESM tests to validate this use. ([Migrated from `sentry-javascript`](https://github.com/getsentry/sentry-javascript/blob/d26b25d18c201572441c4f5885623bc509eaa168/dev-packages/node-integration-tests/suites/express/tracing/test.ts))


plugins/node/opentelemetry-instrumentation-express/test/fixtures/use-express-regex.mjs
```diff
@@ -0,0 +1,67 @@
...
+app.get(['/test/arr/:id', /\/test\/arr[0-9]*\/required(path)?(\/optionalPath)?\/(lastParam)?/], (_req, res) => {
+  res.send({ response: 'response' });
+});
+
+const server = http.createServer(app);
+await new Promise(resolve => server.listen(0, resolve));
+const port = server.address().port;
+
+await new Promise(resolve => {
+  http.get(`http://localhost:${port}/test/arr/requiredPath/optionalPath/lastParam`, (res) => {
```
AbhiPrasad, line 58: Can we table test more regex scenarios to see if it works? good to help against regression

https://github.com/getsentry/sentry-javascript/blob/78797919819f6427d667e3955c439ff8e99eeab2/dev-packages/node-integration-tests/suites/express/tracing/test.ts#L94-L101

plugins/node/opentelemetry-instrumentation-express/test/fixtures/use-express-regex.mjs
```diff
@@ -0,0 +1,67 @@
...
+app.use(async function simpleMiddleware(req, res, next) {
+  // Wait a short delay to ensure this "middleware - ..." span clearly starts
+  // before the "router - ..." span. The test rely on a start-time-based sort
+  // of the produced spans. If they start in the same millisecond, then tests
+  // can be flaky.
+  await promisify(setTimeout)(10);
+  next();
+});
+
+app.get(['/test/arr/:id', /\/test\/arr[0-9]*\/required(path)?(\/optionalPath)?\/(lastParam)?/], (_req, res) => {
```
AbhiPrasad, line 49: can we have a test for some simpler regex like we do here? https://github.com/getsentry/sentry-javascript/blob/78797919819f6427d667e3955c439ff8e99eeab2/dev-packages/node-integration-tests/suites/express/tracing-experimental/server.js#L25-L31

trentm:
> I am trying to confirm if there is specific guidance on when to approve the workflows.

FWIW, I've tended to "Approve and run" after a sanity check that the PR doesn't add a bitcoin miner, credentials exfil or somethign like that. :)

onurtemizkan:
@JamieDanielson, thanks for reviewing. I updated the PR.

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -145,3 +145,27 @@ export const asErrorAndMessage = (
...
+  }
+
+  if (Array.isArray(args[0])) {
+    return args[0]
+      .map(arg => (typeof arg === 'string' || arg instanceof RegExp ? arg : ''))
+      .join(',');
+  }
+
+  return;
+};
```
dyladan, line 171: Why are we using `args: unknown[]` when we only ever use the first argument `args[0]`?

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -145,3 +145,27 @@ export const asErrorAndMessage = (
...
+    return args[0].toString();
+  }
+
+  if (typeof args[0] === 'string') {
+    return args[0];
+  }
+
+  if (Array.isArray(args[0])) {
+    return args[0]
+      .map(arg => (typeof arg === 'string' || arg instanceof RegExp ? arg : ''))
```
dyladan, line 166: Is there any situation where an array entry might be a type other than string or array? If so, it looks like this will result in an empty entry: `getLayerPath(['/test', 6, /test/]) === '/test,,/test/`

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -145,3 +145,27 @@ export const asErrorAndMessage = (
...
     ? [error, error.message]
     : [String(error), String(error)];
+
+/**
+ * Extracts the layer path from the route arguments
+ *
+ * @param args - Arguments of the route
+ * @returns The layer path
+ */
+export const getLayerPath = (args: unknown[]): string | undefined => {
```
dyladan, line 155: This seems a bit fragile to me in the sense that it would be pretty easy to get the same output from a regexp that it would be to get from a string path with a trailing slash. Is there any guardrail against this situation, is this an acceptable risk, or is this not actually an issue if it happens? I'm not an expert in this instrumentation and I'm not 100% sure what the layer path is used for.

dyladan (commented):
Overall looks ok to me but I have a little bit of concern that getLayerPath is not resilient to strings and regexes that have the same stringified represenation.

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
onurtemizkan: Yes, looks like it's possible, I've added support for numbers.

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
onurtemizkan: Updated 👍

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
onurtemizkan: IMHO it would be an acceptable risk. I'm not sure how to work around this without breaking the path pattern.

AbhiPrasad:
@dyladan @JamieDanielson could you take another look at this when you get the time? Thanks!

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
JamieDanielson: Is this still an open question? I'm not sure I'm following the specific concern here, can you provide an example of what we are considering guarding against?

onurtemizkan: My understanding is, that as this PR converts regexp path definitions to a string representation, it may collide with an exact string path which can make both end up under the same transaction name.

Something like:
- `/\/test\/regex/` and
- `/test/regex/` can end up as
- `/test/regex/`

This seems possible, but l don't think it is a common pattern when people define routes.

lforst: Just a thought since I am also following this PR: There is always the possibility for collisions when coercing anything into a string. In my opinion this is worth it. If we want to reduce the likelihood of collisions, we could transform regular expressions like `const re = /\/test\/regex\/` with ``const coerced = `RegEx(${re.toString()})`;``. 🤔

JamieDanielson: Hmm... I think I'm in agreement here that this is an acceptable risk as-is, as we are otherwise guarding against edge cases that are unlikely to occur, and if it does occur it is a fairly minimal impact. Especially compared to the current state, this gets us to a much better place. If we find ourselves getting issues related to this we can reconsider, but it might even just end up being a documentation thing at that point.

PR #2009, opened 2024-03-14
fix(instrumentation-mongoose): Fix instrumentation for Mongoose Model methods, save() and remove(), by passing options argument to original method calls

### LINKED ISSUE #1678: Transaction is not used in Mongoose create operation after OpenTelemetry auto-instrumentation is added
    "@opentelemetry/api": "^1.5.0",
    "@opentelemetry/auto-instrumentations-node": "^0.39.2",

v16.15.1

### What did you do?
When running some database operation inside the transaction
```
const User = db.model('User', new Schema({ name: String }));
const session = await User.startSession();
await session.startTransaction();
const [newStaff] = await User.create([ {name: 'staffA'} ], {session});
await session.abortTransaction();
await session.endSession();

```

It is expected the transaction will be aborted, and user with name "staffA" will not be created in the database.

### What did you see instead?
The transaction was not rolled back, and the record was created in the database.

### Additional Info
I am using "mongoose": "^6.2.1"
The transaction works with `findOneAndUpdate`, for example the user name is not changed to `staffB` in the following case
```
const User = db.model('User', new Schema({ name: String }));
const [newStaff] = await User.create([ {name: 'staffA'} ]);
const session = await User.startSession();
await session.startTransaction();
await User.findOneAndUpdate({name: 'staffA'}, {name: 'staffB'}).session(session);
await session.abortTransaction();
await session.endSession();
```
However, when session is used in the `create` operation, the `lastUse` in `serverSession` is not updated across the whole function, which means the session is not used at all.

ferrelucas:
Fixes #1678 

This PR fixes the issue where transactions cannot be used for the Model methods `save()` and `remove()` (and their aliases eg.  `create()`) because after instrumentation is applied to these methods, specifically when both `options` and `callback` are defined, the final method call loses the first argument `option` making the final call not take the `option.session` into consideration when making database operations.
This could lead to save updates in transactions not being properly rolled back as well as multiple operations that are supposed to share the same session not being able to do so because their calls would eventually be made without the `session` reference.

- Pass the original `args` into `handleCallbackResponse()`
- Updates the implementation of `handleCallbackResponse()` so that we create `newArgs` including the original `options` and the instrumented`callback` function.
- Added tests to cover the base case of empty-object but defined `options` argument with `callback`
- Added tests cases to cover committed and aborted transactions


plugins/node/instrumentation-mongoose/test/mongoose.test.ts
```diff
@@ -63,8 +63,11 @@ describe('mongoose instrumentation', () => {
   beforeEach(async () => {
     instrumentation.disable();
     instrumentation.setConfig({
-      dbStatementSerializer: (_operation: string, payload) =>
-        JSON.stringify(payload),
+      dbStatementSerializer: (_operation: string, payload) => {
+        return JSON.stringify(payload, (key, value) => {
+          return key === 'session' ? '[Session]' : value;
```
ferrelucas, line 68: ℹ️ Note this replacer function was only introduced here to avoid an error due to the Session object having cyclical reference

ferrelucas:
Hi @blumamir, wondering if you could have a look at this PR when you have some time?

ferrelucas:
@open-telemetry/javascript-approvers Hi there, wondering if somebody could have a look at this PR?

plugins/node/instrumentation-mongoose/src/utils.ts
```diff
@@ -99,15 +99,27 @@ export function handleCallbackResponse(
...
-    (err: Error, response: any) => {
-      err
-        ? setErrorStatus(span, err)
-        : applyResponseHook(span, response, responseHook, moduleVersion);
-      span.end();
-      return callback!(err, response);
-    },
-  ]);
+  const newArgs = [];
+  for (const [, argValue] of Object.entries(args[0])) {
```
blumamir, line 106: The argument is an array but we only consume the first value which we expect to be an object. Can you explain the motivation for this?

Is `args[0]` the options object? in case why do we need to desturct it's values into an array? and what if it is being used with both options and a callback?

plugins/node/instrumentation-mongoose/src/utils.ts
```diff
@@ -99,15 +99,27 @@ export function handleCallbackResponse(
...
+    newArgs.push(argValue);
+  }
+
+  let callbackArgumentIndex = 0;
+  if (newArgs.length === 2) {
+    callbackArgumentIndex = 1;
+  }
+
+  newArgs[callbackArgumentIndex] = (err: Error, response: any): any => {
```
blumamir, line 115: are we overwriting a user callback which is now not invoked? should we store it's value before overwriting and invoke it if present from our response callback?

blumamir (commented):
Sorry for the time it took to review and thank you for creating a fix for this issue 🙏

Not sure I fully understood the logic in this PR regarding adding new callback to the existing array.

plugins/node/instrumentation-mongoose/src/utils.ts
ferrelucas: Hi @blumamir , thanks for taking the time to review!

Sorry, this was my mistake. I, for some reason, defined `args` as `...IArguments[]` and assumed it was necessary.
So, just for explaining the code:
Logging the the value of `args` when the method is called with both the `options` object and the `callback` function, we would get the following:
```
args: [ [Arguments] { '0': [Object], '1': [AsyncFunction (anonymous)] } ],
```
In this case, `args` would be an array that has only 1 element - the array of the method arguments: `options` and `callback`. Here is what loging the value of `args[0]` would result:
```
args0: [Arguments] {
    '0': { session: [ClientSession] },
    '1': [AsyncFunction (anonymous)]
  }
```
-----
But, yes, this is not necessary. 
I am pushing an update where we can make the update `args` is defined as `IArguments` and we can directly update `args`.

ferrelucas:
@blumamir , I've update the PR to simplify the usage of `args`

Please let me know what you think
Thanks!

plugins/node/instrumentation-mongoose/src/utils.ts
blumamir: Never mind, I was thinking that the `callback` is already part of the args and we can potentially extract it from it instead of sending it with the function. But now that I think about it again, I think it is more readable the way it is now 👍

blumamir (approved):
LGTM

Thanks again for the fix and sorry it took so long to review 🙏

ferrelucas:
No worries!
Thanks for the review! 🙌

blumamir:
@ferrelucas let me know if you need any help fixing the CI issues. once it is green we can merge and release this fix

ferrelucas:
Hi @blumamir I've taken one of the CI unit tests as example `Run test-all-versions (20)`.
I've seen that the MongoDB setup is in the step called `Initialize containers`, under `Starting mongo service container`.

It is running:
```
/usr/bin/docker create --name fa2fac6564d8478ebd791968ddae63b9_mongo_11c61e --label 62ce9a --network github_network_60aacba5f9dc4f70be079739b8ce42a8 --network-alias mongo -p 27017:27017  -e GITHUB_ACTIONS=true -e CI=true mongo
```

I think we could add the `--replSet` parameter to the command to start the MongoDB instance, [then initialize the replica set](https://www.mongodb.com/resources/products/compatibilities/deploying-a-mongodb-cluster-with-docker).

I believe I don't have permissions to make these changes in the Github Actions
Would it be feasible to do it on your side?
Thanks!

blumamir:
> I believe I don't have permissions to make these changes in the Github Actions
> Would it be feasible to do it on your side?

If I am not mistake, the mongo service is started [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/0cef9fa95d065489b7349bfa0150ff858ed59157/.github/workflows/unit-test.yml#L23) which you can edit, but I really think it would be easier to just use some other option for the test instead of `session`, if one of them is simpler.

ferrelucas:
@blumamir thanks! I've updated the PR unit tests to use `wtimeout` instead of `session` in the options object

blumamir (approved):

ferrelucas:
Hey, @blumamir thanks again for the review! 🙌

PR #2016, opened 2024-03-15
feat(instrumentation-winston): Allow log level to be configured for log sending

hectorhdzg:

- Allow users to configure severity level to be used by OpenTelemetry Winston Transport used in this instrumentation 

Related to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1919


plugins/node/opentelemetry-instrumentation-winston/src/instrumentation.ts
```diff
@@ -206,7 +207,16 @@ export class WinstonInstrumentation extends InstrumentationBase {
               let newTransports = Array.isArray(originalTransports)
                 ? originalTransports
                 : [];
-              const openTelemetryTransport = new OpenTelemetryTransportV3();
+              let transportOptions = {};
+              if (config.logSeverity) {
+                const winstonLevel = winstonLevelFromSeverity(
```
trentm, line 212: (This could be a separate issue/PR, or here.) I gather that the result will confusing if the user configured their Winston logger with levels other than the npm defaults.  The level names returned by `winstonLevelFromSeverity` would result in `undefined` values in the comparison at https://github.com/winstonjs/winston-transport/blob/f4be5c72b5c8d90b0aae1690e077089116d4bb64/modern.js#L80

Perhaps, either add a comment in the README for this option that `logSeverity` only works with the default Winston levels (https://github.com/winstonjs/winston#logging-levels), or update this code to look at `args[0]?.levels` to try to cope. There is a limit to how well this could cope with weird levels. The built in npm, syslog, and cli levels could probably be handled.

I suppose the code could emit a diag.warn if `logSeverity` is given but `args[0].levels` is provided with unexpected level names.

Thoughts?

hectorhdzg: @trentm  Yeah sound good, let me see how we can improve this to support other levels

hectorhdzg:
@trentm this one should be ready for another round, let me know if you have any feedback or concern

trentm (approved):

trentm:
LGTM, thanks!

PR #2019, opened 2024-03-18
perf(instrumentation-pg): reduce temp objects allocations

Samuron:

The amount of allocations that opentelemetry produces is too high

Replaced attributes passed on span creation with span.setAttirbutes to avoid multiple remapping inside startSpan function, reused kind by providing constant and replaced string interpolation with constants

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/src/index.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/index.ts
@@ -17,4 +17,3 @@
 export * from './instrumentation';
 export * from './types';
 export * from './enums/AttributeNames';
-export * from './enums/SpanNames';
--- a/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
@@ -39,19 +39,11 @@ import {
 import { PgInstrumentationConfig } from './types';
 import * as utils from './utils';
 import { AttributeNames } from './enums/AttributeNames';
-import {
-  SemanticAttributes,
-  DbSystemValues,
-} from '@opentelemetry/semantic-conventions';
 import { addSqlCommenterComment } from '@opentelemetry/sql-common';
 import { VERSION } from './version';
 
-const PG_POOL_COMPONENT = 'pg-pool';
-
 export class PgInstrumentation extends InstrumentationBase {
-  static readonly COMPONENT = 'pg';
-
-  static readonly BASE_SPAN_NAME = PgInstrumentation.COMPONENT + '.query';
+  static readonly BASE_SPAN_NAME = 'pg.query';
 
   constructor(config: PgInstrumentationConfig = {}) {
     super(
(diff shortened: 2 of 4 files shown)


plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -141,12 +142,10 @@ export function handleConfigQuery(
   const spanName = getQuerySpanName(dbName, queryConfig);
   const span = tracer.startSpan(spanName, {
     kind: SpanKind.CLIENT,
-    attributes: {
-      [SemanticAttributes.DB_SYSTEM]: DbSystemValues.POSTGRESQL, // required
-      ...getSemanticAttributesFromConnection(connectionParameters),
-    },
   });
 
+  span.setAttributes(getSemanticAttributesFromConnection(connectionParameters));
```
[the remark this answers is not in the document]
Samuron: @Flarna oh, makes sense. Guess all is left is to reduce one temp allocation with spreading then, I've updated the MR

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -367,17 +351,17 @@ export class PgInstrumentation extends InstrumentationBase {
...
-            [AttributeNames.MAX_CLIENT]: this.options.maxClient,
-          },
+          attributes: utils.getSemanticAttributesFromConnection(this.options),
         });
 
+        span.setAttribute(
+          AttributeNames.IDLE_TIMEOUT_MILLIS,
+          this.options.idleTimeoutMillis
+        );
+        span.setAttribute(AttributeNames.MAX_CLIENT, this.options.maxClient);
```
blumamir, line 363: It is generally preferred to have all the attributes that are available set on `startSpan` so they can be used in samplers. I would reconsider setting them after span creation.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -39,19 +39,11 @@ import {
...
 import { addSqlCommenterComment } from '@opentelemetry/sql-common';
 import { VERSION } from './version';
 
-const PG_POOL_COMPONENT = 'pg-pool';
-
 export class PgInstrumentation extends InstrumentationBase {
-  static readonly COMPONENT = 'pg';
-
-  static readonly BASE_SPAN_NAME = PgInstrumentation.COMPONENT + '.query';
+  static readonly BASE_SPAN_NAME = 'pg.query';
```
blumamir, line 46: for `pg.query` there is this `static readonly` on the class, but for other similar span names like `'pg.connect'` we have the value hardcoded in the `startSpan' call.
I would consider aligning the behavior to have consistency.

also, referencing this property from `utils.ts` creates a circular import (utils.ts depends on. iinstrumentation.ts and vice versa). it was already there previously, but since you touch it, might be a good chance to move somewhere else, like `consts.ts`.

blumamir (commented):
Thank you for the contribution and for help improving the code.
Added 2 nit comments

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
Samuron: Done

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
Samuron: Done, I've tried to match the style with enums that already was in place

plugins/node/opentelemetry-instrumentation-pg/src/index.ts
```diff
@@ -17,3 +17,4 @@
 export * from './instrumentation';
 export * from './types';
 export * from './enums/AttributeNames';
+export * from './enums/SpanNames';
```
blumamir, line 20: Actually, I think we should not export it from index, as this would add it to the instrumentation public API which would make it a possible breaking change if we make changes in the future.

blumamir (commented):
Thanks for addressing the previous comments so quickly 🥇 
added one last nit comment about the index.ts

plugins/node/opentelemetry-instrumentation-pg/src/index.ts
Samuron: Done

blumamir (approved):

PR #2028, opened 2024-03-20
feat(koa): Adds support to ignore a span by its layer name

MrFabio:

* Adds `ignoreLayersName` option to instrumentation constructor config to allow layers to be ignored by its name.
* Currently we have the `ignoreLayersType` config, this new one will run after this check.
* It is useful if we want to ignore some layers like logger, allowedMethods, and some ones with no name. It also helps with reducing the number of spans.

* Defines `KoaInstrumentationConfig` with `ignoreLayersName` which is an array of `string` (layer names):
* Utils module with `isLayerNameIgnored` to check if a layer should be ignored by the name
* Modifies `_patchLayer` to check if the layer is ignored by the config


MrFabio:

@blumamir can you help?

plugins/node/opentelemetry-instrumentation-koa/README.md
```diff
@@ -54,6 +54,7 @@ Note that generator-based middleware are deprecated and won't be instrumented.
 | Options            | Type                                | Example              | Description                                                                                              |
 | ------------------ | ----------------------------------- | -------------------- | -------------------------------------------------------------------------------------------------------- |
 | `ignoreLayersType` | `KoaLayerType[]`                    | `['middleware']`     | Ignore layers of specified type.                                                                         |
+| `ignoreLayersName` | `string[]`                          | `['logger']`         | Ignore layers with specified names.                                                                      |
```
trentm, line 57: @MrFabio What would you think about changing this from `ignoreLayersName` to `ignoreLayers` and copying the behaviour of the `ignoreLayers` config from instrumentation-express, given that instrumentation-express already has a very similar config option with (mostly) the same behaviour?

plugins/node/opentelemetry-instrumentation-koa/src/utils.ts
```diff
@@ -36,6 +37,7 @@ export const getMiddlewareMetadata = (
         [SEMATTRS_HTTP_ROUTE]: layerPath?.toString(),
       },
       name: context._matchedRouteName || `router - ${layerPath}`,
+      layerName: context._matchedRouteName || layerPath?.toString() || '',
```
trentm, line 40: Is there a strong need to have a separate `layerName` from the `name` here?  The `name` is the user-visible value used for the span name (IIUC), so it would be a little confusing to not match against that.

See also the matching handling that instrumentation-express does: supporting a string (exact match), regexp, or function for matching:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/80d0c74f8cafdfce681ef39e6cea103dca4e4ccd/plugins/node/opentelemetry-instrumentation-express/src/utils.ts#L120-L133

MrFabio: That was because this logic ran after the name is mutated. This json is what I have when processing the metadata
``` json
{
    "attributes": {
        "koa.name": "/documents",
        "koa.type": "router",
        "http.route": "/documents"
    },
    "name": "router - /documents",
    "layerName": "/documents" <--- original layer name
}
```

so I used the `layerName`. I will change it to use the interpolated name instead.

MrFabio:
Updated ✅

plugins/node/opentelemetry-instrumentation-koa/src/utils.ts
```diff
@@ -56,10 +56,49 @@ export const getMiddlewareMetadata = (
...
+  name?: string
 ): boolean => {
-  return !!(
+  if (
     Array.isArray(config?.ignoreLayersType) &&
     config?.ignoreLayersType?.includes(type)
-  );
+  ) {
+    return true;
+  }
```
trentm, line 67: Given that instrumentation.ts calls `isLayerIgnored` in two different ways -- once to check the `type` and once to check the `name` --  I think it would be cleaner to have two separate `isLayerTypeIgnored` and `isLayerNameIgnored` (or whatever names).

For example, I think the current `isLayerIgnored` has an issue: when it is called without a name: `isLayerIgnored(type, config)`, the code below that calls `satisfiesPattern` is still called, even though `name` is undefined. It shouldn't be called. Separating to two functions would make this cleaner.

plugins/node/opentelemetry-instrumentation-koa/README.md
```diff
@@ -54,6 +54,7 @@ Note that generator-based middleware are deprecated and won't be instrumented.
 | Options            | Type                                | Example              | Description                                                                                              |
 | ------------------ | ----------------------------------- | -------------------- | -------------------------------------------------------------------------------------------------------- |
 | `ignoreLayersType` | `KoaLayerType[]`                    | `['middleware']`     | Ignore layers of specified type.                                                                         |
+| `ignoreLayers`     | `IgnoreMatcher[]`                   | `['logger', /router/]`         | Ignore layers with specified names.                                                                      |
 | `requestHook`      | `KoaRequestCustomAttributeFunction` | `(span, info) => {}` | Function for adding custom attributes to Koa middleware layers. Receives params: `Span, KoaRequestInfo`. |
 
 `ignoreLayersType` accepts an array of `KoaLayerType` which can take the following string values:
```
trentm, line 60: Please also include the notes about `ignoreLayers` from the express README: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5eb61d8ea51371bea1b6030dca9c78861f4cd6d4/plugins/node/opentelemetry-instrumentation-express/README.md#L65-L69

trentm (changes requested):

plugins/node/opentelemetry-instrumentation-koa/README.md
MrFabio: Missed that, thanks!

plugins/node/opentelemetry-instrumentation-koa/src/types.ts
```diff
@@ -69,9 +69,13 @@ export interface KoaInstrumentationConfig<
...
+  /** Ignore specific layers based on their name */
+  ignoreLayers?: IgnoreMatcher[];
   /** Function for adding custom attributes to each middleware layer span */
   requestHook?: KoaRequestCustomAttributeFunction<
     KoaContextType,
     KoaMiddlewareType
   >;
 }
+
+export type IgnoreMatcher = string | RegExp | ((url: string) => boolean);
```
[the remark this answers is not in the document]
MrFabio: That's better!

plugins/node/opentelemetry-instrumentation-koa/test/utils.test.ts
```diff
@@ -45,17 +52,139 @@ describe('Utils', () => {
...
+        utils.isLayerTypeIgnored(KoaLayerType.MIDDLEWARE, {}),
+        false
+      );
+      assert.strictEqual(
+        utils.isLayerTypeIgnored(KoaLayerType.MIDDLEWARE, {
+          ignoreLayers: [],
+        } as KoaInstrumentationConfig),
+        false
+      );
+    });
```
trentm, line 103: Were these meant to have been adapted for `isLayerNameIgnored()`?

plugins/node/opentelemetry-instrumentation-koa/test/utils.test.ts
```diff
@@ -45,17 +52,139 @@ describe('Utils', () => {
...
       assert.strictEqual(
-        utils.isLayerIgnored(KoaLayerType.ROUTER, {
+        utils.isLayerTypeIgnored(KoaLayerType.ROUTER, {
           ignoreLayersType: [KoaLayerType.MIDDLEWARE],
         }),
         false
       );
     });
   });
+  describe('isLayerTypeIgnored()', () => {
```
trentm, line 68:
```suggestion
  describe('isLayerNameIgnored()', () => {
```
?

pichlermarc:
@trentm - I saw you reviewed this PR before - however this PR is targeting an unmaintained component. Are you sponsoring this feature? 🤔 If yes, please apply the `has:sponsor` label.

pichlermarc:
Thanks for your continued effort in keeping this PR up to date. That said, it’s been open for quite a while without much engagement from reviewers or clear interest from end-users. To help keep the project maintainable and avoid long-lived PRs on unmaintained components, I’m going to close this PR.

We're happy to revisit this feature once a new owner for @opentelemetry/instrumentation-koa is in place. If anyone is interested in becoming a component owner, please refer to this comment for more details: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2028#issuecomment-2211680625

PR #2030, opened 2024-03-21
chore: use node 18 in browser test setup

pichlermarc:
May allow us to update tooling, see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2013 where support for Node.js <18 was dropped.


pichlermarc:
Ah, of course. Needs webpack update.

PR #2036, opened 2024-03-24
fix(instrumentation-runtime-node): mixin default config in constructor

satazor:

The constructor of the instrumentation runtime node does not mixin the default config. This means that if I pass an empty object, the `eventLoopUtilizationMeasurementInterval` will be undefined, which has the side-effect of the configured interval to be near 0.

Similarly to all other instrumentations, I've mixed in the default config to make sure default objects are always defined. Please note that the instrumentation file itself didn't have tests, so I haven't added any.


satazor:
cc @d4nyll @pichlermarc @trentm

d4nyll (approved):
LGTM! Thanks @satazor

PR #2037, opened 2024-03-24
feat(instrumentation-koa): identify middleware functions inside route handlers

satazor:

Considering the following route handler:

```JS
const someMiddleware = () => async function some(ctx, next) => {
  await next();
});

router.get('/post/:id', someMiddleware(), (ctx) => {
  ctx.body = `Post id: ${ctx.params.id}`;
});
```

...will generate two spans with the same name: `router - /post/:id`, even though the first handler is actually a named middleware (`some`).

Ideally, it would generate two different spans:

- `middleware - some`
-  `router - /post/:id`
 
The changes are quite simple: when iterating these handlers, we assume they are a middleware if it's not the last handler on the stack and if it has a name.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
@@ -123,7 +123,9 @@ export class KoaInstrumentation extends InstrumentationBase<typeof koa> {
       const pathStack = pathLayer.stack;
       for (let j = 0; j < pathStack.length; j++) {
         const routedMiddleware: KoaMiddleware = pathStack[j];
-        pathStack[j] = this._patchLayer(routedMiddleware, true, path);
+        const isRouter = !routedMiddleware.name || j === pathStack.length - 1;
+
+        pathStack[j] = this._patchLayer(routedMiddleware, isRouter, path);
       }
     }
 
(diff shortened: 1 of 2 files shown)


satazor:
cc @dyladan

Not sure who can look into this since it seems there's no maintainer for this package.

satazor:
Shamelessly pinging @pichlermarc and @trentm 😅

diogotorres97 (approved):

satazor:
cc @blumamir

satazor:
Bump :/

diogotorres97:
Any news here?

pichlermarc:
`@opentelemetry/instrumentation-koa` is an [unmaintained component](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#unmaintained) (see [README.md](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-koa#status))

Reviews for new features for unmaintained components may be delayed until a new owner has been found as we try to prioritize work in the following ways:
- OTel API
- OTel SDK
- components that have owners 
- unmaintained components

We're currently working on better communicating our process as the existing documentation (a note in the readme and the link to the contributing.md) on this is not sufficient.

Issue #2025, opened 2024-03-19
Update usage of semantic conventions to new exported strings


JamieDanielson:
This is somewhat related to #1778 , which is for documenting current semantic convention versions for each instrumentation package.

This issue is intended to track the necessary change of updating each package to use the new exports for semantic conventions that were added in https://github.com/open-telemetry/opentelemetry-js/pull/4298 and now available in [v1.22.0](https://github.com/open-telemetry/opentelemetry-js/releases/tag/v1.22.0).

Each package will need to

- update the dependency for `@opentelemetry/semantic-conventions` to `"^1.22.0"`
- update the imported semantic convention to match the new version, i.e. `HTTP_ROUTE` becomes `SEMATTRS_HTTP_ROUTE`

Please limit PRs to a single package in order to make them quickly and easily reviewable.
If you are working on one of the below packages, please comment so others know the issue is being worked on.

- [x] `@opentelemetry/instrumentation-amqplib` #2086 
- [x] `@opentelemetry/instrumentation-cucumber` #2143
- [x] ~`@opentelemetry/instrumentation-dataloader`~
- [x] `@opentelemetry/instrumentation-fs` #2141 
- [x] `@opentelemetry/instrumentation-lru-memoizer` #2182
- [x] `@opentelemetry/instrumentation-mongoose` #2103 
- [x] `@opentelemetry/instrumentation-socket.io` #2147 
- [x] `@opentelemetry/instrumentation-tedious` #2144
- [x] `@opentelemetry/instrumentation-aws-lambda` #2181 
- [x] `@opentelemetry/instrumentation-aws-sdk` #2142 
- [x] `@opentelemetry/instrumentation-bunyan` #2195 
- [x] `@opentelemetry/instrumentation-cassandra-driver` #2139 
- [x] `@opentelemetry/instrumentation-connect` #2154
- [x] `@opentelemetry/instrumentation-dns` #2182
- [x] `@opentelemetry/instrumentation-express` #2031 
- [x] `@opentelemetry/instrumentation-fastify` #2078 
- [x] `@opentelemetry/instrumentation-generic-pool` #2182
- [x] `@opentelemetry/instrumentation-graphql` #2156 
- [x] `@opentelemetry/instrumentation-hapi` #2190 
- [x] `@opentelemetry/instrumentation-ioredis` #2191 
- [x] `@opentelemetry/instrumentation-knex` #2109 
[...]

PR #1816, opened 2023-11-20
chore: use `@web/test-runner` instead of `karma`

SimenB:
## Which problem is this PR solving?

- Attempting to unblock #1152 - i.e. bringing back the browser extension
- Note that this just affects `@opentelemetry/instrumentation-document-load` for now - I wanna get some feedback from the team that that this is something you want before spending more time on it 🙂 

## Short description of the changes

- As mentioned in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1152#issuecomment-1818373047, `karma` is deprecated, and they [recommend](https://blog.angular.io/moving-angular-cli-to-jest-and-web-test-runner-ef85ef69ceca) using Jest or `@web/test-runner` as a replacement. Seeing as I guess these theses _should_ run in the browser, I elected to go with `@web/test-runner`
- Under the hood `wtr` already uses `mocha`, so that migration is free. However, we also need to replace `webpack` with `rollup`/`esbuild` - that requires some more changes. I was unable to get `assert` to work, so I replaced it with `chai`.
- Coverage is collected, but I'm unsure if CI will correctly pick it up (EDIT: See later comments - the coverage is picked up, but something's weird with the merged report on codecov)

PR #2039, opened 2024-03-26
refactor(instr-document-load): use exported strings for semconv

JamieDanielson:

- updates #1778 
- updates #2025 

In`@opentelemetry/instrumentation-document-load` package:

- Update `@opentelemetry/semantic-conventions` to `^1.22`
- Replace deprecated `SemanticAttributes.*` with new exported strings `SEMATTRS_*`
- Update `README` with new semantic convention package version and keys

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38726,7 +38726,7 @@
         "@opentelemetry/instrumentation": "^0.49.1",
         "@opentelemetry/sdk-trace-base": "^1.0.0",
         "@opentelemetry/sdk-trace-web": "^1.15.0",
-        "@opentelemetry/semantic-conventions": "^1.0.0"
+        "@opentelemetry/semantic-conventions": "^1.22.0"
       },
       "devDependencies": {
         "@babel/core": "7.22.17",
@@ -45970,7 +45970,7 @@
         "@opentelemetry/instrumentation": "^0.49.1",
         "@opentelemetry/sdk-trace-base": "^1.0.0",
         "@opentelemetry/sdk-trace-web": "^1.15.0",
-        "@opentelemetry/semantic-conventions": "^1.0.0",
+        "@opentelemetry/semantic-conventions": "^1.22.0",
         "@rollup/plugin-commonjs": "^25.0.7",
         "@rollup/plugin-node-resolve": "^15.2.3",
         "@types/chai": "^4.3.10",
--- a/plugins/web/opentelemetry-instrumentation-document-load/README.md
+++ b/plugins/web/opentelemetry-instrumentation-document-load/README.md
@@ -111,6 +111,17 @@ registerInstrumentations({
 
 See [examples/tracer-web](https://github.com/open-telemetry/opentelemetry-js/tree/main/examples/tracer-web) for a short example.
 
+## Semantic Conventions
+
+This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+
+Attributes collected:
+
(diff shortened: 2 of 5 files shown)


JamieDanielson:
I got a timeout error suggesting to set/increase `browserStartTimeout` which defaults to 30000 (30 sec), so I pushed a commit that increases that by 5 seconds... before realizing I probably could've tried to just rerun the failed job (I didn't see the option until the rest of the checks finished). Will see what this does and can revert that change if it's unnecessary - haven't looked yet to see if this is a common "flaky" issue for browser tests or a one-off.

JamieDanielson:
🤔  Hmm a [different error](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8442296849/job/23123401306?pr=2039) this time:

> Failed to launch local browser installed at /usr/bin/google-chrome-stable. This could be because of a mismatch between the version of puppeteer and Chrome or Chromium. Try updating either of them, or adjust the executablePath option to point to another browser installation. Use the --puppeteer flag to run tests with bundled compatible version of Chromium.

It works locally so I'm curious to see what might be causing the error.

```sh
➜  opentelemetry-js-contrib git:(jamie.instr-document-load-semconv) npm run test:browser

> opentelemetry-contrib@0.1.0 test:browser
> lerna run test:browser --concurrency 1

lerna notice cli v6.6.2

    ✔  @opentelemetry/instrumentation-document-load:test:browser (3s)
    ✔  @opentelemetry/instrumentation-user-interaction:test:browser (4s)
    ✔  @opentelemetry/propagator-aws-xray:test:browser (3s)
    ✔  @opentelemetry/id-generator-aws-xray:test:browser (3s)
    ✔  @opentelemetry/instrumentation-long-task:test:browser (3s)
    ✔  @opentelemetry/plugin-react-load:test:browser (4s)
    ✔  @opentelemetry/propagator-instana:test:browser (3s)
    ✔  @opentelemetry/propagator-ot-trace:test:browser (3s)
    ✔  @opentelemetry/auto-instrumentations-web:test:browser (3s)

 ————————————————————————————————————————————————————————————————

 >  Lerna (powered by Nx)   Successfully ran target test:browser for 9 projects (29s)
```

JamieDanielson:
I added this option in the `web-test-runner.config.mjs` file to specify a [chromeLauncher with `--no-sandbox` based on the documentation](https://modern-web.dev/docs/test-runner/browser-launchers/overview/#configuring-browser-launchers), found from an issue online. Seems to fix the browser test error. If no other ideas I'm inclined to keep it? 

`browsers: [chromeLauncher({ launchOptions: { args: ['--no-sandbox'] } })],`

Looks like this test runner was updated in this package in #1816 so pinging @SimenB in case this seems familiar at all or would cause any known issues.

JamieDanielson:
> Did you try the puppeteer option like the error suggested?


Ah, yes I forgot about that. So the only reason I didn't use the `--puppeteer` flag is because it required another dependency and I wasn't sure if we wanted that. 🤔  I don't have strong feelings there as I'm not super familiar with the test runner, but it seemed good if a flag worked without adding another dependency 😅 

```
Error: You need to add @web/test-runner-puppeteer as a dependency of your project to use the --puppeteer flag.
```

pichlermarc:
Looks good. :+1: 

Odd that this error surfaced during this PR and not earlier. We ran it quite a few times automatically on `main` without any hiccups. :thinking: 

> Failed to launch local browser installed at /usr/bin/google-chrome-stable. This could be because of a mismatch between the version of puppeteer and Chrome or Chromium.

I wonder if this second error has something to do with the [installed browser version from the image used by the runner](https://github.com/actions/runner-images/blob/ubuntu22/20240317.1/images/ubuntu/Ubuntu2204-Readme.md). :thinking: I think it's worth to investigate this if we see this happening more frequently.

pichlermarc (approved):

PR #2048, opened 2024-03-27
chore(resource-detector-azure): use exported strings for attributes

maryliag:

- Part Of #2025 

On package `opentelemetry-resource-detector-azure`:
- Update `@opentelemetry/semantic-conventions` from `^1.0.0` to `^1.22.0`
- Use exported strings for Semantic Resource Attributes, Cloud Platform Values and Cloud Provider Values.


JamieDanielson:
Just a note, I recently updated the [CONTRIBUTING doc](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#changelog) because we no longer need to update the changelog on our own in PRs - it gets updated in the releasing step!

JamieDanielson (commented):
Looks like the lint step failed. You can ignore the warnings but locally try running `npm run lint:fix` and that should sort out the errors!

maryliag:
Updated to remove the changes on changelog.
I was having some issues making my lint run locally, after a few more tries and changes of versions (looks like my Node was too new), I was able to make it work now and fixed the changes 😄

JamieDanielson (approved):
Thanks for this!

maryliag:
@JamieDanielson thank you for the review! I don't have permission to merge it, so hopefully someone else can merge it for me 😄

trentm:
> I was having some issues making my lint run locally, after a few more tries and changes of versions (**looks like my Node was too new**)

🤨  Really? This was 2w ago, so you may not have details. If you do, I'm a little curious if there is an issue with a very new Node.js version.

trentm:
> If you do, I'm a little curious if there is an issue with a very new Node.js version.

FWIW, I was able to lint with Node v21.7.2:

```
nvm install v21.7.2
npm ci --ignore-scripts
npm run lint
```

maryliag:
I don't have the error messages anymore, but if it happens again in the future I will open an issue with details so it can be properly investigated 😄

PR #2049, opened 2024-03-27
chore(resource-detector-container): use exported strings for attributes

maryliag:

- Part Of #2025 

On package `opentelemetry-resource-detector-container`:
- Update `@opentelemetry/semantic-conventions` from `^1.0.0` to `^1.22.0`
- Use exported strings for Semantic Resource Attributes.

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-container/package.json
+++ b/detectors/node/opentelemetry-resource-detector-container/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@opentelemetry/resources": "^1.0.0",
-    "@opentelemetry/semantic-conventions": "^1.0.0"
+    "@opentelemetry/semantic-conventions": "^1.22.0"
   },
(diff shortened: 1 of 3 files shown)


JamieDanielson:
Just a note, I recently updated the [CONTRIBUTING doc](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#changelog) because we no longer need to update the changelog on our own in PRs - it gets updated in the releasing step!

JamieDanielson:
Hmm I think that browser test is flaky, it looks similar to the failure I experienced on [my PR too](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2039#issuecomment-2021530184).

maryliag:
Updated to remove the changes on changelog.
I'll keep an eye on your PR if I would need to do something similar (or maybe it will pass on the next test run? 🤞 )

pichlermarc (approved):

PR #2052, opened 2024-03-27
chore: add maryliag as CODEOWNER of opentelemetry-instrumentation-pg

maryliag:
Add `maryliag` as CODEOWNER of `plugins/node/opentelemetry-instrumentation-pg` as discussed on SIG meeting of March 27.


pichlermarc (approved):
Thanks for picking up ownership of this component @maryliag :slightly_smiling_face:

Issue #2055, opened 2024-03-27
improve resource detector documentation


maryliag:
On the README of the [resource detectors](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/detectors/node), there are no details about which resource attributes are being set and what are their respective keys, making it hard to identify what we have available.

Improve documentation, adding a table with the info for all resource detectors.

- [x] Alibaba #2056 
- [x] AWS #2060 
- [x] Azure #2057 
- [x] Container #2061 
- [x] GCP #2062 
- [x] Github #2064 
- [x] Instana #2063

trentm:
Done now, I believe.

maryliag:
@trentm thank you for all the reviews and merges!

PR #2056, opened 2024-03-27
chore: update readme for alibaba resource detector

maryliag:

- Part Of #2055

- Update README of Alibaba resource detector to add a table with values getting detected.


legendecas (commented):
Thanks, it would be great if a link to https://www.alibabacloud.com/help/en/ecs/user-guide/use-instance-identities could be added as well. It helps explaining where values like `owner-account-id` come from.

maryliag:
Thank you for the review! Link added

legendecas (approved):
Thank you

PR #2057, opened 2024-03-27
docs(resource-detector-azure): update readme for azure resource detector

maryliag:

- Part Of #2055
- Part Of #2025

- Update README of Azure Resource Detector to clarify the source of the data and also remove values from the table that were not actually being set. Update Semantic Conventions.


JamieDanielson (approved):
On the issue #1778 we discussed maybe not needing the Notes column with the key after all, but it's already here and I don't think it's necessary to do the extra work to remove it. If anything we can remove it when we update semantic conventions (if we need to), to minimize overhead.

maryliag:
column removed

hectorhdzg (approved):
LGTM

detectors/node/opentelemetry-resource-detector-azure/README.md
```diff
@@ -27,48 +27,48 @@ const tracerProvider = new NodeTracerProvider({ resource });
 
 ## Available Detectors
 
+This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
```
trentm, line 30: Ah... I pushed a commit to drop the `Key: CLOUD_RESOURCE_ID_RESOURCE_ATTRIBUTE` column below, but now I wonder if you'd intended to keep that because this is an attribute not in the current semconv package.

Hrm, actually does this mean that this package is implementing a Semantic Conventions version *newer* that 1.7.0? I think it does.

This package was added about 6 mo ago and `cloud.resource_id` was added to SemConv in v1.19.0: https://github.com/open-telemetry/semantic-conventions/blob/main/schemas/1.25.0#L283-L293

So now I wonder if we want this common statement about semconv version to change:
- The version of `@opentelemetry/semantic-conventions` isn't super-relevant to the user of a given package. I think it would be good for the README of the semantic-conventions package to include a table of its version to semconv version, but that's for a separate change.

What about something like:

```suggestion
This package implements Semantic Convention [Version 1.19.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md).
```

I think that's all the information of this package needs to know.

/cc @JamieDanielson

maryliag: Thank you for catching the one was still there! It was there by mistake, I intended to remove all. The new text makes sense to me, so I committed your suggestion.

JamieDanielson: Ooh good catch @trentm ! Yeah, hm. The thought I had around specifying the semantic-convention package version AND the associated semantic convention version was because I initially found it confusing that the two were not aligned and wanted it to be clear what version was implemented with the package. Also the package was always using a caret at the lowest possible value, so it didn't seem like it would change much. But it is ONLY relevant if those are the only semantic convention attributes being referenced - if that is not the case, then mentioning only the implemented version of the conventions makes sense.

detectors/node/opentelemetry-resource-detector-azure/README.md
```diff
@@ -27,48 +27,48 @@ const tracerProvider = new NodeTracerProvider({ resource });
 
 ## Available Detectors
 
+This package implements Semantic Convention [Version 1.19.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md).
```
JamieDanielson, line 30:
```suggestion
This package implements Semantic Convention [Version 1.19.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.19.0/semantic_conventions/README.md).
```

trentm: Doh, I missed updating the link. Take this suggestion. :)

trentm (approved):

PR #2063, opened 2024-03-28
docs(resource-detector-instana): update readme for instana resource detector

maryliag:

- Part Of #2055
- Part Of #2025

- Update README of Instana Resource Detector to clarify the source of the data and add Semantic Conventions

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-instana/README.md
+++ b/detectors/node/opentelemetry-resource-detector-instana/README.md
@@ -42,6 +42,17 @@ const sdk = new NodeSDK({
 sdk.start()
 ```
 
+## Available detectors & Semantic Conventions
+
+This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+
+### Instana Agent Detector
+
+| Resource Attribute           | Description                                                                                                              | Notes                                      |
+|------------------------------|--------------------------------------------------------------------------------------------------------------------------| ------------------------------------------ |
+| process.pid                  | The ID of the process. Value of parameter `pid` from the response to a request to `/com.instana.plugin.nodejs.discovery` | Key: `SEMRESATTRS_PROCESS_PID` |
+| service.instance.id          | The agent UUID. Value of parameter `agentUuid` from the response to a request to `/com.instana.plugin.nodejs.discovery`  | Key: `SEMRESATTRS_SERVICE_INSTANCE_ID` |
+
 ## Useful links
 
 - For more information about Instana Agent, visit: <https://www.ibm.com/docs/en/instana-observability/current?topic=instana-host-agent>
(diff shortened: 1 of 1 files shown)


kirrg001 (changes requested):
The change is not clear to me. 
 
> On the README of the [resource detectors](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/detectors/node), there are no details about which values are being detected and what are their respective keys, making it hard to identify what we have available.
> Improve documentation, adding a table with the info for all resource detectors.

I suggest to choose different headlines and put a sentence, which describes it.

maryliag:
what different headline do you suggest? I'm trying to keep consistent with existing docs for other resource detectors, which use "attributes" and "description"

maryliag:
Since this is about clarifying which resource attributes are being set by the detectors, and what are their sources, I update to. "Resource Attributes" instead of just  "Attributes" and updated the description. Let me know if that is better

kirrg001 (changes requested):
I still don't agree with this table for the public README.

What will a customer do with this table?
The new table you have added is IMO internal logic and should not belong in the public docs. If I am wrong, please explain to me.

The interesting attributes for the customers are:
- the two env variables, which are already documented
- and `[SemanticResourceAttributes.SERVICE_NAME]: "TestService"` (documented as well)

The detector reads the `process.pid` from the process and additionally returns the `agentUuid` based on the env variables (which is`service.instance.id`).

maryliag:
Currently, there is no documentation on what values are being set as the resources, this means that if a user wants to know what options they have available, so they can add filters, different types of charts, and so on, the only way of knowing what they have is to look into the actual code or start sending metrics/traces and check the values there, which is not ideal.
This table is an easy way for the user to know what values are there and where they come from, so they can use accordingly. 

This table exists for other README, but was missing for others, this is why I created PRs to add to the missing ones. This is a matter of keeping the documentation consistent in all READMEs. 

This also helps by showing what users can have by default, so they can also add anything that is no there. For example, the `SERVICE_NAME` that you mentioned doesn't come by default, only if they add it. Instead of them trying to figure out if they need to add something or not, they can check the table and add anything else that is not there by default, which on Instana case is just 2 values.

Examples: [Alibaba](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/detectors/node/opentelemetry-resource-detector-alibaba-cloud) and  [Azure](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/detectors/node/opentelemetry-resource-detector-azure)

kirrg001:
> This table is an easy way for the user to know what values are there and where they come from, so they can use accordingly.

I am not against the table, I just don't agree with how the table is presented and described. 

> This table exists for other README, but was missing for others, this is why I created PRs to add to the missing ones. This is a matter of keeping the documentation consistent in all READMEs.

I'll just approve then.

kirrg001 (approved):

trentm (approved):

trentm:
I have no idea why github-actions decided to add all the pkg: labels to this PR.

It would have been (from the timing) one of these two workflow runs:
- https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8616214861/job/23613343548
- https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8616223307/job/23613369106

However the "actions/labeler" unhelpfully logs no useful information, so there is nothing to go on.

PR #2071, opened 2024-03-30
fix(host-metrics): macOS bundling fix

Netail:

- Fix bundling

- Use specific file, instead of bundling all file of package


Netail:
@pichlermarc can you activate the tests for me?

packages/opentelemetry-host-metrics/src/stats/si.ts
```diff
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-import * as SI from 'systeminformation';
+import { networkStats } from 'systeminformation/lib/network';
```
pichlermarc, line 17: I'm not sure about `systeminformation` but deep-imports like this are usually not covered by stability guarantees. would this work for you?
```suggestion
import { networkStats } from 'systeminformation';
```

Netail: Yes

Netail:
Do the tests even work to begin with?

Netail:
CI/CD should be fixed now :)

david-luna:
Hi @Netail

Thank you for contributing to OTEL :)

We can leverage this PR to help effort the group is doing to update the semantic conventions. Currently we're in [a step](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2025) we need to use the exported strings as you did in this PR and also document which version of the semantic conventions is using.

Would you please add a section in the README file with this info. Here is an example of what I'm referring to.
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2039/files#diff-24b1cca18e57b052dc9ad991566c9e6cc4a10a47d8528e347ddda139d1ce705e

Netail:
> Hi @Netail
> 
> Thank you for contributing to OTEL :)
> 
> We can leverage this PR to help effort the group is doing to update the semantic conventions. Currently we're in [a step](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2025) we need to use the exported strings as you did in this PR and also document which version of the semantic conventions is using.
> 
> Would you please add a section in the README file with this info. Here is an example of what I'm referring to. https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2039/files#diff-24b1cca18e57b052dc9ad991566c9e6cc4a10a47d8528e347ddda139d1ce705e

So if I understand correctly, I have to add the semconvs to https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-semantic-conventions, use the package in host-metrics (replacing the ones specified here) and adding the ones used in the README with the table template?

Netail:
Just checked regarding opentelemetry-semantic-conventions package; it fetches the specifications from opentelemetry-specification, which got removed in v1.21.0, so the generate is broken... :(

david-luna:
> Hi @david-luna So if I understand correctly, I have to add all of https://github.com/Netail/opentelemetry-js-contrib/blob/fix/host-metrics-bundling/packages/opentelemetry-host-metrics/src/enum.ts to https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-semantic-conventions, use the package in host-metrics (replacing the package based ones) and adding the ones used from the semantic-conventions package in the README with the table template?

Semantic conventions already has these enums but the package is not updated yet so no need to contribute to https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-semantic-conventions. Actually this package is generated from the models in https://github.com/open-telemetry/semantic-conventions

We need to inform which version of semantic conventions model (not package) host metrics is using and list them. In this case the package uses v1.24.0 of the model
- system https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/system/system-metrics.md
- process https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/system/process-metrics.md

You could add a section with this message

```
## Semantic Conventions

This package uses Semantic Conventions [Version 1.24.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.24.0/docs/http). As for now the Semantic Conventions
are bundled in this package but eventually will be imported from `@opentelemetry/semantic-conventions` package when it is updated to latest version.
Ref: [opentelemetry-js/issues/4235](https://github.com/open-telemetry/opentelemetry-js/issues/4235)

Attributes collected:

| Attribute          | Short Description    | Notes                   |
| ------------------ | -------------------- | ----------------------- |
| `system.cpu.state` | The state of the CPU | Key: `SYSTEM_CPU_STATE` |
| ... other attributes  |
```

Netail:
> > Hi @david-luna So if I understand correctly, I have to add all of https://github.com/Netail/opentelemetry-js-contrib/blob/fix/host-metrics-bundling/packages/opentelemetry-host-metrics/src/enum.ts to https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-semantic-conventions, use the package in host-metrics (replacing the package based ones) and adding the ones used from the semantic-conventions package in the README with the table template?
> 
> Semantic conventions already has these enums but the package is not updated yet so no need to contribute to https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-semantic-conventions. Actually this package is generated from the models in https://github.com/open-telemetry/semantic-conventions
> 
> We need to inform which version of semantic conventions model (not package) host metrics is using and list them. In this case the package uses v1.24.0 of the model
> 
> * system https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/system/system-metrics.md
> * process https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/system/process-metrics.md
> 
> You could add a section with this message
> 
> ```
> ## Semantic Conventions
> 
> This package uses Semantic Conventions [Version 1.24.0](https://github.com/open-telemetry/semantic-conventions/tree/v1.24.0/docs/http). As for now the Semantic Conventions
> are bundled in this package but eventually will be imported from `@opentelemetry/semantic-conventions` package when it is updated to latest version.
> Ref: [opentelemetry-js/issues/4235](https://github.com/open-telemetry/opentelemetry-js/issues/4235)
> 
> Attributes collected:
> 
> | Attribute          | Short Description    | Notes                   |
> | ------------------ | -------------------- | ----------------------- |
> | `system.cpu.state` | The state of the CPU | Key: `SYSTEM_CPU_STATE` |
> | ... other attributes  |
> ```

Shall I add pull up a PR which generates the semantic concentions from the new repo?

Netail:
Updated the generator here: https://github.com/open-telemetry/opentelemetry-js/pull/4606

david-luna:
> Updated the generator here: https://github.com/open-telemetry/opentelemetry-js/pull/4606

this is indeed one of the tasks to be done for updating semantic conventions but since it a big leap in versions we expect breaking changes. This was discussed in the JavaScript SIG and planned. The plan is summarised and tracked in https://github.com/open-telemetry/opentelemetry-js/issues/4572

the PR looks good but 1st we need to prepare the packages in contrib to be ready for the change. You can help us with these preparation tasks :)

Also you can join the SIG to discuss this topic and others. Here is the doc with the MoMs and also with the info to join https://docs.google.com/document/d/1tCyoQK49WVcE-x8oryZOTTToFm7sIeUhxFPm9g-qL1k/edit

Netail:
> > Updated the generator here: [open-telemetry/opentelemetry-js#4606](https://github.com/open-telemetry/opentelemetry-js/pull/4606)
> 
> this is indeed one of the tasks to be done for updating semantic conventions but since it a big leap in versions we expect breaking changes. This was discussed in the JavaScript SIG and planned. The plan is summarised and tracked in [open-telemetry/opentelemetry-js#4572](https://github.com/open-telemetry/opentelemetry-js/issues/4572)
> 
> the PR looks good but 1st we need to prepare the packages in contrib to be ready for the change. You can help us with these preparation tasks :)
> 
> Also you can join the SIG to discuss this topic and others. Here is the doc with the MoMs and also with the info to join https://docs.google.com/document/d/1tCyoQK49WVcE-x8oryZOTTToFm7sIeUhxFPm9g-qL1k/edit

What kind of preparation do all the packages need? 

Updated https://github.com/open-telemetry/opentelemetry-js/pull/4606 to include v1.25.0 Semantic Conventions. As all packages within the same repo have a different version specified, we can go by them one by one in different PRs?

packages/opentelemetry-host-metrics/README.md
david-luna, line 1: I think the replacements of `=` and `,` to `|` are unintended right? Could you revert that?

packages/opentelemetry-host-metrics/src/BaseMetrics.ts
```diff
@@ -39,18 +37,18 @@ const DEFAULT_NAME = '@opentelemetry/host-metrics';
...
   constructor(config: MetricsCollectorConfig) {
     this._name = config.name || DEFAULT_NAME;
     const meterProvider =
-      config.meterProvider || api.metrics.getMeterProvider();
+      config.meterProvider || metrics.getMeterProvider();
     if (!config.meterProvider) {
       this._logger.warn('No meter provider, using default');
     }
-    this._meter = meterProvider.getMeter(this._name, VERSION);
+    this._meter = meterProvider.getMeter(this._name);
```
david-luna, line 51: Although is optional I think the `VERSION` here is useful. If I'm not mistaken this info will be added as metadata in the metrics so we will know which version of the package sent the data. There might be use cases for this

packages/opentelemetry-host-metrics/global.d.ts
david-luna, line 1: I guess with the change suggested by @pichlermarc in `./src/stats/si.ts`

```diff
- import { networkStats } from 'systeminformation/lib/network';
+ import { networkStats } from 'systeminformation';
```

this file is no needed. I see the `.d.ts` file from `systeminformation` exporting `networkStats` function already

david-luna:
> What kind of preparation do all the packages need?

@Netail please check https://github.com/open-telemetry/opentelemetry-js/issues/4572 for more info on the update plan, thanks :)

packages/opentelemetry-host-metrics/src/BaseMetrics.ts
Netail: Reverted

packages/opentelemetry-host-metrics/README.md
Netail: Oops, reverted

packages/opentelemetry-host-metrics/global.d.ts
Netail: When I use `import { networkStats } from 'systeminformation';`, it still tries to import the underlying cpu.js file, thus throwing an error on macOS. By targeting the file directly it won't

packages/opentelemetry-host-metrics/src/stats/si.ts
Netail: Sadly not, it still imports the cpu.js file in the package. Resulting in having to add the unnecessary dependency `osx-temperature-sensor`, which is not even used by networkStats

Netail:
@david-luna added the sem-conv table and added a comment on why the import has to be directly in order to make it work for macOS users

david-luna (approved):
LGTM

cc: @pichlermarc @legendecas

Netail:
@pichlermarc ready to merge? 😅

packages/opentelemetry-host-metrics/src/enum.ts
```diff
@@ -32,11 +32,9 @@ export enum ATTRIBUTE_NAMES {
   SYSTEM_CPU_STATE = 'system.cpu.state',
   SYSTEM_MEMORY_STATE = 'system.memory.state',
   SYSTEM_DEVICE = 'system.device',
-  SYSTEM_NETWORK_DIRECTION = 'system.network.direction',
+  NETWORK_NETWORK_DIRECTION = 'network.io.direction',
```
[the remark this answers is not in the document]
Netail: Ah yes, my bad

Netail:
@trentm in theory yes, but when I try to use the package it somehow triggers the require ;/

trentm:
> in theory yes, but when I try to use the package it somehow triggers the require ;/

Can you run it again with `NODE_DEBUG=module` set? That should show the chain of requires that leads to loading that module.

Netail:
Doesn't show anything else besides the original stack trace;
```
 ⚠ ../../node_modules/systeminformation/lib/cpu.js
Module not found: Can't resolve 'osx-temperature-sensor' in '<dir>/project/node_modules/systeminformation/lib'

Import trace for requested module:
../../node_modules/systeminformation/lib/cpu.js
../../node_modules/systeminformation/lib/index.js
../../node_modules/@opentelemetry/host-metrics/build/src/stats/si.js
../../node_modules/@opentelemetry/host-metrics/build/src/metric.js
../../node_modules/@opentelemetry/host-metrics/build/src/index.js
../../packages/ah-next-core/src/metrics/instrumentation.node.ts
../../packages/ah-next-core/src/metrics/instrumentation.ts
```

trentm (changes requested):
@Netail

1. Please run 'npm run lint:fix' to update style / reflow-indentation after your changes. That is why the "Lint" check is failing.

2. I also got a failure in `npm run compile`. This fixed it for me:

```diff
--- a/packages/opentelemetry-host-metrics/src/stats/common.ts
+++ b/packages/opentelemetry-host-metrics/src/stats/common.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */

-import { type CpuInfo, cpus, totalmem, freemem } from 'os';
+import type { CpuInfo } from 'os';
+import { cpus, totalmem, freemem } from 'os';
```


3. I don't know what you are experiencing with the `osx-temperature-sensor` thing. Things work fine for me (I am on macOS) after this change:

```diff
--- a/packages/opentelemetry-host-metrics/src/stats/si.ts
+++ b/packages/opentelemetry-host-metrics/src/stats/si.ts
@@ -14,10 +14,7 @@
  * limitations under the License.
  */

-// Import from network file directly as importing from the root imports the /lib/cpu file,
-// resulting in also having to add osx-temperature-sensor as a dependency for macOS,
-// while /lib/cpu isn't even used by this package (deep-importing not working as expected)
-import { networkStats } from 'systeminformation/lib/network';
+import { networkStats } from 'systeminformation';
```

In fact, the tests fail if I don't make that change, because the tests are currently mocking `systeminformation.networkStats` (see https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b903bce15d6ba4fa44daeaf6785abd6ed94cb5ff/packages/opentelemetry-host-metrics/test/metric.test.ts#L140). That mocking doesn't work when `import { networkStats } from 'systeminformation/lib/network';` is used.


* * *

Can you try with the above changes and see if `npm run compile` and `npm test` (in the packages/opentelemetry-host-metrics dir) work for you now?

If not, can you please provide more details on what exactly what commands you are running that lead to the attempt and failure to `require('osx-temperature-sensor')`.

Netail:
Thanks for the feedback @trentm ,
I fixed 1 & 2. 

3 still doesn't work for me, so I created a demo project; https://github.com/Netail/host-metrics-macos

Netail:
@trentm were you able to reproduce the issue?

trentm:
> so I created a demo project; https://github.com/Netail/host-metrics-macos

Ah, Next.js. When I ran `yarn install` and then `yarn dev` the first time in your example repo, I got:

<details>
<summary>first time</summary>

```
% yarn dev
yarn run v1.22.22
$ next dev
  ▲ Next.js 14.2.2
  - Local:        http://localhost:3000
  - Experiments (use with caution):
    · instrumentationHook

 ✓ Starting...
Attention: Next.js now collects completely anonymous telemetry regarding usage.
This information is used to shape Next.js' roadmap and prioritize features.
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://nextjs.org/telemetry

 ○ Compiling /instrumentation ...
 ⚠ ./node_modules/@opentelemetry/instrumentation/build/esm/platform/node/instrumentation.js
Critical dependency: the request of a dependency is an expression

Import trace for requested module:
./node_modules/@opentelemetry/instrumentation/build/esm/platform/node/instrumentation.js
./node_modules/@opentelemetry/instrumentation/build/esm/platform/node/index.js
./node_modules/@opentelemetry/instrumentation/build/esm/platform/index.js
./node_modules/@opentelemetry/instrumentation/build/esm/index.js
./instrumentation.node.ts

./node_modules/systeminformation/lib/cpu.js
Module not found: Can't resolve 'osx-temperature-sensor' in '/Users/trentm/tmp/host-metrics-macos/node_modules/systeminformation/lib'

Import trace for requested module:
./node_modules/systeminformation/lib/cpu.js
./node_modules/systeminformation/lib/index.js
./node_modules/@opentelemetry/host-metrics/build/src/stats/si.js
./node_modules/@opentelemetry/host-metrics/build/src/metric.js
./node_modules/@opentelemetry/host-metrics/build/src/index.js
./instrumentation.node.ts
TypeError: An error occurred while loading instrumentation hook: instrumentationHook.register is not a function
    at DevServer.runInstrumentationHookIfAvailable (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/dev/next-dev-server.js:437:43)
    at async Span.traceAsyncFn (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/trace/trace.js:154:20)
    at async DevServer.prepareImpl (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/dev/next-dev-server.js:214:9)
    at async NextServer.prepare (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/next.js:161:13)
    at async initializeImpl (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/lib/render-server.js:98:5)
    at async initialize (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/lib/router-server.js:423:22)
    at async Server.<anonymous> (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/lib/start-server.js:249:36)

✨  Done in 3.48s.
```

</details>

Which included this in the output:

```
./node_modules/systeminformation/lib/cpu.js
Module not found: Can't resolve 'osx-temperature-sensor' in '/Users/trentm/tmp/host-metrics-macos/node_modules/systeminformation/lib'
```

There is also this error output:

```
TypeError: An error occurred while loading instrumentation hook: instrumentationHook.register is not a function
```

The `yarn dev` exited with a `0` (i.e. success) status code. Pardon my Next.js ignorance, but is that a breakage?

---

After an edit to "node_modules/@opentelemetry/host-metrics/build/src/stats/si.js" to use:

``` js
const SI = require("systeminformation/lib/network");
```

I get:

<details>
<summary>after edit</summary>

```
% yarn dev
yarn run v1.22.22
$ next dev
  ▲ Next.js 14.2.2
  - Local:        http://localhost:3000
  - Experiments (use with caution):
    · instrumentationHook

 ✓ Starting...
 ○ Compiling /instrumentation ...
 ⚠ ./node_modules/@opentelemetry/instrumentation/build/esm/platform/node/instrumentation.js
Critical dependency: the request of a dependency is an expression

Import trace for requested module:
./node_modules/@opentelemetry/instrumentation/build/esm/platform/node/instrumentation.js
./node_modules/@opentelemetry/instrumentation/build/esm/platform/node/index.js
./node_modules/@opentelemetry/instrumentation/build/esm/platform/index.js
./node_modules/@opentelemetry/instrumentation/build/esm/index.js
./instrumentation.node.ts
TypeError: An error occurred while loading instrumentation hook: instrumentationHook.register is not a function
    at DevServer.runInstrumentationHookIfAvailable (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/dev/next-dev-server.js:437:43)
    at async Span.traceAsyncFn (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/trace/trace.js:154:20)
    at async DevServer.prepareImpl (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/dev/next-dev-server.js:214:9)
    at async NextServer.prepare (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/next.js:161:13)
    at async initializeImpl (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/lib/render-server.js:98:5)
    at async initialize (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/lib/router-server.js:423:22)
    at async Server.<anonymous> (/Users/trentm/tmp/host-metrics-macos/node_modules/next/dist/server/lib/start-server.js:249:36)

✨  Done in 2.01s.
```

</details>

This again exits with status 0, and has the `TypeError: An error occurred while loading instrumentation hook: instrumentationHook.register is not a function` message in the output.

Does that `TypeError` need to be dealt with first?

---

So my (limited) understanding and guess is that this is related to Next.js doing *bundling*?
Is that `Can't resolve 'osx-temperature-sensor'` message actually fatal for Next.js? Or will the Next.js app still work because we don't expect a runtime call to `require('osx-temperature-sensor')`?

Could the Next.js app be configured to stub out or mock that package? (I understand it would be poor of OTel JS to require Next.js users of `@opentelemetry/host-metrics` to have to know to do that.)   I'm not that familiar with bundler issues and Next.js, so I'm trying to understand.

trentm:
> Could the Next.js app be configured to stub out or mock that package? (I understand it would be poor of OTel JS to require Next.js users of `@opentelemetry/host-metrics` to have to know to do that.) I'm not that familiar with bundler issues and Next.js, so I'm trying to understand.

See https://github.com/sebhildebrandt/systeminformation/issues/701

trentm:
Actually there are a bunch of (closed) issues on systeminformation:

https://github.com/sebhildebrandt/systeminformation/issues?q=osx-temperature-sensor

that all (at least the 5 or so that I read) point to this issue of getting

```
Module not found: Can't resolve 'osx-temperature-sensor' in '<dir>/project/node_modules/systeminformation/lib'
```

when using this module with webpack.

That error message is from webpack code here:
https://github.com/webpack/webpack/blob/v5.91.0/lib/ModuleNotFoundError.js#L53-L54


(Side note: There is a comment about removing that in webpack 6, for what it is worth. I'm not sure exactly what is implied by that.)

(Side note: The amount of confusion that webpack could reduce or could have reduced by just putting "webpack" in their error messages is large.)

Netail:
> This again exits with status 0, and has the TypeError: An error occurred while loading instrumentation hook: instrumentationHook.register is not a function message in the output.

My mistake, oopsie. Changed this before pushing, should be fixed now.

Netail:
As far as I know, `instrumentation.ts` is the only way currently in NextJS. 

Also recommended according to the docs; https://nextjs.org/docs/app/building-your-application/optimizing/open-telemetry

Netail:
I noticed the `Module not found: Can't resolve 'osx-temperature-sensor' in '<dir>/project/node_modules/systeminformation/lib'` also occurs when building the nextjs app in GitHub actions with linux runners

r34son:
Getting same error:
```sh
 ✓ Starting...
 ○ Compiling /instrumentation ...
 ⚠ ./node_modules/.pnpm/systeminformation@5.22.8/node_modules/systeminformation/lib/cpu.js
Module not found: Can't resolve 'osx-temperature-sensor' in '/Users/seitasanov/ownProjects/profile/node_modules/.pnpm/systeminformation@5.22.8/node_modules/systeminformation/lib'

Import trace for requested module:
./node_modules/.pnpm/systeminformation@5.22.8/node_modules/systeminformation/lib/cpu.js
./node_modules/.pnpm/systeminformation@5.22.8/node_modules/systeminformation/lib/index.js
./node_modules/.pnpm/@opentelemetry+host-metrics@0.35.1_@opentelemetry+api@1.8.0/node_modules/@opentelemetry/host-metrics/build/src/stats/si.js
./node_modules/.pnpm/@opentelemetry+host-metrics@0.35.1_@opentelemetry+api@1.8.0/node_modules/@opentelemetry/host-metrics/build/src/metric.js
./node_modules/.pnpm/@opentelemetry+host-metrics@0.35.1_@opentelemetry+api@1.8.0/node_modules/@opentelemetry/host-metrics/build/src/index.js
```

r34son:
@Netail You can also get rid of this message by modifying nextjs webpack config:
```ts
  webpack: (config, { webpack }) => {
    config.plugins.push(
      new webpack.IgnorePlugin({ resourceRegExp: /osx-temperature-sensor/ }),
    );
    return config;
  },
```

Netail:
> @Netail You can also get rid of this message by modifying nextjs webpack config:
> 
> ```ts
>   webpack: (config, { webpack }) => {
>     config.plugins.push(
>       new webpack.IgnorePlugin({ resourceRegExp: /osx-temperature-sensor/ }),
>     );
>     return config;
>   },
> ```

I know, but does bloat your next config. Would prefer a solution to mitigate this closer at the source

Netail:
@trentm any update? 👀

trentm:
Sorry, I haven't had a chance to get back to this.

Partly I'm uncertain about the boundaries of what soft rules OTel JS code should break to accommodate different usage by bundlers.  The debate here is between (a) doing a deep require into a dependency that might break in a future update vs. (b) having a warning (error?) in usage of the package with bundlers.


I still wasn't certain from my [repro attempt above](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2071#issuecomment-2088655072) if these messages from the Next.js build were *warnings* or *errors*.

> Is that `Can't resolve 'osx-temperature-sensor'` message actually fatal for Next.js? Or will the Next.js app still work because we don't expect a runtime call to `require('osx-temperature-sensor')`?

trentm:
I'll ask in the OTel JS SIG today.

trentm:
> on every page render

Every page render is re-resolving every dependency? Or is that when running in `next dev` Dev-mode that might be re-compiling each time to get recent changes?

packages/opentelemetry-host-metrics/README.md
```diff
@@ -39,6 +39,39 @@ const hostMetrics = new HostMetrics({ meterProvider, name: 'example-host-metrics
...
+Ref: [opentelemetry-js/issues/4235](https://github.com/open-telemetry/opentelemetry-js/issues/4235)
+
+Metrics collected:
+
+| Metric                      | Short Description                                         |
+| --------------------------- | --------------------------------------------------------- |
+| `system.cpu.time`           | Seconds each logical CPU spent on each mode               |
+| `system.cpu.utilization`    | Difference in system.cpu.time since the last measurement  |
+| `system.memory.usage`       | Reports memory in use by state                            |
+| `system.memory.utilization` |                                                           |
```
maryliag, line 55: can you add a description for it?

packages/opentelemetry-host-metrics/README.md
```diff
@@ -39,6 +39,39 @@ const hostMetrics = new HostMetrics({ meterProvider, name: 'example-host-metrics
...
+| `system.network.dropped`    | Count of packets that are dropped                         |
+| `system.network.errors`     | Count of network errors detected                          |
+| `system.network.io`         |                                                           |
```
maryliag, line 58: can you add a description for it?

packages/opentelemetry-host-metrics/README.md
```diff
@@ -39,6 +39,39 @@ const hostMetrics = new HostMetrics({ meterProvider, name: 'example-host-metrics
...
+| `process.cpu.time`          | Total CPU seconds                                         |
+| `process.cpu.utilization`   | Difference in process.cpu.time since the last measurement |
+| `process.memory.usage`      | The amount of physical memory in use                      |
+
+Attributes collected:
+
+| Metric                      | Short Description                  | Notes |
```
maryliag, line 65: Since there is nothing on the Notes column, it can be removed

Netail: Applied

packages/opentelemetry-host-metrics/README.md
Netail: Applied

Netail:
> > on every page render
> 
> Every page render is re-resolving every dependency? Or is that when running in `next dev` Dev-mode that might be re-compiling each time to get recent changes?

Not completely sure, I believe it runs the instrumentation.ts on every initial page render in development (local) mode.

legendecas:
systeminformation doesn't want to solve this issue at their side: https://github.com/sebhildebrandt/systeminformation/issues/230, and the module `'osx-temperature-sensor'` can be externalized to avoid being bundled.

```js
// webpack.config.js
module.exports = {
  externals: ['osx-temperature-sensor'],
};
```

I'd avoid using subpath imports which are not considered as public interface.

Additionally, I would suggest splitting the PR for distinct purposes.

trentm:
We discussed in the SIG a couple days ago (https://docs.google.com/document/d/1tCyoQK49WVcE-x8oryZOTTToFm7sIeUhxFPm9g-qL1k/edit#heading=h.bxgyx12a4wsc) and decided that if we pinned the systeminformation dep, we'd be okay using the deep import.

From memory, my summary of the justification is:
- A macOS user of `@opentelemetry/host-metrics` *and* Webpack may not be able to easily understand the impact of the error/warning message and know how to cope with it.
- We could document the work-around in the host-metrics README or perhaps more visible "Troubleshooting" docs on opentelemetry.io, I suppose.
- The use of host-metrics could possibly be indirect from the user's p.o.v. if, for example, the package is included in a custom vendor SDK.
- The use of webpack could be indirect from the user's point of view. E.g. for Next.js users.
- Certainly attributing the error message about `osx-temperature-sensor` to something used by `host-metrics` is indirect.
- The structure of `systeminformation` looks like a top-level index that re-exports a number of independent functional sub-packages. I'm guessing (this is my opinion here) that it is unlikely that a future version would break the deep import we are proposing to use.  At a guess, the most likely thing that would break this is if `systeminformation/package.json` added an "exports" section for unrelated reasons.
- Pinning and using a deep import kind of sucks, yes.
- If was mentioned that considering pulling in just the functionality we want from `systeminformation` into this repo was **undesired**. Having an effective fork means we lose the connection for possible fixes. There could be licensing issues/pain.


Aside: I'm not sure if bundlers other than Webpack result in a similar error/warning.


TODO:
- Pin the systeminformation version in package.json.
- Document some of the above concerns in the code where the deep import is being made to help future maintenance.
- Splitting into two PRs, as legendecas suggested, *would* really help clarify for future maintenance, I agree.

(@Netail I see you recently commited removing the deep import. Sorry for back and forth on this. I should have posted notes from the SIG a couple days ago.)

Netail:
> We discussed in the SIG a couple days ago (https://docs.google.com/document/d/1tCyoQK49WVcE-x8oryZOTTToFm7sIeUhxFPm9g-qL1k/edit#heading=h.bxgyx12a4wsc) and decided that if we pinned the systeminformation dep, we'd be okay using the deep import.
> 
> From memory, my summary of the justification is:
> 
> * A macOS user of `@opentelemetry/host-metrics` _and_ Webpack may not be able to easily understand the impact of the error/warning message and know how to cope with it.
> * We could document the work-around in the host-metrics README or perhaps more visible "Troubleshooting" docs on opentelemetry.io, I suppose.
> * The use of host-metrics could possibly be indirect from the user's p.o.v. if, for example, the package is included in a custom vendor SDK.
> * The use of webpack could be indirect from the user's point of view. E.g. for Next.js users.
> * Certainly attributing the error message about `osx-temperature-sensor` to something used by `host-metrics` is indirect.
> * The structure of `systeminformation` looks like a top-level index that re-exports a number of independent functional sub-packages. I'm guessing (this is my opinion here) that it is unlikely that a future version would break the deep import we are proposing to use.  At a guess, the most likely thing that would break this is if `systeminformation/package.json` added an "exports" section for unrelated reasons.
> * Pinning and using a deep import kind of sucks, yes.
> * If was mentioned that considering pulling in just the functionality we want from `systeminformation` into this repo was **undesired**. Having an effective fork means we lose the connection for possible fixes. There could be licensing issues/pain.
> 
> Aside: I'm not sure if bundlers other than Webpack result in a similar error/warning.
> 
> TODO:
> 
> * Pin the systeminformation version in package.json.
> * Document some of the above concerns in the code where the deep import is being made to help future maintenance.
> * Splitting into two PRs, as legendecas suggested, _would_ really help clarify for future maintenance, I agree.
> 
> (@Netail I see you recently commited removing the deep import. Sorry for back and forth on this. I should have posted notes from the SIG a couple days ago.)

Alright! Thanks. I've created a seperate PR for the semconv alignment & documentation; https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2240

Netail:
Alright, I've put the deep import back & the semconv alignment PR has been merged in a seperate PR. :)

packages/opentelemetry-host-metrics/package.json
```diff
@@ -56,7 +56,7 @@
   },
   "dependencies": {
     "@opentelemetry/sdk-metrics": "^1.8.0",
-    "systeminformation": "^5.21.20"
+    "systeminformation": "5.22.9"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/opentelemetry-host-metrics#readme"
-}
+}
```
legendecas, line 62: Please add back the EOF.

packages/opentelemetry-host-metrics/tsconfig.json
```diff
@@ -6,6 +6,7 @@
   },
   "include": [
     "src/**/*.ts",
-    "test/**/*.ts"
+    "test/**/*.ts",
+    "global.d.ts"
   ]
-}
+}
```
legendecas, line 12: Please add EOF.

packages/opentelemetry-host-metrics/test/metric.test.ts
```diff
@@ -25,7 +26,7 @@ import {
   MetricReader,
 } from '@opentelemetry/sdk-metrics';
 import * as assert from 'assert';
-import * as os from 'os';
+import * as os from 'node:os';
```
legendecas, line 29: I'd avoid mixing the style of with and without the prefix `node:`. Could you remove unrelated changes?

legendecas (commented):
LGTM % nits.

packages/opentelemetry-host-metrics/package.json
Netail: Interesting, you would think that linting would pick this up

packages/opentelemetry-host-metrics/test/metric.test.ts
Netail: I will remove them, but we do need to add them for the future though incase people use different js runtimes e.g. bun or deno.

And to bypass require cache.

https://github.com/import-js/eslint-plugin-import/issues/2717#issue-1584570798

https://stateful.com/blog/node-18-prefix-only-modules

trentm: If, for example, we needed to use `node:test` -- which is only available with the `node:`-prefix -- then yes we'd need to use the prefix.

However, currently we support `"engines": { "node": ">=14" }` in most package.json files. That is relatively loose on defining *which* Node.js v14 versions we support, but at least v14.0.0 didn't support `node:` prefixes:

```
Welcome to Node.js v14.0.0.
> require('node:os')
Uncaught Error: Cannot find module 'node:os'
Require stack:
- <repl>
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:1014:15)
    at Function.Module._load (internal/modules/cjs/loader.js:884:27)
    at Module.require (internal/modules/cjs/loader.js:1074:19)
    at require (internal/modules/cjs/helpers.js:72:18) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '<repl>' ]
}
>
```

Also, there is no rush on the usage with different runtimes. My guess (granted a guess), is that there is a ton of other work that would be required to get many of the OTel packages working reasonably with bun, deno, etc. We'd want to coordinate that in separate PRs.

packages/opentelemetry-host-metrics/package.json
trentm: We currently have linting of .ts and .md files, but not of the package.json files.

packages/opentelemetry-host-metrics/test/metric.test.ts
Netail: Oh that's true, do note that 14 and 16 hit their eol. So should we still support it tho?

Netail: Btw having some trouble with adding an eof :(

legendecas (approved):

packages/opentelemetry-host-metrics/src/stats/si.ts
trentm: I'm resolving this conversation. It was discussed in separate comments on this PR and in the SIG call recently.

trentm (approved):
Thanks for persisting.

packages/opentelemetry-host-metrics/src/BaseMetrics.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import * as api from '@opentelemetry/api';
-import * as metrics from '@opentelemetry/sdk-metrics';
+import { Meter, diag, metrics } from '@opentelemetry/api';
+import { MeterProvider } from '@opentelemetry/sdk-metrics';
```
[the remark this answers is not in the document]
Netail: Yes they are indeed to avoid import * for tree shaking purposes. It's not necessarily for the bug I encountered, but is related to bundling tho

Netail:
fyi; I am not allowed to merge, so when you guys are ready. feel free to merge :)

legendecas:

PR #2072, opened 2024-03-30
fix(instrumentation-redis): Take host and port used for connection

Zirak:

The Redis client allows specifying connection options in several ways, with sensible defaults. The following all translate into `127.0.0.1:6379`:

```js
createClient('redis://127.0.0.1:6379');
createClient({ host: '127.0.0.1', port: NaN });
createClient({})
createClient()
```

Redis somewhat normalises these separate options into its `options` member, and stores the properties it uses to connect to the server in `connection_options`. Examples of the difference between the two in the examples preceding (slightly redacted for ease of reading):

```js
createClient('redis://127.0.0.1:6379');
// options = { port: '6379', host: '127.0.0.1' }
// connection_options = { port: 6379, host: '127.0.0.1', family: 4 }

createClient({ host: '127.0.0.1', port: NaN });
// options = { host: '127.0.0.1', port: NaN }
// connection_options = { port: 6379, host: '127.0.0.1', family: 4 }

createClient()
// options = { host: undefined }
// connection_options = { port: 6379, host: '127.0.0.1', family: 4 }
```

The instrumentation before this commit looks at the `options` property, which contains caller-supplied values before they're fully normalised and smoothed over by Redis. This means that for these weird cases, the instrumentation would populate `NET_PEER_NAME` and `NET_PEER_PORT` with erroneous values.

This commit has the instrumentation the values the Redis client uses to connect to the server, mirroring actual use.


plugins/node/opentelemetry-instrumentation-redis/test/redis.test.ts
```diff
@@ -47,7 +47,7 @@ const memoryExporter = new InMemorySpanExporter();
 
 const CONFIG = {
   host: process.env.OPENTELEMETRY_REDIS_HOST || 'localhost',
-  port: process.env.OPENTELEMETRY_REDIS_PORT || '63790',
+  port: process.env.OPENTELEMETRY_REDIS_PORT || 63790,
```
Zirak, line 50: Testing these changes proved difficult. The container listens on `63790`, while the default redis port is `6379`. The fix revolves around what the client and instrumentation do when dealing with default values. So if we provide an invalid port, the client will try (and fail) to connect, and the test will fail before getting to the instrumentation.

So I'm unsure what to do here.

david-luna:
@Zirak thanks for the contribution to this package

tests are failing because a formatting issue in the `net.peer.port` attribute. Test expects a string value instead of a number but semantic conventions says it should be a number
https://github.com/open-telemetry/opentelemetry-specification/blob/13358927bf91aa07ada15fdd2c2ab5c4c83e7628/semantic_conventions/trace/general.yaml#L42

I think the test should be changed to expect a number in `net.peer.port`

plugins/node/opentelemetry-instrumentation-redis/test/redis.test.ts
blumamir: sounds not straightforward to test. I think the original intent of using non standard port was not to conflict with users environments that are already running redis.

If we try to send redis command to `6379`, will it end the span with an error, or simply not create any span? perhaps we can verify the recording of those default attributes on the erroneous span?

Zirak: That's a good idea, but this Redis instrumentation (unlike redis4) only creates spans for commands, and not for the connection. The connection happens when the client is created, before any commands are sent

If it's desirable I can add a span for making the connection - but that feels like it should be done for its own sake, and not just for testing for default values (though we can think of something like e.g. having it disabled by default, like a `createConnectionSpan` configuration value)

david-luna (approved):

blumamir (approved):

PR #2076, opened 2024-04-03
feat(deps): update otel-js experimental to 0.50.0

pichlermarc:
This PR updates core dependencies to [v0.50.0](https://github.com/open-telemetry/opentelemetry-js/releases/tag/experimental%2Fv0.50.0).


Flarna (approved):
headline is not really correct as nothing is updated to 1.23.0, only the experimental deps are changed.

pichlermarc:
> headline is not really correct as nothing is updated to 1.23.0, only the experimental deps are changed.

Yep, that's true, thanks for pointing this out.  :+1: 
I changed to just mention the experimental deps (v0.50.0)

PR #2080, opened 2024-04-03
chore(test-utils): update semantic resource attributes

maryliag:

- Part Of #2025

- Update @opentelemetry/semantic-conventions from ^1.0.0 to ^1.22.0
- Use exported strings for Semantic Resource Attributes

### DIFF
--- a/packages/opentelemetry-test-utils/README.md
+++ b/packages/opentelemetry-test-utils/README.md
@@ -69,37 +69,37 @@ That's it - supper short and easy.
 
 This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
 
-| Attributes              | Description                                                                                   | Notes                                     |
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-test-utils/README.md
```diff
@@ -65,6 +65,42 @@ it('some test', () => {
...
+| process.executable.name | The name of the process executable.                                                           |
+| process.pid             | Process identifier (PID).                                                                     |
+| service.instance.id     | The string ID of the service instance.                                                        |
+| service.name            | Logical name of the service.                                                                  |
+| service.namespace       | A namespace for `service.name`.                                                               |
+| service.version         | The version string of the service API or implementation.                                      |
+| telemetry.sdk.language  | The language of the telemetry SDK.                                                            |
+| telemetry.sdk.name      | The name of the telemetry SDK.                                                                |
+| telemetry.sdk.version   | The version string of the telemetry SDK.                                                      |
```
trentm, line 102: I'd be inclined to drop the table for this package, an internal test utils package. But totally fine to have it in.

maryliag: Agree. I removed the table, but keep the message of semantic version, to keep consistent with other places.

PR #2085, opened 2024-04-04
chore(instr-undici): updates/tweaks for new undici instrumentation

trentm:
- 'codecov' npm script was copypasta from core repo
- update to same test-all-versions ver as other pkgs in the workspace
- update undici to latest (it is a security update)
- add label-prs workflow support pkg:instrumentation-undici
- reduce number of versions tested with TAV
- skip a particular failing test case for very old undici versions

* * *
I also added the pkg: label for this instr: https://github.com/open-telemetry/opentelemetry-js-contrib/labels?q=undici

### DIFF
--- a/plugins/node/instrumentation-undici/.tav.yml
+++ b/plugins/node/instrumentation-undici/.tav.yml
@@ -1,8 +1,12 @@
 undici:
   jobs:
-    - versions: ">=5 <6"
+    - versions:
+        include: ">=5 <6"
+        mode: max-7
       node: '>=14'
       commands: npm run test
-    - versions: ">=6 <7"
+    - versions:
+        include: ">=6 <7"
+        mode: max-7
       node: '>=18'
       commands: npm run test
--- a/plugins/node/instrumentation-undici/.tav.yml
+++ b/plugins/node/instrumentation-undici/.tav.yml
@@ -1,7 +1,7 @@
 undici:
   jobs:
     - versions:
-        include: ">=5 <6"
+        include: ">=5.12.0 <6"
         mode: max-7
       node: '>=14'
       commands: npm run test
--- a/plugins/node/instrumentation-undici/README.md
+++ b/plugins/node/instrumentation-undici/README.md
@@ -18,7 +18,6 @@ npm install --save @opentelemetry/instrumentation-undici
 
 - `undici@>=5.12.0`
 
-
 ## Usage
 
 OpenTelemetry Undici/fetch Instrumentation allows the user to automatically collect trace data and export them to their backend of choice, to give observability to distributed systems.
--- a/plugins/node/instrumentation-undici/README.md
+++ b/plugins/node/instrumentation-undici/README.md
@@ -50,11 +50,11 @@ Undici instrumentation has few options available to choose from. You can set the
 
 | Options | Type | Description |
 | ------- | ---- | ----------- |
-| [`ignoreRequestHook`](./src/types.ts#63) | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function. |
-| [`requestHook`](./src/types.ts#65) | `RequestHookFunction` | Function for adding custom attributes before request is handled. |
-| [`startSpanHook`](./src/types.ts#67) | `StartSpanHookFunction` | Function for adding custom attributes before a span is started. |
-| [`requireParentforSpans`](./src/types.ts#69) | `Boolean` | Require a parent span is present to create new span for outgoing requests. |
-| [`headersToSpanAttributes`](./src/types.ts#71) | `Object` |  List of case insensitive HTTP headers to convert to span attributes. Headers will be converted to span attributes in the form of `http.{request\|response}.header.header-name` where the name is only lowercased, e.g. `http.response.header.content-length`|
+| [`ignoreRequestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#63) | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function. |
+| [`requestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#65) | `RequestHookFunction` | Function for adding custom attributes before request is handled. |
+| [`startSpanHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#67) | `StartSpanHookFunction` | Function for adding custom attributes before a span is started. |
+| [`requireParentforSpans`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#69) | `Boolean` | Require a parent span is present to create new span for outgoing requests. |
+| [`headersToSpanAttributes`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#71) | `Object` |  List of case insensitive HTTP headers to convert to span attributes. Headers will be converted to span attributes in the form of `http.{request\|response}.header.header-name` where the name is only lowercased, e.g. `http.response.header.content-length`|
 
 ### Observations
 
--- a/plugins/node/instrumentation-undici/README.md
+++ b/plugins/node/instrumentation-undici/README.md
@@ -46,11 +46,11 @@ Undici instrumentation has few options available to choose from. You can set the
 
 | Options | Type | Description |
 | ------- | ---- | ----------- |
-| [`ignoreRequestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#63) | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function. |
-| [`requestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#65) | `RequestHookFunction` | Function for adding custom attributes before request is handled. |
-| [`startSpanHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#67) | `StartSpanHookFunction` | Function for adding custom attributes before a span is started. |
-| [`requireParentforSpans`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#69) | `Boolean` | Require a parent span is present to create new span for outgoing requests. |
-| [`headersToSpanAttributes`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#71) | `Object` |  List of case insensitive HTTP headers to convert to span attributes. Headers will be converted to span attributes in the form of `http.{request\|response}.header.header-name` where the name is only lowercased, e.g. `http.response.header.content-length`|
+| [`ignoreRequestHook`](./src/types.ts#63) | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function. |
+| [`requestHook`](./src/types.ts#65) | `RequestHookFunction` | Function for adding custom attributes before request is handled. |
+| [`startSpanHook`](./src/types.ts#67) | `StartSpanHookFunction` | Function for adding custom attributes before a span is started. |
+| [`requireParentforSpans`](./src/types.ts#69) | `Boolean` | Require a parent span is present to create new span for outgoing requests. |
+| [`headersToSpanAttributes`](./src/types.ts#71) | `Object` |  List of case insensitive HTTP headers to convert to span attributes. Headers will be converted to span attributes in the form of `http.{request\|response}.header.header-name` where the name is only lowercased, e.g. `http.response.header.content-length`|
 
 ### Observations
 
--- a/plugins/node/instrumentation-undici/README.md
+++ b/plugins/node/instrumentation-undici/README.md
@@ -14,6 +14,11 @@ If you're looking the instrumentation for browser's `fetch` API it is located at
 npm install --save @opentelemetry/instrumentation-undici
 ```
 
+## Supported Versions
+
+- `undici@>=5.12.0`
+
+
 ## Usage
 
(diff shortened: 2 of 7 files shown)


trentm:
This .tav.yml change reduces the num,ber of undici versions tested from 77 to 14:

Before: 

```
% npx tav --dry-run --verbose | rg match
-- 20 package versions matching filter: 6.0.0, 6.0.1, 6.1.0, 6.2.0, 6.2.1, 6.3.0, 6.4.0, 6.5.0, 6.6.0, 6.6.1, 6.6.2, 6.7.0, 6.7.1, 6.8.0, 6.9.0, 6.10.0, 6.10.1, 6.10.2, 6.11.0, 6.11.1
-- 57 package versions matching filter: 5.0.0, 5.1.0, 5.1.1, 5.2.0, 5.3.0, 5.4.0, 5.5.0, 5.5.1, 5.6.0, 5.6.1, 5.7.0, 5.8.0, 5.8.1, 5.8.2, 5.9.0, 5.9.1, 5.10.0, 5.11.0, 5.12.0, 5.13.0, 5.14.0, 5.15.0, 5.15.1, 5.15.2, 5.16.0, 5.17.0, 5.17.1, 5.18.0, 5.19.0, 5.19.1, 5.20.0, 5.21.0, 5.21.1, 5.21.2, 5.22.0, 5.22.1, 5.23.0, 5.24.0, 5.25.0, 5.25.1, 5.25.2, 5.25.3, 5.25.4, 5.26.0, 5.26.1, 5.26.2, 5.26.3, 5.26.4, 5.26.5, 5.27.0, 5.27.1, 5.27.2, 5.28.0, 5.28.1, 5.28.2, 5.28.3, 5.28.4
```

After: 

```
% npx tav --dry-run --verbose | rg match
-- 7 package versions matching filter: 6.0.0, 6.1.0, 6.4.0, 6.6.1, 6.8.0, 6.10.2, 6.11.1
-- 7 package versions matching filter: 5.0.0, 5.7.0, 5.15.0, 5.21.1, 5.26.0, 5.28.2, 5.28.4
```

trentm:
There is a test failure in the TAV tests for earlier 5.x versions. This test case:

```js
    it('should create valid spans for different request methods', async function () {
```

(Note: I isolated just this test case for my debugging, because I think there might be some pollution between tests... or perhaps just a timing issue because the issue is with an async socket close event.)

If you change that to `it.only` to run just that test, here is a failure:

```
% npm install --no-save undici@5.11.0
...

% npm test
...
  UndiciInstrumentation `undici` tests
    enable()
      1) should create valid spans for different request methods

  0 passing (142ms)
  1 failing

  1) UndiciInstrumentation `undici` tests
       enable()
         should create valid spans for different request methods:
     Uncaught SocketError: closed
      at Socket.onSocketClose (node_modules/undici/lib/client.js:967:31)
      at Socket.emit (node:events:517:28)
      at TCP.<anonymous> (node:net:350:12)
      at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
...
```

This passes with undici@5.12.0.

Here is a small test script showing the behaviour difference:

```
% cat use-undici-with-weird-method.js
const undici = require('undici')
const anActualServer = 'http://127.0.0.1:8200/'
async function main() {
  let res
  try {
    res = await undici.request(anActualServer, {
      method: 'get'
    })
  } catch (reqErr) {
    console.log('reqErr: ', reqErr);
  }
  console.log('res:', res.statusCode, res.headers);
  let body
  try {
    body = await res.body.text()
  } catch (bodyErr) {
    console.log('bodyErr: ', bodyErr);
  }
  console.log('body:', body);
}
main()
```

The failing case (undici@5.11.0):

```
% node use-undici-with-weird-method.js
res: 400 { connection: 'close' }
bodyErr:  SocketError: closed
    at Socket.onSocketClose (/Users/trentm/tm/opentelemetry-js-contrib12/plugins/node/instrumentation-undici/node_modules/undici/lib/client.js:967:31)
    at Socket.emit (node:events:517:28)
    at TCP.<anonymous> (node:net:350:12) {
  code: 'UND_ERR_SOCKET',
  socket: {
    localAddress: undefined,
    localPort: undefined,
    remoteAddress: undefined,
    remotePort: undefined,
    remoteFamily: undefined,
    timeout: undefined,
    bytesWritten: 64,
    bytesRead: 47
  }
}
body: undefined
```

The passing test case (undici@5.12.0):

```
% npm install --no-save undici@5.12.0
...

% node use-undici-with-weird-method.js
res: 400 { connection: 'close' }
body:
```

The undici.test.ts code cannot trap this error event with its current try/catch.


I stopped digging in here. We are talking really old undici versions:

```
  '5.11.0': '2022-10-03T15:42:59.406Z',
  '5.12.0': '2022-10-27T15:01:02.770Z',
```

I suggest we just limit support to >=5.12.0.
Thoughts?

trentm:
Options for this test failure:

1. Make this instr only work for undici 5.12.0 and later.
2. Skip this test for less than undici 5.11.0.
3. Someone dig into the test case.

I am inclined to do #2.

trentm:
^^ that commit does option 2

trentm:
Here is a passing TAV run:

```
% npm run test-all-versions -- -q

> @opentelemetry/instrumentation-undici@0.1.0 test-all-versions
> tav -q

-- required packages ["undici@6.11.1"]
-- installing ["undici@6.11.1"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@6.10.2"]
-- installing ["undici@6.10.2"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@6.8.0"]
-- installing ["undici@6.8.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@6.6.1"]
-- installing ["undici@6.6.1"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@6.4.0"]
-- installing ["undici@6.4.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@6.1.0"]
-- installing ["undici@6.1.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@6.0.0"]
-- installing ["undici@6.0.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.28.4"]
-- installing ["undici@5.28.4"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.28.2"]
-- installing ["undici@5.28.2"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.26.0"]
-- installing ["undici@5.26.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.21.1"]
-- installing ["undici@5.21.1"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.15.0"]
-- installing ["undici@5.15.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.7.0"]
-- installing ["undici@5.7.0"]
-- running test "npm run test" with undici (env: {})
-- required packages ["undici@5.0.0"]
-- installing ["undici@5.0.0"]
-- running test "npm run test" with undici (env: {})
-- ok
```

plugins/node/instrumentation-undici/test/undici.test.ts
```diff
@@ -207,7 +210,13 @@ describe('UndiciInstrumentation `undici` tests', function () {
       assert.ok(spans.length === 0, 'ignoreRequestHook is filtering requests');
     });
 
-    it('should create valid spans for different request methods', async function () {
+    it.only('should create valid spans for different request methods', async function () {
```
david-luna, line 213: I guess this is a leftover. This `.only` will skip the rest of the tests always

david-luna:
@trentm there is code to catch this case and skip the test. I think I've made the check too specific for the error message and this could change between versions of `undici` and/or `node`

the intent was to skip only for the versions that throw. I'm not 100% sure if versions prior to v5.11.0 work but the might

the comment: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/fe18e2fbb2a6535cb72f314fdb1550a3a4160403/plugins/node/instrumentation-undici/test/undici.test.ts#L220

the check: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/fe18e2fbb2a6535cb72f314fdb1550a3a4160403/plugins/node/instrumentation-undici/test/undici.test.ts#L244

We could relax that check to skip on socket error since is an expected exception from the lib.

david-luna:
> I suggest we just limit support to >=5.12.0.

I think is a good idea

plugins/node/instrumentation-undici/test/undici.test.ts
```diff
@@ -207,7 +210,13 @@ describe('UndiciInstrumentation `undici` tests', function () {
...
+      if (semver.lt(UNDICI_VERSION, '5.12.0')) {
```
david-luna, line 214: if we drop support for `<5.12.0` this won't be necessary. Neither the try/catch of this test

david-luna:
there is also another failure in TAV testing. Test fails for version 6.11.0 where the headers changed format but request does not have `addHeader` API. I guess this was considered a bug since a new version bas released shortly

```bash
npm info undici time

{
  ...
  '6.11.0': '2024-04-02T10:40:33.293Z',
  '5.28.4': '2024-04-02T16:36:40.050Z',
  '6.11.1': '2024-04-02T16:44:56.065Z'
}
```

so in the test we got the following error
```
 1) UndiciInstrumentation `undici` tests
       enable()
         should create valid spans for "dispatch" method:
     Uncaught TypeError: request.addHeader is not a function
      at UndiciInstrumentation.onRequestCreated (src/undici.ts:77:4)
      at Channel.publish (node:diagnostics_channel:56:9)
      at new Request (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/undici/lib/core/request.js:192:23)
      at Client.[dispatch] (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/undici/lib/dispatcher/client.js:304:21)
      at Intercept (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/undici/lib/interceptor/redirect-interceptor.js:11:16)
      at Client.[Intercepted Dispatch] (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/undici/lib/dispatcher/dispatcher-base.js:158:12)
      at Client.dispatch (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/undici/lib/dispatcher/dispatcher-base.js:179:40)
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/plugins/node/instrumentation-undici/test/undici.test.ts:502:16
      at new Promise (<anonymous>)
      at Context.<anonymous> (test/undici.test.ts:501:13)
      at processImmediate (node:internal/timers:476:21)
      at process.callbackTrampoline (node:internal/async_hooks:130:17)
```
In that case the headers are placed in an array of strings so we can check and push the headers there. This is the change

```diff
diff --git a/plugins/node/instrumentation-undici/src/undici.ts b/plugins/node/instrumentation-undici/src/undici.ts
index 262bea42..e69015b7 100644
--- a/plugins/node/instrumentation-undici/src/undici.ts
+++ b/plugins/node/instrumentation-undici/src/undici.ts
@@ -270,10 +270,12 @@ export class UndiciInstrumentation extends InstrumentationBase {
     for (let i = 0; i < headerEntries.length; i++) {
       const [k, v] = headerEntries[i];

-      if (typeof request.headers === 'string') {
-        request.headers += `${k}: ${v}\r\n`;
-      } else {
+      if (typeof request.addHeader === 'function') {
         request.addHeader(k, v);
+      } else if (Array.isArray(request.headers)) {
+        request.headers.push(k, v);
+      } else if (typeof request.headers === 'string') {
+        request.headers += `${k}: ${v}\r\n`;
       }
     }
     this._recordFromReq.set(request, { span, attributes, startTime });
```

Conditions now are more strict and if none match instrumentation will assume and do nothing to avoid runtime errors.

plugins/node/instrumentation-undici/.tav.yml
```diff
@@ -1,8 +1,12 @@
 undici:
   jobs:
-    - versions: ">=5 <6"
+    - versions:
```
david-luna, line 3: I think is a good idea to reduce the number of versions

trentm:
> We could relax that check to skip on socket error since is an expected exception from the lib.

If I'm understanding the cases that fail for me, I don't think the try/catch will catch this exception. It currently results in an uncaughtException. It might be an "error" event on one of the streams. I'm not sure.



> > I suggest we just limit support to >=5.12.0.
> 
> I think is a good idea

That is what I called Option 1.   I thought it was perhaps  a shame to drop support for all of `[5.0.0, 5.12.0)` just because this one test of an edge case wasn't passing our tests.  However, I think either way is fine. No one should  be using such an old undici.

> if we drop support for `<5.12.0` this won't be necessary. Neither the try/catch of this test

I'll give that a try.

trentm:
[Re: the "request.addHeader" failure]

> I guess this was considered a bug since a new version bas released shortly

Yes. Discussed in this issue: https://github.com/nodejs/undici/issues/3043

plugins/node/instrumentation-undici/test/undici.test.ts
trentm: Thanks. My bad. commit 2c3ce0ce

david-luna (approved):

plugins/node/instrumentation-undici/README.md
```diff
@@ -46,11 +50,11 @@ Undici instrumentation has few options available to choose from. You can set the
...
-| [`startSpanHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#67) | `StartSpanHookFunction` | Function for adding custom attributes before a span is started. |
-| [`requireParentforSpans`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#69) | `Boolean` | Require a parent span is present to create new span for outgoing requests. |
-| [`headersToSpanAttributes`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#71) | `Object` |  List of case insensitive HTTP headers to convert to span attributes. Headers will be converted to span attributes in the form of `http.{request\|response}.header.header-name` where the name is only lowercased, e.g. `http.response.header.content-length`|
+| [`ignoreRequestHook`](./src/types.ts#63) | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function. |
```
pichlermarc, line 53: Have not verified but I think this won't work for npmjs.com (It's published there)

pichlermarc (approved):

plugins/node/instrumentation-undici/README.md
trentm: Ah, phooey. You are correct.

Another example are the links in the "Status" section here:
- on GH: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-dns/README.md#status
- on npmjs.com: https://www.npmjs.com/package/@opentelemetry/instrumentation-dns

npmjs.com *is* attempting to rewrite those links, but doesn't get the right for GH.

I'll revert this part of the change.

trentm: commit 705b83d0

PR #2086, opened 2024-04-05
refactor(instr-amqplib): use exported strings for attributes

david-luna:

- Updates #2025 

- Replace `SemanticAttributes.*` with exported strings `SEMATTRS_*`
- Update README with new semantic convention package version and key

### DIFF
--- a/plugins/node/instrumentation-amqplib/src/utils.ts
+++ b/plugins/node/instrumentation-amqplib/src/utils.ts
@@ -19,10 +19,17 @@ import {
   diag,
   HrTime,
   Span,
-  SpanAttributes,
-  SpanAttributeValue,
+  Attributes,
+  AttributeValue,
 } from '@opentelemetry/api';
-import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
+import {
+  SEMATTRS_MESSAGING_PROTOCOL,
+  SEMATTRS_MESSAGING_PROTOCOL_VERSION,
+  SEMATTRS_MESSAGING_SYSTEM,
+  SEMATTRS_MESSAGING_URL,
+  SEMATTRS_NET_PEER_NAME,
+  SEMATTRS_NET_PEER_PORT,
+} from '@opentelemetry/semantic-conventions';
 import type * as amqp from 'amqplib';
 
 export const MESSAGE_STORED_SPAN: unique symbol = Symbol(
@@ -41,9 +48,9 @@ export const CONNECTION_ATTRIBUTES: unique symbol = Symbol(
 export type InstrumentationPublishChannel = (
   | amqp.Channel
   | amqp.ConfirmChannel
-) & { connection: { [CONNECTION_ATTRIBUTES]: SpanAttributes } };
+) & { connection: { [CONNECTION_ATTRIBUTES]: Attributes } };
 export type InstrumentationConsumeChannel = amqp.Channel & {
-  connection: { [CONNECTION_ATTRIBUTES]: SpanAttributes };
+  connection: { [CONNECTION_ATTRIBUTES]: Attributes };
   [CHANNEL_SPANS_NOT_ENDED]?: {
     msg: amqp.ConsumeMessage;
     timeOfConsume: HrTime;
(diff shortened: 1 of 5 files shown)


maryliag (approved):

plugins/node/instrumentation-amqplib/src/utils.ts
```diff
@@ -19,10 +19,17 @@ import {
   diag,
   HrTime,
   Span,
-  SpanAttributes,
-  SpanAttributeValue,
+  Attributes,
+  AttributeValue,
```
JamieDanielson, line 23: I think we want the Attributes and AttributeValue change targeting Next branch, separate from this PR. Looks like that change is [part of this issue](https://github.com/open-telemetry/opentelemetry-js/issues/4175).

JamieDanielson: Actually that looks to be specifically pointing out packages in core that have a lower API version, and there may not be a next branch on contrib. 🤔

JamieDanielson (commented):
Thanks! I left a comment about maybe doing the Attributes and AttributeValue change separate... but I'm not sure if it's necessary as I [look at it more](https://github.com/open-telemetry/opentelemetry-js/issues/4175). If anything that should maybe be another issue similar to this, where we have a breakdown of issues that need to have deprecated SpanAttributes and SpanAttributeValue replaced with the new Attributes and AttributeValue here in contrib 🤔  It's a breaking change if minimum API version needs to the change, but it's not necessary here so it wouldn't be breaking here.

JamieDanielson (approved):
This look good. I don't think the SpanAttribute change needs to be separate here. On the issue #1778 we discussed maybe not needing the Notes column with the key after all, but it's already here and I don't think it's necessary to do the extra work to remove it. If anything we can remove it when we update semantic conventions, to minimize overhead.

david-luna:
Thanks for the review @JamieDanielson :)

I guess the breaking change would happen when `SpanAttributes` and `SpanAttributeValue` objects are finally removed from `@opentelemetry/semantic-conventions`. All packages without the imports updated will fail to compile.

> Actually that looks to be specifically pointing out packages in core that have a lower API version, and there may not be a next branch on contrib. 🤔

I thought so and it makes sense to remove deprecated object from current instrumentations. I wonder which kind of version bump we should do when the instrumentations update the API/SDK dependencies to next branch

JamieDanielson:
> SpanAttributes and SpanAttributeValue objects are finally removed from @opentelemetry/semantic-conventions.
These are actually in the API package, not semantic convention, but either way I think we're good here. 

> I wonder which kind of version bump we should do when the instrumentations update the API/SDK dependencies to next branch
I wonder this too! I think we may have talked about it briefly during one of the sig meetings, but as we get closer to release, we will have to revisit.

plugins/node/instrumentation-amqplib/src/utils.ts
blumamir: I think that since the change was introduced in https://github.com/open-telemetry/opentelemetry-js-api/pull/142, and it was published with `@opentelemetry/api` version [1.1.0](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/CHANGELOG.md#110-2022-01-25). then it is safe to replace the types here, as this package already has a peer dependency on `"@opentelemetry/api": "^1.3.0"` which is guarentee to include this change

trentm (approved):

trentm:
@david-luna Should this one have also updated the instr-amqplib test files?

```
% rg SemanticAttributes
test/amqplib-promise.test.ts
40:  SemanticAttributes,
150:        publishSpan.attributes[SemanticAttributes.MESSAGING_SYSTEM]
153:        publishSpan.attributes[SemanticAttributes.MESSAGING_DESTINATION]
156:        publishSpan.attributes[SemanticAttributes.MESSAGING_DESTINATION_KIND]
...
```

david-luna:
Indeed. Good catch :) 

I'll create a new PR for it

Thanks!

PR #2087, opened 2024-04-05
perf(instrumentation-nestjs-core): extract reusable span attributes to outer scope

Samuron:

Reduce overhead of nestjs instrumentation

Pulled some parts of instrumentation to enclosing scope to reuse and reduce allocations


blumamir (approved):
LGTM, thank you for the contribution

Issue #1845, opened 2023-12-03
Can we please add the opentelemetry-instrumentation-kafkajs (https://www.npmjs.com/package/opentelemetry-instrumentation-kafkajs) to nod.js auto-instrumention, please?


biswajit-nanda:
### Is your instrumentation request related to a problem? Please describe
It is not related to a problem, just to address inconvenience while doing Node.js auto injection and auto-instrumentation via OpenTelemetry operator in k8s. As opentelemetry-instrumentation-kafkajs is not included in the auto-instrumentation library, the kafka calls cannot be instrumented while using k8s opentelemetry operator and the instrumentation has to be manually injected using a tracing.js file.

### Is it applicable for Node or Browser or both?
It is mostly for node.

### Do you expect this instrumentation to be commonly used?
Yes, a lot.

### What version of instrumentation are you interested in using?
Versions: 0.40.0

PR #2089, opened 2024-04-08
feat: kafkajs instrumentation

seemk:

Prerequisite for https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1845

Adds the [kafkajs instrumentation](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-kafkajs) created by Aspecto which does not seem to be maintained any longer.

The codebase has been converted to TypeScript and differs from the original how the propagation headers are read (in this case they are read case insensitively, this fixes propagation issues across language SDKs). Added a NOTICE file attributing the original work and describing the changes.


plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,452 @@
...
+import { bufferTextMapGetter } from './propagator';
+import {
+  InstrumentationBase,
+  InstrumentationModuleDefinition,
+  InstrumentationNodeModuleDefinition,
+  safeExecuteInTheMiddle,
+  isWrapped,
+} from '@opentelemetry/instrumentation';
+
+export class KafkaJsInstrumentation extends InstrumentationBase<
```
Flarna, line 60: Is it possible to hide the kafkajs types from public API?
see https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,452 @@
...
+  constructor(config: KafkaJsInstrumentationConfig = {}) {
+    super(
+      '@opentelemetry/instrumentation-kafkajs',
+      VERSION,
+      Object.assign({}, config)
+    );
+  }
+
+  override setConfig(config: KafkaJsInstrumentationConfig = {}) {
+    this._config = Object.assign({}, config);
```
Flarna, line 75: is this needed? config is copied and stored in base class

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,452 @@
...
+  typeof kafkaJs
+> {
+  protected override _config!: KafkaJsInstrumentationConfig;
+  private moduleVersion?: string;
+
+  constructor(config: KafkaJsInstrumentationConfig = {}) {
+    super(
+      '@opentelemetry/instrumentation-kafkajs',
+      VERSION,
+      Object.assign({}, config)
```
Flarna, line 70: is this needed? I think defaulting/copy is done in base class

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,452 @@
...
+    operation: string,
+    context: Context | undefined,
+    link?: Link
+  ) {
+    const span = this.tracer.startSpan(
+      topic,
+      {
+        kind: SpanKind.CONSUMER,
+        attributes: {
+          [SEMATTRS_MESSAGING_SYSTEM]: 'kafka',
```
Flarna, line 398: Could we use constants from SemConv here?

plugins/node/instrumentation-kafkajs/src/propagator.ts
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright The OpenTelemetry Authors
```
pichlermarc, line 2: We need to add a Copyright line to keep Aspecto's original license header here and in all other files to comply with the license.

pichlermarc (commented):
Please add some owners to the `.github/component_owners.yaml`. 
Ideally that's two people so that PRs opened by one can be reviewed by the other.

plugins/node/instrumentation-kafkajs/src/types.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Span } from '@opentelemetry/api';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+import type { Message } from 'kafkajs';
```
blumamir, line 18: we should verify that these types do not leak into the public API where it might import types from `kafkajs` and fail if the package is not found in end user node_modules.

blumamir:
Thank you @seemk

plugins/node/instrumentation-kafkajs/src/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface KafkaProducerCustomAttributeFunction {
+  (span: Span, topic: string, message: Message): void;
+}
+
+export interface KafkaConsumerCustomAttributeFunction {
+  (span: Span, topic: string, message: Message): void;
+}
```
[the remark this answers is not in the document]
david-luna: @blumamir what do you think about my comment on using generics and having the `Info` as a default?

plugins/node/instrumentation-kafkajs/test/kafkajs.test.ts
```diff
@@ -0,0 +1,941 @@
...
+              partition: 0,
+              errorCode: 123,
+              offset: '18',
+              timestamp: '123456',
+            },
+          ];
+        });
+        instrumentation.disable();
+        instrumentation.enable();
+        producer = kafka.producer();
```
david-luna, line 146: Producer here does not send messages to any Kafka instance but rather keeps them in a memory. There are other instrumentations (ioredis, mongodb, mysql, ...) that use docker to get closer to a real use case scenario. They are slower but IMO there is value on it. What do you think?

cc: @pichlermarc @blumamir

david-luna (changes requested):

plugins/node/instrumentation-kafkajs/src/types.ts
blumamir: I support this, will give us greater flexibility and robustness over time 👍

plugins/node/instrumentation-kafkajs/src/types.ts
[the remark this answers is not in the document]
blumamir: I think it's a good idea and we should apply it in all instrumentations, especially new ones.

Regardless of the type of `Message` object itself, there are also additional info that is captured in this hook: the topic name which is not part of the `message` and perhaps additional data that will be helpful for users in the future.

The motivation for the `Info` object was that we don't end up with hooks with 5-6 arguments that all have to be written even if just one is to be used.
For example, in some instrumentations the hook can also capture the instrumented package versions (e.g. `kafkajs`), like [`amqplib`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/1af49014b8a77d34058750ce2a87bd58211ebe70/plugins/node/instrumentation-amqplib/src/types.ts#L33). If this would be added later here, and possibly more arguments in the future that users might need, the hook signature can grow over time and become harder to manage as the number of arguments grows. Doing this refactor for existing instrumentations might be a breaking change for users, thus it's a good chance to introduce this pattern here if we think it is appropriate.

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,452 @@
...
+
+  override setConfig(config: KafkaJsInstrumentationConfig = {}) {
+    this._config = Object.assign({}, config);
+  }
+
+  protected init(): InstrumentationModuleDefinition<typeof kafkaJs> {
+    const unpatch: InstrumentationModuleDefinition<
+      typeof kafkaJs
+    >['unpatch'] = moduleExports => {
+      diag.debug('Removing patch for kafkajs');
```
blumamir, line 82: It is preferable to use `this._diag` which is printed with the instrumentation library name (here and in few other places). might be a good chance to align it for this instrumentation

plugins/node/instrumentation-kafkajs/test/kafkajs.test.ts
blumamir: I think writing the tests with a real kafka broker instance can improve the tests and catch more issues, but it is also a lot of work. If there is motivation for someone to work on it, I support and encourage. If the task is too time-consuming or complicated, I am also good with keeping the tests as they are now.

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
seemk: I saw that `SemanticAttributes` has a deprecation notice, so I converted to these constants. Or did you have something else in mind?

plugins/node/instrumentation-kafkajs/test/kafkajs.test.ts
david-luna: let's keep it put of the scope of this PR :)

I'll try to allocate som time for it and create a PR for that. Thanks @blumamir

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
trentm: I'm confused. @Flarna The code has:

```js
import {
  MESSAGINGDESTINATIONKINDVALUES_TOPIC,
  MESSAGINGOPERATIONVALUES_PROCESS,
  MESSAGINGOPERATIONVALUES_RECEIVE,
  SEMATTRS_MESSAGING_SYSTEM,
  SEMATTRS_MESSAGING_DESTINATION_KIND,
  SEMATTRS_MESSAGING_DESTINATION,
  SEMATTRS_MESSAGING_OPERATION,
```

above, so constants from semconv *are* being used here, right?

Flarna: In [SemConv](https://github.com/open-telemetry/semantic-conventions/blob/v1.25.0/docs/messaging/messaging-spans.md) there are values defined for well known messaging systems like `kafka`. I was assuming there are generated constants for it.
So it's not about the keys it's about value `kafka`.

trentm: Okay, understood.

Given that the JS `@opentelemetry/semantic-conventions` package is way back on v1.7.0, semconv did not define well-known messaging systems at that version. IIUC, a set of well-known messaging systems was only added about 5mo ago in https://github.com/open-telemetry/semantic-conventions/pull/517 -- part of semconv v1.24.0 (2023-12-15).

The Java semconv package has this class for that enum:
https://github.com/open-telemetry/semantic-conventions-java/blob/7da24068eea69dff11a78d59750b115dc4c5854d/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/MessagingIncubatingAttributes.java#L369-L401
but AFAICT there is no Java code in https://github.com/open-telemetry/opentelemetry-java-instrumentation using them yet.

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
seemk: Thanks, fixed it

plugins/node/instrumentation-kafkajs/src/types.ts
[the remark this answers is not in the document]
seemk: I've refactored it so the function takes in an info object (with the type defaulting to `KafkajsMessage`)

plugins/node/instrumentation-kafkajs/src/types.ts
seemk: Combined the generic version with the unified `info` argument

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
seemk: `kafka` is not in the semconvs package yet, or am I missing something?

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
seemk: Removed it

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
seemk: Removed

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,452 @@
...
+                  payload.batch.topic,
+                  message,
+                  MESSAGINGOPERATIONVALUES_PROCESS,
+                  undefined,
+                  origSpanLink
+                );
+              }
+            );
+            const batchMessagePromise: Promise<void> = original!.apply(
+              this,
```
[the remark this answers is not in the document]
seemk: Isn't this inherited from the parent scope (in _this_ case the `this` argument of `function eachBatch(this: unknown, ...args)`)?

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
trentm: @seemk I don't think you are missing anything, no. `kafka` isn't defined in the current `@opentelemetry/semantic-conventions` package as a string value.

Flarna: Correct, it's not included. I was assuming it is there because it is in spec.
Completely forgot out the fact that the version used here is several years old where `kafka` as not yet part of the spec.

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
seemk: Removed the internal type from exports

plugins/node/instrumentation-kafkajs/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "test": "ts-mocha --require @opentelemetry/contrib-test-utils -p tsconfig.json 'test/**/*.test.ts'",
+    "tdd": "npm run test -- --watch-extensions ts --watch",
+    "clean": "rimraf build/*",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "eslint . --ext .ts --fix",
+    "precompile": "tsc --version && lerna run version:update --scope @opentelemetry/instrumentation-kafkajs --include-dependencies",
+    "prewatch": "npm run precompile",
+    "prepublishOnly": "npm run compile",
+    "version:update": "node ../../../scripts/version-update.js",
+    "compile": "npm run version:update && tsc -p ."
```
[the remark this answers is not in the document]
seemk: Removed it

plugins/node/instrumentation-kafkajs/src/propagator.ts
seemk: Added Aspecto to the copyright line (and modified the eslint pattern)

david-luna:
@seemk 

looks very good now :)

please fill the table in the readme as requested in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2089/files#r1562439567 and you have my 👍

blumamir (commented):
Thanks again @seemk 

Added a few more minor comments.

blumamir:
Another thought, should this new instrumentation be added to "@opentelemetry/auto-instrumentations-node" as part of this PR, or should it be done separately in a followup PR?

plugins/node/instrumentation-kafkajs/src/propagator.ts
```diff
@@ -0,0 +1,43 @@
...
+export const bufferTextMapGetter: TextMapGetter = {
+  get(carrier, key) {
+    if (!carrier) {
+      return undefined;
+    }
+
+    const keys = Object.keys(carrier);
+
+    for (const carrierKey of keys) {
+      if (carrierKey.toLowerCase() === key) {
```
Flarna, line 32: Is `toLowerCase()` really needed here?
W3C tracecontext spec is quite clear to inject in lowercase so it seems redundant.
What about propagators which want to use mixed case keys? They wont work I guess.

seemk: [Some propagators](https://github.com/open-telemetry/opentelemetry-java/blob/51ff6bcf335a488ff04b49514f9360f59f27677a/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation/B3Propagator.java#L48-L50) inject in mixed case and [others](https://github.com/open-telemetry/opentelemetry-js/blob/d99e211a311ceefdfed67c501ea99129f5488434/packages/opentelemetry-propagator-b3/src/constants.ts) expect it in lowercase.

There's an unresolved issue for it: https://github.com/open-telemetry/opentelemetry-js/issues/3150

Flarna: ok, seems like all options are wrong :o)

What about trying both like [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/eb6e8ef6b4c72710a98ed3590392b85c8c3b79e5/propagators/opentelemetry-propagator-instana/src/InstanaPropagator.ts#L47-L51)?

plugins/node/instrumentation-kafkajs/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+You can set the following:
+
+| Options                      | Type                                   | Description                                                                                                          |
+| ---------------------------- | -------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
+| `producerHook`               | `KafkaProducerCustomAttributeFunction` | Function called before a producer message is sent. Allows for adding custom attributes to the span.                  |
+| `consumerHook`               | `KafkaConsumerCustomAttributeFunction` | Function called before a consumer message is processed. Allows for adding custom attributes to the span.             |
+| `moduleVersionAttributeName` | `string`                               | When set, the kafkajs module version will be added to span attributes under the key of `moduleVersionAttributeName`. |
+
+## Useful links
```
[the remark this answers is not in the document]
seemk: Added the attributes, however I noticed that with the current schema, `messaging.destination.kind` (previously `messaging.destination_kind`) is [removed completely](https://github.com/open-telemetry/opentelemetry-specification/pull/3214). So I wonder whether we should drop this attribute from the instrumentation to avoid breaking changes in the future.

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,416 @@
...
+        this._unwrap(moduleExports.Kafka.prototype, 'producer');
+      }
+      if (isWrapped(moduleExports?.Kafka?.prototype.consumer)) {
+        this._unwrap(moduleExports.Kafka.prototype, 'consumer');
+      }
+    };
+
+    const module = new InstrumentationNodeModuleDefinition(
+      'kafkajs',
+      ['*'],
```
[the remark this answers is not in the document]
seemk: Fixed

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,416 @@
...
+      (moduleExports: typeof kafkaJs) => {
+        this._diag.debug('Applying patch for kafkajs');
```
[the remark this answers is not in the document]
seemk: Removed them, thanks!

plugins/node/instrumentation-kafkajs/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "nyc": "15.1.0",
+    "rimraf": "5.0.5",
+    "sinon": "15.2.0",
+    "ts-mocha": "10.0.0",
+    "typescript": "4.4.4"
+  },
+  "dependencies": {
+    "@opentelemetry/instrumentation": "^0.50.0",
+    "@opentelemetry/semantic-conventions": "^1.23.0"
+  },
```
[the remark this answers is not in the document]
seemk: Updated

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -0,0 +1,416 @@
...
+  protected init() {
+    const unpatch = (moduleExports: typeof kafkaJs) => {
+      this._diag.debug('Removing patch for kafkajs');
+      if (isWrapped(moduleExports?.Kafka?.prototype.producer)) {
+        this._unwrap(moduleExports.Kafka.prototype, 'producer');
+      }
+      if (isWrapped(moduleExports?.Kafka?.prototype.consumer)) {
+        this._unwrap(moduleExports.Kafka.prototype, 'consumer');
+      }
+    };
```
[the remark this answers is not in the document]
seemk: `unpatch` is used at the beginning of `patch` function, think making `unpatch` a member function would help?

seemk:
> Please add some owners to the `.github/component_owners.yaml`. Ideally that's two people so that PRs opened by one can be reviewed by the other.

Added myself, any ideas for the other one?

plugins/node/instrumentation-kafkajs/README.md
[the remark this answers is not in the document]
seemk: Removed the attribute completely

Flarna (approved):

seemk:
> Another thought, should this new instrumentation be added to "@opentelemetry/auto-instrumentations-node" as part of this PR, or should it be done separately in a followup PR?

Will do it in a followup

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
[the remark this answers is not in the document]
david-luna: you're right as per MDN docs at the end of section https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#cannot_be_used_as_methods says

> For similar reasons, the [call()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call), [apply()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply), and [bind()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) methods are not useful when called on arrow functions, **because arrow functions establish `this` based on the scope the arrow function is defined within, and the `this` value does not change based on how the function is invoked.**

david-luna (approved):

blumamir (approved):

PR #2091, opened 2024-04-11
chore: experimental 0.51.0, remove instrumentations generic type to align with upstream

blumamir:
This PR applied the changes in https://github.com/open-telemetry/opentelemetry-js/pull/4598 to the contrib repo.

Once we merge the upstream change in code and release a core version, this PR can be used to apply the removal of the generic type from `InstrumentationBase`, `InstrumentationNodeModuleDefinition` and `InstrumentationNodeModuleFile`.

It was tested to `npm run compile` against the up-to-date `@opentelemetry/instrumentation` package with `npm link` locally.

BEGIN_COMMIT_OVERRIDE
feat(deps): update otel-js to 0.51.0
feat: remove generic type from instrumentations
END_COMMIT_OVERRIDE


JamieDanielson (commented):
Thanks for preparing this PR and making all these changes! This will make it much easier to release quickly after the updated instrumentation package is available to use.

I was also able to test this locally and everything tested fine, except the mongoose test which timed out that I believe is unrelated.

One question I have is around the instrumentations that did use a type other than `any`. You mentioned in the [instrumentation PR](https://github.com/open-telemetry/opentelemetry-js/pull/4598) that some instrumentation packages are using types packages that may not always be up to date... but I also think that if they currently have those, perhaps this PR should keep using them and we can look at a follow-up to remove them. For example, should cucumber have `(moduleExports: Cucumber, moduleVersion)`, express have `(moduleExports: typeof express, moduleVersion)`, etc? Or do we think those types were only used in the `InstrumentationNodeModuleDefinition` because it was necessary to type _something_ and people sort of forced it in there? I see you [added it for http instrumentation](https://github.com/open-telemetry/opentelemetry-js/pull/4598/files#diff-08310811c8e3d80f91f1a3370cc56ca1e560af4f087a608321fc884eb8498ff1R118), which makes me think the same should be done here for the packages using something other than `any` type. What do you think about this?

plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
```diff
@@ -65,10 +65,10 @@ export class FastifyInstrumentation extends InstrumentationBase {
 
   init() {
     return [
-      new InstrumentationNodeModuleDefinition<any>(
+      new InstrumentationNodeModuleDefinition(
         'fastify',
         ['^3.0.0', '^4.0.0'],
-        (moduleExports, moduleVersion) => {
+        (moduleExports: any, moduleVersion) => {
```
JamieDanielson, line 71: I notice this is here but not anywhere else. Is that intentional? Do we want to explicitly type out any for the moduleExports?

blumamir: good catch, not sure how this got here. removed 👍

plugins/node/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -73,8 +73,7 @@ export class DataloaderInstrumentation extends InstrumentationBase {
             this._unwrap(dataloader.prototype, 'loadMany');
           }
         }
-        // cast it to module definition of unknown type to avoid exposing Dataloader types on public APIs
-      ) as InstrumentationNodeModuleDefinition<unknown>,
```
blumamir, line 77: Since this comment is here, I think it's not safe to type the `moduleExports` with the now-removed generic type, which does not give a lot of value IMO.

plugins/node/opentelemetry-instrumentation-connect/src/instrumentation.ts
```diff
@@ -52,7 +52,7 @@ export class ConnectInstrumentation extends InstrumentationBase<Server> {
 
   init() {
     return [
-      new InstrumentationNodeModuleDefinition<any>(
```
blumamir, line 55: This is already typed as `any` so the `moduleExports` type is not typed at the moment. Can be improved in a future PR. want to keep this one minimal to the relevant changes

plugins/node/opentelemetry-instrumentation-dns/src/instrumentation.ts
```diff
@@ -27,26 +27,24 @@ import { DnsInstrumentationConfig } from './types';
 import * as utils from './utils';
 import { VERSION } from './version';
 import {
-  Dns,
-  DnsPromises,
```
blumamir, line 31: these imports invalidated the `internal-types.ts` guarntee, as they were exported in the public API from an internal file

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -95,10 +95,8 @@ export class GraphQLInstrumentation extends InstrumentationBase {
     return module;
   }
 
-  private _addPatchingExecute(): InstrumentationNodeModuleFile<
-    typeof graphqlTypes
```
blumamir, line 99: here the patch function argument is already typed as `any` so the generic type had no effect. trying to type `moduleExports` yields a ts error

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -71,7 +69,7 @@ export class KnexInstrumentation extends InstrumentationBase<any> {
   }
 
   private getRunnerNodeModuleFileInstrumentation(basePath: string) {
-    return new InstrumentationNodeModuleFile<typeof knex>(
```
blumamir, line 74: The patch function is already typed with `any`, so this has no effect

plugins/node/opentelemetry-instrumentation-koa/src/instrumentation.ts
```diff
@@ -52,7 +52,7 @@ export class KoaInstrumentation extends InstrumentationBase<typeof koa> {
   }
 
   protected init() {
-    return new InstrumentationNodeModuleDefinition<typeof koa>(
```
blumamir, line 55: the generic has no effect, `moudleExports` is typed with `any`

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -91,51 +91,51 @@ export class MongoDBInstrumentation extends InstrumentationBase {
...
 
     return [
-      new InstrumentationNodeModuleDefinition<any>(
+      new InstrumentationNodeModuleDefinition(
         'mongodb',
         ['>=3.3 <4'],
         undefined,
         undefined,
         [
-          new InstrumentationNodeModuleFile<WireProtocolInternal>(
```
blumamir, line 100: The patch function, `v3PatchConnection` is already typed correctly

plugins/node/opentelemetry-instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -91,51 +91,51 @@ export class MongoDBInstrumentation extends InstrumentationBase {
...
         ]
       ),
-      new InstrumentationNodeModuleDefinition<any>(
+      new InstrumentationNodeModuleDefinition(
         'mongodb',
         ['4.*', '5.*', '6.*'],
         undefined,
         undefined,
         [
-          new InstrumentationNodeModuleFile<V4Connection>(
```
blumamir, line 114: The patch function, `v4PatchConnectionCallback` is already typed as `any`

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -52,7 +52,7 @@ export class PgInstrumentation extends InstrumentationBase {
   }
 
   protected init() {
-    const modulePG = new InstrumentationNodeModuleDefinition<typeof pgTypes>(
```
blumamir, line 55: `moudle` below is already `any` this type has no effect

blumamir:
Thanks @JamieDanielson 

did another round of fixes thanks to your great comment, added back the types to cucumber, tedious, dns, graphql, hapi, Memcached, mysql, net, winston. If you spot anyone I missed please comment.

I also added comments in the PR documenting few places where the generic type had no effect because the `moduleExports` argument type was override with `any`

My intention in this PR is to keep everything in the same state it is now with types, and only introduce the changes from core. I support a followup work for deciding on guidelines on how the types should be used, and then align all instrumentations according to some common rules. I guess we still need some freedom to define as `any` if the type is too complex or impossible to acquire, and also allow typing the argument where it makes sense and is accessible with minimum issues.

blumamir:
The browsers test currently fails, seems unrelated to this PR.
@pichlermarc  mentioned it might be related to https://github.com/open-telemetry/opentelemetry-js/pull/4486

pichlermarc:
@blumamir I had a look and what's happening here is actually related to https://github.com/open-telemetry/opentelemetry-js/pull/4486. In fact, it's that change behaving as expected. :slightly_smiling_face: 

We're dropping events that happen at `0` now as that suggests falsely suggests that they happen out of order, but what's actually happening is that we don't have any data on them. See `entriesFallback`: there's `unloadEventStart` and `unloadEventEnd`, and they're both `0` and therefore they happen "before" `fetchStart` and they get dropped.

We'll need to adapt both tests to not expect `unloadEventStart` and `unloadEventEnd` for the two failing tests and then we should be good to merge.

JamieDanielson (approved):
Thank you for working on this! The updates look great. Also I appreciated the notes on the types that weren't being applied as expected, it was helpful for reviewing.

We'll need to get the test fixed up in a separate PR and then we should be good to go here.

trentm:
> We'll need to adapt both tests to not expect `unloadEventStart` and `unloadEventEnd` for the two failing tests and then we should be good to merge.

I have a start at a PR for this. Happy for someone else to beat me to it, if you like, though.

trentm:
Please review https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2145 to fix the browser-test.

pichlermarc (approved):
Thank you for working on this @blumamir. :tada: 
Thank you @trentm for fixing the browser tests. :slightly_smiling_face:

PR #2101, opened 2024-04-15
feat(auto-instrumentation-node): add azure detector

maryliag:

- The Azure detectors were not being added to the list of possible detectors. All other cloud providers were possible values, so this commit adds the missing detectors for Azure.

- Adds dependency for azure detector and adds to the list of possible detector that can be added on the auto instrumentation node


trentm (approved):
LGTM

hectorhdzg (approved):
LGTM, just I think this auto-instrumentation-node package is a little too much in my opinion, you don't expect customers to be running in Alibaba cloud, AWS and Azure at the same time, I know it can be configurable but having everything in same package could be overwhelming or confusing for some users.

trentm:
> you don't expect customers to be running in Alibaba cloud, AWS and Azure at the same time

I believe the intent is that the user doesn't need to think about configuring detectors. Whatever cloud env they run their app in, the relevant resource attrs will be detected.

I think right now the cloud detectors are noisy on diag.warn if they don't find anything. I'd like that to change to have them be silent if they don't find anything. However, that's a discussion for a separate issue.

JamieDanielson (approved):

blumamir (approved):

PR #2107, opened 2024-04-17
chore: remove patch and unpatch diag from instrumentations

blumamir:
This PR implements https://github.com/open-telemetry/opentelemetry-js/pull/4641 to the contrib instrumentations, so once the core repo PR is merged, we can quickly rebase this PR and apply the changes in contrib so we can release them together.


trentm (approved):

plugins/node/instrumentation-cucumber/src/instrumentation.ts
```diff
@@ -72,11 +69,8 @@ export class CucumberInstrumentation extends InstrumentationBase {
           });
           return moduleExports;
         },
-        (moduleExports, moduleVersion) => {
+        moduleExports => {
```
JamieDanielson, line 72: There are a few instrumentations still that do not have the moduleVersion removed even though they are no longer being used. Is that intentional? If not, can we update those as well for consistency? Happy to help take on that extra bit if you need a hand, there's a lot here 😃  Also I realize we'll probably have some merge conflicts to deal with between this and #2091 depending on which one merges first.

JamieDanielson (commented):
I notice there are a few instrumentations not included here that I think are intentional - can you confirm I have this right? For example, mongoose has a `patchAggregateExec` that includes `this._diag.debug('patched mongoose Aggregate exec function');`. I think this will at the top level use the shared diag from the main instrumentation package for patching the module itself, and there was just nothing to remove from there because it didn't have a general patch log... and then it will still have additional diag logging for those specific types of patches when it comes up, which seems to be by design already within that instrumentation. Does that sound accurate? A similar situation seems to exist for fastify and a couple others, though there were main patch logs to remove here that you took care of.

I think mongoose will still be able to use the shared unpatch diag log, and koa is missing an update because its syntax is a bit different `api.diag.debug('Patching Koa');`

blumamir:
Thanks for the thorough review.

> can you confirm I have this right? For example, mongoose has a `patchAggregateExec` that includes `this._diag.debug('patched mongoose Aggregate exec function');`. I think this will at the top level use the shared diag from the main instrumentation package for patching the module itself, and there was just nothing to remove from there because it didn't have a general patch log... and then it will still have additional diag logging for those specific types of patches when it comes up, which seems to be by design already within that instrumentation. Does that sound accurate? A similar situation seems to exist for fastify and a couple others, though there were main patch logs to remove here that you took care of.

That is correct, the intention was to address only the patch and unpatch of the module (or module file) itself, and not the specific function patchs from the module exports which I intend to cleanup in a different PR.

Not sure how much value we get from reporting specific function patches like mongoose and hapi does. I think we should either:

1. clean this up from all existing instrumentations.
2. hoist it to the base instrumentation and apply to all instrumentations from the base class `_wrap` and `_unwrap` functions.
3. Understand when it is needed and when not, and align the existing prints to a documented practice.

blumamir:
> and koa is missing an update because its syntax is a bit different `api.diag.debug('Patching Koa');`

Thanks! I updated the koa and also few other places. I think now it is removed from all relevant places. please share if you spot one that I missed

blumamir:
> As a follow-up after this, I think we could get these instrumentations consistently using `this._diag.debug` for their remaining debug statements 🤔 Those could be done separately in small individual PRs though to limit the scope of this.

Absolutely, I intend to align everything to `this._diag` in a followup PR, which isn't always trivial as some diag useages are coming from `utils.ts` and are lacking the instrumentation `this`, but this will be handled in a different PR after some diag usages are removed

plugins/node/instrumentation-cucumber/src/instrumentation.ts
blumamir: Right! thanks for spotting it. I removed the `moduleVersion` everywhere it is not needed anymore. Hope I didn't miss somewhere

JamieDanielson:
> Not sure how much value we get from reporting specific function patches like mongoose and hapi does. I think we should either:
> 
> 1. clean this up from all existing instrumentations.
> 2. hoist it to the base instrumentation and apply to all instrumentations from the base class `_wrap` and `_unwrap` functions.
> 3. Understand when it is needed and when not, and align the existing prints to a documented practice.

With the function patches I suppose it may be helpful when debugging to see if something specific is not getting patched as expected, similar to the purpose here for the module itself, but it's maybe more helpful for some instrumentations vs others.. It is something to consider whether we can get a more consistent practice there too, but I don't see it being quite as impactful as this change for the whole module log, or for general usage of `this._diag.debug`.

JamieDanielson (commented):
Thanks for working on this! It's looking great. I'm eager to see these tests pass with the updated instrumentation package once it's available 🚀

trentm:
Compile error in instr-mongodb:

```
 > @opentelemetry/instrumentation-mongodb@0.42.0 compile
> tsc -p .

src/instrumentation.ts:18:3 - error TS6133: 'diag' is declared but its value is never read.

18   diag,
     ~~~~
```

trentm:
> Not sure how much value we get from reporting specific function patches like mongoose and hapi does. I think we should either:
> 
>     1. clean this up from all existing instrumentations.

Speaking from a little experience with instr-hapi, I think those extra `diag.debug` messages are not useful for debugging at all. +1 for doing option 1 here.

blumamir:
Thanks @JamieDanielson and @trentm I support addressing the individual diag prints as well, mainly I think we should remove them unless there is some non-standard patching logic which can aid debugging?

As this PR already introduces quite some changes, I want to address the remaining patch diag prints in a followup PR.

I rebased this branch from main after merging #2091 and fixed all the conflicts. Thanks again for helping iterate this and the great suggestions.

pichlermarc (approved):
Looks good.  :+1: 

I agree that we should address the function-specific logs in a follow-up and not increase the scope of this PR for now. Merging this will unblock the release :slightly_smiling_face:

PR #2129, opened 2024-04-19
chore(examples/web): use exported strings for attributes

maryliag:

- Part Of #2025 

On package `examples/web`:

### DIFF
--- a/examples/web/package.json
+++ b/examples/web/package.json
@@ -46,7 +46,7 @@
     "@opentelemetry/instrumentation-xml-http-request": "^0.39.1",
     "@opentelemetry/propagator-b3": "^1.13.0",
     "@opentelemetry/sdk-trace-web": "^1.13.0",
-    "@opentelemetry/semantic-conventions": "^1.13.0"
+    "@opentelemetry/semantic-conventions": "^1.23.0"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib#readme"
 }
--- a/examples/web/examples/document-load/index.js
+++ b/examples/web/examples/document-load/index.js
@@ -9,11 +9,11 @@ import { B3Propagator } from '@opentelemetry/propagator-b3';
 import { CompositePropagator, W3CTraceContextPropagator } from '@opentelemetry/core';
 import { registerInstrumentations } from '@opentelemetry/instrumentation';
 import { Resource } from '@opentelemetry/resources';
-import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+import { SEMRESATTRS_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
 
 const provider = new WebTracerProvider({
   resource: new Resource({
-    [SemanticResourceAttributes.SERVICE_NAME]: 'web-service-dl',
+    [SEMRESATTRS_SERVICE_NAME]: 'web-service-dl',
   }),
 });
 
--- a/examples/web/examples/meta/index.js
+++ b/examples/web/examples/meta/index.js
@@ -6,11 +6,11 @@ import { B3Propagator } from '@opentelemetry/propagator-b3';
 import { getWebAutoInstrumentations } from '@opentelemetry/auto-instrumentations-web';
 import { registerInstrumentations } from '@opentelemetry/instrumentation';
 import { Resource } from '@opentelemetry/resources';
-import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+import { SEMRESATTRS_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
 
 const providerWithZone = new WebTracerProvider({
   resource: new Resource({
-    [SemanticResourceAttributes.SERVICE_NAME]: 'web-service-meta',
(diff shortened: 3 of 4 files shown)


JamieDanielson (approved):

examples/web/package.json
```diff
@@ -46,7 +46,7 @@
     "@opentelemetry/instrumentation-xml-http-request": "^0.39.1",
     "@opentelemetry/propagator-b3": "^1.13.0",
     "@opentelemetry/sdk-trace-web": "^1.13.0",
-    "@opentelemetry/semantic-conventions": "^1.13.0"
+    "@opentelemetry/semantic-conventions": "^1.23.0"
```
trentm, line 49: This only *requires* v1.22.0, but I **think** we can have (perhaps even want) the examples all to be updated to be using the very latest versions of deps? Not sure how strict we need to be here. If we need to back this up to `^1.22.0` later, we can do another change.

trentm (approved):

examples/web/package.json
Flarna: This definitely brings the example in bad state. `"@opentelemetry/instrumentation-xml-http-request": "^0.39.1"` depends on `"@opentelemetry/semantic-conventions": "1.13.0"`.
So user ends up with a duplicate.

In general I think examples should be only touched if it is verified manually that result is fine as there are no automated tests.

trentm: Does having two versions of `@opentelemetry/semantic-conventions` necessarily mean a bad state for the example? 

The current state already has three different versions of `@opentelemetry/resource`, FWIW:

```
[23:37:57 trentm@peach:~/src/opentelemetry-js-contrib/examples/web (git:main)]
% npm ls @opentelemetry/resources
web-examples@0.26.0 /Users/trentm/src/opentelemetry-js-contrib/examples/web
├─┬ @opentelemetry/auto-instrumentations-web@0.32.3
│ ├─┬ @opentelemetry/instrumentation-fetch@0.40.0
│ │ └─┬ @opentelemetry/sdk-trace-web@1.14.0
│ │   └─┬ @opentelemetry/sdk-trace-base@1.14.0
│ │     └── @opentelemetry/resources@1.14.0
│ └─┬ @opentelemetry/instrumentation-xml-http-request@0.40.0
│   └─┬ @opentelemetry/sdk-trace-web@1.14.0
│     └─┬ @opentelemetry/sdk-trace-base@1.14.0
│       └── @opentelemetry/resources@1.14.0
├─┬ @opentelemetry/exporter-trace-otlp-http@0.39.1
│ ├─┬ @opentelemetry/otlp-transformer@0.39.1
│ │ ├── @opentelemetry/resources@1.13.0 deduped
│ │ ├─┬ @opentelemetry/sdk-logs@0.39.1
│ │ │ └── @opentelemetry/resources@1.13.0 deduped
│ │ └─┬ @opentelemetry/sdk-metrics@1.13.0
│ │   └── @opentelemetry/resources@1.13.0 deduped
│ ├── @opentelemetry/resources@1.13.0
│ └─┬ @opentelemetry/sdk-trace-base@1.13.0
│   └── @opentelemetry/resources@1.13.0 deduped
└─┬ @opentelemetry/sdk-trace-web@1.23.0
  └─┬ @opentelemetry/sdk-trace-base@1.23.0
    └── @opentelemetry/resources@1.23.0
```

and 7 installs of 3 different versions of `@opentelemetry/semantic-conventions`:

```
% npm ls @opentelemetry/semantic-conventions | rg -v deduped | rg semantic
│ │ └── @opentelemetry/semantic-conventions@1.14.0
│   └── @opentelemetry/semantic-conventions@1.14.0
│ │ └── @opentelemetry/semantic-conventions@1.13.0
│ │ └── @opentelemetry/semantic-conventions@1.13.0
│   └── @opentelemetry/semantic-conventions@1.13.0
│ └── @opentelemetry/semantic-conventions@1.13.0
└── @opentelemetry/semantic-conventions@1.23.0
```

Anyway, it is a bit of a mess. I'm not sure these PRs updating the `examples/*` make it worse, but I understand your point.

Getting into attempting to maintain many of these examples is a huge pain. Some of them are ancient or obsolete (B3Propagator). The examples/web/README.md, for example, explains nothing about what to expect, whether to run the docker-compose file that is included, etc.

MSNev (approved):

examples/web/package.json
trentm: This was discussed in the SIG today (https://docs.google.com/document/d/1tCyoQK49WVcE-x8oryZOTTToFm7sIeUhxFPm9g-qL1k/edit#heading=h.dvqq8u408ifx) and it was felt that we should go ahead with these semconv updates in the `examples/*` dirs.   Trying to summarize the discussion from memory:

- Possible multiple separate `@opentelemetry/semantic-conventions` versions won't break the example apps because the semconv package has no deps.  
- To the degree that users look at the examples for how to do things, it is helpful to update the semantic-conventions package usage to the new exports from v1.22.0.
- There was some expression of hope for automated verification that examples are basically working (perhaps no more than a check that they install/compile fine). Verifying beyond that might be difficult because the examples aren't generally written to be testable. Also contributor bandwidth is an issue.

Issue #1950, opened 2024-02-21
Node - TypeError: Cannot read properties of undefined (reading 'forEach')


Starefossen:
## Steps to Reproduce

Follow the guide for automatic instrumentation: https://opentelemetry.io/docs/languages/js/automatic/

The source code for this application is here: https://github.com/nais/unleash/tree/unleash-v5 and is based on `express` `^4.18.2`.

Environment variables:

```yaml
    - name: NODE_OPTIONS                                                                                                                      
      value: --require @opentelemetry/auto-instrumentations-node/register                                                                               
    - name: OTEL_EXPORTER_OTLP_PROTOCOL                                     
      value: grpc                                                           
    - name: OTEL_NODE_RESOURCE_DETECTORS                                    
      value: env,host,container                                                                                                                         
    - name: OTEL_SERVICE_NAME                                                                                                                           
      value: my-demo                                                                                                                                  
    - name: OTEL_EXPORTER_OTLP_ENDPOINT                                                                                                                 
      value: http://opentelemetry-collector.my-system:4317     
    - name: OTEL_RESOURCE_ATTRIBUTES_POD_NAME                               
      valueFrom:                  
        fieldRef:                                                           
          apiVersion: v1                                                    
          fieldPath: metadata.name
    - name: OTEL_RESOURCE_ATTRIBUTES_NODE_NAME                              
      valueFrom:                                                                                                                                        
        fieldRef:                                                                                                                                       
          apiVersion: v1                                                                                                                                
          fieldPath: spec.nodeName                                          
    - name: OTEL_PROPAGATORS 
      value: tracecontext,baggage,b3                                        
    - name: OTEL_RESOURCE_ATTRIBUTES                                        
      value: k8s.container.name=unleash,k8s.deployment.name=my-demo,k8s.namespace.name=my-unleash,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_N
AME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),k8s.replicaset.name=my-demo-f95cddd45,service.version=v5.9.6-20240221-210653-d96e3be           
    image: europe-north1-docker.pkg.dev/nais-io/nais/images/unleash-v4:v5.9.6-20240221-210653-d96e3be                                                   
```

## Expected Result

It should start properly.

## Actual Result

Application 

### OpenTelemetry Setup Code

```JavaScript
// automatic instrumentation
```

[...]

Issue #9, opened 2020-03-28
express plugin: doesnt work with express-async-errors module


vmarchaud:
`express-async-errors` is a module that wrap express layers so they can be async, however to do so it change the handle fn location (https://github.com/davidbanham/express-async-errors/blob/master/index.js#L36)

sk-:
@vmarchaud any progress on this? The blocking PR was already merged.

vmarchaud:
I stopped using `express-async-errors` and i don't have the time to fix this, feel free to make a PR though

david-luna:
Hi @vmarchaud 

Latest version includes #2137 which also fixes this issue by exposing the `handler` properties in the patched function. I've done a sanity check with the sample given in the package's readme file. This is the code

`pacakge.json`
```json
{
  "name": "xpress-async-errors",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "start": "node --require @opentelemetry/auto-instrumentations-node/register index.js"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@opentelemetry/auto-instrumentations-node": "^0.47.1",
    "express": "^4.19.2",
    "express-async-errors": "^3.1.1"
  }
}
```

`index.js`
```js
const express = require('express');
require('express-async-errors');
const app = express();
const port = 3000;

const User = {
  list: [
    { id: 1, name: 'John Doe', token: '12345' },
  ],
  findAll() {
    return Promise.resolve(this.list);
  },
  findByToken(t) {
    return Promise.resolve(this.list.find(u => u.token === t));
  },
};


app.use(async (req, res, next) => {
  const user = await User.findByToken(req.get('authorization') || 'no-auth');

  if (!user) throw Error("access denied");

  next();
});

app.get('/users', async (req, res) => {
  const users = await User.findAll();
  res.send(users);
});


app.use((err, req, res, next) => {
  if (err.message === 'access denied') {
    res.status(403);
    res.json({ error: err.message });
  }

  next(err);
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
});
```

and starting the application with auto-instrumentations
```sh
npm start
```

then on a separate shell I can request the API without issues, errors are handled correctly

```sh
curl -H "authorization: 125" http://localhost:3000/users
{"error":"access denied"}

curl -H "authorization: 12345" http://localhost:3000/users
[...]

PR #2137, opened 2024-04-22
fix(instr-express): keep hidden properties in layer handlers

david-luna:

- The original `handle` function of an express layer may contain certain metadata, specially in the `router` layer. When instrumentation does the "patch" the patched function does not have the same properties so any code relying on those props will find nothing. That may not be the case in `express` itself but 3rd party tools that look into that metadata are failing.

Fixes: #1950 

- The patched function now exposes same properties proxying the ones from the original `handle` function.
- A test has been added to check the `stack` property is available for the router layer.


plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -165,8 +165,7 @@ export class ExpressInstrumentation extends InstrumentationBase<
       ) {
         const route = original.apply(this, args);
         const layer = this._router.stack[this._router.stack.length - 1];
-        instrumentation._applyPatch.call(
-          instrumentation,
+        instrumentation._applyPatch(
```
david-luna, line 168: Note for reviewer: this usage of `Function.protptype.call` comes from this commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/355/commits/0723d0666b914e132962744b6454c736e5bd165e as a new addition.

I did not dig further to get the original implementation but I think there is no need to use `.call(...)` since we have access to the container object and therefore calling `_applyPatch` with the right context.

david-luna:
> ... will decrease coverage by 0.65%.

This is due to no tests are manipulating the hidden properties of the `handle` function of the layers. so the setter is not covered in the tests.

Screenshot 2024-04-23 at 10 18 27


I can remove the setter since until now not having the property at all was not breaking apps. But I think is a matter of time to get a new issue because of a lib trying to update those rops

JamieDanielson (commented):
Thanks @david-luna ! I will review this week. In the meantime can you run `npm run lint:fix` to update the lint errors?

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -313,6 +310,25 @@ export class ExpressInstrumentation extends InstrumentationBase {
...
       };
+
+      // `handle` isn't just a regular function in some cases. It also contains
+      // some properties holding metadata and state so we need to proxy them
+      // through through patched function
+      // ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1950
+      Object.keys(original).forEach(key => {
+        Object.defineProperty(patched, key, {
+          get() {
+            // @ts-expect-error -- the original function has hidden props indexed by strings
```
JamieDanielson, line 321: While I understand the purpose of what we're trying to achieve, the error that comes up here feels very relevant and makes me wonder if there's a better way to do this.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -313,6 +310,25 @@ export class ExpressInstrumentation extends InstrumentationBase {
           }
         }
       };
+
+      // `handle` isn't just a regular function in some cases. It also contains
+      // some properties holding metadata and state so we need to proxy them
+      // through through patched function
+      // ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1950
+      Object.keys(original).forEach(key => {
+        Object.defineProperty(patched, key, {
```
blumamir, line 319: I wonder why define a property with getter and setter.

Will this work as well?

```js
Object.keys(original).forEach(key => patched[key] = original[key])
```

Or

```js
Object.entries(original).forEach([key, value] => patched[key] = value)
```

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
david-luna: @JamieDanielson 

Thanks for your review :) 

> ...makes me wonder if there's a better way to do this.

For "better" you meant at the type level or you think the usage of `Object.defineProperty` is not a good solution? I'll put my thoughts on both.

**Type error**
The type error we get is:
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'Function'.
  No index signature with a parameter of type 'string' was found on type 'Function'.ts(7053)
```
The internal typings define the `handle` property of the layer as a Function. I've also checked `@types/express` to see if we can get the type from there but the router stack (where the layers are stored) is defines as `any[]`. So at the time of writing it adding comment to highlight express is "leveraging" the Object nature of functions to hide metadata felt the right thing to do. Note: the type error also arises with @blumamir's proposal since its a type level error.

We may get rid of the error by doing some improvements in the internal types so the compiler will know the function being patched is also used as a record. Something like the type below should work.
```ts
type HandlerType = Function & Record <string,any>
```

**Object.defineProperty vs copy**

Doing the shallow copy proposed by @blumamir fixes this issue as well but those properties represents the state of the layer. If that state is modified for the `handler` to change it's behaviour the original one will not notice that change and keep behaving according the original state (before the copy). With `defineProperty` we keep patched and original in sync (sharing state) so the instrumentation will not interfere with any code accessing or changing the state.

david-luna:
Note to self: [this issue](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/9) will probably get fixed by this PR. Check it

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
blumamir: Saw your comment on why this might not work. Thanks for the explanation 🙏 makes sense now

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
JamieDanielson: Ooh, that sounds like a good idea to use that new `HandlerType`. Would you be open to adding that into `internal-types`, and using that instead of ignoring the TypeScript error?

Also thank you for the detailed explanation, that helps me understand much better!

blumamir (approved):

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
blumamir: > Ooh, that sounds like a good idea to use that new `HandlerType`. Would you be open to adding that into `internal-types`, and using that instead of ignoring the TypeScript error?
> 
> Also thank you for the detailed explanation, that helps me understand much better!

I support doing this

david-luna: I'll update the types :)

david-luna: Done @JamieDanielson :)

PR #2145, opened 2024-04-24
test(instr-document-load): fix test to allow missing network span events

trentm:
The change in https://github.com/open-telemetry/opentelemetry-js/pull/4486
means that a addSpanNetworkEvent() in  v1.24.0 and
later might get dropped -- if its time value is before the fetchStart time.
Typically this happens if the event time value is 0.


trentm:
This should fix the browser-test failure on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2091

trentm:
This was written to have the test pass with sdk-trace-web versions before and after 1.24.0 where the change in behaviour happened, because instr-documentation-load has this dep:  `"@opentelemetry/sdk-trace-web": "^1.15.0",`.

JamieDanielson (approved):
Thanks for picking this up so quickly! This looks like a better solution than what I was thinking, which was just to set the `unloadEventStart` and `unloadEventEnd` to `1571078170306` (one tick higher than `fetchStart`)

PR #2154, opened 2024-04-26
refactor(instr-connect): use exported strings for attributes

david-luna:

- Updates #2025 

- Replace `SemanticAttributes.*` with exported strings `SEMATTRS_*`
- Update README with new semantic convention package version and key


maryliag (approved):

plugins/node/opentelemetry-instrumentation-connect/README.md
```diff
@@ -51,6 +51,16 @@ registerInstrumentations({
...
+## Semantic Conventions
+
+This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+
+Attributes collected:
+
+| Attribute    | Short Description                  |
+| ------------ | ---------------------------------- |
+| `http.route` | The matched route (path template). |
+
```
blumamir, line 63: I wonder if it can be helpful to also enumerate the other, non-spec-compliant attributes that are emitted by this instrumentation, as a reference for users and for them to know what to expect when using this instrumentation.

There seems to be 2 custom attributes at the moment:
```js
export enum AttributeNames {
  CONNECT_TYPE = 'connect.type',
  CONNECT_NAME = 'connect.name',
}
```

Do you think eventually we will need to either remove them or add them to the specs? Or eventually, each instrumentation is allowed to record extra attributes for which we also need some sort of documentation and stability guarantee?

blumamir (approved):
LGTM

Left a general question about the current/future goal for the non-semantic attributes

plugins/node/opentelemetry-instrumentation-connect/README.md
pichlermarc: > Do you think eventually we will need to either remove them or add them to the specs? 

I'm not sure. I reached out to @joaopgrassi (semconv maintainer) to ask what the general rules are for instrumentations that emit non-semconv telemetry. It'll be brought up in the semconv SIG meeting today :slightly_smiling_face:

blumamir: Thanks @pichlermarc  and @joaopgrassi .

Any update on this issue?

I think it should not block the current PR. The non-semantic attributes can also be removed / added to docs in followup PRs once we get more guidelines.

JamieDanielson: True we can probably open this as a new issue to look into and follow-up on.

JamieDanielson (approved):

plugins/node/opentelemetry-instrumentation-connect/README.md
JamieDanielson: I've created #2170 for follow-up

Issue #1612, opened 2023-07-26
opentelemetry-instrumentation-graphql doesn't follow GraphQL semantic conventions


SonjaChevre:
- [ ] This only affects the JavaScript OpenTelemetry library
- [x] This may affect other libraries, but I would like to get opinions here first

Hi! in the process of writing a blog post about GraphQL & OpenTelemetry, I have noticed that the GraphQL instrumentation doesn't respect the current (basic) GraphQL semantic convention. 

A couple of findings:

- One example: using graphql.source for the query and not graphql.document
- On the other side, opentelemetry-instrumentation-graphql provides lots of insights that could be standardised to the semantic conventions
- Last, adding information about GraphQL errors on the spans (missing from the instrumentation and the semantic conventions) could have a lot of value for users - see here: https://tyk.io/blog/observing-graphql-errors-with-opentelemetry/

Making a note here, to let you know I have created an issue to suggest working on better semantic conventions for GraphQL: https://github.com/open-telemetry/semantic-conventions/issues/182 - would welcome anybody that join this initiative.

Thanks,
Sonja

PR #2156, opened 2024-04-26
refactor(instr-graphql): use exported strings for attributes

david-luna:

- Updates #2025 

- Replace `SemanticAttributes.*` with exported strings `SEMATTRS_*` in test files

Note: I did not add the "semantic conventions" section in the readme file since this instrumentation is using it only in tests but not for span attribs


blumamir (approved):
LGTM

Should we add some note in the README about this package not implementing any semantic convention at the moment?

david-luna:
> LGTM
> 
> Should we add some note in the README about this package not implementing any semantic convention at the moment?

I thought the absence of the section in the README was enough but I based on your comment I guess it would be better that all packages have a section about semconv

blumamir:
> > LGTM
> > Should we add some note in the README about this package not implementing any semantic convention at the moment?
> 
> I thought the absence of the section in the README was enough but I based on your comment I guess it would be better that all packages have a section about semconv

That's my personal opinion 😄 Feel free to add one or resolve if you prefer the READMEs not to state it in this case

JamieDanielson:
> > > LGTM
> > > Should we add some note in the README about this package not implementing any semantic convention at the moment?
> > 
> > 
> > I thought the absence of the section in the README was enough but I based on your comment I guess it would be better that all packages have a section about semconv
> 
> That's my personal opinion 😄 Feel free to add one or resolve if you prefer the READMEs not to state it in this case

Oh interesting! It looks like graphql semantic conventions do exist, and were recently [added to the registry in semantic-conventions repo](https://github.com/open-telemetry/semantic-conventions/pull/882), though do not appear to exist in the schema itself yet unless I am missing it.

Also seeing there was an [issue opened (closed for stale)](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1612) referencing unexpected attributes. So this does seem like something that we explicitly should state that we are not using semantic conventions here for graphql attributes... and a new issue should perhaps be opened to review that. I'm not clear on whether this would be included in the new generated semantic conventions if it doesn't exist in the schema yaml files, but it is something that appears to have expected conventions at this point.

JamieDanielson (approved):
On #2182 I tried out this wording; if we like it we can use it here.

```markdown
## Semantic Conventions

This package does not currently generate any attributes from semantic conventions.
```

Also as far as the other attributes that do seem like they should be pulled from semantic conventions and possibly updated as well... let's deal with that separately. This is strictly documenting the current state.

david-luna:
@JamieDanielson I've added your suggestion. Thanks :)

plugins/node/opentelemetry-instrumentation-graphql/package.json
```diff
@@ -46,7 +46,7 @@
   "devDependencies": {
     "@opentelemetry/api": "^1.3.0",
     "@opentelemetry/sdk-trace-base": "^1.8.0",
-    "@opentelemetry/semantic-conventions": "^1.3.1",
+    "@opentelemetry/semantic-conventions": "^1.22.1",
```
trentm, line 49: This should be:

```suggestion
    "@opentelemetry/semantic-conventions": "^1.22.0",
```

and then package-lock updated. There isn't a 1.22.1 release.

david-luna: done!

trentm (approved):

PR #2161, opened 2024-04-27
fix: remove unuseful patch message from instrumentations

blumamir:
#2107 removed the diag prints on `InstrumentationNodeModuleDefinition` and `InstrumentationNodeModuleFile`, `patch` and `unpatch` callbacks, as this message is now hoisted to be emitted consistently from the base class. The scope of this PR was narrow to handle specific types of messages that became redundant, but it was noted that there are other diag prints for patching and unpatching specific function from moduleExports.

This PR is a followup to #2107 

- Cleanup diag patch messages for trivial function which created noise and added no real value for troubleshooting to users.
- Document guidelines on using diag channel to emit patch messages, based on the examples I observed in the repo.


blumamir:
@JamieDanielson @trentm @pichlermarc  thanks again for the great feedback and discussion on the previous PR.

Since those changes were intentionally left out of the previous one, I opened a follow-up to apply the issues we talked about in a dedicated PR. I also added a section to the GUIDELINES.md file for future reference to capture the decisions.

Please let me know if this work correctly reflects the discussion in #2107 🙏

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
...
+
+It can also be useful to know when the instrumentation is loaded and patched, to understand the order of operations in the application.
+
+Instrumentation packages should use the `@opentelemetry/instrumentation` package `BaseInstrumentation` class to register patches and unpatch callbacks for specific require/import of the instrumented package, it's dependency or an internal module file. When this mechanism is used, the base class will automatically emit a debug message on instrumentation diag component logger, looking like this:
+
+```shell
+@opentelemetry/instrumentation-foo Applying instrumentation patch for module on require hook {
+  module: 'foo',
+  version: '1.2.3',
+  baseDir: '<your directory>/node_modules/mongoose'
```
JamieDanielson, line 187:
```suggestion
  baseDir: '<your directory>/node_modules/foo'
```
nit: for consistency match module name

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
...
+}
+```
+
+Instrumentation should not add additional debug messages for triggering the patching and unpatching callbacks, as the base class will handle this.
+
+Instrumentation may add additional patch/unpatch messages for specific functions if it is expected to help in troubleshooting issues with the instrumentation. Few examples:
+
+- If the patch logic is conditional, and user can benefit from ensuring the condition is met and the patch happened. `koa` patching logic examine an object and branch between patching it as router vs middleware, which is applied at runtime. `awslambda` will aboard patching if the environment is not configured properly.
```
JamieDanielson, line 195:
```suggestion
- If the patch logic is conditional, and user can benefit from ensuring the condition is met and the patch happened. `koa` patching logic examine an object and branch between patching it as router vs middleware, which is applied at runtime. `aws-lambda` will abort patching if the environment is not configured properly.
```
Let me know if I have that right, I believe the intent is to say it will not patch, so abort is the word we want here.

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
...
+- When the patch is not applied directly on a `moduleExports` object in the `BaseInstrumentation` callbacks, but rather from an event in the package, like creating new client instance, registering a listener, etc. `fastify` instrumentation applies a patch when an hook is added to the fastify app instance, which is patched from `moduleExports`.
```
JamieDanielson, line 196:
```suggestion
- When the patch is not applied directly on a `moduleExports` object in the `BaseInstrumentation` callbacks, but rather from an event in the package, like creating new client instance, registering a listener, etc. `fastify` instrumentation applies a patch when a hook is added to the fastify app instance, which is patched from `moduleExports`.
```

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
...
+- In situations where the patch logic is not trivial and it helps to specify patch events in the right context and nuances. `aws-lambda` logs additional properties extracted from the labmda framework and exposes them for troubleshooting.
```
JamieDanielson, line 197:
```suggestion
- In situations where the patch logic is not trivial and it helps to specify patch events in the right context and nuances. `aws-lambda` logs additional properties extracted from the lambda framework and exposes them for troubleshooting.
```

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
...
+
+The cases above are not covered by the base class and offer additional context to the user trouble shooting an issue with the instrumentation.
```
JamieDanielson, line 199:
```suggestion
The cases above are not covered by the base class and offer additional context to the user troubleshooting an issue with the instrumentation.
```

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
     };
     ...
     ```
+
+## Diag Channel
```
JamieDanielson, line 168: 🤩  Love having this guideline written out here

JamieDanielson (approved):
I think this looks great - I especially love the addition to the guidelines explaining where it might be useful to have logging about patches. I left a few nits but nothing blocking from me.

GUIDELINES.md
```diff
@@ -164,3 +164,36 @@ To support this use case, you can choose one of the following options:
...
+@opentelemetry/instrumentation-foo Applying instrumentation patch for module on require hook {
+  module: 'foo',
+  version: '1.2.3',
+  baseDir: '<your directory>/node_modules/mongoose'
+}
+```
+
+Instrumentation should not add additional debug messages for triggering the patching and unpatching callbacks, as the base class will handle this.
+
+Instrumentation may add additional patch/unpatch messages for specific functions if it is expected to help in troubleshooting issues with the instrumentation. Few examples:
```
JamieDanielson, line 193: Oh, another thought I had that I think will probably be updated in the follow-up PR that replaces inconsistent usages of `api.diag.debug` and similar that aren't using the instrumentation `this` to allow for `this._diag.debug`. When we make that change we should update guidelines for that as well, though it will already be easier to follow suit once these are all more consistent 😄

blumamir: certainly, I aim to first remove or hoist unneeded diag messages, and then will transform all remaining diag message to use `this._diag` and update the GUIDELINES.md 👍 

https://github.com/open-telemetry/opentelemetry-js/pull/4663 will also help in hosting many diag prints and remove the repeated code from  dozens of places in contrib.

blumamir:
Thank you @JamieDanielson

Your suggestions are always valuable. addressed all the comments 🙏

GUIDELINES.md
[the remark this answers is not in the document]
blumamir: I remember that "diag channel" is how we used to refer to it when speaking about it, but I now understand it was not formal name. 
Changed to `Diag Logging`. Thanks!

JamieDanielson (approved):

PR #2162, opened 2024-04-27
chore: align config constructor pattern across instrumentations

blumamir:
There are dozens of instrumentations in this repo, written at different times by different people.
Few different patterns emerged on how to handle the config object in Instrumentation constructor. This PR aims to introduce a common pattern to have consistency across the repo. 

This PR aligns all the instrumentations to use:
```js
  constructor(config: FooInstrumentationConfig = {}) {
    super('@opentelemetry/instrumentation-foo', VERSION, config);
  }
```


blumamir:
After https://github.com/open-telemetry/opentelemetry-js/pull/4659 lands, we can introduce even more consistency and cleanups on the handling of config in instrumentations.
This PR only targets the `constructor` and `super` invocation usage

plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts
```diff
@@ -65,9 +64,7 @@ const DEFAULT_CONFIG: GraphQLInstrumentationConfig = {
 const supportedVersions = ['>=14 <17'];
 
 export class GraphQLInstrumentation extends InstrumentationBase {
-  constructor(
-    config: GraphQLInstrumentationConfig & InstrumentationConfig = {}
```
blumamir, line 69: I removed it because `GraphQLInstrumentationConfig` already extends `InstrumentationConfig` so it was redundant

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -77,8 +77,10 @@ export class AwsLambdaInstrumentation extends InstrumentationBase {
   private _traceForceFlusher?: () => Promise<void>;
   private _metricForceFlusher?: () => Promise<void>;
 
-  constructor(protected override _config: AwsLambdaInstrumentationConfig = {}) {
-    super('@opentelemetry/instrumentation-aws-lambda', VERSION, _config);
+  protected override _config!: AwsLambdaInstrumentationConfig;
```
david-luna, line 80: This is another bit we may want to align, which is the way we type config in our instrumentations. Here we use the `override` keyword so we tell typescript to treat the property as an `AwsLambdaInstrumentationConfig` type. And in other instrumentations we have a getter that casts the config property to the type we need


```ts
getConfig(): AwsLambdaInstrumentationConfig {
  return this._config as AwsLambdaInstrumentationConfig;
}
```

Should we encourage one over he other?

david-luna (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
blumamir: Once https://github.com/open-telemetry/opentelemetry-js/pull/4659 lands, we can cleanup all the config handling variations as the base class would already be typed correctly. Eager to apply this cleanup in all contrib instrumentations

JamieDanielson (commented):
Looks like we are just missing Mongoose and Koa, feel free to [grab this commit](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/a0ecb23de1ca80345e8cee591e64fcfc9ec0d4f9) if it's easiest. Otherwise this looks good.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -161,7 +157,6 @@ export class PgInstrumentation extends InstrumentationBase {
   private _getClientQueryPatch() {
     const plugin = this;
     return (original: typeof pgTypes.Client.prototype.query) => {
-      this._diag.debug('Patching pg.Client.prototype.query');
```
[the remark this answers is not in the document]
blumamir: oops. reverted 👍

blumamir:
> Looks like we are just missing Mongoose and Koa, feel free to [grab this commit](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/a0ecb23de1ca80345e8cee591e64fcfc9ec0d4f9) if it's easiest. Otherwise this looks good.

right! I missed those. thanks for spotting. added them to the PR

JamieDanielson (approved):

trentm:
The TAV test for Node.js v16 failed on instr-undici with undici@5.28.2.

https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/8899428327/job/24438782853?pr=2162

```
-- required packages ["undici@5.28.2"]
-- installing ["undici@5.28.2"]
-- running test "npm run test" with undici (env: {})

> @opentelemetry/instrumentation-undici@0.2.0 test
> nyc ts-mocha -p tsconfig.json test/**/*.test.ts



  UndiciInstrumentation `fetch` tests
    disable()
      - should not create spans when disabled
    enable()
      - should create valid spans even if the configuration hooks fail
      - should create valid spans with empty configuration
      - should create valid spans with the given configuration
      - should not create spans without parent if required in configuration
      - should not create spans with parent if required in configuration
      - should capture errors using fetch API
      - should capture error if fetch request is aborted

  UndiciInstrumentation metrics tests
    with fetch API
      - should report "http.client.request.duration" metric
      - should have error.type in "http.client.request.duration" metric

  UndiciInstrumentation `undici` tests
    disable()
      ✓ should not create spans when disabled (38ms)
    enable()
      ✓ should ignore requests based on the result of ignoreRequestHook
      - should create valid spans for different request methods
      ✓ should create valid spans for "request" method
      ✓ should create valid spans for "fetch" method
      ✓ should create valid spans for "stream" method
      ✓ should create valid spans for "dispatch" method
      ✓ should create valid spans even if the configuration hooks fail
      ✓ should not create spans without parent if required in configuration
      ✓ should create spans with parent if required in configuration
      1) should capture errors while doing request
      ✓ should capture error if undici request is aborted


  10 passing (2s)
  11 pending
  1 failing

  1) UndiciInstrumentation `undici` tests
       enable()
         should capture errors while doing request:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/plugins/node/instrumentation-undici/test/undici.test.ts)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

I'll re-run that test. I can't see how this PR would ahve caused it.

trentm (approved):

PR #1985, opened 2024-03-02
feat(instrumentation-hapi): support v21

nlochschmidt:
## Which problem is this PR solving?

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1383

## Short description of the changes

Continuation of stale #1491

- Updates Hapi dependency to v21
- Uses types shipped with library instead of `@types/hapi__hapi`
- Adds Hapi v21 to the list of tested versions
- Supports tracing when route.options is a function

PR #2165, opened 2024-04-29
chore: remove skipLibCheck: true from tsconfigs where it is not necessary

pichlermarc:
`skipLibCheck: true` is set a few instrumentation libraries, even though they're not necessary anymore. The only remaining instrumentation that needs it is `@opentelemetry/instrumentation-lru-memoizer`.

### DIFF
--- a/metapackages/auto-instrumentations-web/tsconfig.esm.json
+++ b/metapackages/auto-instrumentations-web/tsconfig.esm.json
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "build/esm",
-    "skipLibCheck": true,
     "tsBuildInfoFile": "build/esm/tsconfig.esm.tsbuildinfo"
   },
   "include": [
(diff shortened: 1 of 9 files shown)


david-luna (approved):
> even though they're not necessary anymore...

LGTM but could I get more context on this? Thanks!

pichlermarc:
> > even though they're not necessary anymore...
> 
> LGTM but could I get more context on this? Thanks!

We've seen this cause some confusion over at #1985. 

It looks like some instrumentations had set `skipLibCheck: true`. I assume this to be either an unintentional copy-paste from an existing instrumentation that had this option set, or it was necessary back then due to actually missing types. Either way, nowadays, the instrumentation compile fine without that flag, so I think it's best to revert to the more common strict config that we use.

blumamir (approved):

blumamir:
> The only remaining instrumentation that needs it is `@opentelemetry/instrumentation-lru-memoizer`.

curious why this instrumentation still need the setting

pichlermarc:
> > The only remaining instrumentation that needs it is `@opentelemetry/instrumentation-lru-memoizer`.
> 
> curious why this instrumentation still need the setting

There's some problem with the `lru-cache` types. I looked into it briefly but couldn't find any way to resolve the situation. I think this may be a problem with the upstream packages we use.

Issue #1672, opened 2023-09-07
NodeJS auto-instrumentation: Allow disabling specific instrumentations with environment variables


FredrikAugust:
### Describe the solution you'd like to see

We would like to keep all instrumentations active _except for_ `redis-4`. [This is something that is possible with the Java SDK](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#suppressing-specific-agent-instrumentation), but currently not with the JavaScript SDK, using the `OTEL_INSTRUMENTATION_[NAME]_ENABLED` environment variables.

[Something like this is possible for the resource detectors](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/99ce679f8839d113088d340d2a448b0098d646fc/metapackages/auto-instrumentations-node/src/utils.ts#L170) so a solution akin to that would be amazing.

### Describe alternatives you've considered

As we're using the Kubernetes OTEL Agent Operator we are AFAIK limited to environment variables in terms of configuration of the autoinstrumentation so the only other solution I've thought of is disabling the instrumentation all-together.

### Additional context

I'll happily contribute to this if you'd like:)

FredrikAugust:
Alright, we just confirmed that disabling the OTEL instrumentation no longer produces the error described above. Should I create a separate issue for that as it's likely a redis issue?

peskic93:
I second this. It's really annoying having to implement manual configuration file for tracing just to disable/enable single instrumentation. Implementation for this should not be complex at all, as @FredrikAugust mentioned, it's already done for resource detectors.

FredrikAugust:
I don't think this would be a lot of work, and would allow people to control what instrumentations are being used so I would love to see this implemented!

hamin:
Can we please get this in?

atsu85:
# I need this feature as well
because application startup is a lot slower with all integrations enabled (specially if CPU is limited)

I'm using k8s opentelemetry operator, that [loads all instrumentations](https://github.com/open-telemetry/opentelemetry-operator/blob/d9e1d161366dbf967a9cb47f3d6ee6137ec4d8d6/autoinstrumentation/nodejs/src/autoinstrumentation.ts#L37), so applications that use it can't customize it.

# Alternative solutions
One option for me would be to solve [the issue in opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator/issues/2622#issuecomment-1956683397), but i think it would be bad idea:
[...]

PR #2174, opened 2024-05-02
feat(auto-instrumentations-node): disabling instrumentations via env var

### LINKED ISSUE #1331: Support to disable specific library or middleware instrumentation via environment variable

My application is based on NodeJs  express library and I tried to achieve auto instrumentation using different approaches as mentioned below

Consider the trace shown in the image below when the application is instrumented  (see Additional context section for dependencies used below)
image

I also have a different setup in Kubernetes cluster based on [OpenTelemetry Operator](https://github.com/open-telemetry/opentelemetry-operator)
image

On this occasion, the purple comes from the instrumented library `@opentelemetry/instrumentation-net`.

Even though the spans from auto-instrumentation is useful, they can clutter application specific spans and often digging through the span mesh to identify application specific spans can be frustrating.

From the screen shot above, I would like to have only the span highlighted in red displayed and the other spans(middleware or other libraries) disabled by some configuration or environment variables. Even though the spans provide useful information, but most of the times it adds unnecessary information cluttering the application specific spans. Often it becomes very hard to identify the application specific spans among the several library instrumented spans. I would like to see some support for environment variables to disable spans for specific instrumented libraries (Similar thing is possible with JavaAgent based instrumentation, see [suppressing-specific-auto-instrumentation](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#suppressing-specific-auto-instrumentation)). For example, there could be an environment variable to disable spans when not required for a specific library like `@opentelemetry/instrumentation-net` or support to disable middleware spans in any instrumented library.

In another approach in Kubernetes cluster, I used [OpenTelemetry Operator](https://github.com/open-telemetry/opentelemetry-operator) and the corresponding [annotations on the pod](https://github.com/open-telemetry/opentelemetry-operator#opentelemetry-auto-instrumentation-injection) to inject instrumentation libraries for NodeJS applications. Since, the OpenTelemetry Operator is a "client" of OpenTelemetry upstream instrumentation,  it would really be helpful to have environment variables to disable auto-instrumentation of specific libraries.

Not aware of any. Only thing I tried is to use a individual dependencies instead of [auto-instrumentations-node](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/auto-instrumentations-node). Even while using  `auto-instrumentation-node` package, it would be nice to have support for environment variables to disable auto-instrumentation for specific libraries.

### Additional context

The express application has following dependencies used for instrumentation:
```
"@opentelemetry/api": "1.3.0",
"@opentelemetry/exporter-trace-otlp-http": "0.34.0",
"@opentelemetry/instrumentation": "0.34.0",
"@opentelemetry/instrumentation-express": "0.32.0",
"@opentelemetry/instrumentation-http": "0.32.0",
"@opentelemetry/resources": "1.8.0",
"@opentelemetry/sdk-trace-base": "1.8.0",
"@opentelemetry/sdk-trace-node": "1.8.0",
"@opentelemetry/semantic-conventions": "1.8.0",
```

Thanks in advance!!!

**Best,**
Santosh

JamieDanielson:

- Similar to #1953 this provides the alternative option discussed in some comments for disabling some instrumentations via environment variable.
- Fixes #1331 
- Finishes fix for #1672
- Helps to fix https://github.com/open-telemetry/opentelemetry-operator/issues/2622

- Disable specific instrumentations using `OTEL_NODE_DISABLED_INSTRUMENTATIONS` environment variable. This is similar to the use case when instrumenting in code using `getNodeAutoInstrumentations()` where someone may want to start with every instrumentation and pare down from there.


metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -203,6 +205,23 @@ function getEnabledInstrumentationsFromEnv() {
...
 
+/**
+ * Returns the list of instrumentations that are disabled based on the environment variable.
+ */
+function getDisabledInstrumentationsFromEnv() {
+  if (!process.env.OTEL_NODE_DISABLED_INSTRUMENTATIONS) {
+    return [];
+  }
+
+  const instrumentationsFromEnv =
```
david-luna, line 216: nit: both vars have the same format and map the same way. Could we move it to a separate function?

JamieDanielson: Good point, I did a small cleanup here

JamieDanielson:
> I see that adding this env var was already commented in the linked PR [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1953#issuecomment-1981131555). I guess you already have the use case for this :)

I do keep getting the request for the simplest possible setup, and iterating from there. So this is similar to how many folks have gotten started with the `auto-instrumentations-node` package as it is. Start with everything enabled by default, and then disable one by one as you decide what you don't want, instead of trying to figure out everything you do want. Later as you understand your telemetry more it may be possible to switch from one to the other (start with disabling specific ones, then later only need to enable specific ones). But yes, that is the reason for this. Also other languages have options for both disable and enable specific instrumentations, including [dotnet](https://opentelemetry.io/docs/languages/net/automatic/instrumentations/#instrumentations) and [java](https://opentelemetry.io/docs/languages/java/automatic/configuration/#suppressing-specific-agent-instrumentation), and Python has [Disabled](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fdcbbddb6c753e5e9d494ba399a5b4bcab4afc3f/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation/_load.py#L60) instead of Enabled.

blumamir (approved):
I strongly support this feature, which I find more usable than the `enabled` list option.

The code changes looks good to me.

Thank you for working on this and adding this option

JamieDanielson:
@open-telemetry/javascript-approvers Just wanted to ping to see if anyone else wanted a review before merging, since this is the auto-instr package so it is a bit more far-reaching than a single instrumentation. Otherwise will merge on Monday!

pichlermarc:
> @open-telemetry/javascript-approvers Just wanted to ping to see if anyone else wanted a review before merging, since this is the auto-instr package so it is a bit more far-reaching than a single instrumentation. Otherwise will merge on Monday!

I'm neutral on that feature. I'm usually a big proponent of people knowing which instrumentations are enabled, that's why I prefer explicitly enabling them - if we have enough that asked for this after we added the enabled env var then adding this should be okay. :)

gnz00:
> > @open-telemetry/javascript-approvers Just wanted to ping to see if anyone else wanted a review before merging, since this is the auto-instr package so it is a bit more far-reaching than a single instrumentation. Otherwise will merge on Monday!
> 
> 
> 
> I'm neutral on that feature. I'm usually a big proponent of people knowing which instrumentations are enabled, that's why I prefer explicitly enabling them - if we have enough that asked for this after we added the enabled env var then adding this should be okay. :)

This is essential for an out of the box setup using the Kubernetes  instrumentation CRD to auto-instrument existing services. We switched from using the datadog-agent to the opentelemetry-collector and the fs instrumentation overran our budget. I think it also adds parity to the other instrumentation libraries, e.g. 

OTEL_DOTNET_AUTO_METRICS_{0}_INSTRUMENTATION_ENABLED

Most of the guides you'll find for the NodeSdk look like this:

```
const sdk: NodeSDK = new NodeSdk({
  instrumentations: [    
    getNodeAutoInstrumentations({
      // We recommend disabling fs automatic instrumentation because 
      // it can be noisy and expensive during startup
      '@opentelemetry/instrumentation-fs': {
        enabled: false,
      },
    }),
  ],
});
```

By adding this environment variable, you can replicate this configuration using the opentelemetry-collector's injector without modifying existing services.

nuclearpidgeon:
Thanks for pushing through this change Jamie - where I work we were doing a lot of OpenTelemetry research earlier in the year and the whole "starting up an otel-instrumented hello world nodejs express app spits out 600 filesystem traces from module loading" thing was a big early problem I came across. I luckily also came across your example app repo that had [the magic `requireParentSpan: true` line](https://github.com/JamieDanielson/node-otel-example-app/blob/main/instrumentation.ts#L44) which stopped this, but having an environment variable now for the zero-code instrumentation case will make adopting OTel more attractive :-)

PR #2176, opened 2024-05-03
test(instrumentation-pino): add test for ESM

JamieDanielson:

- updates #1942 

- add back test for ESM support in pino instrumentation, which was inadvertently dropped in a separate PR


trentm (approved):

trentm:
Thanks for re-adding this!

PR #2131, opened 2024-04-19
feat(auto-instrumentation-node): add undici instrumentation

JamieDanielson:
## Which problem is this PR solving?

- auto-instrumentations-node is expected to include all instrumentations in js and js-contrib, but didn't yet have the newly created and published undici

## Short description of the changes

- add undici instrumentation to auto-instrumentations-node package

Note: As I understand it, we haven't had an official "release" of undici instrumentation, but we recently found that new packages like this still [got published on npm](https://www.npmjs.com/package/@opentelemetry/instrumentation-undici) due to a [misconfiguration in our workflows](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2020). 

Also, I couldn't recall if there was a reason to delay adding this to the package (like since it's still so new), but figured I'd open the PR either way because I didn't see it otherwise.

PR #2177, opened 2024-05-04
feat: Add Azure Functions instrumentation

ejizba:

The Azure Functions team plans to announce preview support soon for Open Telemetry across several languages (see [this GitHub issue](https://github.com/Azure/azure-functions-host/issues/9273) for full details). I work specifically on the Node.js experience, including the `@azure/functions` npm package [here](https://www.npmjs.com/package/@azure/functions). A lot of the implementation for OTel was done in our shared host component which is written in .NET (PR [here](https://github.com/Azure/azure-functions-host/pull/9985)).

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/245

Added new npm package that can do the following for Azure Functions users:
1. Correlate the traces generated from the user's node.js code to the trace context provided by our shared host component
2. Send logs directly from Node.js instead of our shared host component (this results in better correlation, more flexibility for the user, etc.)

Added my coworker @castrodd and manager @AnatoliB as additional owners.


plugins/node/opentelemetry-instrumentation-azure-functions/README.md
```diff
@@ -0,0 +1,58 @@
+# OpenTelemetry Azure Functions Instrumentation for Node.js
+
+[![NPM Published Version][npm-img]][npm-url]
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for the [`@azure/functions`](https://learn.microsoft.com/azure/azure-functions/functions-reference-node?pivots=nodejs-model-v4) module, which may be loaded using the [`@opentelemetry/sdk-trace-node`](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-sdk-trace-node) package and is included in the [`@opentelemetry/auto-instrumentations-node`](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node) bundle.
+
+If total installation size is not constrained, it is recommended to use the [`@opentelemetry/auto-instrumentations-node`](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node) bundle with [@opentelemetry/sdk-node](`https://www.npmjs.com/package/@opentelemetry/sdk-node`) for the most seamless instrumentation experience.
```
hectorhdzg, line 8: This package is not being added as a dependency to auto-instrumentations-node project, maybe add the dependency in the project or remove the documentation and add it at a later time when available.

plugins/node/opentelemetry-instrumentation-azure-functions/README.md
```diff
@@ -0,0 +1,58 @@
+# OpenTelemetry Azure Functions Instrumentation for Node.js
+
+[![NPM Published Version][npm-img]][npm-url]
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for the [`@azure/functions`](https://learn.microsoft.com/azure/azure-functions/functions-reference-node?pivots=nodejs-model-v4) module, which may be loaded using the [`@opentelemetry/sdk-trace-node`](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-sdk-trace-node) package and is included in the [`@opentelemetry/auto-instrumentations-node`](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node) bundle.
```
hectorhdzg, line 6: Describing that the module does would be great here, looks like you are doing only context propagation right now, I wonder if it makes sense to have this package as part of the [propagators](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/propagators) in this repo instead. @open-telemetry/javascript-approvers any suggestions here?

plugins/node/opentelemetry-instrumentation-azure-functions/README.md
ejizba: Yes, we would like it added to the auto-instrumentations package. Not sure on the ordering if I can add it as a dependency in this PR or would need to wait until after the first release

plugins/node/opentelemetry-instrumentation-azure-functions/README.md
ejizba: Interesting, I see a lot fewer examples for propagators but I will look into that.

For the record, this isn't all the package will ever do. Unfortunately I'm blocked from adding some more functionality while I wait for some Azure rollouts. For example, here's a small commit that illustrates how we also want to emit logs from here: https://github.com/ejizba/opentelemetry-js-contrib/commit/21b8e1d85e6b0f785bebb1a4accabb8b6519d023

hectorhdzg: We discussed this in today's SIG meeting, and it should be fine to have this and other future functionalities here as an instrumentation package.

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
```diff
@@ -0,0 +1,59 @@
...
+        return moduleExports;
+      },
+      () => disposable?.dispose()
+    );
+  }
+
+  private _preInvocationHook(context: azFunc.PreInvocationContext): void {
+    const traceContext = context.invocationContext.traceContext;
+    if (traceContext) {
+      context.functionHandler = otelContext.bind(
```
hectorhdzg, line 50: It would be good to add some comment describing what this functionHandler does

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
```diff
@@ -0,0 +1,59 @@
...
+        propagation.extract(otelContext.active(), {
```
hectorhdzg, line 51: I believe this only works if there is a W3CTraceContext propagator registered right?, this is not my area of expertise to be honest here, but I would expect that if you wanted to change current context you can do it using existing trace APIs, if you look at W3CTraceContextPropagator code, it will parse the traceParent and traceState, it will create a spanContext and then would be calling trace.setSpanContext

plugins/node/opentelemetry-instrumentation-azure-functions/README.md
trentm: I think it could be added as part of this change, if you like. Add a dependency to auto-instrumentations-node on version 0.1.0 of this package.  I've not seen a new instrumentation added to auto-instrumentations-node in the first commit before, but then they don't get added that frequently.

If it helps, here is a diff showing the last instr added to auto-instrumentations-node: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2131/files

plugins/node/opentelemetry-instrumentation-azure-functions/.eslintignore
trentm, line 1: I haven't reviewed this PR, but a first note is that the directory name should not ahve the "opentelemetry-" prefix. I.e. use `plugins/node/instrumentation-azure-functions`. I hope that doesn't cause too much pain for your PR.

Some recently added packages are: https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/instrumentation-runtime-node/ and https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/instrumentation-undici

We'd like to move all current packages to directories without the "opentelemetry-" prefix, but that hasn't been done yet. (It'll break README.md and other source code deep links and also cause some pain for open PRs, so there is some hesitance.)

ejizba: No problem, will update

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
ejizba: I've never explicitly registered a W3CTraceContext propagator and it's worked just fine for me 🤷‍♂️. [Here's a sample app](https://github.com/ejizba/azure-functions-nodejs-opentelemetry/blob/advanced-previewpackage/src/index.ts) where I'm using this. I am using auto instrumentation stuff, so not sure if that pulls it in

It's a little weird because we don't want to _create_ a span context, just bind to an existing context. The Azure Functions Host (our shared .NET component) creates the span for every function invocation not us

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
ejizba: This functionHandler is _the_ function from Azure Functions that the user defines in their app. People familiar with Azure Functions would know, but I suppose not your average Otel person not familiar with Azure Functions

plugins/node/opentelemetry-instrumentation-azure-functions/README.md
ejizba: Removed readme note for now. I'll add to auto package separately in the future

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
ejizba: added a few comments

plugins/node/opentelemetry-instrumentation-azure-functions/README.md
ejizba: Our changes finished rolling out in Azure, so I was able to finally add the rest of the main functionality we wanted.

plugins/node/instrumentation-azure-functions/src/instrumentation.ts
```diff
@@ -0,0 +1,113 @@
...
+    azFunc.app.setup({
+      capabilities: {
+        WorkerOpenTelemetryEnabled: true,
+      },
+    });
+
+    // Send logs directly from Node.js
+    disposables.push(
+      azFunc.app.hook.log(context => {
+        this.logger.emit({
```
hectorhdzg, line 57: You may want to consider adding a configuration option for log auto generation, is not enforced but we do it in other instrumentation like bunyan and winston, some customers may not want this telemetry

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
hectorhdzg: The W3CTraceContext is actually registered in BasicTracerProvider that NodeTracerProvider extends from, so not sure if there is an actual scenario where this is not available, I cannot find some actual documentation about how we expect people to work with the context, @pichlermarc do you know if such documentation exists?

plugins/node/instrumentation-azure-functions/src/instrumentation.ts
ejizba: I'd prefer to wait for people to ask for a config option before adding it

plugins/node/opentelemetry-instrumentation-azure-functions/src/instrumentation.ts
gunzip: Looks similar to the wrapper we're using around http Azure Functions handlers to obtain e2e tracing: https://github.com/gunzip/tracing-test-app/blob/main/src/wrapper.ts#L72

ejizba: Correct me if I'm wrong, but my takeaway is that this is probably fine in the vast majority of cases. Thanks!

hectorhdzg (approved):
LGTM

ejizba:
Closing related to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2259

gunzip:
where can we follow progress about this now that is closed? thanks!

ejizba:
https://github.com/Azure/azure-functions-nodejs-library/issues/245

Issue #1656, opened 2023-08-29
Support new tedious version 16


blumamir:
Tedious current latest major is 16.
Since tedious 16 dropped support for node 14, it makes our CI fail currently.

I attempted to fix that in #1638 in few ways, but they all were dead ends. I am unable to run tests locally as mssql has no docker image version for my Mac Apple chip which means I cannot run tests locally. After a few failed attempts to fix the test via the CI, I decided it was too much for me currently so I documented the task in this issue instead.

david-luna:
@blumamir @mderriey 

I can take this one after https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2144 gets merged :)

kirrg001:
@david-luna That would be nice. If you need any help, let me know.

kirrg001:
@david-luna Could you please let me know if you want to add the support for v18 after the [merge](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2178)? Otherwise I will do it.

kirrg001:
@david-luna 

> @david-luna Could you please let me know if you want to add the support for v18 after the https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2178? Otherwise I will do it.

kirrg001:
https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2266
I will try to add support for v18.

PR #2178, opened 2024-05-06
feat(instr-tedious): add support for v16 and v17

david-luna:

- adds support for newer versions of `tedious`

Closes: #1656

- add newer versions into the instrumentation
- add a version check in the tests to avoid testing with incopatible nodejs versions
- update `tav.yml`


plugins/node/instrumentation-tedious/README.md
```diff
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-tedious
 
 ## Supported Versions
 
-- `>=1.11.0 <=15`
+- `>=1.11.0 <=19`
```
kirrg001, line 20: _issue_: 
I think this needs to be `<=18` or `<19`
Latest version of tedious is 18.2.0

david-luna: right! my bad

david-luna: changed to `<19`

david-luna:
I've been too optimistic here. `tedious@18` came with breaking changes at type level https://github.com/tediousjs/tedious/releases/tag/v18.0.0

Going to adjust the support the the range `>=16 <18` and add support to v18 in a follow-up PR.

kirrg001:
I will try to add v18 support after the merge 👌

plugins/node/instrumentation-tedious/README.md
```diff
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-tedious
 
 ## Supported Versions
 
-- `>=1.11.0 <=15`
+- `>=1.11.0 <=18`
```
kirrg001, line 20:
```suggestion
- `>=1.11.0 <=17`
```

packages/opentelemetry-test-utils/src/test-utils.ts
```diff
@@ -158,6 +159,27 @@ export const getPackageVersion = (packageName: string) => {
...
+  //   ...
+  // }
+  // resolving `packagePath` to `/path/to/opentelemetry-js-contrib/node_modules/tedious/lib/tedious.js`
+  let packageJsonDir = path.dirname(packagePath);
+  let packageJsonPath = path.join(path.dirname(packagePath), 'package.json');
+
+  while (
+    !existsSync(packageJsonPath) &&
+    !packageJsonDir.endsWith('node_modules') &&
+    packageJsonDir !== '/'
```
kirrg001, line 179: issue: I guess the check `packageJsonDir !== '/'` is not reliable for all operation systems. You could mabye use `packageJsonDir !== path.resolve(packageJsonDir, '..')` or similar. 

I was wondering why do we need to check at all? 🤔  IMO `!packageJsonDir.endsWith('node_modules')` is enough for the current use case of the function.

kirrg001 (approved):

packages/opentelemetry-test-utils/src/test-utils.ts
david-luna: >  IMO !packageJsonDir.endsWith('node_modules') is enough for the current use case of the function.

I think you're right because `require.resolve` would fail if the string provided does not resolve to a package. So I?m removing the lats check. Thanks!

trentm: An alternative might be this:

```diff
diff --git a/packages/opentelemetry-test-utils/src/test-utils.ts b/packages/opentelemetry-test-utils/src/test-utils.ts
index aae38f16..7c7914d9 100644
--- a/packages/opentelemetry-test-utils/src/test-utils.ts
+++ b/packages/opentelemetry-test-utils/src/test-utils.ts
@@ -28,6 +28,7 @@ import {
   hrTimeToMilliseconds,
   hrTimeToMicroseconds,
 } from '@opentelemetry/core';
+import * as fs from 'fs';
 import * as path from 'path';

 const dockerRunCmds = {
@@ -153,11 +154,12 @@ export const getPackageVersion = (packageName: string) => {
   // "test-all-versions" tests that tend to install conflicting package
   // versions. Prefix the search paths with the cwd to include the workspace
   // dir.
-  const packagePath = require?.resolve(packageName, {
+  const mainPath = require?.resolve(packageName, {
     paths: [path.join(process.cwd(), 'node_modules')].concat(
       require?.main?.paths || []
     ),
   });
-  const packageJsonPath = path.join(path.dirname(packagePath), 'package.json');
-  return require(packageJsonPath).version;
+  const idx = mainPath.lastIndexOf('node_modules');
+  const pjPath = path.join(mainPath.slice(0, idx), 'node_modules', packageName, 'package.json');
+  return JSON.parse(fs.readFileSync(pjPath, 'utf8')).version;
 };
```

1. For the `pjPath` it:
    - takes resolved "main" path, e.g. `/Users/trentm/tm/opentelemetry-js-contrib4/plugins/node/instrumentation-tedious/node_modules/tedious/lib/tedious.js`, 
    - slices off from the last `node_modules`, e.g.: `/Users/trentm/tm/opentelemetry-js-contrib4/plugins/node/instrumentation-tedious`
    - then tacks on the `node_modules/$packageName/package.json`.
2. I've also changed from `require(pjPath)` to using JSON.parse(fs.read). It isn't *needed* here, but it avoids the rare case where a package defines "exports" and those exports don't allow `require`ing its package.json.

The technique for "1." is basically what the module-details-from-path mini-module is doing. Both RITM and IITM use that module to do basically this.

plugins/node/instrumentation-tedious/package.json
```diff
@@ -56,7 +56,8 @@
     "mocha": "7.2.0",
     "nyc": "15.1.0",
     "rimraf": "5.0.5",
-    "tedious": "15.1.3",
+    "semver": "^7.6.0",
+    "tedious": "^17.0.0",
```
trentm, line 60: @pichlermarc Did you have strong preference (or particular reasons) for wanting pinned devDep versions? Now that we are using package-lock.json I'm curious if this pinning is an artifact of pre-package-lock days.

trentm (commented):
Otherwise, LGTM. I'll re-review after the test-utils.ts change conversation.

packages/opentelemetry-test-utils/src/test-utils.ts
david-luna: thanks @trent, I think you're right in the assumption that a resolved package must have a `node_modules` and getting the last index short circuits the lookup. Also I did not think about 2 and it's a good addition IMO.

Applying this suggestion in next commit :)

plugins/node/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -78,7 +78,7 @@ export class TediousInstrumentation extends InstrumentationBase {
     return [
       new InstrumentationNodeModuleDefinition(
         TediousInstrumentation.COMPONENT,
-        ['>=1.11.0 <=15'],
+        ['>=1.11.0 <18'],
```
trentm, line 81: nit: It would perhaps be nice to have this version string be the same as is used in the README. However, it is totally fine to leave that for now. 

@blumamir was talking about a separate PR to pick on preferred form (`>=${theMinVer} <=${theMaxMajorVer}`  or >=${theMinVer} <${theMaxMajorVer+1}`) and update many/most of the isntrumentations.

trentm (approved):

plugins/node/instrumentation-tedious/package.json
david-luna: @pichlermarc could you put your thoughts here? I also want to know if there is a reason for having pinned versions

plugins/node/instrumentation-tedious/.tav.yml
```diff
@@ -1,4 +1,4 @@
 tedious:
   # 4.0.0 is broken: https://github.com/tediousjs/tedious/commit/4eceb48
-  versions: "1.11.0 || 1.14.0 || 2.7.1 || 3.0.1 || 4.2.0 || ^6.7.0 || 8.3.0 || 9.2.3 || 11.0.9 || 11.2.0 || 11.4.0 || ^11.8.0 || ^12.3.0 || ^13.2.0 || ^14.0.0 || ^15.0.0"
+  versions: "1.11.0 || 1.14.0 || 2.7.1 || 3.0.1 || 4.2.0 || ^6.7.0 || 8.3.0 || 9.2.3 || 11.0.9 || 11.2.0 || 11.4.0 || ^11.8.0 || ^12.3.0 || ^13.2.0 || >=14 <18"
```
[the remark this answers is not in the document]
david-luna: > Using mode: latest-majors we could reduce the testing 17 versions

I think is a good suggestion. It will reduce the time in TAV :)

plugins/node/instrumentation-tedious/package.json
pichlermarc: @trentm @david-luna it is a pre-package lock artifact. 

One benefit of keeping it pinned even after package-lock is added is that is that updates can be handled directly by renovate bot. That means arenovate PR will be scoped to that one package. Otherwise it would be lumped in together with lockfile updates. In such a lockfile PR it may become hard to pinpoint what's wrong.

trentm: > That means arenovate PR will be scoped to that one package.

Sounds like a good reason to me. I've changed over to pinned versions in commit ddde198d.

plugins/node/instrumentation-tedious/src/instrumentation.ts
blumamir: I will rebase my PR after this one merges 👍

trentm:
@kirrg001 Merging this is currently failing on the CLA check failing for the email address used for your commited suggestion. Are you able to sign the CLA?  Or, if you already have, can you look into why the easycla comment above indicates a failure?

kirrg001:
@trentm Done. When I added the code suggestion, it seems GH took my private email.

trentm:
@kirrg001 Hrm, it is still unresolved here.

kirrg001:
I did it again. And the "not linked" email is also part of my Github account. Its even my primary at the moment. 
Maybe the problem is that I already have other commits in the repo with a different email (katharina.irrgang@ibm.com) and it can't link **this** commit to my account, because its a different email.

https://confluence.linuxfoundation.org/pages/viewpage.action?pageId=86641302

kirrg001:
I guess the easiest fix is to remove the commit. Its only a tiny change. And then re-add the change with a new commit.
@david-luna

trentm:
> I guess the easiest fix is to remove the commit.

I can look into rebasing-away that commit and force-pushing later.

blumamir (approved):

Issue #2170, opened 2024-05-01
Should we document non-spec attribute names in READMEs?


JamieDanielson:
I wonder if it can be helpful to also enumerate the other, non-spec-compliant attributes that are emitted by this instrumentation, as a reference for users and for them to know what to expect when using this instrumentation.

There seems to be 2 custom attributes at the moment:
```js
export enum AttributeNames {
  CONNECT_TYPE = 'connect.type',
  CONNECT_NAME = 'connect.name',
}
```

Do you think eventually we will need to either remove them or add them to the specs? Or eventually, each instrumentation is allowed to record extra attributes for which we also need some sort of documentation and stability guarantee?

_Originally posted by @blumamir in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2154#discussion_r1582544261_

joaopgrassi:
We didn't have time to discuss it this week, but it's at the top of the agenda for next week (May 6th). I will keep you updated.

JamieDanielson:
This was discussed in today's Semantic Conventions WG meeting. There is some general guidance about stable instrumentation NOT providing telemetry at all outside of semantic conventions. This would mean that we should open PRs to add these specific attributes to semantic conventions for frameworks, along with any other attributes not currently defined in semantic conventions, and have that codified before declaring instrumentations stable (instrumentations are not yet stable).

This is sort of a "newer" problem as we are growing as a project and trying to ensure relative stability of attributes, so there is actually not a ton of precedence in documenting and guaranteeing attributes specific to a framework that doesn't exist in other languages or frameworks. There is also discussion about what attributes should potentially be opt-in vs enabled by default. There will be an issue opened in semantic conventions repo to document the current approaches and try to provide workarounds where deemed appropriate.

PR #2182, opened 2024-05-07
docs: remove unused semconv packages and document

JamieDanielson:

- Updates #2025 
- Updates #1778 

- For each of the following instrumentations, document that no semantic convention attributes are included in the instrumentation, and also uninstall `@opentelemetry/semantic-conventions` package that is unused:
  - `@opentelemetry/instrumentation-lru-memoizer`
  - `@opentelemetry/instrumentation-dns`
  - `@opentelemetry/instrumentation-generic-pool`
  - `@opentelemetry/instrumentation-pino`
  - `@opentelemetry/instrumentation-winston`
  - `@opentelemetry/plugin-react-load`
  - `@opentelemetry/sampler-aws-xray`

Note these are intentionally done in separate commits to make it easier to see each change, and also so I can easily move these into separate PRs if that seems easier to deal with. I combined for now because of the minimal impact overall but very open to feedback.

Also, part of this will need to be addressed in some follow-ups after discussion from #2170 ... namely that some of these are still emitting attributes, just not semantic-convention attributes. This does not address that in any way, only documenting the current state.


david-luna (approved):

trentm:
This could potentially add instr-bunyan. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2195#pullrequestreview-2053975474

trentm:
The instr-bunyan update went in (#2195) without a README reference. @JamieDanielson If you like you could add the "This package does not currently generate" README blurb to instr-bunyan as well here. Else I can in a separate PR.

JamieDanielson:
> The instr-bunyan update went in (#2195) without a README reference. @JamieDanielson If you like you could add the "This package does not currently generate" README blurb to instr-bunyan as well here. Else I can in a separate PR.

OK yeah I'll add that here real quick 👍

PR #2186, opened 2024-05-08
fix: fix timeout in aws resource detector

arunp0:
Current timeout is 1 sec, which is causing EC2 metadata api request timed out

The current Ec2 metadata API timeout is 1 sec, which is causing frequent "EC2 metadata API request timed out" issues.

Increased the metadata API timeout to 5 sec.


detectors/node/opentelemetry-resource-detector-aws/test/detectors/AwsEc2Detector.test.ts
```diff
@@ -109,7 +109,8 @@ describe('awsEc2Detector', () => {
       scope.done();
     });
 
-    it('should throw when timed out', async () => {
+    it('should throw when timed out', function (done) {
+      this.timeout(7000);
```
maryliag, line 113: if you're changing the timeout value for the test, you're not really testing the new value you changed to. Don't you want to test with the 5s?

arunp0: Updated the timeouts

blumamir (approved):
LGTM.

I wonder - can it practically take this much time for the endpoint to respond? shouldn't it be a localhost call?

PR #2187, opened 2024-05-08
feat: typeorm instrumentation

seemk:

Adds the [typeorm instrumentation](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-typeorm) created by Aspecto.

Only contains minor changes: upgraded to the latest SDK and semantic conventions package, modified the tests to use `assert`, changed the hook type signature.


plugins/node/instrumentation-typeorm/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import type { Span } from '@opentelemetry/api';
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+
+export enum ExtendedDatabaseAttribute {
+  DB_STATEMENT_PARAMETERS = 'db.statement.parameters',
```
blumamir, line 20: While this isn't a speced name similar to many others, there is still [this rule for semantic attributes](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/) which the existing name violates:

> Names SHOULD NOT coincide with namespaces. For example if service.instance.id is an attribute name then it is no longer valid to have an attribute named service.instance because service.instance is already a namespace

Since we already have `db.statement`, I think it is better to use the opportunity and rename it to `db.typeorm.parameters` (similar to `'db.postgresql.values'`, `'db.mysql.values'`). or some other good name that does not make `db.statement` a namespace.

plugins/node/instrumentation-typeorm/src/typeorm.ts
```diff
@@ -0,0 +1,471 @@
...
+      };
+    };
+  }
+
+  private getOperationName(statement: string) {
+    let operation = 'raw query';
+    if (typeof statement === 'string') {
+      statement = statement.trim();
+      try {
+        operation = statement.split(' ')[0].toUpperCase();
```
blumamir, line 342: [The semantic conventions for database](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#common-attributes) says:

> [2]: When setting this to an SQL keyword, it is not recommended to attempt any client-side parsing of db.statement just to get this property, but it should be set if the operation name is provided by the library being instrumented

Since this code does not cover many valid cases, can introduce performance hit, and the semantic conventions specifically advise aginst it, I would consider removing it from the instrumentation at the moment, or introducing a config option to opt-in to it.

If we end up keeping it, we can consider refactoring it for better performance - so that we don't need to parse the entire string and create an array just to get the first word:
```
let firstSpaceIndex = statement.indexOf(' ');
operation = firstSpaceIndex === -1 ? statement.toUpperCase() : statement.substring(0, firstSpaceIndex).toUpperCase();
```

plugins/node/instrumentation-typeorm/README.md
```diff
@@ -0,0 +1,86 @@
...
+| `collectParameters` | boolean | set to `true` if you want to capture the parameter values for parameterized SQL queries (**may leak sensitive information**)
+
+## Semantic Conventions
+
+This package uses `@opentelemetry/semantic-conventions` version `1.24+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+
+Attributes collected:
+
+  SEMATTRS_NET_PEER_NAME,
+  SEMATTRS_NET_PEER_PORT,
```
blumamir, line 59: Should these be here?

blumamir (commented):
Thank you for working on this 🙏🎉

Left few comments and suggestions

The original instrumentation had a [test-all-versions](https://github.com/aspecto-io/opentelemetry-ext-js/blob/master/packages/instrumentation-typeorm/.tav.yml) setup. I wonder if it was left out on purpose or just overlooked.

plugins/node/instrumentation-typeorm/README.md
seemk: Nope, removed them, thanks!

plugins/node/instrumentation-typeorm/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+   * Setting the `suppressInternalInstrumentation` config value to `true` will
+   * cause the instrumentation to suppress instrumentation of underlying operations.
+   */
+  suppressInternalInstrumentation?: boolean;
+  /** Some methods such as `getManyAndCount` can generate internally multiple spans.
+   * To instrument those set this to `true`
+   */
+  enableInternalInstrumentation?: boolean;
+  /** set to `true` if you want to capture the parameter values for parameterized SQL queries (**may leak sensitive information**) */
+  collectParameters?: boolean;
```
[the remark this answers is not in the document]
seemk: Renamed it

plugins/node/instrumentation-typeorm/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.3.0"
+  },
+  "devDependencies": {
+    "@opentelemetry/api": "^1.3.0",
+    "@opentelemetry/contrib-test-utils": "^0.39.0",
+    "@opentelemetry/sdk-trace-base": "^1.24.0",
+    "@types/mocha": "7.0.2",
+    "@types/node": "18.6.5",
+    "typeorm": "^0.3.17",
```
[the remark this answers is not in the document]
seemk: Updated the versions

seemk:
> Thank you for working on this 🙏🎉
> 
> Left few comments and suggestions
> 
> The original instrumentation had a [test-all-versions](https://github.com/aspecto-io/opentelemetry-ext-js/blob/master/packages/instrumentation-typeorm/.tav.yml) setup. I wonder if it was left out on purpose or just overlooked.

Overlooked, going to re-add it, thanks!

plugins/node/instrumentation-typeorm/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Installation
+
+```bash
+npm install --save @opentelemetry/instrumentation-typeorm
+```
+
+### Supported versions
+
+- `>0.2.28`
```
blumamir, line 20:
```suggestion
- `>=0.2.29 <1`
```

plugins/node/instrumentation-typeorm/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@opentelemetry/instrumentation-typeorm",
+  "version": "0.0.1",
+  "description": "OpenTelemetry instrumentation for `typeorm` database data-wrapper ORM",
```
blumamir, line 4:
```suggestion
  "description": "OpenTelemetry instrumentation for `typeorm` database data-mapper ORM",
```

blumamir (commented):
Thanks for addressing the comments. There are still few threads from previous review I think would be best to address now as this is already a breaking change from the original instrumentation and doing it later will be yet another breaking change for users.

If you prefer that we leave it as is, and address these less trivial cases sometime else, that is also ok with me.

plugins/node/instrumentation-typeorm/src/types.ts
seemk: Changed it, thanks

plugins/node/instrumentation-typeorm/src/typeorm.ts
seemk: Removed it completely for now

seemk:
Briefly messed up a commit and GitHub actions added various tags, shouldn't be a problem

seemk:
Should've addressed all the feedback nwo

seemk:
@blumamir Anything else to resolve here?

CrazyBaran:
@blumamir Up, could you review I could also help with that as I am waiting for proper typeorm instrumentation.

pichlermarc (commented):
Sorry for not being able to drive this for so long, adding new packages never becomes a priority as all existing packages (core and contrib) are always taking up a lot of time which led to [this section in the CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#new-instrumentation) about new instrumentation.

Most questions from my side are preliminary ones around maintenance and support for old versions. I'm fine with adding this without going though the formal process outlined in CONTRIBUTING.md as this PR was opened before we added these guidelines.

However, I still see two owners as a hard-requirement for any contrib package to be viable long-term.

(note: I did not do a technical review and a review wrt. semconv yet, to follow once we have the two owners)

.github/component_owners.yml
```diff
@@ -69,6 +69,8 @@ components:
     - mottibec
   plugins/node/instrumentation-tedious: []
     # Unmaintained
+  plugins/node/instrumentation-typeorm:
+    - seemk
```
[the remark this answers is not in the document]
weyert: I am happy to join as a component owner. I am currently maintaining an internal fork at work.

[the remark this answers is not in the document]
seemk: @weyert Thanks! Added you as a component owner.

plugins/node/instrumentation-typeorm/src/typeorm.ts
```diff
@@ -0,0 +1,455 @@
...
+            this._unwrap(moduleExports.SelectQueryBuilder.prototype, method);
+          }
+        });
+        return moduleExports;
+      }
+    );
+
+    const connection = new InstrumentationNodeModuleFile(
+      'typeorm/connection/Connection.js',
+      ['>0.2.28 <0.3.0'],
```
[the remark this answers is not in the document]
seemk: I don't mind, dropped support for <0.3

pichlermarc (commented):
Glad to see there's two owners now, thank you @weyert for stepping up. 🙂 

I asked @blumamir to take a look again as well as he was involved with this instrumentation in the past and therefore has some unique insight on this package's history.

With the current state of semconv for db clients (release candidate) I see an opportunity to save some work by updating semconv attribute names to the latest before releasing the first version of this package. I think it's worth doing now as the dual-emit thing can be painful to do later.

plugins/node/instrumentation-typeorm/src/typeorm.ts
```diff
@@ -0,0 +1,432 @@
...
+import { suppressTracing } from '@opentelemetry/core';
+import {
+  ATTR_DB_NAME,
+  ATTR_DB_OPERATION,
+  ATTR_DB_SQL_TABLE,
+  ATTR_DB_STATEMENT,
+  ATTR_DB_SYSTEM,
+  ATTR_DB_USER,
+  ATTR_NET_PEER_NAME,
+  ATTR_NET_PEER_PORT,
```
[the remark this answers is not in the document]
seemk: Converted to the new semconv. It seems `db.user` was completely dropped, so I removed this attribute from the instrumentation.

plugins/node/instrumentation-typeorm/NOTICE
```diff
@@ -0,0 +1,9 @@
+[Based on the instrumentation written by Aspecto](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-typeorm).
+
+The library contains the following changes compared to the original:
+* Removed `moduleVersionAttributeName` configuration option.
+* Changed the function signature of `responseHook`.
+* Removed dependency on `is-promise` package.
+* Removed usage of SemanticAttributes in favor of explicit values.
+* Removed parsing of db operation from raw queries.
+* Changed db.database.parameters to db.typeorm.parameters.
```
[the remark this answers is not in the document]
seemk: Added a note about it

plugins/node/instrumentation-typeorm/src/typeorm.ts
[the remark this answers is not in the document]
seemk: @weyert Feel free to take a look at the new semconv changes

plugins/node/instrumentation-typeorm/test/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const rawQueryOptions: any = {
+  type: 'sqlite',
+  database: ':memory:',
+  dropSchema: true,
+  synchronize: true,
+  entities: [User],
+  name: 'rawQuery',
+};
```
pichlermarc, line 53: I think this can be simplified since we now only supprot >=0.3.0

plugins/node/instrumentation-typeorm/src/typeorm.ts
```diff
@@ -0,0 +1,428 @@
...
+          }
+        } catch {
+          /* */
+        }
+
+        Object.entries(attributes).forEach(([key, value]) => {
+          if (value === undefined) delete attributes[key];
+        });
+
+        const span: Span = self.tracer.startSpan(`TypeORM ${opName}`, {
```
pichlermarc, line 226: Hmm, I think this may not match the current semconv requirements for DB client span names:
https://opentelemetry.io/docs/specs/semconv/database/database-spans/#name

pichlermarc:
cc @blumamir - I would appreciate you having another look if you find some time 🙂

plugins/node/instrumentation-typeorm/test/utils.ts
seemk: Removed rawQueryOptions and the comment. Moved the tests from deprecated `createConnection` to `DataSource`.

seemk:
@pichlermarc Added @t2t2 and @mhennoch as additional component owners, getting more people to help out with JS contrib, especially for the upcoming and/or ported instrumentations.

pichlermarc (approved):
Looks good, thank you for working on this. 🙌
 
@t2t2 @mhennoch - thank you for joining as owners (I assume @seemk has reached out offline), please indicate that you're okay with this by approving the PR. 🙂 

I'll leave this open for @blumamir to take a look until Monday - if there are no more comments by then I will merge this PR. 🙂

t2t2 (approved):

mhennoch (approved):

pichlermarc:
@seemk conflicts need to be resovled one last time then this is good to merge 🙂

seemk:
@pichlermarc Thanks, there's one job that failed due to `npm ERR! code ECONNRESET`, can you trigger a rerun of this?

seemk:
@pichlermarc Nvm, pushed an empty commit to trigger it

PR #2188, opened 2024-05-09
feat(instrumentation-document-load): Add ability to disable span events

pkanal:
Adds the option to turn off sending span events generated by the document load instrumentation.

This is useful because this instrumentation can generate 8-10 span events per span which can become noisy. There is precedent for this API in the [`instrumentation-fetch`](https://github.com/open-telemetry/opentelemetry-js/blob/3a426e8c32b9a3db691ba4225dadc52de62e660f/experimental/packages/opentelemetry-instrumentation-fetch/src/fetch.ts#L69) and [`instrumentation-xml-http`](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts#L78) packages so I decided to keep it consistent with that API.

- Added a `ignoreNetworkEvents` optional option. If set to `true`, does not send network span events on any spans.
- Added a `ignorePerformanceEvents` optional option. If set to `true`, does not send performance events span events.

## Open questions
- Should these be configurable per type of span? (e.g. have the ability to turn off network spans only for `documentLoad` events but not `resourceFetch` events)


trentm (approved):

plugins/web/opentelemetry-instrumentation-document-load/src/types.ts
```diff
@@ -35,4 +35,6 @@ export interface DocumentLoadInstrumentationConfig
     documentFetch?: DocumentLoadCustomAttributeFunction;
     resourceFetch?: ResourceFetchCustomAttributeFunction;
   };
+  ignoreNetworkEvents?: boolean;
+  ignorePerformanceEvents?: boolean;
```
blumamir, line 39: Please add a section in this package README.md to document these options for package users, like the one for [`applyCustomAttributesOnSpan`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-instrumentation-document-load/README.md#optional--add-custom-attributes-to-spans-if-needed) We can also include a comment here to help users understand what each option does and when and why to use it. Similar to the comment few lines above the `applyCustomAttributesOnSpan?:` option.

blumamir (commented):
LGTM, added one minor comment about documentation of the new options 

I am not sure how a real span looks like and if it can be helpful to enumarte the events that might be recorded/ignored by each option, thus helping users decide if they want each one on or off. But this can also be addressed separately later if there is motivation.

MSNev (approved):

plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
```diff
@@ -135,21 +137,30 @@ export class DocumentLoadInstrumentation extends InstrumentationBase {
...
+          entries
+        );
+        addSpanNetworkEvent(rootSpan, PTN.DOM_COMPLETE, entries);
+        addSpanNetworkEvent(rootSpan, PTN.LOAD_EVENT_START, entries);
+        addSpanNetworkEvent(rootSpan, PTN.LOAD_EVENT_END, entries);
+      }
+
+      if (!this._getConfig().ignorePerformanceEvents) {
+        addSpanPerformancePaintEvents(rootSpan);
+      }
```
scheler, line 162: There is a way to do better and avoid calling the Performance API altogether. The function `getPerformanceNavigationEntries` and in utils.js could return an empty `entries` map when this flag is turned off.

scheler (changes requested):

plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
scheler: Sorry I put this comment in the wrong place. I meant it can be avoided even here  - https://github.com/honeycombio/opentelemetry-js-contrib/blob/feat/user-interaction-instrumentation-span-events/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts#L103

JamieDanielson:
It looks like the changes suggested have been implemented. @pkanal is there anything else you wanted to add here? If not I think this is good to go 🚀

plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
pkanal: It looks like the `entries` variable is actively used by the [`_startSpan` function to determine the start time](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts#L230) so I'm inclined to leave this as is.

scheler (approved):

PR #2191, opened 2024-05-10
chore(instrumentation-ioredis): use exported strings for attributes

maryliag:

- Part Of #2025 

On package `opentelemetry-instrumentation-ioredis`:


plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts
```diff
@@ -121,11 +125,8 @@ export class IORedisInstrumentation extends InstrumentationBase {
       const span = instrumentation.tracer.startSpan(cmd.name, {
         kind: SpanKind.CLIENT,
         attributes: {
-          [SemanticAttributes.DB_SYSTEM]: DbSystemValues.REDIS,
-          [SemanticAttributes.DB_STATEMENT]: dbStatementSerializer(
-            cmd.name,
-            cmd.args
-          ),
+          [SEMATTRS_DB_SYSTEM]: DbSystemValues.REDIS,
```
trentm, line 128: This `DbSystemValues` usage should be udpated as well, I think.

maryliag: Updated

plugins/node/opentelemetry-instrumentation-ioredis/package.json
```diff
@@ -69,7 +69,7 @@
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.51.0",
     "@opentelemetry/redis-common": "^0.36.2",
-    "@opentelemetry/semantic-conventions": "^1.0.0"
+    "@opentelemetry/semantic-conventions": "^1.23.0"
```
trentm, line 72: nit: This should be `^1.22.0` right?

```suggestion
    "@opentelemetry/semantic-conventions": "^1.22.0"
```

trentm (approved):

plugins/node/opentelemetry-instrumentation-ioredis/package.json
maryliag: I just though if we are making an update, might as well update to the latest, so I did this on purpose. The message on readme says about 1.22+ to accommodate when this would happen.

PR #2194, opened 2024-05-10
chore(instrumentation-net): use strings exported for attributes

maryliag:

- Part Of #2025 

On package `opentelemetry-instrumentation-net`:


trentm (commented):
Was the intent to add the `## Semantic Conventions` section with the attributes table to the README in this PR (and in some of the other PRs in this round) or to do that separately?

maryliag:
updated the readme (on this PR and the other ones too)

plugins/node/opentelemetry-instrumentation-net/package.json
```diff
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.51.0",
-    "@opentelemetry/semantic-conventions": "^1.0.0"
+    "@opentelemetry/semantic-conventions": "^1.23.0"
```
trentm, line 62: nit:

```suggestion
    "@opentelemetry/semantic-conventions": "^1.22.0"
```

Thought perhaps I'm wrong and 1.23 is fine.

trentm (approved):

plugins/node/opentelemetry-instrumentation-net/package.json
maryliag: Same comment as the other PR, I decided to get the latest version, since it also works

PR #2195, opened 2024-05-10
chore(instrumentation-bunyan): use strings exported for attributes

maryliag:

- Part Of #2025 

On package `opentelemetry-instrumentation-bunyan:


trentm (approved):

plugins/node/opentelemetry-instrumentation-bunyan/README.md
```diff
@@ -118,6 +118,10 @@ const logger = bunyan.createLogger({
 });
 ```
 
+## Semantic Conventions
+
+
```
trentm, line 124: I kinda think this should be dropped. The only use a semconv in this package is in the *tests* to use the `SERVICE_NAME` constant.

trentm: See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2182#pullrequestreview-2052361574

maryliag: done

PR #2196, opened 2024-05-11
fix!: standardize supported versions and set upper bound limit

blumamir:
This PR is not breaking any existing behavior, it only aligns everywhere to use a consistent versioning syntax and documentation, as well as capping supportedVersions for any existing instrumentation

- make sure there is upper limit on all supported versions so that they do not attempt to patch a new major version
- align the range to a common format. It seems many instrumentation where using `>=x.y.z <w` writing so I used it
- README.md: added where missing, moved to always be after `Installation` section, added package name and npm  link before version range to give more context.
- added section to GUIDELINES.md
- replace all the use of astrix `*` for unordered list style in makrdown to use dash `-`, it causes errors in markdown linter after introducing new sections.

related: https://github.com/open-telemetry/opentelemetry-js/pull/4693
related: https://github.com/open-telemetry/opentelemetry-js/pull/4696


plugins/node/instrumentation-undici/README.md
```diff
@@ -16,7 +16,7 @@ npm install --save @opentelemetry/instrumentation-undici
 
 ## Supported Versions
 
-- `undici@>=5.12.0`
+[`undici`](https://www.npmjs.com/package/undici) version `>=5.12.0`
```
blumamir, line 19: for `unidici` the instrumentation seems to not target a specific version range, but using `diagnosticChannel`. therefore I don't think we can say anything about the upper limit we support.

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
 The cases above are not covered by the base class and offer additional context to the user troubleshooting an issue with the instrumentation.
+
+## Supported Versions
+
+### Usage
+
+Instrumentation can specify the supported version to patch for each instrumented module and instrumented module file by specifying it in the `InstrumentationNodeModuleFile` and `InstrumentationNodeModuleDefinition` like so:
+
+```js
+const supporterVersions = ['>=1.2.3 <3'];
```
trentm, line 208:
```suggestion
const supportedVersions = ['>=1.2.3 <3'];
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+  protected init() {
+
+    const someModuleFile = new InstrumentationNodeModuleFile(
+      'foo/lib/some-file.js',
+      supporterVersions,
```
trentm, line 214:
```suggestion
      supportedVersions,
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+      myFilePatch,
+      myFileUnpatch,
+    );
+
+    const module = new InstrumentationNodeModuleDefinition(
+      'foo',
+      supporterVersions,
```
trentm, line 221:
```suggestion
      supportedVersions,
```

plugins/node/opentelemetry-instrumentation-redis-4/README.md
```diff
@@ -17,8 +17,7 @@ npm install --save @opentelemetry/instrumentation-redis-4
 
 ### Supported Versions
 
-This package supports `redis@^4.0.0`
-For versions `redis@^2.6.0` and `redis@^3.0.0`, please use `@opentelemetry/instrumentation-redis`
```
trentm, line 21: This extra info about redis v2 and v3 was useful, no?

propagators/README.md
```diff
@@ -22,8 +22,8 @@ api.propagation.setGlobalPropagator(new GrpcCensusPropagator());
 
 ## Useful links
 
-* For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
-* For more about OpenTelemetry JavaScript: <https://github.com/open-telemetry/opentelemetry-js>
+- For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
+- For more about OpenTelemetry JavaScript: <https://github.com/open-telemetry/opentelemetry-js>
```
trentm, line 26: nit: It would clarify this PR to not have the bullet changes included. Not a big problem, however.

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -74,6 +74,8 @@ import {
 } from './utils';
 import { VERSION } from './version';
 
+const supporterVersions = ['>=0.5.5 <1'];
```
trentm, line 77:
```suggestion
const supportedVersions = ['>=0.5.5 <1'];
```

Also the usages below.

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+      myModulePatch,
+      myModuleUnpatch,
+      [someModuleFile]
+    );
+    return module
+  }
+```
+
```
trentm, line 229: The following feels like a lot of prose for the topic. I'm worried that if GUIDELINES get too long then they don't get read. Would you consider a (possibly much) shorter proposed alternative for this section?

plugins/node/opentelemetry-instrumentation-redis-4/README.md
blumamir: I was using this info to create automatic READMEs for the "@opentelemetry/auto-instrumentation-node" package, and it causes the text to be confusing, as this information is not relevant for a list of instrumentations listing both 
`redis` and `redis-4` one next to each others. the information about supported versions is covered in #2196.

My future plan is to merge `redis-4` into `redis`. I was the one to introduce it back then, as it felt right at the moment, but I think it harms usability and I want to revert it and use one package for all redis versions.

propagators/README.md
blumamir: The issue is that I created the `Supported Versions` section in the READMEs, and was using common format of dash. and then the linter said I was using mixed unordered list format in the same markdown.

I can split it up into a separate PR if you prefer, was trying to save time by just doing it all at once after spotted.

GUIDELINES.md
blumamir: Sure I am open to making it shorter. 

Wondering if the guideline file is meant to be read fully by contributors, or more as a reference for a specific topic that documents our decisions and reasoning so it can be used for review, changes etc.

Feel free to suggest new shorter text 🙏

propagators/README.md
blumamir: I extracted this change into #2205 

Will rebase this PR once it's merged to focus it only on supported versions documentation changes and nothing else

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+Instrumentation usually targets some user-facing package that is used by the end user.
+However, to achieve the instrumentation goals, it may need to patch one or more modules of the same or different packages, with potentially different patch implementations depending on the patched version.
+
+We will refer to them as "instrumented package" which is the user-facing package, and "patched package" which is the package/s that are patched by the instrumentation.
+
+The contrib repo has a few examples of different variations of common cases
+
+#### Single Module
+
+This is the easy and common case, where the instrumented package is also also the patched package, thus the supported versions align.
```
JamieDanielson, line 241:
```suggestion
This is the typical case, where the instrumented package is also the patched package, thus the supported versions align.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+For versions that are a closed range, instrumentations should prefer to specify the supported versions of the instrumented package as `>=x.y.z <w` to promote consistency and readability across the code-base.
+
+If Instrumentations supports just one major version of the instrumented package, it can specify the version range as `^x.y.z` or `^x` which are equivalent but more readable.
+
+Instrumentations for nodejs internal modules can specify version range of `['*']`.
+
+Instrumentation should use an upper and lower bounds for the version ranges it uses for patches. This is to ensure that any new major versions of the instrumented package are not automatically patched by the instrumentation, which could lead to unexpected behavior.
+
+New major versions should be reviewed and tested by before being added to the supported versions list.
```
JamieDanielson, line 269:
```suggestion
New major versions should be reviewed and tested before being added to the supported versions list.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+Instrumentation should have a "## Supported Versions" section in the README file that lists the supported versions range of the instrumented package - the user-facing package which makes most sense to human consumer.
+
+This range should hide any internal implementation details like the use of internal modules, different patch logic for different versions, etc. It should focus on the relevance to the human consumer.
+
+For cases where the range is not capped by an upper limit, the README should specify a version with an open upper limit `>=4.0.0`. This should be revisited when new versions are released.
+
+### Add New Supported Versions
+
+When a new major version of the instrumented package is released, renovate bot will open a PR in contrib which is an easy was to become aware of it. The instrumentation maintainer should review the new version and check compatibility with existing code. It can then be added to the supported versions list to be released in the next version of the instrumentation.
```
JamieDanielson, line 281:
```suggestion
When a new major version of the instrumented package is released, renovate bot will open a PR in contrib which helps maintainers to become aware of it. The instrumentation maintainer should review the new version and check compatibility with existing code. It can then be added to the supported versions list to be released in the next version of the instrumentation.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+Checklist for adding a new version to the supported versions list:
+
+- [ ] Review which functions are patched by the instrumentation and if they were changed in the new version that need support in code.
+- [ ] Check for breaking changes in the new version that could affect the instrumentation.
+- [ ] Test the instrumentation with the new version to ensure it works as expected.
+- [ ] Update the supported versions list in the instrumentation code, perhaps with different patches and additional `InstrumentationNodeModuleDefinition` that target the new version.
```
JamieDanielson, line 288:
```suggestion
- [ ] Update the supported versions list in the instrumentation code, perhaps with different patches and additional `InstrumentationNodeModuleDefinition`s that target the new version.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+- [ ] Update the README file to reflect the support for new versions.
+- [ ] For instrumentations that uses test-all-verions `.tav.yaml`, add the new version to the list of versions to test.
```
JamieDanielson, line 290:
```suggestion
- [ ] For instrumentations that use test-all-versions `.tav.yaml`, add the new version to the list of versions to test.
```

GUIDELINES.md
JamieDanielson: I am torn on this - on one hand, all of this detail is really useful context. On the other, it indeed is a lot of text. What do you think about leading with the common example start to finish up front, then a lower section for "Variations" which can be linked from the common example for folks who know they have something different. A "tldr" does tend to be useful when there is a lot of content.

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+#### Single Module
+
+This is the easy and common case, where the instrumented package is also also the patched package, thus the supported versions align.
+
+#### Different Modules
+
+In this case, the instrumentation patches some internal modules (like with `redis-4` instrumentations, or `aws-sdk`). In this case the supported versions range is set on the patched packages, but the supported version upper range of the instrumented package is unknown as it depends on future versioning of the instrumented package dependencies.
+
+#### Nodejs Core Modules
```
JamieDanielson, line 247:
```suggestion
#### Node.js Core Modules
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+In this case, the instrumentation patches a nodejs internal module, which is not versioned (carry the same version of the node runtime). The supported versions range is set to `['*']`, and the supported versions is the same as the support of OpenTelemetry for Nodejs versions. The README should specify the supported versions of NodeJS and align with the "engines" field in `package.json`.
```
JamieDanielson, line 249:
```suggestion
In this case, the instrumentation patches a Node.js internal module, which is not versioned because it carries the same version of the Node.js runtime. The supported versions range is set to `['*']`, and the supported versions will match OpenTelemetry for Node.js supported versions. The README should specify the supported versions of Node.js and align with the "engines" field in `package.json`.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+#### Multiple Modules
+
+One instrumentation package can potentially instrument multiple modules of different packages and version ranges. This makes sense if they are related, for example: `pg` and `pg-pool`, `aws-sdk` dozens of client packages etc. In this case, the instrumentation should specify the supported versions range for each module (if possible), and the README should list them.
```
JamieDanielson, line 253:
```suggestion
One instrumentation package can potentially instrument multiple modules of different packages and version ranges. This makes sense if they are closely related, such as `pg` and `pg-pool`. In this case, the instrumentation should specify the supported versions range for each module (if possible), and the README should list them.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+#### Different Patch Logic
+
+In some cases, instrumentations does not use the moduleExports in order to patch, it can hook up to nodejs diagnostics channel, patch globals (like browser instrumentations that patches the `window`), patch arbitrary lambda function handler, etc. In this cases, the use of supported versions can sometimes be more flexible, and the README should specify useful versioning information.
```
JamieDanielson, line 257:
```suggestion
In some cases, instrumentation does not use the moduleExports to patch, and instead hooks up to the Node.js diagnostics channel, or patches globals (like `window` being patched in browser instrumentations, or patches arbitrary lambda function handlers, etc. In these cases, the use of supported versions can sometimes be more flexible, and the README should specify useful versioning information.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+### Range Specification
+
+For versions that are a closed range, instrumentations should prefer to specify the supported versions of the instrumented package as `>=x.y.z <w` to promote consistency and readability across the code-base.
+
+If Instrumentations supports just one major version of the instrumented package, it can specify the version range as `^x.y.z` or `^x` which are equivalent but more readable.
```
JamieDanielson, line 263:
```suggestion
If an instrumentation supports just one major version of the instrumented package, it can specify the version range as `^x.y.z` or `^x`, which are equivalent but more readable.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+Instrumentations for nodejs internal modules can specify version range of `['*']`.
```
JamieDanielson, line 265:
```suggestion
Instrumentations for Node.js internal modules can specify version range of `['*']`.
```

GUIDELINES.md
```diff
@@ -197,3 +197,94 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+Instrumentation should use an upper and lower bounds for the version ranges it uses for patches. This is to ensure that any new major versions of the instrumented package are not automatically patched by the instrumentation, which could lead to unexpected behavior.
+
+New major versions should be reviewed and tested by before being added to the supported versions list.
+
+### Documentation
+
+Instrumentation should have a "## Supported Versions" section in the README file that lists the supported versions range of the instrumented package - the user-facing package which makes most sense to human consumer.
```
JamieDanielson, line 273:
```suggestion
Instrumentation should have a "## Supported Versions" section in the README file that lists the supported versions range of the instrumented package.
```

propagators/README.md
JamieDanielson: Thank you for separating that change out!

plugins/node/instrumentation-lru-memoizer/README.md
```diff
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-lru-memoizer
 
 ## Supported Versions
 
-- `>=1.3 <3`
+- [`lru-memorizer`](https://github.com/jfromaniello/lru-memoizer) versions `>=1.3 <3`
```
JamieDanielson, line 20:
```suggestion
- [`lru-memorizer`](https://www.npmjs.com/package/lru-memoizer) versions `>=1.3 <3`
```

plugins/node/opentelemetry-instrumentation-connect/README.md
```diff
@@ -23,7 +23,7 @@ npm install --save @opentelemetry/instrumentation-http @opentelemetry/instrument
 
 ### Supported Versions
 
-- `^3.0.0`
+- [`connect`](https://github.com/senchalabs/connect) versions `^3.0.0`
```
JamieDanielson, line 26:
```suggestion
- [`connect`](https://www.npmjs.com/package/connect) versions `^3.0.0`
```

plugins/node/opentelemetry-instrumentation-router/README.md
```diff
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-router
 
 ### Supported Versions
 
-- `>=1.0.0`
+- [`router`](https://www.npmjs.com/package/route) versions `>=1.0.0 <2`
```
JamieDanielson, line 20:
```suggestion
- [`router`](https://www.npmjs.com/package/router) versions `>=1.0.0 <2`
```

JamieDanielson (commented):
Another pass-through of versions review, left a few suggestions. I notice some instrumentations reference the github link in the description and I wonder if we should update all such occurrences to be the npm link instead? For now at least this is good to have supported versions link to the npm package.

`instrumentation-runtime-node` package has a supported versions section near the bottom, we may want to move that up above the Example section. Also note it specifically targets >=14.10.0 because it uses private properties... so it sort of breaks this rule ☹️

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -74,6 +74,8 @@ import {
 } from './utils';
 import { VERSION } from './version';
 
+const supportedVersions = ['>=0.5.5 <1'];
```
JamieDanielson, line 77: I love having this pulled out at the top, instead of hardcoded multiple times throughout. I don't know if it's necessary here because you're already going to be dealing with a lot of merge conflicts, but it might be nice to do this across all instrumentations too, so versions are easier to find and update instead of digging through the code.

plugins/node/opentelemetry-instrumentation-knex/README.md
```diff
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-knex
 
 ### Supported Versions
 
-- `>=0.10.0`
+- [`knex`](https://www.npmjs.com/package/knex) versions `>=0.10.0 <4`
```
JamieDanielson, line 20: Should we also remove the constants file for instrumentation-knex and move the supported versions into instrumentation, similar to what was done in instrumentation-router?

plugins/node/opentelemetry-instrumentation-router/README.md
JamieDanielson, line 20:
```suggestion
- [`router`](https://www.npmjs.com/package/router) versions `^1.0.0`
```

JamieDanielson (commented):
Ok I just did a final pass and left a few more suggestions!

plugins/node/opentelemetry-instrumentation-redis-4/README.md
blumamir: Reverted this change. Will address the text in a different PR for merging both redis instrumentation packages

plugins/node/instrumentation-lru-memoizer/README.md
blumamir: Thanks, I was copying this text from the text in the beginning of the README. makes sense to align it to `npm` at least for the new additions

propagators/README.md
blumamir: #2205 is merged and this PR is rebased.

closing this thread

plugins/node/opentelemetry-instrumentation-router/README.md
blumamir: fixed

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: I think so too, support doing this in a follow-up PR

plugins/node/opentelemetry-instrumentation-redis-4/README.md
```diff
@@ -17,8 +17,7 @@ npm install --save @opentelemetry/instrumentation-redis-4
...
+- [`redis`](https://www.npmjs.com/package/redis) versions `>=4.0.0`
```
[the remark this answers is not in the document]
blumamir: The instrumentation patches a different package then the user-facing one.

The patched packages are `@redis/client` and `@node-redis/client`, with version `^1.0.0` but the user-facing package has a different name and version `redis` with version 4.

I think otel users don't care about the internal packages and it's versioning, therefore I addressed `redis` in the README

Since we can't tell if a future major version of `redis` will continue using v1 of `@node-redis/client` or not, there is no way to tell the upper limit of the supported version for the 'redis' package itself

plugins/node/opentelemetry-instrumentation-nestjs-core/src/instrumentation.ts
```diff
@@ -42,12 +42,12 @@ export class Instrumentation extends InstrumentationBase {
   init() {
     const module = new InstrumentationNodeModuleDefinition(
       Instrumentation.COMPONENT,
-      ['>=4.0.0']
+      ['>=4.0.0 <11']
```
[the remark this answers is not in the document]
blumamir: done :)

plugins/node/opentelemetry-instrumentation-router/README.md
blumamir: fixed

plugins/node/opentelemetry-instrumentation-knex/README.md
blumamir: I support a cleanup phase to the code to make all `supportedVersions` consistent and applied few fixes here and there in this PR. But I want to postpone a thorough cleanup to a later PR dedicated to this improvement and keep this one focused on the READMEs.

GUIDELINES.md
blumamir: I made changes to the GUIDELINES. making it shorter, moving stuff around. might not be perfect yet, but please have a second look when you get a chance

blumamir:
@trentm @JamieDanielson  I think I addressed all the comments from your reviews. Since these are a lot of changes, please have a look again. I think EasyCLA is having hard time with the commits done via the `code suggestion` feature of PR comments in the UI. I will squash and merge all the commits after your review so to not lose and existing comments. 

The one last thing I am not sure about is if we should choose caret `^` range when the version range is just one major version. I am split on this, but tend to prefer replacing it with `>=1.0.0 <2` instead of `^1.0.0`. WDYT?

Thanks again for those great comments suggestions and catches.

plugins/node/instrumentation-amqplib/src/amqplib.ts
trentm, line 77: nit: Perhaps this is academic because there isn't an amqplib 1.x release, and I'm guessing there isn't one in the short term. However, this is a "docs:" PR that changes instr-amqplib code. Perhaps mention this change is adding an upper limit for this instr?

plugins/node/opentelemetry-instrumentation-generic-pool/src/instrumentation.ts
```diff
@@ -40,7 +40,7 @@ export default class Instrumentation extends InstrumentationBase {
     return [
       new InstrumentationNodeModuleDefinition(
         MODULE_NAME,
-        ['>=3'],
+        ['^3.0.0'],
```
trentm, line 43: nit: Similar to the amqplib point above. This PR is slightly more than just a docs PR.

plugins/node/opentelemetry-instrumentation-graphql/README.md
```diff
@@ -21,7 +21,7 @@ npm install @opentelemetry/instrumentation-graphql
 
 ### Supported Versions
 
-`^14.0 | ^15.0 | ^16.0`
+- [`graphql`](https://www.npmjs.com/package/graphql) versions `'>=14.0.0 <17'`
```
trentm, line 24:
```suggestion
- [`graphql`](https://www.npmjs.com/package/graphql) versions `>=14.0.0 <17`
```

GUIDELINES.md
```diff
@@ -198,6 +198,84 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+
+If an instrumentation supports just one major version of the instrumented package, it can specify the version range as `^x.y.z` or `^x`, which are equivalent but more readable.
+
+Instrumentation should use an upper and lower bounds for the version ranges it uses for patches. This is to ensure that any new major versions of the instrumented package are not automatically patched by the instrumentation, which could lead to unexpected behavior.
+
+New major versions should be reviewed and tested before being added to the supported versions list.
+
+Specific guidelines for different cases:
+
+- For `Different Modules`, instrumentations can use an upper limit on patched packages but it is unknown which future versions of the instrumented package will continue to use it. Thus it is ok to use an open upper limit `>=1.2.3` for the instrumented package.
```
trentm, line 261:
```suggestion
- For `Different Modules`, instrumentations can use an upper limit on patched packages but it is unknown which future versions of the instrumented package will continue to use it. Thus it is ok to use an open upper limit, for example `>=1.2.3`, for the instrumented package.
```

GUIDELINES.md
```diff
@@ -198,6 +198,84 @@ Instrumentation may add additional patch/unpatch messages for specific functions
...
+When a new major version of the instrumented package is released, renovate bot will open a PR in contrib which helps maintainers to become aware of it. The instrumentation maintainer should review the new version and check compatibility with existing code. It can then be added to the supported versions list to be released in the next version of the instrumentation.
+
+Checklist for adding a new version to the supported versions list:
+
+- [ ] Review which functions are patched by the instrumentation and if they were changed in the new version that need support in code.
+- [ ] Check for breaking changes in the new version that could affect the instrumentation.
+- [ ] Test the instrumentation with the new version to ensure it works as expected.
+- [ ] Update the supported versions list in the instrumentation code, perhaps with different patches and additional `InstrumentationNodeModuleDefinition`s that target the new version.
+- [ ] Update the README file to reflect the support for new versions.
+- [ ] For instrumentations that use test-all-versions `.tav.yaml`, add the new version to the list of versions to test.
```
trentm, line 277:
```suggestion
- [ ] For instrumentations that use test-all-versions `.tav.yml`, add the new version to the list of versions to test.
```

trentm (approved):
Some nits. I think it would be good to not have this labelled as purely a "docs" PR because it does include code changes for a few modules to add an upper limit on the instrumented versions.

plugins/node/instrumentation-amqplib/src/amqplib.ts
blumamir: I allowed myself to do that, because I was looking at the current major and capped it there. While very unlikely that any user will be affected by this, I don't mind documenting it just in case. Do you think the scope should change from `docs` to `fix!`?

trentm: Can the scope be `docs, chore: ...`?  I suppose it could be `fix` as well. I don't think it is realistically a breaking change for any of the instrumentations changed, but I'm not opposed to a very conservative `fix!`.

JamieDanielson:
> The one last thing I am not sure about is if we should choose caret `^` range when the version range is just one major version. I am split on this, but tend to prefer replacing it with `>=1.0.0 <2` instead of `^1.0.0`. WDYT?

I actually prefer `>=1.0.0 <2` over `^1.0.0` and would support using that. It also makes it more consistent with the other larger ranges.

plugins/node/instrumentation-amqplib/src/amqplib.ts
JamieDanielson: FWIW I prefer to err on the side of conservative, and point out _possible_ behavior changes even if they are unlikely. Renaming from `docs` to`chore` could be sufficient here (...depending on the definition of `chore`, which differs across resources as it could be "non-source changes" - disqualifying this - or "routine tasks" - where this fits... but that might be too far off-course to have that convo here 😅 )

trentm:
> I actually prefer `>=1.0.0 <2` over `^1.0.0`

Same.

JamieDanielson (approved):
I have similar nits to what Trent pointed out, aside from those LGTM 👍

blumamir:
Addressed all the comments (I think). @JamieDanielson @trentm can you please have one last look so this can merge? thanks again for the great feedback

trentm (approved):

PR #2201, opened 2024-05-13
chore(propagator-grpc-cencus-binary): archive package

pichlermarc:

- We've not had any changes to this package for a while. `grpc` has been deprecated since 2021. I propose we deprecate/archive the propagator to enable #2169, where this package is causing a build failure. This will also help contributors on macOS (on ARM) where it seems like the package needs to build the native grpc library every time it is pulled as no pre-built version is included.


blumamir (approved):
I support deprecating it.

Regarding moving the files to `archieve`, the one and only project there states that:

> This project has been archived by the maintainers on August 3 2022 because it was causing a failure in the CI and does not have any maintainer.
> Anyone who would like to maintain this project should open a PR to move it back into the lerna project and add themselves to the component owners file.

So I wonder, since we never plan to rejuvenate this package in the future, should we just remove it from the codebase altogether? It can always be picked up from git history if needed, but any reason we should keep the code?

trentm:
> So I wonder, since we never plan to rejuvenate this package in the future, should we just remove it from the codebase altogether?

I was **going to say** that it would be nice to update the npmjs.com page for this package to clarify where the source went, then... but I can't find any published-to-npm versions for that package name:

```
% npm info @opentelemetry/browser-extension-autoinjection
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@opentelemetry%2fbrowser-extension-autoinjection - Not found
npm ERR! 404
npm ERR! 404  '@opentelemetry/browser-extension-autoinjection@*' is not in this registry.
...
```

Anyway, handling the `browser-extension-autoinjection` package is a separate issue.

trentm (approved):
Woot. Thanks, Marc.

Can we also add a deprecation message to the package per https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions ?

trentm:
Some background and motivation for this PR are here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2169#issuecomment-2089162355

pichlermarc:
> I support deprecating it.
> 
> Regarding moving the files to `archieve`, the one and only project there states that:
> 
> > This project has been archived by the maintainers on August 3 2022 because it was causing a failure in the CI and does not have any maintainer.
> > Anyone who would like to maintain this project should open a PR to move it back into the lerna project and add themselves to the component owners file.
> 
> So I wonder, since we never plan to rejuvenate this package in the future, should we just remove it from the codebase altogether? It can always be picked up from git history if needed, but any reason we should keep the code?

I think removing it is also okay. I'd keep it around for a bit and then remove it when there's no activity for one month or so after we mark it as deprecated in npm, we can remove it. I'll open a follow-up issue to track it.

> Can we also add a deprecation message to the package per https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions ?

Yep, I'll do that.

PR #2203, opened 2024-05-13
refactor: use package json auto generated export also for npm package name

blumamir:
This PR is a suggestion to replace this instrumentation base constructor call:

```js
  constructor(config: CucumberInstrumentationConfig = {}) {
    super("@opentelemetry/instrumentation-foo", VERSION, config);
  }
```

With this one:
```js
  constructor(config: CucumberInstrumentationConfig = {}) {
  }
```

Where the auto generated file `version.ts` will also include the package name from `package.json`.

This PR also rename the cryptic `VERSION` symbol to `NPM_PACKAGE_VERSION` which I think is more readable.

I think the new format is better since:
- It makes it clear that the value we use is the package name and not something else and carries semantic meaning.
- The value is guaranteed to match the content of the `package.json`, just like we currently use the `VERSION` from package.json in instrumentation base constructor. no room for mistakes or misunderstandings. 
- The code is better formatted and more consistent


plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
```diff
@@ -28,19 +28,15 @@ import * as http from 'http';
...
 import LayerType from './enums/LayerType';
 
 export default class RouterInstrumentation extends InstrumentationBase {
   constructor(config: InstrumentationConfig = {}) {
-    super(
-      `@opentelemetry/instrumentation-${constants.MODULE_NAME}`,
-      VERSION,
-      config
-    );
+    super(PACKAGE_NAME, PACKAGE_VERSION, config);
```
JamieDanielson, line 39: As part of this change I think we could remove the `MODULE_NAME` constant from the `constants.ts` file. Then we can also rename in the `init()` there is another `constants.MODULE_NAME`.

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
```diff
@@ -37,11 +37,7 @@ import type { RestifyInstrumentationConfig } from './types';
 
 export class RestifyInstrumentation extends InstrumentationBase {
   constructor(config: RestifyInstrumentationConfig = {}) {
-    super(
-      `@opentelemetry/instrumentation-${constants.MODULE_NAME}`,
-      VERSION,
-      config
-    );
+    super(PACKAGE_NAME, PACKAGE_VERSION, config);
```
JamieDanielson, line 40: same as router - As part of this change I think we could remove the `MODULE_NAME` constant from the `constants.ts` file. Then we can also rename in the `init()` there is another `constants.MODULE_NAME`.

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -50,7 +50,7 @@ function defaultShouldPreventSpanCreation() {
  * addEventListener of HTMLElement
  */
 export class UserInteractionInstrumentation extends InstrumentationBase {
-  readonly version = VERSION;
+  readonly version = PACKAGE_VERSION;
```
JamieDanielson, line 53: I think we should also update the `moduleName` here on the next line to be `PACKAGE_NAME`. It gets used in the `enable()` and `disable()` methods.

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```diff
@@ -58,7 +58,7 @@ export class BaseOpenTelemetryComponent extends React.Component {
   static setTracer(name: string, version?: string): void {
     BaseOpenTelemetryComponent._tracer = api.trace.getTracer(
       name,
-      version ? version : VERSION
+      version ? version : PACKAGE_VERSION
```
JamieDanielson, line 61: I think we can also update the line higher up in the class for `readonly component: string = 'react-load';` to instead be `readonly component: string = PACKAGE_NAME;`

packages/opentelemetry-host-metrics/src/BaseMetrics.ts
```diff
@@ -17,7 +17,7 @@
 import * as api from '@opentelemetry/api';
 import * as metrics from '@opentelemetry/sdk-metrics';
 
-import { VERSION } from './version';
+import { PACKAGE_VERSION } from './version';
```
JamieDanielson, line 20: There is also a `const DEFAULT_NAME = '@opentelemetry/host-metrics';` that can be updated to `PACKAGE_NAME`

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -45,8 +45,8 @@ const DEFAULT_CONFIG: types.KnexInstrumentationConfig = {
 export class KnexInstrumentation extends InstrumentationBase {
   constructor(config: types.KnexInstrumentationConfig = {}) {
     super(
-      `@opentelemetry/instrumentation-${constants.MODULE_NAME}`,
-      VERSION,
+      PACKAGE_NAME,
```
JamieDanielson, line 48: same as router - As part of this change I think we could remove the `MODULE_NAME` constant from the `constants.ts` file. Then we can also rename in the `init()` there is another `constants.MODULE_NAME`.

JamieDanielson (commented):
This is another change I'm a fan of as it increases consistency and reduces errors. I've left a few suggestions, though some may fall slightly out of scope (package names used elsewhere in the instrumentation).

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
blumamir: I totally agree, this pattern should be simplified and refactored to a common one across the repo 👍 

the `MODULE_NAME` export in `constants.ts` currently looks like this:

```js
export const MODULE_NAME = 'router';
```

And still being used in instrumentation [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/0bec64e2e9a73735c7cbecd7212f5be6ef575308/plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts#L50). We cannot utilize the new `PACKAGE_NAME` to remove it altogether at the moment.

This PR touches on the topic of "Instrumentation Package" and the `MODULE_NAME` we refer to is the "Instrumented Package". In order not to introduce too many things into one PR which can make it harder to review and merge in, I suggest we will target the instrumented package patterns in a follow-up PR, which might also include more cleanups and aligning across the codebase.

plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts
blumamir: agree. will handle as follow-up PR

plugins/web/opentelemetry-instrumentation-user-interaction/src/instrumentation.ts
blumamir: This suggested change looks right to me, but it will introduce a change to the instrumentation (it's `name`) which is probably ok but I don't want to bundle it into this PR, as it will change the scope from `refactor` to `fix`.

plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
blumamir: same comment as above - I support doing this as a dedicated `fix` PR which only targets the name of the instrumentation (which is an end user public API content).

packages/opentelemetry-host-metrics/src/BaseMetrics.ts
blumamir: good catch 👍  fixed

blumamir:
@JamieDanielson Thanks for another thorough review. 

You justifiably find many opportunities for further improvement, which I choose to keep out of the scope of this PR so it is very specific on one single change.

plugins/node/opentelemetry-instrumentation-router/src/instrumentation.ts
JamieDanielson: Yeah makes sense to have a new PR for the additional cleanup 👍

JamieDanielson (approved):
LGTM. As discussed in comments we can do further cleanup in a follow-up PR to minimize scope.

trentm (approved):

PR #2205, opened 2024-05-14
ci: add md linter rule to only use dash for unordered lists

blumamir:
During working on #2196 as I was adding the "Supported Versions" part to every README.md, the markdown linter warned me that I am using `dash` for unorder list item style in mix with asterisk, see https://github.com/open-telemetry/opentelemetry-js/issues/4698.

This PR extracts the aligning of readmes to use dash from #2196 so that each Pr focuses on one topic.

Added this rule to the CI


blumamir:
FYI, If the style is incorrect, we get this error from the new linter rule in CI:

```
./plugins/node/opentelemetry-instrumentation-cassandra/README.md:69:1 MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
```

trentm (approved):
I'm fine with this. I don't have a strong opinion. I'd defer to some of the maintainers that have been working in this repo longer.  Marc already indicated being in favour at https://github.com/open-telemetry/opentelemetry-js/issues/4698#issuecomment-2106980898

JamieDanielson (approved):
Ah, I love linters! Let the robots help us when it makes sense 😃 .

Similar to Trent's comment, maybe see if there is any strong objection here or on the issue within like a day, but generally I'd say let it merge. It's pretty minor and consistent formatting is really the desired outcome.

PR #2210, opened 2024-05-15
chore: update last deprecated usages of semantic-conventions pkg

trentm:
Refs: #2025

---

This updates the last uses (in READMEs and test files) of the deprecated exports from the semantic-conventions package for #2025 -- with the exception of a usage in "archive/...".


incubator/opentelemetry-sampler-aws-xray/README.md
```diff
@@ -32,19 +29,19 @@ const _tracerConfig = {
...
+        new HttpInstrumentation(),
+        new AwsInstrumentation({
+            suppressInternalInstrumentation: true
+        }),
+    ],
+    spanProcessor: _spanProcessor,
+    traceExporter: _traceExporter,
+});
+
+sdk.configureTracerProvider(_tracerConfig, _spanProcessor);
```
trentm, line 44: Reviewer note: This just drops the `Resource` usage in favour of the `serviceName` config option to `NodeSDK`. It also cleans up indentation. https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2210/files?w=1 (without whitespace) shows a smaller diff.

packages/opentelemetry-test-utils/src/resource-assertions.ts
```diff
@@ -335,26 +335,26 @@ export const assertEmptyResource = (resource: Resource) => {
   assert.strictEqual(Object.keys(resource.attributes).length, 0);
 };
 
+/**
+ * Assert that the `resource` has at least one known attribute with the given
+ * `prefix`. By "known", we mean it is an attribute defined in semconv.
+ */
 const assertHasOneLabel = (prefix: string, resource: Resource): void => {
```
trentm, line 342: Reviewer note: I changed this function usage to take a `prefix` in lowercase, i.e. what the actual semconv string values are.

I also changed the implementation to something that reads clearer for me.
FWIW, the `.reduce(...)` usage in the removed code had a small bug in it: it didn't provide an initial accumulator value, so the assert message wasn't correct.

trentm:
ping @maryliag @JamieDanielson. I'd appreciate a review on this. This is the last of changing over to the new semconv package usage in the contrib repo. (I.e. I think we can close https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2025 after this.)

trentm:
> I had already approved a couple of hours ago

Oh duh, thanks. I hadn't refreshed this page before pinging.

packages/opentelemetry-test-utils/src/resource-assertions.ts
JamieDanielson: 🤔  I'd prefer to do this in a way that doesn't require bring in semconv as * - isn't service name an expected "known" attribute here that could be safely used in the test? This isn't blocking since it's a test-utils package though.

JamieDanielson (approved):
I left one comment about resource-assertions in the test-utils package, but it's not blocking because of the package it's in, and would be okay to revisit in a follow-up if needed. It would be good to get that issue closed.

packages/opentelemetry-test-utils/src/resource-assertions.ts
trentm: @JamieDanielson I'm not sure exactly what you mean. Take the call `assertHasOneLabel('k8s', resource);`, for example. This function needs to see if `resource.attributes` has at least one attribute that starts with `k8s.` and is an attribute value exported by `@opentelemetry/semantic-conventions`. How else could that be done without importing semconv as `*`?

JamieDanielson: Hmm looking at this again I'm not sure what I had in mind 😅  I was thinking that we could scope the imports of attributes to only pull in a few for each test, but it seems like it could unwieldy. Probably best to ignore me on this one.

PR #2212, opened 2024-05-16
feat: Add optional predicate func to baggage span processor

### LINKED ISSUE #2166: Baggage span processor key predicate
Enhancement idea to the Baggage Span Processor to allow configuration of a user-provided key predicate function used to determine whether a baggage entry should be added to newly started spans.

              I think this behavior can be considered not safe for some organizations.

Wondering if you considered as an alternative adding a config option that can filter out attributes for collection. Perhaps some function that receives an attribute name and return if it should be recorder or not?

```ts
export interface BaggageSpanProcessorConfig {
  attributeRecording(baggaeAttributeName: string): boolean;
}
```

and then users can opt-in to collect only attributes they know are safe, or specifically feed a function that returns `true` to record everything?

This can shift the responsibility of what is being collected to the user, allowing safe flexibility in collection without introducing breaking changes in the future.

_Originally posted by @blumamir in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2114#discussion_r1582606506_

MikeGoldsmith:
- Closes #2166 

- Add optional key filter function as constructor parameter to baggage span processor
- Add unit test to verify behaviour
- Update README with example of setting a filter

### DIFF
--- a/packages/baggage-span-processor/README.md
+++ b/packages/baggage-span-processor/README.md
@@ -46,7 +46,7 @@ const sdk = new NodeSDK({
 sdk.start();
 ```
 
-A predicate function that determines what baggage entries are copied can be provided as a contructor paramter.
+A predicate function that determines what baggage entries are copied can be provided as a constructor paramter.
 
 For example, to only copy baggage entries that start with `my-key`:
 
--- a/packages/baggage-span-processor/README.md
+++ b/packages/baggage-span-processor/README.md
@@ -46,6 +46,16 @@ const sdk = new NodeSDK({
 sdk.start();
 ```
 
+A predicate function that determines what baggage entries are copied can be provided as a contructor paramter.
+
+For example, to only copy baggage entries that start with `my-key`:
+
+```javascript
+new BaggageSpanProcessor((key: string) =>
+    key.startsWith('my-key')
+  )
+```
+
 ## Useful links
 
 - For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
--- a/packages/baggage-span-processor/src/baggage-span-processor.ts
+++ b/packages/baggage-span-processor/src/baggage-span-processor.ts
@@ -50,7 +50,7 @@ export class BaggageSpanProcessor implements SpanProcessor {
    * @param keyFilter An optional function that can be used to filter keys
    * added to the span. If the function returns true, the key will be added
    * to the span. If the function returns false, the key will not be added
-   * to the span.
+   * to the span. The default behavior is to add all keys to the span.
    */
   constructor(keyFilter?: (key: string) => boolean) {
     this._keyFilter = keyFilter ?? (() => true);
--- a/packages/baggage-span-processor/src/baggage-span-processor.ts
+++ b/packages/baggage-span-processor/src/baggage-span-processor.ts
@@ -43,6 +43,19 @@ import {
  * values will appear in all outgoing HTTP headers from the application.
  */
 export class BaggageSpanProcessor implements SpanProcessor {
+  private _keyFilter: (key: string) => boolean;
+
+  /**
+   * Constructs a new BaggageSpanProcessor instance.
+   * @param keyFilter An optional function that can be used to filter keys
+   * added to the span. If the function returns true, the key will be added
+   * to the span. If the function returns false, the key will not be added
+   * to the span.
+   */
+  constructor(keyFilter?: (key: string) => boolean) {
+    this._keyFilter = keyFilter ?? (() => true);
+  }
+
   /**
    * Forces to export all finished spans
    */
@@ -59,7 +72,9 @@ export class BaggageSpanProcessor implements SpanProcessor {
   onStart(span: Span, parentContext: Context): void {
     (propagation.getBaggage(parentContext)?.getAllEntries() ?? []).forEach(
       entry => {
-        span.setAttribute(entry[0], entry[1].value);
+        if (this._keyFilter(entry[0])) {
+          span.setAttribute(entry[0], entry[1].value);
+        }
       }
     );
   }
--- a/packages/baggage-span-processor/test/baggage-span-processor.test.ts
+++ b/packages/baggage-span-processor/test/baggage-span-processor.test.ts
@@ -72,3 +72,42 @@ describe('BaggageSpanProcessor', () => {
     await expect(baggageProcessor.shutdown()).resolves.not.toThrow();
   });
 });
+
+describe('BaggageSpanProcessor with custom key filter', () => {
+  const baggageProcessor = new BaggageSpanProcessor((key: string) =>
+    key.startsWith('brand')
+  );
+
+  const bag = propagation.createBaggage({
+    brand: { value: 'samsonite' },
+    color: { value: 'blue' },
+  });
+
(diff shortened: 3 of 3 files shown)


packages/baggage-span-processor/README.md
```diff
@@ -46,6 +46,16 @@ const sdk = new NodeSDK({
 sdk.start();
 ```
 
+A predicate function that determines what baggage entries are copied can be provided as a constructor paramter.
```
MikeGoldsmith, line 49:
```suggestion
A predicate function that determines what baggage entries are copied can be provided as a constructor parameter.
```

packages/baggage-span-processor/src/baggage-span-processor.ts
```diff
@@ -43,6 +43,19 @@ import {
...
+  private _keyFilter: (key: string) => boolean;
+
+  /**
+   * Constructs a new BaggageSpanProcessor instance.
+   * @param keyFilter An optional function that can be used to filter keys
+   * added to the span. If the function returns true, the key will be added
+   * to the span. If the function returns false, the key will not be added
+   * to the span. The default behavior is to add all keys to the span.
+   */
+  constructor(keyFilter?: (key: string) => boolean) {
```
blumamir, line 55: perhaps instead of `keyFilter` we can call it predicate? I am thinking something like: `keyRecordingPredicate` or `baggageKeyPredicate`. The use of predicate already self explains that this is a function that receives a key to filter and returns boolean to include or filter it.

packages/baggage-span-processor/README.md
```diff
@@ -46,6 +46,16 @@ const sdk = new NodeSDK({
 sdk.start();
 ```
 
+A predicate function that determines what baggage entries are copied can be provided as a constructor parameter.
```
blumamir, line 49: In other READMEs in this repo, we usually have a "Component Options" section which explains each options. something like:

```
### Baggage SpanProcessor Options
```

And then describe the option/s one by one, possibly in a markdown table (but maybe overkill for one option).

blumamir, line 49: Suggestion for alternative description:

```suggestion
### `keyFilter` Option // or any other name

A predicate function that determines what baggage entries keys are recorded as span attributes.
The default behavior is to add all baggage keys as span attributes. // or other default behavior if changed
```

blumamir (commented):
I think it is very much needed. Thank you for working on this.
Added few comments

packages/baggage-span-processor/src/baggage-span-processor.ts
```diff
@@ -43,6 +43,19 @@ import {
...
+    this._keyFilter = keyFilter ?? (() => true);
```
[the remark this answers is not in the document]
MikeGoldsmith: I'm not sure. This is an opt-in processor, ie you have to configure it as part of the tracer provider. If it were to default to not copy all keys and the key filter was optional, I think users will very often add the processor and believe it not to be working as no key-value pairs are added to new spans.

I think we should either:
- defaults to copy all baggage keys and a key filter is an optional param (on a config object suggested above)
- key filter is a required parameter - maybe provide a constant that is `AllowAllKeysFilter` that defaults to true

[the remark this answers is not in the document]
blumamir: That makes sense. I am good with either of the options you think is best or aligns best with implementation in other languages. Regardless of what you choose, I would add a notice to the README to make it clear that users are expected to populate the list of keys to record, and/or the default will capture all keys which might leak sensitive information.

I am thinking if this might be included in OpenTelemetry distributions later on, as an enhancement feature available out-of-the-box to use, what could be a good safe way to set it up via env or other ways (so that the easy way to include it is also a safe one). But we can postpone this discussion until it is needed in practice.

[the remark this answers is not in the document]
MikeGoldsmith: I'm having the env var configuration discussion with the Java SIG so we can make it the Java version accessible from the JavaAgent.

I think I'm leaning towards secure by default and have a required constructor param. I'll update to reflect that.

MikeGoldsmith:
@blumamir I've updated the PR to make the predicate a required parameter, and make some of the suggested changes. Please take a look 👍🏻

packages/baggage-span-processor/src/baggage-span-processor.ts
```diff
@@ -43,6 +56,16 @@ import {
  * values will appear in all outgoing HTTP headers from the application.
  */
 export class BaggageSpanProcessor implements SpanProcessor {
+  private _keyPredicate: BaggageKeyPredicate;
+
+  /**
+   * Constructs a new BaggageSpanProcessor instance.
+   * @param keyPredicate A predicate that determines whether a baggage key-value pair should be added to new spans as a span attribute.
+   */
+  constructor(keyPredicate: BaggageKeyPredicate) {
```
blumamir, line 65: I really love the new approach on being explicit about the predicate. Thank you for introducing it.

One other thing that I think is useful, will be to have the `keyPredicate` inside a `BaggageSpanProcessorConfig` object, so we can pass just one object with all the configuration if more configuration options will be introduced in the future.

This way we won't find yourself potentially having a constructor with many arguments, it's easier to introduce more configuration options later on, and it is also the pattern we use in many other places in this codebase.

blumamir (commented):
Thanks again for addressing my comments, this feature looks great 🥇 

Added one last comment on sending a `config` object in the constructor

packages/baggage-span-processor/src/baggage-span-processor.ts
blumamir: As an example, you can see the [BatchSpanProcessorBase](https://github.com/open-telemetry/opentelemetry-js/blob/4e778b1257c8e8b7fbeb33e8eeba22b78ac33dd6/packages/opentelemetry-sdk-trace-base/src/export/BatchSpanProcessorBase.ts#L52) which receives a `config` argument with all the relevant configuration options inside:

```js
  constructor(
    private readonly _exporter: SpanExporter,
    config?: T
  ) {
```

In this case, we don't need the generic and can just use the config type itself, something like:

```js
// types.ts

export type BaggageKeyPredicate = (baggageKey: string) => boolean;

export const ALLOW_ALL_BAGGAGE_KEYS: BaggageKeyPredicate = (
  baggageKey: string
) => true;

export interface BaggageSpanProcessorConfig {
  keyPredicate: BaggageKeyPredicate
}
```

MikeGoldsmith: Thanks for your feedback @blumamir.

My concern with an options interface is that the key predicate is no longer optional, it's a required parameter. Similar to how `BatchSpanProcessorBase` requires a spanExporter parameter then the options interface.

Making the keyPredicate a required arg was to make sure it's super visible that users must provide it. If it's in an options object, I feel it's kinda hidden.

blumamir: ok that makes sense 👍

blumamir (approved):

MikeGoldsmith:
Thanks @blumamir !

JamieDanielson (approved):

PR #2232, opened 2024-05-22
feat(auto-instrumentations-node): add getPropagator to get propagators from environment

pichlermarc:

According to the spec we MUST NOT maintain the aws propagator be part of the core repo, so we also cannot introduce it as a dependency to `sdk-node`. However, we've had `getResourceDetectors()` in `@opentelemetry/auto-instrumentations-node` which also includes all third-party resource-detectors. This PR adds `getPropagator()` which allows you to get a propagator based on the environment variable `OTEL_PROPAGATORS`. It includes third-party ones like `xray` or `xray-lambda` without having to add them as a core feature. 

- adds `getPropagator()` which builds a propagator based on `OTEL_PROPAGATORS`
- adds that propagator automatically to the `@opentelemetry/auto-instrumentations-node/register` export, so it overrides what the SDK would do.


metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -253,3 +262,56 @@ export function getResourceDetectorsFromEnv(): Array<Detector | DetectorSync> {
...
+  if (process.env.OTEL_PROPAGATORS == null) {
+    return new CompositePropagator({
+      propagators: [
+        new W3CTraceContextPropagator(),
+        new W3CBaggagePropagator(),
+      ],
+    });
+  }
+
+  const propagatorsFromEnv = process.env.OTEL_PROPAGATORS?.split(',').map(
```
trentm, line 297: FWIW, `OTEL_PROPAGATORS` parsing used by sdk-node trims each entry to allow `foo, bar` with spaces: https://github.com/open-telemetry/opentelemetry-js/blob/4e778b1257c8e8b7fbeb33e8eeba22b78ac33dd6/packages/opentelemetry-core/src/utils/environment.ts#L303

It also dedupes values: https://github.com/open-telemetry/opentelemetry-js/blob/4e778b1257c8e8b7fbeb33e8eeba22b78ac33dd6/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L240-L243

trentm (commented):
Drive by.

metapackages/auto-instrumentations-node/src/utils.ts
pichlermarc: good point, aligned the behavior in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2232/commits/c67eda1cd17374a7c7e56d4ce97c2c96c3024259 :slightly_smiling_face:

metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -253,3 +262,56 @@ export function getResourceDetectorsFromEnv(): Array<Detector | DetectorSync> {
...
+  ['jaeger', () => new JaegerPropagator()],
+  ['xray', () => new AWSXRayPropagator()],
+  ['ottrace', () => new OTTracePropagator()],
+]);
+
+/**
+ * Get a propagator based on the OTEL_PROPAGATORS env var.
+ */
+export function getPropagator(): TextMapPropagator {
+  if (process.env.OTEL_PROPAGATORS == null) {
```
trentm, line 288: Should empty string be treated as "use the default" as well? (That's what I infer from the language in this section: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#enum-value)

Also the spec mentions handling "none" as a value to say "no propagators": https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

metapackages/auto-instrumentations-node/test/utils.test.ts
```diff
@@ -161,4 +161,36 @@ describe('utils', () => {
...
+        'tracestate',
+      ]);
+    });
+
+    it('should return the selected propagators when multiple are in the list', () => {
+      process.env.OTEL_PROPAGATORS = 'b3,jaeger';
+      assert.deepStrictEqual(getPropagator().fields(), ['b3', 'uber-trace-id']);
+    });
+
+    it('should return no-op propgator if propagator all propagators are unknown', () => {
```
trentm, line 191: nit: 
```suggestion
    it('should return no-op propagator if all propagators are unknown', () => {
```

metapackages/auto-instrumentations-node/src/utils.ts
pichlermarc: >  Should empty string be treated as "use the default" as well? (That's what I infer from the language in this section: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#enum-value)

Yes that's correct, good catch :slightly_smiling_face: I added that to the default case. https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2232/commits/cf3bb0d733fd56c2daa0ec183c62be28b8a34743

> Also the spec mentions handling "none" as a value to say "no propagators": https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration

Also good catch, it would've incorrectly logged that it's unkown before (the Propgator would still have been a no-op CompositePropagator with no sub-propgators, so I think that would've been spec compliant). I added an info log that lets the user know that `none`. Though I'm not sure what to do if there's `none` among other valid values so I've opted to still take the rest of the propagators in that case. My guess is that users would be least surpirsed if they set `tracecontext, none` and the `W3CTraceContextPropagator` would be the one that gets set. https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2232/commits/cf3bb0d733fd56c2daa0ec183c62be28b8a34743

trentm: Some comparison points on "none" handling:
- auto-instrumentations-node handling of `OTEL_NODE_RESOURCE_DETECTORS` handles `foo,none,bar` and "none" https://github.com/open-telemetry/opentelemetry-js-contrib/blob/caf7cb503760d17ac6455dcf712f677ef665f07a/metapackages/auto-instrumentations-node/src/utils.ts#L238-L244
- ditto for sdk-node's handling of that same envvar
- sdk-node's handling of `OTEL_TRACES_EXPORTER` is slightly weird: `OTEL_TRACES_EXPORTER=none,otlp` results in different behaviour to `OTEL_TRACES_EXPORTER=otlp,none` :) https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-sdk-node/src/TracerProviderWithEnvExporter.ts#L101-L126 

Anyway, this is something that could perhaps share some common processing at some point. Doesn't have to be this PR, and perhaps would wait for work related to supporting file-config.

trentm (approved):
Just the `npm run lint:fix` to run first.

JamieDanielson (approved):
Looks great! (just needs lint as per the previous comment)

JamieDanielson:
also we could note that this updates https://github.com/open-telemetry/opentelemetry-js/issues/4494

pichlermarc:
Thanks for the reviews. :bow: 

I'm moving this back to draft as I talked to @martinkuba about some other options to solve this issue. There will likely be a propagator-autoconfigure package that only contains the spec propagators (all from core and including `lambda` and `lambda-xray`) - then we'll be able to move these propagator packages back to contrib. The code used in that package will look very similar to what I proposed here.

pichlermarc:
closing in favor of #2299

PR #2233, opened 2024-05-22
fix: Update Azure Functions Detector Collected Attributes

JacksonWeber:

- This PR updates the fields collected by the Azure Functions resource detector to be in line with Azure Functions' expectations. Fixes improper population of fields like `faas.name` and properly populates `service.name` instead for example. We also now collect the process PID and cloud resource ID. 
- These changes come with updated tests & README.


detectors/node/opentelemetry-resource-detector-azure/src/detectors/AzureFunctionsDetector.ts
```diff
@@ -49,28 +50,23 @@ const AZURE_FUNCTIONS_ATTRIBUTE_ENV_VARS = {
 class AzureFunctionsDetector implements DetectorSync {
   detect(): IResource {
     let attributes = {};
-    const functionName = process.env[WEBSITE_SITE_NAME];
+    const serviceName = process.env[WEBSITE_SITE_NAME];
     const functionVersion = process.env[FUNCTIONS_VERSION];
```
blumamir, line 54: is it used for anything now that the attribute is removed?

detectors/node/opentelemetry-resource-detector-azure/src/detectors/AzureFunctionsDetector.ts
```diff
@@ -49,28 +50,23 @@ const AZURE_FUNCTIONS_ATTRIBUTE_ENV_VARS = {
...
+      if (serviceName) {
         attributes = {
           ...attributes,
-          [SEMRESATTRS_FAAS_NAME]: functionName,
-        };
-      }
-      if (functionVersion) {
-        attributes = {
-          ...attributes,
-          [SEMRESATTRS_FAAS_VERSION]: functionVersion,
```
blumamir, line 73: Can you share the motivation to remove this attribute?

detectors/node/opentelemetry-resource-detector-azure/src/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+import { WEBSITE_OWNER_NAME, WEBSITE_RESOURCE_GROUP } from './types';
+
+export function getAzureResourceUri(
+  websiteSiteName: string
+): string | undefined {
+  const websiteResourceGroup = process.env[WEBSITE_RESOURCE_GROUP];
+  const websiteOwnerName = process.env[WEBSITE_OWNER_NAME];
+
+  let subscriptionId = websiteOwnerName;
+  if (websiteOwnerName && websiteOwnerName.indexOf('+') !== -1) {
```
blumamir, line 26: should we add a test for this case?

detectors/node/opentelemetry-resource-detector-azure/src/detectors/AzureFunctionsDetector.ts
JacksonWeber: Yes, it's now correctly reported under the semantic resource attribute `SERVICE_NAME`, instead of the `FAAS_NAME` since this represents the name of the app service instance that azure functions runs within, not the name of the azure function that contains this worker.

detectors/node/opentelemetry-resource-detector-azure/src/detectors/AzureFunctionsDetector.ts
JacksonWeber: The `FUNCTIONS_EXTENSION_VERSION` will be set twice, once by the worker process and again by the host process in Azure Functions. This is a problem because the worker process will have a different (and not the value we want to set for the `FAAS_VERSION` value). Until a solution is determined for this, we shouldn't record this attribute.

detectors/node/opentelemetry-resource-detector-azure/src/detectors/AzureFunctionsDetector.ts
blumamir: I was actually referring to the `cons functionVersion` variable, which used to be populated into `SEMRESATTRS_FAAS_VERSION`, but now we don't use it for anything (except the `if` condition below that does not access it anymore).
If I am not missing anything, can this code line be safely removed?

blumamir: Thanks! I never worked with azure function, so it doesn't mean a lot to me, but I suggest adding a comment next to this line to explain why it's there and why it's important, for the future maintainers and readers.

detectors/node/opentelemetry-resource-detector-azure/src/utils.ts
JacksonWeber: Absolutely, added.

detectors/node/opentelemetry-resource-detector-azure/src/detectors/AzureFunctionsDetector.ts
JacksonWeber: Makes sense. I've added a note to make this more clear.

blumamir (approved):

hectorhdzg (approved):
LGTM

Issue #2207, opened 2024-05-14
errant `### Dependencies` sections at the top of a number of CHANGELOG.md files


trentm:
As noticed [in this PR review comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2172#discussion_r1600599514), a number of the CHANGELOG.md files in this repo have `### Dependencies` section incorrectly at the top, instead of under a particular release '## ...' h2 section.

1. It was working up to and including: "chore: release main" (#1590)
2. then starting with this release PR it started incorrectly adding that section: "chore: release main" (#1683)
3. and continued for a number of releases, up to and including: "chore: release main" (#1890)
4. until fixed by the time this release PR was made: "chore: release main" (#1935)

That last range (i.e. when it was "fixed") included #1901. A *guess* is that this bump resolved the issue:

```diff
-      - uses: google-github-actions/release-please-action@v3
+      - uses: google-github-actions/release-please-action@v4
```

# symptoms

It looks like this resulted in the h2 header not being added for packages that only had a "Dependencies" section for that release. E.g. there was no `## [1.3.2](...)` section header added here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1683/files#diff-40eb27dccbec82d0089702356abe45de22ff0b558e3668389a3d94782c8bc109

# trying to see what release-please change resulted in the fix

... and failing.

Looking at the timing of #1890 and #1935, matching to the tags on release-please-action.git (`v3.7.13...v4.0.2`) and to the included release-please.git versions (`v15.13.0..v16.5.0`), nothing jumped out as an obviously related PR.

# fixing

A quick fix would to just drop all those `### Dependencies` sections and move on. It means that a number of older releases that only had deps updates will not have a changelog section.

A longer fix would be to work through each of the following PRs and recreate the `## [${version}] ...` header for each package with an error and properly move that section down:

* 32204a36 - chore: release main (#1890) (4 months ago)
* 90928231 - chore: release main (#1855) (4 months ago)
* c7e7000b - chore: release main (#1829) (5 months ago)
* a757b5e4 - chore: release main (#1798) (6 months ago)
* 607d3755 - chore: release main (#1742) (6 months ago)
* 659d7baf - chore: release main (#1737) (7 months ago)
* a8c225d2 - chore: release main (#1683) (7 months ago)

trentm:
[...]

PR #2234, opened 2024-05-23
chore(lint): refactor Markdown linting to use markdownlint-cli2

trentm:
This switches from `markdownlint-cli` to `markdownlint-cli2` and drops
usage of the https://github.com/avto-dev/markdown-lint GitHub action in CI.

- The `avto-dev/markdown-lint` action was using a 4y old version of `markdownlint`.
  AFAICT that action is not being maintained.
- There is a new `npm run lint:markdown` and the `lint.yml` CI workflow uses it.
- The switch from `markdownlint-cli` to `markdownlint-cli2` and `.markdownlint-cli2.jsonc`
  as the config allows using the https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint
  VSCode plugin for in-editor Markdown lint warnings and intellisense in the config file.

Refs: https://github.com/open-telemetry/opentelemetry-js/pull/4713
Refs: #2207

---

The same was done in the core repo in https://github.com/open-telemetry/opentelemetry-js/pull/4713


trentm:
This updated markdown linting actually catches the errant `### Dependencies` sections described in #2207.

.markdownlint-cli2.jsonc
```diff
@@ -0,0 +1,16 @@
...
+{
+  "config": {
+    // https://github.com/DavidAnson/markdownlint/blob/main/README.md#rules--aliases
+    "MD013": false,
+    "MD024": false,
+    "MD033": false,
+    "MD041": false,
+    // XXX Rules to discuss:
+    "MD012": false // no-multiple-blanks, common in CHANGELOG.md files, ~1700 hits in changelog files
+    // "MD004": { "style": "dash" } // ul-style, ~3800 hits in 63 changelog files, XXX discuss just using "consistent" new default for dash
```
trentm, line 11: @blumamir I'd like to discuss these two rules.

## `"MD012": false`

This is not disabled in the core repo. If re-enabled in the contrib repo here, there are approximately 1700 hits (a) in most of the 63 CHANGELOG.md files and (b) a few hits in these files:

```
   1 plugins/node/opentelemetry-instrumentation-mysql/README.md
   3 .github/ISSUE_TEMPLATE/plugin_request.md
   4 .github/ISSUE_TEMPLATE/feature_request.md
   6 .github/ISSUE_TEMPLATE/bug_report.md
```

I think the cases in the CHANGELOG.md files will just keep happening because they are generated by release-please, we do not want this rule enabled for them. There are two options:

1. globally disable the rule in this config file
2. locally disable the rule for each CHANGELOG.md file. This is slightly burdensome (just a onetime edit) because there are 63 changelog files.

Thoughts?

## `MD004`

Earlier we changed to `style: "dash"` to prefer dashes for bullets. In the core repo we disabled this rule locally in the *2* changelog files. Options:

1. Disable is locally in the 63 changelog files -- similar to how we did in the core repo.
2. Use the newer default of `"consistent"`, which ensures that each file is consistent in the bullets it uses. That means the rule could apply to CHANGELOG.md files as well, but would enforce consistent `*` usage. 
Thoughts?

blumamir: > I think the cases in the CHANGELOG.md files will just keep happening because they are generated by release-please, we do not want this rule enabled for them. There are two options:
> 
> 1. globally disable the rule in this config file
> 2. locally disable the rule for each CHANGELOG.md file. This is slightly burdensome (just a onetime edit) because there are 63 changelog files.
> 
> Thoughts?

Another alternative would be to use 2 markdown configs, one for CHANGELOG and one for other markdowns. I don't mind disabling it across the repo, but could be nice to have it if not too complex.



> Earlier we changed to `style: "dash"` to prefer dashes for bullets. In the core repo we disabled this rule locally in the _2_ changelog files. Options:

If we have different config file for CHANGELOGs than this is addressed as well, but if not straight-forward than we can disable it locally in each file or use the `"consistent"` option. What ever is easier and makes more sense to you

trentm: I'd prefer to not have two separate config files, because:
- it would be more likely for those config files to diverge,
- it is a little messy to have yet another config file at the top,
- it will mean markdownlint editor plugins (like the VSCode one referred to above) won't work properly -- the plugin will just use the default `.markdownlint-cli2.jsonc` file

In commit 2e03f596 I've opted to avoid `<!-- markdownlist-disable ... -->` directives in every CHANGELOG.md file by selecting options that work for all .md files. That means `ul-style: consistent` and `no-multiple-blanks: false`.

(A personal preference note: I *like* using multiple blank lines sometimes, both in Markdown and code, to separate bigger sections vs. a single line for related blocks. So I was never a fan of `no-multiple-blank` lines and the eslint equivalent in code.)

detectors/node/opentelemetry-resource-detector-alibaba-cloud/CHANGELOG.md
```diff
@@ -1,77 +1,5 @@
 # Changelog
 
-### Dependencies
```
blumamir, line 3: Are these removed manually? will we get more of them with the next releases?

trentm: The removal of these `### Dependencies` sections were all done in this one commit if you want to review them together: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2234/commits/4733ea4b037d0eb80ed0723414d9648a32423ba8

The issue is described here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2207

tl;dr: This *was* a some bug in some old release-please-action that we were using. It isn't happening anymore, so I do **not** expect to get more of these sections.

Removing these sections here is effectively the "A quick fix ..." option described at https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2207

blumamir: 👍

blumamir (approved):

PR #2242, opened 2024-05-25
chore: add script to verify READMEs and apply few fixes

blumamir:
I am looking into opportunities to improve how we manage READMEs.

This PR adds a script to verify few peaces of instrumentation README, as well as introducing few fixes where the content was wrong, missing, or not consistent with other packages.

Currently, this is done for instrumentation packages, but can be extended to other packages later on


plugins/node/opentelemetry-instrumentation-nestjs-core/README.md
```diff
@@ -83,6 +83,4 @@ Apache 2.0 - See [LICENSE][license-url] for more information.
 [license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
 [npm-url]: https://www.npmjs.com/package/@opentelemetry/instrumentation-nestjs-core
 [npm-img]: https://badge.fury.io/js/%40opentelemetry%2Finstrumentation-nestjs-core.svg
-[pkg-repo-url]: https://github.com/nestjs/nest
-[pkg-npm-url]: https://www.npmjs.com/package/@nestjs/core
```
blumamir, line 87: these were not used in the readme

maryliag:
It would be really nice if this also had an option of `readme:fix`, similar to `lint:fix`

.github/workflows/lint.yml
```diff
@@ -26,3 +26,6 @@ jobs:
       - run: npm ci
       - name: Lint
         run: npm run lint
+
+      - name: Verify READMEs
+        run: npm run readme:verify
```
trentm, line 31: nit: Perhaps call it `lint:readme` to fit in with lint:markdown, lint:examples?

detectors/node/opentelemetry-resource-detector-container/README.md
```diff
@@ -51,3 +51,5 @@ Apache 2.0 - See [LICENSE][license-url] for more information.
 [discussions-url]: https://github.com/open-telemetry/opentelemetry-js/discussions
 [license-url]: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/LICENSE
 [license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
+[npm-url]: https://www.npmjs.com/package/@opentelemetry/resource-detector-container
+[npm-img]: https://badge.fury.io/js/%40opentelemetry%2Fresource-detector-container.svg
```
trentm, line 55: Note that the markdownlint-cli2 change in #2234 caught this one as well.

scripts/verify-readme.js
```diff
@@ -0,0 +1,82 @@
+const fs = require('fs');
+const path = require('path');
+
+const appRoot = process.cwd();
+
+// identify if it's node or web
+const isNode = appRoot.includes('node');
+const isWeb = appRoot.includes('web');
```
trentm, line 8: This logic will break if we move to a single `./packages/...` dir (option 3 of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223#issuecomment-2125925648).

An alternative option would be to read `auto-instrumentations-*/package.json`, and look at their "dependencies" to see if a given `pjson.name` is included.

scripts/verify-readme.js
```diff
@@ -0,0 +1,82 @@
...
+`;  
+
+  if (!currentReadmeContent.includes(distText)) {
+    throw new Error(
+      `README.md dist text is not valid. Please add the following text to the README.md file:\n\n${distText}`
+    );
+  }
+}
+
+console.log('README.md is verified!');
```
trentm, line 82: nit: IMO linters should be silent if there are no issues (e.g. as eslint & markdownlint-cli2 are).

trentm (approved):
Some nits.

.github/workflows/lint.yml
blumamir: I like it. updated the PR

scripts/verify-readme.js
blumamir: done 👍

JamieDanielson (approved):
Good old fashioned homemade linter 🤖  Love this for consistency

scripts/lint-readme.js
```diff
@@ -0,0 +1,84 @@
+const fs = require('fs');
+const path = require('path');
+
+const packageRoot = process.cwd();
+const monorepoRoot = path.resolve(`${packageRoot}/../../..`);
```
trentm, line 5: nit: This works currently because this is only being called in workspaces 3-levels deep: `plugins/*/*`.  A possible alternative would be:

```suggestion
const monorepoRoot = path.resolve(__dirname, '..');
```

trentm (approved):

scripts/lint-readme.js
blumamir: Thanks, I didn't know the `__dirname` trick. learned something new :)

PR #2247, opened 2024-05-27
test(instrumentation-pino): refactor Pino tests for correctness

trentm:
Before this change the 'PinoInstrumentation' was being created
one or more times per test case, which results in the pino
module getting wrapped *each* time. The pino instrumentation
wraps the top-level/default export, so it cannot use the typical
`instrumentation._wrap` along with `isWrapped`/`_unwrap` handling.
The result was a stack of pino() being wrapped over a dozen
times, with each of those `patchedPino`s being active.

This caused debugging complexity. For example, every test
case after the `logKeys` test case would get span context
fields being added as `trace_id/span_id` *and* `traceId/spandId`
because the earlier `PinoInstrumentation` was still active.
I suspect the 'pino-disabled.test.ts' tests were only
working because they ran *first*, because 'disabled' luckily
comes before 'enabled' alphabetically.

This refactors the tests back to a single file
(because mocha runs separate test files in the same
process there is some cross-talk between the files), and
works with a single instance of PinoInstrumentation.
FWIW the instr-bunyan and instr-winston tests work
similarly.


plugins/node/opentelemetry-instrumentation-pino/package.json
```diff
@@ -45,7 +45,6 @@
   },
   "devDependencies": {
     "@opentelemetry/api": "^1.3.0",
-    "@opentelemetry/context-async-hooks": "^1.8.0",
```
trentm, line 48: Reviewer note: This devDep was being used for:

```js
context.setGlobalContextManager(new AsyncHooksContextManager());
```

which isn't necessary in the tests because the only tracing being done is "create a span context and then immediately/synchronously log in that context". There is no need for tracking context through any async actions.

plugins/node/opentelemetry-instrumentation-pino/test/pino.test.ts
trentm, line 1: Reviewer note: This file was created by moving pino-enabled.test.ts and refactoring it. The top and bottom sections of the file are largely the same. If the diff of the two is useful for review, it is here: https://gist.github.com/trentm/ca146ab41d5193d6db1a24f5faa34114

Every test case that was in pino-enabled.test.ts and pino-disabled.test.ts is reproduced in this file.
- The `describe('enabled instrumentation'` section was renamed `describe('log correlation'`.
- Some of the tests in that section have been renamed and re-ordered. Here is a diff of the test output from before to after these changes, if that helps review: https://gist.github.com/trentm/27b0d589742c64f2f1d3b722c98696b6


This refactoring does away with a number of the "common" functions from common.ts. The basic argument for doing that is that I find this:

```ts
    it('does not inject span context', () => {
      tracer.startActiveSpan('abc', span => {
        logger.info('a message');
        span.end();

        sinon.assert.calledOnce(writeSpy);
        const record = JSON.parse(writeSpy.firstCall.args[0].toString());
        assert.strictEqual(record['msg'], 'a message');
        assert.strictEqual(record['trace_id'], undefined);
        assert.strictEqual(record['span_id'], undefined);
        assert.strictEqual(record['trace_flags'], undefined);
      });
    });
```

to be much more straightforward and easier to debug than this:

```ts
    it('does not inject span context', () => {
      const span = tracer.startSpan('abc');
        testNoInjection(testContext);
      });
    });
```

As well, the squishy `TestContext`, `TestInstrumentation`, and `TestInstrumentationAndContext` types were unhelpful, IMO.

In the end, this test file is slightly fewer lines of code than before -- not that lines of code is necessarily a reasonable metric for debuggability.

pichlermarc (approved):
This looks good, thanks :slightly_smiling_face: 

I admit to not looking into the tests in much detail, but it looks like most of them are still there. Some changes were made to the naming and functions from `common.ts` were inlined into the tests. I prefer the new test layout.

trentm:
Thanks for the review!

PR #2249, opened 2024-05-30
feat(instrumentation-pino): add log sending to Logs Bridge API

trentm:
This adds "log sending" -- sending log records to the Logs SDK -- for the `pino` logger.
(The same was done for Bunyan in #1713 and Winston in #1837.)

- When a Pino logger is created, this wraps the logger instance's output stream with a `pino.multistream` that sends to the original and to the Logs SDK (`class OTelPinoStream`).
- This adds `disableLogSending` and `disableLogCorrelation` options, the same as in instr-bunyan and instr-winston

Limitations:
- This does *not yet* add support for the `logSeverity` option. I prefer to do that in a separate change, because it might be involved with Pino loggers configured with custom levels.
- This does *not yet* test log sending in the *ESM tests*. I'll add that separately, because it requires some updates to the `contrib-test-utils` package.

# reviewer notes

The *test file* changes on this PR assume that https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2247 gets merged first. Please review that PR first. After that is merged, then the test file changes here will be additive and more clear.
**Update (2024-06-10): This is done now, so this PR is now ready for review.**

# checklist

- [x] passing CI
- [x] TAV tests
- [x] discuss [pino-opentelemetry-transport](https://github.com/pinojs/pino-opentelemetry-transport) alternative and trade-offs 
- [ ] separate issue for `logSeverity`
- [ ] separate issue for log-sending tests with ESM
- [ ] consider a separate issue for translating the default Pino errSerializer format to https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-logs.md (ditto for Bunyan and Winston?)


plugins/node/opentelemetry-instrumentation-pino/README.md
```diff
@@ -15,60 +15,105 @@ Compatible with OpenTelemetry JS API and SDK `1.0+`.
...
+| `disableLogSending`     | `boolean`         | Whether to disable [log sending](#log-sending). Default `false`. |
+| `disableLogCorrelation` | `boolean`         | Whether to disable [log correlation](#log-correlation). Default `false`. |
+| `logKeys`               | record            | A record with keys `traceId`, `spanId`, and `traceFlags` string fields giving the field names to use for log-correlation span data. |
+| `logHook`               | `LogHookFunction` | An option hook to inject additional context to a log record after trace-context has been added. This requires `disableLogCorrelation` to be false. |
 
 ## Semantic Conventions
 
 This package does not currently generate any attributes from semantic conventions.
 
+## Alternative log sending with `pino-opentelemetry-transport`
```
david-luna, line 107: I wonder if the instrumentation would be capable to know if `pino-opentelemetry-transport` is used when creating a new logger and disable sending in the main thread. Or maybe warn the user if this transport is detected and suggest to change its `disableSending` configuration if applies.

plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts
```diff
@@ -48,30 +49,73 @@ export class PinoInstrumentation extends InstrumentationBase {
...
+          const otelMixin = instrumentation._getMixinFunction();
+          const mixinSym = moduleExports.symbols.mixinSym;
+          const origMixin = logger[mixinSym];
+          if (origMixin === undefined) {
+            logger[mixinSym] = otelMixin;
+          } else {
+            logger[mixinSym] = (ctx: object, level: number) => {
+              return Object.assign(
+                otelMixin(ctx, level),
+                origMixin(ctx, level)
```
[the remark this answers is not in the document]
trentm: Yes. This was also the behaviour before this change. I basically just moved this same logic from the (now gone) `_combineOptions` function.

david-luna (approved):

plugins/node/opentelemetry-instrumentation-pino/README.md
trentm: First, I don't believe there is a way for instrumentation-pino to know if `pino-opentelemetry-transport` is being used.  Here is an example script:

```
const otel = require('@opentelemetry/api');
const pino = require('pino');

const transport = pino.transport({ target: 'pino-opentelemetry-transport' });
const logger = pino({level: 'debug'}, transport);
console.log('XXX The logger stream is:', logger[pino.symbols.streamSym]);

logger.debug({foo: 'bar'}, 'hi');

const tracer = otel.trace.getTracer('example');
tracer.startActiveSpan('manual-span', span => {
  logger.info('in a span');
  span.end();
});
```

If `instrumentation-pino` is active, the only thing it has to look at is that `logger[pino.symbols.streamSym]` value -- the destination stream of the Pino Logger. This is the thing that the instrumentation calls `origStream` and wraps in a `multistream`.

That thing is a `ThreadStream` (from https://github.com/pinojs/thread-stream/blob/main/index.js):

```
XXX The logger stream is: <ref *1> ThreadStream {
  _events: [Object: null prototype] {
    message: [Function (anonymous)],
    ready: [Function: onReady],
    close: [Function (anonymous)]
  },
  _eventsCount: 3,
  _maxListeners: undefined,
  worker: Worker {
    _events: [Object: null prototype] {
      newListener: [Function (anonymous)],
      removeListener: [Function (anonymous)],
      message: [Function: onWorkerMessage],
      exit: [Function: onWorkerExit]
    },
    _eventsCount: 4,
    _maxListeners: undefined,
    performance: { eventLoopUtilization: [Function: bound eventLoopUtilization] },
    stream: FakeWeakRef { _value: [Circular *1] },
    [Symbol(kCapture)]: false,
...
```

That `ThreadStream` provides no way to know what the `target` package or Node.js file is being run in the Worker thread.  In other words, instr-pino can know that a Pino *transport* is being used (i.e. something running in a separate thread), but has no idea whether that is `pino-opentelemetry-transport`.

I think it would be overkill to look into instrumenting `pino.transport()` to tag any returned transport that is using `pino-opentelemetry-transport` to try to tell.

---

Second, I'm not sure that we'd want instr-pino to automatically do anything even if it *did* know.

- Automatically disabling log-sending if `pino-opentelemetry-transport` (POT) was detected would potentially be surprising behaviour. I think to be explicit we'd then want `disableLogSending` to be a tri-state config: true, false, "true-if-POT-detected".
- Automatically warning if POT was detected? I suppose this would be fair. *Most* users would never want both active on a given Pino logger. (Eventually perhaps we'd want the ability for users to silence particular warnings from the OTel SDK that they don't care above -- something like Python's warnings filter https://docs.python.org/3/library/warnings.html#the-warnings-filter)

These are why I opted for just have documentation about this so that users are at least made aware of the options (assuming they get as far as reading this README). Perhaps a future "Troubleshooting" guide would merit a section on  "I'm using Pino and getting double logs. Why?!".

david-luna: Those are valid points. Thanks for taking the time to give your reasoning :)

trentm:
@hectorhdzg Would you be interested in reviewing this as well? I'm not sure if Pino is as important to some of the work you are doing as Winston.

plugins/node/opentelemetry-instrumentation-pino/README.md
```diff
@@ -15,60 +15,105 @@ Compatible with OpenTelemetry JS API and SDK `1.0+`.
...
-});
+  ]
+})
+sdk.start();
 
 const pino = require('pino');
-const logger = pino();
-logger.info('foobar');
-// {"msg":"foobar","trace_id":"fc30029f30df383a4090d3189fe0ffdf","span_id":"625fa861d19d1056","trace_flags":"01", ...}
+const logger = pino(});
```
hectorhdzg, line 40:
```suggestion
const logger = pino();
```

hectorhdzg:
@trentm changes look good, my only question is related to pino-opentelemetry-transport usage , sorry if this was discussed somewhere else, why not using this package and patching the logger construction similar to Winston? is this related to having older pino version support?

trentm:
> my only question is related to pino-opentelemetry-transport usage , sorry if this was discussed somewhere else, why not using this package and patching the logger construction similar to Winston? is this related to having older pino version support?

@hectorhdzg It hasn't been discussed elsewhere, so it is a good question.

The decision to not just suggest users use `pino-opentelemetry-transport` (POT) for log sending was *not* about having older pino version support.  The log-sending support implemented here requires pino@7 or later.  The current POT requires pino@8 or later, so I guess there *is* a difference in Pino versions supported. However, that wasn't a reason for my choice. (Pino 7 isn't being maintained anymore -- last release was in 2022 -- so there isn't much value is in support it with a new feature.)

> why not using this package and patching the logger construction similar to Winston?

Please correct me if I miss something, but I see this as two separate questions.

## instrumentation no longer patches the logger constructor args

First, why I moved away from patching the logger constructor to instead creating the Logger instance and then tweaking it.

I found that coping with all the possible input combinations to `pino(...)` and patching them to a form that does what we need was difficult.  https://getpino.io/#/docs/api?id=destination describes a lot of possible inputs:

> destination (Number | String | Object | DestinationStream | SonicBoomOpts | WritableStream)

This is the Pino internal code for normalizing the given `pino()` args: https://github.com/pinojs/pino/blob/a41b4d3c409b31529c8bcb4bf3cd4e016da8a8de/lib/tools.js#L290-L340
It has been a few weeks, but IIRC, because that `buildSafeSonicBoom` is internal I'm not sure that it is *possible* to accurately patch the input arguments and never change the semantics. As a minor example, the current instrumentation-pino throws on `pino(2)` -- which granted is an invalid pino() argument, but the instrumentation shouldn't break usage:

```
% cat sadpino.js
const pino = require('pino')
const log = pino(2)
log.info('hi')

% node sadpino.js
{"level":30,"time":1719270084687,"pid":83775,"hostname":"peach.local","msg":"hi"}

% node -r @opentelemetry/auto-instrumentations-node/register sadpino.js
OTEL_TRACES_EXPORTER is empty. Using default otlp exporter.
OpenTelemetry automatic instrumentation started successfully
/Users/trentm/tmp/asdf.20240624T155854/node_modules/@opentelemetry/instrumentation-pino/build/src/instrumentation.js:122
            options.mixin = this._getMixinFunction();
                          ^

TypeError: Cannot create property 'mixin' on number '2'
    at PinoInstrumentation._combineOptions (/Users/trentm/tmp/asdf.20240624T155854/node_modules/@opentelemetry/instrumentation-pino/build/src/instrumentation.js:122:27)
    at /Users/trentm/tmp/asdf.20240624T155854/node_modules/@opentelemetry/instrumentation-pino/build/src/instrumentation.js:55:47
    at Object.<anonymous> (/Users/trentm/tmp/asdf.20240624T155854/sadpino.js:2:13)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.18.2
```

After the pino logger is created, it has a well-defined shape that can be patched cleanly.

## why not just use 'pino-opentelemetry-transport'?

This could either mean that we get instrumentation-pino to automatically use
'pino-opentelemetry-transport' (POT), or just document to users that they should
use POT for log-sending with Pino.

On the "for using pino-opentelemetry-transport" side:

- It is maintained by the Pino community. In addition to being associated with Pino, it also would theoretically mean less maintenance for OTel JS. Both are positives.
- The design of using a worker thread means that in most situations (environments that aren't constrained to a single CPU), there will be less `JSON.parse` and OTLP serialization CPU work *on the main thread*.

My arguments for implementing it without POT are:

- Architecturally, POT differs *a lot* from our other instrumentations. It runs code in a separate worker-thread. AFAIK no other OTel instrumentations do that. In depends on (https://gist.github.com/trentm/95c1821bf84335eac0ba8c9a5dbf948d) and starts its own OTel SDK components. I think the dependencies alone rule it out as something that we could include by default in auto-instrumentations-node.
- Also on the architectural difference: There is a lot more complexity with POT that I worry would lead to more support effort. It starts its own OTel SDK components in the separate thread. The current usage of `node -r @opentelemetry/auto-instrumentations-node/register ...` with **also start an SDK in any worker threads**, because `-r ...` preload modules are passed to worker threads unless the code starting the worker does something special. Granted this is an issue orthogonal to Pino transport usage, that we should discuss separately. (E.g. Should the auto-instrumentations-node/register entry point have a guard to only run in the main thread?)
- A lesser complexity: For POT to include traceId/spanId in LogRecords, it relies on the `trace_id` / `span_id` fields to be on the log record. The implementation in this PR does not. The fairly-recently added `logKeys` option to instrumentation-pino allows users to change the log fields to `myTraceId` or whatever. This will break having span context fields in LogRecords in POT.
- This is perhaps more my personal opinion: I'm not convinced that *most* users of a logger need the more complex design of a sending on a separate thread to free up CPU cycles on the main thread.
- Assuming that we cannot include POT in instrumentation-pino by default, then the current implementation (not using POT) offers the benefit of being something that can be enabled by default -- for "zero-code" instrumentation. Granted we *could* get instrumentation-pino to "do log-sending only if `require('POT')` works" -- but I think that is subtle behaviour that is best avoided if possible.

Downsides of adding log-sending support to instrumentation-pino:

- It adds some potential confusion for users wanting to use Pino and OTel. They may need to now understand *both* options and choose between them, or ensure they work together.


Sorry, that got long.

hectorhdzg (approved):
LGTM, @trentm I appreciate your detailed response on my latest question.

PR #2251, opened 2024-05-31
fix(azure-functions-resource-detector): Update Azure Functions Detector to Consider WEBSITE_SKU

JacksonWeber:

- It is possible for the Azure Functions detector to not detect the Azure Functions environment correctly in certain edge cases.

- There are some edge cases in which the `FUNCTIONS_EXTENSION_VERSION` may be undefined, in which case we should also check the `WEBSITE_SKU` for the value `FlexConsumption` to determine if we still may be running in an Azure Function.


detectors/node/opentelemetry-resource-detector-azure/src/types.ts
```diff
@@ -27,6 +27,7 @@ export const WEBSITE_SITE_NAME = 'WEBSITE_SITE_NAME';
 export const WEBSITE_SLOT_NAME = 'WEBSITE_SLOT_NAME';
 
 export const FUNCTIONS_VERSION = 'FUNCTIONS_EXTENSION_VERSION';
+export const WEBSITE_SKU = 'WEBSITE_SKU';
```
maryliag, line 30: nit: move this to the group above to align with all others `WEBSITE_*`

detectors/node/opentelemetry-resource-detector-azure/test/detectors/AzureFunctionsDetector.test.ts
```diff
@@ -102,3 +102,26 @@ describe('AzureFunctionsDetector', () => {
...
 });
+
+it('should detect azure functions if websiteSku is defined as FlexConsumption', () => {
+  process.env.WEBSITE_SITE_NAME = 'test-service';
+  process.env.REGION_NAME = 'test-region';
+  process.env.WEBSITE_INSTANCE_ID = 'test-instance-id';
+  process.env.WEBSITE_SKU = 'FlexConsumption';
+  process.env.WEBSITE_MEMORY_LIMIT_MB = '1000';
+  process.env.WEBSITE_OWNER_NAME = 'test-owner-name';
+  process.env.WEBSITE_RESOURCE_GROUP = 'test-resource-group';
```
maryliag, line 113: can you make sure that the value of `FUNCTIONS_VERSION` is unset (just in case it gets added in another test by accident and your test starts with it already setup), so we know the results you're getting is because is doing the proper check on the `WEBSITE_SKU` and not because it has a function version?

maryliag (commented):
I added a few nits :)

detectors/node/opentelemetry-resource-detector-azure/test/detectors/AzureFunctionsDetector.test.ts
JacksonWeber: Added an assertion to ensure that these values are not defined before this test.

maryliag (approved):

hectorhdzg (approved):
LGTM

Issue #2254, opened 2024-06-04
ioredis instrumentation span.end getting called twice


peterabbott:
### What version of OpenTelemetry are you using?
 "@opentelemetry/instrumentation-ioredis"@0.40.0 (probably earlier ones too)

### What version of Node are you using?
20

### What did you do?

Enable IORedisInstrumentation class. On error querying Redis, causes warning around calling end on span already ended.

Looking at the code [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-ioredis/src/instrumentation.ts#L174-L186), end is getting called both on reject but again in the catch.  

### What did you expect to see?
no errors

### What did you see instead?

Messages like the follwing
```
"@opentelemetry/diag - get TRACE_ID-SPAN_ID - You can only call end() on a span once."
```

### Additional context

There is obviously an issue with our Redis setup, but using the IORedis instrumentation is actually causing issues. When we remove the instrumentation class the issues we see with our Redis go away (as well obviosuly of the error mentioned above).

peterabbott:
Potentially the utils endSpan method could have a `isRecording` checking

AbhiPrasad:
Opened a PR for this https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2256

pichlermarc:
@AbhiPrasad can I assign you? :slightly_smiling_face:

AbhiPrasad:
@pichlermarc yes please feel free to assign me! I have the fix, just need some advice on how to best test it.

pdeveltere:
@AbhiPrasad this seems to be a problem still. any updates on this?

PR #2256, opened 2024-06-05
fix(ioredis): Make sure span.end is only called once

AbhiPrasad:

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2254

I moved the `endSpan` helper inline and made it mutate a `spanHasEnded` flag to enforce `span.end` is only called once.

Not sure how to best test this change, advice there appreciated!


JacksonWeber:
Once some tests are added, this looks good to me. As for how you might go about doing that - would it be possible to recreate the scenario with the ioredis package that was calling `endSpan` multiple times and then use `sinon` to assert that with the new changes the `endSpan` method is only ever called once under those conditions?

AbhiPrasad:
> use sinon to assert that with the new changes the endSpan method is only ever called once under those conditions

I'm struggling figuring out a good way to spy on the returned span here. I have something like so:

```ts
let endSpanSpy;
const original = instrumentation['tracer'].startSpan;
sinon
  .stub(instrumentation['tracer'], 'startSpan')
  .callsFake((name, options, context) => {
    const span = original(name, options, context);
    endSpanSpy = sinon.spy(span, 'end');
    return span;
  });
  ```
  
  but that seems to be causing issues for some reason.

**Comment by AbhiPrasad** [comment]
Alright I have it something like so:

```
const endSpanSpy = sinon.spy(OtelSpan.prototype, 'end');
const span = provider.getTracer('ioredis-test').startSpan('test span');
try {
  await context.with(trace.setSpan(context.active(), span), () => {
    return client.set('non-int-key', 'non-int-value').then(() => {
      return client.incr('non-int-key', () => {
        throw new Error('This is an error');
      });
    });
  });
} catch (e) {
  // ignore
}

console.log(endSpanSpy.callCount);
```

but the `mocha` `after` hooks don't seem to like this 😢. My inexperience with mocha/sinon really showing here 😓

PR #2258, opened 2024-06-05
feat(instrument-mongoose-7): Added instrumentation package for mongoose v7 (#1606)

### LINKED ISSUE #1606: Support for Mongoose 7
Currently, instrumentation-mongoose explicitly does not support Mongoose 7:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/82267ab02f8d9b27613b5926089c42d04d4a4a7e/plugins/node/instrumentation-mongoose/src/mongoose.ts#L78-L83

Testing on one of my recently upgraded services that are now span-less, things appear to work out of the box if I just change that constraint. However, I suspect there might be edge cases or breakages I'm not triggering.

Renovate seems to have given up on the automatic bump here, #1420, but the `lerna` error is pretty opaque to me. Is outside help useful here, or mainly a question of a maintainer having time to look at it?

cc @blumamir since I saw you in the component_owners file

usfdmiller:

Fixes #1606 

Following the comment in the issue by pichlermarc creating a new package for Mongoose v7 was the easiest path forward for me, given the significant, non backwards compatible changes introduced by removing callbacks.

- Added `instrument-mongoose-7` package to `release-please-config.json`.
- Removed code and tests for the `remove` and `update` methods that were deprecated in Mongoose v6 and removed in Mongoose v7.
- Removed code and tests that were specific to the callbacks removed in Mongoose v7.
- Updated test for `findOneAndUpdate`, which appears to generate only a single span now.
- Updated .tav.yaml to test the latest minor versions of v7.

Honestly, I mostly just removed callbacks, nonexistent functions, and unnecessary tests.  Remaining tests pass and everything seems to build without error.


david-luna:
Support was added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2353

Issue #2238, opened 2024-05-23
Q: adding an auto-instrumentation from a different monorepo


maorleger:
Hiya friends, apologies in advance if this is documented somewhere but I own the [@azure/opentelemetry-instrumentation-azure-sdk](https://www.npmjs.com/package/@azure/opentelemetry-instrumentation-azure-sdk) client library and would love to add Azure instrumentation to the [auto-instrumentation-node](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node) metapackage.

The library is currently hosted in [our monorepo](https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/instrumentation/opentelemetry-instrumentation-azure-sdk) where we can absorb any core changes and release independently - are there options or existing libraries that contribute to the auto-instrumentation-node metapackage from a separate repository?

If not, any guidance or ideas y'all can share? 

Finally, as a last resort I _can_ move the logic here and deprecate the instrumentation package under the `@azure` namespace but would prefer not to. I'm very interested in adding support for Azure so would love to hear from y'all!

- [x] This only affects the JavaScript OpenTelemetry library

hectorhdzg:
In my opinion main benefit of having an instrumentation in this repo would be about visibility and maintainability making it fully open source, Azure Functions team is in the process of adding an instrumentation in this repo, we also have Azure specific resource detectors available here. 
We also have several Azure specific packages, like Azure Monitor Exporter and Azure Monitor Distro, that live outside of this repository, that is actually same Azure SDK repo as your instrumentation, these being our main products, we need to have bigger control of updates and releases, hot fixes being a possibility :) To solve discoverability issue, OpenTelemetry offers a registry of external components so users can find them, our Azure packages are currently available there. 
https://opentelemetry.io/ecosystem/registry/?language=js

lmolkova:
The benefit of having Azure SDK instrumentation here would be that default-distro would provide it and save users from explicitly adding a dependency. 

In this sense exporter is different because it's an exporter to a specific backend, while Azure SDK instrumentation is no different than AWS one and should be part of the default distro.

In case of Java, Azure SDK instrumentation is part of the vanilla otel Java agent - https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/azure-core

So I think it's not a question of a repository, but more of a default distro content.
[...]

PR #2259, opened 2024-06-06
docs: add contributing guidelines for new instrumentations

pichlermarc:

It has become clear that adding instrumentation packages to this repository puts a lot of strain on the OpenTelemetry JavaScript SIG. 

Currently Triagers, Approvers and Maintainers of this repository need to keep up-to-date with **40+ different packages and frameworks**, used across both the browser and Node.js, sometimes even other, officially unsupported runtimes. All of these packages are being constantly updated and improved and therefore the stream of required maintainance work is never-ending. As a reviewers, SIG members also have to be familiar with the with the internals of the instrumented package AND how this package is used in the real world. While this would be feasible for a handful of packages, it does not scale to the amount of packages that are being hosted in this repository.

The main goal of the SIG is develop and improve the OpenTelemetry Client libraries (the API, the SDK and the extensions defined by the OTel Specification) and to provide the community with tools to instrument their applications and libraries. Adding and maintaining more and more instrumentation poses a significant threat to the overall health of the project.

This PR adds a guideline [largely copied from the one the Go SIG](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation#new-instrumentation) has used over the past few years. It provides guidance on where new instrumentation should live:
- in the same package that's being instrumented ("native instrumentation"), via a direct dependency on `@opentelemetry/api` which is stable and is expected to stay stable for the foreseeable future.
- as a separate package, similar to the ones in this repository, provided by either by the authors of the instrumented package, or by community members that wish to instrument that package, but not hosted on this repository
- only if none of these options are not possible it should be hosted here.


CONTRIBUTING.md
```diff
@@ -232,6 +233,38 @@ please ping ([@open-telemetry/javascript-approvers](https://github.com/orgs/open
...
+If possible, OpenTelemetry instrumentation should be included in the instrumented package.
+This will ensure the instrumentation reaches all package users, and is continuously maintained by developers that understand the package.
+
+If instrumentation cannot be directly included in the package it is instrumenting, it should be hosted in a dedicated public repository owned by its maintainer(s).
+This will appropriately assign maintenance responsibilities for the instrumentation and ensure these maintainers have the needed privilege to maintain the code.
+
+The last place instrumentation should be hosted is here in this repository.
+Maintaining instrumentation here hampers the development of OpenTelemetry for JavaScript and therefore should be avoided.
+When instrumentation cannot be included in a target package and there is good reason to not host it in a separate and dedicated repository an [instrumentation request](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/new/choose) should be filed.
+Note that new instrumentation needs at least two contributors assigned to it as a code-owner.  It is the responsibility
```
JamieDanielson, line 260:
```suggestion
Note that new instrumentation needs at least two contributors assigned to it as code-owners.  It is the responsibility
```

JamieDanielson (approved):

trentm (approved):

CONTRIBUTING.md
```diff
@@ -232,6 +233,38 @@ please ping ([@open-telemetry/javascript-approvers](https://github.com/orgs/open
...
+Therefore, to reach our goal, we have the following recommendations for where instrumentation packages should live.
+
+1. Native to the instrumented package
+2. A dedicated public repository
+3. Here in the opentelemetry-js-contrib repository
+
+If possible, OpenTelemetry instrumentation should be included in the instrumented package.
+This will ensure the instrumentation reaches all package users, and is continuously maintained by developers that understand the package.
+
+If instrumentation cannot be directly included in the package it is instrumenting, it should be hosted in a dedicated public repository owned by its maintainer(s).
```
maorleger, line 254: This is what we (the Azure SDK) team do today with our [azure sdk instrumentation package](https://www.npmjs.com/package/@azure/opentelemetry-instrumentation-azure-sdk)

What is not clear to me, and I would love to hear from you all, is how would I then onboard our instrumentation to the auto-instrumentation-node metapackage?

looking at the code it sounds trivial:
- Add @azure/opentelemetry-instrumentation-azure-sdk as a runtime dependency of auto-instrumentation-node [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/package.json#L51)
- Add our [Instrumentation class](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/instrumentation/opentelemetry-instrumentation-azure-sdk/src/instrumentation.ts#L22) to the [instrumentation map](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/src/utils.ts#L95)

Is that what you'd expect to see? If not, is there any documentation for onboarding to the metapackage? I'm very interested in seeing us added there and would prefer to use our existing package (sounds like y'all prefer that as well).

> Caveat is we need to export the class at the top level (today we only export a factory function), but that's an easy change to make 😄

maryliag (approved):

david-luna (approved):

CONTRIBUTING.md
pichlermarc: We'd do not plan depend on thrid-party instrumentations from an OpenTelemetry metapackage. That's mainly for security reasons (I'm not worrying about any Azure packages, but there might be third-parties requesting this treatment that we're not so confident about).  

I'm currently prototyping a way to 
- auto-load instrumentations that are present in `node_modules` as defined by an environment variable. Users would have to explicitly opt-in to the thrid-party instrumentation instead of it being loaded by default, but that's an acceptable trade-off IMO. Essentially one would be able to set something like `OTEL_NODE_LOAD_INSTRUMENTATIONS=http,grpc,npm:@azure/opentelemetry-instrumentation-azure-sdk` and that would load the http/grpc instrumentations (if present in `node_modules`) and `@azure/opentelemetry-instrumentation-azure-sdk`
- generate code for a getNodeAutoInstrumentations (or similar) function. The code-generation can run as a build step and uses the same environment variable on run-time to control which instrumentations are active. This is intended to help with bundled apps and any cases where loading from `node_modules` is not an option (that's less interesting for instrumentations than it is for propagators, resource detectors, etc as `require` is gone in bundled apps). The build step would generate code that imports the instrumentation based on a config file.

I'm not far enough along to figure out how this would work for the OTel Operator, but I'm fairly confident that we can find a solution for this as well. These tools would basically replace the functionality of the auto-instrumentation metapackage. As far as I understand, the reason that metapackage exists today is that we did not have such functionality before.

maorleger: Thanks for the information! I understand that the auto-instrumentation metapackage may not be needed in the future however it is still the recommended approach today and I'd like to get our instrumentation on there. The guidance is a bit conflicting though - I am reading that:

- OTel doesn't want to pull in 3rd party instrumentations into the OTel metapackage
- OTel doesn't want folks to add 3rd party instrumentations into this repo (that is the 3rd option in order of preference) because of the maintenance burden

I also asked this in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2238 if you prefer to engage somewhere but a set of clear next steps would be much appreciated!

If not, I can also open a PR with what I think makes sense and let y'all provide feedback that way - I'd rather know what you prefer beforehand though 

Just to be concrete: I am interested in getting the Azure SDK instrumentation into the @opentelemetry/auto-instrumentation-node metapackage. I don't particularly care which approach I need to take and am happy to follow your guidance, but am confused about which path y'all prefer.

Thanks again!

PR #2261, opened 2024-06-06
docs(undici): Update option links to point to right link

AbhiPrasad:

Current options table points to https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici

Change to https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-undici and fix how line numbers are referenced

### DIFF
--- a/plugins/node/instrumentation-undici/README.md
+++ b/plugins/node/instrumentation-undici/README.md
@@ -50,11 +50,11 @@ Undici instrumentation has few options available to choose from. You can set the
 
 | Options | Type | Description |
 | ------- | ---- | ----------- |
-| [`ignoreRequestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#63) | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function. |
-| [`requestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-undici/src/types.ts#65) | `RequestHookFunction` | Function for adding custom attributes before request is handled. |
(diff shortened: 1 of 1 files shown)


plugins/node/instrumentation-undici/README.md
```diff
@@ -25,7 +25,9 @@ OpenTelemetry Undici/fetch Instrumentation allows the user to automatically coll
 To load a specific instrumentation (Undici in this case), specify it in the Node Tracer's configuration.
 
 ```js
-const { UndiciInstrumentation } = require('@opentelemetry/instrumentation-undici');
+const {
+  UndiciInstrumentation,
```
AbhiPrasad, line 29: my prettier formatter did these changes, lmk if you want them to be reverted.

JacksonWeber (approved):

plugins/node/instrumentation-undici/README.md
```diff
@@ -41,20 +43,19 @@ provider.register();
...
+| Options                                                                                                                                                | Type                    | Description                                                                                                                                                                                                                                                  |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
```
trentm, line 53: You have a formatter that does this for you, do you?

nit: I'm not sure this helps readablility for those reading README.md in a text editor. ;)
However, I know why it is being done.

trentm (approved):

plugins/node/instrumentation-undici/README.md
trentm: All good.

plugins/node/instrumentation-undici/README.md
AbhiPrasad: like the code changes above, this was also formatted via a markdown extension for vscode that I have enabled by default for all JS related projects.

here's what this looks like in the plain diff

image

yeah unfortunately markdown formatting here is super subjective, I know other formatters take a different approach.

david-luna (approved):

Issue #2266, opened 2024-06-07
Add support for tedious v18


kirrg001:
Hi there!

We have recently added support for [v16 and v17](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1656).
[Tedious](https://www.npmjs.com/package/tedious) released v18 **four** month ago.

I will try to create pull request.

Cheers

PR #2267, opened 2024-06-07
feat(instr-tedious): added support for v18

kirrg001:
closes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2266

**Why?**
- Tedious v18 moved @types/tedious to their repository.
- Tedious v18 uses [Typescript v5](https://github.com/tediousjs/tedious/blob/v18.2.0/package.json#L82) and [es2022](https://github.com/tediousjs/tedious/issues/1612)

**Problems:**
- We need to use different typescript versions in the project. This creates conflicts and errors. I think because by default TS checks all TS Definitions in the root folder, which are not compatible against the TS v5.
- We are using an outdated module ts-mocha. Won't build with newer typescript targets.
- NPM Workspaces do not support [nohoisting](https://github.com/npm/rfcs/issues/287). Using nohoist makes it possible to move all @types/* and typescript dependencies into the packages and would avoid errors we are getting from typescript regarding root dependencies.
- Eslint packages no longer compatibale. We had to update eslint packages for typescript too (for the package)
- I had to use ts-ignore flag to import different type notations into the test. Not optimal, but easiest way. And using dynamic imports based on the tedious version did not work for me. And doing it like [mongo](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/caf7cb503760d17ac6455dcf712f677ef665f07a/plugins/node/opentelemetry-instrumentation-mongodb/test) is soo much code duplications. But I am also not an expert in TS. I am open to full working solutions.

**Possible follow up tickets:**
- There is a root typescript dependency and every sub package has its own TS version. That doesn't make sense to me. Either one root ts version which every packages use and a package can override it on its own or the other way around.
- Use tilde for dev dependencies to speed up installation (consistency change). 
- Replace ts-mocha with ts-node everywhere.
- Update mocha to latest version everywhere.
- Find a better way to avoid using ts-ignore.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -31,6 +31,8 @@ module.exports = {
       rules: {
         "@typescript-eslint/no-floating-promises": "error",
         "@typescript-eslint/no-this-alias": "off",
+        '@typescript-eslint/no-explicit-any': "warn",
+        '@typescript-eslint/no-unused-vars': "warn",
         "@typescript-eslint/naming-convention": [
           "error",
           {
--- a/plugins/node/instrumentation-tedious/package.json
+++ b/plugins/node/instrumentation-tedious/package.json
@@ -67,7 +67,8 @@
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.52.0",
-    "@opentelemetry/semantic-conventions": "^1.22.0"
+    "@opentelemetry/semantic-conventions": "^1.22.0",
+    "@types/tedious": "^4.0.10"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/instrumentation-tedious#readme"
 }
--- a/plugins/node/instrumentation-tedious/package.json
+++ b/plugins/node/instrumentation-tedious/package.json
@@ -14,7 +14,7 @@
     "prewatch": "npm run precompile",
     "prepublishOnly": "npm run compile",
     "tdd": "npm run test -- --watch-extensions ts --watch",
-    "test": "nyc ts-mocha -p tsconfig.json 'test/**/*.test.ts'",
+    "test": "nyc mocha --require ts-node/register 'test/**/*.test.ts'",
     "test-all-versions": "tav",
     "version:update": "node ../../../scripts/version-update.js"
   },
@@ -47,25 +47,27 @@
     "@opentelemetry/api": "^1.3.0"
   },
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "6.21.0",
+    "@typescript-eslint/parser": "6.21.0",
+    "eslint": "8.9.0",
     "@opentelemetry/api": "^1.3.0",
     "@opentelemetry/context-async-hooks": "^1.8.0",
     "@opentelemetry/contrib-test-utils": "^0.40.0",
     "@opentelemetry/sdk-trace-base": "^1.8.0",
     "@types/mocha": "7.0.2",
     "@types/node": "18.6.5",
-    "mocha": "7.2.0",
+    "mocha": "^10.4.0",
     "nyc": "15.1.0",
     "rimraf": "5.0.5",
     "semver": "7.6.0",
-    "tedious": "17.0.0",
+    "tedious": "18.2.0",
     "test-all-versions": "6.1.0",
-    "ts-mocha": "10.0.0",
-    "typescript": "4.4.4"
+    "ts-node": "^10.9.2",
+    "typescript": "5.3.3"
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.52.0",
-    "@opentelemetry/semantic-conventions": "^1.22.0",
-    "@types/tedious": "^4.0.10"
+    "@opentelemetry/semantic-conventions": "^1.22.0"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/instrumentation-tedious#readme"
 }
--- a/plugins/node/instrumentation-tedious/src/instrumentation.ts
+++ b/plugins/node/instrumentation-tedious/src/instrumentation.ts
@@ -78,7 +78,7 @@ export class TediousInstrumentation extends InstrumentationBase {
     return [
       new InstrumentationNodeModuleDefinition(
         TediousInstrumentation.COMPONENT,
-        ['>=1.11.0 <18'],
+        ['>=1.11.0 <19'],
         (moduleExports: typeof tedious) => {
           const ConnectionPrototype: any = moduleExports.Connection.prototype;
           for (const method of PATCHED_METHODS) {
--- a/plugins/node/instrumentation-tedious/test/api.ts
+++ b/plugins/node/instrumentation-tedious/test/api.ts
@@ -16,30 +16,35 @@
 
 import * as assert from 'assert';
 import { promisify } from 'util';
+import * as semver from 'semver';
+
 import type {
   Connection,
   Request,
   TYPES,
-  ConnectionConfiguration,
+  // @ts-ignore
+  ConnectionConfig as ConnectionConfigLegacy,
+  // @ts-ignore
+  ConnectionConfiguration as ConnectionConfigNew,
 } from 'tedious';
 
 type Method = keyof Connection & ('execSql' | 'execSqlBatch' | 'prepare');
+type ConnectionConfig = ConnectionConfigLegacy | ConnectionConfigNew;
+
 export type tedious = {
   Connection: typeof Connection;
   Request: typeof Request;
   TYPES: typeof TYPES;
-  ConnectionConfig: ConnectionConfiguration;
+  ConnectionConfig: ConnectionConfig;
 };
 
-export const makeApi = (tedious: tedious) => {
+export const makeApi = (tedious: tedious, tediousVersion: string) => {
   const fullName = (resource: string) => {
     assert.strictEqual(typeof resource, 'string');
     return `[dbo].[${resource}]`;
   };
 
-  const createConnection = (
-    config: ConnectionConfiguration
-  ): Promise<Connection> => {
+  const createConnection = (config: ConnectionConfig): Promise<Connection> => {
     return new Promise((resolve, reject) => {
       const connection = new tedious.Connection(config);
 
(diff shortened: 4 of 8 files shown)


plugins/node/instrumentation-tedious/package.json
```diff
@@ -47,20 +47,23 @@
     "@opentelemetry/api": "^1.3.0"
   },
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "6.21.0",
+    "@typescript-eslint/parser": "6.21.0",
+    "eslint": "8.9.0",
```
kirrg001, line 52: 8.9.0 includes support for es2022

plugins/node/instrumentation-tedious/package.json
```diff
@@ -47,20 +47,23 @@
     "@opentelemetry/api": "^1.3.0"
   },
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "6.21.0",
```
kirrg001, line 50: [v6 is still compatible with eslint v8 and adds support for es2022 as well.](https://github.com/typescript-eslint/typescript-eslint/pull/8377)

plugins/node/instrumentation-tedious/test/api.ts
```diff
@@ -16,17 +16,31 @@
...
+import type {
+  Connection,
+  Request,
+  TYPES,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  ConnectionConfig as ConnectionConfigLegacy,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  ConnectionConfiguration as ConnectionConfigNew,
```
kirrg001, line 30: I know this is not ideal. I was not able to find a better & easier solution. And I am not very familiar with TS.

Possible long-term:
- write tests in TS
- run tests in JS
- requirement: remove the test build folder from the git ignore list
- needs a follow up ticket IMO

eslint.config.js
```diff
@@ -31,6 +31,8 @@ module.exports = {
       rules: {
         "@typescript-eslint/no-floating-promises": "error",
         "@typescript-eslint/no-this-alias": "off",
+        '@typescript-eslint/no-explicit-any': "warn",
+        '@typescript-eslint/no-unused-vars': "warn",
```
kirrg001, line 35: https://github.com/typescript-eslint/typescript-eslint/commit/5346b5bbdbba81439ba761c282ba9cdcec7b45c8

We need to add these because of using typescript-eslint/typescript-eslint in this pkg.
I know they need to belong to the eslint package file, but I thought this is fine as well and easier, because it was a warning before anyway.

kirrg001:
RFR

There were **lots of challenges** in this PR. I have tried my best to give context for the changes.

kirrg001:
@open-telemetry/javascript-approvers Requesting a review

plugins/node/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -78,7 +78,7 @@ export class TediousInstrumentation extends InstrumentationBase {
     return [
       new InstrumentationNodeModuleDefinition(
         TediousInstrumentation.COMPONENT,
-        ['>=1.11.0 <18'],
+        ['>=1.11.0 <19'],
```
blumamir, line 81: please also update the supported version range in the README of the instrumentation.

plugins/node/instrumentation-tedious/package.json
```diff
@@ -47,20 +47,23 @@
...
     "nyc": "15.1.0",
     "rimraf": "5.0.5",
     "semver": "7.6.0",
-    "tedious": "17.0.0",
+    "tedious": "18.2.0",
     "test-all-versions": "6.1.0",
-    "ts-mocha": "10.0.0",
-    "typescript": "4.4.4"
+    "ts-node": "^10.9.2",
+    "typescript": "5.3.3"
```
blumamir, line 66: I think this might be considered a break change which we try to avoid. @dyladan is it safe to apply?

plugins/node/instrumentation-tedious/package.json
```diff
@@ -47,20 +47,23 @@
...
+    "@typescript-eslint/parser": "6.21.0",
+    "eslint": "8.9.0",
     "@opentelemetry/api": "^1.3.0",
     "@opentelemetry/context-async-hooks": "^1.8.0",
     "@opentelemetry/contrib-test-utils": "^0.40.0",
     "@opentelemetry/sdk-trace-base": "^1.8.0",
     "@types/mocha": "7.0.2",
     "@types/node": "18.6.5",
-    "mocha": "7.2.0",
+    "mocha": "^10.4.0",
```
blumamir, line 59: is it needed for this PR to work?
we aim to align the entire repo to use the same common dependencies, so this should ideally be bumped as a dedicated PR that does that for all packages in the repo

plugins/node/instrumentation-tedious/test/api.ts
```diff
@@ -16,17 +16,31 @@
...
+} from 'tedious';
```
blumamir, line 31: Since we have both types from `tedious` and from the still existing `@types/tedious` dependency, is this import able to join them both?

blumamir (commented):
Thank you for working on adding the support for this new version.

Wrote few comments / concerns

plugins/node/instrumentation-tedious/src/instrumentation.ts
kirrg001: Will do

plugins/node/instrumentation-tedious/package.json
kirrg001: Why is that a breaking change? Its a dev dependency.

Please respect the reasoning in the PR description. We have to update ts

plugins/node/instrumentation-tedious/test/api.ts
kirrg001: I am not aware of any other solution. Please suggest a working alternative approach :)

plugins/node/instrumentation-tedious/package.json
kirrg001: I understand. As far as I remember I had to update mocha because I had to replace mocha-ts with ts-node as described.

But I can double check if you want to 👌

plugins/node/instrumentation-tedious/package.json
blumamir: I am not familiar with the exact details, but I remember discussions that if we update typescript version for the package, exiting users of the instrumentation with typescript 4 might have their transpiration broken.
@open-telemetry/javascript-maintainers can someone please help and supply guidance on the safe way to do that?

plugins/node/instrumentation-tedious/package.json
blumamir: Thank you, @kirrg001.

I reviewed a few other instrumentations to see how they handle tests. Based on their dependencies, I believe our current setup should work without needing to bump the shared monorepo dependency versions unless absolutely necessary.

However, I do support the idea of opening a dedicated PR to update some of the dependencies across the repository. If you have the time and would like to contribute to this effort, it would be greatly appreciated.

plugins/node/instrumentation-tedious/test/api.ts
blumamir: I was actually wondering if you know how it works. Since it's test code, i don't think we need to worry and the current implementation is fine with me, but curious how typescript handles situation where the types exists twice in `node_modules`

kirrg001: If the tests are running, depending on the **current installed** tedious version (types are either in tedious or in @types), the code imports whats available. 

I feel like TS was not designed to fulfill such a use case.

plugins/node/instrumentation-tedious/package.json
kirrg001: I will try to downgrade mocha to v7 again and let you know :)

kirrg001:
@open-telemetry/javascript-maintainers I would love to merge this. Requesting further feedback.

trentm:
Hi. I took a look and tried to work through the same issues. I put up a branch for comparison: 
https://github.com/open-telemetry/opentelemetry-js-contrib/compare/main...trentm:opentelemetry-js-contrib:tm-tedious-18-play
Repeating its commit message (because GitHub likes to hide those):

```
    tedious@18 adds its own .d.ts files. Compiling with them requires
    TypeScript v5, and either target:'es2022' for 'AggregateError'
    or using skipLibCheck:true in tsconfig.json.

    This change limits the update to TypeScript v5 to just while
    testing tedious@18 via test-all-versions' support for
    'peerDependencies'.

    THis also drops ts-mocha in favour of 'mocha --require ts-node/register'
    as was done on opentelemetry-js.git in https://github.com/open-telemetry/opentelemetry-js/pull/4840
```


- It uses some of the work that you've done, and tries to limit some things (like not updating the typescript version in package.json).
- I agree that TypeScript v5 is required to compile the test files with tedious@18. Given that we only need to do that for test-all-versions (TAV) testing, I used `peerDependencies` in ".tav.yml" for that (https://github.com/watson/test-all-versions?tab=readme-ov-file#peer-dependencies). 
- I limited some of the needed type changes in test/api.ts.
- I also split out some of the versions in .tav.yml for minimum support Node.js versions. This wasn't completely necesasry because test/instrumentation.test.ts knows how to skip out if the current version of Node and tedious isn't supported.
- I'm good with the switch to `mocha --require ts-node/register`. This was done a few months back in the opentelemetry-js.git repo. We should follow-up with a separate change to do the same for all the other packages in this repo.

### on skipLibCheck

- I found that `skipLibCheck: true` in "tsconfig.json" was necessary, otherwise the test code won't compile with tedious@18. 

It would fail with:

```
../../../node_modules/tedious/lib/connection.d.ts:761:29 - error TS2304: Cannot find name 'AggregateError'.

761     loginError: undefined | AggregateError | ConnectionError;
                                ~~~~~~~~~~~~~~
```

The alternative would be to update the tsconfig "target" to "es2022". That might have an impact on the shipping code for instr-tedious, so I thought that should be avoided for now, at least.

### options

Please let me know which you would prefer:

1. I could make a PR out of my branch for discussion and see about pushing that through, or
2. You could take the few changes I have from yours. If you do take my changes, please take my (smaller) diff to the top-level package-lock.json file as well.

kirrg001:
@trentm Thanks so much for taking the time to review this.

> I could make a PR out of my branch for discussion and see about pushing that through, or

This option sounds like the easiest. I'd suggest that we just open your PR and close mine :)

kirrg001:
Closing in favor of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2381

PR #2273, opened 2024-06-12
fix(instr-undici): respect requireParent flag when INVALID_SPAN_CONTEXT is used

reberhardt7:

This PR mirrors this fix in the http instrumentation: https://github.com/open-telemetry/opentelemetry-js/pull/4788

> The http instrumentation (and many other instrumentation libraries) create spans with INVALID_SPAN_CONTEXT to create NonRecordingSpans when we enable `requireParent` and no parent is present: https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L769
>
> That works fine; those spans are not exported. However, when we create _child_ spans, those end up becoming real spans that are exported. E.g. if we have an http server, when a request comes in, we create a span with INVALID_SPAN_CONTEXT, and then if the http request handler makes an outbound request, the instrumentation sees that it has a parent span (even though it is invalid) and creates a _real_ child span, initiating a new trace.
>
> The opentelemetry sdk creates new traces for descendants of INVALID_SPAN_CONTEXT: https://github.com/open-telemetry/opentelemetry-js/blob/ecc88a38d85d5d6adf847306d96e7be77b7df8d6/packages/opentelemetry-sdk-trace-base/src/Tracer.ts#L91

This checks whether the parent span is recording, and does not create a real child span if the parent isn't recording.


david-luna:
Hi @reberhardt7

1st of all thanks for contributing to OpenTelemetry :)

Could you please sign the CLA so we can proceed on moving your PR forward?

Thanks!

reberhardt7:
@david-luna I've signed the CLA now. Thank you :)

david-luna (approved):

reberhardt7:
@david-luna thanks for the review! Would you be the right person to take a look at [this analogous instrumentation-http PR](https://github.com/open-telemetry/opentelemetry-js/pull/4788), or would that be someone else?

PR #2282, opened 2024-06-14
fix(instr-undici): wrong user agent reported if no user agent were set

YoannMa:

- undici instrumentation wrongly reports the first header's name as the `user-agent` if no `user-agent` header was set (default behavior of `undici` from what I can tell)

`request.headers.findIndex(h => h.toLowerCase() === 'user-agent')` returns `-1` if no `user-agent` header was present.

I added a check to avoid reporting `headers[0]` as `user-agent`

Added a test, not sure if it's useful? Especially as it might break in the future if `undici` decide to add a default `user-agent`


david-luna (approved):
@YoannMa 

Having no user agent header is a scenario that is likely to happen. Thanks for contributing to OpenTelemetry with this bugfix :)

PR #2286, opened 2024-06-20
fix(instr-knex): handle config provider functions

reberhardt7:

The knex connection configuration can be set with a static object, or with a function. From the [knex docs](https://knexjs.org/guide/#configuration-options):

> A function can be used to determine the connection configuration dynamically. This function receives no parameters, and returns either a configuration object or a promise for a configuration object.
>
> ```
> const knex = require('knex')({
>   client: 'sqlite3',
>   connection: () => ({
>     filename: process.env.SQLITE_FILENAME,
>   }),
> });
> ```

This instrumentation didn't handle the function case previously, resulting in `undefined` in the span name and a bunch of missing attributes.

Instead of accessing properties in `client.config.connection`, the instrumentation should be accessing `client.connectionSettings`. This gets set for static configs [here](https://github.com/knex/knex/blob/f6ea8122b61af25cb20f445ee0d731d0a8eed567/lib/client.js#L76) and for config provider functions [here](https://github.com/knex/knex/blob/f6ea8122b61af25cb20f445ee0d731d0a8eed567/lib/client.js#L256)

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
@@ -123,13 +123,17 @@ export class KnexInstrumentation extends InstrumentationBase {
     return function wrapQuery(original: () => any) {
       return function wrapped_logging_method(this: any, query: any) {
         const config = this.client.config;
+        // connection settings are in config.connection, but since that can
+        // potentially be an async function that resolves to the actual
+        // connection settings, it's easier to access client.connectionSettings
+        const connectionSettings = this.client.connectionSettings;
 
         const table = utils.extractTableName(this.builder);
         // `method` actually refers to the knex API method - Not exactly "operation"
         // in the spec sense, but matches most of the time.
         const operation = query?.method;
         const name =
-          config?.connection?.filename || config?.connection?.database;
+          connectionSettings?.filename || connectionSettings?.database;
         const maxLen = (
           instrumentation._config as types.KnexInstrumentationConfig
         ).maxQueryLength!;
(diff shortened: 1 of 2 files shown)


reberhardt7:
I've been getting test failures caused by network flakes:

> npm ERR! code ECONNRESET
> npm ERR! network aborted
> npm ERR! network This is a problem related to network connectivity.
> npm ERR! network In most cases you are behind a proxy or have bad network settings.
> npm ERR! network 
> npm ERR! network If you are behind a proxy, please make sure that the
> npm ERR! network 'proxy' config is set properly.  See: 'npm help config'

If someone with permissions could rerun the tests for me, that would be much appreciated

blumamir (approved):
LGTM.

Thank you for contributing a fix to this issue, and also for adding useful links to provide more context for the review.

I do have one question - since the connection can be an async function that can resolves sometime in the future, is it possible that if a quesy is made very early, that the connectionSettings are still not resolved? Do you think it's worth adding a test for this case?

reberhardt7:
@blumamir I tried writing a test for this case:

```
  it('works when knex query is made before connection function resolves', async () => {
    let resolveConfig: () => void;
    const configResolved = new Promise<void>((resolve) => {
      resolveConfig = resolve;
    })
    client = knex({
      client: 'sqlite3',
      // Use a function to configure the connection
      connection: async () => {
        await configResolved;
        return {
          filename: ':memory:',
        };
      },
      useNullAsDefault: true,
    });

    const queryResult = client.raw('SELECT 1');
    process.nextTick(resolveConfig!);

    await queryResult;
    const instrumentationSpans = memoryExporter.getFinishedSpans();
    assert.strictEqual(instrumentationSpans.length, 1);
    assert.strictEqual(instrumentationSpans[0].name, 'raw :memory:');
    assert.strictEqual(
      instrumentationSpans[0].attributes['net.transport'],
      'inproc'
    );
  })
```

It does pass, as I believe knex must resolve the config function before attempting to make any queries. However, without being able to hook into the implementation of knex, this test seems inherently racey, since there is no way to guarantee that we start trying to run the query before resolving the config function. Do you think it is still worth adding?

reberhardt7:
@blumamir What are the next steps for trying to land this PR?

dyladan:
This PR has approval but also has conflicts. It is blocked on:

1. Add test suggested by @blumamir 
2. Fix conflicts

When these are completed this can be merged

dyladan:
>without being able to hook into the implementation of knex, this test seems inherently racey, since there is no way to guarantee that we start trying to run the query before resolving the config function. Do you think it is still worth adding?

I would not want a test to be flakey. @blumamir do you have an opinion?

PR #2289, opened 2024-06-20
chore: consistent and typed use of instrumentation config

blumamir:
After https://github.com/open-telemetry/opentelemetry-js/pull/4659 is merged released and updated in contrib, we can benefit from the base class config functions being typed and required.

This PR implements and closes https://github.com/open-telemetry/opentelemetry-js/issues/4668 by only accessing config from the getters and setters. After this one merges, I will make the `_config` property private in base class so that it is not used accidentally. 

This PR apply the relevant changes to the contrib instrumentations, which makes everything consistent, short and well-typed (no type assertions).

There is still plenty of potential improvements and alignments for the way different instrumentations uses the config, but I want to leave this out of this PR so that it is not too large.

All the changes in this PR do not change behavior or logic, and should not be breaking in anyway


trentm:
```
> @opentelemetry/instrumentation-graphql@0.41.0 compile
> tsc -p .

src/internal-types.ts:17:1 - error TS6133: 'InstrumentationConfig' is declared but its value is never read.

17 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

TS error in instr-graphql

blumamir:
> ```
> > @opentelemetry/instrumentation-graphql@0.41.0 compile
> > tsc -p .
> 
> src/internal-types.ts:17:1 - error TS6133: 'InstrumentationConfig' is declared but its value is never read.
> 
> 17 import { InstrumentationConfig } from '@opentelemetry/instrumentation';
>    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> TS error in instr-graphql

oops... fixed 👍

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -76,19 +76,13 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
...
-      PACKAGE_NAME,
-      PACKAGE_VERSION,
-      Object.assign({}, DEFAULT_CONFIG, config)
-    );
+    super(PACKAGE_NAME, PACKAGE_VERSION, { ...DEFAULT_CONFIG, ...config });
   }
 
   override setConfig(config: AmqplibInstrumentationConfig = {}) {
-    this._config = Object.assign({}, DEFAULT_CONFIG, config);
+    super.setConfig({ ...DEFAULT_CONFIG, ...config });
```
blumamir, line 85: review note: I think using the spread operator is more idiomatic way to achieve the merge of those objects in modern js.
We were using a mix of these patterns in the codebase, and I choose to align all those usages to the above syntax.

maryliag:
@pichlermarc looking at the message about the feature-freeze, it makes me think it should probably have the package name instead of "This package does not have..."  and "Are you familiar with this package?". Specially for the cases where the PR touches more than one package and you don't know which one is referring to.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
```diff
@@ -367,12 +366,17 @@ const handleResolveSpanSuccess = (
 
 export function wrapFieldResolver<TSource = any, TContext = any, TArgs = any>(
   tracer: api.Tracer,
-  getConfig: () => Required<GraphQLInstrumentationConfig>,
+  getConfig: () => GraphQLInstrumentationParsedConfig,
```
trentm, line 369: I'm not sure I follow the implications of this change.
Is this change why the `config.depth!` change below is necessary?

trentm (commented):
This looks great. I have one Q, and I haven't looked at why TAV tests are failing.

blumamir:
> This looks great. I have one Q, and I haven't looked at why TAV tests are failing.

Thanks, the TAV failing is from  mongodb instrumentation, and is also failing on `main`. Not related to the changes on this PR.

plugins/node/opentelemetry-instrumentation-graphql/src/utils.ts
blumamir: The `config.depth!` non null assertion is by mistake. fixed 👍 

The type `Required<GraphQLInstrumentationConfig>` is not accurate as the instrumentation parses some of the config options like `depth`, `mergeItems` etc, but some where still optional, like the `enabled` and `ignoreTrivialResolveSpans` which typescript didn't like.
The new implementation makes it so the type accurately reflects which fields are nullable and which are non-null after the `DEFAULT_CONFIG` merge.

trentm:
Updating your branch from main to get the mongodb TAV test fix.

pichlermarc:

Oh wow it's adding and removing the label and always posting the message, that's quite annoying. Sorry about that. 
I'll see if I can make some changes to prevent that from happening.  :face_with_diagonal_mouth:

trentm (approved):

plugins/node/opentelemetry-instrumentation-graphql/src/types.ts
```diff
@@ -80,3 +80,12 @@ export interface GraphQLInstrumentationConfig extends InstrumentationConfig {
    */
   responseHook?: GraphQLInstrumentationExecutionResponseHook;
 }
+
+// Utility type to make specific properties required
+type RequireSpecificKeys<T, K extends keyof T> = T & { [P in K]-?: T[P] };
```
obecny, line 85: can you please tell why this change was needed?, thx

PR #2291, opened 2024-06-21
chore: update all otel deps

pichlermarc:
Updates `package-lock.json` to pull in updated OTel deps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10579,11 +10579,11 @@
       "link": true
     },
     "node_modules/@opentelemetry/core": {
-      "version": "1.25.0",
-      "resolved": "https://registry.npmjs.org/@opentelemetry/core/-/core-1.25.0.tgz",
-      "integrity": "sha512-n0B3s8rrqGrasTgNkXLKXzN0fXo+6IYP7M5b7AMsrZM33f/y6DS6kJ0Btd7SespASWq8bgL3taLo0oe0vB52IQ==",
+      "version": "1.25.1",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/core/-/core-1.25.1.tgz",
+      "integrity": "sha512-GeT/l6rBYWVQ4XArluLVB6WWQ8flHbdb6r2FCHC3smtdOAbrJBIv35tpV/yp9bmYUJf+xmZpu9DRTIeJVhFbEQ==",
       "dependencies": {
-        "@opentelemetry/semantic-conventions": "1.25.0"
+        "@opentelemetry/semantic-conventions": "1.25.1"
       },
       "engines": {
         "node": ">=14"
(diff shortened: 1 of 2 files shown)


trentm (approved):

trentm:
Just noticed the failing test. Timing error. And I see you've also tried re-running the tests.

pichlermarc:
> Just noticed the failing test. Timing error. And I see you've also tried re-running the tests.

Yes, fastify instrumentation failed 3 times, hapi once. I'm looking into it.

trentm:
I couldn't repro the hapi one locally. Guessing: Is it just that the GH Action runners are so slow that the `runTestFixture`-based tests -- at least the hapi test failure was the "ESM" test that uses `runTestFixture` -- can occasionally hit the default 3s mocha timeout?

pichlermarc:
> I couldn't repro the hapi one locally. Guessing: Is it just that the GH Action runners are so slow that the `runTestFixture`-based tests -- at least the hapi test failure was the "ESM" test that uses `runTestFixture` -- can occasionally hit the default 2s mocha timeout?

Yes that might be it. One odd thing though:  looking at the logs it seems like the ESM tests do now consistently take longer to complete. I wonder if some changes are incurring a large performance hit.

From [this run](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/9691925874/job/26744285540) on `main`:
-  fastify 
   - ✓ should work with ESM usage (1278ms)
-  Koa Instrumentation 
   - ✓ should work with ESM usage (1183ms)
-  ExpressInstrumentation 
   - ✓ should work with ESM usage (1215ms)
-  PinoInstrumentation ESM usage
   -  ✓ should work with ESM default import (1109ms)
   -  ✓ should work with ESM named import (1069ms)
- Hapi Instrumentation - Hapi.Plugin Tests
  -  ✓ should work with ESM modules (1677ms)

[Latest run](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/9610162032/job/26747782850?pr=2291) on this PR:
-  fastify 
   - ✓ should work with ESM usage (1872ms) -> +594ms
-  Koa Instrumentation 
   - ✓ should work with ESM usage (1392ms) -> +209ms
-  ExpressInstrumentation 
   - ✓ should work with ESM usage (1843ms) -> +628ms
-  PinoInstrumentation ESM usage
   -  hapi failed so it did not run
-  Hapi Instrumentation - Hapi.Plugin Tests
    -  ✓ should work with ESM modules (>2000ms) -> +>323ms

pichlermarc:
Huh, really interesting, fixed this with commit [67e5564](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2291/commits/67e5564d8d14a4504f04b96e3ad5be41cee31891). It looks like there were a bunch of duplicate packages (`1.25.0` and `1.25.1`) which actually slowed down the startup so much that the tests timed out.

trentm:
> It looks like there were a bunch of duplicate packages (`1.25.0` and `1.25.1`) which actually slowed down the startup so much that the tests timed out.

The only ones I see with dupes:

```
opentelemetry/core@1.25.0
opentelemetry/core@1.25.1
opentelemetry/propagator-b3@1.25.0
opentelemetry/propagator-b3@1.25.1
opentelemetry/semantic-conventions@1.25.0
opentelemetry/semantic-conventions@1.25.1
```

Good job finding those. 
Here is a command to how many installs of each npm package in the tree:

```
% npm ls --package-lock-only -ap 2>/dev/null | awk -F'node_modules/' '{print $NF}' | sort | uniq -c | sort -n | tail
  32 minipass
  33 color-convert
  33 color-name
  33 tslib
  33 yallist
  37 semver
  38 ansi-styles
  39 lru-cache
  39 supports-color
  53 minimatch
```

Love having 53 different installs of minimatch. :)
Just otel packages:

```
% npm ls --package-lock-only -ap 2>/dev/null | rg @opentelemetry | awk -F'node_modules/' '{print $NF}' | sort | uniq -c | sort -n | tail
   1 @opentelemetry/sdk-metrics
   1 @opentelemetry/sdk-node
   1 @opentelemetry/sdk-trace-base
   1 @opentelemetry/sdk-trace-node
   1 @opentelemetry/sdk-trace-web
   1 @opentelemetry/sql-common
   1 @opentelemetry/winston-transport
   2 @opentelemetry/propagator-b3
  21 @opentelemetry/core
  22 @opentelemetry/semantic-conventions
```



So many Qs :)

1. I'm curious how you tracked this down, but don't take much time on it.
2. Was this a failure of `./scripts/update-otel-deps.js`? If so, I could try to dig in to avoid this in the future. Perhaps it could finish with running that command above (or equivalent) and error or warn if there are dupe installs of any otel packages.
3. Hundreds of ms to just load these modules. 😢

pichlermarc:
> So many Qs :)
> 
> 1. I'm curious how you tracked this down, but don't take much time on it.
> 

First I thought that we introduced some issue in import-in-the-middle that increased loading time but that was not what happened. Just updating IITM I did not see an increase in loading time. I started updating one by one as the script does and tried to find with which package does increase the loading time (i still assumed at that point that some code-change was causing the issue, even though we had very few changes to core in that release). The load time ended up increasing with a package that we did not change. I checked the code in the package to verify that nothing fishy was going on. 

Then I checked `package-lock.json` for versions that were left behind and saw that there's a bunch of them. Knowing that the only difference is that there's a duplication of packages, I made the change from https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2291/commits/67e5564d8d14a4504f04b96e3ad5be41cee31891 and saw that the time for the test decreased back to the old levels.

> 2. Was this a failure of `./scripts/update-otel-deps.js`? If so, I could try to dig in to avoid this in the future. Perhaps it could finish with running that command above (or equivalent) and error or warn if there are dupe installs of any otel packages.

Hmm, I think it was really the `@opentelemetry/propagator-b3` update that held back the other packages. Maybe just changing that to a caret version would do the trick already.

> 3. Hundreds of ms to just load these modules. 😢

Yes, I was actually somewhat shocked by this. I think it's something that's worth investigating further. At some point it'd be great to have some benchmarks that measure the added load time of OTel in various scenarios (plain cjs, plain esm, esm with customizaton hook, ...) so that we can work on reducing that.

PR #2294, opened 2024-06-21
fix(instr-express): fix handler patching for already patched router

### LINKED ISSUE #2271: Express instrumentation results in routes with no stack, crashing Ghost
## Steps to Reproduce

Given the following `index.js`, which imitates [Ghost's](https://github.com/TryGhost/Ghost) routing:

```js
const http = require('http');

// https://github.com/TryGhost/Ghost/blob/fefb9ec395df8695d06442b6ecd3130dae374d94/ghost/core/core/frontend/web/site.js#L31
let router;
const SiteRouter = (req, res, next) => router(req, res, next);
function setupSiteApp() {
    const siteApp = express('site');

    // fascimile of ParentRouter
    router = express.Router();
    router.use('/:slug', (_1, _2, next) => {
        console.log('application code - I will never trigger');
        next()
    });
    router.param('slug', triggerBug);

    Object.setPrototypeOf(SiteRouter, router)
    siteApp.use(SiteRouter);

    return siteApp;
}

// this is the bug-inducing function, ParentRouter#_respectDominantRouter
function triggerBug(req, res, next, slug) {
    // ParentRouter#getSiteRouter
    const siteRouter = req.app._router.stack.find((router) => {
        return router.name === 'SiteRouter';
    });

    console.log('stack:', siteRouter.handle.stack);
    // this should throw
    siteRouter.handle.stack.every;
    // ...and app code will never trigger
    next();
}

// now that setup is done, try to trigger the bug
const app = setupSiteApp();
app.listen(5000, () => {
    console.log('listening');
    http.get('http://localhost:5000/foo', (res) => {
        res.resume();
    });
});

```

Run twice:

```sh
# without otel
node index.js
# with otel
env NODE_OPTIONS='--require @opentelemetry/auto-instrumentations-node/register' node index.js
```

## Expected Result

Both runs print a router stack like:

```
❯ node index.js
listening
stack: [
  Layer {
    handle: [Function (anonymous)],
    name: '<anonymous>',
    params: { slug: 'foo' },
    path: '/foo',
    keys: [ [Object] ],
    regexp: /^\/(?:([^\/]+?))\/?(?=\/|$)/i {
      fast_star: false,
      fast_slash: false
    },
    route: undefined
  }
]
```

## Actual Result

When running with otel, the router is `undefined`, causing shenanigans and a server 500:

```
❯ env NODE_OPTIONS='--require @opentelemetry/auto-instrumentations-node/register' OTEL_LOG_LEVEL=none node index.js
listening
stack: undefined
TypeError: Cannot read property 'every' of undefined
    at triggerBug (/home/zirak/work/ghost-every-siterouter-bug/index.js:33:29)
    at paramCallback (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/index.js:415:7)
    at param (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/index.js:395:5)
    at Function.process_params (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/index.js:421:3)
    at next (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/index.js:280:10)
    at Function.handle (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/index.js:175:3)
    at router (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/index.js:47:12)
    at SiteRouter (/home/zirak/work/ghost-every-siterouter-bug/index.js:6:40)
    at /home/zirak/work/ghost-every-siterouter-bug/node_modules/@opentelemetry/instrumentation-express/build/src/instrumentation.js:214:37
    at Layer.handle [as handle_request] (/home/zirak/work/ghost-every-siterouter-bug/node_modules/express/lib/router/layer.js:95:5)
```

## Additional Details

It's also possible to replicate this with Ghost itself. Running Ghost with opentelemetry and visiting http://localhost:2368/about/ results in a 500.

I've yet to successfully debug exactly where this originates.

### OpenTelemetry Setup Code

_No response_

### package.json

_No response_

### Relevant log output

_No response_

david-luna:

- closes #2271 

- use [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) to iterate over all the enumerable properties of the `handler` (own properties and inherited)
- add a test with the same scenario described in #2271

I finally opted to use the `for...in` statement instead of the `Proxy`. I ran some benchmarks following [this article](https://thecodebarbarian.com/thoughts-on-es6-proxies-performance) and the results in node v18.16.0 were

```sh
% node index.js

vanilla x 1,107,664,482 ops/sec ±0.21% (101 runs sampled)
proxy x 3,842,562 ops/sec ±0.34% (75 runs sampled)
defineProperty x 1,108,137,480 ops/sec ±0.25% (101 runs sampled)
Fastest is defineProperty,vanilla
```

Since any middleware can access the `handler` properties I think there could be use cases where those properties are accessed frequently so I preferred the `for...in + defineProperty` approach. The only drawback is non enumerable props are not accessible but I think for now it's fine


plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -315,7 +315,11 @@ export class ExpressInstrumentation extends InstrumentationBase {
       // some properties holding metadata and state so we need to proxy them
       // through through patched function
       // ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1950
-      Object.keys(original).forEach(key => {
+      // Also some apps/libs do their own patching before OTEL and have these propeties
```
AbhiPrasad, line 318:
```suggestion
      // Also some apps/libs do their own patching before OTEL and have these properties
```

AbhiPrasad (approved):

Zirak:
Thank you! I've done the patch locally and tested with Ghost, things are working swimmingly.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -517,6 +517,47 @@ describe('ExpressInstrumentation', () => {
...
+        res,
+        next
+      ) => router(req, res, next);
+      router.use('/:slug', (req, res, next) => {
+        const stack = req.app._router.stack as any[];
+        routerLayer = stack.find(router => router.name === 'CustomRouter');
+        return res.status(200).end('bar');
+      });
+      // The patched router has now express router's own properties in its prototype so
+      // they are not acessible through `Object.keys(...)`
```
JamieDanielson, line 538:
```suggestion
      // The patched router now has express router's own properties in its prototype so
      // they are not accessible through `Object.keys(...)`
```
nit

JamieDanielson (approved):
Thanks for this @david-luna ! Also appreciate the new test and detail in the comments.

PR #2296, opened 2024-06-22
fix!: export instrumentations only as named export

blumamir:
This PR is targets how we name and export instrumentation classes from our packages across the repo, and introduces consistency and best practices to align the code base.

It makes it so that:
- each instrumentation class has the same name as the name it is exported with (few were called `Instrumentation` instead of `FooInstrumentation`)
- align the `index.ts` file to export consistently - everything from `instrumentation.ts` and `type.ts`. This is to remove complexity and promote safe practices. This aligns the few places which were too complex, to an equivalent import and export statements to align with most packages in the repo.
- Remove the default export of the instrumentation class from the few places where it was present. since all instrumentations supports named exports of the class, I assume very few people if any are actually importing an instrumentation this way, but it is possible however, and those users will have their transplation broken when upgrading to the new version and will need to update the import. 

### Benefits

- Consistency across the various packages in the repo - better maintenance and developing experince.
- Shorter code - remove some of the un-needed complexities in the `index.ts` files and replace with shorter equivalents. 
- better practices - the `types.ts` and `instrumentation.ts` files are intended to be user facing, and only export what the user is expected to consume. More on that in [the GUIDELINES](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/GUIDELINES.md#types). Exporting everything from those files will prevent accidentally forgetting to add new types to the `index.ts` as happen in the past

### Breaking Change

This PR is a breaking change by removing the default export for the following instrumentations:
- `fs`
- `lru-memoizer`
- `tedious`
- `generic-pool`
- `knex`
- `memcahed`
- `mysql2`
- `restify`
- `router`

As state above, it is possible breaking change, but very unlikely for real users as this style is supported for only few instrumentations. It will be published as minor patch which should not be pulled automatically by users that has caret dependency, so they need to actively pull new version and will observe the transpilation failing.

### DIFF
--- a/plugins/node/instrumentation-dataloader/src/index.ts
+++ b/plugins/node/instrumentation-dataloader/src/index.ts
@@ -15,4 +15,4 @@
  */
 
 export * from './types';
-export { DataloaderInstrumentation } from './instrumentation';
+export * from './instrumentation';
--- a/plugins/node/instrumentation-fs/src/index.ts
+++ b/plugins/node/instrumentation-fs/src/index.ts
@@ -14,11 +14,5 @@
  * limitations under the License.
  */
 
-import FsInstrumentation from './instrumentation';
-
-export { FsInstrumentation };
-
 export * from './instrumentation';
 export * from './types';
-
-export default FsInstrumentation;
--- a/plugins/node/instrumentation-fs/src/instrumentation.ts
+++ b/plugins/node/instrumentation-fs/src/instrumentation.ts
@@ -51,7 +51,7 @@ function patchedFunctionWithOriginalProperties<
   return Object.assign(patchedFunction, original);
 }
 
-export default class FsInstrumentation extends InstrumentationBase {
+export class FsInstrumentation extends InstrumentationBase {
   constructor(config: FsInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
   }
--- a/plugins/node/instrumentation-fs/test/fs.test.ts
+++ b/plugins/node/instrumentation-fs/test/fs.test.ts
@@ -22,7 +22,7 @@ import {
 } from '@opentelemetry/sdk-trace-base';
 import * as assert from 'assert';
 import { promisify } from 'util';
-import Instrumentation from '../src';
+import { FsInstrumentation } from '../src';
 import * as sinon from 'sinon';
 import type * as FSType from 'fs';
 import tests, { TestCase, TestCreator } from './definitions';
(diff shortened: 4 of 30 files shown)


pichlermarc:
Thanks for taking care of this. :slightly_smiling_face:

As far as aligning export goes, I wonder if we should actually align them to be explicit exports (instead of wildcard exports) wherever possible. It has happened often that we accidentally add to the public interface by adding something to a file which in `index.ts` was then re-exported by a wildcard export.

We can avoid these accidental exports by using explicit exports wherever we can (see also: https://github.com/open-telemetry/opentelemetry-js/issues/4186)

blumamir:
> Thanks for taking care of this. 🙂
> 
> As far as aligning export goes, I wonder if we should actually align them to be explicit exports (instead of wildcard exports) wherever possible. It has happened often that we accidentally add to the public interface by adding something to a file which in `index.ts` was then re-exported by a wildcard export.
> 
> We can avoid these accidental exports by using explicit exports wherever we can (see also: [open-telemetry/opentelemetry-js#4186](https://github.com/open-telemetry/opentelemetry-js/issues/4186))

Thank you for the review

This issue has a bit of history behind it:
- up until #1251 , the `types.ts` files contained both public facing exports, and internal exports.
- one can potentially mix public and private exports into the same file, and apply explicit exports from `index.ts` as suggested
- private types often rely on importing dev dependency, like the instrumented package.
- if an import from dev dependency is found in a `.d.ts` file, it will fail transpilation for the instrumentation package users. which happened very often in the past.
- `types.ts` and `instrumentation.ts` both contains user facing exports, thus the user transpiler will for sure need to `import` both `types.d.ts` and `index.d.ts` and all there transitive dependencies
- It was agreed that both these files MUST NOT contain any private exports, due to how easy it is for dev dependencies to leak into the public api with private exports, even with explicit exports.
- Given the last comment, if all exports from `types.ts` and `instrumentation.ts` are public, then exporting them as wildcard is safe.
- using explicit exports can makes it easier to forget exporting a type, which has happened frequently in the past.

All these decisions where made to prevent painful issues which had happen. I am open to discuss alternatives but personally think that using wildcard exports is the safest option, which also communicate the intent of these files to code readers / future maintainers, and reduces the chance of forgetting to export something important or exporting dev dependencies in the public api.

pichlermarc (approved):
I see, thanks for the detailed response. I think in light of that info aligning it the way you did is the correct way then. :slightly_smiling_face:

plugins/node/instrumentation-dataloader/src/index.ts
```diff
@@ -15,4 +15,4 @@
  */
 
 export * from './types';
-export { DataloaderInstrumentation } from './instrumentation';
+export * from './instrumentation';
```
trentm, line 18: Didn't we actually want to move away from `export * ...` if we didn't need it? https://github.com/open-telemetry/opentelemetry-js/issues/4186

PR #2299, opened 2024-06-26
feat: Add package for automatic propagator configuration

martinkuba:

Follow-up to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2232
Related to https://github.com/open-telemetry/opentelemetry-js/issues/4494.

The [spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/aws-lambda.md#xray-lambda-propagator-functionality) says that the lambda instrumentation should be configurable via the `OTEL_PROPAGATORS` env variable, which MUST include the `xray-lambda` propagator.

The [spec also says](https://github.com/open-telemetry/opentelemetry-specification/blob/6b5e9404a3c574c1239230d0753665b4add0b977/specification/context/api-propagators.md?plain=1#L354-L356) that third-party propagators MUST NOT be maintained as part of the core SDK.

This PR introduces new package that combines all known propagators configurable via the `OTEL_PROPAGATORS` env variable. It exposes a `getPropagator()` function that returns a propagators instance based on what is provided in the env variable.


metapackages/auto-configuration-propagators/README.md
```diff
@@ -0,0 +1,53 @@
...
+import { getPropagator } from '@opentelemetry/auto-configuration-propagators';
+
+const propagator = getPropagator();
+```
+
+Please see the [Supported propagators](#supported-propagators) section for more information.
+
+## Supported propagators
+
+- [@opentelemetry/](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/instrumentation-amqplib)
```
pichlermarc, line 30:
```suggestion
```

metapackages/auto-configuration-propagators/README.md
```diff
@@ -0,0 +1,53 @@
...
+```js
+import { getPropagator } from '@opentelemetry/auto-configuration-propagators';
+
+const propagator = getPropagator();
+```
+
+Please see the [Supported propagators](#supported-propagators) section for more information.
+
+## Supported propagators
+
```
pichlermarc, line 29: Can we put a note here that the only propagators that will be supported are the ones from the spec? 
I'd like to refer to this in the future if someone wants to add additional ones that don't have a well-known value and might clash with future additions to the spec (this is something we don't have do deal with with instrumentations/resource detectors as they're configured unspecified `OTEL_NODE*` env vars).

metapackages/auto-configuration-propagators/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "repository": "open-telemetry/opentelemetry-js-contrib",
+  "scripts": {
+    "clean": "rimraf build/*",
+    "compile": "tsc -p .",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "eslint . --ext .ts --fix",
+    "precompile": "tsc --version && lerna run version:update --scope @opentelemetry/auto-configuration-propagators --include-dependencies",
+    "prewatch": "npm run precompile",
+    "prepublishOnly": "npm run compile",
+    "tdd": "yarn test -- --watch-extensions ts --watch",
```
pichlermarc, line 25:
```suggestion
    "tdd": "npm run test -- --watch-extensions ts --watch",
```

pichlermarc (commented):
Thanks for picking this up. Some nits but overall looks good.
You can add me as a code-owner to the code-owners file. :slightly_smiling_face:

metapackages/auto-configuration-propagators/README.md
martinkuba: Added the note, please resolve if it looks good to you.

pichlermarc: looks good, thanks :+1:

pichlermarc (approved):
Ah there's also one other thing we need to to do make auto-labelling of PRs work: we need to add its path to `.github/component-label-map.yml`.

pichlermarc:
cc @trentm @JamieDanielson since you reviewed my original PR #2232

trentm:
Is the intention that `auto-instrumentations-node` will use this?
And that `sdk-trace-base` will NOT use this because that would be a dependency from the core repo to the contrib repo?

metapackages/auto-configuration-propagators/README.md
```diff
@@ -0,0 +1,53 @@
+# OpenTelemetry Meta Package for Propagators Configuration
+
+[![NPM Published Version][npm-img]][npm-url]
+[![Apache License][license-image]][license-url]
+
+## About
+
+This module provides a way to get a propagator based on the [OTEL_PROPAGATORS environment variable](env-var-url).
```
trentm, line 8:
```suggestion
This module provides a way to get a propagator based on the [`OTEL_PROPAGATORS` environment variable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration).
```

... or pull that URL out to an id:

```md
... [blah blah][env-var-url]

... 
[env-var-url]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration
```

The same URL is also used below in the same file.

metapackages/auto-configuration-propagators/src/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (propagatorName === 'none') {
+      diag.info(
+        'Not selecting any propagator for value "none" specified in the environment variable OTEL_PROPAGATORS'
+      );
+      return [];
+    }
+
+    const propagatorFactoryFunction = propagatorMap.get(propagatorName);
+    if (propagatorFactoryFunction == null) {
+      diag.error(
```
trentm, line 82:
```suggestion
      diag.warn(
```

SHould this perhaps be `diag.warn`?  https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#enum-value says, emphasis mine:

> For variables accepting an enum value, if the user provides a value the implementation does not recognize, the implementation MUST **generate a warning** and gracefully ignore the setting.

pichlermarc: Yes, I think this should be a `diag.warn`. Good catch. :+1:

pichlermarc:
> Is the intention that `auto-instrumentations-node` will use this?

The intention is to make the `register.js` setup of `auto-instrumentations-node` use it eventually.

> And that `sdk-trace-base` will NOT use this because that would be a dependency from the core repo to the contrib repo?

Exactly. People would have to add it manually to either the `NodeSDK` constructor, `TracerProvider#register()` or the `propgator` API, but it will be a one-liner for all of these. The lambda layer can take a dependency on the new package to set up the propagator [here](https://github.com/open-telemetry/opentelemetry-lambda/blob/c55ad8742d9d4714a950fe4eaf11ed3246673577/nodejs/packages/layer/src/wrapper.ts#L120-L124).

Having it seperate will allow us to move the thrid-party propagator packages (xray, xray-lambda) back to contrib.

metapackages/auto-configuration-propagators/src/utils.ts
martinkuba: updated

metapackages/auto-configuration-propagators/README.md
martinkuba: updated

martinkuba:
Looks to me like the build failure is unrelated to the change, but to the latest release of [scheduler](https://www.npmjs.com/package/scheduler).

trentm:
> Looks to me like the build failure is unrelated to this change.

Actually it is related to the (large, possibly accidental) "package-lock.json" update in this PR. Currently it includes this diff hunk:

```diff
     "node_modules/@types/scheduler": {
-      "version": "0.16.8",
-      "resolved": "https://registry.npmjs.org/@types/scheduler/-/scheduler-0.16.8.tgz",
-      "integrity": "sha512-WZLiwShhwLRmeV6zH+GkbOFT6Z6VklCItrDioxUnv+u4Ll+8vKeFySoFyK/0ctcRpOmwAicELfmys1sDc/Rw+A==",
+      "version": "0.23.0",
+      "resolved": "https://registry.npmjs.org/@types/scheduler/-/scheduler-0.23.0.tgz",
+      "integrity": "sha512-YIoDCTH3Af6XM5VuwGG/QL/CJqga1Zm3NkU3HZ4ZHK2fRMPYP1VczsTUqtsf43PH/iJNVlPHAo2oWX7BSdB2Hw==",
       "dev": true
     },
```

That is a transitive dep from here:

```
% npm ls @types/scheduler
opentelemetry-contrib@0.1.0 /Users/trentm/tmp/opentelemetry-js-contrib
└─┬ @opentelemetry/plugin-react-load@0.30.1 -> ./plugins/web/opentelemetry-plugin-react-load
  └─┬ @types/react@17.0.67
    └── @types/scheduler@0.23.0
```

On main this is currently:

```
opentelemetry-contrib@0.1.0 /Users/trentm/src/opentelemetry-js-contrib
└─┬ @opentelemetry/plugin-react-load@0.30.1 -> ./plugins/web/opentelemetry-plugin-react-load
  └─┬ @types/react@17.0.67
    └── @types/scheduler@0.16.8
```

`@types/react@17.0.67` has a wildcard dep on `@types/scheduler`: `  "@types/scheduler": "*",`.

So, something about the steps of `npm ...` commands you took in your working copy resulted in `@types/scheduler` being updated to version 0.23.0. This PR to `@types/scheduler` -- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65812 -- broke `@types/react@17.0.67` when it dropped some types that were being used and the `*` dependency.  There is a discussion of that in the DefinitelyTyped PR with the suggestion to move to latest `@types/react@^18`.

## Options

1. Updating to `@types/react@18` isn't an option for us because, IIUC, `./plugins/web/opentelemetry-plugin-react-load` does not support React 18.
2. I notice that the latest `@types/react@17.*` *does* seem to have a fix for this: `@types/react@17.0.80` now has this dep: `"@types/scheduler": "^0.16"`. (Aside: I'd point to the code for this, but I do not currently follow how DefinitelyTyped.git does things. No branches, no tags.)
3. I have a change that reduces the `package-lock.json` diff for this PR down to approximately 200 lines, instead of 23k lines. The diff (compared to "main") is: https://gist.github.com/trentm/9c9c80bd6915e0a33747e09dfd1bdae1

I think option 3 is the most appropriate for this PR. I'll attempt to push that to this feature branch.

trentm:
Aside: package-lock.json maint/updates are a royal PITA.

martinkuba:
@trentm Thanks for looking into this. I was just working on it myself and was coming to the same conclusion. I think what happened is the PR had a merge conflict in the package-lock.json file, and I tried resolving it by deleting it and doing npm install.

trentm:
> I tried resolving it by deleting it and doing npm install.

For the record I think this is what I did in the working copy:

```
cd ~/tm/opentelemetry-js-contrib8  # my working copy
cp ~/src/opentelemetry-js-contrib/package-lock.json .   # copy in the package-lock file from "main"
npm install
```

trentm (approved):
LGTM

JamieDanielson (approved):
Since our function is `getPropagator`(single) and it returns a `CompositePropagator` or a single propagator, I'm wondering if the package should actually be called `auto-configuration-propagator` (singular instead of plural)? I also thought about going the other way with this being `getPropagators` to match `getResourceDetectors...`.

I don't feel strongly about this, just an observation on the second look. Otherwise though this looks great, thanks for working on this.

trentm:
On the plural/singular:
- The NodeSDK API takes a singular `textMapPropagator`, and plural `resouceDetectors`, so I think it "fine" that the exported API here is `getPropagator`. Perhaps a note in the comment for that function and/or a not in a reference section in the README could clarify that a composite propagator is used to merge multiple independent propagators, but that could be done in a separate PR.
- The "users" of this package are using it to handle the plural `OTEL_PROPAGATORS` envvar, so I think it is "fine" that the package name is plural.

There is no "all singular" or "all plural" naming option here, so ... I think it is fine as is. And that we should merge and march on. :)

(I wonder if this package will be somewhat short-lived, if a more general auto-configuration handling package is written that handles file-config and/or other `OTEL_*` SDK envvars.)

JamieDanielson:
> There is no "all singular" or "all plural" naming option here, so ... I think it is fine as is. And that we should merge and march on. :)

Agreed! 🚀  Naming is hard.

PR #2302, opened 2024-06-26
fix(@opentelemetry/instrumentation-winston): Add extra details about @opentelemetry/winston-transport scenarios

### LINKED ISSUE #2283: [instrumentation-winston] May duplicate transport (and logs) without notice

```
    "@opentelemetry/api": "^1.8.0",
    "@opentelemetry/api-logs": "^0.51.1",
    "@opentelemetry/auto-instrumentations-node": "^0.46.0",
    "@opentelemetry/winston-transport": "^0.3.0",
```

```
node --version
v18.20.3
```

### What did you do?

I found an issue while testing [instrumentation-winston](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-winston) which may lead to duplicated logs (via duplicated transports).
To have logger working with logs exporter we may be tempted to follow [winston-transport/README.md#usage] (https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/winston-transport/README.md#usage) guide, extended with exporter configuration (posted also in #2090)  e.g.
```ts
import * as api from "@opentelemetry/api-logs";
import { OTLPLogExporter } from "@opentelemetry/exporter-logs-otlp-http";
import {
  detectResourcesSync,
  envDetectorSync,
  hostDetectorSync,
  processDetectorSync,
} from "@opentelemetry/resources";
import {
  BatchLogRecordProcessor,
  LoggerProvider
} from "@opentelemetry/sdk-logs";
import { OpenTelemetryTransportV3 } from "@opentelemetry/winston-transport";

import * as winston from "winston";

const logExporter = new OTLPLogExporter();
const loggerProvider = new LoggerProvider({
  resource: detectResourcesSync({
    detectors: [envDetectorSync, processDetectorSync, hostDetectorSync],
  }),
});

loggerProvider.addLogRecordProcessor(
  new BatchLogRecordProcessor(logExporter)
);
api.logs.setGlobalLoggerProvider(loggerProvider);

export const logger = winston.createLogger({
  level: "info",
  transports: [
    new winston.transports.Console(),
    new OpenTelemetryTransportV3(),
  ],
});

logger.info('Hello world');
```
And that will already cause a problem combined with [instrumentation-winston](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-winston) because transport aren't deduplicated and [instrumentation.ts#L220](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-winston/src/instrumentation.ts#L220) adds it without deduplication even if transport is already there. Which in the end leads to duplicated logs in collector / APM.

I expect that I won't have duplicated logs at the end.

### What did you see instead?

I observed duplicated logs because instrumentation added another `OpenTelemetryTransportV3` to list of transports in winston instance.

### Additional context

Instrumentation code that adds transport:
```ts
const openTelemetryTransport = new OpenTelemetryTransportV3(
  transportOptions
);
if (originalTransports && !Array.isArray(originalTransports)) {
  newTransports = [originalTransports];
}
newTransports.push(openTelemetryTransport);
args[0].transports = newTransports;
```
[instrumentation.ts#L220](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-winston/src/instrumentation.ts#L220)

hectorhdzg:
Fixes #2283

Adding extra documentation about scenarios where both winston transport is added and winston instrumentation log sending is enabled.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-winston/README.md
+++ b/plugins/node/opentelemetry-instrumentation-winston/README.md
@@ -140,7 +140,6 @@ const logger = winston.createLogger({
 
 Duplication of logs could happen if @opentelemetry/winston-transport is added as a transport in Winston logger and @opentelemetry/instrumentation-winston is also enabled with disableLogSending config as false.
 
-
 ## Semantic Conventions
 
 This package does not currently generate any attributes from semantic conventions.
--- a/plugins/node/opentelemetry-instrumentation-winston/README.md
+++ b/plugins/node/opentelemetry-instrumentation-winston/README.md
@@ -138,8 +138,7 @@ const logger = winston.createLogger({
 });
 ```
 
-[!WARNING]  
-> Duplication of logs could happen if @opentelemetry/winston-transport is added as a transport in Winston logger and @opentelemetry/instrumentation-winston is also enabled with disableLogSending config as false.
+Duplication of logs could happen if @opentelemetry/winston-transport is added as a transport in Winston logger and @opentelemetry/instrumentation-winston is also enabled with disableLogSending config as false.
 
 
 ## Semantic Conventions
--- a/plugins/node/opentelemetry-instrumentation-winston/README.md
+++ b/plugins/node/opentelemetry-instrumentation-winston/README.md
@@ -107,7 +107,7 @@ Log injection can be disabled with the `disableLogCorrelation: true` option.
 ### Using OpenTelemetryTransportV3 without instrumentation
 
 [@opentelemetry/winston-transport](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/winston-transport) package exports the Winston transport class that is used to send records to the
-OpenTelemetry Logs SDK. It can be used directly when configuring a Winston logger. 
+OpenTelemetry Logs SDK. It can be used directly when configuring a Winston logger.
 For example:
 
 ```js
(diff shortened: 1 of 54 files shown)


plugins/node/opentelemetry-instrumentation-winston/README.md
```diff
@@ -104,6 +104,42 @@ logHook: (span, record) => {
...
+const logger = winston.createLogger({
+  level: 'info',
+  transports: [
+    new winston.transports.Console(),
+    new OpenTelemetryTransportV3()
+  ]
+});
+```
+
+Duplication of logs could happen if @opentelemetry/winston-transport is added as a transport in Winston logger and @opentelemetry/instrumentation-winston is also enabled with disableLogSending config as false.
```
pichlermarc, line 141: non-blocking: we can highlight it with something like this (some editorial changes in my suggestion, feel free to ignore those):

```suggestion
> [!IMPORTANT]
> Logs will be duplicated if `@opentelemetry/winston-transport` is added as a transport in `winston` and `@opentelemetry/instrumentation-winston` is configured with `disableLogSending: false`.
```

pichlermarc (approved):
Thanks :+1:

trentm (approved):

PR #2303, opened 2024-06-26
fix(koa)!: use generic config hook types and move dep to dev

### LINKED ISSUE #2228: opentelemetry-js-contrib has a hard dependency version for @types/koa -> 2.14.0 which creates duplication in yarn.lock
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5ecab2cf903fe56ab83ee7f43bd5b0d4bbaa4ba1/plugins/node/opentelemetry-instrumentation-koa/package.json#L71

Version should be changed to track minor change aka "2.\*.\*" aka "^2.14.0"

The same goes for @types/koa__router.

blumamir:
fixes #2228 

This PR aims to remove the instrumentation package dependency on the koa types packages, by making the relevant types generic and defaulting to `any`. That allows users who choose to implement the hook, to either use type `any`, or install the relevant types packages and populate the generic parameters.

- [x] moved koa types packages to dev dependency instead of regular dependency
- [x] documented how to use in README
- [x] updated tests to verify it compiles correctly when used in recommended way
- [x] compiled the new code and verified that `d.ts` files transitively imported from `index.d.ts` do not import types from `@types/koa`, `@types/koa__router` or any other dev dependency

**Breaking Change** 

- this PR makes it so that `KoaContext` is no longer exported as public type from the package 
- some hook info properties type will change to `any` by default, which should not break anyone I believe, but will no not verified  this variable typing as before.

**Prior Art**

defaults to any and allow to use generics to type correctly:

- [express instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-express#using-requesthook) see concensus [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1804#issuecomment-1843639947)

vendors part of the type as default, but allows to specify the full type when desired:
- [kafkajs](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/0ed40384287a8d06549c2a9c98a26ea9b068c472/plugins/node/instrumentation-kafkajs/src/types.ts#L32)
- [undici](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/0ed40384287a8d06549c2a9c98a26ea9b068c472/plugins/node/instrumentation-undici/src/types.ts#L46)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12632,6 +12632,7 @@
       "version": "1.3.7",
       "resolved": "https://registry.npmjs.org/@types/accepts/-/accepts-1.3.7.tgz",
       "integrity": "sha512-Pay9fq2lM2wXPWbteBsRAGiWH2hig4ZE2asK+mm7kUzlxRTfL961rj89I6zV/E3PcIkDqyuBEcMxFT7rccugeQ==",
+      "dev": true,
       "dependencies": {
         "@types/node": "*"
       }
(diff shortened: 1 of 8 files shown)


pichlermarc (approved):
Looks great, thanks :+1: 
Another instrumentation that had the instrumented package's types part of their public interface crossed of the list :slightly_smiling_face:

Issue #2311, opened 2024-07-02
When using auto-instrumentations-node, some values of `OTEL_NODE_RESOURCE_DETECTORS` are treated as invalid


qbedard:
### What version of OpenTelemetry are you using?

@opentelemetry/api@1.9.0
@opentelemetry/auto-instrumentations-node@0.47.1
@opentelemetry/exporter-trace-otlp-http@0.52.1
@opentelemetry/instrumentation-express@0.40.1
@opentelemetry/instrumentation-http@0.52.1
@opentelemetry/instrumentation-undici@0.3.0
@opentelemetry/resource-detector-aws@1.5.1
@opentelemetry/sdk-node@0.52.1
@opentelemetry/sdk-trace-node@1.25.1

### What version of Node are you using?

18.12.1

### What did you do?

Use [auto-instrumentations-node](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/auto-instrumentations-node) with `OTEL_NODE_RESOURCE_DETECTORS` set to `aws,env,host,os`.

### What did you expect to see?

AWS resource detector is enabled.

### What did you see instead?

> Invalid resource detector "aws" specified in the environment variable OTEL_NODE_RESOURCE_DETECTORS

### Additional context

[This conditional](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/0c46dfe406b22d4e83399812b421514b50a58573/metapackages/auto-instrumentations-node/src/utils.ts#L277) is evaluating to `false`-y and appears to do so for all resource detectors that are of type `Detector` or `Detector[]`. Those that have values of `DetectorSync` or `DetectorSync[]` _do_ work, so the issue is likely that a `Detector` instance is evaluating to `false`-y and the linked conditional or the detector map needs to be updated to evaluate those values properly.

trentm:
@qbedard Thanks for the issue, and the PR.

I believe what is happening here is that those `Invalid resource detector "aws" specified ...` diag.error messages are actually coming from the underlying `@opentelemetry/sdk-node` package that `@opentelemetry/auto-instrumentations-node` is using.

For example:

```
% cat foo.js
console.log('hi')
[16:59:00 trentm@peach:~/tmp/asdf]
% OTEL_NODE_RESOURCE_DETECTORS=foo,gcp,alibaba,aws,env,host,os node -r @opentelemetry/auto-instrumentations-node/register foo.js
[...]

PR #2313, opened 2024-07-02
fix(auto-instrumentations-node): check if detector key in map rather than value false-iness

qbedard:

https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2311

In auto-instrumentations-node, the check for detector configuration values is evaluating the `false`-iness of the map value, causing it to incorrectly treat `Detector | Detector[]` values as `false`-y. This PR changes that check to instead check for the the presence of the key value in the map.

### DIFF
--- a/metapackages/auto-instrumentations-node/src/utils.ts
+++ b/metapackages/auto-instrumentations-node/src/utils.ts
@@ -273,12 +273,11 @@ export function getResourceDetectorsFromEnv(): Array<Detector | DetectorSync> {
   }
 
   return resourceDetectorsFromEnv.flatMap(detector => {
-    const resourceDetector = resourceDetectors.get(detector);
-    if (!resourceDetector) {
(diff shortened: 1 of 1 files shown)


trentm (changes requested):
See https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2311#issuecomment-2207586470
I believe a more appropriate fix would be on the sdk-node package in https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-sdk-node/

blumamir:
@qbedard thank you for documenting this issue and working on a fix. I too think the fix should be in sdk-node package. would you like to implement the fix there?

qbedard:
Should be fixed in https://github.com/open-telemetry/opentelemetry-js/pull/4879. Closing.

Issue #2309, opened 2024-06-28
[ci][instrumentation-mongodb] tav tests are failing


pichlermarc:
Tests for `instrumentation-mongodb` are currently failing.

See https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/9709436608/job/26798116241.

pichlermarc:
Looks like the problem is `mongodb@6.8.0` that was released 18 hours ago (see https://github.com/mongodb/node-mongodb-native/releases/tag/v6.8.0)

I assume it to be a side effect of "Cursor responses are now parsed lazily"

The error from the tests:

```
2024-06-28T09:28:20.2247855Z      MongoRuntimeError: Unexpected null cursor id. A cursor creating command should have set this
2024-06-28T09:28:20.2249200Z       at FindCursor.getMore (node_modules/mongodb/src/cursor/abstract_cursor.ts:652:13)
2024-06-28T09:28:20.2250498Z       at FindCursor.getMore (node_modules/mongodb/src/cursor/find_cursor.ts:107:34)
2024-06-28T09:28:20.2251796Z       at FindCursor.fetchBatch (node_modules/mongodb/src/cursor/abstract_cursor.ts:730:35)
2024-06-28T09:28:20.2253263Z       at processTicksAndRejections (node:internal/process/task_queues:96:5)
2024-06-28T09:28:20.2254587Z       at FindCursor.next (node_modules/mongodb/src/cursor/abstract_cursor.ts:398:7)
2024-06-28T09:28:20.2255966Z       at FindCursor.[Symbol.asyncIterator] (node_modules/mongodb/src/cursor/abstract_cursor.ts:319:26)
2024-06-28T09:28:20.2257359Z       at FindCursor.toArray (node_modules/mongodb/src/cursor/abstract_cursor.ts:461:22)
```

pichlermarc:
Bumping to P1 as it's causing problems in end-user apps.

david-luna:
@pichlermarc indeed there is an internal change in the feature you mentioned which the instrumentation does not handle and raises the error
https://github.com/mongodb/node-mongodb-native/pull/4112/files#diff-588dd05062f39c3e44bcc04d3fe0ca9b370bdfb7469e4ce02d568373b2685cae

`Connection#command` now accepts an extra parameter and instrumentation does not pass it to the original function.

pichlermarc:
[...]

PR #2314, opened 2024-07-03
fix(instr-mongodb): fix  function patch missing one argument introduced in v6.8.0

david-luna:
…

`mongodb@6.8.0` introduced a new feature for lazy parsing documents. This feature relies on a new parameter on an internal API that the instrumentation was not passing to the original method resulting in an exception being thrown. Ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2309

this PR contains a fix to make sure the `command` original API gets all parameters when called. It uses `Function.prototype.apply` to call the original function so we do not stump on the same issue if any other param is added.

- use `Function.prototype.apply` to not lose any parameters when calling the original API
- added some extra types
- updated tests since one of them was inspecting data which is not parsed yet


plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v5-v6.test.ts
```diff
@@ -499,12 +499,23 @@ describe('MongoDBInstrumentation-Tracing-v5', () => {
...
         create({
-          responseHook: (span: Span, result: MongoResponseHookInformation) => {
-            span.setAttribute(dataAttributeName, JSON.stringify(result.data));
+          responseHook: (span: Span, result: any) => {
+            const { data } = result;
+            if (data.n) {
+              span.setAttribute('mongodb_insert_count', result.data.n);
+            }
+            // from v6.8.0 the cursor preoperty is not an object but an intance of
+            // `CursoResponse`. We need to use the `toObject` method to be able to inspect the data
```
pichlermarc, line 511:
```suggestion
            // `CursorResponse`. We need to use the `toObject` method to be able to inspect the data
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v5-v6.test.ts
```diff
@@ -499,12 +499,23 @@ describe('MongoDBInstrumentation-Tracing-v5', () => {
...
       beforeEach(() => {
         create({
-          responseHook: (span: Span, result: MongoResponseHookInformation) => {
-            span.setAttribute(dataAttributeName, JSON.stringify(result.data));
+          responseHook: (span: Span, result: any) => {
+            const { data } = result;
+            if (data.n) {
+              span.setAttribute('mongodb_insert_count', result.data.n);
+            }
+            // from v6.8.0 the cursor preoperty is not an object but an intance of
```
pichlermarc, line 510:
```suggestion
            // from v6.8.0 the cursor property is not an object but an instance of
```

plugins/node/opentelemetry-instrumentation-mongodb/test/mongodb-v5-v6.test.ts
```diff
@@ -533,7 +543,7 @@ describe('MongoDBInstrumentation-Tracing-v5', () => {
         });
       });
 
-      it('should attach response hook data to the resulting span for find function', done => {
+      it.only('should attach response hook data to the resulting span for find function', done => {
```
pichlermarc, line 546:
```suggestion
      it('should attach response hook data to the resulting span for find function', done => {
```

pichlermarc (approved):
@david-luna thanks for working on this. The fix looks good, just a few nits from my side. :+1:

trentm (approved):

Issue #1993, opened 2024-03-05
Express instrumentation does not properly handle router usage


JamieDanielson:
I don't think the current express instrumentation handles usage of a `express.Router()`.   Perhaps change `use-express.mjs` for now to not use a Router and we add a separate issue about handling `Router` in instr-express?

_Originally posted by @trentm in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1966#discussion_r1505871458_

I expected it to show up as `router - /post/:id` but instead was just `router - /`

JohannesHuster:
@JamieDanielson This is related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1947 

At the moment,`http.route` attributes of router and middleware spans are set to the "[mount path](https://expressjs.com/en/guide/using-middleware.html#:~:text=mount%20path)" (in the cases I could see).

```JavaScript
// Tests PR example (root mount path)
const routerA = express.Router();
routerA.get('/post/:id', (req, res) => {
  res.send(`Post id: ${req.params.id}`);
});
app.use(routerA);

// Adapted example (non-root mount path)
const routerB = express.Router();
routerB.get('/:id', (req, res) => {
  res.send(`Post id: ${req.params.id}`);
});
app.use('/post', routerB);
```

| | Tests PR example (root mount path) | Adapted example (non-root mount path) |
|--------|--------|--------|
| __router span__ | http.route: "/"<br>name: "router - /" | http.route: "/post"<br>name: "router - /post" |
| __request span__ |  http.route: "/post/:id"<br>name: "GET /post/:id" | http.route: "/post/:id"<br>name: "GET /post/:id" |

<details><summary>Spans for the example from tests PR (using a minimal express app)</summary>

```
{
  traceId: 'a1c7faafb2135a3d970d31093ee63437',
  parentId: 'fb4451520d0fa2b5',
  traceState: undefined,
  name: 'middleware - query',
  id: '788673a2a133abce',
  kind: 0,
  timestamp: 1710156373205000,
  duration: 375.252,
  attributes: {
    'http.route': '/',
    'express.name': 'query',
    'express.type': 'middleware'
  },
  status: { code: 0 },
  events: [],
  links: []
}
{
  traceId: 'a1c7faafb2135a3d970d31093ee63437',
  parentId: 'fb4451520d0fa2b5',
  traceState: undefined,
  name: 'middleware - expressInit',
  id: '7ece8b4ea1cf76fc',
[...]

Issue #1613, opened 2023-07-26
fix wrong rpcMetadata.route value when route.use(<middleware>) handle nested router


biuboombiuboom:
### What version of OpenTelemetry are you using?

    opentelemetry: 1.4.0
    node: 16.14.2

### What did you do?
        const app = express();
        const router = express.Router();
        app.use('/users', router);
        router.get('/:userId', (req, res) => {
          res.status(200).end('user-'+req.params.userId);
        });
        router.use(customMiddleware);
        const bookRouter = express.Router({ mergeParams: true });
        router.use('/:userId/books', bookRouter);
        bookRouter.get('/:bookId', (req, res) => {
          setImmediate(() => {
            res.status(200).end('book-' + req.params.bookId);
          });
        });       

    request http://localhost:8080/proxy/v1/bar/1

### What did you expect to see?

    rpcMetadata.route='/users/:userId/books/:bookId'

### What did you see instead?

    rpcMetadata.route='/:userId/books/:bookId'

mikesimons:
Looks like OP decided to drop the PR but this is definitely still an issue.

@blumamir @chigia001 Is the approach taken in #1615 the right direction?
Would consider picking this up with some direction on how you think it could be solved.

PR #2319, opened 2024-07-08
fix(instrumentation-express): Get `router` path from layer stack.

onurtemizkan:
Related: https://github.com/getsentry/sentry-javascript/issues/12103

- `router` spans are falling back to `/` as their span name in the current implementation. (Issue: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1993)
- `requestHandler` spans are not showing the full route in their span name when nested routers are used. (This is not fixing exactly but it is related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1613)

- This PR adds a new utility (`getRouterPath`) to recursively search the (possibly nested) parameterised router path when getting the layer metadata. This only affects the span name, not the `http.route` attribute of the span. `http.route` is still showing the mount path. (Active discussion on this: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1993#issuecomment-2159288594)

- Uses `route` primarily to assign the `requestHandler` span name instead of `layerPath`. `layerPath` does not represent the full path, only the path of the last matching router when nested routers are used.

So if we have a nested router setup like:

```javascript
const userRouter = express.Router();
const postsRouter = express.Router();

postsRouter.get('/:postId', (req, res, next) => {
  res.json({ hello: 'yes' });
  res.end();
  next();
});

userRouter.get('/api/user/:id', (req, res, next) => {
  res.json({ hello: 'yes' });
  res.end();
  next();
});

userRouter.use('/api/user/:id/posts', postsRouter);

app.use(userRouter);

```

The new behaviour is as below:

| Current Implementation      	| New Implementation                            	|
|-----------------------------	|-----------------------------------------------	|
| `router - /`                	| `router - /api/user/:id`                      	|
| `router - /`                	| `router - /:postId`                           	|
| `requestHandler - /:postId` 	| `requestHandler - /api/user/:id/posts/:postId`	|


plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -44,33 +44,64 @@ export const storeLayerPath = (request: PatchedRequest, value?: string) => {
...
   if (layer.name === 'router') {
+    let extractedRouterPath = '';
+
+    if (layer.handle?.stack) {
+      extractedRouterPath = getRouterPath(extractedRouterPath, layer);
+    }
+
+    if (!extractedRouterPath) {
+      extractedRouterPath = layerPath || route || '/';
+    }
```
AbhiPrasad, line 88:
```suggestion
    const maybeRouterPath = getRouterPath('', layer);
    const extractedRouterPath = maybeRouterPath ? maybeRouterPath : layerPath || route || '/';
```

AbhiPrasad (approved):

JamieDanielson:
Thank you for your work on this! Overall I like the idea of this and it seems to resolve the issue of span names from basic router usage as I had flagged in the issue, properly renaming `-router - /` to `-router - /post/:id` in the `router.get` usage.

What I'm currently stuck on is the nested router use case. The second router span in this PR is rewritten to `router - /:postId`, which matches the `postsRouter.get('/:postId')`. The current (on main) implementation is actually using the path from `userRouter.use('/api/user/:id/posts')`, and is fully excluding the `postsRouter.get('/:postId')`. So the table is a bit inaccurate there (it's not currently `router -/`), but I am not sure of what we should expect for that second router span. Should it be `router - /:postId` or should it be `router - /api/user/:id/posts/:postId`? 

| Current Implementation         | New Implementation                             |
| ------------------------------ | ---------------------------------------------- |
| `router - /`                   | `router - /api/user/:id`                       |
| ~`router - /`~ `router - /api/user/:id/posts`                   | `router - /:postId`                            |
| `requestHandler - /:postId`    | `requestHandler - /api/user/:id/posts/:postId` |

onurtemizkan:
>  So the table is a bit inaccurate there (it's not currently router -/)

@JamieDanielson, yes you're right. Thanks for pointing that out.

>  not sure of what we should expect for that second router span. Should it be router - /:postId or should it be router - /api/user/:id/posts/:postId

`/:postId` actually is the lower-level router. So IMHO it feels more accurate to have it as the `router` span name. We get the full path from the `requestHandler` span in this case.

JamieDanielson (approved):
This is titled as a fix, but it could perhaps be considered a feature as it properly implements the router instrumentation 😁  🤷 . I lean toward feature because it is changing span names and is more or less implementing something that didn't properly exist before. What do you think about a title like "feat(instrumentation-express):Use router path in router span names"?

JamieDanielson (changes requested):
I want to hold onto this for a bit longer until this other PR #2408 is merged that fixes the instrumentation of `http.get` and changes the [output of the ESM spans](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2408#discussion_r1735862915). It's going to require a few more changes to the ESM tests but they will make more sense now with the client and server spans.

After that PR merges and we update this branch we'll see the ESM tests fail. We'll just need to update the ESM tests here and then this will be good to go 🚀

onurtemizkan:
@JamieDanielson, sorry for the label/assignee mess caused by my bad merge. 😅 I had to force-push the branch with the updated tests.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -641,6 +641,86 @@ describe('ExpressInstrumentation', () => {
...
+        assert.strictEqual(spans[3].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[4].name, 'middleware - simpleMiddleware');
+        assert.strictEqual(spans[4].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[4].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[5].name, 'router - /api/user/:id');
+        assert.strictEqual(spans[5].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[5].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[6].name, 'request handler - /api/user/:id');
+        assert.strictEqual(spans[6].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[6].parentSpanId, spans[1].spanId);
```
pichlermarc, line 675: these are all internal spans, I think.

JamieDanielson, line 675:
```suggestion
        assert.strictEqual(spans[3].kind, testUtils.OtlpSpanKind.INTERNAL);
        assert.strictEqual(spans[3].parentSpanId, spans[1].spanId);
        assert.strictEqual(spans[4].name, 'middleware - simpleMiddleware');
        assert.strictEqual(spans[4].kind, testUtils.OtlpSpanKind.INTERNAL);
        assert.strictEqual(spans[4].parentSpanId, spans[1].spanId);
        assert.strictEqual(spans[5].name, 'router - /api/user/:id');
        assert.strictEqual(spans[5].kind, testUtils.OtlpSpanKind.INTERNAL);
        assert.strictEqual(spans[5].parentSpanId, spans[1].spanId);
        assert.strictEqual(spans[6].name, 'request handler - /api/user/:id');
        assert.strictEqual(spans[6].kind, testUtils.OtlpSpanKind.INTERNAL);
        assert.strictEqual(spans[6].parentSpanId, spans[1].spanId);
```

JamieDanielson:
> I want to hold onto this for a bit longer until this other PR #2408 is merged that fixes the instrumentation of `http.get` and changes the [output of the ESM spans](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2408#discussion_r1735862915). It's going to require a few more changes to the ESM tests but they will make more sense now with the client and server spans.
> 
> After that PR merges and we update this branch we'll see the ESM tests fail. We'll just need to update the ESM tests here and then this will be good to go 🚀

I've merged in main and updated the test to use `otlpSpanKind`. Once tests pass we can merge.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
JamieDanielson: I've updated this (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2319/commits/b1902dc08e067be20c884b1af3c4b23506b4906c), running tests now

JamieDanielson (approved):

JamieDanielson:
The failed tests may be intermittent and unrelated to this. I created #2436 to track but after some re-running they are working. Hope to be able to merge soon!

biuboombiuboom:
em.....if layerPath is undefined,  the router span name still does not contain the full route
utils.ts>storLayerPath
```
           if (value === undefined) return;
           (request[_LAYERS_STORE_PROPERTY] as string[]).push(value);
```
instrumentation.ts line 276-277, pop  layerStore  does`t check layerPath 
```
            if (!(req.route && isError)&&layerPath) {
              (req[_LAYERS_STORE_PROPERTY] as string[]).pop();
            }
```

test code 
```
it('should full route with route use(<middleware>)', async () => {
    const rootSpan = tracer.startSpan('rootSpan');
    const customMiddleware: express.RequestHandler = (req, res, next) => {
      for (let i = 0; i < 1000000; i++) {
        continue;
      }
      return next();
    };
    let finishListenerCount: number | undefined;
    let rpcMetadata: RPCMetadata | undefined;
    const httpServer = await serverWithMiddleware(tracer, rootSpan, app => {
      app.use(express.json());
      app.use((req, res, next) => {
        rpcMetadata = getRPCMetadata(context.active());
        res.on('finish', () => {
          finishListenerCount = res.listenerCount('finish');
        });
        next();
      });
      app.use('/users', router);
      router.get('/:userId', (req, res) => {
        res.status(200).end('user-' + req.params.userId);
      });
      router.use(customMiddleware);
      const bookRouter = express.Router({ mergeParams: true });
      router.use('/:userId/books', bookRouter);
      bookRouter.get('/:bookId', (req, res) => {
        setImmediate(() => {
          res.status(200).end('book-' + req.params.bookId);
        });
      });
    });
   let port = httpServer.port;
    assert.strictEqual(memoryExporter.getFinishedSpans().length, 0);
    await context.with(
      trace.setSpan(context.active(), rootSpan),
      async () => {
        const response = await httpRequest.get(
          `http://localhost:${port}/users/1/books/2`
        );
        assert.strictEqual(response, 'book-2');
        rootSpan.end();
        assert.strictEqual(finishListenerCount, 2);
        assert.notStrictEqual(
          memoryExporter
            .getFinishedSpans()
            .find(span => span.name.includes('customMiddleware')),
          undefined
        );
        assert.notStrictEqual(
          memoryExporter
            .getFinishedSpans()
            .find(span => span.name.includes('jsonParser')),
          undefined
        );
        const requestHandlerSpan = memoryExporter
        .getFinishedSpans()
        .find(span => span.name.includes('request handler'));
      assert.notStrictEqual(requestHandlerSpan, undefined);
        assert.notStrictEqual(requestHandlerSpan, undefined);
        assert.strictEqual(
          requestHandlerSpan?.attributes['http.route'],
          '/users/:userId/books/:bookId'
        );
   
        assert.strictEqual(
          rpcMetadata?.route,
          '/users/:userId/books/:bookId'
        );
      }
    );
```
image

Issue #2320, opened 2024-06-19
traces of resource detectors being sent


david-luna:
## Steps to Reproduce

- Get the express example from https://github.com/open-telemetry/opentelemetry-js-contrib
- instrument it with `@opentelemetry/auto-instrumentations`
- enable all resource detectors

```sh
OTEL_TRACES_EXPORTER=console \
  OTEL_NODE_RESOURCE_DETECTORS=all \
  node --require '@opentelemetry/auto-instrumentations-node/register' index.js
```

## Expected Result

Exported spans should belong only to application traces and not to any internal activity of the SDK or its components.

## Actual Result

HTTP spans are exported for the requests done by GCP while it's doing its detection logic. 

## Additional Details

This is related to open-telemetry/opentelemetry-js#1989 since the instrumentations do the patching when constructor is called (before the detectors being used). IMO a user would expect to start collecting/exporting traces once the SDK has completely initialized and also to only export traces/logs/metrics activity related to the application. Self instrumentation would be a feature that could be activated or not.

### Application Code

```JavaScript
// index.js code no need for tracing.js 
// since is using --require '@opentelemetry/auto-instrumentations-node/register'

// Usage:
//  OTEL_NODE_RESOURCE_DETECTORS=all node --require '@opentelemetry/auto-instrumentations-node/register' index.js

// Require in rest of modules
const express = require('express');
const axios = require('axios');

// Setup express
const app = express();
const PORT = 8080;

const getCrudController = () => {
  const router = express.Router();
  const resources = [];
  router.get('/', (req, res) => res.send(resources));
  router.post('/', (req, res) => {
    resources.push(req.body);
    return res.status(201).send(req.body);
  });
  return router;
};

const authMiddleware = (req, res, next) => {
  const { authorization } = req.headers;
  if (authorization && authorization.includes('secret_token')) {
    next();
  } else {
    res.sendStatus(401);
  }
};

app.use(express.json());
app.get('/health', (req, res) => res.status(200).send("HEALTHY")); // endpoint that is called by framework/cluster
app.get('/run_test', async (req, res) => {
  // Calls another endpoint of the same API, somewhat mimicking an external API call
  const createdCat = await axios.post(`http://localhost:${PORT}/cats`, {
    name: 'Tom',
    friends: [
      'Jerry',
    ],
  }, {
    headers: {
      Authorization: 'secret_token',
    },
  });

  return res.status(201).send(createdCat.data);
});
[...]

PR #2321, opened 2024-07-08
fix(detector-gpc): suppress tracing when requesting for GCP metadata

david-luna:

This is a 1st PR for fixing #2320. Some of the detectors are using `http` or other instrumented modules to get resource related data. In this case GCP resource detector is using `http` module to get info and this results on some spans being exported. These spans are not related to the instrumented service so it shouldn't be there

- Update `GcpDetector` class to implement `DetectorSync` interface
- use `suppresTracing` and `unsuppresTracing` APIs to make sure spans are not exported

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-gcp/test/detectors/GcpDetector.test.ts
+++ b/detectors/node/opentelemetry-resource-detector-gcp/test/detectors/GcpDetector.test.ts
@@ -31,7 +31,6 @@ import {
   assertContainerResource,
   assertEmptyResource,
 } from '@opentelemetry/contrib-test-utils';
-import { Resource } from '@opentelemetry/resources';
 
 const HEADERS = {
   [HEADER_NAME.toLowerCase()]: HEADER_VALUE,
@@ -85,7 +84,10 @@ describe('gcpDetector', () => {
       const secondaryScope = nock(SECONDARY_HOST_ADDRESS)
         .get(INSTANCE_PATH)
         .reply(200, {}, HEADERS);
-      const resource: Resource = await gcpDetector.detect();
+
+      const resource = gcpDetector.detect();
+      await resource.waitForAsyncAttributes?.();
+
       secondaryScope.done();
       scope.done();
 
@@ -121,7 +123,10 @@ describe('gcpDetector', () => {
       const secondaryScope = nock(SECONDARY_HOST_ADDRESS)
         .get(INSTANCE_PATH)
         .reply(200, {}, HEADERS);
-      const resource = await gcpDetector.detect();
+
+      const resource = gcpDetector.detect();
+      await resource.waitForAsyncAttributes?.();
+
       secondaryScope.done();
       scope.done();
 
@@ -155,7 +160,10 @@ describe('gcpDetector', () => {
       const secondaryScope = nock(SECONDARY_HOST_ADDRESS)
         .get(INSTANCE_PATH)
         .reply(200, {}, HEADERS);
-      const resource = await gcpDetector.detect();
+
+      const resource = gcpDetector.detect();
+      await resource.waitForAsyncAttributes?.();
+
       secondaryScope.done();
       scope.done();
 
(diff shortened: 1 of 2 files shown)


detectors/node/opentelemetry-resource-detector-gcp/test/fixtures/use-gcp-detector.js
trentm, line 1: nit: missing license header, and mixed 4- and 2-space indents in this file. Unfortunately the eslint setup doesn't lint .js files.

trentm (approved):

detectors/node/opentelemetry-resource-detector-gcp/test/fixtures/use-gcp-detector.js
david-luna: thanks! maybe we could also make eslint work in js (but not in this PR)

trentm: Definite not this PR.

detectors/node/opentelemetry-resource-detector-gcp/test/detectors/GcpDetector.test.ts
```diff
@@ -167,8 +179,77 @@ describe('gcpDetector', () => {
...
     it('returns empty resource if not detected', async () => {
-      const resource = await gcpDetector.detect();
+      const resource = gcpDetector.detect();
+      await resource.waitForAsyncAttributes?.();
+
       assertEmptyResource(resource);
     });
   });
+
+  describe('internal tracing', () => {
```
trentm, line 189: (I guess I forgot to hit "save" in my first review.)

Wow this was a fair amount of effort to write a test case for this.
I don't know how others would feel, but I wouldn't require similar test code be added for suppressTracing usage in other detectors.

david-luna: Yup it was a fair amount of work 😓. But I needed
- the `sdk` running so I had to create the fixture
- to avoid `nock` since real requests won't happen so we won't be really testing
- `gcp-metadata` to get responses. I could let it fail but I found that the request time vary in a way that makes the test flaky.

there is another approach I can take but 1st I wanted some feeedback for this one. If we finally use this one I may be inclined to move the mock server to `@opentelemetry/contrib-test-utils`

david-luna: About the 3rd point in my previous comment I found that if the async attributes of the resource are resolved after calling `sdk.shutdown()` we loose some spans. Without knowing to much about the mechanics of the processors/exporters I'd guess that these spans are not yet in the queue that is flushed during shutdown. I think this could be a potential issue for apps that have a short execution time (thinking on lamba or similar)

david-luna: @trentm the other approach seems to work and it's shorter 🎉 (although we're getting a resource with empty values). Changes are
- remove the mock server
- trigger the detection again and `waitForAsyncAttributes` 

```diff
diff --git a/detectors/node/opentelemetry-resource-detector-gcp/test/fixtures/use-gcp-detector.js b/detectors/node/opentelemetry-resource-detector-gcp/test/fixtures/use-gcp-detector.js
index 0c71a081..f88d2ae4 100644
--- a/detectors/node/opentelemetry-resource-detector-gcp/test/fixtures/use-gcp-detector.js
+++ b/detectors/node/opentelemetry-resource-detector-gcp/test/fixtures/use-gcp-detector.js
@@ -78,6 +78,9 @@ server.listen(0, '127.0.0.1', async function () {
   });

   // flush any left spans
+  // NOTE: this adds extra requests but its necessary to make sure
+  // spans have the resouce and are queued in the exporter
+  await gcpDetector.detect().waitForAsyncAttributes();
   await sdk.shutdown();
   await new Promise(resolve => server.close(resolve));
 });
```

A couple of observations:
- test becomes slower since the requests are made again at the end of the fixture code
- this detector just falls back to empty string if the request fails but I'm not sure if this will work on other detectors

david-luna:
After a discussion with @trentm I've decided to do a different approach:
- update detectors to the new `DetectrSync` interface. a PR per detector so they are small and easy to review.
- once merged add `suppressTracing` to the ones that need it

PR #2325, opened 2024-07-11
test(auto-instr-node): show result of setting both env vars

JamieDanielson:

- More clarity around what happens if both environment variables are set for enabling and disabling instrumentations

- Add debug message describing disabled instrumentations if both `OTEL_NODE_ENABLED_INSTRUMENTATIONS` and `OTEL_NODE_DISABLED_INSTRUMENTATIONS` environment variables are both set.
- Add test asserting that instrumentation is disabled when both environment variables are set.

Note this test is intended to show current state. If we want to change the behavior so that one wins over the other, that is an option, but is not the purpose of this PR.


metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -149,6 +149,12 @@ export function getNodeAutoInstrumentations(
   const enabledInstrumentationsFromEnv = getEnabledInstrumentationsFromEnv();
   const disabledInstrumentationsFromEnv = getDisabledInstrumentationsFromEnv();
 
+  if (enabledInstrumentationsFromEnv && disabledInstrumentationsFromEnv) {
```
maryliag, line 152: Can you add this clarification to the [docs](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/auto-instrumentations-node)? This behaviour was not clear (at least to me)

metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -149,6 +149,12 @@ export function getNodeAutoInstrumentations(
...
+    diag.debug(
+      'OTEL_NODE_ENABLED_INSTRUMENTATIONS and OTEL_NODE_DISABLED_INSTRUMENTATIONS environment variables are mutually exclusive. All instrumentations are disabled.'
+    );
+  }
```
trentm, line 156: But this isn't the actual behaviour, is it?

```
% cd ~/src/opentelemetry-js-contrib/metapackages/auto-instrumentations-node
% node
> var mod = require('./')
undefined
> process.env.OTEL_NODE_ENABLED_INSTRUMENTATIONS = 'dns,fs,hapi'
'dns,fs,hapi'
> mod.getNodeAutoInstrumentations().map(instr => instr.instrumentationName)
[
  '@opentelemetry/instrumentation-dns',
  '@opentelemetry/instrumentation-fs',
  '@opentelemetry/instrumentation-hapi'
]
> process.env.OTEL_NODE_DISABLED_INSTRUMENTATIONS = 'bunyan,fs'
'bunyan,fs'
> mod.getNodeAutoInstrumentations().map(instr => instr.instrumentationName)
[
  '@opentelemetry/instrumentation-dns',
  '@opentelemetry/instrumentation-hapi'
]
```

(See related comment at https://github.com/open-telemetry/opentelemetry.io/pull/4824#discussion_r1680436482

trentm (changes requested):

metapackages/auto-instrumentations-node/src/utils.ts
JamieDanielson: Huh yeah that was a poor test, not sure what I was doing 😅 . Will update!

metapackages/auto-instrumentations-node/src/utils.ts
JamieDanielson: Good callout! I've updated the README to match what will live in public docs. Also behavior probably wasn't clear because... it wasn't accurate 😅  I had some things mixed up when I was working on this but should be fixed up now!

trentm: nit (<--- only a nit because I don't want to derail this PR): I wouldn't say no to a change that drops this diag.debug completely. I think the behaviour (start with enabled-list, then subtract disabled-list) isn't surprising, and I don't think it is unreasonable usage to have both envvars set.   Keeping it is fine, too. :)

metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -151,7 +151,7 @@ export function getNodeAutoInstrumentations(
 
   if (enabledInstrumentationsFromEnv && disabledInstrumentationsFromEnv) {
     diag.debug(
-      'OTEL_NODE_ENABLED_INSTRUMENTATIONS and OTEL_NODE_DISABLED_INSTRUMENTATIONS environment variables are mutually exclusive. All instrumentations are disabled.'
+      'OTEL_NODE_ENABLED_INSTRUMENTATIONS and OTEL_NODE_DISABLED_INSTRUMENTATIONS are both set.\nAny instrumentation that exists in both lists will be disabled.'
```
trentm, line 154: nit:
```suggestion
      'OTEL_NODE_ENABLED_INSTRUMENTATIONS and OTEL_NODE_DISABLED_INSTRUMENTATIONS are both set. Any instrumentation that exists in both lists will be disabled.'
```

Multiline log message strings are a mild pain for processing (somewhat-)structured logs.  This is perhaps just personal preference, however.

trentm (approved):

maryliag (approved):

metapackages/auto-instrumentations-node/src/utils.ts
JamieDanielson: Per your other message... I'm totally onboard with dropping the debug altogether. I only added it when I was surprised by the disabling behavior. Now that I realize my brain was somewhere else that day, I agree that it doesn't help much. Will drop.

metapackages/auto-instrumentations-node/src/utils.ts
JamieDanielson: I'm totally onboard with dropping the debug altogether. I only added it when I was surprised by the disabling behavior. Now that I realize my brain was somewhere else that day, I agree that it doesn't help much. Will drop.

metapackages/auto-instrumentations-node/src/utils.ts
JamieDanielson: Also good to know re: pain for logs. I didn't initially have this as two lines when I updated it, but wasn't sure if it was too long so I split it. But will keep in mind for future.

blumamir (approved):

PR #2329, opened 2024-07-12
feat(detector-aws): add DetectorSync implementation for Lambda detector

david-luna:
## Context

Another PR moving away from the deprecated `Detector` interface in favour to the `DetectorSync` interface. This time is for AWS Lambda detector and it might help a bit in https://github.com/open-telemetry/opentelemetry-js/issues/3582

- add `AwsLambdaDetectorSync` class and instance
- make `AwsLambdaDetector` reuse the new sync instance
- update tests

Note: since this package is marked as stable (major version is 1) we are doing a non breaking change by adding the new class and mark the current as `deprecated`. The async detector changes behavior, it does not throw anymore, and it was discussed in last SIG meeting (2024-07-31). The outcome was we're okay with that.

A couple of PRs will follow:
- one to suppress internal tracing.
- one to add the sync detector to auto-instrumentations package.


metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -235,7 +235,7 @@ function getDisabledInstrumentationsFromEnv() {
 export function getResourceDetectorsFromEnv(): Array<Detector | DetectorSync> {
   const resourceDetectors = new Map<
     string,
-    Detector | DetectorSync | Detector[] | DetectorSync[]
+    Detector | DetectorSync | Detector[] | DetectorSync[] | Array<Detector | DetectorSync>
```
david-luna, line 238: note for reviewer: this is necessary to allow this single refactor instead of doing all AWS detector at once.

metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -235,7 +235,11 @@ function getDisabledInstrumentationsFromEnv() {
 export function getResourceDetectorsFromEnv(): Array<Detector | DetectorSync> {
   const resourceDetectors = new Map<
     string,
-    Detector | DetectorSync | Detector[] | DetectorSync[]
+    | Detector
```
david-luna, line 238: note for reviewer: this is necessary to allow this single refactor instead of doing all AWS detector at once.

legendecas:
The @opentelemetry/resources version requirement in the package.json needs to be updated as ^[1.10.0](https://github.com/open-telemetry/opentelemetry-js/releases/tag/v1.10.0).

metapackages/auto-instrumentations-node/src/utils.ts
blumamir: Why not do all the AWS detectors at once? What is the gain in opening so many PRs with the exact same refactor?

I personally would prefer to have all detectors in the repo updated in one PR, but at least the ones from the same npm package can be grouped together, no?

david-luna: >Why not do all the AWS detectors at once? What is the gain in opening so many PRs with the exact same refactor?
>
>I personally would prefer...

as you said is more a preference although I find some benefits of splitting. We can discuss each detector independently and allow other detectors to be merged in case one discussion takes long time.

it's my preference but I'm fine in adding up all the changes in a single big PR. I preferred that way to let the owners

david-luna:
closed in favour of #2376 following [this suggestion](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2329#discussion_r1687558761)

PR #2331, opened 2024-07-12
feat(detector-aws): add DetectorSync implementation for ECS detector

david-luna:
## Context

Another PR moving away from the deprecated `Detector` interface in favour to the `DetectorSync` interface. This time is for AWS ECS detector and it might help a bit in https://github.com/open-telemetry/opentelemetry-js/issues/3582

- add `AwsEcsDetectorSync` class and instance
- make `AwsEcsDetector` reuse the new sync instance
- update tests

Note: since this package is marked as stable (major version is 1) we are doing a non breaking change by adding the new class and mark the current as `deprecated`. The async detector changes behavior, it does not throw anymore, and it was discussed in last SIG meeting (2024-07-31). The outcome was we're okay with that.

A couple of PRs will follow:
- one to suppress internal tracing.
- one to add the sync detector to auto-instrumentations package.


metapackages/auto-instrumentations-node/src/utils.ts
```diff
@@ -235,7 +235,11 @@ function getDisabledInstrumentationsFromEnv() {
 export function getResourceDetectorsFromEnv(): Array<Detector | DetectorSync> {
   const resourceDetectors = new Map<
     string,
-    Detector | DetectorSync | Detector[] | DetectorSync[]
+    | Detector
```
david-luna, line 238: note for reviewer: this is necessary to allow this single refactor instead of doing all AWS detector at once.

david-luna:
closed in favour of #2376 following [this suggestion](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2329#discussion_r1687558761)

PR #2328, opened 2024-07-12
feat(detector-alibaba)!: change implementation to DetectorSync interface

david-luna:
## Context

In order to fix #2320 we need to suppress tracing when this detector is doing HTTP requests or using any other instrumented API. IMO this would be easier if we already have the detectors implementing the `DetectorSync` interface moving away from the deprecated `Detector`. This PR is adding a new sync detector that will be used by the SDK and auto-instrumentations package (to be changed in a follow-up PR). 

Follow-up PRs will be created to:
- suppressTracing
- use the sync detector in auto-instrumentations

## Short description of the changes

- Make the current detector implement the `DetectorSync` interface
- update tests

### Test changes

The new detector's `detect` method becomes synchronous and therefore there is no way to throw an exception in it for an async HTTP error. The refactor makes this detector to return and empty resource if an async error occurs, so the test does not expect an exception anymore but an empty resource.

Also this aligns with `@opentelemetry/resources` logic used by the NodeSDK which catches any exception from the detector and returns an empty resource.
https://github.com/open-telemetry/opentelemetry-js/blob/54b14fbbe33e0cf38115ee8c5c934a4e27786186/packages/opentelemetry-resources/src/detect-resources.ts#L32

PR #2334, opened 2024-07-12
feat(detector-container)!: change implementation to DetectorSync interface

david-luna:
## Context

Another PR moving away from the deprecated `Detector` interface in order to prepare the codebase for a fix for #2320. This time is for `Container` detector.

My reasons to doing this change:

- update tests


pichlermarc:
FYI @abhee11(compoenent owner)

blumamir (approved):

PR #2335, opened 2024-07-12
feat(detector-gcp)!: change implementation to DetectorSync interface

david-luna:
## Context

Another PR moving away from the deprecated `Detector` interface in order to prepare the codebase for a fix for #2320. This time is for `GCP` detector.

My reasons to do this change:

- update tests


blumamir (commented):
LGTM but I think we should consider changing it to `feat!` instead

blumamir (approved):

PR #2336, opened 2024-07-12
feat(detector-github)!: change implementation to DetectorSync interface

david-luna:
## Context

Another PR moving away from the deprecated `Detector` interface in order to prepare the codebase for a fix for #2320. this time is for `GitHub` detector.

My reasons to categorize it as a refactor are:


pichlermarc:
I'll sponsor this.

pichlermarc (approved):
Thanks for updating this :+1:

pichlermarc:
(changed the title to reflect that this is a breaking change :slightly_smiling_face:

trentm:
@pichlermarc FYI: David was going to bring up some Qs regarding all the Detector -> DetectorSync PRs in the JS SIG tomorrow. There is some more discussion on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2328

david-luna:
> @pichlermarc FYI: David was going to bring up some Qs regarding all the Detector -> DetectorSync PRs in the JS SIG tomorrow. There is some more discussion on #2328

On the SIG we agreed to break for non stable detectors like this one. So I'm merging this one :)

PR #2337, opened 2024-07-12
feat(detector-instana)!: change implementation to DetectorSync interface

david-luna:
## Context

Another PR moving away from the deprecated `Detector` interface in order to prepare the codebase for a fix for #2320. This time is for `Instana` detector.

My reasons to categorize it as a refactor are:

- update tests

### Test changes

Previously the test where expecting the detector to throw but with the new implementation is not possible throw in a synchornous `detect` method when an async HTTP request fails. The behaviour has been changed to return an empty resource instead. This change is compatible with the current SDK since it actually does the same logic when a detector fails. You can check the `detectResources` implementation link above.


detectors/node/opentelemetry-resource-detector-instana/src/detectors/InstanaAgentDetector.ts
```diff
@@ -13,31 +13,44 @@
...
+} from '@opentelemetry/resources';
 import { diag } from '@opentelemetry/api';
 import {
   SEMRESATTRS_PROCESS_PID,
   SEMRESATTRS_SERVICE_INSTANCE_ID,
 } from '@opentelemetry/semantic-conventions';
 import * as http from 'http';
 
-class InstanaAgentDetector implements Detector {
+class InstanaAgentDetector implements DetectorSync {
```
blumamir, line 29: While very unlikely, there might be users out there which can potentially break due to this change. 
Is there any reason to avoid marking it with `feat!` just in case?

detectors/node/opentelemetry-resource-detector-instana/test/InstanaAgentDetectorUnitTest.test.ts
```diff
@@ -90,28 +91,24 @@ describe('[UNIT] instanaAgentDetector', () => {
...
+      const resource = instanaAgentDetector.detect();
+      await resource.waitForAsyncAttributes?.();
+
+      assert.deepStrictEqual(resource.attributes, {});
 
       scope.done();
     });
 
-    it('agent timeout', async () => {
+    it('agent retuen emrty resource if timeout', async () => {
```
blumamir, line 108:
```suggestion
    it('agent return empty resource if timeout', async () => {
```

blumamir (commented):
LGTM, but I wonder if we should still mark it as breaking just in case.

david-luna:
@basti1302 @kirrg001 is it okay for you to have this breaking change? If not we I can instead add a new detector implementing the sync interface

basti1302:
Hi David, I do no longer work for IBM/Instana, so I don't have any particular opinion about this.

david-luna:
Thanks @basti1302. Then we will go for the breaking change

blumamir (approved):

blumamir:
As discussed in the last SIG and documented on the PR, it is very unlikely that people out there are calling the `InstanaAgentDetector` directly and will be affected by changing the `implements Detector` to `implements DetectorSync`.

Since the package is still experimental (major version 0), we can make this change as breaking.

@open-telemetry/javascript-approvers any objections before I merge this PR?

pichlermarc:
> As discussed in the last SIG and documented on the PR, it is very unlikely that people out there are calling the `InstanaAgentDetector` directly and will be affected by changing the `implements Detector` to `implements DetectorSync`.
> 
> Since the package is still experimental (major version 0), we can make this change as breaking.
> 
> @open-telemetry/javascript-approvers any objections before I merge this PR?

I'm mainly waiting for someone from Instana to comment. Maybe we give it a couple more days until we merge this to ensure they have had enough time to comment.

blumamir:
> > As discussed in the last SIG and documented on the PR, it is very unlikely that people out there are calling the `InstanaAgentDetector` directly and will be affected by changing the `implements Detector` to `implements DetectorSync`.
> > Since the package is still experimental (major version 0), we can make this change as breaking.
> > @open-telemetry/javascript-approvers any objections before I merge this PR?
> 
> I'm mainly waiting for someone from Instana to comment. Maybe we give it a couple more days until we merge this to ensure they have had enough time to comment.

 @kirrg001 can you please take a look?

david-luna:
> > > As discussed in the last SIG and documented on the PR, it is very unlikely that people out there are calling the `InstanaAgentDetector` directly and will be affected by changing the `implements Detector` to `implements DetectorSync`.
> > > Since the package is still experimental (major version 0), we can make this change as breaking.
> > > @open-telemetry/javascript-approvers any objections before I merge this PR?
> > 
> > 
> > I'm mainly waiting for someone from Instana to comment. Maybe we give it a couple more days until we merge this to ensure they have had enough time to comment.
> 
> @kirrg001 can you please take a look?

@kirrg001 did you have the chance to look at this PR?

Issue #2339, opened 2024-07-15
Container detector - cgroupv1 and v2 - prefix support & fix


abhee11:
### Is your feature request related to a problem? Please describe

There is an issue where sometimes using lightweight k8 cgroup v1 or v2 - prefix may be there in front of the containerId - current implementation does not account for it. 

### Describe the solution you'd like to see

Will need to add a check ( regex to filter crio, crio-containerd and containers prefix before the containerId extraction

abhee11:
Please assign it to me.

PR #2341, opened 2024-07-17
feat: added prefix to look for containerid

abhee11:

-https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2339

- adding some prefix check - if they exist to truncate them from the container id string -


abhee11:
@blumamir @rauno56 @dyladan Can I get a review or you can assign it to a relevant person

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -94,22 +81,70 @@ export class ContainerDetector implements Detector {
...
+        if (lastSection.startsWith(this.CRIO)) {
+          lastSection = this.truncatePrefix(lastSection, this.CRIO);
+        } else if (lastSection.startsWith(this.DOCKER)) {
+          lastSection = this.truncatePrefix(lastSection, this.DOCKER);
+        } else if (lastSection.startsWith(this.CRI_CONTAINERD)) {
+          lastSection = this.truncatePrefix(lastSection, this.CRI_CONTAINERD);
+        }
+        // Remove anything after the first period
+        if (lastSection.includes('.')) {
+          lastSection = lastSection.split('.')[0];
```
blumamir, line 123: the original code is using the last dot:

```js
        let startIdx = lastSection.lastIndexOf('-');
        let endIdx = lastSection.lastIndexOf('.');

        startIdx = startIdx === -1 ? 0 : startIdx + 1;
        if (endIdx === -1) {
          endIdx = lastSection.length;
        }
        if (startIdx > endIdx) {
          continue;
        }
```

Can you elaborate about the motivation for this change, perhaps also include some examples of what this`section` might be formatted like to give more context for the review and future maintainers?

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -51,35 +56,17 @@ export class ContainerDetector implements Detector {
...
-        continue;
-      }
-      const lastSection = line.substring(lastSlashIdx + 1);
-      const colonIdx = lastSection.lastIndexOf(':');
-      if (colonIdx !== -1) {
-        // since containerd v1.5.0+, containerId is divided by the last colon when the cgroupDriver is systemd:
-        // https://github.com/containerd/containerd/blob/release/1.5/pkg/cri/server/helpers_linux.go#L64
-        return lastSection.substring(colonIdx + 1);
-      } else {
-        let startIdx = lastSection.lastIndexOf('-');
```
blumamir, line 72: was the search for `-` intended to remove the prefixes (`docker-` etc)?
can the new code now fail to remove such prefixes for something which is not in the predefined list of know prefixes?

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -94,22 +81,70 @@ export class ContainerDetector implements Detector {
...
+  protected extractContainerIdFromLine(line: string): string | null {
+    if (line) {
+      const sections = line.split('/');
+      if (sections.length > 1) {
+        let lastSection = sections[sections.length - 1];
+
+        // Handle containerd v1.5.0+ format with systemd cgroup driver
+        const colonIndex = lastSection.lastIndexOf(':');
+        if (colonIndex !== -1) {
+          lastSection = lastSection.substring(colonIndex + 1);
```
blumamir, line 110: previous code behavior was to return in this case, and avoid any other manipulations on the `lastSection`.

Since I don't have any prior knowledge on how containers are named and formatted, it's difficult to review this change.
Can you please describe the reasoning behind these changes, preferably with examples

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -94,22 +81,70 @@ export class ContainerDetector implements Detector {
       .trim()
       .split('\n')
       .find(s => s.includes(this.HOSTNAME));
-    const containerIdStr = str
-      ?.split('/')
-      .find(s => s.length === this.CONTAINER_ID_LENGTH);
-    return containerIdStr || '';
+
+    const strArray = str?.split('/') ?? [];
+    for (let i = 0; i < strArray?.length; i++) {
```
blumamir, line 86: if `strArray` is `undefined`, then `strArray?.length` is also `undefined`. not sure what will come up from `<` a number to `undefined`, but I don't think we should keep it.
Actually, since strArray is nullish coalesced, it cannot be `undefined` so using optional chaining is not needed here.

you can choose to refactor this section to simplify the flow in case `str` is empty, and then return instead of iterating an empty list.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -94,22 +81,70 @@ export class ContainerDetector implements Detector {
...
+      if (
+        strArray[i] === this.MARKING_PREFIX &&
+        strArray[i + 1]?.length === this.CONTAINER_ID_LENGTH
```
blumamir, line 89: for the last iteration, `i+1` is always out of range and will be undefined.
We can change the loop so it ends at `strArray.length - 1` instead of accessing index out of range.

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
```diff
@@ -146,5 +146,75 @@ describe('ContainerDetector', () => {
...
     });
+
+    describe(' extractContainerId from line tests', () => {
+      const containerDetector = new ContainerDetector();
+      it('should extract container ID from crio-prefixed line', () => {
+        const line =
+          '11:devices:/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod5c5979ec_6b2b_11e9_a923_42010a800002.slice/crio-1234567890abcdef.scope';
+        const expected = '1234567890abcdef';
+        assert.strictEqual(
+          containerDetector['extractContainerIdFromLine'](line),
```
blumamir, line 157: Accessing the private / protected function this way feels wrong. 
WDYT about extracting the function to `utils.ts` instead of it being class member function. then we can test it directly in an idiomatic way.

blumamir (commented):
I have very little knowledge about how containers are stored and formatted in files. added few general suggestions / concerns.

The new tests seems not to cover some of the flows introduced by this PR. Please consider adding more tests.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
abhee11: we are taking care of more prefixes :   
  readonly CRIO = 'crio-';
  readonly CRI_CONTAINERD = 'cri-containerd-';
  readonly DOCKER = 'docker-';

earlier code only took care of docker - this change handles additional cases.

detectors/node/opentelemetry-resource-detector-container/test/ContainerDetector.test.ts
abhee11: it does feel wrong. I shall move the helper methods as per suggestion.

detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
abhee11: So the motivation for this change is as follows:

1. Add more cases to be handled ( more prefixes ) - previous case only handled a specific case
2. Adding more filter - knowing that v1 id is hexadecimal in nature - adding that check 
3. The new logic is essentially doing very similar but with more checks
4. Added a lot of tests and example in utils.test.ts

abhee11:
@blumamir Thank you for the review - I have added some comments and added more test cases in utils.test.ts This PR 

1. Covers more prefixes  ( previous was only docker ) this has more coverage.
2. Added additional hexadecimal checks. 
3. Added suffix check and improved the start and end consideration of what might be containerID 
4. Added more tests to cover new functionality .

blumamir (approved):
Thank you for working on this and addressing all the comments.
LGTM

Issue #2348, opened 2024-07-24
Prototype for Semantic Conventions for Database Metrics


maryliag:
Prototype implementation for [Semantic Conventions for Database Metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md).

Required for stabilization of semantic convention.

This first prototype will be implemented using Postgres.

PR #2349, opened 2024-07-24
initial implementation of prototype

maryliag:

- Part Of #2348 

- This PR implements part of the [Semantic Conventions for Database Metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md). I will break down the implementation on different PRs for each metric to facilitate the review process.
- This PR focus on the [db.client.connection.count](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientconnectioncount) metric

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
@@ -26,6 +26,8 @@ import {
   Span,
   SpanStatusCode,
   SpanKind,
+  MeterProvider,
+  UpDownCounter,
 } from '@opentelemetry/api';
 import type * as pgTypes from 'pg';
 import type * as pgPoolTypes from 'pg-pool';
@@ -43,8 +45,30 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 import { SpanNames } from './enums/SpanNames';
 
 export class PgInstrumentation extends InstrumentationBase {
+  private _connectionsCount!: UpDownCounter;
+  private _connectionsCounter: utils.poolConnectionsCounter = {
+    used: 0, idle: 0
+  };
+
   constructor(config: PgInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
+    this._setMetricInstruments();
+  }
+
+  override setMeterProvider(meterProvider: MeterProvider) {
+    super.setMeterProvider(meterProvider);
+    this._setMetricInstruments();
+  }
+
+  private _setMetricInstruments() {
+    this._connectionsCount = this.meter.createUpDownCounter(
+      'db.client.connection.count',
+      {
+        description:
+          'The number of connections that are currently in state described by the state attribute.',
+        unit: '{connection}',
+      }
+    );
   }
 
   protected init() {
@@ -96,16 +120,28 @@ export class PgInstrumentation extends InstrumentationBase {
         if (isWrapped(moduleExports.prototype.connect)) {
           this._unwrap(moduleExports.prototype, 'connect');
         }
+        if (isWrapped(moduleExports.prototype.end)) {
+          this._unwrap(moduleExports.prototype, 'end');
+        }
+
         this._wrap(
           moduleExports.prototype,
           'connect',
           this._getPoolConnectPatch() as any
+        );
+        this._wrap(
+          moduleExports.prototype,
+          'end',
+          this._getPoolEndPatch() as any
         );
         return moduleExports;
       },
       (moduleExports: typeof pgPoolTypes) => {
         if (isWrapped(moduleExports.prototype.connect)) {
           this._unwrap(moduleExports.prototype, 'connect');
+        }
+        if (isWrapped(moduleExports.prototype.end)) {
+          this._unwrap(moduleExports.prototype, 'end');
         }
       }
     );
@@ -323,7 +359,6 @@ export class PgInstrumentation extends InstrumentationBase {
               });
             });
         }
-
         // else returns void
         return result; // void
       };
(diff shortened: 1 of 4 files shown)


plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -482,4 +509,91 @@ describe('pg-pool', () => {
...
+              'idle'
+            );
+
+            assert.strictEqual(
+              metrics[1].descriptor.name,
+              'db.client.connection.pending_requests'
+            );
+            assert.strictEqual(
+              metrics[1].descriptor.description,
+              'The number of current pending requests for an open connection.'
```
pichlermarc, line 592: is it on purpose that this test never asserts on the actual data point values? :thinking:

david-luna:
@maryliag 

A part from @pichlermarc suggestion about separators the rest LGTM :)

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -258,6 +264,50 @@ export function patchCallback(
   };
 }
 
+export function getPoolName(pool: PgPoolOptionsParams): string {
+  let poolName = '';
+  poolName += pool?.host ? `${pool.host}:` : 'unknown_host';
+  poolName += pool?.port ? `${pool.port}/` : 'unknown_port';
```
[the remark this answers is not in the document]
maryliag: Good catch! Fixed

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -482,4 +509,91 @@ describe('pg-pool', () => {
...
+    function initMeterProvider() {
+      otelTestingMeterProvider = new MeterProvider();
+      inMemoryMetricsExporter = new InMemoryMetricExporter(
+        AggregationTemporality.CUMULATIVE
+      );
+      const metricReader = new PeriodicExportingMetricReader({
+        exporter: inMemoryMetricsExporter,
+        exportIntervalMillis: 100,
+        exportTimeoutMillis: 100,
+      });
```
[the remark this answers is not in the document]
maryliag: Thanks for the info, I'll take a look into it!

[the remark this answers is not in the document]
maryliag: Btw, I copied this implementation from both mongo and mysql tests, since I'm changing on a similar place. Should all these places be update to use the approach you mentioned then?

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -258,6 +264,50 @@ export function patchCallback(
...
+    [SEMATTRS_CLIENT_CONNECTION_POOL_NAME]: poolName,
+  });
+
+  connectionPendingRequests.add(pending - latestCounter.pending, {
+    [SEMATTRS_CLIENT_CONNECTION_POOL_NAME]: poolName,
+  });
+
+  return { used: used, idle: idle, pending: pending };
+}
+
```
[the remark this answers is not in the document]
maryliag: I can always bring this feedback to the group. For this PR I'll keep as the `UpDownCounter` and we can always change if comes to it (since it's experimental)

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
maryliag: @pichlermarc  Not on purpose. I added the check on the datapoints now, but I'm having an issue. When I tested manually, everything was working as expected, but now this test is failing because it says the data point for used is `-1`, instead of the correct `1`. I printed the values getting added to the metric and it has all the right values, adding the correct 1 when it starts and removing it when the connection is no longer being used.

The data point for the used connection is like this, which I noticed has start and end times, could this aggregation be causing the value to change? I can't figure it out what else could be wrong on this. Open to suggestions and ideas
```
{
    attributes: {
      'db.client.connection.state': 'used',
      'db.client.connection.pool.name': 'localhost:54320/postgres'
    },
    startTime: [ 1723575108, 137000000 ],
    endTime: [ 1723575108, 486000000 ],
    value: -1
  },
```

maryliag: After some updates, now the value is 0, instead of -1, but still not the expected 1

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -43,8 +45,46 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
...
+
+  private _setMetricInstruments() {
+    this._connectionsCount = this.meter.createUpDownCounter(
+      'db.client.connection.count',
+      {
+        description:
+          'The number of connections that are currently in state described by the state attribute.',
+        unit: '{connection}',
+      }
+    );
```
pichlermarc, line 79: Took me quite a while to figure that out, the test is failing as we don't reset the internally kept state when the MeterProvider is reset. This means the internal state assumes the counter is at `1` when it really is at `0` as the old counter which was at `1` has been replaced by a new one.

```suggestion
    this._connectionsCounter = {
      idle: 0,
      pending: 0,
      used: 0,
    }
    this._connectionsCount = this.meter.createUpDownCounter(
      'db.client.connection.count',
      {
        description:
          'The number of connections that are currently in state described by the state attribute.',
        unit: '{connection}',
      }
    );
```

Another reason why I think the semconv should switch to a synchronous gauge instead for this. :sweat_smile:

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -43,8 +45,51 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
...
+
   constructor(config: PgInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
+    this._setMetricInstruments();
+  }
+
+  override setMeterProvider(meterProvider: MeterProvider) {
+    super.setMeterProvider(meterProvider);
+    this._setMetricInstruments();
+  }
```
pichlermarc, line 69: Hmm, this overrides the base method but basically does the same as the base implementation - would we loose any functionality by using `_updateMetricInstruments`? :thinking:

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -43,8 +45,51 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
...
+  // when the value change.
+  private _connectionsCounter: utils.poolConnectionsCounter = {
+    used: 0,
+    idle: 0,
+    pending: 0,
+  };
+
   constructor(config: PgInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
+    this._setMetricInstruments();
```
pichlermarc, line 63: If overwritten, `_updateMetricInstruments()` would already be called by `super()`.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
maryliag: Not lose anything, so I changed to use the `_updateMetricInstruments` instead

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
maryliag: Updated and removed this line, now that I'm using `_updateMetricInstruments`

pichlermarc (approved):

david-luna (approved):

maryliag:
@pichlermarc friendly ping if this can be merged

PR #2356, opened 2024-07-29
feat(instrumentation-undici): Add `responseHook`

### LINKED ISSUE #2179: [@opentelemetry/instrumentation-undici] make http.response.status_code available for use in hooks

Good morning all, I tried to use the 2 exposed hooks `requestHook` and `startSpanHook` from @opentelemetry/instrumentation-undici in order to access `http.response.status_code` from the span and copy its value into `http.status_code` for compatibility with our systems. From my understanding, this hooks run at a point in time before the monitored request actually concludes, so there is no way currently to achieve this.

A hook that allows you to modify the span in terms of attributes after the request succeeded or failed.

### Additional context

```js
new UndiciInstrumentation({
        requestHook: (span, request) => {
          span.setAttribute(
            'http.host',
            span.attributes['server.address'] +
              ':' +
              span.attributes['server.port']
          )
          span.setAttribute(
            'http.method',
            span.attributes['http.request.method']
          )
          span.setAttribute(
            'http.target',
            span.attributes['url.path'] + span.attributes['url.query']
          )
          span.setAttribute('http.url', span.attributes['url.full'])
          
          // http.response.status_code does not exist yet
          span.setAttribute(
            'http.status_code',
            parseInt(span.attributes['http.response.status_code'])
          )
        },
      })
```

### LINKED ISSUE #2265: Add hook to access response in instrumentation-undici
It would be great to have a hook that allows us access to request & response in instrumentation-undici. 

Would you be open to add support for this? If so, I could work on a PR adding this. I am thinking about something like this which exists in instrumentation-http:

```ts
applyCustomAttributesOnSpan: (
  span,
  request,
  response,
) : void
```

Alternatively, if we could make certain things protected (e.g. `subscribeToChannel`) we could also extend the instrumentation class and add such things ourselves. If that's preferable I could also PR this?

timfish:
- Closes #2265
- Closes #2179

This PR adds a `responseHook` to the instrumentation config


mydea (approved):
Looks great from my POV! Would be super helpful.

plugins/node/instrumentation-undici/src/types.ts
```diff
@@ -51,18 +52,29 @@ export interface RequestHookFunction<T = UndiciRequest> {
   (span: Span, request: T): void;
 }
 
+export interface ResponseHookFunction<
+  RequestType = UndiciResponse,
+  ResponseType = UndiciResponse
+> {
+  (span: Span, request: RequestType, response: ResponseType): void;
```
blumamir, line 59: Please use an `Info` object to store all relevant hook data. 
In the future, more arguments might be added which will make the usage harder.

See example [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/80304bcd1bb4654499fb4f67e804d142f0284d21/plugins/node/instrumentation-amqplib/src/types.ts#L44)

plugins/node/instrumentation-undici/src/types.ts
```diff
@@ -51,18 +52,29 @@ export interface RequestHookFunction<T = UndiciRequest> {
...
+export interface UndiciInstrumentationConfig<
+  RequestType = UndiciRequest,
+  ResponseType = UndiciResponse
+> extends InstrumentationConfig {
   /** Not trace all outgoing requests that matched with custom function */
   ignoreRequestHook?: IgnoreRequestFunction<RequestType>;
   /** Function for adding custom attributes before request is handled */
   requestHook?: RequestHookFunction<RequestType>;
+  /** Function called once response headers have been received */
+  responseHook?: ResponseHookFunction<RequestType, ResponseType>;
```
blumamir, line 77: need to also be added to [README](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/instrumentation-undici#undicifetch-instrumentation-options)

plugins/node/instrumentation-undici/src/types.ts
```diff
@@ -41,6 +41,7 @@ export interface UndiciRequest {
 export interface UndiciResponse {
   headers: Buffer[];
   statusCode: number;
+  statusText: string;
```
blumamir, line 44: What is the motivation to add this field? #2179 mentions status code and not text

blumamir (commented):
LGTM.

added few minors

plugins/node/instrumentation-undici/src/types.ts
timfish: The response object contains this property so why not expose it in the types? 

Users could spread the response into a new object not knowing it has this property:
```ts
const data = {
  ...someOtherObj,
  ...response,
}
```

plugins/node/instrumentation-undici/src/types.ts
david-luna: If I may ask something is to keep the generics since they give flexibility to the devs to pass the type specific to the `undici` version they're using

plugins/node/instrumentation-undici/src/types.ts
david-luna: I'd say it's fine to have this field if it's available in all supported versions. If not you can make it optional.

plugins/node/instrumentation-undici/test/fetch.test.ts
```diff
@@ -281,6 +287,11 @@ describe('UndiciInstrumentation `fetch` tests', function () {
         'hook-value',
         'startSpanHook is called'
       );
+      assert.strictEqual(
+        span.attributes['test.response-hook.attribute'],
+        'OK',
+        'startSpanHook is called'
```
david-luna, line 293: copy paste from previous assertion?

david-luna (commented):
Hi @timfish

Thanks for contributing to the `undici` instrumentation. The PR looks very good :)

The new hook is executed upon `undici:request:headers` diagnostics channel publishes a message when headers are received but there is no guarantee the body is available yet. I wonder if `responseHook` name will let devs think they can inspect the whole response (body included).

plugins/node/instrumentation-undici/src/types.ts
timfish: As far as I can tell, this has existed as long as the diagnostic channel has had type defs:
https://github.com/nodejs/undici/blame/9bc9b44a24d7869f568103b4c7cf428755601157/types/diagnostics-channel.d.ts#L16

timfish:
> I wonder if `responseHook` name will let devs think they can inspect the whole response (body included).

The http instrumentation uses a hook with the same name which is called before the body is received:
https://github.com/open-telemetry/opentelemetry-js/blob/30a46ae5471f9268c2cd241a805e8e68e3ab4f28/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L332

Although the readme docs for it say:
> Function for adding custom attributes before response is handled

So maybe that is a better description?

plugins/node/instrumentation-undici/src/types.ts
blumamir: Ok, I thought it was related to an issue.
Not sure when it is useful, but don't mind having it in the hook 👍

blumamir:
> > I wonder if `responseHook` name will let devs think they can inspect the whole response (body included).
> 
> The http instrumentation uses a hook with the same name which is called before the body is received: https://github.com/open-telemetry/opentelemetry-js/blob/30a46ae5471f9268c2cd241a805e8e68e3ab4f28/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L332
> 
> Although the readme docs for http say:
> 
> > Function for adding custom attributes before response is handled
> 
> So maybe that is a better description than mine?
> 
> > Function for adding custom attributes after the response headers are received.

I think hooks are very popular to record payloads. with http instrumentation, it is possible to collect the payload in the `responseHook`, but is it possible with `undici`? or will we need a different hook for accessing the response payload?

timfish:
> I think hooks are very popular to record payloads. with http instrumentation, it is possible to collect the payload in the responseHook, but is it possible with undici?

How are they doing this? As far as I can see, the http instrumentation does not offer an easy way to access the response payload from outgoing requests. 

It's possible to get the response buffers from undici via the [`undici:request:trailers`](https://undici.nodejs.org/#/docs/api/DiagnosticsChannel?id=undicirequesttrailers) diagnostic channel event which in the undici plugin currently just calls `onDone` to end the span. The downside to this event is that it doesn't send pay you the request object. It's nice to have a single event with both request and response at the same time so you can make more granular choices.

trentm:
[Amir]
> with http instrumentation, it is possible to collect the payload in the `responseHook`

Are you sure? I don't think it is. The `responseHook` is being called synchronously in the 'response' event handler (https://nodejs.org/api/http.html#event-response). The response body is not yet available at this point.

timfish:
The body isn't available at that point, but you have access to the response which is a stream, so you could `response.on('data', (d) => {})`.

Is anyone actually doing this just access the resulting buffer in the `responseHook`? What are the use cases?

plugins/node/instrumentation-undici/README.md
```diff
@@ -51,11 +51,12 @@ Undici instrumentation has few options available to choose from. You can set the
...
-| [`headersToSpanAttributes`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-undici/src/types.ts#L71) | `Object`                | List of case insensitive HTTP headers to convert to span attributes. Headers will be converted to span attributes in the form of `http.{request\|response}.header.header-name` where the name is only lowercased, e.g. `http.response.header.content-length` |
+| [`ignoreRequestHook`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-undici/src/types.ts#L73)       | `IgnoreRequestFunction` | Undici instrumentation will not trace all incoming requests that matched with custom function.                                                                                                                                                               |
```
trentm, line 54: IIUC, these updates are to all the line numbers. Thanks for doing that.
In a separate change I think we should drop those links. It is too much of a maintenance burden.

trentm (approved):
LGTM

Waiting for a bit before merging to give @blumamir a chance to follow-up on earlier comments and @david-luna  to review.

plugins/node/instrumentation-undici/README.md
david-luna: I agree

david-luna (approved):
I guess if we get questions/issues about body not being available in the hook we can always update the description and highlight it.

foklepoint:
heyo!
Wanted to share that Rye needed undici response bodies in our traces. Throwing our implementation of using these newly added hooks (Thnks @timfish). It's not a great implementation by any standard and lots of jankiness trying to access body. But just in case someone wants to improve upon this or use it.  Tried the path with the undici::trailers mentioned in this thread and just wanted to call out had a really hard time trying to get the body off that. 

Drawbacks of the approach below - couldn't quite get the request body to be added to the trace spans and janky async added on hooks to make type system happy. 

```
import { UndiciRequest, UndiciResponse } from '@opentelemetry/instrumentation-undici';
import stringify from 'fast-safe-stringify';
import * as zlib from 'zlib';

// Tees the body then restores it back to the original request object
async function readAndRestoreRequestBody(request: UndiciRequest) {
  const originalBody = request.body;
  const chunks: Uint8Array[] = [];

  for await (const chunk of originalBody) {
    chunks.push(chunk);
  }

  // Restore the body
  request.body = {
    async *[Symbol.asyncIterator]() {
      yield* chunks;
    },
  };

  return Buffer.concat(chunks).toString();
}

// Constants
const MAX_BODY_SIZE = 1024 * 1024; // 1MB

/**
 * Represents the options for the Undici instrumentation configuration.
 */
interface UndiciInstrumentationOptions {
  requestHeadersSpanAttrName?: string;
  responseHeaderSpanAttrName?: string;
  requestBodySpanAttrName?: string;
  responseBodySpanAttrName?: string;
}

/**
 * Represents the capture context for response data.
 */
interface CaptureContext {
  rawResponse: Buffer;
  span?: Span;
}

/**
 * Represents the handler for Undici instrumentation.
 */
class UndiciInstrumentationResponseBodyHandler {
  private context: CaptureContext;
  private responseHeaderSpanAttrName: string;
  private responseBodySpanAttrName: string;

  constructor(span: Span, responseHeaderSpanAttrName: string, responseBodySpanAttrName: string) {
    this.responseHeaderSpanAttrName = responseHeaderSpanAttrName;
    this.responseBodySpanAttrName = responseBodySpanAttrName;
    this.context = {
      rawResponse: Buffer.alloc(0),
      span,
    };
  }

  /**
   * Handles each data chunks from the response as they come in.
   */
  private handleChunk(chunk: Buffer): void {
    if (chunk.length > MAX_BODY_SIZE) {
      throw new Error(`Response body exceeded ${MAX_BODY_SIZE} bytes. Skipping further capturing.`);
    }
    this.context.rawResponse = Buffer.concat([this.context.rawResponse, chunk]);
  }

  /**
   * The overridden onData method to capture response chunks.
   */
  public onData(chunk: Buffer): void {
    try {
      this.handleChunk(chunk);
    } catch (error) {
      console.error(`Error in undici telemetry onData:`, error);
    }
  }

  /**
   * The overridden onFinally method to finalize capturing.
   */
  public onFinally(_error: Error | null): void {
    try {
      if (this.context.rawResponse.length === 0) {
        return;
      }
      const stringifiedBody = this.stringifiedResponseBody();
      this.context.span?.setAttribute(this.responseBodySpanAttrName, stringifiedBody);
    } catch (error) {
      console.error(`Error in undici telemetry onFinally:`, error);
    }
  }

  private stringifiedResponseBody(): string {
    // If not content-encoded, return utf8 of rawResponse buffer
    // Read response headers from the span, very hacky but it works. Unfortunatly the response object
    // is not available here because this class hooks into the requestHook, not the responseHook and it
    // has to do that so we can listen to onData and onFinally BEFORE the response is returned.
    if (!this.context.span) {
      throw new Error('Span is not set in the context.');
    }

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const responseHeadersJson = (this.context.span as any).attributes[
      this.responseHeaderSpanAttrName
    ];
    if (!responseHeadersJson) {
      throw new Error('Response headers not found in span attributes.');
    }

    const responseHeaders: Record<string, string> = JSON.parse(responseHeadersJson);
    const contentEncoding = responseHeaders['content-encoding'];
    if (!contentEncoding) {
      return this.context.rawResponse.toString('utf8');
    }

    // Decompress the rawResponse buffer
    return this.decompressBody(this.context.rawResponse, contentEncoding);
  }

  private decompressBody(rawResponse: Buffer, contentEncoding: string): string {
    switch (contentEncoding) {
      case 'gzip':
      case 'x-gzip':
        return zlib.gunzipSync(rawResponse).toString('utf8');
      case 'br':
        return zlib.brotliDecompressSync(rawResponse).toString('utf8');
      case 'deflate':
        return zlib.inflateSync(rawResponse).toString('utf8');
      default:
        throw new Error(`Unsupported content encoding: ${contentEncoding}`);
    }
  }
}

/**
 * Converts an array of headers to a map/object.
 * @returns {Record<string, string>} - A map of header keys to their values.
 */
function headerArrayToMap(headers: Buffer[] | string[]): Record<string, string> {
  const headersMap: Record<string, string> = {};
  for (let i = 0; i < headers.length; i += 2) {
    const key = headers[i].toString().toLowerCase();
    const value = headers[i + 1].toString();
    headersMap[key] = value;
  }
  return headersMap;
}

/**
 * Factory function to get UndiciInstrumentation with custom hooks.
 * @returns {object} - The instrumentation configuration with hooks.
 */
export function getUndiciInstrumentationConfig(options: UndiciInstrumentationOptions = {}) {
  const {
    requestHeadersSpanAttrName = 'http.request.headers',
    responseHeaderSpanAttrName = 'http.response.headers',
    requestBodySpanAttrName = 'http.request.body',
    responseBodySpanAttrName = 'http.response.body',
  } = options;

  return {
    /**
     * Hook to handle requests.
     */
    requestHook: async (span: Span, request: UndiciRequest) => {
      try {
        const handler = new UndiciInstrumentationResponseBodyHandler(
          span,
          responseHeaderSpanAttrName,
          responseBodySpanAttrName,
        );
        const requestHeaders = headerArrayToMap(request.headers as string[]);
        span.setAttribute(requestHeadersSpanAttrName, stringify(requestHeaders));

        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        const originalOnData = (request as any).onData.bind(request);
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        (request as any).onData = (chunk: Buffer) => {
          handler.onData(chunk);
          return originalOnData(chunk);
        };

        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        const originalOnFinally = (request as any).onFinally.bind(request);
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        (request as any).onFinally = (error: Error | null) => {
          handler.onFinally(error);
          return originalOnFinally(error);
        };
      } catch (error) {
        console.error(`Error in undici telemetry requestHook:`, error);
      }
    },

    /**
     * Adds custom attributes to the span on response.
     */
    responseHook: async (
      span: Span,
      { request, response }: { request: UndiciRequest; response: UndiciResponse },
    ) => {
      try {
        const responseHeaders = headerArrayToMap(response.headers as Buffer[]);
        span.setAttribute(responseHeaderSpanAttrName, stringify(responseHeaders));

        // Since the response hook is not async we need to read the body AFTER we are done sending the request and have
        // received the response so we can be sure we aren't going to be consuming the body mid request. Doing so
        // leads to mismatched content-length errors.
        const requestBody = await readAndRestoreRequestBody(request);
        span.setAttribute(requestBodySpanAttrName, requestBody);
      } catch (error) {
        console.error(
          `Error in undici telemetry responseHook: Failed to set headers as span attributes:`,
          error,
        );
      }
    },
  };
}
```

Issue #2360, opened 2024-07-31
[opentelemetry-instrumentation-aws-sdk] Add S3 Bucket Name and Kinesis Stream Name to aws-sdk span attribute if present


jj22ee:
### Is your feature request related to a problem? Please describe

I created S3 (`ListObjects`) and Kinesis (`CreateStream`) Client calls, but did not observe S3 Bucket Name or Kinesis Stream Name in the respective spans that were generated for these calls.

### Describe the solution you'd like to see

I want S3 Bucket Name and Kinesis Stream Name to be added to aws-sdk span attribute if present

### Describe alternatives you've considered

N/A

### Additional context

S3 semantic conventions: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/object-stores/s3.md
No such semantic conventions for Kinesis Streams exist in OTel JS repo, but AWS ADOT team [hopes that such conventions are eventually added to upstream](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/2c00bd07eaaf703880a24a2fcfe874cdb4196678/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_attribute_keys.py#L15-L19).

PR #1464, opened 2023-04-10
chore: update component owner for AWS to carolabadeer

willarmiros:
## Short description of the changes

I am transferring ownership of AWS X-Ray Components to @carolabadeer, who will help maintain them going forward per `CONTRIBUTING.md`.

PR #2361, opened 2024-07-31
feat(aws-sdk): add s3 and kinesis service extensions for aws-sdk instrumentation

jj22ee:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2360

- Add service extensions for S3 and Kinesis in AWS SDK instrumentation.
  - For S3, add bucket name to span if present
  - For Kinesis, add stream name to span if present
- This is a simple starting point for these new service extensions. The bucket/stream names are most important, but in the future, perhaps more data can be collected as span attributes.

Regarding some added constants to the `utils` file:
```
export const _AWS_S3_BUCKET = 'aws.s3.bucket';
export const _AWS_KINESIS_STREAM_NAME = 'aws.kinesis.stream.name';
```
- `_AWS_S3_BUCKET` should be added to [OTel JS Semantic Attributes in the future](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-semantic-conventions/src/trace/SemanticAttributes.ts).
  - See S3 specification: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/object-stores/s3.md
- There is no related specification for `_AWS_KINESIS_STREAM_NAME`, but [some AWS repos are already using `aws.kinesis.stream.name`](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/2c00bd07eaaf703880a24a2fcfe874cdb4196678/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/)

## Testing

1. Unit tests
2. Tested with:
  - Kinesis Client: observed `'aws.kinesis.stream.name': myNewKinesisStream`
```
    const client = new KinesisClient({ region: 'us-east-1' });
    const input = { // CreateStreamInput
      StreamName: "myNewKinesisStream", // required
      ShardCount: Number(3),
      StreamModeDetails: { // StreamModeDetails
        StreamMode: "ON_DEMAND", // required
      },
    };
    const command = new CreateStreamCommand(input);
    await client.send(command).then(data => {
      console.log(data);
    })
```
  - S3 Client: observed `'aws.s3.bucket': test-bucket-not-exist-or-accessible`
```
  const s3Client = new S3Client({ region: 'us-east-1' });
  const bucketName = 'test-bucket-not-exist-or-accessible';
  try {
    await s3Client.send(
      new ListObjectsCommand({
        Bucket: bucketName,
      }),
    ).then((data) => {
      console.log(data);
    });
  } catch (e) {
    if (e instanceof Error) {
      console.error('Exception thrown: ', e.message);
    }
  } finally {
    res.send('done aws sdk s3 request');
  }
```


srprash (approved):

jj22ee:
Pinging @blumamir for review!

Unrelated heads-up, I'm looking to take over component ownership of AWS JS components from @carolabadeer. I think I just need to make a PR like https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1464.

trentm:
> I think I just need to make a PR like #1464.

@jj22ee Yup that's correct. Great to have you helping.

plugins/node/opentelemetry-instrumentation-aws-sdk/package.json
```diff
@@ -52,6 +52,7 @@
   },
   "devDependencies": {
     "@aws-sdk/client-dynamodb": "3.85.0",
+    "@aws-sdk/client-kinesis": "3.85.0",
```
blumamir, line 55: is it used anywhere? 
seems the tests for kinesis uses v2 (`aws-sdk`) and this dependency is redundant.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/s3.test.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+import {
+  getTestSpans,
+  registerInstrumentationTesting,
+} from '@opentelemetry/contrib-test-utils';
+import { AwsInstrumentation } from '../src';
+registerInstrumentationTesting(new AwsInstrumentation());
+
+import * as AWS from 'aws-sdk';
```
blumamir, line 24: awesome to have tests for that.

Please note that those tests are for v2 which [will soon be end of life](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/).
Consider also adding assertions on these new attributes to v3 tests in `test/aws-sdk-v3.test.ts` [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/60a99c98fd3a5594c7c2234184f06166b375e707/plugins/node/opentelemetry-instrumentation-aws-sdk/test/aws-sdk-v3.test.ts#L66)

blumamir (commented):
Thank you for the contribution and the info in the PR.

Left few comments. Some of them I wrote on s3 implementation but they also apply to kinesis, didn't want to duplicate the same comments twice :)

jj22ee:
Hi, looking for another round of review! I removed some of the aws-sdk v2 tests I had previously given their end-of-life and soon-to-be-deprecated statuses, and removed the latency with the Kinesis Client test (enforced HTTP instead of HTTP2 which isn't supported by `nock`)

blumamir (approved):
LGTM 💯 

Thank you for the contribution, the code and tests looks great and the description on the PR is very helpful.

PR #2366, opened 2024-08-02
fix(amqplib)!: instrumentation produced high cardinality span names

cedricziel:

This change amends the amqplib instrumentation so it adheres to span name conventions mentioned here https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name

Ref: #2367 

Ensure span name is low cardinality and uses a standard operation name

fixes: #2367


maryliag (approved):
small nit, otherwise LGTM

plugins/node/instrumentation-amqplib/test/amqplib-callbacks.test.ts
```diff
@@ -136,10 +137,14 @@ describe('amqplib instrumentation callback model', () => {
         const [publishSpan, consumeSpan] = getTestSpans();
 
         // assert publish span
+        expect(publishSpan.name).toMatch('<default> publish'); // span name needs to be low cardinality
```
cedricziel, line 140:
```suggestion
        expect(publishSpan.name).toMatch('<default> publish');
```

david-luna (approved):

trentm (approved):
LGTM.

I'll wait for a bit to let Amir, as component owner, to have a chance to review.

FWIW, the `... -> ${routingKey} ` in the span.name dates back to the original commit: https://github.com/aspecto-io/opentelemetry-ext-js/commit/e3bd5ea377732df3085ed7177922caf574ec7616#diff-f1a7a2a54feb09ebd4bde5c788752e112f0d521fa6de4b05d725a887b2cf743dR305

JamieDanielson (commented):
I've not reviewed this in full but it seems like it should be marked as a breaking change, even if it is technically a fix.

blumamir (commented):
Thank you @cedricziel 
overall LGTM, left 2 nit suggestions / concerns

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -655,21 +655,19 @@ export class AmqplibInstrumentation extends InstrumentationBase<AmqplibInstrumen
...
-          [SEMATTRS_MESSAGING_MESSAGE_ID]: options?.messageId,
-          [SEMATTRS_MESSAGING_CONVERSATION_ID]: options?.correlationId,
-        },
-      }
-    );
+    const span = self.tracer.startSpan(`publish ${normalizedExchange}`, {
+      kind: SpanKind.PRODUCER,
+      attributes: {
+        ...channel.connection[CONNECTION_ATTRIBUTES],
+        [SEMATTRS_MESSAGING_OPERATION]: 'publish',
```
cedricziel, line 662:
```suggestion
```

plugins/node/instrumentation-amqplib/test/amqplib-callbacks.test.ts
```diff
@@ -30,6 +30,7 @@ import {
   MESSAGINGDESTINATIONKINDVALUES_TOPIC,
   SEMATTRS_MESSAGING_DESTINATION,
   SEMATTRS_MESSAGING_DESTINATION_KIND,
+  SEMATTRS_MESSAGING_OPERATION,
```
cedricziel, line 33:
```suggestion
```

plugins/node/instrumentation-amqplib/src/amqplib.ts
```diff
@@ -655,21 +655,19 @@ export class AmqplibInstrumentation extends InstrumentationBase<AmqplibInstrumen
...
-          [SEMATTRS_MESSAGING_MESSAGE_ID]: options?.messageId,
-          [SEMATTRS_MESSAGING_CONVERSATION_ID]: options?.correlationId,
-        },
-      }
-    );
+    const span = self.tracer.startSpan(`${normalizedExchange} publish`, {
+      kind: SpanKind.PRODUCER,
+      attributes: {
+        ...channel.connection[CONNECTION_ATTRIBUTES],
+        [SEMATTRS_MESSAGING_OPERATION]: 'publish',
```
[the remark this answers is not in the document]
cedricziel: removed, thanks!

blumamir (approved):

blumamir:
> I've not reviewed this in full but it seems like it should be marked as a breaking change, even if it is technically a fix.

Do you think it's breaking because someone might rely on the current name for some logic?
I would personally not consider it breaking, but since the current major is `0` I don't see any harm in marking it as breaking just in case

PR #2371, opened 2024-08-06
fix(detector-azure): suppress tracing for AzureVmDetector

david-luna:

`AzureVmDetector` is making a HTTP request to get metadata from the VM and depending on how the instrumentation is configured it may export HTTP spans related to that requet. Internal tracing should not be sent to the collector because it may confuse customers and the data is not related to the service being instrumented.

Refs: #2320 

- execute the HTTP request with `suppressTracing`
- add tests to check no spans are being exported.


pichlermarc (changes requested):
I think we need to add a dependency on `@opentelemetry/core` :slightly_smiling_face:

pichlermarc (approved):

blumamir (approved):

Issue #2226, opened 2024-05-20
[undici] add all headers


mbrevda:
### Is your feature request related to a problem? Please describe

Headers must be whitelisted to be added to the trace

### Additional context

An option should be added to allow adding _all_ headers, with perhaps a blacklist option of headers not to add (such as sensitive headers)

Thanks!

david-luna:
Hi @mbrevda

I guess the instrumentation can assume that if there is a `block list` for headers it should send them all but the ones in the config. The behaviour could be described this way:

- no headers config => send nothing (for backwards compatibility)
- only `headersToSpanAttributes` present => send only the headers defined there
- only `blockHeadersToSpanAttributes` present => send all headers but the ones defined in the config
- both properties defined => sends the ones defined in `headersToSpanAttributes` if not present in `blockheadersToSpanAttributes`


so the following config would make undici to send all headers but `autorization`
```json
{
  "blockheadersToSpanAttributes": {
    "requestHeaders": ["authorization"]
  }
}
```

Maybe allowing to have a function would be more straight forward but IMO this way eventually we can have this configuration via env vars.

Thoughts?

trentm:
OTel semconv says this about collecting HTTP request and response headers:
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-client

```
**[11]:** Instrumentations SHOULD require an explicit configuration of which headers are to be captured. Including all request headers can be a security risk - explicit configuration helps avoid leaking sensitive information.
The `User-Agent` header is already captured in the `user_agent.original` attribute. Users MAY explicitly configure instrumentations to capture them even though it is not recommended.
The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.
```

I'm sorry I didn't note this before you made a PR for this, David.

This doesn't disallow adding a relatively simple config option to collect all headers (like `blockHeadersToSpanAttributes` or similar), but it discourages it.
Thoughts?
I am inclined to think twice before adding this functionality. And if we do, then we add a big security warning.

@mbrevda I believe it is possible to get what you want by using the existing `requestHook` config option. E.g.: this config option passed to `getNodeAutoInstrumentations()`

```js
            '@opentelemetry/instrumentation-undici': {
                requestHook: (span, req) => {
                    for (let i = 0; i < req.headers.length; i += 2) {
[...]

PR #2379, opened 2024-08-13
feat(instr-undici): add `blockHeadersToSpanAttributes` config option

david-luna:

- add a new option `blockHeadersToSpanAttributes` to configure a block list, all headers that are not in the list will be converted to span attributes.
- if set in combination with `headersToSpanAttributes` the new config will take precedence and if a header is present in both config options the header will not be converted.
- if no config is set the instrumentation won't convert any header so the change is backwards compatible.

Closes: #2226

- add new config option
- update request/response headers conversion
- add tests


trentm:
See my comment: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2226#issuecomment-2297734212

PR #2380, opened 2024-08-13
feat(instrumentation-pg): implementation of metric operation duration

maryliag:

- Part Of #2348 

- This PR implements part of the [Semantic Conventions for Database Metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md). I will break down the implementation on different PRs to facilitate the review process.
- This PR focus on the metrics:
   - [db.client.operation.duration](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientoperationduration)

This PR does not contain all attributes, because all the ones that require parsing of the query will be more complex and need dedicated testing for performance and be opt-in, so I will leave those for the next PR

Note for reviewers: I still need to add tests to this PR


plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -236,15 +237,16 @@ export function handleExecutionResult(
...
+  attributes: Attributes
 ): PostgresCallback {
   return function patchedCallback(
     this: PgClientExtended,
     err: Error,
     res: object
   ) {
     if (err) {
-      // span.recordException(err);
+      attributes['error.type'] = err.message;
```
david-luna, line 249: Why the `error.type` here is resolved in a different way that the previous cases?

david-luna (commented):
@maryliag thanks for your contribution :)

PR looks very good. I left one question I'm curious about before giving approval.

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
maryliag: oh nice catch! actually the others should be updated to message. Fixed!

JacksonWeber (approved):
LGTM

david-luna (approved):

maryliag:
@pichlermarc friendly ping if this can be merged

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -960,4 +962,102 @@ describe('pg', () => {
...
+        );
+        const dataPoint = metrics[0].dataPoints[0];
+        assert.strictEqual(
+          dataPoint.attributes[SEMATTRS_DB_SYSTEM],
+          DBSYSTEMVALUES_POSTGRESQL
+        );
+        assert.strictEqual(
+          dataPoint.attributes[ATTR_ERROR_TYPE],
+          'function test() does not exist'
+        );
```
pichlermarc, line 1044: Hmm, does the instrumentation parse the actual error message to add it to the attribute? :thinking: 
This sounds like it could potentially be a source of a high-cardinality attribute - do the semantic conventions limit this in any way?

maryliag: Ah good point, I might actually use the error code instead, since there is a limited number of possibilities. We didn't discuss a limitation for it, but you're right, this could cause a high cardinality issue

I may add a option to opt-in to get the message, we will need an opt-in case for the query itself, so I can work on a similar scenario (this will all be on a different PR anyway)

plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -236,15 +237,17 @@ export function handleExecutionResult(
...
+  recordDuration: { (): void }
 ): PostgresCallback {
   return function patchedCallback(
     this: PgClientExtended,
     err: Error,
     res: object
   ) {
     if (err) {
-      // span.recordException(err);
+      attributes['error.type'] = getErrorMessage(err);
```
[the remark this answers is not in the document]
maryliag: Nice catch! :) 

This ended up getting removed per my other commend

pichlermarc (approved):

pichlermarc:
Failing tests are actually caused by #2445 - I'm not sure how to fix that yet. :disappointed:

pichlermarc:
The [test failures now](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/11013435305/job/30581787109?pr=2380) seem related to the previous PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2349 which seems to have introduced some flakiness in the tests.

I've re-started, but it's something that we need to address.

PR #2381, opened 2024-08-14
feat(instrumentation-tedious): support tedious@18

trentm:
This builds on work by @kirrg001 in #2267.
tedious@18 added its own types, dropping the need to use
@types/tedious. However, it requires TypeScript v5 to use with TS,
which was challenging.

- This uses test-all-versions' support for 'peerDependencies' to only
  install TS 5 when testing tedious@18.
- This drops ts-mocha, in favour of 'mocha --require ts-node/register'
  as is being used in the core repo.
- This adds skipLibCheck:true to the tsconfig, needed to compile
  the *tests* when using tedious@18. It was that or bump to
  target:es2022, which could impact the published code. JS SIG
  discussion decided using skipLibCheck is probably fine.

More discussion here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2267#issuecomment-2278923319

Obsoletes: #2267
Closes: #2266


trentm:
For the record: some earlier skipLibCheck discussion here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2165

kirrg001 (approved):
LGTM

david-luna (approved):

PR #2388, opened 2024-08-19
feat(opentelemetry-test-utils): create TestMetricReader

maryliag:
Creation of a TestMetricReader that can be used during tests involving metric collection.

Created based on comment: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2349#discussion_r1708887639

For usage, the function `initMeterProvider` can be called on the `beforeEach` of tests, and the return of `metricReader` can be used such as `const { resourceMetrics, errors } = await metricReader.collect();`

Once a new version of the package `opentelemetry-test-utils` is created, I will update the PR mentioned above with its usage.

### DIFF
--- a/packages/opentelemetry-test-utils/src/test-utils.ts
+++ b/packages/opentelemetry-test-utils/src/test-utils.ts
@@ -24,12 +24,14 @@ import {
 } from '@opentelemetry/api';
 import * as assert from 'assert';
 import { ReadableSpan } from '@opentelemetry/sdk-trace-base';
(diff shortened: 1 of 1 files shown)


maryliag:
why this added so many labels? 👀 Is it because all of them use this package?

trentm:
> Is it because all of them use this package?

Yup.

maryliag:
I also forgot to export the class itself, so I created a PR for it: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2393

But even with that change, I still can't use the function from here

Issue #2383, opened 2024-08-15
Zero code instrumentation doesn't flush telemetry on normal exit


aabmass:
### What version of OpenTelemetry are you using?
main branch

### What version of Node are you using?
Node 22

### What did you do?
Zero code instrumentation of a script which exits normally, using `--require @opentelemetry/auto-instrumentations-node/register`

```ts
async function main() {
  const res = await fetch('https://opentelemetry.io');
  console.log('Got status %s', res.status);
}

main();
```

To repro, checkout this branch https://github.com/open-telemetry/opentelemetry-js-contrib/compare/main...aabmass:opentelemetry-js-contrib:auto-flush-repro and

```sh
cd flush-repro/
docker compose up --build --abort-on-container-exit
```

(you may need to `chmod a+r config.yaml` if the collecter container can't read the mounted file)

### What did you expect to see?
When the program finishes running normally, the SDK should shutdown and flush telemetry. In the repro, the collector should print the traces it receives.

### What did you see instead?
The process exits without ever calling `NodeSDK.shutdown()`. No telemetry is sent to the collector and nothing is printed.

### Additional context
There is already a SIGTERM handler which flushes when the process is gracefully killed
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b169191a718248aa5f2445e5945987f4628daaba/metapackages/auto-instrumentations-node/src/register.ts#L43-L45

but this doesn't work when the process exits normally. I included a fix in a comment, which is adding a `process.on('beforeExit')` hook that calls `sdk.shutdown()`. I'm happy to send a PR

trentm:
Yes, I think it should. I use a `beforeExit` handler typically myself.

(One thing to potentially watch out for is that `beforeExit` **in an AWS Lambda function** should not be used to shutdown the SDK, because it is sometimes used to watch for when the freeze an instance. Anyway, this doesn't matter here because the OTel Lambda layer for Node.js does **not** use auto-instrumentations-node: https://github.com/open-telemetry/opentelemetry-lambda/tree/main/nodejs/packages/layer/)

A PR would be great.

aabmass:
Thanks @trentm, can you assign this to me?

PR #2394, opened 2024-08-21
fix(auto-instrumentations-node): shutdown the SDK when the process exits normally

aabmass:

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2383

Adds a one-shot `beforeExit` hook to call `NodeSDK.shutdown()`. In the event of SIGTERM  and the process gracefully exiting, shutdown may be called twice. I believe the additional calls to `shutdown()` should no-op so that is not a concern but may make for spammy logs.

I updated the tests to capture the expected behavior
- Shutdown on SIGTERM
- Shutdown when exiting normally


trentm (approved):
Thanks!

trentm:
I'll leave some time for others to review as well.

pichlermarc (approved):

Issue #2391, opened 2024-08-21
`instrumentation-undici` does not work when `enabled: true` is set in the configuration


enikey87:
### What version of OpenTelemetry are you using?

```json
  "dependencies": {
    "@opentelemetry/api": "1.9.0",
    "@opentelemetry/auto-instrumentations-node": "0.49.1",
    "@opentelemetry/sdk-metrics": "1.25.1",
    "@opentelemetry/sdk-node": "0.52.1"
  },
```
### What version of Node are you using?
```sh
node -v
v20.16.0
```

### What did you do?

Tried to instrument fetch with Node.js auto-instrumentations with explicitly enabled `instrumentation-undici`:
```javascript
import { NodeSDK } from '@opentelemetry/sdk-node'
import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node'
import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-base'
import { register } from 'module'

const sdk = new NodeSDK({
  traceExporter: new ConsoleSpanExporter(),

  instrumentations: [
    getNodeAutoInstrumentations({
      '@opentelemetry/instrumentation-undici': { enabled: true }
    })
  ]
})

sdk.start()

register('@opentelemetry/instrumentation/hook.mjs', import.meta.url)
```

### What did you expect to see?

Expected to see spans exported to console for node's global [fetch()](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch).

### What did you see instead?

No spans in console.

### Additional context
- [I put repo that illustrates the issue and workaround here](https://github.com/enikey87/otel-node-instrumentation-undici-reproduction?tab=readme-ov-file).
- No UndiciInstrumentation is applied because it [uses config.enabled](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/8b35f786158767967597cfc70e053dac4abd72ab/plugins/node/instrumentation-undici/src/undici.ts#L91) for both its internal state and the configuration itself.

PR #2395, opened 2024-08-21
fix(instr-undici): fix issue with config in constructor

david-luna:

Fixes: #2391 

- removed `enabled` internal config from the logic of the instrumentation. Now it relies only in enable/disable methods
- updated tests

### DIFF
--- a/plugins/node/instrumentation-undici/src/undici.ts
+++ b/plugins/node/instrumentation-undici/src/undici.ts
@@ -84,6 +84,12 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     // This method is called by the `InstrumentationAbstract` constructor before
     // ours is called. So we need to ensure the property is initalized
     this._channelSubs = this._channelSubs || [];
+
+    // Avoid to duplicate subscriptions
+    if (this._channelSubs.length > 0) {
+      return;
+    }
+
     this.subscribeToChannel(
       'undici:request:create',
       this.onRequestCreated.bind(this)
--- a/plugins/node/instrumentation-undici/src/undici.ts
+++ b/plugins/node/instrumentation-undici/src/undici.ts
@@ -76,13 +76,25 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
   }
 
   override disable(): void {
+    super.disable()
     this._channelSubs.forEach(sub => sub.channel.unsubscribe(sub.onMessage));
     this._channelSubs.length = 0;
   }
 
   override enable(): void {
-    // This method is called by the `InstrumentationAbstract` constructor before
-    // ours is called. So we need to ensure the property is initalized
+    // "enabled" handling is currently a bit messy with InstrumentationBase.
+    // If constructed with `{enabled: false}`, this `.enable()` is still called,
+    // and `this.getConfig().enabled !== this.isEnabled()`, creating confusion.
+    //
+    // For now, this class will setup for instrumenting if `.enable()` is
+    // called, but use `this.getConfig().enabled` to determine if
+    // instrumentation should be generated. This covers the more likely common
+    // case of config being given a construction time, rather than later via
+    // `instance.enable()`, `.disable()`, or `.setConfig()` calls.
+    super.enable()
+
+    // This method is called by the super-class constructor before ours is
+    // called. So we need to ensure the property is initalized.
     this._channelSubs = this._channelSubs || [];
 
     // Avoid to duplicate subscriptions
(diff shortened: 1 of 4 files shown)


plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -113,16 +106,6 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     this.subscribeToChannel('undici:request:error', this.onError.bind(this));
   }
 
-  override setConfig(config: UndiciInstrumentationConfig = {}): void {
```
david-luna, line 116: Note for reviewer: here I was mixing `UndiciInstrumentationConfig` with `InstrumentationConfig`. This side effect shouldn't be there

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -77,26 +76,20 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
   }
 
   override disable(): void {
-    if (!this.getConfig().enabled) {
-      return;
-    }
-
     this._channelSubs.forEach(sub => sub.channel.unsubscribe(sub.onMessage));
     this._channelSubs.length = 0;
-    super.disable();
```
david-luna, line 86: Note for reviewer: no need to call super since the parent class method deals with patching/unpatching and does not apply here

plugins/node/instrumentation-undici/test/fetch.test.ts
```diff
@@ -57,6 +54,9 @@ describe('UndiciInstrumentation `fetch` tests', function () {
       this.skip();
     }
 
+    instrumentation = new UndiciInstrumentation();
```
david-luna, line 57: Note for reviewer: the test runner was loading all the test suites at once and also evaluating the 1st `describe` block. As a consequence there were 3 instances of the instrumentation before any test was ran. Leading to errors because all instances where getting messages from diagnostics channels

plugins/node/instrumentation-undici/src/undici.ts
trentm: super.disable() is *also* setting `this._enabled = false`, so I wonder if it still should be called.
Those calls to `super.` in the enable() and disable() were added in this PR: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2289/files#diff-8e6dd9963dd47a60e25c9ac92d7a5ae25ce79c40340cb4721eefd203bec81a60R86

trentm:
Slightly related, I've opened https://github.com/open-telemetry/opentelemetry-js/pull/4941 to eventually deal with a confusion that can come from using `instr.setConfig({})`.

plugins/node/instrumentation-undici/test/fetch.test.ts
trentm: Nice find!

trentm:
Here is my test script for manual testing of the usage via NodeSDK, with getNodeAutoInstrumentations (as shown in the original reported issue), and without `getNodeAutoInstrumentations` in a few different cases:

`use-undici-request.js`
```
const path = require('path');
const {NodeSDK} = require('@opentelemetry/sdk-node');
const {getNodeAutoInstrumentations} = require('@opentelemetry/auto-instrumentations-node');
const {UndiciInstrumentation} = require('./');

// Instrument undici.
process.env.OTEL_TRACES_EXPORTER = 'console'
const sdk = new NodeSDK({
  serviceName: path.parse(process.argv[1]).name,
  instrumentations: [
    // 1.
    // new UndiciInstrumentation(),
    // 2.
    // new UndiciInstrumentation({enabled: false}),
    // 3.
    // new UndiciInstrumentation({enabled: true}),
    // 4.
    getNodeAutoInstrumentations({
      '@opentelemetry/instrumentation-undici': {
        enabled: true,
      },
    }),
  ],
});
process.once('beforeExit', () => {
  sdk.shutdown();
});
sdk.start();

// Use undici.
const undici = require('undici');
async function main() {
  const res = await undici.request('http://www.google.com/');
  console.log('response: %s', res.statusCode);
  const body = await res.body.text();
  console.log('response body: %s...', body.slice(0, 40));
}
main();
```

All four cases work as I'd expect. Case 3 results in no undici span; all other cases result in an undici span.

trentm:
> All four cases work as I'd expect.

... they work *after* I get the commit in that I had. Still trying to get that pushed.

plugins/node/instrumentation-undici/src/undici.ts
trentm: I've restored the `super.disable()` and `super.enable()` calls, to allow InstrumentationBase to maintain its internal `this._enabled` setting.

trentm (approved):

pichlermarc (approved):

Issue #2389, opened 2024-08-19
App crashes with a `TypeError` when using the Redis instrumentation


rowanmanning:
### What version of OpenTelemetry are you using?

  * @opentelemetry/sdk-node `v0.52.1`
  * @opentelemetry/auto-instrumentations-node `v0.49.1`

### What version of Node are you using?

  * Node.js `v20.16.0`

### What did you do?

@metart43 and I tried to use the `redis` package alongside Node.js auto-instrumentations. We're connecting to Redis with an empty string as a URL which the `redis` client accepts.

### What did you expect to see?

We expected to see a log (`Redis connected`) and the app to continue to run.

### What did you see instead?

The app crashes with a `TypeError`:

```
/redacted/node_modules/@opentelemetry/instrumentation-redis-4/build/src/utils.js:34
        diag.error('failed to sanitize redis connection url', err);
             ^

TypeError: Cannot read properties of undefined (reading 'error')
    at removeCredentialsFromDBConnectionStringAttribute (/redacted/node_modules/@opentelemetry/instrumentation-redis-4/build/src/utils.js:34:14)
    at getClientAttributes (/redacted/node_modules/@opentelemetry/instrumentation-redis-4/build/src/utils.js:11:65)
    at Commander.patchedConnect (/redacted/node_modules/@opentelemetry/instrumentation-redis-4/build/src/instrumentation.js:218:68)
    at Object.<anonymous> (/redacted/index.js:11:8)
    at Module._compile (node:internal/modules/cjs/loader:1358:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Module._load (node:internal/modules/cjs/loader:1024:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12)
    at node:internal/main/run_main_module:28:49
```

### Additional context

  * [**I put together a reduced test case in a Gist here**](https://gist.github.com/rowanmanning/0c84d824d12405f00761f47a3d33db43). This illustrates the issue consistently
  * It seems like the issue is in some cases `diag` is `undefined` by the time it reaches [this line](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b169191a718248aa5f2445e5945987f4628daaba/plugins/node/opentelemetry-instrumentation-redis-4/src/utils.ts#L60)

trentm:
[...]

PR #2399, opened 2024-08-22
fix(instrumentation-redis-4): avoid diag.error spam when configured client URL is the empty string

trentm:
Issue #2389 showed that this instrumentation would *crash* on a Redis
client configured with `{url: ''}` (an empty string). The crash was
fixed in #2397. This change avoids the once-crashy code, and hence
the diag.error spam, by using the same guard before attempting to parse
the configured client URL that the Redis client itself does, `if
(options?.url)`,

Refs: #2389


JacksonWeber (approved):
LGTM

pichlermarc (approved):

PR #2403, opened 2024-08-26
feat: Record AWS Lambda coldstarts

serkan-ozal:

Adds recording AWS Lambda invocation coldstarts as Lambda invocation span attributes.

Decides whether an AWS Lambda invocation is coldstart or not by taking care of the following cases:
- Non-first invocations are warm, not coldstart
- If sandbox environment is initialized with provisioned concurrency, every invocations are warm, not coldstart
- If sandbox has been initialized [proactively](https://aaronstuyvenberg.com/posts/understanding-proactive-initialization) before the actual request, every invocations are warm, not coldstart


tylerbenson (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -164,16 +174,47 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+        if (
+          process.env.AWS_LAMBDA_INITIALIZATION_TYPE ===
+          'provisioned-concurrency'
+        ) {
+          // If sandbox environment is initialized with provisioned concurrency,
+          // even the first requests should not be considered as coldstart.
+          requestIsColdStart = false;
+        } else {
+          // Check whether it is proactive initialization or not:
+          // https://aaronstuyvenberg.com/posts/understanding-proactive-initialization
```
pragmaticivan, line 204: 🤯

pragmaticivan (approved):

tylerbenson (approved):

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -44,6 +44,7 @@ import {
   AWSXRayPropagator,
 } from '@opentelemetry/propagator-aws-xray';
 import {
+  SEMATTRS_FAAS_COLDSTART,
```
dyladan, line 47: Nit: Would be nice to update the semconv package and use the new constant. Not blocking

dyladan (approved):
Looks reasonable to me. I think this can be merged. I left a minor suggestion, but if you don't want to do it that's fine. Just comment and we'll merge this when you're ready.

serkan-ozal:
> Looks reasonable to me. I think this can be merged. I left a minor suggestion, but if you don't want to do it that's fine. Just comment and we'll merge this when you're ready.

@dyladan Are you suggesting to update semconv package and update all semconv attributes in all places? If so, I think it should be done in a separate PR. I have used `SEMATTRS_...` properties, because as far as I see, it is the general approach  by other instrumentations too.

dyladan:
> > Looks reasonable to me. I think this can be merged. I left a minor suggestion, but if you don't want to do it that's fine. Just comment and we'll merge this when you're ready.
> 
> @dyladan Are you suggesting to update semconv package and update all semconv attributes in all places accordingly? If so, I think it should be done in a separate PR. I have used `SEMATTRS_...` properties, because as far as I see, it is the general approach used by other instrumentation plugins too.

I had been thinking only the newly introduced attribute would use the new format. The old formats are still exported for backward compatibility. but leaving it for a followup is also fine.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
serkan-ozal: @dyladan ok, done. just used that new attribute (`faas.coldstart`) from new constant (experimental attribute through incubating)

PR #2407, opened 2024-08-28
feat!: use modern diagnostics_channel API

secondfry:

I'm observing that `instrumentation-undici` does not work in Node.js 18.16.0 if code was started with no Internet access. I nailed it down to usage of `channel.subscribe(onMessage)` which was deprecated in Node.js 18.7.0.

Interestingly enough new API does not pass tests on all Node.js versions between 18.7.0 and 18.18.2. I think this is related to https://github.com/nodejs/node/pull/47520. Thus I'm setting minimal engine version to 18.19.0.

Use [diagnostics_channel.subscribe(name, onMessage)](https://nodejs.org/api/diagnostics_channel.html#diagnostics_channelsubscribename-onmessage) instead of [channel.subscribe(onMessage)](https://nodejs.org/api/diagnostics_channel.html#channelsubscribeonmessage).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38476,7 +38476,7 @@
         "@opentelemetry/sdk-trace-base": "^1.8.0",
         "@opentelemetry/sdk-trace-node": "^1.8.0",
         "@types/mocha": "7.0.2",
-        "@types/node": "18.6.5",
+        "@types/node": "18.19.47",
         "mocha": "7.2.0",
         "nyc": "15.1.0",
         "rimraf": "5.0.10",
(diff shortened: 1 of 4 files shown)


david-luna:
@secondfry did you have time to think about my suggestion on nodejs support?

david-luna:
Hi @secondfry

If you have no bandwidth I can take this one although I'd go for not dropping support of Node.js versions

david-luna:
@secondfry this PR includes your proposal along a few changes to keep compatibility
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2457

secondfry:
@david-luna I'm terribly sorry for not following through with requested changes and thank you so much for your PR

PR #2408, opened 2024-08-28
chore(deps): update deps matching "@opentelemetry/"

pichlermarc:
0.52.1 -> 0.53.0 @opentelemetry/api-logs (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/instrumentation (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/instrumentation-fetch (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/instrumentation-grpc (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/instrumentation-http (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/instrumentation-xml-http-request (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/propagator-aws-xray-lambda (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/sdk-logs (range-bump)
    0.52.1 -> 0.53.0 @opentelemetry/sdk-node (range-bump)
    1.25.1 -> 1.26.0 @opentelemetry/context-async-hooks
    1.25.1 -> 1.26.0 @opentelemetry/context-zone
    1.25.1 -> 1.26.0 @opentelemetry/context-zone-peer-dep
    1.25.1 -> 1.26.0 @opentelemetry/core
    1.25.1 -> 1.26.0 @opentelemetry/exporter-jaeger
    1.25.1 -> 1.26.0 @opentelemetry/propagator-aws-xray
    1.25.1 -> 1.26.0 @opentelemetry/propagator-b3
    1.25.1 -> 1.26.0 @opentelemetry/propagator-jaeger
    1.25.1 -> 1.26.0 @opentelemetry/resources
    1.25.1 -> 1.26.0 @opentelemetry/sdk-metrics
    1.25.1 -> 1.26.0 @opentelemetry/sdk-trace-base
    1.25.1 -> 1.26.0 @opentelemetry/sdk-trace-node
    1.25.1 -> 1.26.0 @opentelemetry/sdk-trace-web
    1.25.1 -> 1.27.0 @opentelemetry/semantic-conventions

BEGIN_COMMIT_OVERRIDE
fix(instrumentation-bunyan): avoid overrding parent _config in setConfig()
feat(contrib-test-utils): export OtlpSpanKind as it does not match the Trace SDK SpanKind enum
test(instrumentation-express): fix up tests to match fixed http instrumentation, use correct span kind
 test(instrumentation-koa): assert for correct SpanKind, account for http client span
END_COMMIT_OVERRIDE

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-bunyan/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-bunyan/src/instrumentation.ts
@@ -96,7 +96,7 @@ export class BunyanInstrumentation extends InstrumentationBase<BunyanInstrumenta
   }
 
   override setConfig(config: BunyanInstrumentationConfig = {}) {
-    this._config = { ...DEFAULT_CONFIG, ...config };
+    super.setConfig({ ...DEFAULT_CONFIG, ...config });
   }
 
   private _getPatchedEmit() {
--- a/plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
@@ -620,20 +620,23 @@ describe('ExpressInstrumentation', () => {
         //     `- span 'middleware - simpleMiddleware'
         //     `- span 'request handler - /post/:id'
         const spans = collector.sortedSpans;
-        assert.strictEqual(spans[0].name, 'GET /post/:id');
-        assert.strictEqual(spans[0].kind, SpanKind.CLIENT);
-        assert.strictEqual(spans[1].name, 'middleware - query');
-        assert.strictEqual(spans[1].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[0].name, 'GET');
+        assert.strictEqual(spans[0].kind, testUtils.OtlpSpanKind.CLIENT);
+        assert.strictEqual(spans[1].name, 'GET /post/:id');
+        assert.strictEqual(spans[1].kind, testUtils.OtlpSpanKind.SERVER);
         assert.strictEqual(spans[1].parentSpanId, spans[0].spanId);
-        assert.strictEqual(spans[2].name, 'middleware - expressInit');
-        assert.strictEqual(spans[2].kind, SpanKind.SERVER);
-        assert.strictEqual(spans[2].parentSpanId, spans[0].spanId);
-        assert.strictEqual(spans[3].name, 'middleware - simpleMiddleware');
-        assert.strictEqual(spans[3].kind, SpanKind.SERVER);
-        assert.strictEqual(spans[3].parentSpanId, spans[0].spanId);
-        assert.strictEqual(spans[4].name, 'request handler - /post/:id');
-        assert.strictEqual(spans[4].kind, SpanKind.SERVER);
-        assert.strictEqual(spans[4].parentSpanId, spans[0].spanId);
+        assert.strictEqual(spans[2].name, 'middleware - query');
+        assert.strictEqual(spans[2].kind, testUtils.OtlpSpanKind.INTERNAL);
+        assert.strictEqual(spans[2].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[3].name, 'middleware - expressInit');
+        assert.strictEqual(spans[3].kind, testUtils.OtlpSpanKind.INTERNAL);
+        assert.strictEqual(spans[3].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[4].name, 'middleware - simpleMiddleware');
+        assert.strictEqual(spans[4].kind, testUtils.OtlpSpanKind.INTERNAL);
+        assert.strictEqual(spans[4].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[5].name, 'request handler - /post/:id');
+        assert.strictEqual(spans[5].kind, testUtils.OtlpSpanKind.INTERNAL);
+        assert.strictEqual(spans[5].parentSpanId, spans[1].spanId);
       },
     });
   });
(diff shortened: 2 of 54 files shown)


plugins/node/opentelemetry-instrumentation-bunyan/src/instrumentation.ts
```diff
@@ -96,7 +96,7 @@ export class BunyanInstrumentation extends InstrumentationBase<BunyanInstrumenta
   }
 
   override setConfig(config: BunyanInstrumentationConfig = {}) {
-    this._config = { ...DEFAULT_CONFIG, ...config };
```
pichlermarc, line 99: This did override the `enabled` config and actually disabled the instrumentation, this went unnoticed until the change from https://github.com/open-telemetry/opentelemetry-js/commit/0ee398d1e3f84059dc648597aa1b2f7be368f262 which now uses this method instead of setting it directly.

maryliag (commented):
you mention on the description the update on `@opentelemetry/semantic-conventions`, but I don't see the updates on this dependency

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -620,20 +620,23 @@ describe('ExpressInstrumentation', () => {
         //     `- span 'middleware - simpleMiddleware'
         //     `- span 'request handler - /post/:id'
         const spans = collector.sortedSpans;
-        assert.strictEqual(spans[0].name, 'GET /post/:id');
-        assert.strictEqual(spans[0].kind, SpanKind.CLIENT);
-        assert.strictEqual(spans[1].name, 'middleware - query');
-        assert.strictEqual(spans[1].kind, SpanKind.SERVER);
+        assert.strictEqual(spans[0].name, 'GET');
+        assert.strictEqual(spans[0].kind, testUtils.OtlpSpanKind.CLIENT);
```
pichlermarc, line 624: reviewer note: with https://github.com/open-telemetry/opentelemetry-js/pull/4866 http/https client instrumentation was fixed and now produces a client span.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -620,20 +620,23 @@ describe('ExpressInstrumentation', () => {
...
+        assert.strictEqual(spans[1].name, 'GET /post/:id');
+        assert.strictEqual(spans[1].kind, testUtils.OtlpSpanKind.SERVER);
```
pichlermarc, line 626: reviewer note: this was previously asserted to be a client span, which it actually was not. Since the SDK-internal representation for a client span is `2` (`SpanKind.CLIENT`) and the OTLP represenation for a server span is `2` (`OtlpSpanKind.SERVER`), the test passed anyway.

It now asserts on the correct kind.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -620,20 +620,23 @@ describe('ExpressInstrumentation', () => {
...
-        assert.strictEqual(spans[3].kind, SpanKind.SERVER);
-        assert.strictEqual(spans[3].parentSpanId, spans[0].spanId);
-        assert.strictEqual(spans[4].name, 'request handler - /post/:id');
-        assert.strictEqual(spans[4].kind, SpanKind.SERVER);
-        assert.strictEqual(spans[4].parentSpanId, spans[0].spanId);
+        assert.strictEqual(spans[2].name, 'middleware - query');
+        assert.strictEqual(spans[2].kind, testUtils.OtlpSpanKind.INTERNAL);
+        assert.strictEqual(spans[2].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[3].name, 'middleware - expressInit');
+        assert.strictEqual(spans[3].kind, testUtils.OtlpSpanKind.INTERNAL);
```
pichlermarc, line 632: reviewer note: the same thing I described above also happened with `SpanKind.SERVER` and `OtlpSpanKind.INTERNAL`. I also adapted all following tests.

plugins/node/opentelemetry-instrumentation-koa/test/koa.test.ts
```diff
@@ -741,14 +741,16 @@ describe('Koa Instrumentation', () => {
...
+        assert.strictEqual(spans[0].name, 'GET');
+        assert.strictEqual(spans[0].kind, testUtils.OtlpSpanKind.CLIENT);
+        assert.strictEqual(spans[1].name, 'GET /post/:id');
+        assert.strictEqual(spans[1].kind, testUtils.OtlpSpanKind.SERVER);
+        assert.strictEqual(spans[2].name, 'middleware - simpleMiddleware');
+        assert.strictEqual(spans[2].kind, testUtils.OtlpSpanKind.INTERNAL);
         assert.strictEqual(spans[2].parentSpanId, spans[1].spanId);
+        assert.strictEqual(spans[3].name, 'router - /post/:id');
+        assert.strictEqual(spans[3].kind, testUtils.OtlpSpanKind.INTERNAL);
+        assert.strictEqual(spans[3].parentSpanId, spans[2].spanId);
```
pichlermarc, line 753: reviewer note: the same thing as in the express instrumentation is happening here as well, SpanKind SDK vs. OTLP-representation and an additional span that is inserted at positon 0.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
JamieDanielson: Oh wow, this makes so much more sense. Thanks for fixing this!

JamieDanielson (approved):
This is great, thank you for fixing up the confusion with span kind and sorry for causing some of that headache with the tests! @robbkidd and I were just looking at that earlier today thinking something was off there.... This makes much more sense now.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
trentm: Thanks for chasing this down, Marc!

Sorry for the confusion.
FWIW, the `testUtils.runTestFixture()` stuff was based on a similar thing we have at work on a separate repo. In that case the equivalent of the TestCollector class **normalizes the span objects somewhat**. One of those normalizations is to translate the OTLP `span.kind` and `span.status` values to their enum strings. So, test asserts are doing `assertEqual(span.kind, 'SPAN_KIND_CLIENT')` and `assertEqual(span.status.code, 'STATUS_CODE_ERROR')`.

trentm: Again, FWIW, here is the normalization code I mean: https://github.com/elastic/elastic-otel-node/blob/main/packages/mockotlpserver/lib/normalize.js#L114-L154
I'm not proposing we use this in this repo.
I'd vaguely recall seeing something about a canonical JSON representation of OTLP service requests or of OTel signals. Given that, I think that would be a good target for normalization, and then tests could assert against that normalized form.

trentm: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2421 to get one more occurrence of `SpanKind.*` usage in this test file.

PR #2418, opened 2024-09-03
feat(auto-configuration-web): add configuration helper functions for web

martinkuba:
Addresses https://github.com/open-telemetry/opentelemetry-js/issues/4702
Alternative to https://github.com/open-telemetry/opentelemetry-js/pull/4325

This is an alternative to creating a Web SDK package (similar to [Node SDK](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-sdk-node)). Instead of having an SDK class, this PR introduces helper functions for configuring different parts of web instrumentation.

### DIFF
--- a/metapackages/auto-configuration-web/.eslintignore
+++ b/metapackages/auto-configuration-web/.eslintignore
@@ -1,0 +1,2 @@
+build
+.eslintrc.js
--- a/metapackages/auto-configuration-web/.eslintrc.js
+++ b/metapackages/auto-configuration-web/.eslintrc.js
@@ -1,0 +1,8 @@
+module.exports = {
+    "env": {
+        "commonjs": true,
+        "node": true,
+        "mocha": true,
+    },
+    ...require('../../eslint.config.js')
+}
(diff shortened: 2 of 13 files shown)


pichlermarc (dismissed):
@martinkuba overall structure looks promising :slightly_smiling_face:  I do have a few questions I have around allowing multiple ways to set the same thing and the complexity that entails. :thinking: Having occasionally worked on `NodeSDK`, I've seen this get out of control quite quickly. Therefore I'd recommend doing an iterative approach (starting out with very few duplicate configurations, then adding more where necessary to make it simpler for the user). This can make maintenance easier going forward (and possibly also reduce user-confusion).

Another note: before we can continue we'll need to find owners that will be assigned to this package and will maintain it going forward.

dyladan:
@martinkuba this is quite an old draft. Is this something that can be safely closed or are you still working on this?

martinkuba:
> @martinkuba this is quite an old draft. Is this something that can be safely closed or are you still working on this?

We might want to revisit this in the future, but I am ok with closing this for now.

PR #2424, opened 2024-09-09
fix(detector-gcp): suppress internal tracing

david-luna:

- this detector performs some request to get metadata from GCP. We should suppress the tracing for these requests

ref: #2320 

- suppress tracing when resolving attributes
- add test

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
+++ b/detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
@@ -51,7 +51,7 @@ class GcpDetector implements DetectorSync {
       );
       return new Resource({}, attributes);
     }
-    
+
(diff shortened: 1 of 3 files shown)


detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
```diff
@@ -44,7 +45,10 @@ import {
  */
 class GcpDetector implements DetectorSync {
```
pichlermarc, line 46: I just saw that this depends on `@opentelemetry/resource@^1.0.0` but the `DetectorSync` was added in a later version. I think we may need to update the package to the version where `DetectorSync` was first added to avoid compile issues. :grimacing:

pichlermarc (approved):
thanks :+1:

PR #2425, opened 2024-09-09
fix: auto-config of baggage propagator

joekhoobyar:

- When supplying `OTEL_PROPAGATORS` via an env var, `baggage` is mistakenly mapped to to the `W3CTraceContextPropagator`

- properly map `baggage` to to the `W3CBaggagePropagator `


pichlermarc (approved):
oof, good catch - thanks for fixing this :+1:

PR #2429, opened 2024-09-12
fix(detector-alibaba): suppres internal tracing

david-luna:

- this detector performs some request to get metadata from Alibaba cloud. We should suppress the tracing for these requests

ref: #2320

- suppress tracing when resolving attributes
- add test

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
+++ b/detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
@@ -56,7 +56,7 @@
     "@opentelemetry/api": "^1.0.0"
   },
   "dependencies": {
-    "@opentelemetry/resources": "^1.0.0",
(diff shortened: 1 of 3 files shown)


detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
```diff
@@ -56,7 +56,7 @@
     "@opentelemetry/api": "^1.0.0"
   },
   "dependencies": {
-    "@opentelemetry/resources": "^1.0.0",
+    "@opentelemetry/resources": "^1.10.0",
```
david-luna, line 59: note for reviewer: `DetectorSync` interface appeared in `v1.10.0`. this is a leftover from #2328

pichlermarc (approved):

legendecas (approved):
Thanks!

PR #2439, opened 2024-09-20
ci: pin mssql server version

pichlermarc:
Currently containers are failing to initialize - checking to see if this is related to a recent push of the 2017-latest mssql server image.


pichlermarc:
hmm, maybe something changed with the runners.

PR #2441, opened 2024-09-20
fix: update references to logging exporter

codeboten:
This exporter has been replaced by the debug exporter and will be removed soon. Related to https://github.com/open-telemetry/opentelemetry-collector/pull/11037


trentm:
Thanks. Do you think we'll also need to update the version from otel/opentelemetry-collector-contrib:0.61.0. I'm guessing that version in ancient at this point.

trentm:
> Do you think we'll also need to update the version from otel/opentelemetry-collector-contrib:0.61.0.

I found it was necessary to update, along with a number of other changes to get the example to work.
I hope it is okay that I lifted your changes. See #2465

codeboten:
Closing in favour of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2465

PR #2444, opened 2024-09-23
chore(detector-contrainer): remove orphan codecov script

pichlermarc:

- we've switched to the codecov action, this script is unused and the dependency has also been removed already


pichlermarc:
failing tests will be fixed by #2446

david-luna (approved):

PR #2446, opened 2024-09-23
c: remove orphan codecov script

pichlermarc:
Trying to see if tests work with the 2019 version of mssql server

### DIFF
--- a/.github/workflows/test-all-versions.yml
+++ b/.github/workflows/test-all-versions.yml
@@ -26,7 +26,7 @@ jobs:
         ports:
         - 27017:27017
       mssql:
-        image: mcr.microsoft.com/mssql/server:2017-latest
+        image: mcr.microsoft.com/mssql/server:2019-latest
         env:
           SA_PASSWORD: mssql_passw0rd
           ACCEPT_EULA: Y
--- a/packages/opentelemetry-test-utils/src/test-utils.ts
+++ b/packages/opentelemetry-test-utils/src/test-utils.ts
@@ -39,7 +39,7 @@ const dockerRunCmds = {
   memcached:
     'docker run --rm -d --name otel-memcached -p 11211:11211 memcached:1.6.9-alpine',
   mssql:
-    'docker run --rm -d --name otel-mssql -p 1433:1433 -e SA_PASSWORD=mssql_passw0rd -e ACCEPT_EULA=Y mcr.microsoft.com/mssql/server:2017-latest',
+    'docker run --rm -d --name otel-mssql -p 1433:1433 -e SA_PASSWORD=mssql_passw0rd -e ACCEPT_EULA=Y mcr.microsoft.com/mssql/server:2019-latest',
   mysql:
     'docker run --rm -d --name otel-mysql -p 33306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=test_db -e MYSQL_USER=otel -e MYSQL_PASSWORD=secret mysql:5.7 --log_output=TABLE --general_log=ON',
   postgres:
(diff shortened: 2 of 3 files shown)


.github/workflows/test-all-versions.yml
```diff
@@ -26,14 +26,14 @@ jobs:
...
-        image: mcr.microsoft.com/mssql/server:2017-latest
+        image: mcr.microsoft.com/mssql/server:2022-latest
         env:
           SA_PASSWORD: mssql_passw0rd
           ACCEPT_EULA: Y
         ports:
         - 1433:1433
         options: >-
-          --health-cmd "/opt/mssql-tools/bin/sqlcmd -U sa -P $SA_PASSWORD -Q 'select 1' -b -o /dev/null"
+          --health-cmd "/opt/mssql-tools18/bin/sqlcmd -S localhost -U sa -P $SA_PASSWORD -C -Q 'select 1' -b -o /dev/null"
```
pichlermarc, line 36: note for reviewers: that's the path from MSSQL server 2018 and up - as we're using MSSQL server 2022 I assumed there would be a `/opt/mssql-tools22/` directory but inspecting the file system of a running container shows that there's only `/opt/mssql-tools18/`

`-C` ignores the self-signed certificate

packages/opentelemetry-test-utils/src/test-utils.ts
```diff
@@ -39,7 +39,7 @@ const dockerRunCmds = {
   memcached:
     'docker run --rm -d --name otel-memcached -p 11211:11211 memcached:1.6.9-alpine',
   mssql:
-    'docker run --rm -d --name otel-mssql -p 1433:1433 -e SA_PASSWORD=mssql_passw0rd -e ACCEPT_EULA=Y mcr.microsoft.com/mssql/server:2017-latest',
+    'docker run --rm -d --name otel-mssql -p 1433:1433 -e SA_PASSWORD=mssql_passw0rd -e ACCEPT_EULA=Y mcr.microsoft.com/mssql/server:2019-latest',
```
trentm, line 42: Curious why "2019" here, but "2022" above.

.github/workflows/test-all-versions.yml
trentm: The best docs I've found for this image is the "Overview" section here: https://hub.docker.com/r/microsoft/mssql-server

> Note Starting with SQL Server 2022 CU 14, we are updating SQL Server 2022 container images to include the new mssql-tools18 package. With the introduction of SQL Server 2022 CU 14, and in all future container images, the previous directory /opt/mssql-tools/bin will be phased out.

Where by "phased out" they mean: just gone. ;)

The version for the "mssql-tools" dir is, at least somewhat, independent of the mssql-server package. I believe that's why it is "mssql-tools18" even for versions of MSSQL server beyond 2018.

trentm (commented):
FWIW, the `SA_PASSWORD` envvar is (long since) deprecated in favour of `MSSQL_SA_PASSWORD`.
https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-configure-environment-variables?view=sql-server-2017

However, that could be updated in a separate PR.

packages/opentelemetry-test-utils/src/test-utils.ts
trentm:
```suggestion
    'docker run --rm -d --name otel-mssql -p 1433:1433 -e SA_PASSWORD=mssql_passw0rd -e ACCEPT_EULA=Y mcr.microsoft.com/mssql/server:2022-latest',
```

trentm (approved):

JamieDanielson (approved):

PR #2448, opened 2024-09-23
fix(detector-container): properly detect container ID when using Podman

sjinks:

When using Podman, the container detector is unable to determine the ID of the container. This happens because the detector expects `/containers/<container ID>` substring in the `mountinfo` but Podman uses `/overlay-containers/<container ID>` instead.

For example:
```
$ podman run -it --rm alpine grep hostname /proc/self/mountinfo
4282 4276 0:94 /containers/overlay-containers/b9bb2fc96c57ef741ffbaaed3c0ac16a70705ba27443ce1a7d73554c9a63d125/userdata/hostname /etc/hostname rw,nosuid,nodev,relatime - tmpfs tmpfs rw,size=6557728k,nr_inodes=1639432,mode=700,uid=1000,gid=1000,inode64
```

For docker:
```
$ docker run -it --rm alpine grep hostname /proc/self/mountinfo
4204 4195 259:2 /var/lib/docker/containers/0848110e7e6a6fe6317d104e099286860b03962119a1a124d9dd6ec4c4dc4b11/hostname /etc/hostname rw,relatime - ext4 /dev/nvme0n1p2 rw,errors=remount-ro
```

Fix Container ID detection in Podman containers


david-luna:
@sjinks it seems you have a linting issue

sjinks:
@david-luna I have fixed them now

david-luna (approved):

PR #2449, opened 2024-09-24
feat(instrumentation-dataloader): Enhance `dataloader` instrumentation.

onurtemizkan:

- Adds missing instrumentation for `prime`, `clear` and `clearAll` methods of `dataloader`
- Adds new attributes to `dataloader` spans.

Related: https://github.com/getsentry/sentry-javascript/pull/13664 and https://github.com/getsentry/sentry-javascript/issues/13724

The current implementation of dataloader instrumentation creates spans for only `load`, `loadMany` and `batch` operations. With this PR, the instrumentation starts creating spans for synchronous 'non-read' operations, `prime`, `clear`, and `clearAll`. 

This PR also attaches new attributes to `load`, `loadMany` and newly added `prime`, `clear` spans.

- `cache.key` keeps an array of keys involved in the operation
- `cache.hit` marks if the read operation is not a miss
- `cache.item_size` marks the size of the read / set / deleted value in the cache.

Note to the reviewer(s): I added these attributes to be able to supply information about cache operations in Sentry. IMHO this information is still valuable anywhere else. Alternatively, we can add a `requestHook` to this integration and set those attributes in Sentry SDK.


plugins/node/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -242,10 +250,139 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
         // .loadMany never rejects, as errors from internal .load
         // calls are caught by dataloader lib
         return original.call(this, ...args).then(value => {
+          span.setAttribute('cache.key', Array.from(args[0]));
+          span.setAttribute(
+            'cache.hit',
```
AbhiPrasad, line 255: Unfortunately these are not official OTEL semantic conventions, so I think we cannot add these here. I think we'll need `requestHook` and add these ourselves. For now we should remove this.

There are no cache conventions for OTEL yet. We're thinking about donating our conventions upstream for this.

onurtemizkan: @AbhiPrasad - Should we add the `requestHook` with these updates, or in a separate PR?

AbhiPrasad: Let's add `requestHook` in a separate PR

onurtemizkan:
@open-telemetry/javascript-approvers - can I have 👀 on this please?

pichlermarc:
@onurtemizkan the failing build is likely what's holding people off from reviewing the PR.

I was also unable to reach the current component owner (@henrinormak), so it's likely that this component will soon be marked as unmaintained.

pichlermarc:
(removing assignees as they were added due to a previous diff of this PR including changes to `@opentelemetry/instrumentation-express`)

pichlermarc (approved):
proxy-approving for component-owner @henrinormak 

(@henrinormak, if you find some time, please head over to #2600 and go though the process of joining the GitHub org so that we can assign you triager permissions 🙂)

PR #2456, opened 2024-09-27
fix(pg): Do not add SQLCommenter comments to prepared statements

raphael-theriault-swi:

Currently the `pg` instrumentation will add SQLCommenter comments to any query it can, even prepared ones. In the case of prepared statements this usually results in the trace context within the comment being incorrect for all but the first instance of the query being run.

This adds a simple check to skip adding the comment for prepared statements. See https://node-postgres.com/features/queries#prepared-statements.


maryliag:
@pichlermarc is this consider a breaking change? Just want to make sure everything is properly flagged before giving any approvals

raphael-theriault-swi:
Wouldn't this be a bugfix rather than a breaking change given this behaviour is always incorrect ?

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -297,10 +297,11 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
 
         // Modify query text w/ a tracing comment before invoking original for
         // tracing, but only if args[0] has one of our expected shapes.
+        // Also omit the tracing comment if args[0] is a prepared query object.
         if (instrumentationConfig.addSqlCommenterCommentToQueries) {
           args[0] = firstArgIsString
             ? addSqlCommenterComment(span, arg0)
-            : firstArgIsQueryObjectWithText
+            : firstArgIsQueryObjectWithText && !('name' in arg0)
```
blumamir, line 304: This section is already quite complex IMO with the double js ternary operator.

While the comment above is useful, I can see how future readers might get confused figuring out that the `name` in `arg0` when it's an object is an indication for prepared statement.

The link you gave in the PR body says that:

> PostgreSQL has the concept of a [prepared statement](https://www.postgresql.org/docs/9.3/static/sql-prepare.html). node-postgres supports this by supplying a name parameter to the query config object

I suggest refactoring this part to make it clear why this check is applied

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -948,6 +948,37 @@ describe('pg', () => {
...
+          const [span] = memoryExporter.getFinishedSpans();
+          const commentedQuery = addSqlCommenterComment(
+            trace.wrapSpanContext(span.spanContext()),
+            query
+          );
+
+          const executedQueries = getExecutedQueries();
+          assert.equal(executedQueries.length, 1);
+          assert.equal(executedQueries[0].text, query);
+          assert.notEqual(query, commentedQuery);
```
blumamir, line 975: Does this line assert something meaningful?

If I get it right, `query` is a static text `SELECT NOW()` and `commentedQuery` will contain the comment even with your fix since the PR does not change it.

blumamir (commented):
Thanks for contributing this fix @raphael-theriault-swi 

This LGTM, added 2 nits. I don't think this is a breaking change, as current behavior is a bug.

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
raphael-theriault-swi: Yeah you're correct this doesn't assert anything useful, removing it.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
raphael-theriault-swi: Should be a lot clearer now.

maryliag (approved):
Thank you for working on these changes and the feedback!

blumamir (approved):
Thank you for the contribution.

PR #2457, opened 2024-09-30
refactor(instr-undici): avoid use of deprecated APIs for diagnostic channels

david-luna:

`undici` instrumentation is used deprecated APIs to subscribe to and unsubscribe from diagnostic channels. Eventually these APIs will be removed and the instrumentation won't work with newer versions of Node.js. I've considered to classify it as a fix but the deprecated APIs are still present so I think refactor is better.

- use new `diagnostics_channel.subscribe(name, handler)` and `diagnostics_channel.unsubscribe(name, handler)`  APIs if available and fallback to `diagnostics_channel.channel(name)` if not

NOTE: this PR supersedes #2407


plugins/node/instrumentation-undici/src/internal-types.ts
```diff
@@ -13,14 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { Channel } from 'diagnostics_channel';
+
+import type { Channel, ChannelListener } from 'diagnostics_channel';
 
 import { UndiciRequest, UndiciResponse } from './types';
 
+export interface DiagnosticChannelApi {
```
trentm, line 21: Perhaps "DiagnosticsChannelApi" (i.e. plural "Diagnostics" to match `diagnostics_channel`).

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -139,12 +140,35 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
+    // greater or equal
+    // ref: https://github.com/nodejs/node/pull/47520
+    const [major, minor] = process.version
+      .replace('v', '')
+      .split('.')
+      .map(n => Number(n));
+    const isCompatible = major > 18 || (major === 18 && minor >= 19);
+
+    let unsubscribe: () => void;
+    if (typeof diagchApi.subscribe === 'function' && isCompatible) {
```
trentm, line 159:
```suggestion
    const useNewSubscribe = major > 18 || (major === 18 && minor >= 19);

    let unsubscribe: () => void;
    if (useNewSubscribe) {
```

nit: I think you could drop the `typeof diagchApi.subscribe === 'function'` check, now that it is being gated on being >=18.19.

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -139,12 +140,35 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     diagnosticChannel: string,
     onMessage: ListenerRecord['onMessage']
   ) {
-    const channel = diagch.channel(diagnosticChannel);
-    channel.subscribe(onMessage);
+    // NOTE: later versions of `@types/node` include proper type definitions
+    // for diagnostics channel. However updating it breaks compilation :(
```
trentm, line 144: Should there be a separate issue to follow-up on that? Or is there one already?

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -139,12 +140,35 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
+    // Hence we have a helper interface to deal with it
+    const diagchApi = diagch as unknown as DiagnosticChannelApi;
+
+    // NOTE: diagnostic channel had an issue tht was solved in v18.19.0 so
+    // instrumentation is going to use the new API when the process version is
+    // greater or equal
+    // ref: https://github.com/nodejs/node/pull/47520
```
trentm, line 151:
```suggestion
    // `diagnostics_channel` had a ref counting bug until v18.19.0.
    // https://github.com/nodejs/node/pull/47520
```

nit

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -139,12 +140,35 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
+    } else {
+      const channel = diagchApi.channel(diagnosticChannel);
+      channel.subscribe(onMessage);
+      unsubscribe = () => channel.unsubscribe(onMessage);
+    }
+
     this._channelSubs.push({
       name: diagnosticChannel,
-      channel,
       onMessage,
```
trentm, line 170:
```suggestion
```

I guess the `onMessage` could be dropped from the ListenerRecord now, because it is no longer used. It is captured in the `unsubscribe` closure.

plugins/node/instrumentation-undici/src/undici.ts
david-luna: I think is related to typescript because it spits typing errors from other modules. I guess we could ref to https://github.com/open-telemetry/opentelemetry-js/issues/4870

david-luna:
@trentm thanks for the review. I've applied your suggestions myself so it's only necessary your approval (not sure of the approve settings in this repo)

trentm (approved):

PR #2458, opened 2024-10-01
chore: remove usages of Span constructor

david-luna:

The Span constructor from ' @opentelemetry/sdk-trace-base` is planned for removal from the public API ([ref](https://github.com/open-telemetry/opentelemetry-js/issues/3597)).

This PR anticipates the possible breaking changes derived from it

- Remove usage of `new Span` constructor in favour of `Tracer.startSpan`

### DIFF
--- a/packages/baggage-span-processor/src/index.ts
+++ b/packages/baggage-span-processor/src/index.ts
@@ -14,5 +14,5 @@
  * limitations under the License.
  */
 
-export * from './baggage-span-processor';
-export * from './types';
+export { BaggageSpanProcessor } from './baggage-span-processor';
+export { ALLOW_ALL_BAGGAGE_KEYS, BaggageKeyPredicate } from './types';
--- a/packages/baggage-span-processor/test/baggage-span-processor.test.ts
+++ b/packages/baggage-span-processor/test/baggage-span-processor.test.ts
@@ -16,12 +16,7 @@
 
 import { BaggageSpanProcessor } from '../src/baggage-span-processor';
 import { ALLOW_ALL_BAGGAGE_KEYS } from '../src/types';
-import {
-  propagation,
-  ROOT_CONTEXT,
-  SpanKind,
-  TraceFlags,
-} from '@opentelemetry/api';
+import { propagation, ROOT_CONTEXT, SpanKind } from '@opentelemetry/api';
 import { BasicTracerProvider, Span } from '@opentelemetry/sdk-trace-base';
 import { expect } from 'expect';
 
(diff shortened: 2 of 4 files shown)


packages/baggage-span-processor/test/baggage-span-processor.test.ts
```diff
@@ -131,17 +120,14 @@ describe('BaggageSpanProcessor with regex key filter', () => {
...
       {
-        traceId: 'e4cda95b652f4a1592b449d5929fda1b',
-        spanId: '7e0c63257de34c92',
-        traceFlags: TraceFlags.SAMPLED,
+        kind: SpanKind.SERVER,
       },
-      SpanKind.SERVER
-    );
+      ROOT_CONTEXT
+    ) as unknown as Span;
```
MikeGoldsmith, line 130: Do we need both type coercions here - can we coerce straight into `Span`?

plugins/node/opentelemetry-instrumentation-dns/test/functionals/utils.test.ts
```diff
@@ -154,13 +154,14 @@ describe('Utility', () => {
...
+      const span = tracer.startSpan(
         'test',
-        { spanId: '', traceId: '', traceFlags: TraceFlags.NONE },
-        SpanKind.INTERNAL
-      );
+        {
+          kind: SpanKind.INTERNAL,
+        },
+        ROOT_CONTEXT
+      ) as unknown as Span;
```
MikeGoldsmith, line 164: Same as above.

MikeGoldsmith (approved):
Looks good to me, thanks @david-luna. I've left one query regarding type coercion.

MikeGoldsmith:

Looks like the warning will be for [opentelemetry-instrumentation-dns](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-dns), not baggage-span-procesor. We may need to break up the PR to address the different components due to their level of ownership / maintenance.

pichlermarc:
@MikeGoldsmith I'll sponsor the PR, then we don't need to break it up :slightly_smiling_face:

plugins/node/opentelemetry-instrumentation-dns/test/functionals/utils.test.ts
```diff
@@ -154,13 +154,14 @@ describe('Utility', () => {
   describe('setError()', () => {
     it('should have error attributes', () => {
       const errorMessage = 'test error';
-      const span = new Span(
-        new BasicTracerProvider().getTracer('default'),
-        ROOT_CONTEXT,
+      const tracer = new BasicTracerProvider().getTracer('default');
+      const span = tracer.startSpan(
```
pichlermarc, line 158: It looks like all this test does is to check if `Span#setStatus()` and `Span#addAttributes()` is called. I wonder if we should just create mock span using `sinon` and assert that these methods are called with the expected data. This way we could eliminate the need for this test to use the SDK completely.

david-luna: @pichlermarc I think is a good idea. I'll add it in the next commit :)

packages/baggage-span-processor/test/baggage-span-processor.test.ts
david-luna: Not necessary at all. Removing it in the next commit. Good catch :)

packages/baggage-span-processor/src/index.ts
```diff
@@ -14,5 +14,5 @@
  * limitations under the License.
  */
 
-export * from './baggage-span-processor';
-export * from './types';
+export { BaggageSpanProcessor } from './baggage-span-processor';
+export { ALLOW_ALL_BAGGAGE_KEYS, BaggageKeyPredicate } from './types';
```
david-luna, line 18: note to reviewer: https://github.com/open-telemetry/opentelemetry-js/issues/4186

plugins/node/opentelemetry-instrumentation-dns/test/functionals/utils.test.ts
david-luna: Done!

packages/baggage-span-processor/test/baggage-span-processor.test.ts
david-luna: Removed!

plugins/node/opentelemetry-instrumentation-dns/test/functionals/utils.test.ts
david-luna: Removed!

pichlermarc (approved):
Thanks @david-luna :slightly_smiling_face: 
approving for `@opentelemetry/instrumentation-dns` :slightly_smiling_face:

MikeGoldsmith (approved):
Looks good for me too - thanks @david-luna 🎉

PR #2462, opened 2024-10-04
chore: update @types/node to 18.18.14

david-luna:

- bump `@types/node` in all packages to `18.18.14`
- fix a type error in `plugins/node/instrumentation-amqplib/src/utils.ts`


pichlermarc (approved):
thanks :+1:

PR #2467, opened 2024-10-09
feat(auto-instrumentations-node)!: disable @opentelemetry/instrumentation-fs by default

### LINKED ISSUE #2453: [auto-instrumentations-node] disabling/removing `fs` instrumentation
## Description 

The `@opentelemetry/autoinstrumentations-node` package is usually used to get a quick-start on instrumenting with OpenTelemetry JS, however, there are a few instrumentations that behave in very verbose ways:

- `@opentelemetry/instrumentation-fs`
- `@opentelemetry/instrumentation-net`
- `@opentelemetry/instrumentation-dns`

This proposal is limited to `@opentelemetry/instrumentation-fs`, as it is the most verbose of them, creating error spans that may be picked up by error reporting even though the error is expected by the user. An example of this includes trying to access a file that does not exist, then catching the error.  The `@opentelemetry/instrumentation-fs` package will then create an error span, even though this behavior is completely intended by the user. Acting in such a way is even recommened by the Node.js documentation, [as checking with `fs.access()` before accessing the file can introduce a race condition](https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsaccesspath-mode-callback).

In addition to the error spans, `@opentelemetry/instrumentation-fs` also generates a lot of spans during startup, when files are `require`d. This causes increased memory pressure for `fs` calls that are most likely not relevant for the user.

## Proposal

(1) I propose we remove `@opentelemetry/instrumentation-fs` from the automatically enabled instrumentations that are returned by `getNodeAutoinstrumentations()` unless:
- `OTEL_JS_ENABLED_INSTRUMENTATIONS` constains `fs`
- or `@opentelemetry/instrumentation-fs` is explicitly enabled when calling `getNodeAutoinstrumentations()`

This has the drawback of breaking users that rely on the telemetry generated by this instrumentation. If they want to get it back, they have to explicitly enable it via one of the above options. `OTEL_JS_ENABLED_INSTRUMENTATIONS` is an allow-list so they will have to also explicitly list every other instrumentation they want to have enabled, which may be tedious.

If there are no objections, I will go forward with this approach 2 weeks after this issue was opened.

**Other options:**
- (2) removing `@opentelemetry/instrumentation-fs` from `@opentelemetry/autoinstrumentations-node` (this will break `getNodeAutoinstrumentations()` users, but they will not be able to get the instrumentation back by listing the instrumentation via configuration only. They will have to add a dependency to the package and instantiate and register the instrumentation themselves)
- (3) removing `@opentelemetry/instrumentation-fs` but introducing "profiles" (or "groups" of instrumentations) that can be enabled 
  - `OTEL_JS_ENABLED_INSTRUMENTATIONS=group:legacy` (would allow users to go back to the "previous" state)
  - this has the drawback of adding yet another way of configuring the `@opentelemetry/autoinstrumentations-node`, which means additional maintenance overhead, possible user-confusion.
- (4) other options I did not think of yet (I will add them here as they come up in discussion)

Once we have decided on a way forward, we can give a similar treatment to `@opentelemetry/instrumentation-net` and `@opentelemetry/instrumentation-dns`

## Additional context:
- An issue about memory use: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1344
- `@opentelemetry/instrumentation-fs` acting as the example in a description of how to disable instrumentations: 
  - https://opentelemetry.io/docs/languages/js/libraries/#registration
  - https://opentelemetry.io/docs/zero-code/js/configuration/#disable-specific-instrumentations
  - https://opentelemetry.io/docs/demo/services/payment/#initializing-opentelemetry
  - https://github.com/open-telemetry/opentelemetry-js-contrib/blob/7898458b4d15335f376cc1aae94cc73c0d7d96c7/metapackages/auto-instrumentations-node/README.md?plain=1#L94-L97

pichlermarc:

see #2453 - many users seek to disable `@opentelemetry/instrumentation-fs` due to it's verbosity. This is so common that we have added instructions on how to disable it everywhere. 

This PR disables the instrumentation by default. To get back to the previous behavior, end-users can add 'fs' to a comma seperated list in `OTEL_NODE_ENABLED_INSTRUMENTATIONS` when using enviornment variable configuration, or they can explicitly enable the instrumentation like so when using `getNodeAutoInstrumentations()`:
```ts
    getNodeAutoInstrumentations({
      '@opentelemetry/instrumentation-fs': {
        enabled: true,
      },
    }),
```
### I need `@opentelemetry/instrumentation-fs` back, how do I get back to the previous state?

**If you're using `@opentelemetry/auto-instrumentations-node/register` without `OTEL_NODE_ENABLED_INSTRUMENTATIONS` set**
- set `OTEL_NODE_ENABLED_INSTRUMENTATIONS=amqplib,aws-lambda,aws-sdk,bunyan,cassandra-driver,connect,cucumber,dataloader,dns,express,fastify,fs,pool,graphql,grpc,hapi,http,ioredis,kafkajs,knex,koa,memoizer,memcached,instrumentation-mongodb,instrumentation-mongoose,instrumentation-mysql2,instrumentation-mysql,instrumentation-nestjs-core,instrumentation-net,instrumentation-pg,instrumentation-pino,instrumentation-redis,instrumentation-redis-4,instrumentation-restify,instrumentation-router`

**If you're using `getAutoInstrumentationsNode()` without an options object that contains `@opentelemetry/fs` and without `OTEL_NODE_ENABLED_INSTRUMENTATIONS` set**
- set `OTEL_NODE_ENABLED_INSTRUMENTATIONS=amqplib,aws-lambda,aws-sdk,bunyan,cassandra-driver,connect,cucumber,dataloader,dns,express,fastify,fs,pool,graphql,grpc,hapi,http,ioredis,kafkajs,knex,koa,memoizer,memcached,instrumentation-mongodb,instrumentation-mongoose,instrumentation-mysql2,instrumentation-mysql,instrumentation-nestjs-core,instrumentation-net,instrumentation-pg,instrumentation-pino,instrumentation-redis,instrumentation-redis-4,instrumentation-restify,instrumentation-router`

**If you're using `getAutoInstrumentationsNode()` but you don't want to use enviornment variables**
- change the code to
  ```ts
    getNodeAutoInstrumentations({
      '@opentelemetry/instrumentation-fs': {
        enabled: true,
      },
    }),
  ```

Fixes #2453


trentm:
> To get back to the previous behavior, end-users can add 'fs' to a comma seperated list in `OTEL_NODE_ENABLED_INSTRUMENTATIONS`

nit: The linked issue points out this change for the (expected to be very rare) user that relies on the `fs` instrumentation (emphasis mine):

> This has the drawback of breaking users that rely on the telemetry generated by this instrumentation. If they want to get it back, they have to explicitly enable it via one of the above options. **OTEL_NODE_ENABLED_INSTRUMENTATIONS is an allow-list so they will have to also explicitly list every other instrumentation they want to have enabled, which may be tedious.**

Ideally it would be nice to point this out in the changelog. However, I don't recall if we have a way for a PR to add some explicit prose to the changelog.  A fallback is to perhaps clearly point out what the user needs to do *in the PR description*?

trentm (approved):

david-luna (approved):

pichlermarc:
> > To get back to the previous behavior, end-users can add 'fs' to a comma seperated list in `OTEL_NODE_ENABLED_INSTRUMENTATIONS`
> 
> nit: The linked issue points out this change for the (expected to be very rare) user that relies on the `fs` instrumentation (emphasis mine):
> 
> > This has the drawback of breaking users that rely on the telemetry generated by this instrumentation. If they want to get it back, they have to explicitly enable it via one of the above options. **OTEL_NODE_ENABLED_INSTRUMENTATIONS is an allow-list so they will have to also explicitly list every other instrumentation they want to have enabled, which may be tedious.**
> 
> Ideally it would be nice to point this out in the changelog. However, I don't recall if we have a way for a PR to add some explicit prose to the changelog. A fallback is to perhaps clearly point out what the user needs to do _in the PR description_?

Good point. I've added some info to the PR description. I have not found a simple way to have the auto-generated changelog include details, but I'll modify the Github Release to include that info as well 🙂

ilushaR:
@pichlermarc  the fix isn’t working as expected.

Even when I use ```'@opentelemetry/instrumentation-fs': { enabled: true }```, the instrumentation still doesn’t enable because it isn’t included in the `enabledInstrumentationsFromEnv` list.

[This](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/src/utils.ts#L167) check seems incorrect.

pichlermarc:
> @pichlermarc the fix isn’t working as expected.
> 
> Even when I use `'@opentelemetry/instrumentation-fs': { enabled: true }`, the instrumentation still doesn’t enable because it isn’t included in the `enabledInstrumentationsFromEnv` list.
> 
> [This](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/src/utils.ts#L167) check seems incorrect.

Hi @ilushaR - thanks for letting me know. :slightly_smiling_face:  I'll have a look. I opened #2515 to follow up.

(FYI: in the future please always open a new issue rather then commenting on a closed PR/issue, opening a new one will ensure that we see a bug report - I usually filter any activity on closed PRs/issues. :slightly_smiling_face: )

PR #2470, opened 2024-10-10
refactor(instrumentation-mysql2): improve performance of getSpanName using substring

geotry:

We noticed that `getSpanName()` from `instrumentation-mysql2` was taking non-negligeable time from our profiler in production, considering that it's in a critical path.

This PR replaces split() by substring() to avoid instanciating a new array everytime.

Here is a benchmark with the difference before and after (~20 times faster for me): https://jsbm.dev/YtumA6nRNvjVJ

Use substring() with indexOf() instead of split() to extract the SQL verb from a sql query. It also checks for queries with no spaces like "COMMIT" and fallback to return the raw query.


david-luna:
@geotry 

Thanks for contributing to OpenTelemetry :) 

Could you run `npm run lint:fix` and push the changes to pass all checks?

geotry:
thanks for the review

@maryliag I implemented the change in instrumentation-mysql and added tests to cover getSpanName().
I c/c the same code from instrumentation-mysql2 for consistency but it was slightly different as it was returning the full query string in case of an object instead of just the verb. I'm not sure wether it's safe or not to apply this change, let me know and I can revert it 👍 

I also fix the lint issues, sorry about that @david-luna

maryliag:
Thank you for updating the other package too! 😄 
I'm waiting for a maintainer to approve the workflows so we can see what is failing and get a fix for it, then we can get this merged

david-luna (approved):

Issue #2472, opened 2024-10-10
@opentelemetry/auto-instrumentations-node 0.50.1 breaks when using Yarn PnP because of missing dependencies 


joekrill:
### What version of OpenTelemetry are you using?

0.50.1

### What version of Node are you using?

22.8.0

### What did you do?

Upgraded @opentelemetry/auto-instrumentations-node from 0.50.0 to 0.50.1

### What did you expect to see?

No errors.

### What did you see instead?

The following error:

> Error: @opentelemetry/instrumentation-pg tried to access @opentelemetry/core, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

I "fixed" that by adding the following to my `.yarnrc.yml`:

```yaml
packageExtensions:
  "@opentelemetry/instrumentation-pg@*":
    dependencies:
      "@opentelemetry/core": "*"
```

But that just caused this error: 

> Error: @opentelemetry/resource-detector-alibaba-cloud tried to access @opentelemetry/core, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

I didn't go further because I suspect fixing that would just cause similar errors from other related packages. Ultimately `@opentelemetry/instrumentation-pg`, `@opentelemetry/resource-detector-alibaba-cloud`, and (potentially) others _should_ include the dependency on ` @opentelemetry/core` in their `package.json`.

pichlermarc:
Thanks for reporting this @joekrill 

> I didn't go further because I suspect fixing that would just cause similar errors from other related packages. Ultimately @opentelemetry/instrumentation-pg, @opentelemetry/resource-detector-alibaba-cloud, and (potentially) others should include the dependency on  @opentelemetry/core in their package.json.

Yes, you're right, these two were missing the dependencies as well - I did not find any other instances. I opened a PR #2473 to fix this.

pichlermarc:
@joekrill a new release is out now, please let me know if that fixed the issue.

PR #2473, opened 2024-10-10
fix: add missing @opentelemetry/core dependency

pichlermarc:

See #2472, we're using `@opentelemetry/core` but don't have it listed as a dependency in some packages.

- adds the dependency to the packages where it's missing (identified through comparing https://github.com/open-telemetry/opentelemetry-js-contrib/compare/auto-instrumentations-node-v0.50.0...auto-instrumentations-node-v0.50.1)
- pins the `@opentelemetry/semantic-conventions` dependency for `@opentelemetry/instrumentation-pg`  as it's using the `incubating` entrypoint, which may break on minor bumps.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/package.json
+++ b/plugins/node/opentelemetry-instrumentation-pg/package.json
@@ -73,8 +73,9 @@
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.53.0",
-    "@opentelemetry/semantic-conventions": "^1.27.0",
+    "@opentelemetry/semantic-conventions": "1.27.0",
     "@opentelemetry/sql-common": "^0.40.1",
+    "@opentelemetry/core": "^1.26.0",
     "@types/pg": "8.6.1",
     "@types/pg-pool": "2.0.6"
   },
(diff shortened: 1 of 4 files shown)


maryliag (approved):
Thanks for fixing this!

plugins/node/opentelemetry-instrumentation-pg/package.json
```diff
@@ -73,8 +73,9 @@
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.53.0",
-    "@opentelemetry/semantic-conventions": "^1.27.0",
+    "@opentelemetry/semantic-conventions": "1.27.0",
```
pichlermarc, line 76: drive-by-fix: we're using the `@opentelemetry/semantic-conventions/incubating` entrypoint which may break in minor versions - pinning this to avoid any surprises when we release the `1.28.0`.

trentm: Might want the same at https://github.com/open-telemetry/opentelemetry-js-contrib/blob/645ac2e9fa322503e88283985f3f3bb2349a9729/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json#L63

We should have a lint check for this perhaps.

trentm (approved):

PR #2476, opened 2024-10-10
feat(instrumentation-pg): add error type to db duration metric

maryliag:

- Part Of #2348 

- Add error type to db duration operation metric. It uses the error codes from postgres, so we can limit cardinality and they're well known error codes.


plugins/node/opentelemetry-instrumentation-pg/src/utils.ts
```diff
@@ -252,6 +255,10 @@ export function patchCallback(
     res: object
   ) {
     if (err) {
+      if (Object.keys(err).includes('code')) {
```
david-luna, line 258: nit: maybe is not necessary to create an array of keys to check the presence of `code`. I think you could make use of `Object.hasOwnProperty`

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -1013,5 +1014,53 @@ describe('pg', () => {
...
+        assert.strictEqual(
+          metrics[0].descriptor.description,
+          'Duration of database client operations.'
+        );
+        const dataPoint = metrics[0].dataPoints[0];
+        assert.strictEqual(
+          dataPoint.attributes[SEMATTRS_DB_SYSTEM],
+          DBSYSTEMVALUES_POSTGRESQL
+        );
+        // TODO: add once #2475 is merged
```
david-luna, line 1042: It's approved :)

david-luna (approved):

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
pichlermarc: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2475 is now merged too :slightly_smiling_face:

maryliag: Merged with main and added the code

PR #2464, opened 2024-10-08
chore: update test-all-versions configs to test fewer package versions

trentm:
The TAV tests in CI are taking 2.5h - 3h, for a total usage of over 13h
compute time. E.g.: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/11233652488/usage

This change reduces the number of tested versions from 839 to 317
(for a TAV run with Node.js v20). For a few of the .tav.yml files the
tested version *range* was actually increased to capture some versions
that the instrumentation supported, but were not being tested by TAV.

PR #2478, opened 2024-10-11
chore(instrumentation-pg): use done on test

maryliag:

- Trying to solve the flakiness of the test

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
@@ -491,7 +491,7 @@ describe('pg-pool', () => {
       metricReader = testUtils.initMeterProvider(instrumentation);
     });
 
-    it('should generate `db.client.connection.count` and `db.client.connection.pending_requests` metrics', async () => {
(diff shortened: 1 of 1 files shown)


plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -563,6 +563,7 @@ describe('pg-pool', () => {
             0,
             'expected to have 0 pending requests'
           );
+          done();
```
trentm, line 566: Ah, this is a clearer fix.
I have a similar fix for this test failure here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2464/files?diff=unified&w=1#diff-682483846290c7e14fba934e5f5effaaa2ff8e84e94f511300af6d0120a59251R495-R570
as part of a larger PR

trentm (approved):

trentm:
@maryliag I just now noticed your comment on my other PR: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2464#issuecomment-2407353350

> But got the same issue.


Hrm, I had *thought* that that change made the issue go ahead. I haven't yet seen that error happen after this change (your version or mine). We shall see if it comes up again.

PR #2480, opened 2024-10-11
fix(instrumentation-pg): ensure db.client.operation.duration metric is recorded for Promises API usage of pg

trentm:
Refs: #2380

---

While working on pg tests in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2464 I accidentally noticed that the duration metric added to instr-pg in #2380 didn't work if the Promises API of the pg client was used.


plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -277,6 +277,7 @@ describe('pg-pool', () => {
         await client.query('SELECT NOW()');
       } finally {
         client.release();
+        await newPool.end();
```
trentm, line 280: This is an unrelated change. It fixes a few-second hang in the test file completing (presumably on a pool connection timeout).

trentm:
The codecov issue is that the `recordDuration()` call in the promise rejection case wasn't covered. Meh.

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -567,5 +573,39 @@ describe('pg-pool', () => {
...
+        [],
+        'expected no errors from the callback during metric collection'
+      );
+
+      // We just test the expected metric *names* here. The particulars of the
+      // metric values are already tested in other test cases.
+      const metrics = resourceMetrics.scopeMetrics[0].metrics;
+      assert.strictEqual(
+        metrics[0].descriptor.name,
+        'db.client.operation.duration'
```
maryliag, line 599: nit: can you use the variable here? 
I made a similar update on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2475/files#diff-682483846290c7e14fba934e5f5effaaa2ff8e84e94f511300af6d0120a59251R53

maryliag:
That was an awesome catch! Thanks for finding and fixing it!

maryliag (approved):
a little nit, otherwise LGTM

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
trentm: done in commit 2120edf4

JamieDanielson (approved):

Issue #2479, opened 2024-10-11
Memory leak in instrumentation-pg


yandrushchak:
### What version of OpenTelemetry are you using?
`@opentelemetry/instrumentation-pg: 0.45.1`

### What version of Node are you using?
`v20.15.1`

### What did you do?
Used `@opentelemetry/instrumentation-pg` with TypeORM and pg-pool. Received warning about possible memory leak, which happens since listeners are constantly attached for each query. Partial stack trace:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connect listeners added to [BoundPool]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at _addListener (node:events:593:17)
    at BoundPool.addListener (node:events:611:10)
    at BoundPool.connect (c:/test/node_modules\@opentelemetry\instrumentation-pg\src\instrumentation.ts:461:12)
    at <anonymous> (c:/test/node_modules\typeorm\driver\src\driver\postgres\PostgresDriver.ts:1175:25)
    at new Promise (<anonymous>)
    at PostgresDriver.obtainMasterConnection (c:/test/node_modules\typeorm\driver\src\driver\postgres\PostgresDriver.ts:1174:16)
    at PostgresQueryRunner.connect (c:/test/node_modules\typeorm\driver\src\driver\postgres\PostgresQueryRunner.ts:111:18)
    at PostgresQueryRunner.query (c:/test/node_modules\typeorm\driver\src\driver\postgres\PostgresQueryRunner.ts:248:47)
....
```

### What did you expect to see?
No memory leak.

### Additional context
This issue was introduced in this commit - https://github.com/open-telemetry/opentelemetry-js-contrib/commit/12adb4354f09ade438cd96340bdfd1f715b5fed3#diff-1360de1aed58c38c5cbd3a1759a8af761049ea2c562a477964623fb2358b2730R379
Since [pg.Pool.connect](https://node-postgres.com/apis/pool#poolconnect) is called for each query to obtain connection, duplicated event listeners are attached each time, which would probably also lead to wrong metrics data.

It seems in order to fix it, additional check is needed to make sure event listeners are added only on a first `connect` call.

astrot1988:
same issue

clintonb:
[...]

PR #2484, opened 2024-10-15
fix(instrumentation-pg): not add duplicate listeners to pg pool

maryliag:

- Addresses #2479

- Checks if listeners for pg pool events were already added before trying to add them. 
- Generate poolName only once, and use the same on all updates, instead of calculating every time

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
@@ -66,7 +66,6 @@ import {
 } from '@opentelemetry/semantic-conventions/incubating';
 
 export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConfig> {
-  private _pgPoolListenersSet: boolean = false;
   private _operationDuration!: Histogram;
   private _connectionsCount!: UpDownCounter;
   private _connectionPendingRequests!: UpDownCounter;
@@ -86,7 +85,6 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
   }
 
   override _updateMetricInstruments() {
-    this._pgPoolListenersSet = false;
     this._operationDuration = this.meter.createHistogram(
       METRIC_DB_CLIENT_OPERATION_DURATION,
       {
@@ -437,8 +435,8 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
     };
   }
 
-  private setPoolConnectEventListeners(pgPool: PgPoolExtended) {
-    if (this._pgPoolListenersSet) return;
+  private _setPoolConnectEventListeners(pgPool: PgPoolExtended) {
+    if (pgPool.eventListenersSet) return;
     const poolName = utils.getPoolName(pgPool.options);
 
     pgPool.on('connect', () => {
@@ -480,7 +478,7 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
         this._connectionsCounter
       );
     });
-    this._pgPoolListenersSet = true;
+    pgPool.eventListenersSet = true;
   }
 
   private _getPoolConnectPatch() {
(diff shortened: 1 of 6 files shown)


plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -435,6 +437,52 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
     };
   }
 
+  private setPoolConnectEventListeners(pgPool: PgPoolExtended) {
```
raphael-theriault-swi, line 440: This should probably be prefixed with an underscore to match the other private methods.

maryliag: done

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -435,6 +437,52 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
...
+    if (this._pgPoolListenersSet) return;
```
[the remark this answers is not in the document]
maryliag: I added a new test case to test this scenario and it was indeed failing, so I moved the flag to be owned by each pool. So now each pool is responsible for knowing if their instrumentation listeners have been set properly.

With this all tests are passing:
- calling connect multiple times on the same pool continues to have a single listener
- adding custom listeners worked as expected (with custom + instrumentation listeners being executed)
- all pool created are properly instrumented

plugins/node/opentelemetry-instrumentation-pg/src/internal-types.ts
```diff
@@ -57,6 +57,7 @@ export interface PgPoolOptionsParams {
 
 export interface PgPoolExtended extends pgPoolTypes<pgTypes.Client> {
   options: PgPoolOptionsParams;
+  eventListenersSet: boolean; // flag to identify if the event listeners for instrumentation have been set
```
pichlermarc, line 60: This is not a property that actually exists so we'd be defining it as we go - I think to avoid any problems downstream with adding an enumerable property we should a Symbol as a key like we do in other instrumentations, which won't be enumerable and should be a bit safer for the user.

See [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/6dfe93c01c6a5750d69045a1ae037b8712857b8f/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts#L70-L73) for an example.

pichlermarc (commented):
thanks for taking care of this. :+1: 
One comment about the added property - otherwise this looks good to be merged.

plugins/node/opentelemetry-instrumentation-pg/src/internal-types.ts
maryliag: Done

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -222,6 +225,8 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
   }
 
   private recordOperationDuration(attributes: Attributes, startTime: HrTime) {
+    const emit = utils.getEmitType();
+    if (emit === utils.emitType.OLD_SEM_CONV) return;
```
pichlermarc, line 229: With this the semconv stability env-var this actually becomes a breaking change. I'd recommend splitting it form this PR and only introducing the bugfix first - then we can still follow up with the stability opt-in feature that was requested in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2479#issuecomment-2415615704

pichlermarc (commented):
Thanks for working on this. There's one more comment I have about he semconv stability flag, once addressed we can merge this.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
maryliag: sounds good, working on it

maryliag: done

pichlermarc (approved):
thank you for taking care of this :+1:

PR #2487, opened 2024-10-16
fix: update eslint to use flat config (#1635)

### LINKED ISSUE #1635: Upgrade eslint to flat config
Recently, renovate opened those 2 PRs regarding prettier: #1630  and #1629 

Apparently, eslint has a new way of handling configs called flat configs.
More info here: https://eslint.org/docs/latest/use/configure/configuration-files-new

In #1629 , part of the changelog includes:

> Added: The CLI helper tool now works with eslint.config.js (flat config). Just like ESLint itself, the CLI tool automatically first tries eslint.config.js and then eslintrc, and you can force which one to use by setting the [ESLINT_USE_FLAT_CONFIG] environment variable. Note that the config of eslint-config-prettier has always been compatible with eslint.config.js (flat config) – it was just the CLI tool that needed updating. On top of that, the docs have been updated to mention how to use both eslint.config.js (flat config) and eslintrc, and the tests now test both config systems.

It seems they have moved to a simpler config structure, but for us to use the new versions we have to follow the [Configuration Migration Guide](https://eslint.org/docs/latest/use/configure/migration-guide).

This issue is to track this task. It should probably include:
- bumping the versions of everything eslint and prettier related in the project to latest versions
- Update the project to use the new simplified formats
- Remove all the references to old config objects which should now be part of the root `eslint.configs.js`. For example, the package.json scripts should not include the `--ext .ts` part for all packages in the monorepo in `"lint": "eslint . --ext .ts",`
- Run linting on the project and either fix new lint issues, or apply relevant configuration to disable them.

walter-0:

- Resolves #1635

- Upgrades eslint, @typescript-eslint, and all related plugins
- Rewrites the root eslint config to use the flat style
- Updates lint script in each package.json
- Removes all deprecated .eslintignores
- Removes all eslintrc.js files within each package
- Cleans up the eslint plugins and rules to remove deprecated ones
- Adds `@copyright` and `@license` JSDoc tags to header comment.


walter-0 (commented):
Hi, I'm a new contributor and I have a couple of questions to get this ready.
I updated eslint and ran lint:fix on everything, but there are still some typescript related linting errors. Many of them are `no-explicit-any`, `no-missing-import`. What is the convention for type-related linting errors, and other code smells such as `unused-vars`, `no-unsafe-function-type`.
I think the `no-missing-import` may have something to do with the root tsconfig using `moduleResolution: node`, which probably should be considered deprecated since it's for Node versions older than v10. I wasn't able to get that working on my end, but perhaps there's something I'm not aware of with this project.

PR #2489, opened 2024-10-16
chore: add readme and markdown to default lint

maryliag:
Add the lint checks for markdown and readme to the `npm run lint` command, to align the local command with what is being executed on github workflow.

Without this change the lint was taking 31s, and after it took 33s, so I think the change is acceptable.


package.json
```diff
@@ -24,12 +23,13 @@
     "test-all-versions": "npm run --if-present --workspaces test-all-versions",
     "bump": "lerna publish",
     "changelog": "lerna-changelog",
-    "lint": "lerna run lint",
-    "lint:fix": "lerna run lint:fix",
+    "lint": "lerna run lint && lerna run lint:readme && markdownlint-cli2 $(git ls-files '*.md')",
```
david-luna, line 26: With this change I think the [lint workflow](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/a4fcd26f825278fa769443a0d619705004af948f/.github/workflows/lint.yml#L19) could be simplified. Otherwise we may be running readme and markdown linting twice

david-luna (approved):

PR #2493, opened 2024-10-18
chore: use nx commands

david-luna:

This PR tries to work towards the migration from `lerna` to `nx` https://github.com/open-telemetry/opentelemetry-js/issues/4240. Although the issue is on the core repository I think both repos should be aligned in the tooling. I've just started with this one because it seems more straight-forward.

## Changes

This PR surfaces the `nx@15.9.7` package required by `lerna`. I haven't tried to use a higher version since `lerna` is not completely removed yet and it might have version conflicts.  The `package.json`'s  script section has been changed to use the equivalent `nx` commands for building and testing the packages, with some nuances:
- a `nx.json` file has been added defining task dependencies so some package scripts like `precompile` are not needed
- a cache has been defined for some tasks like `version:update`, `lint` and `compile`. More details below

The workflow does not change. Indeed the development guidelines needs no changes since already recommends to run `npm run compile` at the root folder before start doing changes in a package.

## Cache

`nx` can cache the output of a task like the compilation. Having it active may come handy for developers since you can switch branches and rebuild only what changed just by `npm run compile` at the root. The rest will come from the cache. Cache is located inside `node_modules` so its easy to invalidate via `npm run ci` or just removing the `node_modules/.cache/nx` folder.

### Using it in CI

For unit and TAV tests there is a compilation step before running the test suite. Turns out that the tests are using the `tsconfig.json` to compile the tests before running. The file has the following options
```json
{
  "extends": "../../../tsconfig.base",
  "compilerOptions": {
    "rootDir": ".",
    "outDir": "build"
  },
  "include": [
    "src/**/*.ts",
    "test/**/*.ts"
  ]
}
```
so the tests are compiling again the code in order to run. This is an unnecessary use of resources.

Test jobs in workflows do a clean checkout before testing so running `npm run test` in a package would fail because one or more of this reasons:
- `version.ts` file is missing
- package may depend on other in the same repo (this dependency needs to be compiled)
- package is referring to build folder in tests ([example](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/30e8fc5a2264ffef31804a502af51bc2f79bd2a4/detectors/node/opentelemetry-resource-detector-container/test/ContainerDetectorIntegration.test.ts#L49))

Test workflows run `npm run compile` for all packages

And because of that workflows run `npm run compile` for all packages before testing consuming time.

This PR leverages `nx` cache to avoid this extra compilation given the assumption that TypeScript is giving the same output regardless of the nodejs version used for compilation:
- TypeScript version is pinned and the same in all workflows
- TS configuration does not change depending on nodejs version

So we can change the workflows to
- have a single compilation step and storing the cache. Compilation is done with the same setup done for publishing (nodejs v16) so the build results are "production like".
- restore the cache when test jobs are running. This way the needed compilation takes less time

The gain is significant since the compilation takes around 5~7min and with the cache is around 3~6sec (2sec downloading + 2~3sec compiling).


.github/workflows/test-all-versions.yml
```diff
@@ -13,8 +13,35 @@ on:
...
+        with:
+          node-version: 16
+      - name: Install
+        run: npm ci
+      - name: Build
+        run: npm run compile
+      - name: Upload Build Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: tav-build-cache-${{ github.run_number }}
```
david-luna, line 38: note for reviewer: `github.run_number` is unique for the PR so each time the workflow is run this tasks overwrites the artifact saving space. `retention-days` is set to 1 so the artifact will be removed the next day after the last run of this workflow.

.github/workflows/test-all-versions.yml
```diff
@@ -13,8 +13,35 @@ on:
         type: string
 
 jobs:
+  build:
```
david-luna, line 16: note for reviewer: this is the build step that is going to be cached. `github.run_number` is unique for the PR so each time the workflow is run this tasks overwrites the artifact saving space. `retention-days` is set to 1 so the artifact will be removed the next day after the last run of this workflow.

.github/workflows/unit-test.yml
```diff
@@ -5,7 +5,34 @@ on:
   pull_request:
 
 jobs:
+  build:
```
david-luna, line 8: note for reviewer: same here for unit tests. I guess a single compile step can be done for both workflows but I think this will add complexity to them. This way both workflows are independent and still we reduce the resource usage

detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
```diff
@@ -10,7 +10,6 @@
     "compile": "tsc -p .",
     "lint": "eslint . --ext .ts",
     "lint:fix": "eslint . --ext .ts --fix",
-    "precompile": "tsc --version && lerna run version:update --scope @opentelemetry/resource-detector-alibaba-cloud --include-dependencies",
```
david-luna, line 13: note for reviewer: not necessary since `nx.json` has defined the task dependencies. Same applies for other packages

nx.json
```diff
@@ -0,0 +1,43 @@
...
+          "version:update"
+        ]
+      }
+    }
+  },
+  "targetDefaults": {
+    "compile": {
+      "dependsOn": [
+        "version:update",
+        "^compile"
```
david-luna, line 18: note for reviewer: this is the syntax to tell that dependencies within the monorepo should be compiled before compiling the current package

package.json
```diff
@@ -12,24 +12,21 @@
...
-    "clean": "lerna run clean",
-    "precompile": "tsc --version && npm run version:update",
-    "version:update": "lerna run version:update",
-    "compile": "lerna run compile",
-    "prewatch": "npm run precompile",
-    "test": "lerna run test",
-    "test:ci:changed": "lerna run test --since origin/main",
-    "test:browser": "lerna run test:browser --concurrency 1",
-    "test-all-versions": "npm run --if-present --workspaces test-all-versions",
-    "bump": "lerna publish",
```
david-luna, line 24: note for reviewer: I think this bump script is not used since release-please is handling the versioning

trentm:
I'll sponsor to not have the "package unmaintained" logic.

trentm:
> The gain is significant since the compilation takes around 5~7min and with the cache is around 3~6sec (2sec downloading + 2~3sec compiling).

I got confused by the mark up here. Are you trying to say the timing improved from `5-7 minutes` to `3-6 seconds`? If so, that's pretty nice. :)

david-luna:
> > The gain is significant since the compilation takes around 5~7min and with the cache is around 3~6sec (2sec downloading + 2~3sec compiling).
> 
> I got confused by the markup here. Are you trying to say the timing improved from `5-7 minutes` to `3-6 seconds`? If so, that's pretty nice. :)

Also I think I'm not clear enough. There is improvement in resource (CPU, Memory, ...) usage but not in the total time spent in the workflow. I'll try to elaborate

Currently each job on the matrix does a compilation (necessary to generate version file and have dependencies solved) before testing. Each one of them taking its 5-7 minutes of CPU & Memory consumption.

With the change we compile only once at the beginning, still taking 5-7min but only taking resources from 1 node. Then the job matrix for testing reuses the output from that previous step thanks to the cache. It might be represented like this

```
Before
test-matrix
├─ node 14/ --- compile (5-7min)---test(?min)---(reports and others)
├─ node 16/ --- compile (5-7min)---test(?min)---(reports and others)
├─ ...

Now
build --- compile with node v16 (5-7min)
├─ test-matrix
   ├─ node 14/ --- compile (3-6sec from cache)---test(?min)---(reports and others)
   ├─ node 16/ --- compile (3-6sec from cache)---test(?min)---(reports and others)
   ├─ ...
```

so total time is the same but using less resources.

trentm (approved):
- might want to add 'nx' to `ignoreDeps` in "renovate.json" for now

One thing that is lost here with `nx run-many ...` vs `lerna run ...` is the number of workers to run. When doing `npm run compile`, for example, with `lerna run ...` it would run 11 tasks in parallel on my machine. With `nx run-many ...` it runs 3 tasks in parallel. `lerna` is setting the "parallel / concurrency" based on the number of CPUs on the machine (https://github.com/lerna/lerna/blob/6.6.2/libs/core/src/lib/command/index.ts#L15). `nx` is hardcoding the default `--parallel` value to 3. I think even the latest nx version does that.

I was **going** to say that this means the initial `npm run compile` takes longer for me. However, I unscientifically compared `time npm run compile` (after running `npm ci`) in git clones with and without this PR, and even with the parallel=3 it is *faster* with nx. 

```
# with this PR
npm run compile  453.13s user 31.79s system 601% cpu 1:20.59 total

# without this PR (latest main)
npm run compile  655.85s user 86.89s system 880% cpu 1:24.35 total
```

I'll take it.
For me it means that the initial `npm run compile` takes longer.

.github/workflows/unit-test.yml
```diff
@@ -5,7 +5,34 @@ on:
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 16
+      - name: Install
+        run: npm ci
+      - name: Build
+        run: npm run compile
```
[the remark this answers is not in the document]
david-luna: > I'm a little scared of the "if npm ci and npm run compile layout is the same for all node versions". :)

We can do even another twist on this because:
- `ts-node` compiles `src` & `tests` folders so "in theory" is not necessary to compile, just to have the `version.ts` file available
- but some packages depend on another in the same monorepo like `@opentelemetry/resouce-detector-container` and these dependencies need to be build.

Since the published packages are compiled using nodejs v16 maybe is better to have a "production" build as tests assets instead of a different build with each nodejs version (14, 18, 20, ...)?

plugins/node/opentelemetry-instrumentation-bunyan/package.json
```diff
@@ -10,7 +10,6 @@
     "compile": "tsc -p .",
     "lint": "eslint . --ext .ts",
     "lint:fix": "eslint . --ext .ts --fix",
-    "precompile": "tsc --version && lerna run version:update --scope @opentelemetry/instrumentation-bunyan --include-dependencies",
```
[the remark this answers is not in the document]
david-luna: I've relied on the docs recommendation of running `npm run compile` at the root level after a clean checkout.

> As a general rule, installing and then compiling from the root directory should always be done first before anything else.

ref: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#general-guidance

.github/workflows/test-all-versions.yml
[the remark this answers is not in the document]
david-luna: fair point, I'll change it in a following commit

detectors/node/opentelemetry-resource-detector-container/package.json
```diff
@@ -34,6 +33,7 @@
   "devDependencies": {
     "@opentelemetry/api": "^1.0.0",
     "@opentelemetry/contrib-test-utils": "^0.42.0",
+    "@opentelemetry/instrumentation-fs": "*",
```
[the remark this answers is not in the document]
david-luna: > Should we give it a try?

I answer myself here with a NO. In this repo we have `@opentelemetry/auto-instrumentations-node` which has as dependencies most of the packages so I think for simplicity is better to compile all of them

trentm (approved):

package.json
pichlermarc: Yep that seems correct. I'd be surprised if this was used by anything. 
Thanks for removing it :+1:

.github/workflows/unit-test.yml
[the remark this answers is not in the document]
pichlermarc: > I'm not sure if we have binary deps -- i.e. ones that are built at install time via a gyp file or install or post-install npm scripts.

We don't have them anymore AFAICT. The last was removed with grpc-cencus-binary-propagator (#2276) because it always built on macOS (it had pre-built binaries for linux+glibc so building sometimes does not happen based on which system one is using). I think if there was another one left we'd have noticed it by now as it tends to take a lot of time on install. :slightly_smiling_face: 

  > Since the published packages are compiled using nodejs v16 maybe is better to have a "production" build as tests assets instead of a different build with each nodejs version (14, 18, 20, ...)?

I think having the Node.js version matched with the published packages makes sense from a testing standpoint. This way we can preempt surprises if there are any. @david-luna would you mind putting a comment in the publish workflow and here that the versions should always be updated together? :slightly_smiling_face:

.github/workflows/unit-test.yml
```diff
@@ -166,6 +199,11 @@ jobs:
         run: npm install -g npm@9 # npm@9 supports node >=14.17.0
       - name: Install
         run: npm ci
+      - name: Dowload Artifacts
```
pichlermarc, line 202: nit: typo
```suggestion
      - name: Download Artifacts
```

nx.json
```diff
@@ -0,0 +1,43 @@
...
+    "version:update": {
+      "inputs": [
+        "{projectRoot}/package.json"
+      ],
+      "outputs": [
+        "{projectRoot}/src/version.ts"
+      ]
+    }
+  }
+}
```
pichlermarc, line 43: nit: missing newline
```suggestion
}

```

pichlermarc (approved):
Looks good % some nits, thanks for working on this. :tada:  
Looking forward to seeing it in action :slightly_smiling_face:

Issue #1937, opened 2024-02-14
Socket.io instrumentation span name not following semantic convention


makeavish:
### What version of OpenTelemetry are you using?
@opentelemetry/auto-instrumentations-node@^0.39.4

### What version of Node are you using?
18

### What did you do?
Instrumented with latest socket.io otel instrumentation: https://www.npmjs.com/package/@opentelemetry/instrumentation-socket.io

### What did you expect to see?
According to otel spec guideline the span name should not contain a unique identifier and should be general: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.22.0/specification/trace/api.md#span

### What did you see instead?
I saw unique span names for spans: eg: `/[--3s4wJNxrT5DwjTABGg] send`

david-luna:
Hi @makeavish

Is it possible to have a little more context?
- is this happening for both sending and receiving messages? or just only one?
- could you share here the `package.json` of your app or at least the list of dependencies?
- could you share a minimal reproduction of the issue?

Thanks! :)

david-luna:
I've just tried with the chat example but could not reproduce. This is the repo if you want to add some code that may reproduce the issue
https://github.com/david-luna/socket-io-chat-example/tree/dluna/add-otel-instrumentation

What I can see in Kibana is a trace with the name `chat message receive`
![Screenshot 2024-04-26 at 13 46 29](https://github.com/open-telemetry/opentelemetry-js-contrib/assets/999029/910d0509-8bf7-42bb-b4ff-bb35e8ebbd6d)

prashanth-nani:
In our case, the Socket.io span name includes the room ID making every span name different as the room IDs are auto-generated GUIDs. This restricts us from using the span name for filtering/aggregations.

It clutters up the list of operations like this
![image](https://github.com/user-attachments/assets/2399c660-c96f-4c5c-964e-9bfb87d043f1)

The span names without dynamic data
![image](https://github.com/user-attachments/assets/7057af29-8ff0-4ce3-bf43-bd2bc85b7a03)


**Example Span in a trace:**
```
name: /[d31f44fd-11f3-43ba-a378-7b09df3bf900-cc762f38-2a17-4c51-887a-1a6f2ba1a53f] send
kind: producer
```
_Span Tags:_
```
messaging.destination: /
messaging.destination_kind: topic
messaging.socket.io.event_name: notification
messaging.socket.io.namespace: /
[...]

PR #2495, opened 2024-10-21
fix: change span names for socket-io

povilasv:

Socket io span names might contain high cardinality data. I'm removing roomId from span names so it doesn't.

Issue: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1937

- I refactored names so It follows messaging semconv,  it's "send ${namespace}" / "receive ${namespace}"  instead of `... receive`. 

One question about should it be "receive", I think we might need to change it to "process"?

Given:

> “Receive” spans SHOULD be created for operations of passing messages to the application when those operations are initiated by the application code (pull-based scenarios).

>“Process” spans SHOULD be created for operations of passing messages to the application when those operations are not initiated by the application code (push-based scenarios). Such “Process” span covers the duration of such an operation, which is usually a callback or handler.

Ref: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans

### DIFF
--- a/plugins/node/instrumentation-socket.io/src/socket.io.ts
+++ b/plugins/node/instrumentation-socket.io/src/socket.io.ts
@@ -292,14 +292,9 @@ export class SocketIoInstrumentation extends InstrumentationBase<SocketIoInstrum
         }
         const wrappedListener = function (this: any, ...args: any[]) {
           const eventName = ev;
-          const defaultNamespace = '/';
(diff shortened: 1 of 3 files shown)


blumamir (approved):
LGTM, the room should not be part of the span name and the operation should be first according to the span name of messaging spans.

Thanks for fixing it.

blumamir:
@povilasv please fix the lint error with `npm run lint` so CI is green

PR #2501, opened 2024-10-24
chore(propagator-aws-xray): update npm scripts to align with other pkgs in contrib repo

trentm:
Also drop the cross-var dep, now unused.


trentm:
A little story about `cross-var`, if you would like a little crazy in your life:


```bash
git clone git@github.com:open-telemetry/opentelemetry-js.git this-src-is-crazy
cd this-src-is-crazy
npm ci
npx cross-var echo hi
```

Does that fail with something like this?

```
/Users/trentm/tmp/this-src-is-crazy/node_modules/babel-core/lib/transformation/file/options/option-manager.js:328
        throw e;
        ^

ReferenceError: Unknown plugin "transform-runtime" specified in "/Users/trentm/tmp/this-src-is-crazy/node_modules/babel-plugin-transform-regenerator/node_modules/regenerator-transform/package.json" at 0, attempted to resolve relative to "/Users/trentm/tmp/this-src-is-crazy/node_modules/babel-plugin-transform-regenerator/node_modules/regenerator-transform" (While processing preset: "/Users/trentm/tmp/this-src-is-crazy/node_modules/babel-preset-es2015/lib/index.js")
    at /Users/trentm/tmp/this-src-is-crazy/node_modules/babel-core/lib/transformation/file/options/option-manager.js:180:17
...
```

We use `cross-var` in the core repo so that we can have the odd npm script that uses `$foo` for envvars, e.g. `cross-var echo this is my $SOMEVAR`, and it'll do the right thing for Windows (transform '$FOO' to '%FOO%' and spawn it).

Turns out the author wanted to use ESM code, but unfortunately was living in 2017 when this was written (before current versions of Node.js supported ESM).  "No problem," they said, "I'll use Babel to down compile to JS that the current node can run." At runtime. Everytime. "And I'll use the handy `babel-register` command to do it."

`babel-register` **installs a require hook** that attempts to transpile every file that is loaded.
"I don't want that," says the developer, "so I will use its `only` option to only transpile the files in my `src/` dir"

https://github.com/elijahmanor/cross-var/blob/e5aaa1d2de34f6b18a1251ad774c3ec41a219255/index.js#L3-L6

```js
require( "babel-register" )( {
    ignore: false,
    only: /src/
} );
```

Unfortunately that means that **any match for "src" in the full path is a hit** -- and "src" is in the top dirname I specified above. And babel blows up on ".../this-src-is-crazy/node_modules/babel-plugin-transform-regenerator/node_modules/regenerator-transform/lib/index.js" for reasons I didn't get in to.

(In my personal case, I sometimes work on a clone of opentelemetry-js.git and opentelemetry-js-contrib.git in my `~/src/` directory. Boom.)

If the above was changed to:

```js
require( "babel-register" )( {
    ignore: false,
    only: /cross-var\/src\/index.js$/
} );
```

it would be fine.
But cross-var hasn't been updated in 7y.
There is a PR for this: https://github.com/elijahmanor/cross-var/pull/18
And an npm-published fork with the languishing PR: https://www.npmjs.com/package/cross-var-src-patch

I would like to just remove cross-var from our repos.
This PR does so for the contrib repo, where cross-var usage snuck back in with the returning `propagator-aws-xray` package.

blumamir (approved):

dyladan (approved):

PR #2508, opened 2024-10-28
feat(instrumentation-aws-lambda): take care of ESM based (`.mjs`) handlers

serkan-ozal:

Currently, ESM based handlers with `.mjs` file extension are not taken care of by AWS Lambda instrumentation, but ESM based handlers are supported since long time (https://aws.amazon.com/tr/about-aws/whats-new/2022/01/aws-lambda-es-modules-top-level-await-node-js-14/).

For the fix, while handler file is being resolved, in addition to `.js` and `.cjs` file extensions, `.mjs` file extension is also checked.

This PR is the contrib counter part of the https://github.com/open-telemetry/opentelemetry-js/pull/5094

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
@@ -97,14 +97,28 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     // Lambda loads user function using an absolute path.
     let filename = path.resolve(taskRoot, moduleRoot, module);
     if (!filename.endsWith('.js')) {
-      // its impossible to know in advance if the user has a cjs or js file.
-      // check that the .js file exists otherwise fallback to next known possibility
+      // Its impossible to know in advance if the user has a js, cjs or mjs file.
+      // Check that the .js file exists otherwise fallback to the next known possibilities (.cjs, .mjs).
       try {
         fs.statSync(`${filename}.js`);
         filename += '.js';
       } catch (e) {
-        // fallback to .cjs
-        filename += '.cjs';
+        try {
+          fs.statSync(`${filename}.cjs`);
+          // fallback to .cjs (CommonJS)
+          filename += '.cjs';
+        } catch (e2) {
+          try {
+            fs.statSync(`${filename}.mjs`);
+            // fallback to .mjs (ESM)
+            filename += '.mjs';
+          } catch (e3) {
+            diag.error(
+              'No handler file was able to resolved with one of the known extensions for the file',
+              filename
+            );
+          }
+        }
       }
     }
 
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
@@ -97,23 +97,23 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     // Lambda loads user function using an absolute path.
     let filename = path.resolve(taskRoot, moduleRoot, module);
     if (!filename.endsWith('.js')) {
-      // Its impossible to know in advance if the user has a js, cjs or mjs file.
-      // Check that the .js file exists otherwise fallback to the next known possibilities (.cjs, .mjs).
+      // It's impossible to know in advance if the user has a js, mjs or cjs file.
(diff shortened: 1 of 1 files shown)


raphael-theriault-swi:
Looks like a subset of #2000, thanks for addressing the issue in the instrumentation code wrt. absolute paths !

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -97,14 +97,28 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     // Lambda loads user function using an absolute path.
     let filename = path.resolve(taskRoot, moduleRoot, module);
     if (!filename.endsWith('.js')) {
-      // its impossible to know in advance if the user has a cjs or js file.
-      // check that the .js file exists otherwise fallback to next known possibility
+      // Its impossible to know in advance if the user has a js, cjs or mjs file.
+      // Check that the .js file exists otherwise fallback to the next known possibilities (.cjs, .mjs).
```
trentm, line 101: Note that the Lambda Node.js runtime prefers this order: .js, .mjs, .cjs
(https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/v3.2.1/src/UserFunction.js#L149-L174)
I think this means that if a user has a Lambda with both a 'handler.mjs' *and* 'handler.cjs' that instrumentation won't work: OTel will prepare to instrument the .cjs file, but the Lambda runtime will load the .mjs file.

I think this is definitely an improvement, though, so if we wanted to get this in *now* for a release it is good.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -97,14 +97,28 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+          fs.statSync(`${filename}.cjs`);
+          // fallback to .cjs (CommonJS)
+          filename += '.cjs';
+        } catch (e2) {
+          try {
+            fs.statSync(`${filename}.mjs`);
+            // fallback to .mjs (ESM)
+            filename += '.mjs';
+          } catch (e3) {
+            diag.error(
```
trentm, line 116: Or this could be a diag.warn. I'm not sure if we have clear/clean policy on when to use which level.


Suggest use the instrunmentation's scoped `this._diag` so the insrtumentation name is included in the diag message.

```suggestion
            this._diag.error(
```

trentm (approved):

trentm:
@pichlermarc I haven't merged this yet, in case OP would like to respond to my comments above. I think it is good to merge if you are doing a release that could include this tomorrow.

trentm:
@raphael-theriault-swi Ah, sorry I hadn't seen your earlier PR ... and I hadn't really been digging into Lambda-related PRs. I'll try to take a look in the next few days.

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
serkan-ozal: Good catch

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -117,14 +117,28 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     // Lambda loads user function using an absolute path.
     let filename = path.resolve(taskRoot, moduleRoot, module);
     if (!filename.endsWith('.js')) {
-      // its impossible to know in advance if the user has a cjs or js file.
-      // check that the .js file exists otherwise fallback to next known possibility
+      // Its impossible to know in advance if the user has a js, cjs or mjs file.
```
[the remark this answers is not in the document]
serkan-ozal: done

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
serkan-ozal: done

plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
serkan-ozal: done

serkan-ozal:
@trentm Thanks for the review. I have updated PR according to your comments

trentm (approved):

PR #2517, opened 2024-11-05
docs(instrumentation-router): correct import statement

Netail:

- Router instrumentation import in the documentation


pichlermarc (approved):
thanks :+1:

PR #2525, opened 2024-11-13
test(instrumentation-mysql2): Add test for mysql2/promise

### LINKED ISSUE #650: Add support for mysql2/promise
When using library mysql2, instrumentation does not pickup calls made through mysql2/promise (https://www.npmjs.com/package/mysql2#using-promise-wrapper).

AFAIK, it does not look for promises require (https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts#L51), only 'mysql2'.

This is only for Node since databases are not normally used in browsers.

Weekly Downloads: ???
I would expect a large percentage of people using mysql2 would want promises. Over time. the usage of promises will continue to increase.

Versions: latest

Karibash:
close #650

Add test for mysql2/promise.

I added a test to confirm that `@opentelemetry/instrumentation-mysql2` works with `mysql2/promise`.


trentm:
I'll sponsor.

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql2-promise.test.ts
```diff
@@ -0,0 +1,980 @@
...
+    memoryExporter.reset();
+    instrumentation.setConfig({});
+    instrumentation.disable();
+    await connection.end();
+    await pool.end();
+    if (isPoolClusterEndIgnoreCallback()) {
+      await poolCluster.end();
+    } else {
+      await poolCluster.end();
+    }
```
trentm, line 165: I'm am guessing you could drop `function isPoolClusterEndIgnoreCallback()` below and simplify this block to:

```suggestion
    await poolCluster.end();
```

trentm (approved):
Thanks!
One nit for a possible simplification.

trentm:
@Karibash You'll need to run 'npm run lint:fix' to fix a style error in your added file.

trentm:
@Karibash Then some of the "TAV" checks are failing. TAV stands for "test-all-versions". The TAV tests effectively run `npm test` for a number of version of the `mysql2` dependency.

Looking at some of the TAV test runs above, it looks like:
- the tests passed with `mysql2@2.3.2`
- the tests failed with `mysql2@2.2.5`


Do you know what `mysql2` version added support for `mysql2/promises`? We may want to get the tests in `mysql2-promises.test.ts` to be skipped in older version of `mysql2`. I can help with that if you need.

Some mysql2 release dates:

```
  '2.2.5': '2020-09-21T12:28:13.157Z',
  '2.3.0': '2021-08-05T13:47:23.743Z',
  '2.3.1': '2021-10-15T11:42:14.888Z',
  '2.3.2': '2021-10-16T06:08:46.680Z',
  '2.3.3': '2021-11-14T04:17:46.192Z',
  '3.0.0': '2023-01-12T04:51:01.542Z',
```

Perhaps mysql2@2.3.0 added promises support, or changed how its promises support works, such that this instrumentation works with it?

You should be able to test locally by running:

```sh
cd .../plugins/node/opentelemetry-instrumentation-mysql2
npm install --no-save mysql2@2.2.5    # or some other version
RUN_MYSQL_TESTS=1 npm test
```

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql2-promise.test.ts
Karibash: Fixed in the following commit. In addition, the ESLint error is also fixed.
[ef5d5a6](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2525/commits/ef5d5a6d210aa7821c2ad9524d858840a90cf66b)

Karibash:
@trentm 
Thank you for your detailed explanation!
The reason the test was failing was that the Promise wrapper for PoolCluster did not exist in versions prior to v2.3.0.
In the following commit, I have modified the test so that the Promise tests for PoolCluster are only performed for v2.3.0 and later.
[8877f75](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2525/commits/8877f75ef6fe884ed8642af971db8fdee4b22d36)

david-luna:
@Karibash 

https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2662 already added some tests for `mysql2/promise` and conflicts with your PR. Do you think that's enough or do you wish to move this PR forward?

Karibash:
@david-luna 
If it has already been resolved, then it can't be helped. I'll close this PR. It's a bit disappointing that it was left unmerged for two months in the end :cry:

PR #2526, opened 2024-11-13
chore: remove lerna

david-luna:

This is a follow up PR of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2493. It removes `lerna` as a dependency and also from the CI workflows. All scripts and `lerna` commands have been replaced by its equivalent in `nx` (lerna was using them internally) or custom scripts.

## Changes

- remove unnecessary scripts in packages (precompile & prewatch)
- remove `lerna` dependency
- update `scripts/peer-api-check.mjs` to work at the root level and the dependant workflows
- add `publish-to-npm.mjs` script and update the release workflow
- added a `focus` npm script for people who does not want to compile all project to start developing
- added `scripts/bump-package-versions.mjs` to help with manual releases
- update docs


nx.json
```diff
@@ -18,17 +18,22 @@
         "^compile"
       ],
       "inputs": [
-        "{projectRoot}/src",
-        "{projectRoot}/test"
+        "{projectRoot}/src/**/*.ts",
```
david-luna, line 21: note to reviewer: although `nx` documentation says using folder names is okay I found this option to be more sensitive to changes.

.github/workflows/peer-api.yaml
```diff
@@ -19,4 +19,4 @@ jobs:
         run: npm ci
 
       - name: Check API dependency semantics
-        run: npx lerna exec "node \$LERNA_ROOT_PATH/scripts/peer-api-check.js"
+        run: node ./scripts/peer-api-check.js
```
david-luna, line 22: note for reviewer: this script was used only here so now it's changed to walk over all packages to do the API check.

.github/workflows/release-please-validate.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install lerna
-        run: npm install -g lerna@6.6.2
+      - name: Install glob
+        run: npm install glob
```
david-luna, line 19: note for reviewer: necessary dependency for the next step

RELEASING.md
```diff
@@ -90,19 +90,14 @@ Create a pull request titled `chore: x.y.z release proposal`. The commit body sh
 
 Merge the PR, and pull the changes locally (using the commands in the first step). Ensure that `chore: x.y.z release proposal` is the most recent commit.
 
-## Publish all packages
+## Compile all packages
```
david-luna, line 93: note to reviewer: IMO this step was missing since `npm publish` will pack and publish the packages in whatever state it is. With a full compilation everything should be in the right state for publishing

RELEASING.md
david-luna, line 1: note to self: although manual release is unlikely to happen we should provide a way to do it. So we need to provide an alternative to the command `lerna publish --skip-npm --no-git-tag-version --no-push`

Do not merge this PR until this is resolved.

david-luna: This is now implemented in `./scripts/bump-package-versions.mjs`

trentm:
I'll "sponsor" to stop the "unmaintained" workflows.

david-luna:
ping to @open-telemetry/javascript-maintainers since the changes in this PR affects the release process

david-luna:
Decided in the SIG to upgrade `lerna` instead of removing it. Closing in favour of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2836

PR #2533, opened 2024-11-16
feat(instrumentation-aws-lambda): In AWS Lambda instrumentation, extend invocation span start time back to Lambda environment startup time on coldstart

serkan-ozal:

Currently, AWS Lambda instrumentation doesn't take Lambda startup time into account while calculating invocation span start time on coldstarts. So, this behavior leads to reporting dramatically incorrect invocation span duration.

In this PR, Lambda environment start time is used as invocation span start time on coldstarts, so reported invocation span time also includes initialization time too.


dyladan:
@serkan-ozal this is quite an old draft. Is it something that can be safely closed?

PR #2536, opened 2024-11-18
ci: add workflow to update otel dependencies

pichlermarc:

We currently have to update `@opentelemetry/*` dependencies by manually opening a PR to do so (see #2535, for example). With the script that @trentm added (`scripts/update-otel-deps.js`) this process is almost fully automated already, and just has to be reviewed by a human. 

This PR adds a workflow that makes @opentelemetrybot open the PR based on the changes made by the script, similar to how [release PRs work in the core repo](https://github.com/open-telemetry/opentelemetry-js/blob/main/.github/workflows/create-or-update-release-pr.yml).

Ideally this workflow will help us spread the load of releasing even further by making it a bit simpler. Once merged and once we've tried running it - I'll update the core repo's releasing.md to include the instructions to run this workflow. We may even be able to have it run automatically when there's a new OTel JS core release. 🙂


.github/workflows/update-otel-deps.yaml
```diff
@@ -0,0 +1,48 @@
+name: Create or Update OpenTelemetry Update PR
+
+on:
+  workflow_dispatch:
```
trentm, line 4: - The first time this workflow fails I'll want to see the output with `DEBUG=1 ...` turned on. I guess we could run it locally to try to reproduce the failure. Eventually we may want a way to run the workflow with `DEBUG=1`.
- Currently update-otel-deps.js will `console.warn(...)` in some cases and carry on. Should we (I could) change the script (possibly optionally, or make it the default) to *fail* for those cases? One warning case is when the script notices that not all otel deps were updated (usually for some complex transitive dep issue).
- Would we *notice* if this update script failed? I don't regularly look at https://github.com/open-telemetry/opentelemetry-js-contrib/actions for failed workflow runs. Do you know if there an prior art in any OTel repos for notifications from GH Action workflows?

trentm (approved):

.github/workflows/update-otel-deps.yaml
pichlermarc: > The first time this workflow fails I'll want to see the output with DEBUG=1 ... turned on. I guess we could run it locally to try to reproduce the failure. Eventually we may want a way to run the workflow with DEBUG=1.

Yep, I think having an option to do so would be good - I'll follow up with a PR to add a toggle so that we can turn it on if needed.

> Currently update-otel-deps.js will console.warn(...) in some cases and carry on. Should we (I could) change the script (possibly optionally, or make it the default) to fail for those cases? One warning case is when the script notices that not all otel deps were updated (usually for some complex transitive dep issue).

I think we could do that - we have to carefully review the PR anyway for now so I'd definitely review it in-depth before merging

> Would we notice if this update script failed? I don't regularly look at https://github.com/open-telemetry/opentelemetry-js-contrib/actions for failed workflow runs. Do you know if there an prior art in any OTel repos for notifications from GH Action workflows?

For workflow runs that fail usually the person that triggered it gets a GitHub notification - are you looking for something more verbose like an automation to create issues, perhaps? 🤔

PR #2540, opened 2024-11-19
fix: losing this context on register handler function for hapi

Cheprer:

- When someone is using also a wrapper on the return function from register handler, due to wrapping by opentelemetry before it without retain the context, the following wrapper loses it.

- Use apply to retain the `this` context of the original function.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-hapi/src/instrumentation.ts
@@ -287,7 +287,7 @@ export class HapiInstrumentation extends InstrumentationBase {
           pluginName
         );
       });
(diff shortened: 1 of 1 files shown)


Cheprer:
Hi @pichlermarc can we move on with this. What is needed here please?

pichlermarc:
@Cheprer this is already fixed in the latest version AFAIK. #2625

pichlermarc:
> @pichlermarc nice, shame that someone else did the same thing again, where this could be merged and fixed much sooner.

Yes, you're right. I really try to do my best but I have to hande too many threads a day on GitHub and Slack - so on my end things tend to fall through the cracks especially on instrumentation packages like `@opentelemetry/instrumentation-hapi` that have been abandoned for a while.

PR #2549, opened 2024-11-25
chore(deps): bump @opentelemetry/semantic-conventions to 1.28

trivikr:

- https://github.com/open-telemetry/opentelemetry-js/releases/tag/semconv%2Fv1.28.0

- Updates `@opentelemetry/semantic-conventions` to 1.28

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-pg/package.json
+++ b/plugins/node/opentelemetry-instrumentation-pg/package.json
@@ -71,7 +71,7 @@
   "dependencies": {
     "@opentelemetry/core": "^1.26.0",
     "@opentelemetry/instrumentation": "^0.55.0",
-    "@opentelemetry/semantic-conventions": "1.27.0",
+    "@opentelemetry/semantic-conventions": "^1.28.0",
     "@opentelemetry/sql-common": "^0.40.1",
     "@types/pg": "8.6.1",
     "@types/pg-pool": "2.0.6"
--- a/plugins/node/opentelemetry-instrumentation-pg/package.json
+++ b/plugins/node/opentelemetry-instrumentation-pg/package.json
@@ -71,7 +71,7 @@
   "dependencies": {
     "@opentelemetry/core": "^1.26.0",
     "@opentelemetry/instrumentation": "^0.55.0",
-    "@opentelemetry/semantic-conventions": "^1.28.0",
+    "@opentelemetry/semantic-conventions": "1.28.0",
     "@opentelemetry/sql-common": "^0.40.1",
     "@types/pg": "8.6.1",
     "@types/pg-pool": "2.0.6"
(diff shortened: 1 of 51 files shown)


maryliag (approved):

plugins/node/opentelemetry-instrumentation-pg/package.json
```diff
@@ -71,7 +71,7 @@
   "dependencies": {
     "@opentelemetry/core": "^1.26.0",
     "@opentelemetry/instrumentation": "^0.55.0",
-    "@opentelemetry/semantic-conventions": "1.27.0",
+    "@opentelemetry/semantic-conventions": "^1.28.0",
```
trentm, line 74: This should be discussed.
Over in https://github.com/open-telemetry/opentelemetry-js/issues/5182 we are discussing the advice/recommendations/policy for using the "incubating" entry-point of the semconv JS package.

The rough concensus is that we will *pin* the semconv dep for a package when it is using the incubating entry-point, because in general there *can* be (hopefully rare) breaking changes in the incubating entry-point for a minor version release.

That's the reason this particular dep was pinned.

Currently the only packages in the contrib repo using the incubating endpoint are instrumentation-pg and intrumentation-aws-lambda. The latter isn't pinning the dep, mainly I assume because we are still working out how we want to handle the semconv dep.

trivikr: Pinned in [`9f2acaf` (#2549)](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2549/commits/9f2acaf8a0a5d0124e67d8d4ec81d633165d7ce2)

trentm:
@trivikr Now that all deps on the semantic-conventions package are unpinned, e.g. `"@opentelemetry/semantic-conventions": "^1.27.0"`, I'm not sure we *need* this kind of update.

- This regular maint PR a few weeks ago already updated the semconv version in the *package-lock* file for all packages: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2594

```bash
% npm ls --package-lock-only @opentelemetry/semantic-conventions -p
/Users/trentm/tm/opentelemetry-js-contrib2/node_modules/@opentelemetry/semantic-conventions

% head /Users/trentm/tm/opentelemetry-js-contrib2/node_modules/@opentelemetry/semantic-conventions/package.json
{
  "name": "@opentelemetry/semantic-conventions",
  "version": "1.28.0",
...
```

I think there is a balance here between:

1. Updating the package.json deps to always be at the latest minor semconv release, which pushes downstream users of the those packages to have to update their `@opentelemetry/semantic-conventions` dep; vs.
2. Leaving the semconv deps at a minimal required version for the given package. So, for example, if instrumentation-mysql2 needed a definition from `@opentelemetry/semantic-conventions@1.29.0`, then it could update that single package.json entry.   (The possible downside of this is that the author of that change might not notice they are using something new in v1.29.0 and forget to update the package.json entry.)

PR #2551, opened 2024-11-26
chore(aws): replace deprecated substr with substring

trivikr:

- The [String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated.

- This PR replaces `String.prototype.substr()` with [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring)


pichlermarc (approved):
Thanks 🙌

Issue #2543, opened 2024-11-19
Run prettier in the development workflow


trivikr:
### Is your feature request related to a problem? Please describe

Prettier was added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1439, but I couldn't find where it's run in the CI.
There don't seem to be IDE specific configurations or precommit hooks.

https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js-contrib+prettier+NOT+path%3A*.md&type=code

Because of this, there are some files committed without prettier being run
```console
$ npx prettier --write $(git ls-files '*.ts' '*.js')
...

$ git status --short | wc -l
     143
```

### Describe the solution you'd like to see

Run prettier somewhere in the development cycle. A good place to start is GitHub CI.

Maybe add `format` and `format:fix` in root package.json similar to `lint:markdown` and `lint:markdown:fix`
```json
    "format": "prettier $(git ls-files '*.ts' '*.js')",
    "format:fix": "prettier --write $(git ls-files '*.ts' '*.js')",
```

### Describe alternatives you've considered

All packages already have `lint` and `lint:fix` scripts. Another option will be add equivalent `format` and `format:fix` which run prettier.

trivikr:
I suggest formatting just the TypeScript files right now, as there are just 18 of them - and add CI check.

```console
$ npx prettier --check $(git ls-files '*.ts')
Checking formatting...
[warn] examples/express/src/client.ts
[warn] examples/express/src/server.ts
[warn] examples/express/src/tracer.ts
[warn] examples/koa/src/client.ts
[warn] examples/koa/src/server.ts
[warn] examples/koa/src/tracer.ts
[warn] examples/mongodb/src/client.ts
[warn] examples/mongodb/src/server.ts
[warn] examples/mongodb/src/tracer.ts
[warn] examples/mongodb/src/utils.ts
[warn] examples/mysql/src/client.ts
[warn] examples/mysql/src/server.ts
[warn] examples/mysql/src/tracer.ts
[warn] examples/redis/src/client.ts
[warn] examples/redis/src/express-tracer-handlers.ts
[warn] examples/redis/src/server.ts
[warn] examples/redis/src/setup-redis.ts
[warn] examples/redis/src/tracer.ts
[warn] Code style issues found in 18 files. Forgot to run Prettier?
```

The CI check will fail if files are not formatted for future PRs.
[...]

PR #2552, opened 2024-11-26
chore: add workflow to format using prettier

trivikr:

- Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2543
- Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2545#issuecomment-2501209745

- The first commit contains `format` command and a workflow. The CI fails as the TypeScript files are not formatted.
- The second PR contains formatting for the 18 files listed in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2543#issuecomment-2486997247. The CI succeeds.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,21 @@
+name: Format
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
+          cache: 'npm'
+      - run: npm ci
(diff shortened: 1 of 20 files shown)


trivikr:
cc @pichlermarc who'd enabled workflow for previous PR.

trentm:
I'd like to propose an alternative so that we can get the examples/* being linted using mostly the same eslitn config as the rest of the repo uses (using the top-level "eslint.config.js").  I have a start at supporting this here:
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2567

dyladan:
@trivikr there is already a lint workflow that is doing everything except the examples. I would prefer to extend that rather than adding a separate workflow that doesn't use our config. Trent already is working on this

PR #2553, opened 2024-11-26
chore: replace deprecated substr() with substring()

trivikr:

- The [String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated.

- This PR replaces `String.prototype.substr()` with [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring)
- Similar change for aws packages was done in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2551


trentm:
I'll sponsor to avoid the "unmaintained" pkg flows.

trentm (approved):
Thanks.

PR #2555, opened 2024-11-26
feat(instrumentation-mysql2): Add hook for setting span name

Just-Sieb:

My company runs a set clause to set a connection SQL variables to the user triggering the action before all our SQL clauses. This means the span name for requests always ends up being set which is not useful. Plus now that WITH statements are a more common SQL feature I am sure plenty of users could benefit from a way to customize the span name.

This adds a new config option the the mysql2 instrumentation which allows people to provide custom behavior for span name. I modeled the API after the similar API that already exists in the express instrumentation.


Just-Sieb:
@trentm Would you be willing to sponsor this PR? I saw you sponsored another one for this package.

Just-Sieb:
@raphael-theriault-swi Now that you are maintaining this package, would you be able to re-open this PR? I can also create a new one if that is necessary.

PR #2563, opened 2024-11-28
feat(pg): Patch client inside `lib` and `lib/pg-native`

onurtemizkan:

Ref:
-  https://github.com/getsentry/sentry-javascript/issues/14238
-  (Potentially related) https://github.com/getsentry/sentry-javascript/issues/14238#issuecomment-2477945642

Debugging the reproduction of https://github.com/getsentry/sentry-javascript/issues/14238, it seems `@opentelemetry/instrumentation-pg` fails to pick up the `pg` module while it successfully picks up and patches `pg-pool`.

Digging into `node_modules` of the reproduction, `pg` package did not have an index on its root, and the whole implementation resides in `lib`. Which made me suspect that relying on the root exports may not be enough with certain bundlers (Vite in this case) and runtimes. Instrumenting the client file instead worked well and spans were created successfully.

The same applies to `pg-native` which `pg` has bindings inside `lib/native`. So it's possible to patch it from the `pg` itself. That is similar to [the approach Sentry was using](https://github.com/getsentry/sentry-javascript/pull/3894) to instrument `pg-native` before migrating to OTEL. 

- Added file patchers for `lib/client.js` and `lib/native/client.js` on top of the original implementation.
- Tested both locally.


danielabel:
I'd very much like to support this PR getting merged. It solves a problem I am seeing where pg-native is not sending traces when the `native` flag is used in Sequelize

pichlermarc:
cc @maryliag (component owner) 🙂

plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
```diff
@@ -114,6 +114,10 @@ describe('pg-pool', () => {
   function create(config: PgInstrumentationConfig = {}) {
     instrumentation.setConfig(config);
     instrumentation.enable();
+
+    // Disable and enable the instrumentation to visit unwrap calls
+    instrumentation.disable();
```
maryliag, line 119: does this means that if a user adds a new instrumentation, they also need to disable and enabled?

onurtemizkan: No, I added this to visit the `unwrap` calls in the tests to ensure they don't error. They were not visited in tests in the coverage report.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -125,45 +132,38 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
...
+    const SUPPORTED_PG_VERSIONS = ['>=8.0.3 <9'];
 
-        if (isWrapped(moduleExports.Client.prototype.connect)) {
-          this._unwrap(moduleExports.Client.prototype, 'connect');
-        }
+    const modulePgNativeClient = new InstrumentationNodeModuleFile(
+      'pg/lib/native/client.js',
+      SUPPORTED_PG_VERSIONS,
+      this._patchPgClient.bind(this),
+      this._unpatchPgClient.bind(this)
```
maryliag, line 141: can you clarify why are you patching just to immediately unpatch? I'm not understanding the need of the unpatch part here

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -125,45 +132,38 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
...
-          module[Symbol.toStringTag] === 'Module'
-            ? module.default // ESM
-            : module; // CommonJS
-        if (isWrapped(moduleExports.Client.prototype.query)) {
-          this._unwrap(moduleExports.Client.prototype, 'query');
-        }
-      }
+        const moduleExports = extractModuleExports(module);
+
+        this._unpatchPgClient(moduleExports.Client);
```
maryliag, line 163: the original was only unpatching the `query`, with this just you're now unpatching both `query` and `connect`. Just want to confirm if that was intentional

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -125,45 +132,38 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
...
-        this._wrap(
-          moduleExports.Client.prototype,
-          'query',
-          this._getClientQueryPatch() as any
-        );
+    const modulePgClient = new InstrumentationNodeModuleFile(
+      'pg/lib/client.js',
+      SUPPORTED_PG_VERSIONS,
+      this._patchPgClient.bind(this),
+      this._unpatchPgClient.bind(this)
```
maryliag, line 148: same comment as the top one, why the need to unpatch?

maryliag (commented):
Hi, nice work working on this. I added a few questions :)

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
onurtemizkan: I moved to `InstrumentationNodeModuleFile` from `InstrumentationNodeModuleDefinition` as we need to patch the client file instead. Just passing the patch and unpatch methods with bound context. So `InstrumentationNodeModuleFile` will handle the unpatch when needed. They are passed as arguments.

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
onurtemizkan: Please see my comment above: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2563/files#r1918841550

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
onurtemizkan: I don't recall any specific reason why I added this unpatch, probably for consistency, as it's patched in `Client` as well. I can remove it.

maryliag: I think it was probably an oversight and both should have been unpatched, so your changed makes sense to me, but I wanted to confirm if it was intentional. You can keep it

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
maryliag: oh I'm so sorry, I read `;` instead of `,`, and it looked like you're making those calls... 🤦 
never mind, you're correct, sorry abut that

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
maryliag: same reply as above 😄

maryliag (approved):

FauxFaux:
FYI: This change appears to fix queries everywhere for us (node 22, esm); I'm guessing it fixes "esm in general".

mkohlmyr:
Is it possible that this issue or something very similar affects redis & express (and possibly mongoose) integrations? I've been observing a similar outcome where these do not appear to get patched and are not reporting to Sentry, which seems to believe we are running in CommonJS mode when in fact our esbuild outputs ESM and we are initialising instrumentation using --import.

Issue #2564, opened 2024-11-28
(instr-koa) Update tests for `@koa/router` v13


david-luna:
`@koa/router` was updated to v13 in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2384

Although all tests passed in all node versions supported (>=v14) the package declares in its `package.json` file that only supports Node.js >=18. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2384/files/d897d9b8bc3bceded82b6106248d3fa8b7575f56#r1857376767

This means that we're not testing the proper version ranges and in future an update of this package may break test.

Checklist
- [ ] skip tests when package is not compatible with node version
- [ ] update `tav.yml`

david-luna:
The min requirement of Node.js `>=20` was added in a commit between `12.0.2` and `13.0.0` ([see](https://github.com/koajs/router/commit/ff02349fe9e7fc7b94bf5b6426413db7233383ac)). Then it was changed to `>=18` in the PR mentioned before.

PR #2566, opened 2024-11-28
chore(instr-koa): update instrumentation-koa tests

david-luna:

Fixes: #2564 

- skip `@koa/router` tests if version is not compatible with nodejs version


plugins/node/opentelemetry-instrumentation-koa/test/koa.test.ts
```diff
@@ -721,7 +740,8 @@ describe('Koa Instrumentation', () => {
     });
   });
 
-  it('should work with ESM usage', async () => {
+  const itFn = isrouterCompat ? it : it.skip;
+  itFn('should work with ESM usage', async () => {
```
trentm, line 744: I'm curious: does `this.skip()` not work in async test functions?

trentm (approved):

PR #2567, opened 2024-11-28
chore(eslint): lint examples/bunyan and prepare for linting .js and .mjs files

trentm:
This adds support for 'npm run lint' in examples/bunyan, using the same
base eslint.config.js used by most of the repo.

There are ulterior motives:
- This is a first step to linting all the examples using the same
  base config (rather than the spotty linting of examples using the
  separate "examples/.eslintrc".
- The eslint.config.js changes support being used to lint .ts,
  .js, and .mjs files reasonably. I'll follow-up with a separate PR
  adding linting of .mjs files used for some ESM testing
  (e.g. "instrumentation-ioredis/test/fixtures/use-ioredis.mjs")
- I'll follow-up separately to get 'npm run lint' working for
  other examples and get the top-level `npm run lint:examples`
  using those.

Hattip to the following for some of the eslint config changes:
  https://github.com/open-telemetry/opentelemetry-js/pull/4446/files#diff-1a88e8e72db982c999acf146817a416a02d93547458e6525610cf69b24751af9

---

If it helps review:
- the first commit is just the config changes
- the second commit is the changes to examples/bunyan to get 'npm run lint' passing

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -5,10 +5,10 @@ module.exports = {
     "node",
     "prettier"
   ],
-  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended", "plugin:prettier/recommended"],
+  extends: ["eslint:recommended", "plugin:prettier/recommended"],
   parser: "@typescript-eslint/parser",
   parserOptions: {
-    "project": "./tsconfig.json"
+    "project": null
   },
   rules: {
     "quotes": ["error", "single", { "avoidEscape": true }],
@@ -28,6 +28,11 @@ module.exports = {
   overrides: [
     {
       files: ['*.ts'],
+      // Enable typescript-eslint for ts files.
+      extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended", "plugin:prettier/recommended"],
+      parserOptions: {
+        "project": "./tsconfig.json"
+      },
       rules: {
         "@typescript-eslint/no-floating-promises": "error",
         "@typescript-eslint/no-this-alias": "off",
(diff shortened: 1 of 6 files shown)


legendecas (commented):
Should `examples/bunyan` be added to the `package.json#workspaces` list in order to run linter on CI regularly?

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/8bfa21feda79cab00a42b6a1bb30f87eb7935841/package.json#L79-L86

trentm:
> Should `examples/bunyan` be added to the `package.json#workspaces` list in order to run linter on CI regularly?

I was initially thinking *no*. Instead I was going to consider adding a `lint:examples` script to the set of `npm run lint` things.   The last I recall we took the examples *out* of the workspace to help with shared package-lock deps, because the examples were so behind.

JamieDanielson:
> I was going to consider adding a `lint:examples` script to the set of `npm run lint` things.

I like this idea. It's super useful to have linters run against examples but shouldn't block any other builds. I would love to have proper linting on all the examples.

JamieDanielson (approved):

JamieDanielson:
> we took the examples _out_ of the workspace to help with shared package-lock deps, because the examples were so behind.

That's right. Similarly to my comment before, while it's useful to have CI run on examples regularly, if they are not being kept up to date, we don't want to have problems with the packages themselves. We need to revisit our examples strategy in general though to ensure they are useful while also not hindering forward progress.

dyladan:
This has a conflict but I think it can be merged after that

Issue #2572, opened 2024-12-02
[instrumentation-mysql2] tests time out for mysql2@3.11.5


pichlermarc:
## Description

On `main` the TAV tests for mysql2@3.11.5 are currently timing out. Since assertions are failing on that version, I'd assume that also telemetry is being dropped, hence categorizing this as `priority:p2`.

See release: https://github.com/sidorares/node-mysql2/releases/tag/v3.11.5
See failing workflow: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/12107327068/job/33754171979

pichlermarc:
I believe that this is related to https://github.com/sidorares/node-mysql2/pull/3081

pichlermarc:
This is a release blocker for core 1.29.0/0.56.0 - see https://github.com/open-telemetry/opentelemetry-js/pull/5232

PR #2573, opened 2024-12-02
feat(contrib-test-utils): copy soon-to-be-removed types from @opentelemetry/otlp-transformer

### LINKED ISSUE #2565: [contrib-test-utils] stop using soon-to-be-removed `@opentelemetry/otlp-transformer` types
## Description

This issue is part of https://github.com/open-telemetry/opentelemetry-js/issues/4582

In https://github.com/open-telemetry/opentelemetry-js/issues/4583 we will remove large parts of the public API of the `@opentelemetry/otlp-transformer` package. Before this change is released we should also get rid of all usages of these soon to be removed APIs from our tests here in this repository to ensure a smooth update-process to the latest `@opentelemetry/*` packages.

**This issue is considered done when:**
- [ ] usages of all types/functions that will be removed in https://github.com/open-telemetry/opentelemetry-js/issues/4583 have been replaced in this repository
  - see PR https://github.com/open-telemetry/opentelemetry-js/pull/5200 for a proposal of what will be removed

pichlermarc:

[We're about to remove the intermediate OTLP types from `@opentelemetry/otlp-transformer`](https://github.com/open-telemetry/opentelemetry-js/issues/4583) as we now have an interface that allows us to serialize directly to OTLP(protobuf) or OTLP(json).

Since they were used for testing here, we need to duplicate the relevant ones so that we're not blocked on the next core update.

Closes #2565 

- changes `export *` to explicit exports to avoid exporting anything we don't intend
- copies types from `@opentelemetry/otlp-transformer`
- moves the `OtlpSpanKind` enum from `test-fixtures.ts` to `otlp-types.ts`


pichlermarc:
Note: TAV tests will fail here due to #2572 - it's unrelated to this PR.

maryliag (approved):

packages/opentelemetry-test-utils/src/otlp-types.ts
```diff
@@ -0,0 +1,167 @@
...
+  message?: string;
+  /** Status code */
+  code: EStatusCode;
+}
+
+/** StatusCode enum. */
+const enum EStatusCode {
+  /** The default status. */
+  STATUS_CODE_UNSET = 0,
+  /** The Span has been evaluated by an Application developers or Operator to have completed successfully. */
```
JacksonWeber, line 129:
```suggestion
  /** The Span has been evaluated by an Application developer or Operator to have completed successfully. */
```

PR #2578, opened 2024-12-04
fix(instr-mysql2): Fix mysql2 instrumentation connection prototype

david-luna:

`mysql2@3.11.5` introduced a internal refactoring to solve circular dependencies within the package.
ref: https://github.com/sidorares/node-mysql2/pull/3081

The changes include a new `BaseConnection` class which contains the methods we need to patch (query, execute). The `Connection` class is extending it therefore the instrumentation gets the wrong prototype when patching.

Fixes: #2572 

- bump `mysql2` to v3.11.5
- added a function to extract the proper prototype when patching/unpatching

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
@@ -215,13 +215,12 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
   private _getConnectionPrototype(moduleExports: any): mysqlTypes.Connection {
     const baseClass = Object.getPrototypeOf(moduleExports.Connection);
 
-    // if `baseClass` has no prototype means connection is not extending
-    // another class so the functions we need to patch are already there
-    if (typeof baseClass.prototype === 'undefined') {
-      return moduleExports.Connection.prototype;
+    // return the base class prototype if it has the method
+    // we need to patch (mysql2@3.11.5+)
+    if (typeof baseClass.prototype?.query === 'function') {
+      return baseClass.prototype;
     }
 
-    // return the BaseConnection prototype (mysql2@3.11.5+)
-    return baseClass.prototype;
+    return moduleExports.Connection.prototype;
   }
 }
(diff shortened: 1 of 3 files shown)


pichlermarc (approved):
ah I also just finished fixing this on my working copy before heading out to lunch, looks like you beat my by half an hour 🙂 

Looks good, thanks 🙌 Let's wait for TAV to complete and then we should be goo to go 🙂

pichlermarc:
Huh, weird - tests in TAV seem to be still failing but it's not yet possible to see which versions they're failing at. 🤔

david-luna:
> ah I also just finished fixing this on my working copy before heading out to lunch, looks like you beat my by half an hour 🙂
> 
> Looks good, thanks 🙌 Let's wait for TAV to complete and then we should be good to go 🙂

Forgot to link the issue in the PR, sorry. TAV tests seem to be stuck so maybe my fix is not solving all issues :(

plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
```diff
@@ -210,4 +211,17 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
       };
     };
   }
+
+  private _getConnectionPrototype(moduleExports: any): mysqlTypes.Connection {
+    const baseClass = Object.getPrototypeOf(moduleExports.Connection);
+
+    // if `baseClass` has no prototype means connection is not extending
+    // another class so the functions we need to patch are already there
+    if (typeof baseClass.prototype === 'undefined') {
```
pichlermarc, line 220: Ah, I think in older versions you get `EventEmitter` as a base. That may be why the test are failing.

david-luna: I've reverted the check to make it more explicit. It will return the base class prototype if it has the `query` function (`execute` goes together). Otherwise it falls back to the `Connection` prototype.

Note: I had trouble running it locally since docker image `mysql:5.7` has no support for `linux/arm64/v8` platform. Is it possible to update the image version in `@opentelemetry/contrib-test-utils`?

david-luna:
closing this one in favor of #2579

PR #2588, opened 2024-12-05
fix(test-utils): Don't swallow assertion errors from `checkResult` and `checkCollector`

onurtemizkan:

- Assertion errors thrown inside `runTextFixture` are swallowed and not reported.

- Reject the `runTestFixture` promise with the thrown error, if there is one.
- Updated Hapi tests that seem to be silently failing before this update.


pichlermarc (approved):
looks good, thanks 👍

PR #2589, opened 2024-12-06
feat(detector-aws): add more lambda semconv attributes

garysassano:

Adding extra attributes: `aws.log.group.names`, `faas.instance`, `faas.max_memory`

See above.


detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
```diff
@@ -19,47 +19,32 @@ import {
...
-      return Resource.empty();
-    }
-
     const functionVersion = process.env.AWS_LAMBDA_FUNCTION_VERSION;
-    const region = process.env.AWS_REGION;
+    const logGroupName = process.env.AWS_LAMBDA_LOG_GROUP_NAME;
+    const logStreamName = process.env.AWS_LAMBDA_LOG_STREAM_NAME;
+    const memorySize = process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE as string;
 
     const attributes: ResourceAttributes = {
```
jj22ee, line 38: Any reason you are preferring to hardcode the attribute keys and not use the key values from `@opentelemetry/semantic-conventions`?

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
```diff
@@ -19,47 +19,32 @@ import {
...
-  CLOUDPLATFORMVALUES_AWS_LAMBDA,
-} from '@opentelemetry/semantic-conventions';
 
 /**
  * The AwsLambdaDetector can be used to detect if a process is running in AWS Lambda
  * and return a {@link Resource} populated with data about the environment.
  * Returns an empty Resource if detection fails.
  */
 export class AwsLambdaDetectorSync implements DetectorSync {
-  detect(_config?: ResourceDetectionConfig): IResource {
```
jj22ee, line 40: Removal of this config parameter is a breaking change ([see other usage](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/resource-detector-aws-v1.9.0/detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetector.ts#L32)). While not used here, it could be useful later, so let's keep it.

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
garysassano: Actually, some of them never existed in `@opentelemetry/semantic-conventions` because the module couldn't keep up with the docs update.

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
```diff
@@ -19,47 +19,32 @@ import {
...
+  detect(): IResource {
+    const awsRegion = process.env.AWS_REGION;
     const functionName = process.env.AWS_LAMBDA_FUNCTION_NAME;
-    if (!functionName) {
```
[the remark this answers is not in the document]
garysassano: Why would someone import the `AwsLambdaDetector` if they're not running in a Lambda environment? It doesn't make sense. Developers should be mindful of what they're including in their projects. Importing random modules without intent or purpose isn’t standard practice, is it?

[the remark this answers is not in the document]
jj22ee: There is an `all` option setting in `OTEL_NODE_RESOURCE_DETECTORS`, uses all resource detectors.

- https://github.com/open-telemetry/opentelemetry-js-contrib/blob/auto-instrumentations-node-v0.54.0/metapackages/auto-instrumentations-node/src/utils.ts#L272-L277

- https://github.com/open-telemetry/opentelemetry-js/blob/v1.29.0/experimental/packages/opentelemetry-sdk-node/src/utils.ts#L55-L60

So the current implementation will cause unexpected behaviour.

Also please see [this doc for zero-code auto-instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-instrumentations-node/README.md):
> By default, all SDK resource detectors are used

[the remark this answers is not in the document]
jj22ee: The specification does touch upon this topic (possibility of restrictions of the default "all" setting) a bit further.
See [Trade-offs and mitigations](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0111-auto-resource-detection.md#trade-offs-and-mitigations).

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
```diff
@@ -19,47 +19,32 @@ import {
...
+    const logStreamName = process.env.AWS_LAMBDA_LOG_STREAM_NAME;
+    const memorySize = process.env.AWS_LAMBDA_FUNCTION_MEMORY_SIZE as string;
 
     const attributes: ResourceAttributes = {
-      [SEMRESATTRS_CLOUD_PROVIDER]: String(CLOUDPROVIDERVALUES_AWS),
-      [SEMRESATTRS_CLOUD_PLATFORM]: String(CLOUDPLATFORMVALUES_AWS_LAMBDA),
+      'aws.log.group.names': [logGroupName],
+      'cloud.provider': 'aws',
+      'cloud.platform': 'aws_lambda',
+      'cloud.region': awsRegion,
```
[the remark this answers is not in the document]
garysassano: In this context, it's unnecessary to check for undefined runtime environment variables. Since you're already ensuring the code runs within a Lambda environment, these variables are guaranteed to be present. Adding extra checks for them would undermine the purpose of AWS's standardization across Lambda runtimes, as detailed in the [documentation](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime).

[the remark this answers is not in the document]
garysassano: In this context, it's unnecessary to check for undefined runtime environment variables. Since you're already ensuring the code runs within a Lambda environment, these variables are guaranteed to be present. Adding extra checks for them would undermine the purpose of AWS's standardization across Lambda runtimes, as detailed in the [docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime).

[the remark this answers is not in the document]
jj22ee: Hey, sorry for the late response. The latest commit looks good, but to followup on this conversation, from your link it states that for SnapStart Lambda functions, `AWS_LAMBDA_LOG_GROUP_NAME` and `AWS_LAMBDA_LOG_STREAM_NAME` are not available. Can these 2 env vars be conditionally added only if they are available?

*Note that SnapStart isn't available (yet?) for Node.js - https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html. So this suggestion is just for future-proofing.

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
jj22ee: Similar to your change in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2642, can you use these attribute keys from `@opentelemetry/semantic-conventions/incubating`?

garysassano: @jj22ee Could you approve the other PR so I can merge those changes? That way, I can limit the scope of this one to just adding the extra attributes.

garysassano: @jj22ee Could you approve the other PR so I can merge those changes? That way, I can limit the scope of this PR to just adding the extra attributes.

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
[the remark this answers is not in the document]
garysassano: I embedded a JSDoc link to the official AWS Lambda envvars docs, added a condition for Lambda SnapStart, replaced hardcoded strings with constants from the semantic convention library, and reorganized the code in both the detector and the test to enhance readability.

pichlermarc:
cc @jj22ee, please take another look if you find the time 🙂

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
```diff
@@ -38,27 +41,37 @@ import {
  */
 export class AwsLambdaDetectorSync implements DetectorSync {
   detect(_config?: ResourceDetectionConfig): IResource {
-    const functionName = process.env.AWS_LAMBDA_FUNCTION_NAME;
-    if (!functionName) {
+    // Check if running inside AWS Lambda environment
+    const executionEnv = process.env.AWS_EXECUTION_ENV;
+    if (!executionEnv?.startsWith('AWS_Lambda_')) {
```
jj22ee, line 46: I just recently learned that the Amazon Linux 2 & 2023 Lambda runtimes (OS-only runtime used for Golang, Rust, C++, custom) does not define the `AWS_EXECUTION_ENV` environment variable. Let's use the older logic that checks for `AWS_LAMBDA_FUNCTION_NAME` instead.

garysassano: > this change will make this resource detector stop working for Go/C++/Rust Lambda users.

This detector is specifically designed for the Lambda Node.js runtime. Since we're in the `opentelemetry-js-contrib` repository, which is written in TypeScript, support for other languages isn't relevant.

Regarding the concern that this change affects Go/C++/Rust Lambda users, that’s not the case. This detector cannot be used with those languages—they have separate implementations. In fact, I recently added an attribute to the Rust AWS detector ([see here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/139)) as well. The implementations are entirely distinct, and `opentelemetry-js-contrib` is not applicable to other languages.

garysassano: > this change will make this resource detector stop working for Go/C++/Rust Lambda users.

This detector is specifically designed for the Lambda Node.js runtime. Since we're in the `opentelemetry-js-contrib` repository, which is written in TypeScript, support for other languages isn't relevant.

Regarding the concern that this change affects Go/C++/Rust Lambda users, that’s not the case. This detector cannot be used with those languages—they have separate implementations. In fact, I recently added an attribute to the Lambda resource detector in Rust as well ([see here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/139)). The implementations are entirely distinct, and `opentelemetry-js-contrib` is not applicable to other languages.

garysassano: > this change will make this resource detector stop working for Go/C++/Rust Lambda users.

This detector is specifically designed for the Lambda Node.js runtime. Since we're in the `opentelemetry-js-contrib` repository, which is written in TypeScript, support for other languages isn't relevant.

Regarding the concern that this change affects Go/C++/Rust Lambda users, that’s not the case. This detector cannot be used with those languages, as each OpenTelemetry SDK has its own separate implementation. In fact, I recently added an attribute to the Lambda resource detector in Rust as well ([see here](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/139)). The implementations are entirely distinct, and `opentelemetry-js-contrib` is not applicable to other languages.

jj22ee: Oh sorry, was a bit absent-minded late last night. Thanks!

jj22ee (approved):

detectors/node/opentelemetry-resource-detector-aws/src/detectors/AwsLambdaDetectorSync.ts
garysassano: No problem, appreciate the late-night review.

garysassano: No problem, I appreciate the late-night review.

pichlermarc (approved):

PR #2599, opened 2024-12-10
chore(instrumentation-pg): Inline incubating constants from @opentelemetry/semantic-conventions

nwalters512:

~~No other instrumentation packages pin `@opentelemetry/semantic-conventions`, and from what I can tell, there's no good reason that `@opentelemetry/instumentation-pg` should be pinning it. When the dependency is pinned, npm can't dedupe it to a higher version, which results in multiple versions of this package being installed.~~

As decided in https://github.com/open-telemetry/opentelemetry-js/issues/5182#issuecomment-2536558790, `incubating` constants from `@opentelemetry/semantic-conventions` should be inlined instead of pinning `@opentelemetry/semantic-conventions` to a particular version.

~~In `@opentelemetry/instrumentation-pg`, `@opentelemetry/semantic-conventions` now uses a `^` semver range to allow newer minor versions to be installed.~~

This PR updates `@opentelemetry/instrumentation-pg` to include inlined copies of the experimental constants it needs. This, in turn, allows us to unpin `@opentelemetry/semantic-conventions`.


maryliag:
The reason for this versioning being pinned, it's because this package is using the `incubating` entrypoint, which may break on minor bumps.
This was added on purpose here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2473
Once it's no longer using the `incubating`, it can be unpinned.

I see other packages that use the `incubating` entrypoint are not pinning the value, like this [one](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json#L60).  @pichlermarc do you know if this was just a mistake or we can now unpin for these cases?

nwalters512:
Ah! Apologies, I didn't go digging deep enough. In that case, can I just change this PR to pin to 1.28.0 so that it will be deduped by package managers (at least until 1.29.0 comes out)?

pichlermarc:
> I see other packages that use the incubating entrypoint are not pinning the value, like this [one](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-lambda/package.json#L60). @pichlermarc do you know if this was just a mistake or we can now unpin for these cases?

Hi 🙂 I'd say it's a mistake, when using the incubating entrypoint, we should always pin for now until we have a decision on https://github.com/open-telemetry/opentelemetry-js/issues/5182. Sorry for the inconvenience.

maryliag:
Hi @nwalters512 , looks like we will go with a different approach of copying those specific values from the incubation, see more here: https://github.com/open-telemetry/opentelemetry-js/issues/5182#issuecomment-2536558790

I would be happy to review if that is something you want to work on.

nwalters512:
@maryliag I'd be happy to open some PRs for this! I'll work on `instrumentation-pg` to start, and I can follow up with other packages that are using the `incubating` entrypoint.

maryliag:
thank you @nwalters512!

plugins/node/opentelemetry-instrumentation-pg/src/instrumentation.ts
```diff
@@ -65,7 +65,7 @@ import {
   METRIC_DB_CLIENT_OPERATION_DURATION,
   ATTR_DB_NAMESPACE,
   ATTR_DB_OPERATION_NAME,
-} from '@opentelemetry/semantic-conventions/incubating';
+} from './semantic-conventions';
```
nwalters512, line 68: https://github.com/open-telemetry/opentelemetry-js/issues/5182#issuecomment-2536558790 suggested using `semconv.ts`; I went with `semantic-conventions.ts` since it directly matches the package name. Happy to use the other suggestion if that's strongly preferred.

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * These constants are considered experimental exports of `@opentelemetry/semantic-conventions`.
+ * They're being inlined until they're officially exported by `@opentelemetry/semantic-conventions`.
+ */
```
nwalters512, line 4: This little comment will appear at the top of this file in other packages; let's workshop it if needed before it proliferates.

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * The number of connections that are currently in state described by the `state` attribute
+ *
+ * @experimental This metric is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
```
nwalters512, line 9: I copied these all in full, including the "experimental" warning. Happy to edit this to remove it desired.

nwalters512 (commented):
@maryliag I repurposed this PR to include the inlining of the relevant constants.

nwalters512:
Sorry about the failing tests, I'll get those fixed up.

maryliag:

@trentm would you mind confirming if that is what you had in mind?

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
trentm: FWIW, in a third-party instrumentation I have at work that is doing this, I have this block comment:
https://github.com/elastic/elastic-otel-node/blob/690896cb2995a349313b37ae6e609a741676946c/packages/instrumentation-openai/src/semconv.ts#L20-L27

That's perhaps a little wordier than necessary, though.

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
trentm: I think either way is fine. 

(Ideally eventually there is some handy tooling for this to help with generation and maintenance of these. I'd played with a *start* at something, but it is far from ready for others to use:
https://github.com/elastic/elastic-otel-node/blob/main/packages/instrumentation-openai/scripts/semconv-gen.js)

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
trentm, line 1: nit: I have a small preference for "semconv.ts" if others think that is reasonable as the filename. (It is the name I mentioned as an example at https://github.com/open-telemetry/opentelemetry-js/blob/main/semantic-conventions/README.md#unstable-semconv)

I'm happy to be overridden by others' opinions on this. The current naming of `.ts` files in this repo *does* tend to use this-style.ts rather than using abbreviations in filenames.

trentm (approved):
> @trentm would you mind confirming if that is what you had in mind?

Yes, this is what I had in mind. Thanks!

I added a couple comments and a nit. Good to go as is, however.

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
nwalters512: I commented on this here too: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2599#discussion_r1880770481

I'll defer to whatever you or the other maintainers think is best 🙂

plugins/node/opentelemetry-instrumentation-pg/src/semantic-conventions.ts
nwalters512: Agreed that it's perhaps a bit wordy. I'd be happy to apply any recommendations to the message I have here!

trentm:
@nwalters512 https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2664 is doing the same thing. There was discussion on Slack to refine that one. I'm merging that one and closing this one. Thanks for the changes!

PR #2612, opened 2024-12-20
feat(oracledb): Add support for Oracle DB instrumentation

sudarshan12s:
Add Oracledb instrumentation

Adding instrumentation support for [node-oracledb ](https://github.com/oracle/node-oracledb)applications connecting to Oracle database. 

From [node-oracledb ](https://github.com/oracle/node-oracledb) version 6.7 onwards,  oracledb module exports a class, 
`TraceHandlerBase` which simulates interface kind of functionality with abstract methods to be overwritten by the derived class.

The current OT module derives this class `TraceHandlerBase` and implements abstract methods; It uses `setTraceInstance` method of `oracledb` module to register this new derived class `OracleTelemetryTraceHandler`
These methods are invoked by `oracledb` module providing the  `traceContext` which is used to dump the data in to the span after converting to standard span attributes.

Following are the abstract methods implemented:

- onEnterFn -> It gets invoked by `oracledb` module when application invokes `getConnection` to create a connection to 
database or `execute` method to run a sql. It is invoked before invoking the public API calls made by application.
- onExitFn -> It gets invoked after the public API call is completed with success/failure.
- onBeginRoundTrip -> It gets invoked before round trip to database which was triggered as part of public API.
- onEndRoundTrip -> It gets invoked after round trip to database completes which was triggered as part of public API..


raphael-theriault-swi (changes requested):
Could you rename the new types to TitleCase to follow the rest of the code ?

sudarshan12s:
> Could you rename the new types to TitleCase to follow the rest of the code ?

Yes renamed to TitleCase for types and internal-types exported.

sudarshan12s:
The earlier triggered CI failure seems to be this. It looks related to some package version dependency ,` package-lock.json` updates and not the changes in PR.   Please suggest if there are some changes needed from my side.  
```

 Exception during run: /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/@aws-sdk/client-sqs/dist-cjs/index.js:185
  static {
         ^

SyntaxError: Unexpected token '{'
```

sudarshan12s:
Please help me with steps for updating` package-lock.json` . This PR lock file  ` package-lock.json` differs from main and i am not able to keep it in sync easily..

sudarshan12s:
> Please help me with steps for updating` package-lock.json` . This PR lock file ` package-lock.json` differs from main and i am not able to keep it in sync easily..

I have fixed this discrepency.

JamieDanielson:
@sudarshan12s as we discussed during SIG meeting yesterday, please review the [requirements here for component ownership](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#component-ownership), as well as our [preferred approach for long-term maintenance of instrumentation libraries](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#new-instrumentation).

marcalff:
> @sudarshan12s as we discussed during SIG meeting yesterday, please review the [requirements here for component ownership](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#component-ownership), as well as our [preferred approach for long-term maintenance of instrumentation libraries](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#new-instrumentation).

To @JamieDanielson 
Cc @open-telemetry/javascript-maintainers

Hi Jamie, All,

To become a member of CNCF, @sudarshan12s will need two sponsors for his application.

I am myself an Oracle employee, one of the [maintainers](https://github.com/open-telemetry/opentelemetry-cpp?tab=readme-ov-file#contributing) of the opentelemetry-cpp repository, and can sponsor Sudarshan.

Could you, or someone in javascript-maintainers, be the second sponsor ?
I think it makes the most sense, since Sudarshan contribution(s) are in the js-contrib space.

Regards,
-- Marc

sudarshan12s:
> Hi @sudarshan12s. I was absent during last week's SIG meeting so I'm not sure if there was an agreement made to not apply the current guidelines, but there are rules and guidelines for new instrumentations. These were introduced since many instrumentations were abandoned in this repository and we would like to avoid this in the future as it diverts resources from development of OTel core components.
> 
> **Before opening a PR,** the following questions should be answered, see [Jamie's link](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/CONTRIBUTING.md#new-instrumentation):
> 
> * who's going to maintain it (2 people minimum, both must be members of the OTel org prior to opening the PR)
> * why not host it in a separate repo? (hosting it here has maintenance implications for the repo, the less we have to host here, the better the quality of other components like the API and SDK is going to be as we get to spend more time on that)
> 
> This is done in the form of an [instrumentation request](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/new?template=instrumentation_request.md).

Thanks @pichlermarc , Yes we were informed on the guidelines/procedure to be followed in the last SIG. I have updated the instrumentation request [here](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2703) with more details. Let me know if this looks fine. For becoming members of OTel org, @marcalff  has agreed for first sponsor. Can someone from javascript-maintainers be the second sponsor.

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
```diff
@@ -0,0 +1,57 @@
...
+  ORACLE_EXEC_OPTIONS = 'db.oracle.exec.options',
+  ORACLE_INSTANCE = 'db.oracle.instance',
+  ORACLE_PDBNAME = 'db.oracle.pdbname',
+  ORACLE_IMPLICIT_RELEASE = 'db.oracle.implicit_release',
+}
+
+// Contains span names produced by instrumentation
+// It lists the RPC names (suffix with _MSG like EXECUTE_MSG) and
+// exported oracledb functions (like EXECUTE).
+export enum SpanNames {
```
blumamir, line 34: please see the guidelines for span names [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#name) and review if we can improve the span name. 
I understand not all the info is available.

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
```diff
@@ -0,0 +1,325 @@
...
+  SEMATTRS_NET_TRANSPORT,
+  SEMATTRS_DB_USER,
+  SEMATTRS_DB_STATEMENT,
+} from '@opentelemetry/semantic-conventions';
+import {
+  ATTR_DB_SYSTEM,
+  ATTR_DB_NAMESPACE,
+  ATTR_DB_OPERATION_NAME,
+} from './semconv';
+import * as oracledbTypes from 'oracledb';
```
blumamir, line 41:
```suggestion
import type * as oracledbTypes from 'oracledb';
```

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
```diff
@@ -0,0 +1,325 @@
...
+/**
+ * OracleTelemetryTraceHandler extends TraceHandlerBase from oracledb module
+ * It implements the abstract methods; `onEnterFn`, `onExitFn`, `onBeginRoundTrip`
+ * and `onEndRoundTrip` of TraceHandlerBase class.
+ * Inside these overriden methods, the input traceContext data is used
+ * to generate attributes for span.
+ */
+export class OracleTelemetryTraceHandler extends newmoduleExports.traceHandler
+  .TraceHandlerBase {
+  private _tracer: Tracer;
```
blumamir, line 60: the tracer should be updated when `setTracerProvider` is called on the instrumentation class.

plugins/node/opentelemetry-instrumentation-oracledb/src/internal-types.ts
```diff
@@ -0,0 +1,46 @@
...
+// Captures the entire span data
+export interface TraceSpanData {
+  operation: string; // RPC or exported function name.
+  error?: oracledbTypes.DBError;
+  connectLevelConfig?: SpanConnectionConfig;
+  callLevelConfig?: SpanCallLevelConfig;
+  additionalConfig?: any; // custom key/values associated with a function.
+  fn: Function; // Replaced with bind function associating the active context.
+  args?: any[]; // input arguments passed to the exported function.
+  userContext: InstrumentationContext;
```
blumamir, line 38: please add a short comment about this field, and how the driver uses it, so it's documented for future maintainers

plugins/node/opentelemetry-instrumentation-oracledb/src/internal-types.ts
```diff
@@ -0,0 +1,46 @@
...
+import type * as api from '@opentelemetry/api';
+import { SpanConnectionConfig } from './types';
+
+// onEnterFn returns this Context(contains only span for now) and it is
+// received in onExitFn to end the span.
+export interface InstrumentationContext {
+  span: api.Span;
+}
+
+// Captures the entire span data
```
blumamir, line 29: you can add a short comment documenting that this interface correlates to a js object that is exported from the instrumented package. preferably with a link

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
```diff
@@ -0,0 +1,57 @@
...
+  ORACLE_POOL_MIN = 'db.oracle.pool_min',
+  ORACLE_POOL_MAX = 'db.oracle.pool_max',
+  ORACLE_POOL_INCR = 'db.oracle.pool_incr',
+  ORACLE_EXEC_OPTIONS = 'db.oracle.exec.options',
+  ORACLE_INSTANCE = 'db.oracle.instance',
+  ORACLE_PDBNAME = 'db.oracle.pdbname',
+  ORACLE_IMPLICIT_RELEASE = 'db.oracle.implicit_release',
+}
+
+// Contains span names produced by instrumentation
```
blumamir, line 31: These span names actually correlates to the values being produced by the instrumented package itself, let's add it as a comment so people will know it's not arbitrary and will break if changed

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
```diff
@@ -0,0 +1,325 @@
...
+    }
+  }
+
+  // This method is invoked after exported function from oracledb module
+  // completes.
+  onExitFn(traceContext: TraceSpanData) {
+    if (
+      this._shouldSkipInstrumentation() ||
+      !traceContext.userContext ||
+      !traceContext.userContext.span
```
blumamir, line 275:
```suggestion
      !traceContext.userContext?.span
```

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
```diff
@@ -0,0 +1,325 @@
...
+      );
+    }
+    if (traceContext.additionalConfig.implicitRelease) {
+      span.setAttribute(
+        AttributeNames.ORACLE_IMPLICIT_RELEASE,
+        traceContext.additionalConfig.implicitRelease
+      );
+    }
+    if (traceContext.error) {
+      span.setStatus({
```
blumamir, line 223: consider:

```
  span.recordException(traceContext.error);
```

which will log a span event for this error

plugins/node/opentelemetry-instrumentation-oracledb/src/internal-types.ts
sudarshan12s: Added following:
```

// This corresponds to js object filled by the 'oracledb' module
// See: https://github.com/oracle/node-oracledb/blob/main/lib/traceHandler.js
```

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
sudarshan12s: I am passing the function (`() => this.tracer` ) to OracleTelemetryTraceHandler constructor instead. Just to avoid override of `setTracerProvider` . looks fine?

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
sudarshan12s: The type script [definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/oracledb) still does not have new exported class [`TraceHandlerBase`](https://github.com/oracle/node-oracledb/blob/main/lib/traceHandler.js)  , Hence importing both types and values. 


Should I keep it for now and remove as types definitions get updated?

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright (c) 2024, Oracle and/or its affiliates.
+ * */
+
+// Oracle specific attributes not covered by semantic conventions
```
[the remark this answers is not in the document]
sudarshan12s: I am working on the [PR](https://github.com/open-telemetry/semantic-conventions/pull/1911) for oracledb semconv.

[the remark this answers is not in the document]
sudarshan12s: Link is also added in the code for the same.

blumamir:
> For becoming members of OTel org, @marcalff has agreed for first sponsor. Can someone from javascript-maintainers be the second sponsor.

Once this PR is finalized and ready to merge, I can be your second sponsor for the organization membership

plugins/node/opentelemetry-instrumentation-oracledb/src/types.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export interface OracleInstrumentationExecutionResponseHook {
+  (span: api.Span, resultInfo: OracleResponseHookInformation): void;
+}
+
+export interface OracleInstrumentationConfig extends InstrumentationConfig {
+  /**
+   * If true, an attribute containing the execute method
+   * bind values will be attached the spans generated.
```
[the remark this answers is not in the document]
sudarshan12s: changed as following:
```
If true, an attribute containing the execute method
   bind values will be attached the spans generated.
  It can potentially record PII data and should be used with caution.
```

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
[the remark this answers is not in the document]
sudarshan12s: Renamed to `OracleTelemetryTraceHandler.ts`

plugins/node/opentelemetry-instrumentation-oracledb/src/types.ts
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @default false
+   */
+  enhancedDatabaseReporting?: boolean;
+
+  /**
+   * If true, db.statement will have sql in the spans generated.
```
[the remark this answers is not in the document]
sudarshan12s: Added the following:
```

If true, db.statement will have sql which could potentially contain
    sensitive unparameterized data in the spans generated.
```

plugins/node/opentelemetry-instrumentation-oracledb/src/instrumentation.ts
```diff
@@ -0,0 +1,68 @@
...
+  constructor(config: OracleInstrumentationConfig = {}) {
+    super(PACKAGE_NAME, PACKAGE_VERSION, config);
+  }
+
+  protected init() {
+    const moduleOracleDB = new InstrumentationNodeModuleDefinition(
+      'oracledb',
+      ['>= 6.7 < 7'],
+      (moduleExports: typeof oracleDBTypes) => {
+        const newmoduleExports: any = moduleExports;
```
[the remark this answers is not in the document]
sudarshan12s: Yes removed `newmoduleExports` variable.

plugins/node/opentelemetry-instrumentation-oracledb/src/instrumentation.ts
```diff
@@ -0,0 +1,68 @@
...
+        const config = this.getConfig();
+        const obj = new utils.OracleTelemetryTraceHandler(
+          () => this.tracer,
+          config
+        );
+        obj.enable();
+
+        // Register the instance with oracledb.
+        newmoduleExports.traceHandler.setTraceInstance(obj);
+        this._tmHandler = obj;
```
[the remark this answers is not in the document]
sudarshan12s: I am using non null check to see if its patched already, then unregister and register again. Please add if i missed anything.

`if (this._tmHandler) {`

plugins/node/opentelemetry-instrumentation-oracledb/src/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  serverMode?: string;
+  pdbName?: string;
+  poolMin?: number;
+  poolMax?: number;
+  poolIncrement?: number;
+}
+
+export interface OracleRequestHookInformation {
+  inputArgs: any;
+  connection: SpanConnectionConfig | undefined;
```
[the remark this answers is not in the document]
sudarshan12s: removed `undefined`.

plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -0,0 +1,439 @@
...
+// to extend the TraceHandlerBase class. This is needed
+// to avoid importing values from oracledb and only import types.
+interface TraceHandlerBaseConstructor {
+  new (): traceHandler['TraceHandlerBase'];
+}
+
+import type { traceHandler } from 'oracledb'; // Import only for type checking
+
+// It returns the TraceHandlerBase class, if oracledb module is available.
+function getTraceHandlerBaseClass(): TraceHandlerBaseConstructor | null {
```
sudarshan12s, line 120: For the issue with requiring oracledb to be present in node_modules with this line:
`import * as oracledbTypes from 'oracledb';`

I made changes to import types instead:
`import type { traceHandler } from 'oracledb';`

Now since `oracledb` module exposes `TraceHandlerBase` as abstract class where some default implementations (`enable`) are available, I can not do `implements TraceHandlerBase`.
Added a function, getTraceHandlerBaseClass which gives the constructor for TraceHandlerBase dynamically during patch avoiding global `require("oracledb")` in generated js files. 

The types added inside this file will be removed once this [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/72060) and release including this change is done.

Please confirm if i am missed anything or any suggestions.


Thanks for pointing it out with details.

sudarshan12s:
> Did another round of review and added few more comments.
> 
> Currently marking as "RequestChanges" due to the following 2 issues:
> 
> 1. the copyright to oracle should be addressed (not sure in what way) @open-telemetry/javascript-maintainers
> 2. the instrumentation must not import from the instrumented package.
> 
> Thanks again

I have addressed the second point by including types , Please see.

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
sudarshan12s: The TraceHandlerBase is modelled as abstract class, hence I need to extend and not implement. I made changes to use dynamic require with a function, Please see and let me know if this is fine

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
```diff
@@ -0,0 +1,330 @@
...
+      );
+      if (
+        this._instrumentConfig.dbStatementDump ||
+        this._instrumentConfig.enhancedDatabaseReporting
+      ) {
+        span.setAttribute(SEMATTRS_DB_STATEMENT, callConfig.statement);
+        if (this._instrumentConfig.enhancedDatabaseReporting && !roundTrip) {
+          const values = this._getValues(callConfig.values);
+          if (values) {
+            span.setAttribute(AttributeNames.ORACLE_BIND_VALUES, values);
```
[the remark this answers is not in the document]
sudarshan12s: I will try. But since it was more for debugging and its a sensitive information, I think its less priority.

plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
raphael-theriault-swi: Note that the instance of `oracledb` that's being `require`d by the instrumentation package here might differ from the one that is being patched. Given you have access to the exports from the patch method it would be a better idea to pass them here as a parameter and get the base class from the instance that's being instrumented.

sudarshan12s: got it. I made the changes accordingly.

dyladan:
https://github.com/cncf/foundation/blob/main/copyright-notices.md#what-if-i-want-my-copyright-notice-included

plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -0,0 +1,372 @@
...
+    diag.error('Failed to load oracledb module.', err);
+    return null;
+  }
+}
+
+export function getOracleTelemetryTraceHandlerClass(
+  obj: typeof oracleDBTypes
+): any {
+  const traceHandlerBase = getTraceHandlerBaseClass(obj);
+  if (traceHandlerBase) {
```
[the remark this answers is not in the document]
sudarshan12s: Done.

sudarshan12s:
Hey @opentelemetry/javascript-approvers, I've incorporated the feedback received so far. If anyone else would like to take a look, I'd appreciate additional inputs!

Thanks.

plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
raphael-theriault-swi: Thank you !

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright (c) 2025, Oracle and/or its affiliates.
+ * */
+
+// Oracle specific attributes not covered by standard semantic conventions.
+// See: https://github.com/open-telemetry/semantic-conventions/pull/1911
+export enum AttributeNames {
```
maryliag, line 21: I don't think this PR should be merged until the [Semantic Conventions PR ](https://github.com/open-telemetry/semantic-conventions/pull/1911) for it is merged, since the definition can change there and you don't want to have something merged that is already out of date even before it gets released. And on that topic, why these attributes were not added to the semantic convention? 
I imagine the goal is to have a similar instrumentation for oracle in the other SDKs, but how will Java or .NET know what names to use if they're not defined in the Semantic convention. They will have no idea that it was already implemented in JS and they should follow here. This is why the Semantic Conventions are so important. So please add a complete definition there with all attributes you're using in this instrumentation 😄 

Also, as I mentioned in the SIG, really try to use already existing name attributes, for example, we have several [attributes for pool](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#connection-pools) already, so there is no need to create specific with oracle in the name. 
You should check if we don't have for the others.

sudarshan12s: Thanks, @maryliag! I've started making the changes based on the inputs suggested in the semantic conventions PR.

I’ll use the existing attributes and retain only one entry in `constants.ts`, as it is specific to Oracle [DRCP](https://node-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html#implicitpool):
`ORACLE_IMPLICIT_RELEASE = 'oracle.db.implicit_release',` 

For pool attributes, I plan to map them as follows:
```
oracledb.pool.max -> db.client.connection.max  
oracledb.pool.min -> db.client.connection.min  
```
I also requested clarification on whether `db.client.connection.max` and `db.client.connection.min` can be used in DB spans when the `oracledb.createPool` API is called, in addition to being included in metrics. They will not be dumped in subsequent DB queries.

For now, I’ve removed `oracledb.pool_incr` (I have not seen in existing attributes) but can add it back to the semantic conventions and instrumentation once I receive clarification on the above.

sudarshan12s: I have incorporated the changes based on the semantic conventions PR feedback received so far. The attribute `oracle.db.implicit_release` is still under review.

plugins/node/opentelemetry-instrumentation-oracledb/src/utils.ts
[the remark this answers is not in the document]
sudarshan12s: I have added it now along with other feedback from semantic conventions PR.

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
sudarshan12s: The semantic conventions PR is merged now. I made changes to incorporate them. Please see. 

Following is the example of span attributes aligning to conventions.

If a sql query, `'select :name from dual'` is executed with name being the placeholder , following attributes would appear.


```
spanName-> oracledb.Connection.execute:SELECT FREE|FREEPDB1|freepdb1 (It has Fnname:operationName db.namespace)
db.operation.name ='SELECT'
db.operation.parameter.name ='0'
db.query.text = 'select :name from dual'
db.system.name ='oracle.db'
db.namespace = 'FREE|FREEPDB1|freepdb1'. (It has instanceName | pdbname | serviceName)
```

maryliag (commented):
Thank you for your work on this!
I added a few comments

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
```diff
@@ -0,0 +1,47 @@
...
+  EXECUTE_MANY = 'oracledb.Connection.executeMany',
+  LOGOFF_MSG = 'oracledb.LogOffMessage',
+  CONNECT_CLOSE = 'oracledb.Connection.close',
+  CREATE_LOB = 'oracledb.Connection.createLob',
+  LOB_MESSAGE = 'oracledb.LobOpMessage',
+  LOB_GETDATA = 'oracledb.Lob.getData',
+}
+
+/*
+ * The semantic conventions defined DBSYSTEMVALUES_ORACLE as oracle, hence
```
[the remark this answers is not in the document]
sudarshan12s: This was sourced from [here](https://github.com/open-telemetry/semantic-conventions/blob/dd6a536db4a2c7d351f0cae72bd253f2955da25b/model/database/registry.yaml#L301) and also reflects a suggestion to follow that format. I think the intent is to avoid including vendor or company names, and instead have the engine or database.

plugins/node/opentelemetry-instrumentation-oracledb/src/types.ts
```diff
@@ -0,0 +1,94 @@
...
+   */
+  enhancedDatabaseReporting?: boolean;
+
+  /**
+   * If true, db.statement will have sql which could potentially contain
+   * sensitive unparameterized data in the spans generated.
+   *
+   * @default false
+   */
+  dbStatementDump?: boolean;
```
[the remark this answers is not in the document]
sudarshan12s: I have updated documentation with more details. `dbStatementDump` only dumps sql stmt but never bind values. But `enhancedDatabaseReporting` dumps sql stmt and bind values.. This was done to enable basic debugging of sql stmts .. Looks fine?

plugins/node/opentelemetry-instrumentation-oracledb/test/oracle.test.ts
```diff
@@ -0,0 +1,1680 @@
...
+  ReadableSpan,
+  TimedEvent,
+} from '@opentelemetry/sdk-trace-base';
+import * as assert from 'assert';
+import { OracleInstrumentation } from '../src';
+import { SpanNames, DB_SYSTEM_VALUE_ORACLE } from '../src/constants';
+
+import {
+  ATTR_SERVER_ADDRESS,
+  SEMATTRS_NET_TRANSPORT,
```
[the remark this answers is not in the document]
sudarshan12s: Yes replaced with ATTR_*

plugins/node/opentelemetry-instrumentation-oracledb/test/oracle.test.ts
```diff
@@ -0,0 +1,1680 @@
...
+}
+
+function getDBNameSpace(instanceName = '', pdbName = '', servicename = '') {
+  return [instanceName, pdbName, servicename].join('|');
+}
+
+// It verifies the spans, its attributes and the parent child relationship.
+function verifySpans(
+  parentSpan: Span | null,
+  attributesList: Attributes[] = [executeAttributes, executeAttributes],
```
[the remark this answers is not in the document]
sudarshan12s: For executing a query, typically we have one span for public API and a child span doing a round trip. 

Both the spans have same attributes dumped, hence we are passing same attributes for both public API and round trip span..

plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -0,0 +1,434 @@
...
+            span,
+            traceContext.callLevelConfig,
+            roundTrip
+          );
+        }
+        if (traceContext.error) {
+          span.recordException(traceContext.error);
+          span.setStatus({
+            code: SpanStatusCode.ERROR,
+            message: traceContext.error.message,
```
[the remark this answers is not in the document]
sudarshan12s: Thanks. I shall confirm on this.

plugins/node/opentelemetry-instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -0,0 +1,434 @@
...
+        if (value == null) {
+          return 'null';
+        }
+        if (value instanceof Buffer || this._isLobInstance(value)) {
+          return value.toString();
+        }
+        if (typeof value === 'object') {
+          return JSON.stringify(value);
+        }
+        // number, string, boolean,
```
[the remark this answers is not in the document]
sudarshan12s: Donw

plugins/node/opentelemetry-instrumentation-oracledb/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+- RoundTrip Spans will not appear for Thick Mode
+- Hostname will not be available in Thick Mode
+
+### Oracle Instrumentation Options
+
+| Options | Type | Default | Description |
+| ------- | ---- | --------| ----------- |
+| `enhancedDatabaseReporting` | `boolean` | `false` | If true, additional details about the sql statement's bind values will be attached to the spans generated to represent the database operations |
+| `dbStatementDump` | `boolean` | `false` | If true, ``db.statement`` will contain the sql string in the spans generated |
```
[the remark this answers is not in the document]
sudarshan12s: Corrected.

sudarshan12s:
> Thank you for your work on this! I added a few comments

I have incorporated the comments..  Please see.

plugins/node/opentelemetry-instrumentation-oracledb/src/constants.ts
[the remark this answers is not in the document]
maryliag: got it, thanks

plugins/node/opentelemetry-instrumentation-oracledb/src/types.ts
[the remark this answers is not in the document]
maryliag: thanks for the explanation!

plugins/node/opentelemetry-instrumentation-oracledb/src/semconv.ts
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_DB_OPERATION_PARAMETER = 'db.operation.parameter';
+
+/**
+ * The connection string used to connect to the database. It is recommended to remove embedded credentials.
+ *
+ */
+export const ATTR_DB_CONNECTION_STRING = 'db.connection_string';
```
maryliag, line 101: this attribute is deprecated
``@deprecated Replaced by `server.address` and `server.port`.``

plugins/node/opentelemetry-instrumentation-oracledb/test/oracle.test.ts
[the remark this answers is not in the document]
maryliag: I'm still seeing a lot of `SEMATTRS_` here

maryliag (commented):
Thank you for addressing all comments. It's almost there :) 
I added a few more comments, all easy to fix!

plugins/node/opentelemetry-instrumentation-oracledb/test/oracle.test.ts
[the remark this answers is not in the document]
sudarshan12s: I had missed the test files earlier. added now—thanks.

plugins/node/opentelemetry-instrumentation-oracledb/src/semconv.ts
sudarshan12s: Can I keep this attribute. 

The reason is : 

The [node-oracledb](https://github.com/oracle/node-oracledb) driver supports a thick mode, in which the address and port are not directly available. While it's technically possible to extract them by parsing the connection string, this approach is avoided. Moreover, if the connection string is an [alias](https://node-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html#tns-aliases-for-connection-strings) referencing an external configuration file (such as a tnsnames.ora entry), the address and port may not be accessible at all . It is planned to expose these attributes in thick mode from node-oracledb driver itself but currently they are not available.

This limitation does not apply to thin mode in [node-oracledb](https://github.com/oracle/node-oracledb), as it provides the parsed address, port, and protocol directly.

sudarshan12s: I have removed the `ATTR_DB_CONNECTION_STRING` attribute as suggested in the SIG.

plugins/node/opentelemetry-instrumentation-oracledb/test/oracle.test.ts
```diff
@@ -0,0 +1,1675 @@
...
+const pno = 1521;
+const serviceName = 'FREEPDB1';
+let serverVersion = 2304000000; // DB version.
+let numExecSpans = 2; // Default number of Spans created for Execute API in thin mode.
+let numConnSpans = 2; // Default number of spans created during connection establishment.
+let poolMinSpanCount = 1; // number of spans created for createPool considering poolMin.
+const CONFIG = {
+  user: process.env.ORACLE_USER || 'demo',
+  password: process.env.ORACLE_PASSWORD || 'demo',
+  connectString: process.env.ORACLE_CONNECTSTRING || 'localhost:1521/freepdb1',
```
maryliag, line 81:
```suggestion
```

sudarshan12s: > 

This is a configuration attribute that contains information such as `host`, `port`, and `servicename`. It's required to establish a database connection and run the tests. However, it is never validated within the span, as it does not appear there.

maryliag (approved):
Great work on this!

sharadraju:
Thanks @maryliag and @blumamir  for your review and thanks @marcalff for the sponsor support!
@sudarshan12s Nice work!

sudarshan12s:
> Great work on this!

Thank you so much...

sudarshan12s:
It looks like the failed test may be flaky (intermittently asserting different connection counts). Would it be possible to either ignore this failure for this PR, or re-run the CI?

```
Expected values to be strictly equal:
2 !== 1
at test/mongodb-v4-v5-v6.metrics.test.ts:127:12
```

Thanks for your time!`

sudarshan12s:
> It looks like the failed test may be flaky (intermittently asserting different connection counts). Would it be possible to either ignore this failure for this PR, or re-run the CI?
> 
> ```
> 
> MongoDBInstrumentation-Metrics
>        Should add connection usage metrics:
> 
>       AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
> 
> 2 !== 1
> 
>       + expected - actual
> 
>       -2
>       +1
>       
>       at Context.<anonymous> (test/mongodb-v4-v5-v6.metrics.test.ts:127:12)
> ```
> 
> Thanks for your time!

@onurtemizkan, I believe this is a flaky error and can be safely ignored.

sharadraju:
Thanks @blumamir for merging this PR.

Issue #2627, opened 2025-01-07
[aws-sdk] test-all-versions script is failing


trivikr:
### What version of OpenTelemetry are you using?

CI failure

### What version of Node are you using?

CI failure in Node.js 

### What did you do?

CI failed for https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2623#issuecomment-2575947974, due to dependency `@aws-sdk/client-sqs`

### What did you expect to see?

CI to run deprecation of AWS SDK for JavaScript v2

### What did you see instead?

CI failure because of `@aws-sdk/client-sqs` dependency, possibly unrelated to PR

trivikr:
Probably related to the SDK?

https://github.com/aws/aws-sdk-js-v3/issues/6779

trivikr:
This issue is specific to Node.js 14.x

The error in Node.js 16.x is different https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/12656564146/job/35269729978?pr=2623, and needs to be handled in the PR.

```console
  1) MessageAttributes
       injectPropagationContext
         should inject context if there are available attributes:
     Error: expect(received).toBeGreaterThan(expected)

Expected: > 5
Received:   5
      at Context.<anonymous> (test/MessageAttributes.test.ts:84:53)
      at processImmediate (node:internal/timers:491:21)
      at process.topLevelDomainCallback (node:domain:161:15)
      at process.callbackTrampoline (node:internal/async_hooks:128:24)
```

PR #2628, opened 2025-01-07
test(aws-sdk): set v3 versions based on Node.js support

trivikr:

- Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2627

- Sets AWS SDK for JavaScript v3 versions based on Node.js version they support.


pichlermarc (approved):
Thank you for tracking that down, much appreciated 🙌

PR #2631, opened 2025-01-08
fix(instrumentation-runtime-node): use the correct unit for eventloop.utilization

### LINKED ISSUE #2630: nodejs.eventloop.utilization under instrumentation-runtime-node is a percentage, not a duration
0.11.0

20.17.0

### What did you do?

I was trying to understand what `event loop utilization` actually meant and i ended up reading the code [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-runtime-node/src/metrics/eventLoopUtilizationCollector.ts) and looking up the [nodejs docs](https://nodejs.org/api/perf_hooks.html#performanceeventlooputilizationutilization1-utilization2). 

The metric unit to be a percentage or number

### What did you see instead?
The metric unit is seconds

### Additional context
[on line 48](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-runtime-node/src/metrics/eventLoopUtilizationCollector.ts#L48) `elu.utilization` _according to the nodejs docs_ returns a value between 0 and 1. This represents the percentage of time spend in an active state. Example 0.5 would mean half of the time spent in the event loop is active and half idle.
What I don't understand is how does OTEL convert this value between 0 and 1 into a duration?
I was reading [this comment thread](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2136#discussion_r1674534637) on the PR that introduced this code and maybe otel is summing the 0->1 values or something? not sure.
Something smells like a bug to me 🤔  

Anyway, I would love to have someone explain to me how this metric works when nodejs is returning a number/percentage but otel reports a duration in seconds

maryliag:

- Fix #2630

- Unit of the metric `eventloop.utilization` was wrongly using `s`, instead of the unit `1`, as defined its [semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/main/model/nodejs/metrics.yaml#L72-L81)


pichlermarc (approved):
looks good, thank you 🙂

Issue #2629, opened 2025-01-07
[aws-sdk] drop support for Node.js 14.x


trivikr:
### Is your feature request related to a problem? Please describe

The AWS SDK for JavaScript v3 announced dropping support of Node.js 14.x in [Oct 2023](https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-node-js-14-x-in-the-aws-sdk-for-javascript-v3/), and dropped it in on May 1, 2024 in https://github.com/aws/aws-sdk-js-v3/pull/6034

The instrumentation still supports Node.js 14.x which has been end-of-life since [April 2023](https://github.com/nodejs/Release/pull/859).

### Describe the solution you'd like to see

End support for Node.js 14.x in instrumentation-aws-sdk

### Describe alternatives you've considered

Keep supporting Node.js 14.x in instrumentation-aws-sdk

### Additional context

- Noticed in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2627
- Can be taken up after JS SDK v2 support is dropped in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2607, and a new minor version is released.
- Once this change is done and published, another issue can be created to drop support for Node.js 16.x

trivikr:
@open-telemetry/javascript-approvers Can an instrumentation drop support for specific Node.js version in configuration, i.e. [engines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) in package.json, while others keep supporting it?

pichlermarc:
> Can an instrumentation drop support for specific Node.js version in configuration, i.e. [engines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) in package.json, while others keep supporting it?

@trivikr unfortunately no (for now) since it is included in the `@opentelemetry/auto-instrumentations-node` metapackage - when dropping support for an included instrumentation we'd also have to drop support for Node.js 14 for the metapackage.

However, we're planning to release 2.x of the core repo in February https://github.com/open-telemetry/opentelemetry-js/issues/5148. 
Once the we publish 2.0 in core, then we can also drop support Node.js 14 in the metapackage and the AWS SDK instrumentation. 🙂

pichlermarc:
SDK 2.0 will also drop support for Node.js 16 so that will also be dropped here.

PR #2632, opened 2025-01-08
test(aws-sdk): drop support for Node.js 14.x [DO NOT MERGE]

trivikr:

- DO NOT MERGE
- Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2629#issuecomment-2577993128

- This PR is just to check if CI fails, if we try to drop support for Node.js 14.x in one of the `@opentelemetry/auto-instrumentations-node` dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37371,7 +37371,7 @@
         "typescript": "4.4.4"
       },
       "engines": {
-        "node": ">=14"
+        "node": ">=16"
       },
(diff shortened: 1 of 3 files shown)


trivikr:
The CI failures are due to some network connectivity. Converting to CI so that it's not merged.
I'll rebase when new commits are pushed to main, and recheck.

PR #2633, opened 2025-01-08
fix(instrumentation-fastify): add missing module export

### LINKED ISSUE #2027: Fastify's _patchConstructor causing errorCodes to be inaccessible
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/9df30ea0fd822a69203b818b8fbe34e1e1c8bced/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts#L203-L222

This overrides the originally exported `fastify` while it should contain `errorCodes` as [assigned here](https://github.com/fastify/fastify/blob/main/fastify.js#L896). This breaks whenever I try to do this:
```ts
import { errorCodes } from 'fastify';
```
The `errorCodes` is always `undefined`.

Is there any workaround for this?

tsmithhisler:

This PR adds an additional export that is part of the fastify public API to the `InstrumentationNodeModuleDefinition` patch function. Without this patch, 

fastify v4.8+ exports an object named `errorCodes` as both a property of the default export and as a named export.

The export is documented at https://github.com/fastify/fastify/blob/4.x/docs/Reference/Errors.md?plain=1#L236.

This closes #2027.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-fastify/src/instrumentation.ts
@@ -27,6 +27,7 @@ import type {
   FastifyInstance,
   FastifyRequest,
   FastifyReply,
+  FastifyErrorCodes,
 } from 'fastify';
 import { hooksNamesToWrap } from './constants';
 import {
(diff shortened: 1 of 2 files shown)


pichlermarc (dismissed):
thanks 👍

pichlermarc:
oh this still needs some lint fixes (`npm run lint:fix`)

tsmithhisler:
It turns out the `errorCodes` export was added in v4.8.0. I updated the test and code to support that. Also the lint fixes should be in as well.

tsmithhisler:
FYI I rebased off of latest main and updated commit body to reflect v4.8 instead of general v4.x.

tsmithhisler:
@pichlermarc anything I can do to help move this along?

pichlermarc (approved):

PR #2636, opened 2025-01-10
feat(detectors): generate esm build files too

serkan-ozal:

As OTEL FAAS SIG, we are working on reducing OTEL Lambda Nodejs layer coldstart overhead and bundling has the biggest impact. But bundling tools like webpack, esbuild don't like CJS modules while tree-shaking, but ESM modules. So, in this PR, we are configuring build process to generate ESM build files too.

To be able to generate ESM build files too, this PR is 
- introducing `tsconfig.esm.json` files
- and updating scripts accordingly in the `package.json`
- and defining ESM file paths in the `files` section of the `package.json` for their discoverability

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
+++ b/detectors/node/opentelemetry-resource-detector-alibaba-cloud/package.json
@@ -3,18 +3,19 @@
   "version": "0.29.7",
   "description": "OpenTelemetry resource detector for Alibaba Cloud",
   "main": "build/src/index.js",
+  "module": "build/esm/index.js",
   "types": "build/src/index.d.ts",
   "repository": "open-telemetry/opentelemetry-js-contrib",
   "scripts": {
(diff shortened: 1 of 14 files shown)


dyladan:
Looks like this is done the same way as our other ESM libraries like the aws propagator etc. You should know that this method of ESM publishing is actually not standards compliant and we've had problems with it in the past. There is an issue in the main repo to revamp our ESM publishing strategy.

For now I don't see a problem applying the strategy we already use to other packages, but you should be aware it might be changing in the near future.

serkan-ozal:
@dyladan Yes, I know. I just wanted to be sync with the current approach and if it will be changed in the future, this one also can be changed along with other together.

dyladan (approved):

dyladan:
@serkan-ozal sorry i was mostly putting that there for people in the future wondering why things might be the way they are, and for other reviewers.

trentm (approved):

PR #2638, opened 2025-01-10
feat(instrumentation-express): propagate context and measure full handler spans

### LINKED ISSUE #2022: @opentelemetry/instrumentation-express handler does not run in correct context
```
├── @opentelemetry/api@1.8.0
└─┬ @opentelemetry/auto-instrumentations-node@0.43.0
  ├── @opentelemetry/api@1.8.0 deduped
  └── @opentelemetry/instrumentation-express@0.36.1
```

v18.19.1

### What did you do?

Run the included express sample https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/express

Child spans of the handler should be children under the `request handler` span

### What did you see instead?

Child spans created under the request handler fall under the top level server span (`GET /run_test`), which I believe is created by the `@opentelemetry/instrumentation-http`. I.e. they appear as siblings of the server span, next to middleware. This is really confusing because you can't tell if the span belongs to a middleware or the handler.

This screenshot demonstrates the problems
image

### Additional context

It looks like the wrapped handler is not executed with the created span set in context:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/9df30ea0fd822a69203b818b8fbe34e1e1c8bced/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L308

Maybe this is intentional for 3-arg handlers that accept a `next()` function, as the next layer would be running in the parent's span context. If that is considered incorrect, you could probably solve this by wrapping the `next()` function to restore the parent context before executing the next layer to keep everything as siblings.

### LINKED ISSUE #2023: @opentelemetry/instrumentation-express handler does not account for async work
```
├── @opentelemetry/api@1.8.0
└─┬ @opentelemetry/auto-instrumentations-node@0.43.0
  ├── @opentelemetry/api@1.8.0 deduped
  └── @opentelemetry/instrumentation-express@0.36.1
```

v18.19.1

### What did you do?

Run the included express sample https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/examples/express

Async work should be accounted for in the handler duration.

### What did you see instead?

The handler span has very short duration, only accounting for synchronous work. This is already a known caveat documented [here](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/9df30ea0fd822a69203b818b8fbe34e1e1c8bced/plugins/node/opentelemetry-instrumentation-express#caveats) but I couldn't find a tracking bug.

This screenshot demonstrates the problems
image

### Additional context

A common pattern for express is to use an async function as the handler, as demonstrated in the sample https://github.com/open-telemetry/opentelemetry-js-contrib/blob/9df30ea0fd822a69203b818b8fbe34e1e1c8bced/examples/express/src/server.ts#L36

For handlers that return a promise, it should be possible to chain a `span.end()` call to the promise. However this also seems related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2022

### LINKED ISSUE #2146: Auto instrumentation of express app is not sending exception event in trace
"_@opentelemetry/api_": "1.8.0"
"_@opentelemetry/auto-instrumentations-node_": "0.44.0"

_v20.12.2_

### What did you do?
Ran a simple express app which raises exception.

```javascript
const app = express();


function ErrorGenerator(a, b) {
    if (b === 0) {
        throw new Error('Division by zero error');
    }
    return a / b;
}


app.get('/', (req, res) => {
    const result = ErrorGenerator(10, 0);
    console.log('Result:', result);
    res.send("Here is results of highest computed maths");
});

app.use((err, req, res, next) => {
    console.error('An error occurred:', err.message);
    res.status(400).send('Bad Request: ' + err.message);
    next(err);
  });

const port = 6002;
app.listen(port, () => {
    console.log(`Server is running on port ${port}`);
});

```

On hitting the endpoint get a span with exception in event array. Somewhat similar to this:

```javascript
{
  traceId: '7d57f374b49081a2c46aee1b9334d1c7',
  parentId: undefined,
  traceState: undefined,
  name: 'GET',
  id: 'dfdebec418911ea9',
  kind: 2,
  timestamp: 1714033150562000,
  duration: 3011281.152,
  attributes: {
    'http.url': 'http://169.254.169.254/computeMetadata/v1/instance',
    'http.method': 'GET',
    'http.target': '/computeMetadata/v1/instance',
    'net.peer.name': '169.254.169.254',
    'http.host': '169.254.169.254:80',
    'http.error_name': 'Error',
    'http.error_message': 'socket hang up'
  },
  status: { code: 2, message: 'socket hang up' },
  events: [
    {
      name: 'exception',
      attributes: {
        'exception.type': 'ECONNRESET',
        'exception.message': 'socket hang up',
        'exception.stacktrace': 'Error: socket hang up\n' +
          '    at Socket.socketCloseListener (node:_http_client:473:25)\n' +
          '    at Socket.emit (node:events:530:35)\n' +
          '    at TCP.<anonymous> (node:net:337:12)\n' +
          '    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)'
      },
      time: [ 1714033153, 573273845 ],
      droppedAttributesCount: 0
    }
  ],
  links: []
}
```

### What did you see instead?

```javascript
{
  traceId: 'f2a2854b4e599d5abb83960e4904fb79',
  parentId: undefined,
  traceState: undefined,
  name: 'GET /',
  id: '46e1fb4811ce2415',
  kind: 1,
  timestamp: 1714033162279000,
  duration: 9807.383,
  attributes: {
    'http.url': 'http://localhost:6002/',
    'http.host': 'localhost:6002',
    'net.host.name': 'localhost',
    'http.method': 'GET',
    'http.scheme': 'http',
    'http.target': '/',
    'http.user_agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4.1 Safari/605.1.15',
    'http.flavor': '1.1',
    'net.transport': 'ip_tcp',
    'net.host.ip': '::1',
    'net.host.port': 6002,
    'net.peer.ip': '::1',
    'net.peer.port': 50140,
    'http.status_code': 400,
    'http.status_text': 'BAD REQUEST',
    'http.route': '/'
  },
  status: { code: 0 },
  events: [],
  links: []
}
```

### LINKED ISSUE #2442: express instrumentation - span end prematurely for non-middleware instrumentations
- [x] This may affect other libraries, but I would like to get opinions here first

Hi, I want to verify if this is a bug or intended behavior of the current instrumentation code.
I notice in the instrumentation code: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L236-L242

The span is ended right after it is created unless it's a middleware. It does not even measure the sync execution time of the actual handler.

Should it be put after this line to measure the original handle execution time? https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L285

I ran a few tests, the time measured for 'request handler' is extremely short, only a couple of microseconds.

Much appreciated for looking into this!

### LINKED ISSUE #2469: Can't update span name when using OpenTelemetry Express Instrumentation for Node.js
```json
    "@opentelemetry/exporter-trace-otlp-http": "^0.53.0",
    "@opentelemetry/instrumentation-express": "^0.42.0",
    "@opentelemetry/instrumentation-grpc": "^0.53.0",
    "@opentelemetry/instrumentation-http": "^0.53.0",
    "@opentelemetry/resources": "^1.26.0",
    "@opentelemetry/sdk-trace-node": "^1.26.0",
    "@opentelemetry/semantic-conventions": "^1.27.0",
```

20.8.0

### What did you do?
I am using OpenTelemetry Express Instrumentation for Node.js (`@opentelemetry/instrumentation-express`) to take advantage of better and more precise span names for routes in Express (e.g. `GET /elements/:elementId` for GET requests with query parameters).
However, I noticed that once I start using Express instrumentation, I can no longer update span names. When I am using just instrumentation for HTTP I can do it without any problems (I can see `GET / (hello world)` span name in Jaeger) but the following code is no longer working when I use Express instrumentation:
```ts
app.get('/', (_req, res) => {
	trace.getActiveSpan()?.updateName('GET / (hello world)');
	res.json({ hello: 'Hello world' });
});
```
**How can I update name of the span when using Express instrumentation?** I was trying to use `spanNameHook` configuration option but this seems to be working only for child spans (e.g. for middleware related spans) or `requestHook` but it also didn't work.
I wanted to use custom names e.g. for some spans related to serving static files (e.g. `GET /index.html` or `GET /app.js`).
Thanks in advance for your help!

I am attaching the minimal reproducible example.
Startup command:  `nodemon --require ./src/tracing.ts --watch ./src ./src/main.ts`
File `main.ts`:
```ts
import { trace } from '@opentelemetry/api';
import express from 'express';

const port = 38000;
const app = express();
app.use(express.json());
app.get('/', (_req, res) => {
	trace.getActiveSpan()?.updateName('GET / (hello world)');
	res.json({ hello: 'Hello world' });
});
app.listen(port);
console.log(`HTTP server started on port: ${port}`);
```

File `tracing.ts`:
```ts
import { diag, DiagConsoleLogger, DiagLogLevel } from '@opentelemetry/api';
import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http';
import { registerInstrumentations } from '@opentelemetry/instrumentation';
import { ExpressInstrumentation } from '@opentelemetry/instrumentation-express';
import { GrpcInstrumentation } from '@opentelemetry/instrumentation-grpc';
import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
import { Resource } from '@opentelemetry/resources';
import { SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base';
import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
import { ATTR_SERVICE_NAME, ATTR_SERVICE_VERSION } from '@opentelemetry/semantic-conventions';

const url = 'http://localhost:4318/v1/traces';

function initializeTracing() {
	diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.WARN);

	const provider = new NodeTracerProvider({
		resource: new Resource({
			[ATTR_SERVICE_NAME]: 'ol-express-test',
			[ATTR_SERVICE_VERSION]: '1.0.0',
		}),
	});

	const exporter = new OTLPTraceExporter({ url });
	console.log(`Initializing trace exporter with URL: '${exporter.url}'`);

	provider.addSpanProcessor(new SimpleSpanProcessor(exporter));
	provider.register();

	registerInstrumentations({
		instrumentations: [
			new GrpcInstrumentation(),
			new HttpInstrumentation(),
			new ExpressInstrumentation(),
		],
	});
}

initializeTracing();

```

I was expecting the same result as without Express instrumentation: a span with `GET / (hello world)` name:

### What did you see instead?
Unfortunately the exported span name is default (`GET /`):

### LINKED ISSUE #2617: Can't get active span in Express instrumentation.
I followed the very simple instrumentation instructions:

```
const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
const { registerInstrumentations } = require('@opentelemetry/instrumentation');
const { HttpInstrumentation } = require('@opentelemetry/instrumentation-http');

provider.register();

registerInstrumentations({
  instrumentations: [
    // Express instrumentation expects HTTP layer to be instrumented
    new HttpInstrumentation(),
    new ExpressInstrumentation(),
  ],
});
```
And in my express code I am trying to get the active span:
```
app.get("/api/results/:id", async (req, res) => {
  const localSpan = opentelemetry.trace.getActiveSpan();
```
But it ends up being null??

raphael-theriault-swi:

Currently the Express instrumentation does not propagate context and does not properly instrument request handlers.

Context is propagated for middleware and request handler spans, while their `next` callback is reset to the parent context to avoid extreme nesting as suggested in #2022. Request handler spans are also no longer ended prematurely.

Router spans are kept as-is since they are broken and propagating context for them would make things even more confusing than they already are.

### Before
example trace before the changes

### After
example trace after the changes

## Related issues

- Fixes #2022
- Fixes #2023
- Fixes #2146
- Fixes #2442
- Fixes #2469
- Fixes #2617 
- Supersedes #2603

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
@@ -215,6 +215,9 @@ export class ExpressInstrumentation extends InstrumentationBase<ExpressInstrumen
           attributes: Object.assign(attributes, metadata.attributes),
         });
 
+        const parentContext = context.active();
+        let currentContext = trace.setSpan(parentContext, span);
+
         const { requestHook } = instrumentation.getConfig();
         if (requestHook) {
           safeExecuteInTheMiddle(
@@ -234,12 +237,15 @@ export class ExpressInstrumentation extends InstrumentationBase<ExpressInstrumen
         }
 
         let spanHasEnded = false;
+        // TODO: Fix router spans (getRouterPath does not work properly) to
+        // have useful names before removing this branch
         if (
-          metadata.attributes[AttributeNames.EXPRESS_TYPE] !==
-          ExpressLayerType.MIDDLEWARE
+          metadata.attributes[AttributeNames.EXPRESS_TYPE] ===
+          ExpressLayerType.ROUTER
         ) {
           span.end();
           spanHasEnded = true;
+          currentContext = parentContext;
         }
         // listener for response.on('finish')
         const onResponseFinish = () => {
@@ -278,12 +284,12 @@ export class ExpressInstrumentation extends InstrumentationBase<ExpressInstrumen
               (req[_LAYERS_STORE_PROPERTY] as string[]).pop();
             }
             const callback = args[callbackIdx] as Function;
-            return callback.apply(this, arguments);
+            return context.bind(parentContext, callback).apply(this, arguments);
           };
         }
 
         try {
-          return original.apply(this, arguments);
+          return context.bind(currentContext, original).apply(this, arguments);
         } catch (anyError) {
           const [error, message] = asErrorAndMessage(anyError);
           span.recordException(error);
@@ -296,7 +302,7 @@ export class ExpressInstrumentation extends InstrumentationBase<ExpressInstrumen
           /**
            * At this point if the callback wasn't called, that means either the
            * layer is asynchronous (so it will call the callback later on) or that
-           * the layer directly end the http response, so we'll hook into the "finish"
+           * the layer directly ends the http response, so we'll hook into the "finish"
            * event to handle the later case.
            */
           if (!spanHasEnded) {
--- a/plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
@@ -96,7 +96,7 @@ describe('ExpressInstrumentation', () => {
           );
           assert.strictEqual(response, 'tata');
           rootSpan.end();
-          assert.strictEqual(finishListenerCount, 2);
+          assert.strictEqual(finishListenerCount, 3);
           assert.notStrictEqual(
             memoryExporter
               .getFinishedSpans()
@@ -201,7 +201,7 @@ describe('ExpressInstrumentation', () => {
           );
           assert.strictEqual(response, 'tata');
           rootSpan.end();
-          assert.strictEqual(finishListenerCount, 2);
+          assert.strictEqual(finishListenerCount, 3);
           assert.notStrictEqual(
             memoryExporter
               .getFinishedSpans()


raphael-theriault-swi:
cc @pichlermarc fixes discussed wednesday

deejay1 (approved):

pichlermarc (approved):
this looks good, thank you for working on this. 🙌 
cc @JamieDanielson and @pkanal (component owners) for comments :)

mikesir87:
Apologies for the outsider coming in on this PR, but reporting that I've manually applied this PR into a local working copy to validate it works with my sample apps and it worked! In fact, it already helped me catch a case or two where I was missing an `await` call on a Kafka publish or DB update that was running after the request handler had returned. Excited to see this contribution! 😄

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -234,12 +237,15 @@ export class ExpressInstrumentation extends InstrumentationBase<ExpressInstrumen
         }
 
         let spanHasEnded = false;
+        // TODO: Fix router spans (getRouterPath does not work properly) to
```
dyladan, line 240: Can you expand on what you mean by not working properly or create an issue so this doesn't get lost?

raphael-theriault-swi: You can see in the [nested routers test](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/instrumentation-express-v0.47.0/plugins/node/opentelemetry-instrumentation-express/test/express.test.ts#L680-L722) that the created router spans never include [the `/posts` section of the route](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/instrumentation-express-v0.47.0/plugins/node/opentelemetry-instrumentation-express/test/fixtures/use-express-nested-router.mjs#L50-L64). If I were to propagate context and nest them we'd end up with `/api/user/:id` -> `/:postId` -> `/api/user/posts/:postId` which is really confusing.

raphael-theriault-swi:
@JamieDanielson @pkanal Sorry, gonna bother you again for an approval on this one :)

JamieDanielson (approved):
Thank you for working on this! Should we mark this as a breaking change (still experimental but mostly for signaling) since it updates parent span ID for some spans?

pichlermarc:
@JamieDanielson good point, I changed the title so that it's categorized as such. 🙂

Issue #2516, opened 2024-11-04
instrumentation-mongodb: perfromance degradation due to _scrubStatement


DanStory:
### What version of OpenTelemetry are you using?
```
@opentelemetry/api: 1.9.0
@opentelemetry/sdk-trace-base: 1.26.0
@opentelemetry/sdk-trace-node: 1.26.0
@opentelemetry/instrumentation-mongodb: 0.48.0
```
### What version of Node are you using?
```
node: v20.17.0
```

### What did you do?
Insert/update mongo documents that have large datasets (BinData, Arrays, etc).
Namely using GridFS to upload large files (which are chunked into 255KiB documents)

### What did you expect to see?
GridFS upload of a 430MiB file takes around a minute (as of `instrumentation-mongodb: 0.44.0`)

### What did you see instead?
GridFS upload of a 430MiB file takes over 5 minutes  (as of `instrumentation-mongodb: 0.45.0` and after)

### Additional context
We were able to deduce that the introduce of `_scrubStatement` as being the culprit: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1728/files#diff-0dd65d8fe0cee39d03f2ddd06f60730b20553e33521842fdb553056bfc1c8231R929

It is iterating the mongo cmd operation (which includes the document) recursively through nested objects and arrays, namely with no disregard of the operation (for GridFs upload) to contain binary data upwards of (by default) `261120` items in an Array/BinData/Buffer. I will also note, the implementation would not (IMO) handle Date, ObjectId, etc. correctly either. 

<img width="1477" alt="image" src="https://github.com/user-attachments/assets/880ad522-33a6-4ca6-a54c-523bec5ee503">

trentm:
I added some notes at https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2281#issuecomment-2489881451 that relate to this issue as well.

Issue #2281, opened 2024-06-14
Circular reference in _scrubStatement causes maximum call stack exceeded in instrumentation-mongodb 


tzjames:
I am using mongoose which seems to have added a __parentArray, to an object that is part of an array.  Hence there is a circular reference and so _scrubStatement then throws:
`@opentelemetry/instrumentation-mongodb Error running dbStatementSerializer hook RangeError: Maximum call stack size exceeded at /Users/james/Library/CloudStorage/Dropbox/growthbookzip/growthbook/node_modules/@opentelemetry/instrumentation-mongodb/build/src/instrumentation.js:581:65 at Array.map (<anonymous>) at MongoDBInstrumentation._scrubStatement`

This was introduced in [instrumentation-mongodb](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1728).

You probably need to add a circular reference check. Or some hard coded check for __parentArray, since I'm assuming mongoose is pretty common.  

### What version of OpenTelemetry are you using?
```
    "@opentelemetry/api": "^1.9.0",
    "@opentelemetry/auto-instrumentations-node": "^0.47.1",
    "@opentelemetry/exporter-metrics-otlp-proto": "^0.52.0",
    "@opentelemetry/sdk-metrics": "^1.25.0"
```

### What version of Node are you using?
`v20.12.2`

### What did you do?
Ran some DB query that got automatically scrubbed. 

My mongoose schema has a property `phases` that is defined as:
```
 phases: [
    {
      _id: false,
      dateStarted: Date,
      dateEnded: Date,
      phase: String,
      name: String,
      reason: String,
      coverage: Number,
      condition: String,
      savedGroups: [
        {
          _id: false,
          ids: [String],
          match: String,
        },
      ],
      prerequisites: [
        {
          _id: false,
          id: String,
          condition: String,
        },
      ],
      namespace: {},
      seed: String,
      variationWeights: [Number],
      groups: [String],
    },
  ],
```
It probably was a create statement.

### What did you expect to see?
No Errors.

### What did you see instead?
The maximum call stack error above.

### Additional context
None

tzjames:
@Fox32 Pinging you since I think you introduced it in your [PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1728) a week ago.  Pinging @david-luna as well as you may have context here having approved it.  

Thanks for your help!

Fox32:
[...]

PR #2653, opened 2025-01-16
feat(instrumentation-mongodb): Lighten scrubbed data

onurtemizkan:

Resolves: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2281
Resolves: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2516

- `_scrubStatement` scrubs all data, including binary data, and with large datasets it can create performance issues.
- Also, it seems it can also end up scrubbing circular references, which can be added by `mongoose`. (eg: `__parentArray`)

This PR updates the behaviour as described in @trentm's [comment](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2281#issuecomment-2489881451).

- Sets a depth limit of `10` for recursion inside deeply nested objects. Please let me know if that needs to be configurable.
- Skips `Buffer` and `Uint8Array` data while scrubbing.
- Skips `mongoose`-added keys, which are marked as `non-enumerable` from mongoose versions after `5.13.14`.

Note: I was not able to test with `mongoose` versions under `5.13.14` in the unit tests due to type mismatches between `mongodb@6.8.0` and `mongoose < 5.13.14` `devDependencies`.


Matth--:
@onurtemizkan should this be re-opened since you have been added as owner? I see this PR as a fix for both open issues linked to this PR.

Issue #2645, opened 2025-01-14
[all] migrate from deprecated `addSpanProcessor` to `*TracerProvider` constructor option


pichlermarc:
# Description

`*TracerProvider#addSpanProcessor()` is deprecated in favor of a constructor option and will be removed in SDK 2.0. To ensure a smoother transition for the contrib repo when we publish SDK 2.0, we should take care of this before we release 2.0 on February 17. 

### This issue is considered done when we have applied these changes in all packages of this repository

- we moved from `*TracerProvider#addSpanProcessor()` to the constructor option
- we made the necessary changes to the documentation

### Packages checklist:

> [!NOTE]
> **For traigers/approvers/maintainers** please link the appropriate PRs and update check boxes as we work through them. Some of these packages may not use this method anywhere. In that case, please simple mark them as done and add a comment "not used"

- [ ] @opentelemetry/resource-detector-alibaba-cloud
- [ ] @opentelemetry/instrumentation-user-interaction
- [ ] @opentelemetry/resource-detector-container
- [ ] @opentelemetry/instrumentation-generic-pool
- [ ] @opentelemetry/instrumentation-document-load
- [ ] @opentelemetry/resource-detector-instana
- [ ] @opentelemetry/instrumentation-nestjs-core
- [ ] @opentelemetry/resource-detector-github
- [ ] @opentelemetry/instrumentation-aws-lambda
- [ ] @opentelemetry/resource-detector-azure
- [ ] @opentelemetry/instrumentation-cassandra-driver
- [ ] @opentelemetry/instrumentation-memcached
- [ ] @opentelemetry/instrumentation-long-task
- [ ] @opentelemetry/resource-detector-aws
- [ ] @opentelemetry/resource-detector-gcp
- [ ] @opentelemetry/instrumentation-aws-sdk
- [ ] @opentelemetry/instrumentation-connect
- [ ] @opentelemetry/instrumentation-express
- [ ] @opentelemetry/instrumentation-fastify
- [ ] @opentelemetry/instrumentation-graphql
- [ ] @opentelemetry/instrumentation-ioredis
- [ ] @opentelemetry/instrumentation-mongodb
- [ ] @opentelemetry/instrumentation-redis-4
- [ ] @opentelemetry/instrumentation-restify
- [ ] @opentelemetry/instrumentation-winston
- [ ] @opentelemetry/instrumentation-bunyan
- [ ] @opentelemetry/instrumentation-mysql2
[...]

PR #2657, opened 2025-01-17
docs(*): Migrate away from deprecated `addSpanProcessor` in README

chancancode:
This migrates from the deprecated `addSpanProcessor` method to the constructor option in preparation for the upcoming SDK 2.0 release.

This commit migrates everything in the README.md of each package, other than that there are no runtime code changes.

Technically this is a user-facing change for users that previously followed the examples, putting it in its own PR/commit so we can label it for learna-changelog if desired (probably a good idea, in case when it comes time to publishing, this ends up being the only change).

Ref #2645 (cc @pichlermarc)


david-luna (approved):

david-luna:
@chancancode I've updated the PR title to pass the lint process. Thanks for your contribution :)

chancancode:
@david-luna updated the branch from the UI here, let me know if you need a proper rebase

PR #2660, opened 2025-01-17
refactor(test-utils): Migrate away from deprecated `addSpanProcessor`

chancancode:
This migrates from the deprecated `addSpanProcessor` method to the constructor option in preparation for the upcoming SDK 2.0 release.

This commit migrates the **runtime code** in the test-utils package and should be included in the package's CHANGELOG.

Ref #2645 (cc @pichlermarc)


chancancode:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/.github/component-label-map.yml#L5

I suppose these rules are to ping maintainers to look at any changes to the shared test utils, which is fair enough (though this is really minor).

However, **for the purpose of lerna-changelog, the pkg labels are definitely wrong**. That should be fixed before merging.

david-luna:
@chancancode once you update the branch I'll proceed to merge it :)

chancancode:
@david-luna pressed the button on the web UI, does that work for you?

PR #2285, opened 2024-06-19
chore: permanently deprecate opentelemetry-browser-extension-autoinjection

svrnm:
As discussed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1152#issuecomment-2173744798 this PR will permanently deprecate the OpenTelemetry Browser Extension in this repository. While this was not a lighthearted decision, I think it is the right thing to do, since it

* reduces the maintanance bruden for the JS maintainers
* allows me to focus on the areas of the OTel project that need my attention
* allows other extensions (like https://github.com/tbrockman/opentelemetry-browser-extension) to shine and take it's place

In this PR I make sure that it is clear that this is permanent, but that there are alternatives and that they can be found via the Registry (and people can add more if they want to)

@SimenB I appreciate the help you provided on fixing and debugging the underlying issue with webpack4&webpack5 and I apologize that this didn't land

Note, that I kicked off a conversation with @tbrockman about the extension at https://github.com/tbrockman/opentelemetry-browser-extension and shared some if the ideas I had for the extension. I also played around with this extension and it is much feature-richer and stable than what I created, so at the end this is a happy ending!

Thank you all

PR #2661, opened 2025-01-17
chore: permanently delete opentelemetry-browser-extension-autoinjection

chancancode:
This package was archived in #2285, the code no logner builds and contains legacy/deprecated patterns that will stop working with SDK 2.0, which we have no intention to fix.

Added a link to the last snapshot of the code for those who are looking to reference the archived code.

Uses [this GitHub-specific markdown annotation](https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/).

* * *

I found this while working on #2645, specifically this usage:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/archive/opentelemetry-browser-extension-autoinjection/src/instrumentation/WebInstrumentation.ts#L45-L81

Since the caller passes in an already-instantiated provider, this cannot be easily fixed. In any case, we clearly have no intention to release a new version of this package so it's a bit moot.

The intention was to freeze the code, but it's annoying that it comes up in the search for these type of migrations/refactor efforts. Looking at the commit history, it was already updated a handful of times by similar efforts despite the stated intention to keep the code frozen.

In #2285, @blumamir said:

> I support removing the code (which one can also find in git by checking out to an early commit) and leaving just a note in the README to communicate the status so that it won't go away for some reasonable time.
> 
> The code pops up in IDE searches and may still promote old practices which I don't see any good reason to keep anymore.

...and @pichlermarc said...

> Let's merge this PR and then we can still follow up on removing the code.

I agree with @blumamir so here I am finishing the job 😀


pichlermarc (approved):
Thank you 🙂

Issue #2590, opened 2024-12-06
Issue with 3.11.5 mysql2 span not being generated


liustve:
### What version of OpenTelemetry are you using?
0.54.0 (latest)

### What version of Node are you using?
v22.10.0

### What did you do?
 
Set up an express server to make a call to a dummy nonexistant database using mysql2:
import mysql from 'mysql2';

  const connection = mysql.createConnection({
    host: process.env.DB_HOST,
    user: process.env.DB_USER,
    password: process.env.DB_PASSWORD
  });
  
  connection.query(
    'CREATE DATABASE IF NOT EXISTS db',
    (err, results) => {
      if (err) {
        console.error('Error creating database:', err);
        return;
      }
      console.log('Database created successfully');
    }
  );

Ran using version 3.11.4, db span exists and gets populated:
<img width="592" alt="image" src="https://github.com/user-attachments/assets/91f4876c-c635-42ee-9580-7f3062b69592">

Ran using version 3.11.5, no db span:
<img width="656" alt="image" src="https://github.com/user-attachments/assets/1c7f7f66-309f-461f-9fed-299d021faf69">

### What did you expect to see?
DB span to populate in newest mysql2 version

### What did you see instead?
No span running on the latest mysql2 version and latest otel js version.

### Additional context
Related issue: https://github.com/aws-observability/aws-otel-js-instrumentation/issues/127

pichlermarc:
huh, weird - I though this was fixed in #2579

pichlermarc:
@liustve thanks for reaching out - assigning P2 (incomplete or missing telemetry).

david-luna:
Hi @liustve 

I've tried to reproduce the issue with no luck in this [small project](https://github.com/david-luna/otel-js-contrib-2590). For both versions of mysql2 (3.11.4 & 3.11.5) the instrumentation gets the db statement.

When you mention the version `0.54.0 (latest)` which package are you referring to? I?m asking because there are several packages involved in the tracing setup.

this is the list of dependencies of the project
```bash
npm ls
otel-js-contrib-2590@1.0.0 /Users/david/Documents/repos/dl/otel-js-contrib-2590
├── @opentelemetry/api@1.9.0
├── @opentelemetry/instrumentation-http@0.56.0
├── @opentelemetry/instrumentation-mysql2@0.44.0
├── @opentelemetry/instrumentation@0.56.0
├── @opentelemetry/sdk-trace-base@1.29.0
[...]

PR #2662, opened 2025-01-17
fix: missing telemetry in mysql2 when importing only promise API

macno:

See #2590

Patch `lib/connection.js` that is always imported


plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -1210,13 +1205,160 @@ describe('mysql2', () => {
...
+
+      mysqlTypesPromReload = await import('mysql2/promise');
+
+      if (!shouldTest) {
+        // this.skip() workaround
+        // https://github.com/mochajs/mocha/issues/2683#issuecomment-375629901
+        this.test!.parent!.pending = true;
+        this.skip();
+      }
+      provider.addSpanProcessor(new SimpleSpanProcessor(memoryExporter));
```
pichlermarc, line 1274: `addSpanProcessor` is deprecated. We'll soon publish 2.0 of `@opentelemetry/sdk-trace-base` where this will be removed, so it's best to avoid it wherever we can to make our lives easier later on.

The change is only required for this new test code here, you can leave the old one as-is. #2659 will take care of the old ones.

Ref #2645

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -1210,13 +1205,160 @@ describe('mysql2', () => {
...
+    let contextManager: AsyncHooksContextManager;
+    const provider = new BasicTracerProvider();
+    const memoryExporter = new InMemorySpanExporter();
+
+    let connection: mysqlTypesProm.Connection;
+    let rootConnection: mysqlTypesProm.Connection;
+    let mysqlTypesPromReload: typeof mysqlTypesProm;
+
+    before(async function () {
+      // cleanup cache for mysql2
```
pichlermarc, line 1253:
```suggestion
      // cleanup cache for 'mysql2'
      // in previous iterations these tests would have passed since 'mysql2' was already loaded,
      // but 'mysql2/promise' does not import 'mysql2'. Therefore, if used without ever loading 'mysql2',
      // the relevant code would never get patched.
```

pichlermarc (commented):
Overall looks good, thanks for chasing this down. 🙌 
Just one comment nit and one blocker around `addSpanProcessor`.

Due to the re-format and change in nesting, the diff here initially looked more daunting than it actually was 🙂

macno:
> Overall looks good, thanks for chasing this down. 🙌 Just one comment nit and one blocker around `addSpanProcessor`.
> 
> Due to the re-format and change in nesting, the diff here initially looked more daunting than it actually was 🙂

yeah, sorry about that, but I had to find a way to use different before/after blocks.

I'll take care of the suggestions, thank you!

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -1210,13 +1205,163 @@ describe('mysql2', () => {
...
+          delete require.cache[key];
+        }
+      });
+
+      const { MySQL2Instrumentation } = await import('../src');
+      instrumentation = new MySQL2Instrumentation();
+      instrumentation.enable();
+      instrumentation.disable();
+
+      mysqlTypesPromReload = await import('mysql2/promise');
```
david-luna, line 1270: nit: by importing the types at the top of the file you can avoid loading twice the module which may improve perf (although not noticeable) and avoid any possible odd scenario. Actually I found quite misfortunate to name `mysqlTypes` and `mysqlTypesProm` the imported modules. The `types` in the names is quite misleading since it's the actual module object

david-luna (approved):

PR #2663, opened 2025-01-17
feat(detectors): mark as side effect free

serkan-ozal:

In my previous [PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2636), I have forgotten to mark resource detector modules as side effect free. So, during my internal tests with custom built resource detector packages, I have noticed that they are not tree-shaked by webpack even though they are not referenced.

Add `sideEffects` property as `false` to the `package.json` files of all the resource detector modules.


pichlermarc (approved):
thanks 👍

PR #2664, opened 2025-01-17
chore(instrumentation-pg): have `@opentelemetry/semantic-conventions` package dependency with caret (`^`) version to sync with other instrumentation packages

serkan-ozal:
… so latest minor versions will be able to used

Currently, `@opentelemetry/instrumentation-pg` package has `@opentelemetry/semantic-conventions` dependency with direct version (`1.27.0`), but other instrumentation packages have that with caret version (`^1.27.0`) of it. And this leads to having multiple copies of the `@opentelemetry/semantic-conventions` package with different versions in the AWS Lambda Node.js layer.  

Change `@opentelemetry/semantic-conventions` package version to `^1.27.0` to be sync with the other instrumentation packages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38732,7 +38732,7 @@
       "dependencies": {
         "@opentelemetry/core": "^1.26.0",
         "@opentelemetry/instrumentation": "^0.57.0",
-        "@opentelemetry/semantic-conventions": "1.27.0",
+        "@opentelemetry/semantic-conventions": "^1.27.0",
         "@opentelemetry/sql-common": "^0.40.1",
         "@types/pg": "8.6.1",
         "@types/pg-pool": "2.0.6"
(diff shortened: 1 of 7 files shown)


pichlermarc (changes requested):
this is pinned on purpose since the `incubating` entrypoint may include breaking changes in minor versions. 

See https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv for recommendations on how to handle this case.

pichlermarc (approved):
thanks 🙂 
cc @maryliag (component owner)

maryliag (approved):

trentm:
tl;dr: I think we should be using the local copy (src/semconv.ts) in *tests* as well.

[The following is copied from a [Slack discussion](https://cloud-native.slack.com/archives/C01NL1GRPQR/p1737137395232259?thread_ts=1737109456.189729&cid=C01NL1GRPQR) to capture this visibly somewhere.]

> If one uses the incubating entry point in tests... then the tests will be somewhat of a check that you properly copied the definuitions to your src/semconv.ts.

I'm changing my mind on this one.

Say one is using .../incubating in tests, then the version of the semantic-conventions dep is updated that drops one of the used unstable semconv definitions (say, ATTR_DB_CLIENT_CONNECTION_STATE). The runtime code will be fine, but the tests will fail. We don't want this test failure to be a reason to force the particular instrumentation to change for the new version of Semantic Conventions right then. (Eventually the instrumentations should migrate to newer Semantic Conventions, but it doesn't have to be right when we update the @opentelemetry/semantic-conventions dep for the whole opentelemetry-js-contrib.git repo.)

IOW, I think we should be using the src/semconv.ts in tests as well.

(Ensuring that the copied defs in src/semconv.ts match what is in the actual .../incubating endpoint should, ideally, be the job of a generator tool or a linting script of some sort.)

trentm (approved):
There was some discussion on Slack to refine this PR. It mimics https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2599 which was doing the same thing.

PR #2667, opened 2025-01-17
fix(auto-instrumentations-node): make SIGTERM shutdown test more reliable

garysassano:

The test `shuts down the NodeSDK when SIGTERM is received` is flaky due to making HTTP requests to `example.com`:

image

The key improvements are:

1. **Controlled Environment**: Instead of depending on `example.com` (which could be slow/unreachable), we use a local server that responds immediately.

2. **Better Event Handling**: Changed from using `req.on('close')` to `res.on('end')` which is more reliable for confirming the request completed.

3. **Proper Cleanup**: Added `server.close()` to the SIGTERM handler to ensure clean shutdown.

4. **Fixed Timing**: The local server responds immediately, reducing the chance of hitting the test's 1500ms timeout.

The main issue was likely that requests to `example.com` were taking too long or failing, causing the test to hit its timeout before seeing the `Finished request` message. By using a local server, we make the test much more deterministic and faster.

### DIFF
--- a/metapackages/auto-instrumentations-node/test/test-app/app-server.js
+++ b/metapackages/auto-instrumentations-node/test/test-app/app-server.js
@@ -17,22 +17,34 @@
 //Used in register.test.ts to mimic a JS app that stays alive like a server.
 const http = require('http');
 
-const options = {
-  hostname: 'example.com',
-  port: 80,
-  path: '/',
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
thanks, much appreciated 🙂

PR #2668, opened 2025-01-17
fix(detector-aws): remove semconv incubating import

garysassano:

The AWS detectors are directly importing experimental attributes from `@opentelemetry/semantic-conventions/incubating`, which can lead to breaking changes in minor versions and increased disk usage due to having multiple package versions installed.

- Created a shared `/src/semconv.ts` file with generated unstable semantic conventions
- Updated imports in AWS detectors to use the local `/src/semconv.ts` file

This follows the OpenTelemetry [best practices](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv) for handling unstable semantic conventions.


jj22ee (approved):

trentm:
@garysassano Thanks. Heads up that I pushed a couple of commits to your feature branch. Happy to discuss the changes:

- I moved src/lib/semconv.ts to src/semconv.ts. I don't see the value of the "lib/" subdir. Other packages will be using "src/semconv.ts" (e.g. instrumentation-pg already does).
- I added the block comments for the copied vars: these are helpful for tooltips and deprecated styling in some IDEs.  I've scripted the generation of semconv.ts (see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2669 for the PR adding the generator script).
- I moved two of the imports that are stable to import directly from `@opentelemetry/semantic-conventions`. The generator script warns about these cases.

garysassano:
> I moved src/lib/semconv.ts to src/semconv.ts. I don't see the value of the "lib/" subdir. Other packages will be using "src/semconv.ts" (e.g. instrumentation-pg already does).

I would have preferred using `/lib/semconv.ts` instead of `/src/semconv.ts` to allow imports from `/tests` as `../lib/semconv` rather than `../src/semconv.ts`. However, I noticed that the NX build fails unless everything is within the `/src` folder, so I'm fine with your changes.

pichlermarc (approved):

PR #2670, opened 2025-01-18
fix(instrumentation-aws-lambda): do not import from semantic-conventions incubating entry-point

trentm:
Refs: https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv


jj22ee (approved):

pichlermarc (approved):
(approving due to owner-approval)

PR #2671, opened 2025-01-20
feat(instrumentation-knex): Use newer semantic conventions

deejay1:

- Updates deprecated attributes with current ones and drop deprecated function

- Uses newer semantic conventions and replaces deprecated `SpanAttributes` call with `Attributes`
- Syncs attributes with PostgreSQL instrumentation


trentm:
I'll sponsor.

trentm:
@maryliag I'd appreciate your guidance here. Part of this PR is updating some `db.*` semconv attribute usage to more recent forms. E.g. changing `db.sql.table` to `db.collection.name` (which was part of semconv 1.25.0).  

Is that something we should be doing *now*? Or is that something we should wait on until after the semconv v1.33.0 release because of the pending stabilization of db semconv: https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.32.0 says:

>  This release is the second release candidate for the Database Semantic Conventions, with **db conventions stability planned to be declared in the subsequent release.**

Do you know if there will be a `db` stability migration thing using `OTEL_SEMCONV_STABILITY_OPT_IN` as was done for http (https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/)?

deejay1:
@trentm as far as I recall knex is already behind other modules regarding the attributes, so I don't see much of an issue making these changes opt-in.

maryliag:
I think it's okay to do them now, because we don't plan on making any other changes to them, at least not in the ones that were updated here. Regarding the opt-in, we do have some guidance on this on [semantic convention pages already](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md):

Existing database instrumentations that are using v1.24.0 of this document (or prior):

- SHOULD NOT change the version of the database conventions that they emit by default until the database semantic conventions are marked stable. Conventions include, but are not limited to, attributes, metric and span names, and unit of measure.
- SHOULD introduce an environment variable OTEL_SEMCONV_STABILITY_OPT_IN in the existing major version which is a comma-separated list of values. The list of values includes:
    - `database` - emit the new, stable database conventions, and stop emitting the old experimental database conventions that the instrumentation emitted previously.
    - `database/dup` - emit both the old and the stable database conventions, allowing for a seamless transition.
    - The default behavior (in the absence of one of these values) is to continue emitting whatever version of the old experimental database conventions the instrumentation was emitting previously.
    - Note: `database/dup` has higher precedence than database in case both values are present
- SHOULD maintain (security patching at a minimum) the existing major version for at least six months after it starts emitting both sets of conventions.
- SHOULD drop the environment variable in the next major version.

deejay1:
> So, I would say you can make the changes now, but with an opt-in for the new names and still sending old values, then create an issue to remove the old names on a following version, just to make sure we keep track of this and don't forget
Ok, I'll adapt this then. Thanks for the feedback.

deejay1:
Yay, I managed to bork the rebase somehow, but should be good now. Unfortunately it seems that the bot didn't remove the labels...

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -134,24 +165,45 @@ export class KnexInstrumentation extends InstrumentationBase<KnexInstrumentation
...
-          [SEMATTRS_DB_NAME]: name,
-          [SEMATTRS_NET_PEER_NAME]: config?.connection?.host,
-          [SEMATTRS_NET_PEER_PORT]: config?.connection?.port,
-          [SEMATTRS_NET_TRANSPORT]:
-            config?.connection?.filename === ':memory:' ? 'inproc' : undefined,
         };
+        const transport =
+          config?.connection?.filename === ':memory:' ? 'inproc' : undefined;
+
+        if ((semConv & SemconvStability.OLD) === SemconvStability.OLD) {
```
maryliag, line 174: I'm only seeing the case where you have old or stable, but you should also have the case for duplicate that sends both

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -45,8 +59,21 @@ const DEFAULT_CONFIG: KnexInstrumentationConfig = {
...
+
+    for (const entry of getStringListFromEnv('OTEL_SEMCONV_STABILITY_OPT_IN') ??
+      []) {
+      if (entry.toLowerCase() === 'database/dup') {
+        // database/dup takes highest precedence. If it is found, there is no need to read the rest of the list
+        this._semconvStability = SemconvStability.DUPLICATE;
+        break;
+      } else if (entry.toLowerCase() === 'database') {
+        this._semconvStability = SemconvStability.STABLE;
+      }
```
maryliag, line 75: I don't see the case where you set the value to `OLD`, but you're checking lower in the code for that case. Wa it suppose to be the default value?

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
deejay1: this is defined in the SemconvStability enum that:
```ts
export const enum SemconvStability {
  /** Emit only stable semantic conventions */
  STABLE = 0x1,
  /** Emit only old semantic conventions*/
  OLD = 0x2,
  /** Emit both stable and old semantic conventions*/
  DUPLICATE = 0x1 | 0x2,
}
```
so if it's defined to DUPLICATE the operations `(semConv & SemconvStability.OLD) === SemconvStability.OLD` and `(semConv & SemconvStability.STABLE) === SemconvStability.STABLE` will both return `true` and both conventions will be emitted

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
deejay1: the default value is defined in line 62:
```ts
export class KnexInstrumentation extends InstrumentationBase<KnexInstrumentationConfig> {
  private _semconvStability: SemconvStability = SemconvStability.OLD;
}
```

maryliag (approved):

maryliag:
cc @trentm

plugins/node/opentelemetry-instrumentation-knex/package.json
```diff
@@ -33,7 +33,9 @@
   "files": [
     "build/src/**/*.js",
     "build/src/**/*.js.map",
-    "build/src/**/*.d.ts"
+    "build/src/**/*.d.ts",
+    "LICENSE",
+    "README.md"
```
trentm, line 38: nit: This is fine, but I'd be inclined to not bother adding this. None of the other instrumentation packages in this repo have these. These two files are in the set of files that are always included by npm: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -35,8 +42,15 @@ import {
...
-import * as utils from './utils';
-import { KnexInstrumentationConfig } from './types';
+import {
+  ATTR_DB_COLLECTION_NAME,
+  ATTR_DB_NAMESPACE,
+  ATTR_DB_OPERATION_NAME,
+  ATTR_DB_QUERY_TEXT,
+  ATTR_DB_SYSTEM_NAME,
+  ATTR_DB_USER,
+} from './semconv';
```
trentm, line 52: Now that DB semconv has stabilized and https://github.com/open-telemetry/opentelemetry-js/releases/tag/semconv%2Fv1.33.0 was released, we can update the dep in package.json:

```diff
-    "@opentelemetry/semantic-conventions": "^1.27.0"
+    "@opentelemetry/semantic-conventions": "^1.33.0"
```

and import all of these (except `ATTR_DB_USER`) from `@opentelemetry/semantic-conventions`.
Similarly in "src/utils.ts".
I have a commit that makes these changes. (I used the [gen-semconv-ts.js](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/scripts/gen-semconv-ts.js) script to regenerate the "src/semconv.ts" file.)
Are you okay with me pushing that to your feature branch? (Or you can grab the diff from here: https://gist.github.com/trentm/5c2cb0cc28d4571f177f96b34ebb23f9)

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
trentm, line 75: FYI: Not for now (unless this PR waits for a bit), but there is a (not yet released) utility method being added to `@opentelemetry/instrumentation` for this: `databaseSemconvStabilityFromStr(process.env.OTEL_SEMCONV_STABILITY_OPT_IN)`. See https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation/src/semconvStability.ts#L17

This will be in `@opentelemetry/instrumentation@2.1.0`. The intent is to release this soon (perhaps this week).

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -122,6 +149,10 @@ export class KnexInstrumentation extends InstrumentationBase<KnexInstrumentation
 
   private createQueryWrapper(moduleVersion?: string) {
     const instrumentation = this;
+
+    // We need to bind it here, because `this` is not the same in the wrapper
+    const semConv = this._semconvStability;
```
trentm, line 154: Or could use `instrumentation._semconvStability` below. This is fine though.

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
trentm, line 174: This can be simplified to:

```suggestion
        if (semConv & SemconvStability.OLD) {
```

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -134,24 +165,45 @@ export class KnexInstrumentation extends InstrumentationBase<KnexInstrumentation
...
+            [SEMATTRS_DB_SQL_TABLE]: table,
+            [SEMATTRS_DB_OPERATION]: operation,
+            [SEMATTRS_DB_USER]: config?.connection?.user,
+            [SEMATTRS_DB_NAME]: name,
+            [SEMATTRS_NET_PEER_NAME]: config?.connection?.host,
+            [SEMATTRS_NET_PEER_PORT]: config?.connection?.port,
+            [SEMATTRS_NET_TRANSPORT]: transport,
+          });
+        }
+        if ((semConv & SemconvStability.STABLE) === SemconvStability.STABLE) {
```
trentm, line 186: This can be simplified to:

```suggestion
        if (semConv & SemconvStability.STABLE) {
```

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -134,24 +165,45 @@ export class KnexInstrumentation extends InstrumentationBase<KnexInstrumentation
...
+          Object.assign(attributes, {
+            [ATTR_DB_SYSTEM_NAME]: mapSystem(config.client),
+            [ATTR_DB_COLLECTION_NAME]: table,
+            [ATTR_DB_OPERATION_NAME]: operation,
+            [ATTR_DB_USER]: config?.connection?.user,
```
trentm, line 191: Should this be dropped? https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/ says `db.user` is removed

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
```diff
@@ -134,24 +165,45 @@ export class KnexInstrumentation extends InstrumentationBase<KnexInstrumentation
...
+            [ATTR_DB_NAMESPACE]: name,
+            [ATTR_SERVER_ADDRESS]: config?.connection?.host,
+            [ATTR_SERVER_PORT]: config?.connection?.port,
+            [ATTR_NETWORK_TRANSPORT]: transport,
```
trentm, line 195: https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/ says `network.transport` is removed.

plugins/node/opentelemetry-instrumentation-knex/README.md
```diff
@@ -70,6 +91,25 @@ Attributes collected:
...
+|----------------------|-----------------------------------------------------------------------------|
+| `db.namespace`       | This attribute is used to report the name of the database being accessed.   |
+| `db.operation.name`  | The name of the operation being executed.                                   |
+| `db.collection.name` | The name of the primary table that the operation is acting upon.            |
+| `db.query.text`      | The database statement being executed.                                      |
+| `db.system`          | An identifier for the database management system (DBMS) product being used. |
+| `db.user`            | Username for accessing the database.                                        |
+| `server.address`     | Remote hostname or similar.                                                 |
+| `server.port`        | Remote port number.                                                         |
+| `network.transport`  | Transport protocol used.                                                    |
```
trentm, line 111: https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/ suggests `network.transport` was removed.

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
deejay1: I grabbed the diff, thanks.

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
deejay1: I can wait for that, although I'll push the other changes on my branch.

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
deejay1: doh, seems kind of obvious in hindsight ;)

plugins/node/opentelemetry-instrumentation-knex/README.md
deejay1: removed

plugins/node/opentelemetry-instrumentation-knex/package.json
deejay1: removed, must have looked at another `package.json` in this repo for this probably

plugins/node/opentelemetry-instrumentation-knex/README.md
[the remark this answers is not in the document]
deejay1: I'll take another look at it tomorrow, as I'm somehow not satisfied with the current fixed version, yet I can't tell what bothers me there.

plugins/node/opentelemetry-instrumentation-knex/src/utils.ts
```diff
@@ -18,7 +18,7 @@ import { Exception } from '@opentelemetry/api';
 import {
   DB_SYSTEM_NAME_VALUE_POSTGRESQL,
   DB_SYSTEM_NAME_VALUE_SQLITE,
-} from './semconv';
+} from '@opentelemetry/semantic-conventions';
```
trentm, line 21: To use `DB_SYSTEM_NAME_VALUE_SQLITE` will require using a local `src/semconv.ts` soon. See:
https://github.com/open-telemetry/opentelemetry-js/pull/5690#issue-3067435791

tl;dr: `DB_SYSTEM_NAME_VALUE_SQLITE` is actually *un*stable and should not have been included in the stable `@opentelemetry/semantic-conventions` entry point.

(Eventually semconv-related things will settle down and this won't be so hard/obtuse/subtle.)

plugins/node/opentelemetry-instrumentation-knex/README.md
```diff
@@ -54,21 +54,38 @@ registerInstrumentations({
 
 ## Semantic Conventions
 
-This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+This package uses `@opentelemetry/semantic-conventions` version `1.33+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+
+This package is capable of emitting both Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md) and [Version 1.33.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.33.0/docs/database/database-metrics.md)
```
trentm, line 59: Did you intend to link the database *metrics* doc page?

plugins/node/opentelemetry-instrumentation-knex/package.json
```diff
@@ -56,8 +56,9 @@
     "typescript": "5.0.4"
   },
   "dependencies": {
+    "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/instrumentation": "^0.200.0",
```
trentm, line 60: `@opentelemetry/instrumentation@0.201.0` was just released, which includes [`semconvStabilityFromStr`](https://github.com/trentm/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation/src/semconvStability.ts#L81)

This dep was updated here as well, so merging changes from `main` will update to `^0.201.0`.

plugins/node/opentelemetry-instrumentation-knex/src/instrumentation.ts
trentm: I got the version wrong, it is in `@opentelemetry/instrumentation@0.201.0`, now published.

plugins/node/opentelemetry-instrumentation-knex/src/utils.ts
trentm: > To use `DB_SYSTEM_NAME_VALUE_SQLITE` will require using a local `src/semconv.ts` soon.

Perhaps easiest to wait for a v1.33.1 release of the semconv package, then create the src/semconv.ts again (perhaps using the https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/scripts/gen-semconv-ts.js tool). I'm happy to help with that.

plugins/node/opentelemetry-instrumentation-knex/README.md
deejay1: hmm, both have the tags ;) I'll link to `database-spans.md` when 1.33.1 comes out

deejay1:
@trentm frankly, I'm stumped why it started failing after the change

trentm:
> Edit: looks like https://ncjamieson.com/dont-export-const-enums/

I think you are right. Good find!
Sorry this PR has become the guinea pig for the `SemconvStability` addition. I'll start digging in. Currently I'm thinking we'll need a new `@opentelemetry/instrumentation@0.200.1` that drops the `const` from that enum.


> Lint

The lint step is failing because you need to now do:

```
npm uninstall @opentelemetry/core
```

That is assuming you'd rather not go back to a local/internal `semconvStability` implementation while `@opentelemetry/instrumentation` sorts out its `const enum` issue.

deejay1:
> Sorry this PR has become the guinea pig for the `SemconvStability` addition

No problem, I'll wait - someone has to test things ;)

As for the linter - I didn't even look at this being occupied with the previous error.

trentm:
> I'll start digging in.

My write-up: https://github.com/open-telemetry/opentelemetry-js/issues/5691
and PR: https://github.com/open-telemetry/opentelemetry-js/pull/5692

It being Friday afternoon, this will likely sit until Monday.

deejay1:
> It being Friday afternoon, this will likely sit until Monday.

Ok, I've updated everything now.

package-lock.json
```diff
@@ -45232,6 +45269,26 @@
...
+          "resolved": "https://registry.npmjs.org/@opentelemetry/instrumentation/-/instrumentation-0.201.1.tgz",
+          "integrity": "sha512-6EOSoT2zcyBM3VryAzn35ytjRrOMeaWZyzQ/PHVfxoXp5rMf7UUgVToqxOhQffKOHtC7Dma4bHt+DuwIBBZyZA==",
+          "requires": {
+            "@opentelemetry/api-logs": "0.201.1",
+            "@types/shimmer": "^1.2.0",
+            "import-in-the-middle": "^1.8.1",
+            "require-in-the-middle": "^7.1.1",
+            "shimmer": "^1.2.1"
+          }
+        },
```
trentm, line 45291:
```suggestion
```

package-lock.json updating sucks. There are multiple layouts of deps that satisfy a given set of package.json files. The one you currently have on this PR duplicates some packages to `plugins/node/opentelemetry-instrumentation-knex/node_modules` that don't need to be there. This happened because you added a dep on `@opentelemetry/instrumentation@0.201.1` in this subdir when the top-level still depended on version `0.201.0`. Then #2835 updated the dep to version `0.201.1` at the top-level.   

A subsequent `npm install` does not clean this up. `npm dedupe` *does* clean this up -- at least this case -- but running that on the current repo state *also* changes the install location of a few other packages, which doesn't belong on this PR.

Instead, I'll suggest a few manual deletions to package-lock.json to clean it up manually.

(The alternative is to revert package-lock.json to the state on main, then run:

```bash
cd plugins/node/opentelemetry-instrumentation-knex
npm install
```

package-lock.json
```diff
@@ -45241,6 +45298,11 @@
             "undici-types": "~5.26.4"
           }
         },
+        "@types/shimmer": {
+          "version": "1.2.0",
+          "resolved": "https://registry.npmjs.org/@types/shimmer/-/shimmer-1.2.0.tgz",
+          "integrity": "sha512-UE7oxhQLLd9gub6JKIAhDq06T0F6FnztwMNRvYgjeQSBeMc1ZG/tA47EwfduvkuQS8apbkM/lpLpWsaCeYsXVg=="
+        },
```
trentm, line 45305:
```suggestion
```

package-lock.json
```diff
@@ -36371,6 +36402,12 @@
         "undici-types": "~5.26.4"
       }
     },
+    "plugins/node/opentelemetry-instrumentation-knex/node_modules/@types/shimmer": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@types/shimmer/-/shimmer-1.2.0.tgz",
+      "integrity": "sha512-UE7oxhQLLd9gub6JKIAhDq06T0F6FnztwMNRvYgjeQSBeMc1ZG/tA47EwfduvkuQS8apbkM/lpLpWsaCeYsXVg==",
+      "license": "MIT"
+    },
```
trentm, line 36410:
```suggestion
```

package-lock.json
```diff
@@ -36362,6 +36362,37 @@
...
+        "require-in-the-middle": "^7.1.1",
+        "shimmer": "^1.2.1"
+      },
+      "engines": {
+        "node": "^18.19.0 || >=20.6.0"
+      },
+      "peerDependencies": {
+        "@opentelemetry/api": "^1.3.0"
+      }
+    },
```
trentm, line 36395:
```suggestion
```

trentm (commented):
LGTM after the suggested package-lock edits.

package-lock.json
trentm: Yes, I'm obsessing. :) It would be fine to merge this as is and later (or not) run npm dedupe.

trentm: In fact, let's just do that.

trentm (approved):
LGTM.
I'll follow up with a separate `npm dedupe` PR.

Thanks for doing this @deejay1!

PR #2674, opened 2025-01-20
fix(detector-aws): replace ecs deprecated semconv consts and hardcoded strings

garysassano:

- Updates AWS ECS detector tests to use proper semantic convention constants instead of deprecated ones
- Removes redundant semantic resource attributes file by using constants from semconv

- Replaced deprecated `SEMRESATTRS_*` and `CLOUDPROVIDERVALUES_*` constants with their `ATTR_*` and `*_VALUE_*` counterparts from semconv following [unstable semconv best practices](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv)
- Removed `SemanticResourceAttributes.ts` file as `ATTR_CLOUD_RESOURCE_ID` is now properly imported from `/src/semconv.ts`
- Organized imports alphabetically for better readability


pichlermarc:
@jj22ee (component owner) PTAL 🙂

detectors/node/opentelemetry-resource-detector-aws/test/detectors/AwsEcsDetector.test.ts
```diff
@@ -65,70 +67,96 @@ const assertEcsResource = (
   resource: Resource,
   validations: EcsResourceAttributes
 ) => {
-  assertCloudResource(resource, {
```
jj22ee, line 68: (applies to both test files) Wondering why you choose not to use this `assertCloudResource` utility method? I'm curious since these tests are still using `assertEmptyResource` and `assertContainerResource` which are also from `contrib-test-utils`.

If the purpose was to avoid having an import that uses `SEMRESATTRS_*` attributes, would it make more sense to remove the usage of those attributes in [opentelemetry-test-utils/src/resource-assertions.ts](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/opentelemetry-test-utils/src/resource-assertions.ts) instead? Otherwise, since the usage of these attributes are just in a test file, it should be okay to continue using `assertCloudResource`?

jj22ee: Also, was the omission of the `assertHasOneLabel('cloud', resource)` test intentional?
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/69cd638b986612b98310be822cd584ebcd93c647/packages/opentelemetry-test-utils/src/resource-assertions.ts#L68

jj22ee (commented):
lgtm! Have a couple questions regarding the tests:

detectors/node/opentelemetry-resource-detector-aws/test/detectors/AwsEcsDetector.test.ts
garysassano: > Wondering why you choose not to use this `assertCloudResource` utility method? I'm curious since these tests are still using `assertEmptyResource` and `assertContainerResource` which are also from `contrib-test-utils`.

I chose not to use `assertCloudResource` because it lacks certain attributes, specifically `cloud_platform` and `cloud_resource_id`, which are part of the [cloud resource](https://opentelemetry.io/docs/specs/semconv/resource/cloud/). I found it confusing to have some attributes validated within the utility and others outside of it.  


```ts
/**
 * Test utility method to validate a cloud resource
 *
 * @param resource the Resource to validate
 * @param validations validations for the resource attributes
 */
export declare const assertCloudResource: (resource: Resource, validations: {
    provider?: string;
    accountId?: string;
    region?: string;
    zone?: string;
}) => void;
```


More broadly, these utility functions are never fully in sync with the actual resource attributes, which is an ongoing issue. I discussed this also with @trentm, as maintaining them efficiently would require a generator similar to the one used for the experimental attributes from the `@opentelemetry/semantic-conventions/incubating` module.

garysassano: Also, why isn't there an `assertFaasResource`? It's all too inconsistent.

jj22ee (approved):

pichlermarc (approved):
thank you @garysassano  🙂 

FYI @jj22ee - we have recently assigned triage permissions to all component owners. If you approve a PR for a component owned by you, then you can now add the `has:owner-approval` label. 

I periodically go and check PRs with these labels and merge them if they were opened by people without permissions to merge them themselves. 🙂

PR #2681, opened 2025-01-27
test(instrumentation-mongodb): Include all versions in coverage

onurtemizkan:

- Test coverage of the MongoDB instrumentation is currently `49.74%` - [Link](https://app.codecov.io/gh/open-telemetry/opentelemetry-js-contrib/tree/main/plugins%2Fnode%2Fopentelemetry-instrumentation-mongodb)

This is because tests for MongoDB versions v3 and v4 are not included in the coverage report. 

- Runs v3 and v4 tests in the `test` command to make sure they are included in the coverage report


plugins/node/opentelemetry-instrumentation-mongodb/package.json
```diff
@@ -7,12 +7,15 @@
...
   "scripts": {
     "docker:start": "docker run -e MONGODB_DB=opentelemetry-tests -e MONGODB_PORT=27017 -e MONGODB_HOST=127.0.0.1 -p 27017:27017 --rm mongo",
-    "test": "npm run test-v5-v6",
-    "test-v3": "nyc mocha --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v3.test.ts'",
-    "test-v4": "nyc mocha --require '@opentelemetry/contrib-test-utils' 'test/mongodb-v4-v5-v6.metrics.test.ts' 'test/**/mongodb-v4.test.ts'",
-    "test-v5-v6": "nyc mocha --require '@opentelemetry/contrib-test-utils' 'test/mongodb-v4-v5-v6.metrics.test.ts' 'test/**/mongodb-v5-v6.test.ts'",
+    "test": "npm run test-v3 && npm run test-v4 && npm run test-v5-v6 && nyc merge .nyc_output ./coverage/coverage-final.json",
+    "test-v3": "tav mongodb 3.7.4 npm run test-v3-run",
+    "test-v4": "tav mongodb 4.17.0 npm run test-v4-run",
+    "test-v5-v6": "tav mongodb 6.8.0 npm run test-v5-v6-run",
```
david-luna, line 13: do you think is necessary to pin here the version? `pacakge.json` file already has the version fixed and if we want to update to a newer version we need to do it in both places.

david-luna:
I have my doubts about this PR. IMHO the wider the support range is the more tests with different versions we would need to do in order to aim for high coverage. We may end up doing a subset of `test-all-versions` 🤔

plugins/node/opentelemetry-instrumentation-mongodb/package.json
onurtemizkan: I just double checked, for the pinned version, it's not needed. 👍

onurtemizkan:
Thanks for the review @david-luna!

The alternative I thought was generating coverage on `test-all-versions` and merging reports in the end. But that seems a bit wider scale update if that's going to be applied to every package (there are just a few packages excluding specific versions as far as I can see).

david-luna:
> Thanks for the review @david-luna!
> 
> The alternative I thought was generating coverage on `test-all-versions` and merging reports in the end. But that seems a bit wider scale update if that's going to be applied to every package (there are just a few packages excluding specific versions as far as I can see).

@onurtemizkan we discussed this approach for coverage in the SIG. We accept the tradeoff of running 3 different tests to get the coverage report. Also
- locally you can get a different version of the package under test if `npm run test` fails in one of the steps.
- we will keep an eye on the CI to be sure is not impacting it significantly

Thanks for your contribution :)

david-luna (approved):

david-luna:
@onurtemizkan sorry for the delay on this PR. I was checking if it's possible to have the coverage in TAV tests. Instead of stalling this PR I'll try it myself in another instrumentation.

Issue #2680, opened 2025-01-27
@opentelemetry/instrumentation-express - incorrect route set to span, if middlewares are being used


RigoTamas:
### What version of OpenTelemetry are you using?

```json
{
    "@opentelemetry/api": "^1.9.0",
    "@opentelemetry/auto-instrumentations-node": "^0.55.3",
    "@opentelemetry/exporter-metrics-otlp-proto": "^0.57.1",
    "@opentelemetry/exporter-trace-otlp-proto": "^0.57.1",
    "@opentelemetry/instrumentation-express": "^0.47.0",
    "@opentelemetry/sdk-metrics": "^1.30.1",
    "@opentelemetry/sdk-node": "^0.57.1",
    "@opentelemetry/sdk-trace-node": "^1.30.1",
    "express": "4.21.1"
  }
```

### What version of Node are you using?

```sh
20.15.0
```

### What did you do?
This repository reproduces the bug:
https://github.com/RigoTamas/opentelemetry-express-instrumentation-span-name-bug

```typescript
// index.ts

import express, { NextFunction, Request, Response, Router } from "express";

const app = express();

const userMiddleware = (req: Request, res: Response, next: NextFunction) => {
  console.log("user middleware running");
  next();
};

const userMiddleware2 = (req: Request, res: Response, next: NextFunction) => {
  console.log("user middleware 2 running");
  next();
};

const userRouter = Router({ mergeParams: true });

userRouter.use(userMiddleware);

userRouter.use(userMiddleware2);

userRouter.get("/", (req: Request, res: Response, next: NextFunction) => {
  res.send({ name: "John" });
});

const v1Router = Router({ mergeParams: true });

v1Router.use("/user", userRouter);

app.use("/v1", v1Router);

app.listen(8000, () => {
  console.log("api listening on 8000");
});
```

```typescript
// instrumentation.ts
import { getNodeAutoInstrumentations } from "@opentelemetry/auto-instrumentations-node";
import { OTLPMetricExporter } from "@opentelemetry/exporter-metrics-otlp-http";
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-proto";
import { Resource } from "@opentelemetry/resources";
import { PeriodicExportingMetricReader } from "@opentelemetry/sdk-metrics";
import { NodeSDK } from "@opentelemetry/sdk-node";
import { ATTR_SERVICE_NAME } from "@opentelemetry/semantic-conventions";

[...]

PR #2682, opened 2025-01-27
fix(express): span name if middleware on nested router is used

RigoTamas:

- Fixes: #2680

- A flag is passed down in the `storeLayerPath` function, which tells the caller wether a new item has been added to the layer store. This is needed, because we only want to pop from the `req[_LAYERS_STORE_PROPERTY]` list if we added something to it. Popping from this list in case we did not push anything to it will result in popping a parent entry, which will skew the layers store. Hence this unnecessary popping will result in a malformed http.route tag.
- This can be reproduced by adding a middleware to a nested router.
- The nested router unit test has been updated to reflect this behavior. The unit test now breaks without this change, which reflects the need for this change.


bsiedleckiallegro:
That is a pretty significant improvement, good job!
@pkanal @JamieDanielson could you please take a look at the changes suggested by @RigoTamas ? Much appreciated :)

allegro-dbargowski:
Hey there, just wanted to share that I've tried out the changes from this PR by tweaking the OTEL libraries in node_modules for our app, and it sorted out the issues we were facing with `http_route` values in our Prometheus metrics. @pkanal @JamieDanielson, it would be awesome if you could take a look at this PR soon. Thanks a bunch!

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -33,15 +33,21 @@ import {
  * @param request The request where
  * @param [value] the value to push into the array
  */
-export const storeLayerPath = (request: PatchedRequest, value?: string) => {
+export const storeLayerPath = (
+  request: PatchedRequest,
+  value?: string
+): { isLayerPathStored: boolean } => {
```
raphael-theriault-swi, line 39: This is extremely minor but I can't really think of any other information that could be returned by this, so personally I'd rather see this skip the object and just return a boolean, the meaning of which can be documented in the docstring.

raphael-theriault-swi (approved):
Left some feedback but I don't feel it's important enough to warrant not approving. Thanks for the contribution !

RigoTamas:
Thanks for approving @raphael-theriault-swi! Who is part of the `open-telemetry/javascript-approvers` group? I see somebody from that group still needs to approve in order to merge this.

raphael-theriault-swi:
@open-telemetry/javascript-approvers ready for workflow approval

allegro-dbargowski:
Hey @raphael-theriault-swi, @pkanal, and @JamieDanielson! Just checking in to see if there's any chance this fix will be merged soon. We're using OTEL for monitoring and alerting based on Prometheus metrics for our app, which is critical for our on-call engineers. Right now, we're seeing issues with invalid `http_route` label values (which in turn makes differentiating endpoints impossible sometimes). If the merge will take a while, we'll consider using a forked version with this fix applied. Any updates would be greatly appreciated—thanks a bunch! :pray:

pichlermarc (commented):
(proxy-appoving since @raphael-theriault-swi has no write permissions on the repo)

pichlermarc (approved):

PR #2684, opened 2025-01-27
test(instrumentation-winston): Include all versions in coverage

onurtemizkan:

- Spin off from: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2681

- Runs `v1` to `v2` and `v3` tests in the test command to make sure they are included in the coverage report


blumamir (approved):
Thank you for adding this.
LGTM

Issue #2490, opened 2024-10-17
Platform-Specific Path Assertion Error: Windows vs. Unix Root Path Mismatch in Tests


Annosha:
### What version of OpenTelemetry are you using?
 "name": "opentelemetry-contrib",
  "version": "0.1.0",
OS: Windows

### What version of Node are you using?
v20.11.1

### What did you do?

 Cloned the repo to setup dev env for contribution
compiled the code without any warnings or errors.
![compile](https://github.com/user-attachments/assets/e482084e-3378-4ead-a3ce-0ae30f6b0e91)

But running `npm test` on root level as well as on module level of mentioned modules give errors:
![failed-test](https://github.com/user-attachments/assets/d91c182e-1c77-4d2c-b3ee-119060acc567)
Detailed log of errors for respective modules is attached below.
[instrumentation-cucumber-test-log.txt](https://github.com/user-attachments/files/17411289/instrumentation-cucumber-test-log.txt)
[instrumentation-fs-test-log.txt](https://github.com/user-attachments/files/17411301/instrumentation-fs-test-log.txt)

### What did you expect to see?
The tests must be modified to handle platform-specific tests

### What did you see instead?
Error due to platform-specific path mismatch. The test expects the root path 'C:\\' on Windows to be loosely equal to '/', the root path on Unix-based systems. This causes an AssertionError as the file system paths differ between operating systems.
![cucumber-test-error](https://github.com/user-attachments/assets/136db588-8497-4b93-b58e-27badf02bd8e)
![fs-failed-test](https://github.com/user-attachments/assets/ced5770c-0406-4706-af1d-cd0658423cee)

### Additional context
 
 The test does not account for Windows-specific path formats, leading to the failure. We can:

-  Check the OS type using `process.platform` and adjusting the expected value accordingly.

- Normalizing the paths to make them comparable across platforms.

PR #2694, opened 2025-02-04
test(cucumber): join all paths with path module

Ugzuzg:

- Addresses #2490 for cucumber instrumentation.

- Join all paths with `path.join`.


blumamir (approved):
Thanks @Ugzuzg

Issue #2696, opened 2025-02-05
Pino instrumentation hiding log lines when child logger configured at lower level than parent


OmriHarary:
### What version of OpenTelemetry are you using?

```
"@opentelemetry/sdk-node": "0.57.1",
"@opentelemetry/instrumentation-pino": "0.46.0",
```

### What version of Node are you using?

Tested with 18, 20, and 22

### What did you do?

With Pino instrumentation enabled, creating a logger at info level and then a child of it at debug level causes debug-level messages from the child not to be logged

Repro:
```
const {NodeSDK, tracing, logs} = require('@opentelemetry/sdk-node');
const {PinoInstrumentation} = require('@opentelemetry/instrumentation-pino');
const sdk = new NodeSDK({
  spanProcessor: new tracing.SimpleSpanProcessor(new tracing.ConsoleSpanExporter()),
  logRecordProcessor: new logs.SimpleLogRecordProcessor(new logs.ConsoleLogRecordExporter()),
  instrumentations: [new PinoInstrumentation()],
  resourceDetectors: [],
});
sdk.start();

const pino = require('pino');
const logger1 = pino({level: 'info'});
const logger2 = pino({level: 'debug'});

const child1 = logger1.child({}, {level: 'debug'});
const child2 = logger2.child({}, {level: 'debug'});

logger1.info('l1 info'); // should log
logger1.debug('l1 debug'); // should not log

logger2.info('l2 info'); // should log
logger2.debug('l2 debug'); // should log

child1.info('c1 info'); // should log
child1.debug('c1 debug'); // should log

child2.info('c2 info'); // should log
child2.debug('c2 debug'); // should log

const otherlogger = pino();
otherlogger.info('l1 = %s | l2 = %s | c1 = %s | c2 = %s', logger1.level, logger2.level, child1.level, child2.level);
```

### What did you expect to see?
(Actual results from running with `PinoInstrumentation` not registered or when registered with `disableLogCorrelation: true`)

```
{"level":30,"time":1738772867952,"pid":61376,"hostname":"CAKNWV959WHV","msg":"l1 info"}
{"level":30,"time":1738772867953,"pid":61376,"hostname":"CAKNWV959WHV","msg":"l2 info"}
{"level":20,"time":1738772867953,"pid":61376,"hostname":"CAKNWV959WHV","msg":"l2 debug"}
[...]

PR #2699, opened 2025-02-05
fix(instrumentation-pino): log-sending was losing records when the level was reduced

trentm:
Before this fix, if a Pino logger level was reduced (e.g. from 'info' to 'debug'),
then log.debug('msg') would *not* work. The same thing would happen if a child
logger was created at a lower-level than its parent.

This issue was that PinoInstrumentation is internally using a pino multistream
(https://getpino.io/#/docs/api?id=pino-multistream) and statically setting
the level of those streams to the current Logger level. If the logger level
was set lower, the multistream levels would end up filtering records
at that original higher level. The fix is to set the stream levels to
0 to never filter out records.

Closes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2696


trentm:
Cycle the instr-pino label to see if that triggers the TAV test run.

maryliag (approved):

PR #2705, opened 2025-02-11
refactor(resource-detector-gcp): migrate away from getEnv()

pichlermarc:

We're planning to remove `getEnv()` as it does not scale well when everyone has to update `@opentelemetry/core` to get their env vars. For `OTEL_*` there will be a replacement utility function. For all others, when the package does not need to be browser-compatible (which seems to be the case here) using `process.env` directly should be fine.

Refs https://github.com/open-telemetry/opentelemetry-js/pull/5443

This PR removes `getEnv()` with `process.env` and inlines the defaults from `@opentelemetry/core`'s `DEFAULT_ENVIRONMENT`

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
+++ b/detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
@@ -24,7 +24,6 @@ import {
   ResourceAttributes,
   IResource,
 } from '@opentelemetry/resources';
-import { getEnv } from '@opentelemetry/core';
 import {
   CLOUDPROVIDERVALUES_GCP,
   SEMRESATTRS_CLOUD_ACCOUNT_ID,
@@ -79,7 +78,7 @@ class GcpDetector implements DetectorSync {
     attributes[SEMRESATTRS_CLOUD_AVAILABILITY_ZONE] = zoneId;
     attributes[SEMRESATTRS_CLOUD_PROVIDER] = CLOUDPROVIDERVALUES_GCP;
 
-    if (getEnv().KUBERNETES_SERVICE_HOST)
+    if (process.env.KUBERNETES_SERVICE_HOST)
       this._addK8sAttributes(attributes, clusterName);
 
     return attributes;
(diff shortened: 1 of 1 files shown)


detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
```diff
@@ -79,7 +79,7 @@ class GcpDetector implements DetectorSync {
     attributes[SEMRESATTRS_CLOUD_AVAILABILITY_ZONE] = zoneId;
     attributes[SEMRESATTRS_CLOUD_PROVIDER] = CLOUDPROVIDERVALUES_GCP;
 
-    if (getEnv().KUBERNETES_SERVICE_HOST)
+    if (process.env.KUBERNETES_SERVICE_HOST)
```
pichlermarc, line 82: **Note for reviewers:** this previously was an empty string when it was unset - now it's `undefined`. The semantics should be the same other than that.

detectors/node/opentelemetry-resource-detector-gcp/src/detectors/GcpDetector.ts
```diff
@@ -90,12 +90,10 @@ class GcpDetector implements DetectorSync {
...
   ): void {
-    const env = getEnv();
-
     attributes[SEMRESATTRS_K8S_CLUSTER_NAME] = clusterName;
-    attributes[SEMRESATTRS_K8S_NAMESPACE_NAME] = env.NAMESPACE;
-    attributes[SEMRESATTRS_K8S_POD_NAME] = env.HOSTNAME;
-    attributes[SEMRESATTRS_CONTAINER_NAME] = env.CONTAINER_NAME;
+    attributes[SEMRESATTRS_K8S_NAMESPACE_NAME] = process.env.NAMESPACE ?? '';
+    attributes[SEMRESATTRS_K8S_POD_NAME] = process.env.HOSTNAME ?? '';
+    attributes[SEMRESATTRS_CONTAINER_NAME] = process.env.CONTAINER_NAME ?? '';
```
pichlermarc, line 96: **Note for reviewers:**

See here for previous defaults:
- `NAMESPACE` ->
https://github.com/open-telemetry/opentelemetry-js/blob/15ba2d7da3bd6663ee6008e189f004bab93ff6fe/packages/opentelemetry-core/src/utils/environment.ts#L164
- `HOSTNAME` -> https://github.com/open-telemetry/opentelemetry-js/blob/15ba2d7da3bd6663ee6008e189f004bab93ff6fe/packages/opentelemetry-core/src/utils/environment.ts#L162
- `CONTAINER_NAME` -> https://github.com/open-telemetry/opentelemetry-js/blob/15ba2d7da3bd6663ee6008e189f004bab93ff6fe/packages/opentelemetry-core/src/utils/environment.ts#L88

trentm (approved):

PR #2708, opened 2025-02-11
refactor(auto-instrumentations-node): migrate away from getEnv()

pichlermarc:

We're planning to remove `getEnv()` as it does not scale well when everyone has to update `@opentelemetry/core` to get their env vars. For `OTEL_*` there will be a replacement utility function, but until this one is available, we can use `process.env` directly.

Refs https://github.com/open-telemetry/opentelemetry-js/pull/5443

Moves code for parsing `OTEL_LOG_LEVEL` to `@opentelemetry/auto-instrumentations-node` here. It may be changed later as we add similar code to `@opentelemetry/sdk-node`.

### DIFF
--- a/metapackages/auto-instrumentations-node/src/register.ts
+++ b/metapackages/auto-instrumentations-node/src/register.ts
@@ -16,14 +16,12 @@
 import * as opentelemetry from '@opentelemetry/sdk-node';
 import { diag, DiagConsoleLogger } from '@opentelemetry/api';
 import {
+  getLogLevelFromEnv,
   getNodeAutoInstrumentations,
   getResourceDetectorsFromEnv,
 } from './utils';
 
-diag.setLogger(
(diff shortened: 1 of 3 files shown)


pichlermarc:
> I'm a little more hesitant on this change, given that this package, sdk-node, and https://github.com/open-telemetry/opentelemetry-lambda/blob/faa6c56645b77d58f170f57ab016e845d815c0a8/nodejs/packages/layer/src/wrapper.ts#L470 would all need this change.
> 
> It might still be the easiest transition away from `getEnv()`, so I'm approving.

My original draft of the PR in core actually had a utility function to do this. I'll look into adding it back when I open the PR to update `@opentelemetry/sdk-node` - if there's three places in which this is used, I think there's a good case to add a utility like this. 👍

blumamir (approved):

maryliag (approved):

PR #2709, opened 2025-02-11
chore: restore lint:deps checking; fix some deps issues

trentm:
Fixes these 'lint:deps' hits:

```
Unused dependencies (2)
@opentelemetry/otlp-transformer  packages/opentelemetry-test-utils/package.json
@opentelemetry/core              plugins/web/opentelemetry-plugin-react-load/package.json
Unlisted dependencies (1)
@opentelemetry/instrumentation  plugins/web/opentelemetry-plugin-react-load/src/BaseOpenTelemetryComponent.ts
```

`lint:deps` was accidentally removed in a merge-from-main in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2493.


pichlermarc (approved):
thanks 👍 
Sorry - I should've checked that. In my mind `@opentelemetry/instrumentation` was a "safe" replacement as all of the instrumentation packages usually depend on it, but I was not considering that the react-load one is not an instrumentation package in that sense.

PR #2718, opened 2025-02-16
fix(aws-sdk): use `custom` client_context prop

smoke:

In every other language the respected property carrying the context is `custom` e.g.
- https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/9a7f7125ab3f77c47a46be524f29e150536fba6f/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/LambdaImpl.java#L41

- rename `Custom` to `custom` property that is used as context carrier for AWS Lambda Client Context


smoke:
@blumamir @jj22ee @trivikr Sorry to bother, please let me know if you need anything more on this PR. 

It is really sad that AWS Documentation is misleading https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html
The documentation is wrong instead of
`Custom – Custom values that are set by the client application.`
it should be
`custom – Custom values that are set by the client application.`

I have just filed a feedback form, hopefully to correct it.
image

smoke:
@blumamir @jj22ee @trivikr The AWS documentation is now updated https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html
image

Can you please merge and release this PR?

YevhenHorbachov (approved):

jj22ee (approved):

jj22ee:
Thanks! You may need to merge with main to get the remaining checks to run.

jj22ee:
Also @smoke, did `injectLambdaPropagationContext` never work properly before? If it worked in some capacity, I was wondering if this should be labelled as a breaking change.

smoke:
> Also @smoke, did `injectLambdaPropagationContext` never work properly before? If it worked in some capacity, I was wondering if this should be labelled as a breaking change.

It never worked, probably.

pichlermarc (approved):

PR #2732, opened 2025-02-25
fix(instrumentation-mysql2)!:Missing masking of sql queries

gflachs:
Based on Issue #1758  and my [comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1857#issuecomment-2668028719) in the PR #1857 this PR is a quickfix for missing masking from sql queries. 

Since OpenTelemetry instrumentation should remain as lightweight as possible, introducing a full SQL parser would be too heavy. As a result, the default masking approach is intentionally simplistic. However, users can define their own `maskStatementHook`, allowing them to leverage a SQL parser or other custom logic for more advanced masking if needed.

Additionally, some use cases may require no masking at all. To accommodate this, the `maskStatement` setting allows users to disable masking entirely, preserving the original behavior of the MySQL2Instrumentation as it was before.

This PR extends MySQL2InstrumentationConfig with two new options:

* `maskStatementHook` → A customizable hook for masking SQL statements (optional). Default:
 ```ts
  return query.replace(/\b\d+\b/g, '?').replace(/(["'])(?:(?=(\\?))\2.)*?\1/g, '?')
 ```
* `maskStatement` → A boolean flag (default: true) that determines whether maskStatementHook is applied at all.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
@@ -23,6 +23,7 @@ import {
   SEMATTRS_NET_PEER_PORT,
 } from '@opentelemetry/semantic-conventions';
 import type * as mysqlTypes from 'mysql2';
+import { MySQL2InstrumentationQueryMaskingHook } from './types';
 
 type formatType = typeof mysqlTypes.format;
 
@@ -107,20 +108,39 @@ function getJDBCString(
 export function getDbStatement(
   query: string | Query | QueryOptions,
   format?: formatType,
-  values?: any[]
+  values?: any[],
+  maskStatement: boolean = true,
+  maskStatementHook: MySQL2InstrumentationQueryMaskingHook = defaultMaskingHook
 ): string {
-  if (!format) {
-    return typeof query === 'string' ? query : query.sql;
+  let formattedQuery: string;
+  try {
+    if (!format) {
+      formattedQuery = typeof query === 'string' ? query : query.sql;
+    } else if (typeof query === 'string') {
+      formattedQuery = values ? format(query, values) : query;
+    } else {
+      // According to https://github.com/mysqljs/mysql#performing-queries
+      // The values argument will override the values in the option object.
+      formattedQuery =
+        values || (query as QueryOptions).values
+          ? format(query.sql, values || (query as QueryOptions).values)
+          : query.sql;
+    }
+
+    return maskStatement
+      ? values
+        ? formattedQuery
+        : maskStatementHook(formattedQuery)
+      : formattedQuery;
+  } catch (e) {
+    return 'Could not determine the query due to an error in masking or formatting';
   }
-  if (typeof query === 'string') {
-    return values ? format(query, values) : query;
-  } else {
-    // According to https://github.com/mysqljs/mysql#performing-queries
-    // The values argument will override the values in the option object.
-    return values || (query as QueryOptions).values
-      ? format(query.sql, values || (query as QueryOptions).values)
-      : query.sql;
-  }
+}
+
+function defaultMaskingHook(query: string): string {
+  return query
+    .replace(/\b\d+\b/g, '?')
+    .replace(/(["'])(?:(?=(\\?))\2.)*?\1/g, '?');
 }
 
 /**
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/instrumentation.ts
@@ -137,13 +137,20 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
         } else if (arguments[2]) {
           values = [_valuesOrCallback];
         }
-
+        const { maskStatement, maskStatementHook, responseHook } =
+          thisPlugin.getConfig();
         const span = thisPlugin.tracer.startSpan(getSpanName(query), {
           kind: api.SpanKind.CLIENT,
           attributes: {
             ...MySQL2Instrumentation.COMMON_ATTRIBUTES,
             ...getConnectionAttributes(this.config),
-            [SEMATTRS_DB_STATEMENT]: getDbStatement(query, format, values),
+            [SEMATTRS_DB_STATEMENT]: getDbStatement(
+              query,
+              format,
+              values,
+              maskStatement,
+              maskStatementHook
+            ),
           },
         });
 
@@ -166,7 +173,6 @@ export class MySQL2Instrumentation extends InstrumentationBase<MySQL2Instrumenta
               message: err.message,
             });
           } else {
-            const { responseHook } = thisPlugin.getConfig();
             if (typeof responseHook === 'function') {
               safeExecuteInTheMiddle(
                 () => {
--- a/plugins/node/opentelemetry-instrumentation-mysql2/src/types.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/src/types.ts
@@ -25,7 +25,26 @@ export interface MySQL2InstrumentationExecutionResponseHook {
   (span: Span, responseHookInfo: MySQL2ResponseHookInformation): void;
 }
 
+export interface MySQL2InstrumentationQueryMaskingHook {
+  (query: string): string;
+}
+
 export interface MySQL2InstrumentationConfig extends InstrumentationConfig {
+  /**
+   * If true, the query will be masked before setting it as a span attribute, using the {@link maskStatementHook}.
+   *
+   * @default true
+   * @see maskStatementHook
+   */
+  maskStatement?: boolean;
+
+  /**
+   * Hook that allows masking the query string before setting it as span attribute.
+   *
+   * @default (query: string) => query.replace(/\b\d+\b/g, '?').replace(/(["'])(?:(?=(\\?))\2.)*?\1/g, '?')
+   */
+  maskStatementHook?: MySQL2InstrumentationQueryMaskingHook;
+
   /**
    * Hook that allows adding custom span attributes based on the data
    * returned MySQL2 queries.
--- a/plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
@@ -146,6 +146,9 @@ describe('mysql2', () => {
       contextManager = new AsyncHooksContextManager().enable();
       context.setGlobalContextManager(contextManager);
       instrumentation.setTracerProvider(provider);
+      instrumentation.setConfig({
+        maskStatement: false,
+      });
       instrumentation.enable();
       connection = createConnection({
         port,
@@ -1118,6 +1121,7 @@ describe('mysql2', () => {
             responseHook: (span, responseHookInfo) => {
               throw new Error('random failure!');
             },
+            maskStatement: false,
           };
           instrumentation.setConfig(config);
         });
@@ -1145,6 +1149,7 @@ describe('mysql2', () => {
                 JSON.stringify(responseHookInfo.queryResults)
               );
             },
+            maskStatement: false,
           };
           instrumentation.setConfig(config);
         });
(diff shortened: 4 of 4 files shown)


pichlermarc:
Please disregard the previous comment by the GH action - this component actually maintained by @raphael-theriault-swi now, we forgot to update the workflow. 

I'll open a PR to make sure this comment does not pop up again for this instrumentation.

plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
```diff
@@ -107,22 +108,41 @@ function getJDBCString(
...
+    return formattedQuery;
+  } catch (e) {
+    return 'Could not determine the query due to an error in masking or formatting';
   }
 }
 
+function defaultMaskingHook(query: string): string {
+  return query
+    .replace(/\b\d+\b/g, '?')
+    .replace(/(["'])(?:(?=(\\?))\2.)*?\1/g, '?');
```
raphael-theriault-swi, line 143: I'm no regex expert so this might be something obvious, but what's the reasoning for using a lookahead here rather than just a literal backslash ?

gflachs: Me neither, but at least my understanding of regex is that the lookahead ((?=(\\?))) is used to detect an optional backslash without consuming the next character right away. This ensures that escaped quotes (like \") inside a string don’t cause the pattern to end prematurely. By capturing the optional backslash in a group (\2) and then matching it with \2., both escaped and regular characters are handled consistently. Using a direct backslash match (e.g., \\?) would make it harder to distinguish between cases where a backslash is present or not, leading to a more complex and less readable regex.

raphael-theriault-swi: Makes sense. Could you maybe just add a comment to describe the behaviour of the default hook so future contributors don't have to parse the regex to understand what's happening ?

plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
```diff
@@ -107,22 +108,41 @@ function getJDBCString(
...
+      // The values argument will override the values in the option object.
+      formattedQuery =
+        values || (query as QueryOptions).values
+          ? format(query.sql, values || (query as QueryOptions).values)
+          : query.sql;
+    }
+    return formattedQuery;
+  } catch (e) {
+    return 'Could not determine the query due to an error in masking or formatting';
   }
```
raphael-theriault-swi, line 137: I find this series of overlapping conditionals to be really hard to follow, and I feel like it could definitely be simplified. Something like this maybe ? I'm also not sure there are any reasons to keep the formatting logic at all apart from backwards compatibility since it does the exact opposite of what the spec asks for.
```suggestion
  const [querySql, queryValues] = typeof query === 'string'
    ? [query, values]
    // According to https://github.com/mysqljs/mysql#performing-queries
    // The values argument will override the values in the option object.
    : [query.sql, values || query.values];
  try {
    if (maskStatement) {
      return maskStatementHook(querySql);
    } else if (format && values) {
      return format(querySql, queryValues);
    } else {
      return querySql;
    }
  } catch (e) {
    return 'Could not determine the query due to an error in masking or formatting';
  }
```

raphael-theriault-swi, line 137: I'm assuming the rationale here is that we don't want to accidentally leak information by including the error message in the attribute value ?

raphael-theriault-swi (changes requested):

plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
gflachs: Excatly. Since there could be some stacktrace with the actual string (like 'abc' ist not valid), its better to have this generic errormessage. Also, since the user can provide its own maskingHook, we never know, if there is any problematic code, which forces the leakage of information trough stacktrace. Therefore I thought this is the best way, to prevent this from happening. I can still adjust the errormessage, if you think its not clear enough :)

gflachs: I adjusted it, but added the `hasValues` Typecheck as well, since Typescript was complaining otherwise (Query don't has the properties values).

The formatting is actual needed, since its replacing the placeholders with the real values. If someone is setting maskStatement to 'false' I assume, the intent is, that all queries are pushed unmasked into the trace, not only already unmasked Queries. Therefore we need to resolve the parameterized Queries as well. I can remove it, if you think, that this is not beneficial at all.

plugins/node/opentelemetry-instrumentation-mysql2/src/utils.ts
gflachs: I added the following comment as js-doc at the defaultMaskingHook

```
/**
 * Replaces numeric values and quoted strings in the query with placeholders ('?').
 *
 * - `\b\d+\b`: Matches whole numbers (integers) and replaces them with '?'.
 * - `(["'])(?:(?=(\\?))\2.)*?\1`:
 *   - Matches quoted strings (both single `'` and double `"` quotes).
 *   - Uses a lookahead `(?=(\\?))` to detect an optional backslash without consuming it immediately.
 *   - Captures the optional backslash `\2` and ensures escaped quotes inside the string are handled correctly.
 *   - Ensures that only complete quoted strings are replaced with '?'.
 *
 * This prevents accidental replacement of escaped quotes within strings and ensures that the
 * query structure remains intact while masking sensitive data.
 */
```

raphael-theriault-swi (approved):
Looks good, left some very minor suggestions but I'm happy with this either way !

gflachs:
> Looks good, left some very minor suggestions but I'm happy with this either way !

Thanks :) Github still claims, that a Review from a Maintainer ist missing :thinking: Do I have to do soemthing else, or is it still the Bug mentioned by @pichlermarc ?

raphael-theriault-swi:
Approvers are still the ones doing the workflow approval and merging, so this is still waiting on Marc or someone else to merge yes.

pichlermarc:
Hmm, looks like the lint  step is failing.
 @gflachs, please run `npm run lint:fix` and commit the results. Thanks! 🙂

raphael-theriault-swi:
Would appreciate the changes to `hasValues` at the same time since another commit is needed.

gflachs:
> Would appreciate the changes to `hasValues` at the same time since another commit is needed.

No problem, I'll do both tomorrow morning (MEZ, UTC +1)

gflachs:
> Would appreciate the changes to `hasValues` at the same time since another commit is needed.

I added it in the current commit :)

maryliag:
Hi @gflachs , thank you for working on this.
I have concerns with the performance of the masking itself. Using regex and replacements can be very costly, specially in large queries (I have seen queries with over 20k characters). Can you provide some benchmark comparing with and without your changes, considering a heavy workload with big queries?

Also, since this could cause a performance issue, I don't think we should have `maskStatement` as default `true`

gflachs:
> Hi @gflachs , thank you for working on this. I have concerns with the performance of the masking itself. Using regex and replacements can be very costly, specially in large queries (I have seen queries with over 20k characters). Can you provide some benchmark comparing with and without your changes, considering a heavy workload with big queries?
> 
> Also, since this could cause a performance issue, I don't think we should have `maskStatement` as default `true`

Hey @maryliag,
Thanks for your reply and for raising your concerns about the performance of the masking itself—I totally get where you’re coming from, though I don’t fully share the same worry since it’s always a tradeoff between security/privacy and performance.

I’ve put together a quick benchmark to dig into this:

```js
const { performance } = require('perf_hooks');
const fs = require('fs');
const os = require('os');

function getSystemInfo() {
    let cpuLimit = "Unlimited";
    let memLimit = "Unlimited";

    try {
        const cpuData = fs.readFileSync('/sys/fs/cgroup/cpu.max', 'utf8').trim().split(' ');
        if (cpuData[0] !== 'max') {
            cpuLimit = (parseInt(cpuData[0]) / 1000) + " ms CPU time per 100ms";
        }
    } catch (err) {
        console.warn("⚠️  Could not retrieve CPU limit.");
    }

    try {
        const memData = fs.readFileSync('/sys/fs/cgroup/memory.max', 'utf8').trim();
        if (memData !== 'max') {
            memLimit = (parseInt(memData) / (1024 ** 2)).toFixed(2) + " MB";
        }
    } catch (err) {
        console.warn("⚠️  Could not retrieve memory limit.");
    }

    return {
        cpu: os.cpus()[0].model,
        cores: os.cpus().length,
        totalMemGB: (os.totalmem() / (1024 ** 3)).toFixed(2) + " GB",
        platform: os.platform(),
        release: os.release(),
        containerCPU: cpuLimit,
        containerMemory: memLimit
    };
}

function regexMasking(query) {
    return query.replace(/\b\d+\b/g, '?').replace(/(["'])(?:(?=(\\?))\2.)*?\1/g, '?');
}

function generateLargeSQLQuery(size) {
    let query = "SELECT id, name, age FROM users WHERE age = 25 AND name = 'John Doe' ";
    while (query.length < size) {
        query += "UNION SELECT id, 'Random Name', 30 FROM users WHERE id = " + Math.floor(Math.random() * 1000) + " ";
    }
    return query.slice(0, size);
}

function benchmarkFunction(fn, query, iterations = 3) {
    let totalTime = 0;
    for (let i = 0; i < iterations; i++) {
        let start = performance.now();
        fn(query);
        totalTime += (performance.now() - start);
    }
    return (totalTime / iterations).toFixed(2);
}

function benchmark() {
    const sizes = [100, 1000, 10000, 50000, 100000, 1000000];

    console.log("\n🔍 **System Specifications:**\n", JSON.stringify(getSystemInfo(), null, 2));
    console.log("\n🚀 **Benchmark: SQL Masking Performance**");
    console.log("| SQL Size | Regex (ms) |");
    console.log("|----------|------------|");

    for (let size of sizes) {
        const query = generateLargeSQLQuery(size);

        let regexTime = benchmarkFunction(regexMasking, query);

        console.log(`| ${size.toLocaleString()} | ${regexTime} |`);
    }
}

benchmark();

```

Benchmarking within a docker-container to simulate different server-enviroments:

docker run --rm -it --cpus="0.25" --memory="128m" benchmark_sql:latest 

🔍 **System Specifications:**
 {
  "cpu": "AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics",
  "cores": 16,
  "totalMemGB": "30.00 GB",
  "platform": "linux",
  "release": "6.13.5-200.fc41.x86_64",
  "containerCPU": "25 ms CPU time per 100ms",
  "containerMemory": "128.00 MB"
}

🚀 **Benchmark: SQL Masking Performance**
| SQL Size | Regex (ms) |
|----------|------------|
| 100 | 0.07 |
| 1,000 | 0.01 |
| 10,000 | 0.08 |
| 50,000 | 1.11 |
| 100,000 | 0.41 |
| 1,000,000 | 7.56 |

---

 docker run --rm -it --cpus="0.1" --memory="50m" benchmark_sql:latest  

🔍 **System Specifications:**
 {
  "cpu": "AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics",
  "cores": 16,
  "totalMemGB": "30.00 GB",
  "platform": "linux",
  "release": "6.13.5-200.fc41.x86_64",
  "containerCPU": "10 ms CPU time per 100ms",
  "containerMemory": "50.00 MB"
}

🚀 **Benchmark: SQL Masking Performance**
| SQL Size | Regex (ms) |
|----------|------------|
| 100 | 0.06 |
| 1,000 | 0.01 |
| 10,000 | 0.07 |
| 50,000 | 0.48 |
| 100,000 | 0.38 |
| 1,000,000 | 67.37 |

---

 docker run --rm -it --cpus="0.5" --memory="1g" benchmark_sql:latest 

🔍 **System Specifications:**
 {
  "cpu": "AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics",
  "cores": 16,
  "totalMemGB": "30.00 GB",
  "platform": "linux",
  "release": "6.13.5-200.fc41.x86_64",
  "containerCPU": "50 ms CPU time per 100ms",
  "containerMemory": "1024.00 MB"
}

🚀 **Benchmark: SQL Masking Performance**
| SQL Size | Regex (ms) |
|----------|------------|
| 100 | 0.03 |
| 1,000 | 0.01 |
| 10,000 | 0.10 |
| 50,000 | 0.56 |
| 100,000 | 0.43 |
| 1,000,000 | 5.67 |


## My Take
I hear your concern about regex and replacements being costly, especially with large queries like the 20k+ character ones you mentioned. The benchmarks show that for smaller queries (up to 10k characters), the impact is negligible (<0.1 ms). For larger queries (e.g., 1M characters), it can climb to 5–67 ms depending on resources, though a 20k-character query would likely fall in the 0.1–0.2 ms range—still fast, but potentially noticeable under heavy load.

That said, I’d like to push back a bit on setting maskStatement to false by default. Data security and privacy are critical, and masking sensitive data (like numbers and strings) in outputs is a key safeguard. The performance hit, even in extreme cases, seems manageable in most realistic scenarios, and I’d argue the security benefits outweigh the cost. Turning it off by default could expose sensitive data unnecessarily, which feels riskier than a slight performance dip.

maryliag:
Thank you for performing the tests. Let me clarify my suggestion, I think the default should be:
- not send this data, since I agree that the security risk is a top priority

then they have the option to:
- enable to send data and with that we would have the mask of data

in the future we want to take advantage of the config file, so there people will be able to select one of the options:
1) I'm already sending data without the values, so you can send as is
2) I'm sending confidential data, so I want to remove everything and not mask, because my system cannot afford performance hits
3) I'm sending confidential data and I want it to be masked, since my system can afford performance hits

we can even have other performance improvements, such as if query is up to X characters mask it, otherwise remove completely, or even cut to X characters and then mask it

My main concern at the moment is to make this behaviour the default, and a lot of people use ORMs and those tend to create really huge queries

raphael-theriault-swi:
I'd be curious to see the delta between the masking and `sqlstring`'s format (which is currently used by default), given the latter also uses regex on top of some other non trivial logic. Don't disagree with the performance concerns but I have a feeling masking might not make it worse than it currently is.

maryliag:
It's not ready to merge.
Current behaviour: the value os `maskStatement` is set to default `true` and sanitization is being made by default. 

Changes needed: the default value should be `false` and no information should be sent on that case. Then IF the value is true, sanitization should be done and the query text can be sent

gflachs:
Please read my comment for the purpose of the maskStatement flag :)

maryliag:
> Turning it off by default could expose sensitive data unnecessarily

This is why I said that the it needs to be aligned, when the value is `false`, nothing it sent (so we don't potentially send sensitive information), and when the value is `true`, you can do the masking + sending the data.

dyladan:
@gflachs I think this is quite close, just needs the off-by-default comment addressed

david-luna:
heads up @gflachs 

A [recent PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2928) that changes the folder structure has been merged. This affects your PR so you need to sync with the main branch.

Happy to help if you need it :)

maryliag (approved):

maryliag:
I updated the default value of maskStatement to false, and made the correspondent changes in the doc and PR description.
That was the only remaining change for this PR, and it should be good to merge now

PR #2751, opened 2025-03-10
feat(web-exception-instrumentation): Add instrumentation for web exceptions

pkanal:
Currently, there's no standardized way to capture and monitor unhandled exceptions and promise rejections in web applications using OpenTelemetry. This makes it difficult for developers to:
- Track runtime errors in production applications
- Understand the frequency and types of errors occurring
- Collect detailed error information including stack traces
- Add custom attributes to error events

- Captures both synchronous errors and unhandled promise rejections
- Records error name, message, and stack trace using OpenTelemetry semantic conventions
- Supports custom attribute callback for extending error information
- Uses the OpenTelemetry Events API for lightweight error reporting
- Maintains proper cleanup of event listeners

## Open questions
- This instrumentation probably shouldn't rely on the events API directly, should the instrumentation class also expose an events method like it does for tracing and logging?
- There is also a dependency on types from the `api` and `api-logs` package, is that okay or should they be redefined?


plugins/web/opentelemetry-instrumentation-web-exception/src/instrumentation.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.instrumentationName,
+        this.instrumentationVersion
+      );
+
+      const customAttributes = this.applyCustomAttributes
+        ? this.applyCustomAttributes(error)
+        : {};
+
+      eventLogger.emit({
+        name: 'exception',
```
MSNev, line 74: Event names should be namespaced so just `exception` doesn't actually match the semantic conventions.

Please work with @Karlie-777 and her PR around defining this exception https://github.com/open-telemetry/semantic-conventions/pull/1942. I agree that using the 3 top level attributes makes sense for these and I've left some comments in the link PR as well.

There are a few other properties that we should also be setting (if possible), which does mean attempting to parse out the exception on the client.

plugins/web/opentelemetry-instrumentation-web-exception/src/instrumentation.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (error) {
+      const message = error.message;
+      const type = error.name;
+      const errorAttributes = {
+        [ATTR_EXCEPTION_TYPE]: type,
+        [ATTR_EXCEPTION_MESSAGE]: message,
+        [ATTR_EXCEPTION_STACKTRACE]: error.stack,
+      };
+
+      const eventLogger = events.getEventLogger(
```
MSNev, line 64: We should also be including the "parsed" filename as well as the current url (hosting page), I also left a [comment in](https://github.com/open-telemetry/semantic-conventions/pull/1942/files#r1989711556) that this should be part of the client exception

MSNev: Note: if we end up using the `url.full` then to [comply with this PR](https://github.com/open-telemetry/semantic-conventions/pull/971) it MUST be redacted and I don't believe that the JS repo currently has any "helpers" to construct a redacted `url.full`.

MSNev: Additional comment, while it would be "nice" to just say that the URL is in the "resources" for SPA's it really needs to be mutable which resources can't currently be 😢

plugins/web/opentelemetry-instrumentation-web-exception/src/instrumentation.ts
martinkuba: Could we be consistent with the [existing semantic conventions for span events](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md#exception-event) (where the event name is "exception")? Or, do we need to define exception event for each separate domain (e.g. "browser.exception", "android.exception" etc.)?

pkanal: I've changed it to `exception` for now since we're not creating any client specific semantic conventions for errors yet. We can change it in the future if this pattern develops.

plugins/web/opentelemetry-instrumentation-web-exception/src/instrumentation.ts
pkanal: For now, since we're sticking with the existing semantic conventions instead of client specific ones I haven't included the parsed file name or URL. I think we can add the URL in a future PR and add some core helpers to do that separately.

plugins/web/opentelemetry-instrumentation-web-exception/src/instrumentation.ts
```diff
@@ -0,0 +1,93 @@
...
+  readonly applyCustomAttributes?: (error: Error) => Attributes;
+  constructor(config: GlobalErrorsInstrumentationConfig = {}) {
+    super('@opentelemetry/instrumentation-web-exception', '0.0.1', config);
+
+    this.applyCustomAttributes = config.applyCustomAttributes;
+  }
+
+  init() {}
+
+  onError = (event: ErrorEvent | PromiseRejectionEvent) => {
```
martinkuba, line 53:
```suggestion
  onError (event: ErrorEvent | PromiseRejectionEvent) {
```
When defined as a variable, then `this.onError` is undefined in the enable() function.

plugins/web/opentelemetry-instrumentation-web-exception/src/instrumentation.ts
```diff
@@ -0,0 +1,93 @@
...
+    const error: Error | undefined =
+      'reason' in event ? event.reason : event.error;
+    if (error) {
+      const message = error.message;
```
martinkuba, line 57: `error` can be a string. In that case, perhaps set the message to be the string and omit type and stack trace.

plugins/web/opentelemetry-instrumentation-web-exception/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+```bash
+npm install --save @opentelemetry/instrumentation-web-exception
+```
+
+## Usage
+
+```typescript
+import { LoggerProvider } from '@opentelemetry/sdk-logs';
+import { EventLoggerProvider, events } from '@opentelemetry/api-events';
```
martinkuba, line 18: The API has just the interface for EventLoggerProvider. Import the class from the Events SDK.

plugins/web/opentelemetry-instrumentation-web-exception/README.md
```diff
@@ -0,0 +1,100 @@
...
+import { WebExceptionInstrumentation } from '@opentelemetry/instrumentation-web-exception';
+import { registerInstrumentations } from '@opentelemetry/instrumentation';
+
+// Set up the logger provider and event logger
+const loggerProvider = new LoggerProvider();
```
martinkuba, line 23: Consider adding a processor to make the example functional:
```
loggerProvider.addLogRecordProcessor(new SimpleLogRecordProcessor(new ConsoleLogRecordExporter())); 
```

martinkuba (changes requested):
Left a few comments

pichlermarc (changes requested):
Hi @pkanal - FYI, we require two component owners (members of the org) that are responsible for PR reviews and issue triage sign up in `.github/component-owners.yml` before merging any new instrumentations.

martinkuba:
I can volunteer to be the second component owner.

dyladan:
@pkanal can you please add the component owners and address martin's comments?

david-luna (commented):
Hi,

IIUC this instrumentation is included in Browser SIG Phase 1 as described [here](https://github.com/open-telemetry/community/blob/main/projects/browser-phase-1.md#step-2-core-instrumentation). So I guess it will remain in this state for a while or at leats get a new round of reviews.

In any case I would give a heads up on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2920 which will change the structure of this repository and place all packages within a `packages` folder. The instrumentation name follows what was commented on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223#issuecomment-2153093042 so its a good start. I think it would be easier to let https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2920 merge 1st and remap this changes (move files) to `pacakges/instrumentation-web-exception` folder.

pichlermarc:
> Hi @pkanal - FYI, we require two component owners (members of the org) that are responsible for PR reviews and issue triage sign up in `.github/component-owners.yml` before merging any new instrumentations.

Hi - seem like the component owners file is still not updated - this is a hard requirement for us that needs to be addressed before we can merge this PR.

plugins/web/opentelemetry-instrumentation-web-exception/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "@types/sinon": "17.0.4",
+    "@web/dev-server-esbuild": "^1.0.1",
+    "@web/dev-server-rollup": "^0.6.1",
+    "@web/test-runner": "^0.18.0",
+    "chai": "^4.3.10",
+    "sinon": "15.2.0",
+    "typescript": "^5.0.4"
+  },
+  "dependencies": {
+    "@opentelemetry/api-events": "^0.57.2",
```
martinkuba: The Events API and SDK have been removed. This PR should be updated to use just the Log API.

martinkuba (changes requested):

wolfgangcodes:
Hey folks! As mentioned at the SIG, I'll work with @pkanal on addressing these comments and move this forward.

packages/instrumentation-web-exception/.eslintignore
overbalance, line 1: You can safely delete this file.

packages/instrumentation-web-exception/.eslintrc.js
overbalance, line 1: You can safely delete this file.

packages/instrumentation-web-exception/package.json
```diff
@@ -0,0 +1,82 @@
...
+  "esnext": "build/esnext/index.js",
+  "types": "build/src/index.d.ts",
+  "repository": "open-telemetry/opentelemetry-js-contrib",
+  "scripts": {
+    "clean": "tsc --build --clean tsconfig.json tsconfig.esm.json tsconfig.esnext.json",
+    "compile:with-dependencies": "nx run-many -t compile -p @opentelemetry/instrumentation-web-exception",
+    "compile": "tsc --build tsconfig.json tsconfig.esm.json tsconfig.esnext.json",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "eslint . --ext .ts --fix",
+    "lint:readme": "node ../../scripts/lint-readme.js",
```
overbalance: These are now handled at the root of the repo.
```suggestion
```

overbalance (commented):
1. I'm not sure what version new instrumentation should have. Do we start out with 0.0.1?
2. I don't love prefixing the name with Web but I know that is a longer conversation. Should we do that now or not? Or mark this as experimental in case we decide to remove Web/Browser prefixes as instrumentations migrates to the browser repo?

wolfgangcodes:
Thanks for Reviewing @overbalance 
> 1. I'm not sure what version new instrumentation should have. Do we start out with 0.0.1?

I didn't see any guidance here. Maybe 0.1.0? I don't have strong opinions, other than it's not a 1,0, since it's still experimental to point 2 below.

> 2. I don't love prefixing the name with Web but I know that is a longer conversation. Should we do that now or not? Or mark this as experimental in case we decide to remove Web/Browser prefixes as instrumentations migrates to the browser repo?

I think it's experimental by default unless flagged otherwise, but we could make this explicit. I'm in favor of merging this if the implementation is acceptable and we can discuss the browser naming conventions, prefixes and disambiguation more wholistically at Browser SIG?

packages/instrumentation-web-exception/src/instrumentation.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export class WebExceptionInstrumentation extends InstrumentationBase<GlobalErrorsInstrumentationConfig> {
+  constructor(config: GlobalErrorsInstrumentationConfig = {}) {
+    super('@opentelemetry/instrumentation-web-exception', '0.0.1', config);
+    this.onError = this.onError.bind(this);
+  }
+
+  init() { }
+
+  onError( event: ErrorEvent | PromiseRejectionEvent) {
```
overbalance:
```suggestion
  onError(event: ErrorEvent | PromiseRejectionEvent) {
```

overbalance:
@wolfgangcodes Agreed, we should merge first and discuss in the SIG meeting and on Slack.

pichlermarc:
Looks good % package-lock conflicts. 🙂 
Re-requesting a review from @martinkuba as it's still marked as `Changes Requested`

PR #2752, opened 2025-03-12
feat(instrumentation-kafkajs): update semantic conventions

brianphillips:

Semantic conventions used in `@opentelemetry/instrumentation-kafkajs` are quite outdated.

- Updated `@opentelemetry/semantic-conventions` in `@opentelemetry/instrumentation-kafkajs` to version 1.30.0.
- Updated code for span creation to use new attributes/values per semantic conventions.
- Added various metrics per the semantic conventions for messaging systems
- Added tests for the above


plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -111,11 +229,39 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
...
+        [ATTR_MESSAGING_SYSTEM]: MESSAGING_SYSTEM_VALUE_KAFKA,
+        [ATTR_MESSAGING_OPERATION_NAME]: `${event.payload.apiName}`, // potentially suffix with @${event.payload.apiVersion}?
+        [ATTR_SERVER_ADDRESS]: address,
+        [ATTR_SERVER_PORT]: Number.parseInt(port, 10),
+      });
+    });
+
+    consumer[EVENT_LISTENERS_SET] = true;
+  }
+
```
brianphillips, line 264: I tried for quite awhile to figure out a way to wrap/stub/spy/etc on these (as well as the producer event listeners below) and failed. Very open to suggestions.

plugins/node/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -221,47 +409,75 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
...
+                  message,
+                  operationType: MESSAGING_OPERATION_TYPE_VALUE_PROCESS,
+                  link: origSpanLink,
+                  attributes: {
+                    [ATTR_MESSAGING_DESTINATION_PARTITION_ID]: String(
+                      payload.batch.partition
+                    ),
+                  },
+                })
+              );
```
brianphillips, line 457: This isn't new code but it feels a little bit wrong when compared to the examples in the semantic conventions doc ([this one](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#batch-receiving), specifically). All of these `process` spans for each message will have the same start/end times as the batch itself and I'm not sure that's helpful or intuitive. But for now, I'm leaving that design choice alone and just adding the corresponding metric that the semantic convention requires for all `process` spans.

seemk: Yeah this might be out of scope for this PR. Think creating a single `receive` span with `messaging.batch.message_count` attribute would be fine and then add the message ID, parent span context as links, not as new spans. (Trying this out with SQS: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2345)

seemk (approved):

pichlermarc (approved):

Issue #2747, opened 2025-03-07
instrumentation-aws-lambda: Error when configuring tracing without metrics


jordankay13:
### What version of OpenTelemetry are you using?
instrumentation-aws-lambda: v0.50.3

### What version of Node are you using?
22

### What did you do?
Instrumented an AWS Lambda Function with opentelemetry
```js
  const provider = new NodeTracerProvider({
    resource: new Resource({
      [ATTR_SERVICE_NAME]: 'my-service'
    }),
    sampler: new AlwaysOnSampler(),
    spanProcessors: [
      new BatchSpanProcessor(myExporter),
    ],
  });

  registerInstrumentations({
    instrumentations: [
      getNodeAutoInstrumentations({
        '@opentelemetry/instrumentation-aws-lambda': {
          disableAwsContextPropagation: true,
        },
      }),
    ],
  });
```

### What did you expect to see?
Error logs should be clean.

### What did you see instead?
Error logs:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b520d048465d9b3dfdf275976010c989d2a78a2c/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L392-L394
`Metrics may not be exported for the lambda function because we are not force flushing before callback.`

### Additional context
Despite not wanting metrics to be configured, the lambda instrumentation is requiring it through this error. We now either have to filter out these errors from our alerting or we have to configure a MeterProvider.

pichlermarc:
Categorized as P4 (logspam) - looks like there's no need to log an error in this case.
cc @jj22ee (component owner)


(side note: It's a bit of a weird feature as it accesses SDK methods from an API instance, violating an OTel design principle. Though I can see why it would do that (avoiding data loss for a something as short-lived as a lambda).)

pichlermarc:
Closing this as this was addressed by #2753

PR #2753, opened 2025-03-13
fix(instrumentation-aws-lambda): Change endSpan error-logs for forceflush into debug-logs

jj22ee:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2747

- This log is not an actual error, because it is really just a warning. Thus, change it to `debug` log.
  - It could be argued that this could be `warn` log instead, but given the default `info` log level, these logs appear in each Lambda invocation when TracerProvider and/or MeterProvider are not configured, which is a bit noisy for logs.


pichlermarc (approved):
thanks! 🙂

PR #2763, opened 2025-03-18
test(aws-sdk): remove runs for Node.js <18.x

trivikr:

- Support for Node.js <18.x as dropped in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2738

- Remove runs for Node.js <18.x


scripts/skip-test-if.js
```diff
@@ -1,56 +0,0 @@
-/*
```
pichlermarc, line 1: please keep this - we'll likely need it for other instrumentations going forward.

pichlermarc (commented):
overall looks good, please keep the `skip-test-if.js` script around though. we'll likely need it for other instrumentations soon.

plugins/node/opentelemetry-instrumentation-aws-sdk/.tav.yml
```diff
@@ -1,67 +1,28 @@
...
-    - node: "16"
-      versions:
-        include: ">=3.6.1 <3.723.0"
-        exclude: "3.529.0 || >=3.363.0 <=3.377.0"
-        mode: "max-7"
-      commands:
-        - mocha --require '@opentelemetry/contrib-test-utils' test/aws-sdk-v3-s3.test.ts
-        - mocha --require '@opentelemetry/contrib-test-utils' test/s3.test.ts
-    - node: "14"
+    - node: '>=18'
```
david-luna, line 11: Test matrix does not have versions older than v18 https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5837997dfb42a6ea6482999c0ca3e00cf68cdb8d/.github/workflows/test-all-versions.yml#L49

plugins/node/opentelemetry-instrumentation-aws-sdk/.tav.yml
```diff
@@ -1,67 +1,28 @@
...
 # - https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1828#issuecomment-1834276719
 
 # node version support in JS SDK v3:
-# - 14.x dropped in v3.567.0 https://github.com/aws/aws-sdk-js-v3/pull/6034
 # - 16.x dropped in v3.723.0 https://github.com/aws/aws-sdk-js-v3/pull/6775
 
-"@aws-sdk/client-s3":
-  env:
-    - SKIP_TEST_IF_DISABLE=true
+'@aws-sdk/client-s3':
```
david-luna, line 8: do you have any formatter in your editor?

plugins/node/opentelemetry-instrumentation-aws-sdk/.tav.yml
trivikr: That's why this PR removes them. It was missed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2738

plugins/node/opentelemetry-instrumentation-aws-sdk/.tav.yml
trivikr: Yes. And it uses project formatter conventions which explicitly mentions single quote https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5837997dfb42a6ea6482999c0ca3e00cf68cdb8d/prettier.config.js#L7

The existing file did not use those conventions.

trivikr:
This change was done in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2700#issuecomment-2735141908

PR #2768, opened 2025-03-20
chore: update auto-instrumentations README.md

unflxw:
The latest version of the auto-instrumentation package is only compatible with 2.x/0.2xx versions. Update the README accordingly.

### DIFF
--- a/metapackages/auto-instrumentations-node/README.md
+++ b/metapackages/auto-instrumentations-node/README.md
@@ -11,7 +11,7 @@ The net result is the ability to gather telemetry data from a Node application w
 
 This module also provides a simple way to manually initialize multiple Node instrumentations for use with the OpenTelemetry SDK.
 
-Compatible with OpenTelemetry JS API and SDK `1.0+`.
+Compatible with OpenTelemetry JS API and SDK `2.0+`.
 
 ## Installation
 


pichlermarc (approved):
thanks!

unflxw:
@pichlermarc No prob! Outside the scope of this PR, but, are there plans to bring the versioning scheme of contrib in line with what the main repo is doing? (`0.2xx` for experimental, `2.x` for non-experimental)

pichlermarc:
> Outside the scope of this PR, but, are there plans to bring the versioning scheme of contrib in line with what the main repo is doing? (`0.2xx` for experimental, `2.x` for non-experimental)

We've not made any plans to do this at this time. I can see that this could be helpful for troubleshooting though. 🤔
It's somewhat difficult to do for contrib since all the packages are all independently versioned. They would quickly diverge again. Also not all packages break compatibility with 1.x - only resource detectors, span-processors and the metapackages are actually broken. Instrumentations and propagators will continue to work, albeit with dropped support for older Node.js and TypeScript versions.

unflxw:
> Instrumentations and propagators will continue to work

Oh, that's interesting. I assumed that the bump to `@opentelemetry/instrumentation@0.200.0` in the instrumentation's dependencies meant that support for 1.x was dropped.

pichlermarc:
> > Instrumentations and propagators will continue to work
> 
> Oh, that's interesting. I assumed that the bump to `@opentelemetry/instrumentation@0.200.0` in the instrumentation's dependencies meant that support for 1.x was dropped.

Hmm, yes it's kind of a complex topic. 😅 

Technically no two experimental versions of `@opentelemetry/instrumentation` are expected to be compatbile with each other as according to semver these are to be treated as if they were different major versions (hence the need to make `@opentelemetry/instrumentation` ready). In reality, it does happen only every once in a while that `@opentelemetry/instrumentation` introduces a breaking change that is actually breaking in real world use. `0.200.0` was not such a change and `registerInstrumentations` from the old instrumentation package *should* still work with instrumentations that are based on the new one.

Also since all of the instrumentation packages interact don't interact with the SDK directly, but rather with the `@opentelemetry/api` package (which did not receive a major version bump) nothing is technically broken for them. This is the original intention of the `@opentelemetry/api` package (and the API/SDK split in the specification): the SDK setup and behavior can change without breaking instrumentation code. 🙂

PR #2769, opened 2025-03-21
feat(instrumentation-aws-sdk): add gen ai conventions for converse stream span

anuraaga:

- Currently only non-streaming `Converse` populates gen ai conventions

- Allows a service extension to override the response, needed to wrap streams
- Don't end the span in middleware if an extension indicates the response is a stream as the extension needs to end it after it populates attributes during stream consumption. Unrelated to semconv, this should also make the timings more accurate instead of only being for the initial request before response consumption
- Handles `ConverseStream` in bedrock extension using above

/cc @trentm @codefromthecrypt


plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
```diff
@@ -400,12 +400,16 @@ export class AwsInstrumentation extends InstrumentationBase<AwsSdkInstrumentatio
...
                     normalizedResponse,
                     span,
                     self.tracer,
                     self.getConfig()
                   );
+                  if (override) {
+                    response.output = override;
+                    normalizedResponse.data = override;
+                  }
                   self._callUserResponseHook(span, normalizedResponse);
```
anuraaga, line 413: I considered whether this should be done for the user hook too but didn't think there's enough use case for it. Currently the change is only internal since AFAIK, users can't define service extensions

codefromthecrypt (approved):
Streaming is tricky but you handled it concisely. Thanks

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/ServiceExtension.ts
```diff
@@ -29,6 +29,10 @@ import {
 export interface RequestMetadata {
   // isIncoming - if true, then the operation callback / promise should be bind with the operation's span
   isIncoming: boolean;
+  // isStream - if true, then the response is a stream so the span should not be ended by the middleware.
+  // the ServiceExtension must end the span itself, generally by wrapping the stream and ending after it is
+  // consumed.
+  isStream?: boolean;
```
trentm, line 35: Do you know if the GenAI SIG discussed/documented wanting this behaviour of ending the span after the full stream is consumed?   I've seen opinions vary when discussing HTTP streaming. See the guidance for HTTP client span duration here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-client-span-duration

Is there any equivalent in the Python GenAI-related instrumentations?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -34,6 +34,10 @@ import {
   NormalizedRequest,
   NormalizedResponse,
 } from '../types';
+import {
+  ConverseStreamOutput,
+  TokenUsage,
+} from '@aws-sdk/client-bedrock-runtime';
```
trentm, line 40: This adds a runtime dependency on the library being instrumented. In general, I don't think we can have that, because it adds coupling between the instrumentation (which is meant to work for a range of aws-sdk versions) and a particular version of the aws-sdk dependency.

These are just types. My understanding is the typical handling is to add a (possibly simplified) *copy* of the types from the module being instrumented, and add those to a local types ts file in the instrumentation. See, for example, the existing "src/aws-sdk.types.ts" in this instrumentation package, which says:

```ts
/*
  These are slightly modified and simplified versions of the actual SQS types included
  in the official distribution:
  https://github.com/aws/aws-sdk-js/blob/master/clients/sqs.d.ts
  These are brought here to avoid having users install the `aws-sdk` whenever they
  require this instrumentation.
*/
```

trentm (commented):
Some initial review. I haven't looked carefully at the implementation yet.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/ServiceExtension.ts
trentm: > While the duration could keep streaming out of it, that would then need to override the end time of that span with the earlier timestamp, which I don't think the JS SDK supports.

Correct, it does not support that. The (documented) behaviour then would be that the Span duration would be just up until the first response from the server, effectively [TTFB](https://en.wikipedia.org/wiki/Time_to_first_byte).

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
trentm: It looks like instrumentation-graphql and some others are using `import type ...` for this kind of thing.
Sounds good as long as it passes the test-all-versions tests.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/ServiceExtension.ts
anuraaga: Ah do you mean that we should use TTFB here? That would mean we couldn't record usage information though.

BTW, I realized that this might be closer to RPC than HTTP which has some specification for streaming

https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/#message-event

> an event for each message sent/received on client and server spans SHOULD be created

I think this also implies the overall span is for the whole stream. FWIW python keeps the span for the entire stream too.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/load-instrumentation.ts
```diff
@@ -21,12 +21,25 @@
...
-} from '@opentelemetry/contrib-test-utils';
+  AggregationTemporality,
+  InMemoryMetricExporter,
+  MeterProvider,
+  PeriodicExportingMetricReader,
+} from '@opentelemetry/sdk-metrics';
 import { AwsInstrumentation } from '../src';
 
 export const instrumentation = new AwsInstrumentation();
-export const metricReader = initMeterProvider(instrumentation);
```
anuraaga, line 31: I realized this pattern seemed to not work for multiple tests. I did try passing `DELTA` temporality in the `MetricReader` constructor since I thought it would fix it but it didn't. So I changed to a pattern inspired by what's used in Java

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/LibraryTestRunner.java#L122

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/ServiceExtension.ts
trentm: > Ah do you mean that we should use TTFB here?

No, I did not mean to imply what this "should" do. I don't have a strong opinion one way or the other. 

The HTTP span guidance from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-client-span-duration says:

>  Because of the potential for confusion around this, HTTP client library instrumentations SHOULD document their behavior around ending HTTP client spans.

I only meant to say that *if it is decided to handle this stream* by having the span end just on the initial response that this should be documented.

> FWIW python keeps the span for the entire stream too.

Sounds good to me to have the intention for the JS instrumentation to be the same.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/bedrock-runtime.test.ts
```diff
@@ -224,6 +352,7 @@ describe('Bedrock', () => {
     });
   });
 
+  // TODO: Instrument InvokeModel
```
trentm, line 355: ? Isn't this already done?

trentm (commented):
LGTM, with a couple small nits.

@jj22ee, @blumamir, @trivikr  It would be good to have one your opinions on this as component owners. @jj22ee I think this PR predates your being added as a component owner for instrumentation-aws-sdk, so it seems likely you haven't seen this.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -342,6 +358,63 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+      }
+      this.setUsage(response, span, usage, startTime);
+    } finally {
+      span.end();
+    }
+  }
+
+  private static setStopReason(span: Span, stopReason: string | undefined) {
+    if (stopReason !== undefined) {
+      console.log(stopReason);
```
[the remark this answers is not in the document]
anuraaga: Thanks! Fixed the two points

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -342,6 +358,62 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+  private responseHookConverseStream(
+    response: NormalizedResponse,
+    span: Span,
+    tracer: Tracer,
+    config: AwsSdkInstrumentationConfig,
+    startTime: HrTime
+  ) {
+    return {
+      ...response.data,
+      stream: this.wrapConverseStreamResponse(
```
jj22ee, line 374: Does `stream: this.wrapConverseStreamResponse(...)` overwrite the `stream` from `...response.data`?
I think it will be nice to have a comment to call this out.

anuraaga: Yup - added a comment about it

trentm (approved):
LGTM. Waiting for an approval from one of the code owners before merging.

jj22ee (approved):
Thanks, also LGTM!

I did initially thought it was weird that the instrumentation has to rely on the user to fully consume the response's stream in order to end the span, since it could leave a "never-ended" span in case the user doesn't consume the stream. But it sounds like it would be weirder if the user didn't consume the entire stream from the earlier discussion.

PR #2771, opened 2025-03-27
feat(instrumentation-aws-sdk): add gen ai metrics for bedrock

anuraaga:

- Gen AI defines metrics that should be applied for operations like bedrock runtime

- Allows propagating meter to service extensions
- Passes start time to service extensions to allow them to compute operation duration as needed
- Implement metrics for bedrock service extension
- Refactor tests to share a single instrumentation instance

/cc @trentm @codefromthecrypt


plugins/node/opentelemetry-instrumentation-aws-sdk/test/load-instrumentation.ts
```diff
@@ -0,0 +1,25 @@
+/*
```
anuraaga, line 1: I tried many attempts to have the meter-enabled instrumentation running for bedrock tests, including stuff like `beforeEach(() => registerInstrumentationForTesting(); instrumentation.enable())` but nothing worked, so I ended up refactoring to make sure there's only a single instrumentation instance among all these tests (since mocha doesn't isolate tests)

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -23,19 +32,57 @@ import {
...
+  hrTimeDuration,
+  hrTimeToMilliseconds,
+} from '@opentelemetry/core';
 
 export class BedrockRuntimeServiceExtension implements ServiceExtension {
+  private tokenUsage!: Histogram;
+  private operationDuration!: Histogram;
+
+  updateMetricInstruments(meter: Meter) {
+    this.tokenUsage = meter.createHistogram('gen_ai.client.token.usage', {
```
trentm, line 60: nit: A comment linking to https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage could be helpful. Ditto for the other metric below.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -98,32 +145,61 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+          ...sharedMetricAttrs,
+          [ATTR_GEN_AI_TOKEN_TYPE]: GEN_AI_TOKEN_TYPE_VALUE_INPUT,
+        });
       }
       if (outputTokens !== undefined) {
         span.setAttribute(ATTR_GEN_AI_USAGE_OUTPUT_TOKENS, outputTokens);
+
+        this.tokenUsage.record(outputTokens, {
+          ...sharedMetricAttrs,
+          [ATTR_GEN_AI_TOKEN_TYPE]: GEN_AI_TOKEN_TYPE_VALUE_COMPLETION,
```
trentm, line 201: `export const GEN_AI_TOKEN_TYPE_VALUE_COMPLETION = "output"` is weird: "COMPLETION" rather than "OUTPUT". 
Ah, looks like that'll be fixed by https://github.com/open-telemetry/semantic-conventions/pull/1757/files#diff-0936f2641601c5b370b10dfb33c8f82d6d731a2663cf042eef2aa5114c720902R177
which is probably in semconv v1.31.0, for which the OTel JS semconv package doesn't have a release yet.
So eventually this'll be `GEN_AI_TOKEN_TYPE_VALUE_OUTPUT`.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/load-instrumentation.ts
trentm: Yah, mocha running all tests in the same single process is a pain, IMHO. (Test frameworks often go this way though because speed.)

Could you add a block comment at the top of the file (after the license block) that explains basically what you've said above.

plugins/node/opentelemetry-instrumentation-aws-sdk/test/lambda.test.ts
```diff
@@ -14,12 +14,8 @@
  * limitations under the License.
  */
 
-import { AwsInstrumentation } from '../src';
-import {
-  getTestSpans,
-  registerInstrumentationTesting,
-} from '@opentelemetry/contrib-test-utils';
-registerInstrumentationTesting(new AwsInstrumentation());
+import { getTestSpans } from '@opentelemetry/contrib-test-utils';
```
trentm, line 17: This import from contrib-test-utils doesn't absolutely need to be before `./load-instrumentation` does it?

plugins/node/opentelemetry-instrumentation-aws-sdk/test/load-instrumentation.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+import {
+  initMeterProvider,
+  registerInstrumentationTesting,
+} from '@opentelemetry/contrib-test-utils';
+import { AwsInstrumentation } from '../src';
+
+export const instrumentation = new AwsInstrumentation();
+export const metricReader = initMeterProvider(instrumentation);
+instrumentation._updateMetricInstruments();
```
trentm, line 24: This method is called in InstrumentationAbstract's constructor. Is it necessary to call it again?
The method is also called in `initMeterProvider` (that calls `instrumentation.setMeterProvider(meterProvider);`, which I think calls `instr._updateMetricInstruments()`).

anuraaga:
Thanks @trentm applied the cleanups. Apologies for the delay finally back to work from kubecon

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
trentm: There *is* a semconv v1.31.0 release now, so you *could*:

1. update package.json to bump the dep to `"@opentelemetry/semantic-conventions": "^1.31.0"`
2. change the usage and import in bedrock-runtime.ts to `GEN_AI_TOKEN_TYPE_VALUE_OUTPUT`
3. regenerate the src/semconv.ts by re-running (I hope this works, it is meant to):

```
cd .../plugins/node/opentelemetry-instrumentation-aws-sdk
~/src/opentelemetry-js-contrib/scripts/gen-semconv-ts.js
```

trentm (commented):
Mostly LGTM. One comment on the `init()` method below.

@blumamir @jj22ee @trivikr Any comments on the ServiceExtension changes adding metrics support?

anuraaga:
Thanks - updated semconv

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
```diff
@@ -67,7 +67,8 @@ type V3PluginCommand = AwsV3Command<any, any, any, any, any> & {
 
 export class AwsInstrumentation extends InstrumentationBase<AwsSdkInstrumentationConfig> {
   static readonly component = 'aws-sdk';
-  private servicesExtensions: ServicesExtensions = new ServicesExtensions();
+  // initialized in callbacks from super constructor for ordering reasons.
+  private declare servicesExtensions: ServicesExtensions;
```
jj22ee, line 71: Can you elaborate on the ordering reasons? But this is OK anyways since you made `servicesExtensions` to be still initialized due to the superclass constructor call to `_updateMetricInstruments`.

jj22ee (commented):
@trentm The metrics support in `ServiceExtension` looks good, no concern from me.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
anuraaga: It was about `declare` - I reworded so hopefully it's clearer

jj22ee: Got it. Just wondering what is the benefit of needing `_updateMetricInstruments` to initialize `servicesExtensions` instead of keeping the initialization as is?

anuraaga: We do need to pass `meter` on `setMeterProvider` to catch updates if it's replaced, but during constructor, it will be called before the child class's so we need to make sure to update metric instruments ourselves. So to keep initialization in the child class, the diff would look like this

```
+++ b/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
@@ -67,11 +67,14 @@ type V3PluginCommand = AwsV3Command<any, any, any, any, any> & {

 export class AwsInstrumentation extends InstrumentationBase<AwsSdkInstrumentationConfig> {
   static readonly component = 'aws-sdk';
-  // initialized in callbacks from super constructor for ordering reasons.
-  private declare servicesExtensions: ServicesExtensions;
+
+  private servicesExtensions: ServicesExtensions;

   constructor(config: AwsSdkInstrumentationConfig = {}) {
+
+    this.servicesExtensions = new ServicesExtensions();
+    this.servicesExtensions.updateMetricInstruments(this.meter);
   }

   protected init(): InstrumentationModuleDefinition[] {
@@ -469,9 +472,8 @@ export class AwsInstrumentation extends InstrumentationBase<AwsSdkInstrumentatio
   }

   override _updateMetricInstruments() {
-    if (!this.servicesExtensions) {
-      this.servicesExtensions = new ServicesExtensions();
+    if (this.servicesExtensions) {
+      this.servicesExtensions.updateMetricInstruments(this.meter);
     }
-    this.servicesExtensions.updateMetricInstruments(this.meter);
   }
 }
```

Keeping the metric instrument update in one place seemed simpler at first but on reading, this doesn't seem too bad either. Any preference?

jj22ee: Thanks for the explanation (and the diff!). You're right, keeping `updateMetricInstruments` in one place is definitely simpler, I prefer your current solution.

jj22ee:
LGTM!

There is a conflict with main branch right now, most likely due to another Bedrock related PR that got merged from https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2777, but I think there should be little to resolve.

anuraaga:
Sorry for the delay @jj22ee, resolved the conflict

jj22ee (approved):
Looks like there's another conflict, but it must likely be `package-lock.json`.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts
trentm: FWIW, doing metrics initialization in Instrumentation classes is commonly done the way @anuraaga has it here. For example:
https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js-contrib%20_updateMetricInstruments&type=code

trentm (approved):

PR #2781, opened 2025-04-03
fix(instrumentation-undici): fix several header handling handling bugs

benjaminjkraft:

The handling for User-Agent had a bunch of bugs, most notably with the handling of multiple-valued headers, which are rare but legal.

Code similar to the added test "another header with multiple values" caused us errors in production from the user-agent code, and is where I started. Reading the code, writing tests, and improving the types revealed several more bugs in the same code as well as the span-to-attributes handling; the added tests "multiple user-agents", "another header with value user-agent" also fail before this PR (the others are just to actually cover the ordinary cases.

Similarly, I also fixed an incorrect case in undici v5 where it would treat a `user-agent-bogus` header as a user agent, but I couldn't write a test case since the tests run with the newer version.

The relevant code is rewritten to handle multiple headers consistently, and refactored so we don't have to write it all twice.

### DIFF
--- a/plugins/node/instrumentation-undici/src/undici.ts
+++ b/plugins/node/instrumentation-undici/src/undici.ts
@@ -165,6 +165,49 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     });
   }
 
+  /**
+   * For each header in the request, call the callback. Skips likely-invalid
+   * headers. Multi-valued headers are passed through. The loop exits early if
+   * the callback returns true.
+   */
+  private forEachRequestHeader(
+    request: UndiciRequest,
+    callback: (key: string, value: string | string[]) => boolean | undefined
+  ): void {
+    if (Array.isArray(request.headers)) {
+      // headers are an array [k1, v2, k2, v2] (undici v6+)
+      for (let i = 0; i < request.headers.length; i += 2) {
+        const key = request.headers[i];
+        if (typeof key !== 'string') {
+          // Shouldn't happen, but the types don't know that, and let's be safe
+          continue;
+        }
+        if (callback(key, request.headers[i + 1])) {
+          break;
+        }
+      }
+    } else if (typeof request.headers === 'string') {
+      // headers are a raw string (undici v5)
+      const headers = request.headers.split('\r\n');
+      for (const line of headers) {
+        if (!line) {
+          continue;
+        }
+        const colonIndex = line.indexOf(':');
+        if (colonIndex === -1) {
+          // Invalid header? Probably this can't happen, but again let's be safe.
+          continue;
+        }
+        const key = line.substring(0, colonIndex);
+        const value = line.substring(0, colonIndex + 1);
+
+        if (callback(key, value)) {
+          break;
+        }
+      }
+    }
+  }
+
   // This is the 1st message we receive for each request (fired after request creation). Here we will
   // create the span and populate some atttributes, then link the span to the request for further
   // span processing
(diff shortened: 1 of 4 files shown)


benjaminjkraft:
(Working on the CLA with my company, should have that fairly soon.)

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -165,6 +165,49 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
+        if (!line) {
+          continue;
+        }
+        const colonIndex = line.indexOf(':');
+        if (colonIndex === -1) {
+          // Invalid header? Probably this can't happen, but again let's be safe.
+          continue;
+        }
+        const key = line.substring(0, colonIndex);
+        const value = line.substring(0, colonIndex + 1);
```
JacksonWeber, line 202: Wouldn't this capture the key again since you're starting from the zeroth index of the `line` string again? Could be something like: `const value = line.substring(colonIndex + 1).trim();`

david-luna:
Hi @benjaminjkraft,

thanks for contributing to opentelemetry and welcome :)

From your message I couldn't understand what are exactly the errors this PR fixes. I see you mention:
- header values provided as an array of strings (multiple-valued headers)
- headers prefixed by `user-agent` may be considered the user agent wrongly.

I'll check your changes assuming both points above. Please let me know if there is something else this PR is fixing soit can be mentioned in the changelog.

Cheers

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -218,26 +261,16 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
-      );
-      if (idx >= 0) {
-        userAgent = request.headers[idx + 1];
+    this.forEachRequestHeader(request, (key, value) => {
+      if (key.toLowerCase() === 'user-agent') {
+        // user-agent should only appear once per the spec, but the library doesn't
+        // prevent passing it multiple times, so we handle that to be safe.
+        const userAgent = Array.isArray(value) ? value[0] : value;
+        attributes[SemanticAttributes.USER_AGENT_ORIGINAL] = userAgent;
+        return true; // no need to keep iterating
```
david-luna, line 270: nit: IMO `forEachXXX` name implies we're executing for all items in the collection. Depending on a previous result to decide to continue the loop does not follow the semantics of the name. Without the comment after the return statement one not familiar with the code will not know that the loop breaks.

The control of the format and access to the headers in a separate function is a good idea but maybe it will serve better if it returns an iterable object. This way the consumer code can loop through it (and break the loop) with the language specifics like [for..of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) and [break](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break)

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -218,26 +261,16 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
-      const idx = request.headers.findIndex(
-        h => h.toLowerCase() === 'user-agent'
-      );
-      if (idx >= 0) {
-        userAgent = request.headers[idx + 1];
+    this.forEachRequestHeader(request, (key, value) => {
+      if (key.toLowerCase() === 'user-agent') {
+        // user-agent should only appear once per the spec, but the library doesn't
+        // prevent passing it multiple times, so we handle that to be safe.
+        const userAgent = Array.isArray(value) ? value[0] : value;
```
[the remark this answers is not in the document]
benjaminjkraft: I could log it, although I dunno where that goes in practice? We would probably also be spec-compliant-ish to do `value.join(", ")`.

plugins/node/instrumentation-undici/src/undici.ts
benjaminjkraft: sure, I can do an iterable, I wasn't totally sure if that was kosher yet for the lib but I guess they're supported pretty much everywhere these days.

plugins/node/instrumentation-undici/src/undici.ts
benjaminjkraft: Eek, you're right!

Which makes me think -- do we have a way to actually unit test the patches against older library versions? I guess that would have to be a separate test package?

benjaminjkraft (commented):
Sorry, for a more explicit list of the issues this fixes:
- (undici v6+) the code crashed if any header (preceding `user-agent`) is multi-valued
- (undici v6+) the code used the wrong value if a header *value* preceding the `user-agent` header was `"user-agent"` (it would log the next header-key as the user-agent)
- (undici v5) the code would treat any header starting with `user-agent`, e.g. `user-agent-bogus` as the `user-agent`

(CLA is in review, I'll get back to this later this week once it's approved.)

benjaminjkraft:
Thanks for the reviews -- pushed updates to fix the v5 typo and swap to generators (which indeed look cleaner). (Also, CLA is now signed.) Let me know if there's anything else you'd like to see!

plugins/node/instrumentation-undici/src/undici.ts
david-luna: > Eek, you're right!
> 
> Which makes me think -- do we have a way to actually unit test the patches against older library versions? I guess that would have to be a separate test package?

The `test-all- versions` script should perform these tests. You can inspect which versions are tested in the `.tav.yml` file at the root of this instrumentation.

david-luna:
> Which makes me think -- do we have a way to actually unit test the patches against older library versions? I guess that would have to be a separate test package?

You can use the script `npm run test-all-versions` which will install and test with different versions of undici. You could even update `.tav.yml` file since Node.js v14 and v16 support was dropped in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2738

You have a compile issue :(
https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/14367208545/job/40308049357?pr=2781

david-luna:
@benjaminjkraft did you have time to look at the compile issue? Do you need any kind of assistance?

david-luna:
@benjaminjkraft are you able to finish this PR? Let me know if you need me to jump in :)

david-luna:
Superseded by #2875

PR #2783, opened 2025-04-10
chore: update `nx.json` for better caching

david-luna:

Fixes `nx.json` configuration for caching. The `input` globs need to have the recursive lookup for typescript files `**/*.ts` otherwise it will compile only once from the root folder even if you do changes in a package later.

The issue impacts only locally and not in the CI.

- modify the glob patter in `nx.json` inputs

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -18,8 +18,8 @@
         "^compile"
       ],
       "inputs": [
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
thanks!

PR #2685, opened 2025-01-29
feat(instrumentation-nestjs-core): add support for NestJS 11

neilime:
## Which problem is this PR solving?

- nest-core instrumentation is not supporting last version of NestJS (11)

## Short description of the changes

- Adding support for NestJS 11

PR #2790, opened 2025-04-14
chore(instrumentation-nestjs): update nestjs supported versions

anuraaga:

- README is not up-to-date with code

- Updates supported versions in nestjs README

#2685 added support for NestJS 11 and I can confirm the latest released artifact works fine with it, so this updates the README to reflect it. I wasn't sure if it's better to remove `<YY` for `YY` that don't exist yet or keep it to not "pre-support" new major versions, let me know which is better.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-nestjs-core/README.md
+++ b/plugins/node/opentelemetry-instrumentation-nestjs-core/README.md
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-nestjs-core
 
 ### Supported Versions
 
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
Thanks @anuraaga 🙂 

> https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2685 added support for NestJS 11 and I can confirm the latest released artifact works fine with it, so this updates the README to reflect it. I wasn't sure if it's better to remove <YY for YY that don't exist yet or keep it to not "pre-support" new major versions, let me know which is better.

Not pre-supporting is better - we do this to avoid breaking end users on new major versions.

PR #2798, opened 2025-04-23
Add Github action to add "has:owner-approval" label for approved PRs

hectorhdzg:
Automate label tagging for owner reviews in PRs 
-

Added scripts to parse component_owners.yml, to get existing components and owners by componer
Added script to get details about a PR, looks for all updated files and determine if one of the owner of that file already approved the PR, then add label.
Added workflow to kick off the script when a PR is reviewed and approved.
-

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -1,0 +1,26 @@
+name: PR review
+# This workflow is triggered when a PR review is submitted. It checks if the review is approved and runs a script to add labels to the PR based on the review.
+
+on:
+  pull_request_target:
+    types:
+      - submitted
+
+jobs:
+  approved:
+    if: github.event.review.state == 'approved'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Run update permissions
+        run: chmod +x ./.github/workflows/scripts/add-labels-to-reviewed-pr.sh
+
+      - name: Run add-labels-to-reviewed-pr.sh
+        run: ./.github/workflows/scripts/add-labels-to-reviewed-pr.sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE: ${{ github.event.issue.number }}
+          COMMENT: ${{ github.event.comment.body }}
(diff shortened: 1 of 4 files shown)


JacksonWeber (approved):
LGTM

.github/workflows/pr-review.yml
```diff
@@ -0,0 +1,22 @@
+name: PR review
+# This workflow is triggered when a PR review is submitted. It checks if the review is approved and runs a script to add labels to the PR based on the review.
+on:
+  pull_request_review:
+    types:
+      - submitted
```
pichlermarc, line 6: Hmm, I think that the `pull_request_review` trigger only provides a read-only `GITHUB_TOKEN` - I guess it would not be able to re-label the PR in that case, right? 🤔 

Thinking more about it, I'm also wondering if there are some security concerns if we were to give the workflow additional permissions. 🤔

hectorhdzg: We can add permissions by workflow right? that would prevent us to change the default permissions and keep them read only for other workflows

pichlermarc: yes, I think you're right  - let's use 
```     
      contents: read
      pull-requests: write
```

on the job as that's what another workflow is using already. Then I think this PR is ready to go. 

The overall workflow permission can be `contents: read`.

pichlermarc:
overall looks good, let's use some explicit permissions and then we can get this merged. 🙂

hectorhdzg:
Updated, thanks for reviewing @pichlermarc  let me know if there is something else

pichlermarc (approved):
Thanks!

pichlermarc (commented):
ah - looks like the script is missing `+x` permissions.

pichlermarc:
(see [this run](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17619131319/job/50060049835))

hectorhdzg:
@pichlermarc I'm not sure if the change is going to work as I cannot kick off the job, we can coordinate if needed to try different things, hopefully the issue is resolved and that is not needed

hectorhdzg:
@pichlermarc can you kick off the workflow when you had a chance?

Issue #1287, opened 2022-11-14
AWS Lambda Instrumentation: Consider allowing shimmer to fail monkeypatching


NPellet:
- [x] This only affects the JavaScript OpenTelemetry library

The problem is a well-known one. There are cases where shimmer fails to patch a function. This can happen when, for example, the lambda handler is compiled with esbuild and using an immutable ESM export `export const handler = function() {}` transpiled to CJS.

I believe we should try/catch the shimmer wrap function here https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts#L121 and return the unwrapped module if shimmer fails.

What do you think ?

PR #2799, opened 2025-04-23
fix(aws-lambda-instrumentation): errors while instrumenting cause instrumented lambda to fail 

mzl-md:

Errors thrown by shimmer cause the instrumented lambda to fail.

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1287

Since the instrumentation should never cause the instrumented lambdas to fail, errors thrown by shimmer should be caught and logged.

I can reproduce the error but I don't really have an idea on how to fix this. @anuraaga could you help?

Output of the test with try-catch enabled:
> npx mocha test/integrations/lambda-handler.moduleError.test.ts
>  lambda handler
    handler with module.exports syntax
      1) should not fail
no original to unwrap to -- has handler already been unwrapped?
>  0 passing (2s)
  1 failing
>  1) lambda handler
       handler with module.exports syntax
         should not fail:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/ms/Repos/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.moduleError.test.ts)
      at listOnTimeout (node:internal/timers:581:17)
      at processTimers (node:internal/timers:519:7)

Without try-catch:
>npx mocha test/integrations/lambda-handler.instrumentationError.test.ts
  lambda handler
    handler with module.exports syntax
      1) should not fail
no original to unwrap to -- has handler already been unwrapped?
  0 passing (9ms)
  1 failing
>  1) lambda handler
       handler with module.exports syntax
         should not fail:
     TypeError: Cannot redefine property: handler

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
@@ -157,11 +157,16 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
               if (isWrapped(moduleExports[functionName])) {
                 this._unwrap(moduleExports, functionName);
               }
+              // try {
               this._wrap(
                 moduleExports,
                 functionName,
                 this._getHandler(lambdaStartTime)
               );
+              // } catch (error) {
+              //   // _wrap is not error safe. We do not want the instrumented lambda to fail if wrapping fails.
+              //   diag.error('patching handler function failed', error);
+              // }
               return moduleExports;
             },
             (moduleExports?: LambdaModule) => {
--- a/plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.instrumentationError.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-aws-lambda/test/integrations/lambda-handler.instrumentationError.test.ts
@@ -1,0 +1,99 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -157,11 +157,16 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+              // try {
               this._wrap(
                 moduleExports,
                 functionName,
                 this._getHandler(lambdaStartTime)
               );
+              // } catch (error) {
+              //   // _wrap is not error safe. We do not want the instrumented lambda to fail if wrapping fails.
+              //   diag.error('patching handler function failed', error);
+              // }
```
jj22ee, line 169: Instead of letting an error get thrown and relying on a try-catch here to catch it, one idea is preemptively determining if `TypeError: Cannot redefine property: handler` will occur, and in that case we just return `[]` and don't attempt to use `InstrumentationNodeModuleDefinition` or `InstrumentationNodeModuleFile`.

I haven't looked into yet, but is it possible to check if the [`"lambda handler is compiled with esbuild and using an immutable ESM export export const handler = function() {} transpiled to CJS."`](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1287)?

mzl-md: I think so, but trying to guess if the wrap-function would throw an error seems not a good approach to me. If the implementation of shimmer changes, we would need to update this as well. Besides, there could be other reasons for wrapping to fail and in my opinion this should never lead to an error in the instrumented code.

mzl-md:
> Note that if there are crashing issues, I would personally go to opentelemetry-js to make sure _wrap and family never throw.

I thought about that but was not sure who else uses _wrap and might depend on it throwing errors. Maybe we can introduce a `_saveWrap()`. I will open a pull requests there.

dyladan:
>I will open a pull requests there.

@mzl-md if this references `shimmer` it is unmaintained, but we have vendored the code in our repo here: https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation/src/shimmer.ts

Feel free to make a PR there for this. I think we can probably change the `_wrap` function as depending on throw behavior is an edge-case but we can discuss there if you make a PR.

Issue #2435, opened 2024-09-17
Add support for Express v5


timfish:
Express v5 has been released and it's not supported:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/80c2f1af6139e43522a944ab496fed68ef912fb2/plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts#L61-L62

I'm looking into the changes required so you can assign this issue to me!

pichlermarc:
@timfish It's yours, thanks for looking into it :slightly_smiling_face:

pichlermarc:
cc @JamieDanielson @pkanal (component owners)

timfish:
I want to retain the v4 tests, duplicate them and test v5 too. To have both versions installed, I was going to use aliases but then realised this does not work because I'm importing `express5` does not get instrumented!

```json
    "express": "4.19.2",
    "express5": "npm:express@5.0.0",
```

I can see for redis, there are two seperate instrumentations for v3 and v4 but it doesn't look this this will be required for express because so far it's looking like not much will need to be changed.

Any tips on what I should do to test both versions? I could add another `package.json` and `node_modules` inside the v5 test directory?

hitsthings:
Hi just curious about the future of this. Movement on the PR looks stalled, but it seems like Express team is looking to emit events that would allow for a simpler implementation. Are you waiting for those events or shipping first as-is? Is there any blocker or is it just one of those "need to find the time" things?

Cheers!

hitsthings:
Awesome, thanks for that update! Much appreciated!

CreativeBrainInc:
Wondering any update on the new PR ??
We are using Express 5 and wanted to implement OpenTelemetry.

PR #2801, opened 2025-04-25
feat(instrumentation-express): add support for Express v5, take 2

trentm:
This PR starts with @timfish's #2437 (Tim did all the hard work) and
makes it a smaller patch -- mostly by avoiding having separate
test/v4 and test/v5 trees.

Now that we are on JS SDK v2 and hence only support Node.js v18 as
a minimum (the same as express@5) the testing story is now simpler:
no need to avoid running some tests with older Node.js versions.
(See #2722 for a general discussion of the pain when having to do that.)

Obsoletes: #2437
Closes: #2435

# Testing Notes

- I've bumped the express version in package.json to v5, so that version is tested via `npm test`.
- I've updated `.tav.yml` so that supported express@4 versions are tested via `npm run test-all-versions`.

(If there is a coverage failure due to this, I'll make the argument that we can ignore it. Given our instrumentations have version-specific code, the only reasonable way to get full coverage is to gather coverage from `npm run test-all-versions` runs and *not* from `npm test` runs.)


plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
```diff
@@ -60,9 +60,13 @@ export class ExpressInstrumentation extends InstrumentationBase<ExpressInstrumen
     return [
       new InstrumentationNodeModuleDefinition(
         'express',
-        ['>=4.0.0 <5'],
+        ['>=4.0.0 <6'],
         moduleExports => {
-          const routerProto = moduleExports.Router as unknown as express.Router;
+          const isExpressV5 =
+            typeof moduleExports?.Router?.prototype?.route === 'function';
```
trentm, line 66: Dev note: I used feature-sniffing to detect the version. We could use `(moduleExports, version) => { ...` and a semver.satisfies() check of the loaded module version if there is a preference for that.

plugins/node/opentelemetry-instrumentation-express/test/express.test.ts
```diff
@@ -852,8 +880,8 @@ describe('ExpressInstrumentation', () => {
     'arr/requiredPath',
     'arr/required/lastParam',
     'arr55/required/lastParam',
-    'arr/requiredPath/optionalPath/',
-    'arr/requiredPath/optionalPath/lastParam',
+    'arr/requiredpath/optionalPath/',
+    'arr/requiredpath/optionalPath/lastParam',
```
trentm, line 884: Dev Note:

Interesting side-bar. The following route in "test/fixtures/use-express-regex.mjs" (added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2008) was testing whether a request path `/test/arr/requiredPath/optionalPath/` matched that route. Notice the capital letter in `requiredPath`. Normally this should NOT match. However it does in express@4, not in express@5 and I think it is an express@4 limitation/bug.

```js
app.get(
  [
    '/test/arr/:id',
    /\/test\/arr[0-9]*\/required(path)?(\/optionalPath)?\/(lastParam)?/,
  ],
  (_req, res) => {
    res.send({ response: 'response' });
  }
);
```

Let's use a simpler example (note to self: /Users/trentm/tmp/asdf.20250424T102536/app.js):

```js
app.get(
  [
    '/foo',
    /^\/another-path$/,
  ],
  (req, res) => {
    res.end('pong');
  }
);
```

- Express routes are case-insensitive by default .
- express@4 used `path-to-regexp@0.1` for its route matching. That version of the lib creates a single regexp to match for any of the entries in the route array. So how to match '/foo' case-insensitively, but '/another-path' case-*sensitively*? The answer is that you don't. This is the regexp used in express@4 for matching that route: `/^\/foo\/?$|^\/another-path$/i`. It is all case-insensitive.  **In other words, but including a regexp route together with a string route in an array to `app.get(...)` results in the regexp route being made case-insensitive.**
- In express@5 the `router@2` module is used for route matching and it handles each of the `'/foo'` and `/^\/another-path$/` entries separately.

The existing test data here was unintentionally relying on that Express routing subtlety. That subtlety changed in express@5, but isn't relevant to the testing here.

trentm (commented):
Some dev notes.

david-luna (approved):

pichlermarc:
cc @raphael-theriault-swi @JamieDanielson @pkanal (component owners)

pichlermarc:
This will likely also fix https://github.com/open-telemetry/opentelemetry.io/issues/6726

pichlermarc (approved):

plugins/node/opentelemetry-instrumentation-express/.tav.yml
```diff
@@ -1,5 +1,5 @@
 express:
   - versions:
-      include: "^4.16.2"
+      include: ">=4.16.2 <6"
```
wesleytodd, line 3: `4.16.2` is not supported anymore, ideally bump this to `4.21.2`.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
wesleytodd: This is very brittle, but I don't know enough about your `InstrumentationNodeModuleDefinition` wrapper to be sure. Is this simply a check to tell which version it is? Or do you have logic somewhere based on if `.route` is a function?

wesleytodd (approved):
I don't have much to include just like on the last one, but if there is some better way we can export the express version from the library I am down for landing that to help prevent you from having checks like that one.

plugins/node/opentelemetry-instrumentation-express/.tav.yml
JamieDanielson: True, although if 4.16.2 still works without issue I'm okay keeping this until it causes us problems, unless it is massively increasing testing time.

wesleytodd: I just mentioned it because of these: https://github.com/expressjs/express/security/advisories?state=published

wesleytodd: I guess I should have posted more context. I doubt any impact from CVEs here. I just like to point it out in case there are parts of the systems I am unaware of that if it is not done for a *real* reason it is best to just avoid the versions with published CVEs. Hope that helps clarify my reasoning for mentioning it.

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
JamieDanielson: I guess the main question here is whether that feature may be backported at any point, removing the reliability of this check. Considering v4.x has been out for... years... I wouldn't rule out backporting features as a possibility. I'd lean toward version if we can, similar to how we do it in MongoDB.

Alternatively if we can comment the reasoning with a link to docs somewhere that might be good enough and we can change to use version if it causes a problem later with a backport.

JamieDanielson: Also just realized you used the semver.satisfies() check in testing. I like the explicitness of it anyway.

plugins/node/opentelemetry-instrumentation-express/.tav.yml
JamieDanielson: > downloads tab on npm

Ah yep, I meant specifically folks who use this instrumentation. But to your point, this version is overall pretty low in terms of downloads compared to even the next minor release. Besides, if someone is still on this old of a version of express, they are less likely to be upgrading this dependency. Let's increase minimum testing version 👍

plugins/node/opentelemetry-instrumentation-express/src/instrumentation.ts
wesleytodd: > I guess the main question here is whether that feature may be backported at any point

No it will not be backborted, but we want to clean up some of these interfaces in the future and so understanding how users are using these api's (especially incorrectly like this for version checking) is important for us to keep in mind.

Sorry, to be clear I am not sure if I was before, but I am one of the TC members for express, that's why I was pinged on the other PR and so got a notification for this one. Feel free to ask me if you have questions for the express side.

trentm: > Is this simply a check to tell which version it is? Or do you have logic somewhere based on if `.route` is a function?

The instrumentation here wraps methods on the Router prototype, including `.route` (just below this `const isExpressV5 = ...` statement), so it feels reasonable to (somehow) check that the thing to be instrumented exists.

If express@5 internals changed such that `Router.prototype.route` went away, then this instrumentation-express implementation would have to change. So, IIUC, a semver check something like `semver.satisfies(version, '^5')` wouldn't really help for future-proofness either.

@wesleytodd Or am I missing a detail in which you think this is more brittle than the semver-based differentiation that the previous PR was doing: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2437/files#diff-81c3e13c3dac53e0dd6aabf6dc165305350eb522a85a2561f9ee51d9249011e7R84-R90

If the var was called `isExpressWithRouterPrototype` or something like that, would that be clearer?

wesleytodd: > The instrumentation here wraps methods on the Router prototype, including .route (just below this const isExpressV5 = ... statement), so it feels reasonable to (somehow) check that the thing to be instrumented exists.

Ah ok yes, with this explanation I agree it makes sense. Calling it `isExpressV5` is what threw me as I thought maybe it could be used for other things related to other parts of the v5 api changes (spoilers for below...).

> If express@5 internals changed such that Router.prototype.route went away,

We cannot do that as it would be breaking. So you are for sure safe on that until the next major. 

> So, IIUC, a semver check

And I agree I am not sure a semver check would be better since you use this for is to patch `.route`. 

> If the var was called isExpressWithRouterPrototype or something like that, would that be clearer?

Yes, after reading your first line that is where my head went before I even read down to the end of your reply. 🎉

trentm: > If the var was called isExpressWithRouterPrototype or something like that, would that be clearer?

Done in commit 87428ca6

Ecksters:
README is still indicating:
>Supported Versions
[express](https://www.npmjs.com/package/express) version >=4.0.0 <5

And I seem to be having trouble getting it to work with Express 5.1, not sure if the README is correct or if I'm doing something wrong on my end.

trentm:
@Ecksters I only fluked upon your comment here.  I've opened https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3159 to update the README.md.



> And I seem to be having trouble getting it to work with Express 5.1,

If you are still having an issue, please open a new issue with details so we can take a look. My understanding is things should be working with express@5.1.0. For example this recentish workflow run includes a test of instr-express with express@5.1.0: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18317653219/job/52162813315

Issue #2759, opened 2025-03-17
pg: cannot instrument an ESM-imported pg-pool


constb:
### What version of OpenTelemetry are you using?

```
    "@opentelemetry/api": "1.9.0",
    "@opentelemetry/core": "1.30.1",
    "@opentelemetry/instrumentation": "0.57.2",
    "@opentelemetry/instrumentation-pg": "0.51.1",
    "@opentelemetry/sdk-node": "0.57.2",
    "@opentelemetry/semantic-conventions": "1.28.0"
```

### What version of Node are you using?

Node.js v22.14.0

### What did you do?

I setup opentelemetry node sdk to instrument a web app with postgresql. I'm using typescript compiled to ES-modules. I'm importing directly from pg-pool with `import Pool from "pg-pool"`.

### What did you expect to see?

Application running.

### What did you see instead?

With OTEL_LOG_LEVEL enabled:

```
@opentelemetry/instrumentation-pg Applying instrumentation patch for module on require hook {
  module: 'pg-pool',
  version: '3.8.0',
  baseDir: '/…/app/node_modules/pg-pool'
}
/…/app/node_modules/@opentelemetry/instrumentation-pg/build/src/instrumentation.js:88
            if ((0, instrumentation_1.isWrapped)(moduleExports.prototype.connect)) {
                                                                         ^

TypeError: Cannot read properties of undefined (reading 'connect')
    at InstrumentationNodeModuleDefinition.patch (/…/app/node_modules/@opentelemetry/instrumentation-pg/build/src/instrumentation.js:88:74)
    at PgInstrumentation._onRequire (/…/app/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:164:39)
    at hookFn (/…/app/node_modules/@opentelemetry/instrumentation/build/src/platform/node/instrumentation.js:227:29)
    at callHookFn (/…/app/node_modules/import-in-the-middle/index.js:29:22)
    at Hook._iitmHook (/…/app/node_modules/import-in-the-middle/index.js:150:11)
    at /…/app/node_modules/import-in-the-middle/lib/register.js:42:31
    at Array.forEach (<anonymous>)
    at register (/…/app/node_modules/import-in-the-middle/lib/register.js:42:15)
    at file:///…/app/node_modules/pg-pool/index.js?iitm=true:25:1
    at ModuleJob.run (node:internal/modules/esm/module_job:271:25)

Node.js v22.14.0

[...]

PR #2807, opened 2025-04-29
fix(pg): fix instrumentation of ESM-imported pg-pool

maryliag:
The ESM imported top-level object is a Module Namespace Object per ECMA262 spec that is, apparently, incompatible with the 'PG' class object that 'pg' returns. The ".defaults" holds the equivalent of require('pg').

Fixes: #2759


pichlermarc:
Thanks for fixing this 🙂 
Could you also add a test akin to the below one for `pg-pool` to avoid regressions?

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/d9e757f585b346767406eabd4f491399dafc15b8/plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts#L1098-L1125

Doing so would also check the box in #1942

trentm (approved):
Thanks!

If you are super-bored sometime ;) a test for this (similar to the runTestFixture usage in test/pg.test.ts) would be great. Then we could check off instrumentation-pg in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1942#issue-2135458236

trentm:
Oh, duh, Marc said the *exact* same thing already.

maryliag:
Yes, I will be adding the test shortly :D

trentm:
Long notes on that ^^ commit:

Setup for testing (docker cmd from "packages/opentelemetry-test-utils/src/test-utils.ts"):

```
% docker run --rm -d --name otel-postgres -p 54320:5432 -e POSTGRES_PASSWORD=postgres postgres:16-alpine
bff6acac2f00014797721861ae2feda794cfb3eeec17988424f4b4bddfeafa9f

% docker logs -f otel-postgres
...
```

One intentional design of the `runTestFixture` thing is that you can run the test fixture *outside* of the test suite to play with it directly. So, for example:

```
% time node --experimental-loader=@opentelemetry/instrumentation/hook.mjs test/fixtures/use-pg-pool.mjs
...
node --experimental-loader=@opentelemetry/instrumentation/hook.mjs  0.37s user 0.14s system 4% cpu 10.525 total
```

The spans are emitted almost immediately.
Then a 10s hang before exiting.
The 10s run time is ... some timeout?

Yup, it is this one:
https://github.com/brianc/node-postgres/blob/6be857e9d3096d526566ee20fdb68c831c4ada03/packages/pg-pool/index.js#L98-L100


```diff
diff --git a/plugins/node/opentelemetry-instrumentation-pg/test/fixtures/use-pg-pool.mjs b/plugins/node/opentelemetry-instrumentation-pg/test/fixtures/use-pg-pool.mjs
index b210666e..de1e23da 100644
@@ -52,6 +52,7 @@ await tracer.startActiveSpan('test-span', async span => {
       assert.ifError(err)
     });
     release();
+    pgPool.end();
     span.end();
     sdk.shutdown();
   });
```

and now it exits immediately after exporting the spans.

This should fix this error currently being seen in CI:

```
   1) pg-pool (ESM)
       should work with ESM usage:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts)
      at listOnTimeout (node:internal/timers:573:17)
      at processTimers (node:internal/timers:514:7)
```

I assume that 2000ms timeout is from *mocha*. Yup, https://mochajs.org/#command-line-usage shows a default timeout of 2000ms.


Now to run again in the test suite:

```
% RUN_POSTGRES_TESTS=true npx mocha test/pg-pool.test.ts
...
  pg-pool (ESM)
    1) should work with ESM usage


  16 passing (421ms)
  1 failing

  1) pg-pool (ESM)
       should work with ESM usage:

+ actual - expected

+ 'test-span'
- 'pgPool.connect'
      + expected - actual

      -test-span
      +pgPool.connect

      at Object.checkCollector (test/pg-pool.test.ts:944:16)
      at done (/Users/trentm/tm/opentelemetry-js-contrib7/packages/opentelemetry-test-utils/src/test-fixtures.ts:262:24)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I assume this is about the *order* of the emitted spans.

Let's look at the emitted spans with this diff:

```diff
diff --git a/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts b/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
index b06d89ff..af772c41 100644
--- a/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
+++ b/plugins/node/opentelemetry-instrumentation-pg/test/pg-pool.test.ts
@@ -917,7 +917,8 @@ describe('pg-pool', () => {
   });
 });

-describe('pg-pool (ESM)', () => {
+// XXX
+describe.only('pg-pool (ESM)', () => {
   after(done => {
     done();
   });
@@ -938,6 +939,9 @@ describe('pg-pool (ESM)', () => {
         const spans = collector.sortedSpans;

         assert.strictEqual(spans.length, 3);
+        console.log('XXX spans:');
+        spans.forEach((s) => { delete s.attributes; delete s.resource; })
+        console.dir(spans, { depth: 50 })

         assert.strictEqual(spans[0].name, 'pgPool.connect');
         assert.strictEqual(spans[0].kind, 3);
```

Then re-run (edited for brevity):

```js
[
  {
    traceId: 'f9809b89d5298f027b948eaee59ad496',
    spanId: '90da3c95adbe1a47',
    name: 'test-span',
    kind: 1,
    startTimeUnixNano: '1746568315750000000',
    endTimeUnixNano: '1746568315778829875',
  },
  {
    traceId: 'f9809b89d5298f027b948eaee59ad496',
    spanId: '8042003c2392a88c',
    parentSpanId: '90da3c95adbe1a47',
    name: 'pg-pool.connect',
    kind: 3,
    startTimeUnixNano: '1746568315750000000',
    endTimeUnixNano: '1746568315777686833',
  },
  {
    traceId: 'f9809b89d5298f027b948eaee59ad496',
    spanId: '79b563b4c489133d',
    parentSpanId: '8042003c2392a88c',
    name: 'pg.connect',
    kind: 3,
    startTimeUnixNano: '1746568315751000000',
    endTimeUnixNano: '1746568315777938000',
  }
]
```

Actually... we don't have a span for the actual query.
I believe this is because we are ending the pool and client before the query can return:

```js
    pgPool.query('SELECT NOW()', (err, res) => {
      assert.ok(res);
      assert.ifError(err)
    });
    release();
    pgPool.end();
    span.end();
    sdk.shutdown();
```

How about this:

```ts
await tracer.startActiveSpan('test-span', async span => {
  const client = await pgPool.connect();
  try {
    const res = await pgPool.query('SELECT NOW()');
    assert.ok(res);
    console.log('rows:', res.rows);
  } finally {
    client.release();
    pgPool.end();
    span.end();
    sdk.shutdown();
  }
});
```

Now running the tests (with `console.dir(spans.map(s => s.name), { depth: 50 })`) we get *six* spans:

```
[
  'test-span',
  'pg-pool.connect',
  'pg.connect',
  'pg-pool.connect',
  'pg.connect',
  'pg.query:SELECT postgres'
]
```

I wonder if the two sets of `*.connect` are because pg-pool does a pool size of 2 by default.
When I try to configure the Pool class with `max: 1` it gets unhappy, so /shrug.
We can write the test assertions to expect these 6 spans.

# diff PG dbName between local tests and CI

Let's also align the Postgres config used between CI and local testing. Let's
use the CI name: `POSTGRES_DB: otel_pg_database`. (Better, because grepping for that name won't have false positives.)

The *start* at changes for that are:

```diff
diff --git a/packages/opentelemetry-test-utils/src/test-utils.ts b/packages/opentelemetry-test-utils/src/test-utils.ts
index a64c9ad5..28a6aa92 100644
--- a/packages/opentelemetry-test-utils/src/test-utils.ts
+++ b/packages/opentelemetry-test-utils/src/test-utils.ts
@@ -43,7 +43,7 @@ const dockerRunCmds = {
   mysql:
     'docker run --rm -d --name otel-mysql -p 33306:3306 -e MYSQL_ROOT_PASSWORD=rootpw -e MYSQL_DATABASE=test_db -e MYSQL_USER=otel -e MYSQL_PASSWORD=secret mysql:5.7 --log_output=TABLE --general_log=ON',
   postgres:
-    'docker run --rm -d --name otel-postgres -p 54320:5432 -e POSTGRES_PASSWORD=postgres postgres:16-alpine',
+    'docker run --rm -d --name otel-postgres -p 54320:5432 -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=otel_pg_database postgres:16-alpine',
   redis: 'docker run --rm -d --name otel-redis -p 63790:6379 redis:alpine',
 };

diff --git a/plugins/node/opentelemetry-instrumentation-pg/test/fixtures/use-pg-pool.mjs b/plugins/node/opentelemetry-instrumentation-pg/test/fixtures/use-pg-pool.mjs
index b210666e..5bdc971a 100644
--- a/plugins/node/opentelemetry-instrumentation-pg/test/fixtures/use-pg-pool.mjs
+++ b/plugins/node/opentelemetry-instrumentation-pg/test/fixtures/use-pg-pool.mjs
@@ -26,7 +26,7 @@ import { PgInstrumentation } from '../../build/src/index.js';
 const CONFIG = {
   user: process.env.POSTGRES_USER || 'postgres',
   password: process.env.POSTGRES_PASSWORD || 'postgres',
-  database: process.env.POSTGRES_DB || 'postgres',
+  database: process.env.POSTGRES_DB || 'otel_pg_database',
   host: process.env.POSTGRES_HOST || 'localhost',
   port: process.env.POSTGRES_PORT
     ? parseInt(process.env.POSTGRES_PORT, 10)
```

trentm:
`unit-test (18)` failure is:

```
> @opentelemetry/auto-instrumentations-node@0.58.1 test
> nyc mocha 'test/**/*.ts'



  Register
    1) can load auto instrumentation from command line
    ✔ shuts down the NodeSDK when SIGTERM is received (1692ms)
```

This is failing on main for me locally. That test case is just really slow. Without `nyc` (on my machine):

```
% npx mocha test/register.test.ts


  Register
    ✔ can load auto instrumentation from command line (1305ms)
    ✔ shuts down the NodeSDK when SIGTERM is received (1194ms)
```

and then nyc pushes it over the edge sometimes:

```
% npx nyc mocha test/register.test.ts


  Register
    1) can load auto instrumentation from command line
    2) shuts down the NodeSDK when SIGTERM is received
```

instr-runtime-node *was* added to auto-instrumentations-node recently. It could be that is the straw that broke the camel's back.

trentm (approved):

pichlermarc (approved):
thank you for fixing this 🙂

Issue #2377, opened 2024-08-07
Use updated attributes with ATTR prefix in all packages


JamieDanielson:
Similar to https://github.com/open-telemetry/opentelemetry-js/issues/4896 in core, but in contrib

This is part of [Semantic Convention Update: Tracking Issue #4572](https://github.com/open-telemetry/opentelemetry-js/issues/4572).

After the new semantic conventions package is ready, all semantic convention attributes will need to be updated to use the new attributes, located in the new experimental(/incubating? tbd) directory.

This issue is for changing `SEMRESATTRS_` and `RESATTRS_` prefixes to just `ATTR_` for attributes. No new attributes should be added in this change.

The next changes will add the env var, update instrumentations for stable http attributes, and use new metric attributes.

Currently blocked on release of semantic conventions

trentm:
Need to sort out https://github.com/open-telemetry/opentelemetry-js/issues/5182 first (how to handle unstable semconv attributes).

trentm:
The set to work through:

```
% npm-ws-ls | while read pkg; do echo -n "$pkg: "; rg 'SEM(RES)?ATTRS_' $pkg | wc -l; done | sort -n -k2
...
detectors/node/opentelemetry-resource-detector-container:        2
metapackages/auto-instrumentations-node:        2
plugins/node/opentelemetry-instrumentation-bunyan:        2
plugins/node/opentelemetry-instrumentation-graphql:        2
plugins/node/opentelemetry-instrumentation-pino:        2
plugins/node/opentelemetry-instrumentation-router:        2
detectors/node/opentelemetry-resource-detector-instana:        6
plugins/node/opentelemetry-instrumentation-connect:        6
plugins/node/opentelemetry-instrumentation-hapi:        6
plugins/node/opentelemetry-instrumentation-nestjs-core:        6
plugins/node/opentelemetry-instrumentation-restify:        7
plugins/node/opentelemetry-instrumentation-express:       10
plugins/node/opentelemetry-instrumentation-fastify:       12
plugins/node/opentelemetry-instrumentation-koa:       12
plugins/web/opentelemetry-instrumentation-document-load:       12
detectors/node/opentelemetry-resource-detector-gcp:       18
plugins/node/instrumentation-cucumber:       18
plugins/node/opentelemetry-instrumentation-knex:       18
plugins/node/opentelemetry-instrumentation-aws-lambda:       19
[...]

PR #2819, opened 2025-05-09
refactor(examples): update semconv usage to ATTR_ exports

trentm:
Refs: #2377

Note: I did *not* work through the steps for each example. Clearly many of these examples are out of date and need work. However, there is no way this change will add *more* breakage, examples/**/package.json include `"@opentelemetry/semantic-conventions": "^1.27.0",` and version 1.27.0 has the `ATTR_SERVICE_NAME` export being used.


JamieDanielson (approved):
Thanks for this. We have a start at describing the updates needed for examples but haven't come back around to it yet. This is still a good change

PR #2821, opened 2025-05-13
feat(test-utils): add assertion helper for faas resources

kaspernissen:
I'm currently trying to expand the resource detector for gcp and aligning it with the rest of the gcp resource detectors in other languages, that currently sets `faas.name`, `faas.version`, and `faas.instance` for Cloud Run services. 

I have another PR open on this topic: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2818 where I will be using this helper. 

This PR adds an assertion helper to the `opentelemtry-test-utils` package. Similarly to the rest of the assertion helpers it uses the semantic conventions and validates the inputs.

### DIFF
--- a/packages/opentelemetry-test-utils/src/resource-assertions.ts
+++ b/packages/opentelemetry-test-utils/src/resource-assertions.ts
@@ -26,6 +26,9 @@ import {
   SEMRESATTRS_CONTAINER_IMAGE_NAME,
   SEMRESATTRS_CONTAINER_IMAGE_TAG,
   SEMRESATTRS_CONTAINER_NAME,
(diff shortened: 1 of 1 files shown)


pichlermarc:
Hi @kaspernissen - I would rather not have this be part of the `@opentelemetry/contrib-test-utils` unless it's commonly used across multiple packages. I think for now it would be better to have the assertion helper in `@opentelemetry/resource-detector-gcp` instead.

kaspernissen:
Sure, I can do that for now. Reasoning behind putting it in the common, was to make it available for other packages. 
I will close this PR and add it with the original PR 👍 Thanks for your swift reply!

PR #2840, opened 2025-05-21
chore: reduce duplicate dep installs with 'npm dedupe'

trentm:
```bash
npm ci
npm dedupe
```

See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2671#discussion_r2101328910 for the motivation for my doing this. From skimming this change I notice that our `typescript@5.0.4` usage was being installed *in every separate package* rather than at the top-level because of some history of how we installed this version and other transitive deps have a dep on a different typescript version.

A surprising result is how much smaller the `npm ci` result is. This is on a clean checkout and `npm ci`:

```
% du -sh *
916M	opentelemetry-js-contrib-after
3.7G	opentelemetry-js-contrib-before
```


deejay1 (approved):

david-luna (approved):
Awesome! Thank you :)

I'll check if we have a similar situation in contrib repo.

Issue #2841, opened 2025-05-22
Replace aws.region with cloud.region in AWS SDK instrumentation


yuliia-fryshko:
Hi team, it is my first issue here, so let me know if I've missed anything or should follow a different practices. Thank you in advance.

### Is your feature request related to a problem? Please describe

I was working on Bedrock instrumentation and noticed that we using `aws.region` as a span attribute. However, per the updated [semantic convention](https://github.com/open-telemetry/semantic-conventions/pull/2238), `aws.region` is replaced by `cloud.region`.
   

### Describe the solution you'd like to see

I'd like to replace `aws.region` with `cloud.region` in the AWS SDK instrumentation. However, this is a breaking change, since existing users might rely on `aws.region` being present in their pipeline or tests. 

### Describe alternatives you've considered

- Keeping both attributes temporary to allow smoother migration.
- Replacing `aws.region` immediately with clear release notes as a breaking change.
- Opening PR only for the Bedrock instrumentation as a start.  

### Additional context

I would like to make sure any refactor here is aligned with project expectations.

yuliia-fryshko:
I created [PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2842/files) to address this issue. 

I will appreciate your feedback and happy to adjust the implementation based on your suggestions.

PR #2842, opened 2025-05-22
refactor(opentelemetry-instrumentation-aws-sdk): rename aws.region to…

yuliia-fryshko:
Fixes [#2841](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2841)

- This PR updates the AWS SDK instrumentation to align with the latest semantic conventions, which introduced generic and standardized `cloud.region` instead of vendor specific `aws.region`. 

- Replaced all usage of `aws.region` with `cloud.region`.  
- This change affects instrumentation for AWS SDK including Bedrock and S3.

##Notes

- This is a breaking change for users relying on the presence of the `aws.region` attribute.


plugins/node/opentelemetry-instrumentation-aws-sdk/src/enums.ts
```diff
@@ -15,7 +15,7 @@
  */
 export enum AttributeNames {
   AWS_OPERATION = 'aws.operation',
-  AWS_REGION = 'aws.region',
+  CLOUD_REGION = 'cloud.region',
```
trentm, line 18: `cloud.region` *does* have a constant in `@opentelemetry/semantic-conventions`. However, we can defer using that constant (and the others in this instrumentation) to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2377

The current constant using the old, deprecated naming (`SEMRESATTRS_CLOUD_REGION`) is/was intended for use as a *Resource* attribute. That was before https://github.com/open-telemetry/semantic-conventions/pull/2238 expanded the use of `cloud.region` beyond resource attributes.  That's why I'm not suggesting this PR use `SEMRESATTRS_CLOUD_REGION`.

trentm (approved):
LGTM. I updated the PR title (following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)) to indicate that this is a breaking change. That's necessary to ensure the next version is 0.54.0 rather than 0.53.1.

Code owners: any strong opinions here?

pichlermarc:
> Code owners: any strong opinions here?

cc @jj22ee @trivikr (component owners)

jj22ee (approved):
LGTM, no strong opinions from me.

PR #2843, opened 2025-05-22
fix(express): Ensure 404 routes don't attach route attribute

s1gr1d:

This is an upstream fix for https://github.com/getsentry/sentry-javascript/pull/16211

Instead of adding `/` to non-existing routes, it sets the attribute to `undefined`, which makes it clear it's a non-existing route and not the index route.

--- 

A tricky thing is that the layers look the same if a non-existing route is requested or the root is requested: `['/', '/']`.
- `/existing` --> `['/', '/existing']`
- `/non-existing` --> `['/', '/']`
- `/` --> `['/', '/']`

So in order to know if it was a 404 route, a utility function that does some additional checking was added.


pichlermarc (approved):
LGTM! thank you for your contribution @s1gr1d 🙂 

cc @raphael-theriault-swi, @JamieDanielson, @pkanal (component owners)

pichlermarc:
oh - looks like there are some lint errors, you can fix them by running `npm run lint:fix` and then pushing the resulting diff 🙂

raphael-theriault-swi (approved):

s1gr1d:
I have to look into this again. I see another test is failing. It doesn't add `http.route` either when the route is `/` (but this should be added in this case).

s1gr1d:
The problem was that the layers look the same if a non-existing route is requested or the root is requested: ['/', '/']`. 

So in order to know if it was a 404 route, I added a utility function that does some additional checking. I made sure that the existing tests still work and added some more tests for the utility function.

pichlermarc (commented):
Hi @s1gr1d, thanks for following up on this.

It seems that this PR is based on a commit that's a little over 2 months old. Would you mind rebasing it on top of `main` it so that we can see how this would fit in with the current state? 🙂

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -213,3 +213,91 @@ const extractLayerPathSegment = (arg: LayerPathSegment) => {
...
+    return constructedRoute;
+  }
+
+  // For RegExp routes or route arrays, return the constructed route
+  // This handles the case where the route is defined using RegExp or an array
+  if (constructedRoute.includes('/') &&
+      (constructedRoute.includes(',') ||
+       constructedRoute.includes('\\') ||
+       constructedRoute.includes('*') ||
+       constructedRoute.includes('['))) {
```
s1gr1d: This condition is for considering regex routes (without this, the regex tests would fail). This should cover most cases.

There is a previous conversation about this: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2008#discussion_r1538081650

s1gr1d:
@pichlermarc  I rebased it and aligned the code so the "regex" tests don't fail anymore.

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -213,3 +213,93 @@ const extractLayerPathSegment = (arg: LayerPathSegment) => {
...
   return;
 };
+
+export function getConstructedRoute(req: {
+  originalUrl: PatchedRequest['originalUrl'];
+  [_LAYERS_STORE_PROPERTY]?: string[];
+}) {
+  const layersStore: string[] = Array.isArray(req[_LAYERS_STORE_PROPERTY])
+    ? (req[_LAYERS_STORE_PROPERTY] as string[])
+    : [];
```
david-luna, line 223: nit: `_LAYERS_STORE_PROPERTY` is set only in `storeLayerPath` function. Meaning that if defined it must be an array. So this const assignment could be simplified to.

```ts
const layersStore = req[_LAYERS_STORE_PROPERTY] || [];
```

I see you added a test specifically for a scenario where that property is defined with a non-array value. Probably to avoid a case where another piece of code outside the instrumentation modifies the property. If you want the extra safety of `Array.isArray` it's okay but there is no need to cast types. TS resolves the type properly from the function signature.

```ts
const layersStore = Array.isArray(req[_LAYERS_STORE_PROPERTY])
    ? req[_LAYERS_STORE_PROPERTY]
    : [];
```

NOTE: maybe if the instrumentation uses a `Symbol` instead of a string it will ensure no other code could change the layer store

plugins/node/opentelemetry-instrumentation-express/src/utils.ts
```diff
@@ -213,3 +213,93 @@ const extractLayerPathSegment = (arg: LayerPathSegment) => {
...
+ *
+ * @param req - The Express request object with layers store
+ * @param layersStoreProperty - The property name where layer paths are stored
+ * @returns The matched route string or undefined if no valid route is found
+ */
+export function getActualMatchedRoute(req: {
+  originalUrl: PatchedRequest['originalUrl'];
+  [_LAYERS_STORE_PROPERTY]?: string[];
+}): string | undefined {
+  const layersStore: string[] = Array.isArray(req[_LAYERS_STORE_PROPERTY])
```
david-luna, line 249: nit: same here

david-luna (approved):
LGTM :) just a couple of nits

PR #2845, opened 2025-05-22
feat: add Bedrock InvokeModelWithResponseStream instrumentation

yuliia-fryshko:

Adds instrumentation of the **InvokeModelWithResponseStreamCommand** in the **AWS Bedrock SDK**.

- Captures GenAI semantic span attributes for **Amazon Titan, Anthropic Claude, Amazon Nova** models.
- Ensures span.end() is deferred until the full stream is consumed. 
- `instrumentAsyncIterable` is used to inspect streamed chunks in real time and extract relevant telemetry. 
- Includes unit tests for Titan, Claude and Nova models.


plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -102,6 +102,13 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
         return this.requestPreSpanHookConverse(request, config, diag, true);
       case 'InvokeModel':
         return this.requestPreSpanHookInvokeModel(request, config, diag);
+      case 'InvokeModelWithResponseStream':
+        return this.requestPreSpanHookInvokeModelWithResponseStream(
```
trentm, line 106: Is there a reason to not re-use `requestPreSpanHookInvokeModel`: add a 4th `isStream` argument and pass in `false` for 'InvokeModel', `true` for 'InvokeModelWithResponseStream', and then make the minor update to the implementation?  This is how it was done for 'Converse' and 'ConverseStream'.

It looks to me like the `requestPreSpanHookInvokeModel` and `requestPreSpanHookInvokeModelWithResponseStream` functions are almost identical ... except that the latter doesn't have blocks for 'meta.llama', 'cohere.*', and 'mistral'.

yuliia-fryshko: Thanks for the suggestion, @trentm ! You are absolutely right! I've updated the code to consolidate `requestPreSpanHookInvokeModel` and `requestPreSpanHookInvokeModelWithResponseStream` into a single method using `isStream` parameter as you suggested.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -579,4 +590,134 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+
+  private async responseHookInvokeModelWithResponseStream(
+    response: NormalizedResponse,
+    span: Span,
+    tracer: Tracer,
+    config: AwsSdkInstrumentationConfig
+  ): Promise<any> {
+    const stream = response.data?.body;
+    const modelId = response.request.commandInput?.modelId;
+    if (!stream || !span.isRecording()) return;
```
jj22ee:
```suggestion
    if (!stream) return;
```

`!span.isRecording()` is already checked before `responseHookInvokeModelWithResponseStream()` is called

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -579,4 +590,134 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+          yield item;
+        }
+      } finally {
+        span.end();
+      }
+    })();
+    return response.data;
+
+    // Tap into the stream at the chunk level without modifying the chunk itself.
+    function instrumentAsyncIterable<T>(
```
jj22ee: Wouldn't it be more efficient to declare these following functions outside of this member function scope?

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
yuliia-fryshko: You are right, @jj22ee! I removed this unnecessary check

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
yuliia-fryshko: Thank you, @jj22ee , for the review and good catches! Yes, these helper functions could be declared outside the method, at the class level, instead of being re-created every time the method is called. 
Changed!

JacksonWeber:
Update this to pull in the main branch and I can take another look over. Thanks!

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -579,4 +590,145 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+            BedrockRuntimeServiceExtension.recordClaudeAttributes(
+              parsedChunk,
+              span
+            );
+          } else if (modelId.includes('amazon.nova')) {
+            BedrockRuntimeServiceExtension.recordNovaAttributes(
+              parsedChunk,
+              span
+            );
+          }
```
trentm, line 656: The non-stream `responseHookInvokeModel()` is adding some attributes for a few more models: `'meta.llama'`, `'cohere.command-r'`, etc.
Should this function also add relevant attributes for those models? Or is it possible streaming is not supported for those models?

Also, perhaps the `recordNovaAttributes()`, `recordClaudeAttributes()`, etc. methods could be used by both `responseHookInvokeModel` and `responseHookInvokeModelWithResponseStream`.  Or perhaps all the `record*Attribute()` methods could be moved to one `setInvokeModelResponseAttributes()` that is used by both `responseHookInvokeModel*()` methods.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -579,4 +590,145 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+        }
+      );
+    // Replace the original response body with our instrumented stream.
+    // - Defers span.end() until the entire stream is consumed
+    // This ensures downstream consumers still receive the full stream correctly,
+    // while OpenTelemetry can record span attributes from streamed data.
+    response.data.body = (async function* () {
+      try {
+        for await (const item of wrappedStream) {
```
trentm: IIUC, the result here is that *two* async iterators are created. Could the chunk handling above be moved into this `for await ...` and then not bother having a `wrappedStream` at all? I haven't tried this.

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -579,4 +590,145 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+    };
+  }
+
+  private static parseChunk(bytes?: Uint8Array): any {
+    if (!bytes || !(bytes instanceof Uint8Array)) return null;
+    try {
+      const str = Buffer.from(bytes).toString('utf-8');
+      return JSON.parse(str);
+    } catch (err) {
+      console.warn('Failed to parse streamed chunk', err);
```
trentm: Shouldn't use `console` in instrumentation code. Instead use the DiagLogger that every intsrumentation instance has.

```suggestion
      this._diag.warn('Failed to parse streamed chunk', err);
```

This will mean you'll need to not use `static` for this method.

pichlermarc:
@yuliia-fryshko do you have bandwidth to work on this? Looks like there are some unaddressed comments that are blocking this from getting merged.

yuliia-fryshko:
> @yuliia-fryshko do you have bandwidth to work on this? Looks like there are some unaddressed comments that are blocking this from getting merged.

Hi @pichlermarc ! Sorry for the long reply, I just back from the vacation today and I will try to address all comments today

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
yuliia-fryshko: Done

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
yuliia-fryshko: Done

plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/bedrock-runtime.ts
yuliia-fryshko: Thank you, @trentm , for the review . I addressed all your comments and would appreciate next round of a review. Thanks

pichlermarc:
Hi @trivikr @blumamir @jj22ee, please take a look. Thanks! :)

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -579,4 +592,214 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+    // This ensures downstream consumers still receive the full stream correctly,
+    // while OpenTelemetry can record span attributes from streamed data.
+    response.data.body = async function* (
+      this: BedrockRuntimeServiceExtension
+    ) {
+      try {
+        for await (const chunk of stream) {
+          const parsedChunk = this.parseChunk(chunk?.chunk?.bytes);
+
+          if (!parsedChunk) return;
```
trentm: @yuliia-fryshko 

nit: If, for whatever reason, a chunk cannot be parsed for adding otel data, should we perhaps *continue* to yield the chunks and eventually span.end() rather than `return;` here?

trentm (approved):
I'm good with this PR now. Sorry, I realize this has dragged on for a long while.

Some notes:

- I have a nit about error handling (see below).
- At this point, because `gen_ai` semconv moves quickly enough, some of the semconv vars here are outdated. E.g. `gen_ai.system` is now replaced with `gen_ai.provider.name`.  However, this instrumentation was already using the older gen-ai semconv version and the current gen-ai semconv docs (https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/) suggest that existing instrumentations should stick to their current version by default:

> SHOULD NOT change the version of the GenAI conventions that they emit by default. Conventions include, but are not limited to, attributes, metric, span and event names, span kind and unit of measure.
> SHOULD introduce an environment variable OTEL_SEMCONV_STABILITY_OPT_IN ...


So any updates of semconv version could be done in a separate PR.

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
yuliia-fryshko: I agree, it will be better to close the span in this case. Done

yuliia-fryshko:
> I'm good with this PR now. Sorry, I realize this has dragged on for a long while.
> 
> Some notes:
> 
> * I have a nit about error handling (see below).
> * At this point, because `gen_ai` semconv moves quickly enough, some of the semconv vars here are outdated. E.g. `gen_ai.system` is now replaced with `gen_ai.provider.name`.  However, this instrumentation was already using the older gen-ai semconv version and the current gen-ai semconv docs (https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/) suggest that existing instrumentations should stick to their current version by default:
> 
> > SHOULD NOT change the version of the GenAI conventions that they emit by default. Conventions include, but are not limited to, attributes, metric, span and event names, span kind and unit of measure.
> > SHOULD introduce an environment variable OTEL_SEMCONV_STABILITY_OPT_IN ...
> 
> So any updates of semconv version could be done in a separate PR.

Thank you, @trentm ! 

I created [an issue ](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3152)to review and upgrade all fields with latest semantic convention. Happy to take it after this MR will be merged.

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
trentm: I think your change in commit 80d93e3 is not correct. It would result in `span.end()` being called and then an attempt to call `BedrockRuntimeServiceExtension.record*Attributes(...)` which would likely break.

I've change it to what I think it should be in commit ee6f51e9.
@yuliia-fryshko Please let me know if my change looks inappropriate.

plugins/node/instrumentation-cucumber/test/current.feature
trentm, line 1: Just noticed that this file got added accidentally at some point.

trentm: removed in commit 35143a44

PR #2836, opened 2025-05-19
chore: upgrade lerna

david-luna:
## Which problem is this PR solving?

This is a follow up PR of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2493. It updates `lerna` to version v8.2.2, which indirectly updates `nx`. Dev scripts are using `nx` directly but publish workflows and tasks keep using `lerna`

## Changes

- remove unnecessary scripts in packages (precompile & prewatch)
- upgrade `lerna` dependency
- update `scripts/peer-api-check.mjs` to work at the root level and the dependant workflows
- update `scripts/check-release-please.mjs`, simplification
- added a `setup:dev` npm script in each package for people who does not want to compile all project to start developing
- update docs

Supersedes: #2526

PR #2854, opened 2025-05-23
fix: remove lerna legacy options

david-luna:

After the upgrade of lerna the `release-please` workflow fails due to legacy configuration that newer version does not accept. The failure message is

```
npx lerna publish from-package --no-push --no-private --no-git-tag-version --no-verify-access --yes
  shell: /usr/bin/bash -e {0}
  env:
    NPM_CONFIG_USERCONFIG: /home/runner/work/_temp/.npmrc
    NODE_AUTH_TOKEN: ***
    NPM_CONFIG_PROVENANCE: true
lerna notice cli v[8](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/15204321834/job/42764004634#step:10:9).2.2
lerna ERR! ECONFIGWORKSPACES The "useWorkspaces" option has been removed. By default lerna will resolve your packages using your package manager's workspaces configuration. Alternatively, you can manually provide a list of package globs to be used instead via the "packages" option in lerna.json
```

Also when fixing this issue another message appears about the `no-verify-access` CLI option.

```
npx lerna publish from-package --no-push --no-private --no-git-tag-version --no-verify-access --yes
lerna notice cli v8.2.2
lerna WARN verify-access --verify-access=false and --no-verify-access are no longer needed, because the legacy preemptive access verification is now disabled by default. Requests will fail with appropriate errors when not authorized correctly.
```

The option is needed since now is the default

- remove `useWorkspaces` from `lerna.json`
- update `lerna publish` command options
- remove `RELEASING.md` as commented in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2836#discussion_r2100832084


pichlermarc (approved):
looks good!
thanks for following up on the broken workflow 🙂

PR #2855, opened 2025-05-25
fix(detector-aws): extract full container ID from ECS Fargate cgroup

### LINKED ISSUE #2455: [detectors-aws] incomplete containerId detected for AWS ECS Fargate
```
    "@opentelemetry/core": "^1.26.0",
    "@opentelemetry/exporter-trace-otlp-proto": "^0.53.0",
    "@opentelemetry/instrumentation-aws-sdk": "^0.44.0",
    "@opentelemetry/instrumentation-http": "^0.53.0",
    "@opentelemetry/instrumentation-ioredis": "^0.43.0",
    "@opentelemetry/instrumentation-winston": "^0.40.0",
    "@opentelemetry/resource-detector-aws": "^1.6.1",
    "@opentelemetry/resource-detector-container": "^0.4.1",
    "@opentelemetry/resources": "^1.26.0",
    "@opentelemetry/sdk-node": "^0.53.0",
    "@opentelemetry/sdk-trace-node": "^1.26.0",
    "@opentelemetry/winston-transport": "^0.6.0",
```

```
20
```

### What did you do?

I initialize my tracing similar to the setup below as one of the first actions within my Node.js express server. I bundle the whole application via Docker and deploy it to AWS ECS on Fargate.

```
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-proto";
import { AwsInstrumentation } from "@opentelemetry/instrumentation-aws-sdk";
import {
  envDetector,
  processDetector,
  Resource,
} from "@opentelemetry/resources";
import { awsEcsDetectorSync } from "@opentelemetry/resource-detector-aws";
import { containerDetector } from "@opentelemetry/resource-detector-container";
import { env } from "@/src/env.mjs";

  const sdk = new NodeSDK({
    resource: new Resource({
      "service.name": env.OTEL_SERVICE_NAME,
    }),
    traceExporter: new OTLPTraceExporter({
      url: `${env.OTEL_EXPORTER_OTLP_ENDPOINT}/v1/traces`,
    }),
    instrumentations: [
      new AwsInstrumentation(),
    ],
    resourceDetectors: [
      envDetector,
      processDetector,
      awsEcsDetectorSync,
      containerDetector,
    ],
  });

  sdk.start();
}

```

I would expect to see something like `<taskId>-<containerId>` as the `container.id` in my span attributes. This would fit the conventions that observability vendors like Datadog use. For a task named `c23e5f76c09d438aa1824ca4058bdcab` I'd expect to see something like `c23e5f76c09d438aa1824ca4058bdcab-1234678` for a single container.

### What did you see instead?

As part of https://github.com/aws/amazon-ecs-agent/issues/1119 AWS apparently shifted their cgroup naming convention to something like `/ecs/<taskId>/<taskId>-<containerId>`. Given the 64 character limit it usually cuts off in the middle of taskId. For the example above, this would yield a `container.id` value like `438aa1824ca4058bdcab/c23e5f76c09d438aa1824ca4058bdcab-1234678`.

Is there some way to consistently receive only the part after the `/` in the cgroup name, i.e. the last chunk? Happy to contribute in case this seems desirable. It should probably follow a regex based approach like in https://github.com/DataDog/dd-trace-js/pull/1176.

seongpil0948:

AWS ECS Fargate changed their cgroup naming convention to `/ecs/<taskId>/<taskId>-<containerId>` format. The existing container ID extraction logic only takes the last 64 characters, causing truncation in the middle of the taskId and resulting in incomplete container IDs like `438aa1824ca4058bdcab/c23e5f76c09d438aa1824ca4058bdcab-1234678` instead of the expected full `c23e5f76c09d438aa1824ca4058bdcab-1234678`.

This breaks compatibility with observability vendors like DataDog that require the complete `taskId-containerId` format for proper container identification.

- **Enhanced container ID extraction**: Extract the last path segment from cgroup paths instead of truncating to 64 characters
- **New `_extractContainerIdFromLine()` method**: Handles various container runtime formats (Docker, containerd, CRI-O) with proper prefix/suffix removal
- **Flexible validation**: Uses permissive regex pattern `[a-zA-Z0-9\-_]+` to support real-world container ID formats
- **Backward compatibility**: Maintains fallback to original 64-char logic for legacy formats
- **Comprehensive tests**: Added test cases covering DataDog compatibility scenarios and edge cases

**Supported formats:**
- New ECS Fargate: `/ecs/<taskId>/<taskId>-<containerId>` → `taskId-containerId`
- Docker: `/docker/containerid.scope` → `containerid` 
- Containerd: `system.slice:cri-containerd:containerid` → `containerid`
- UUID format: `/uuid/34dc0b5e-626f-2c5c-4c51-70e34b10e765` → `34dc0b5e-626f-2c5c-4c51-70e34b10e765`

Fixes #2455


pichlermarc:
@seongpil0948 looks like this PR is in draft, is this intentional? 🤔

seongpil0948:
> @seongpil0948 looks like this PR is in draft, is this intentional? 🤔

i'm sorry to late ..  🫡

david-luna:
@seongpil0948 

There has been a re-organization of the packages in this repository, see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2928

Would you mind to sync your PR with main branch?

dyladan:
Looks like this needs reviews. The conflicts are just from package naming. @jj22ee can you please take a look?

seongpil0948:
i worked on a new branch based on new structured packages
thanks all

jj22ee:
Hi @seongpil0948, wondering why not opt to use a more automated regex based approach (as mentioned in the issue, something like in https://github.com/DataDog/dd-trace-js/pull/1176, which is now [here](https://github.com/DataDog/dd-trace-js/blob/16e5800ded62f8a87e44707685ffad0859cb430c/packages/dd-trace/src/exporters/common/docker.js#L15))?

IMO that would help alleviate some of the burden in the current PR which adds logic to "track and remove known prefixes", execute step-by-step extraction+sanitizing of the lastSegment, and perform two sanity validations based on the length of the extracted segment based on what we think is reasonable. Regex would be more "straight to the point" at a glance, easier to maintain.

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
```diff
@@ -176,6 +174,53 @@ export class AwsEcsDetector implements ResourceDetector {
...
+   * Returns the last segment after the final '/' which should be the complete container ID.
+   */
+  private _extractContainerIdFromLine(line: string): string | undefined {
+    if (!line) {
+      return undefined;
+    }
+
+    // Primary pattern: Match /ecs/taskId/taskId-containerId format (new ECS Fargate)
+    // This captures the full taskId-containerId part after the last slash
+    const ecsPattern = /\/ecs\/[a-zA-Z0-9-]+\/([a-zA-Z0-9-]+)$/;
```
jj22ee:
```suggestion
    const ecsPattern = /\/ecs\/[a-fA-F0-9-]+\/([a-fA-F0-9-]+)$/;
```

I don't see a need to allow characters that are definitely not part of the taskId/containerId.

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
```diff
@@ -176,6 +174,53 @@ export class AwsEcsDetector implements ResourceDetector {
...
+        lastSegment.length <= 128
+      ) {
+        return lastSegment;
+      }
+    }
+
+    // Legacy fallback: original logic for lines that are just the container ID (64 chars)
+    if (
+      line.length > AwsEcsDetector.CONTAINER_ID_LENGTH &&
+      !line.includes('/')
```
jj22ee:
```suggestion
      line.length > AwsEcsDetector.CONTAINER_ID_LENGTH
```

Let's keep the legacy fallback to be exactly the same as the original logic. In scenarios where the line does include `/`, whether or not original logic is better than `undefined` is a bit blurry...

dyladan:
@jj22ee looks like there's been changes since your last comment. I think this is ready for another review.

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
```diff
@@ -176,6 +174,53 @@ export class AwsEcsDetector implements ResourceDetector {
...
+    }
+
+    // Fallback: Extract last segment after slash for any path-like format
+    const segments = line.split('/');
+    if (segments.length > 1) {
+      const lastSegment = segments[segments.length - 1];
+      if (
+        lastSegment &&
+        lastSegment.length >= 12 &&
+        lastSegment.length <= 128
```
jj22ee: >Extract last segment after slash for any path-like format

I think there is a good chance this may extract a string that is not a container ID. All it takes here are strings as short as 12 characters long after just a slash to pass this logic.

Do you think it'd safe to just start with detecting `/ecs/taskId/taskId-containerId`, with the original logic as fallback?

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
jj22ee: >Match `/ecs/taskId/taskId-containerId` format

1. For `/taskId/`, shouldn't the regex be updated such that we don't expect any `-` in taskId?
2. For `/taskId-containerId`, shouldn't the regex be updated such that we only expect one `-`?

jj22ee: According to https://github.com/aws/amazon-ecs-agent/issues/1119, observed contents of `/proc/self/cgroup` that contains `.../ecs/...` seems to be:
```
9:perf_event:/ecs/task-arn/container-sha
8:memory:/ecs/task-arn/container-sha
7:hugetlb:/ecs/task-arn/container-sha
...
```

So does this mean `/ecs/taskId/taskId-containerId` is not what should be matched? Instead `/ecs/task-arn/container-sha` should be matched instead (I suppose taskId is part of the task-arn). If this is the case, this current logic will not work.

Wondering if you have full examples of `/proc/self/cgroup` content in ECS Fargate environment?

seongpil0948: > I don't see a need to allow characters that are definitely not part of the taskId/containerId.
thank you , i mistook it for string with UUID [Specification](https://docs.docker.com/engine/containers/run/#container-identification)

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
seongpil0948: @jj22ee  First of all, according to the related issue [COMMENT](https://github.com/aws/amazon-ecs-agent/issues/1119#issuecomment-348123884) , it doesn't seem necessary to fix it exactly at 64, but I thought it would be best to guarantee at least 12 characters (short identifier)

the aforementioned `/ecs/` may not include ecs depending on options (refer [COMMENT](https://github.com/aws/amazon-ecs-agent/issues/1119#issuecomment-357529259))
If you are worried about other values, i can separate by CRI like docker

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
seongpil0948: > According to [aws/amazon-ecs-agent#1119](https://github.com/aws/amazon-ecs-agent/issues/1119), observed contents of `/proc/self/cgroup` that contains `.../ecs/...` seems to be:
> 
> ```
> 9:perf_event:/ecs/task-arn/container-sha
> 8:memory:/ecs/task-arn/container-sha
> 7:hugetlb:/ecs/task-arn/container-sha
> ...
> ```
> 
> So does this mean `/ecs/taskId/taskId-containerId` is not what should be matched? Instead `/ecs/task-arn/container-sha` should be matched instead (I suppose taskId is part of the task-arn). If this is the case, this current logic will not work.
> 
> Wondering if you have full examples of `/proc/self/cgroup` content in ECS Fargate environment?

this is the content I reffered to when i worked on it

## When Task ARN 

- ARN: `arn:aws:ecs:ap-northeast-2:725129837589:task/Trader-VPC-Stack-Cluster/447438d8540d49dca93b4f0a488ebe90`
- ContainerID: `447438d8540d49dca93b4f0a488ebe90-1364044452`
	- What Container ID specified in AWS Console

### Output
when excute `cat` in **REAL** ECS Container 
```bash
cat /proc/self/cgroup

11:memory:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
10:devices:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
9:freezer:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
8:blkio:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
7:perf_event:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
6:net_cls,net_prio:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
5:cpuset:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
4:pids:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
3:hugetlb:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
2:cpu,cpuacct:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
1:name=systemd:/ecs/447438d8540d49dca93b4f0a488ebe90/447438d8540d49dca93b4f0a488ebe90-1364044452
```


### Fallback or Compatibility

https://github.com/aws/amazon-ecs-agent/issues/1119#issuecomment-357529259
```
1:blkio:/docker/a4d00c9dd675d67f866c786181419e1b44832d4696780152e61afd44a3e02856
2:cpu:/docker/a4d00c9dd675d67f866c786181419e1b44832d4696780152e61afd44a3e02856
3:cpuacct:/docker/a4d00c9dd675d67f866c786181419e1b44832d4696780152e61afd44a3e02856
```


### Metadata

result of `wget 169.254.170.2/v2/metadata` in ECS

```json
{
  "Cluster": "arn:aws:ecs:ap-northeast-2:123456789012:cluster/MyDemo-VPC-Stack-Cluster",
  "TaskARN": "arn:aws:ecs:ap-northeast-2:123456789012:task/MyDemo-VPC-Stack-Cluster/a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6",
  "Family": "sample-app-dev",
  "Revision": "999",
  "DesiredStatus": "RUNNING",
  "KnownStatus": "RUNNING",
  "Containers": [
    {
      "DockerId": "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6-1234567890",
      "Name": "sample-app-dev",
      "DockerName": "sample-app-dev",
      "Image": "123456789012.dkr.ecr.ap-northeast-2.amazonaws.com/sample/app-dev:production-a1b2c3d",
      "ImageID": "sha256:a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2",
      "Labels": {
        "com.amazonaws.ecs.cluster": "arn:aws:ecs:ap-northeast-2:123456789012:cluster/MyDemo-VPC-Stack-Cluster",
        "com.amazonaws.ecs.container-name": "sample-app-dev",
        "com.amazonaws.ecs.task-arn": "arn:aws:ecs:ap-northeast-2:123456789012:task/MyDemo-VPC-Stack-Cluster/a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6",
        "com.amazonaws.ecs.task-definition-family": "sample-app-dev",
        "com.amazonaws.ecs.task-definition-version": "999"
      },
      "DesiredStatus": "RUNNING",
      "KnownStatus": "RUNNING",
      "Limits": {
        "CPU": 1024,
        "Memory": 2048
      },
      "CreatedAt": "2025-08-08T08:08:08.123456789Z",
      "StartedAt": "2025-08-08T08:08:08.123456789Z",
      "Type": "NORMAL",
      "Networks": [
        {
          "NetworkMode": "awsvpc",
          "IPv4Addresses": [
            "10.0.1.23"
          ]
        }
      ]
    }
  ],
  "Limits": {
    "CPU": 1,
    "Memory": 2048
  },
  "PullStartedAt": "2025-08-08T08:08:01.123456789Z",
  "PullStoppedAt": "2025-08-08T08:08:06.123456789Z",
  "AvailabilityZone": "ap-northeast-2a"
}
```

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
jj22ee: 1. Just wondering if 12 character minimum is too short, and could allow strings not counted as containerIds?
2. Even in this "fall-back" scenario, IMO this should still be limited to `[a-fA-F0-9-]+` characters

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
```diff
@@ -163,10 +163,8 @@ export class AwsEcsDetector implements ResourceDetector {
       );
       const splitData = rawData.trim().split('\n');
       for (const str of splitData) {
-        if (str.length > AwsEcsDetector.CONTAINER_ID_LENGTH) {
-          containerId = str.substring(
-            str.length - AwsEcsDetector.CONTAINER_ID_LENGTH
-          );
+        containerId = this._extractContainerIdFromLine(str);
+        if (containerId) {
```
jj22ee: Looking at this again, this will try to match `Primary pattern -> Fallback -> Legacy Fallback` patterns, for each `str of splitData`. However, this will return after the first match (e.g. splitData[0] works with Legacy Fallback and would be return, instead splitData[1] that could have matched with Primary pattern).

I suppose what you really want is to change the logic to be something like:
```ts
for (const str of splitData) {
  // check Primary pattern, return if found
}
for (const str of splitData) {
  // check Fallback pattern, return if found
}
for (const str of splitData) {
  // check Legacy Fallback pattern, return if found
}
```

seongpil0948: out of review, thank you for leading me

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
seongpil0948: 1. It parse the entire cgroup file and evaluate every line in prioritized passes (Primary → Fallback → Legacy), so later primary matches aren’t shadowed by earlier lines. I didn’t identify additional edge cases beyond what the current tests cover.
I’ve raised the lower bound to `CONTAINER_ID_LENGTH_MIN = 32` (ECS taskId length), which reduces false positives while still covering Docker’s 64‑char IDs. In practice, valid IDs are found by `_extractPrimaryEcsContainerId` first, and, if needed, by the stricter `_extractLastSegmentContainerId`
Thanks for pointing this out.

2. thank you for found

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
```diff
@@ -161,21 +162,74 @@ export class AwsEcsDetector implements ResourceDetector {
...
+
+      // Pass 3: Legacy fallback to last 64 chars (Docker-style), keep existing behavior
+      for (const line of lines) {
+        const id = this._extractLegacyContainerId(line);
+        if (id) return id;
       }
     } catch (e) {
       diag.debug('AwsEcsDetector failed to read container ID', e);
     }
     return containerId;
```
jj22ee: `containerId` is now redundant. I guess here you can now return undefined and remove this variable, or change `const id` to just be `containerId`.

jj22ee (commented):
Ty for the fixes & enhancements! I have one remaining nit, and then I can approve.

packages/resource-detector-aws/src/detectors/AwsEcsDetector.ts
seongpil0948: You're right

jj22ee (approved):

jj22ee:
Need help from @open-telemetry/javascript-approvers to run the remaining PR workflows.

jj22ee:
@seongpil0948 There is a lint issue in the PR check. Should be able to be fixed by running:
```
cd packages/resource-detector-aws
npm run lint:fix
```

seongpil0948:
Thank you @pichlermarc !

PR #2859, opened 2025-05-27
refactor(instrumentation-express): update semconv usage to ATTR_ exports

trentm:
Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2377

### DIFF
--- a/detectors/node/opentelemetry-resource-detector-instana/README.md
+++ b/detectors/node/opentelemetry-resource-detector-instana/README.md
@@ -38,7 +38,7 @@ sdk.start()
 
 ## Available detectors
 
-This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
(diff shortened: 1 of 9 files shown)


plugins/node/opentelemetry-instrumentation-express/test/hooks.test.ts
```diff
@@ -179,7 +179,7 @@ describe('ExpressInstrumentation hooks', () => {
 
     it('should call requestHook when set in config', async () => {
       const requestHook = sinon.spy((span: Span, info: ExpressRequestInfo) => {
-        span.setAttribute(SEMATTRS_HTTP_METHOD, info.request.method);
+        span.setAttribute(ATTR_HTTP_REQUEST_METHOD, info.request.method);
```
trentm, line 182: Review note: Yes, this is changing from old HTTP semconv `http.method` to new stable HTTP semconv `http.request.method`. However, this is not testing that instr-express is generating this. It is just testing that the `requestHook` is being called.  We could satisfy the test by using `'my.http.method'` or `'foo'` or whatever.

pichlermarc (approved):

david-luna (approved):

PR #2865, opened 2025-05-29
fix(redis): use new tracer after setTracerProvider

blumamir:

This PR fixes a bug in "redis" instrumentation, where the `tracer` used to create spans is brought into closure scope once at instrumentation "create client" patch time and then used regardless of any new tracer provider being set via the `setTracerProvider` on the instrumentation instance.

```ts
  private _getPatchCreateClient() {
    const tracer = this.tracer;
    return function createClient(original: Function) {
      return getTracedCreateClient(tracer, original);
    };
  }
```

The function taking care of doing the patch and creating the span is now found in `utils.ts` which I think less makes sense logically and also requires passing `tracer`, `config` (and we should also use `diag` and `meter`) to a `util` function which is the cause for the current bug. This pattern is uncommon in the repo since having the function on instrumentation class guarantee simple access to providers instead of passing them as arguments. 

I am working on adding support for redis v5, and wanted to first consolidate `v4` into this package, and run into this bug will hooking up the multi-version support. 

I moved the function from `utils.ts` to a member function on the instrumentation class. the patched function now calls `instrumentation.tracer` which is [a getter function](https://github.com/open-telemetry/opentelemetry-js/blob/5736c498426cacaebf53ef89b2835e61be1cee98/experimental/packages/opentelemetry-instrumentation/src/instrumentation.ts#L165) that returns `_tracer` which is updated on the instrumentation [every time `setTracerProvider` is called](https://github.com/open-telemetry/opentelemetry-js/blob/5736c498426cacaebf53ef89b2835e61be1cee98/experimental/packages/opentelemetry-instrumentation/src/instrumentation.ts#L158).

This guarantee that `startSpan` is called on a tracer from the most up-to-date tracerProvider that was set via instrumentation public api function.

Added a new test to assert this change (failing before the refactor and passing after). The PR only moved the function from `utils.ts` and `instrumentation.ts` and changed the `tracer` being used. the function content itself is not modified


plugins/node/opentelemetry-instrumentation-redis/src/instrumentation.ts
```diff
@@ -96,28 +108,116 @@ export class RedisInstrumentation extends InstrumentationBase<RedisInstrumentati
...
+        const config = instrumentation.getConfig();
+
+        const hasNoParentSpan = trace.getSpan(context.active()) === undefined;
+        if (config.requireParentSpan === true && hasNoParentSpan) {
+          return original.apply(this, arguments);
+        }
+
+        const dbStatementSerializer =
+          config?.dbStatementSerializer || defaultDbStatementSerializer;
+        const span = instrumentation.tracer.startSpan(
```
blumamir, line 138: reviewer note: this is the fix. here we use `instrumentation.tracer` which gets updated in after `setTracerProvider` so we always use the latest one

plugins/node/opentelemetry-instrumentation-redis/src/utils.ts
```diff
@@ -49,17 +28,14 @@ const endSpan = (span: Span, err?: Error | null) => {
   span.end();
 };
 
-export const getTracedCreateClient = (tracer: Tracer, original: Function) => {
+export const getTracedCreateClient = (original: Function) => {
```
blumamir, line 31: reviewer note: removing unused parameter (which is also locked to it's initial value and does not update on `setTracerProvider`

david-luna (approved):

david-luna:
@blumamir,

thanks for the changes. I think we should merge this one before any work related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2867

Also, as a component owner, your feedback on the issue would be valuable :)

PR #2866, opened 2025-05-29
chore: improve test workflow and coverage reports

david-luna:

Changes the PR workflow to a single file to improve it making it faster with less resource consumption. The goal is to provide feedback to the PR author by compiling, testing and reporting coverage only for the affected packages in the PR (as for now it does a compilation, test, and coverage report for all in nodejs v18). More details on the plan can be found in #2870 

Closes: #2870 

- add new PR workflow with:
  - single compile step for affected packages. results are cached
  - unit and TAV test for affected packages for a matrix of nodejs versions. Reusing compilation results from cache. coverage reports are cached
  - get coverage reports from all nodejs versions (in cache) and merge them. then upload them to codecov
- add scripts to support such workflow
  - `*:ci:affected` script at root to execute tasks only for affected packages (leveraging `nx` CLI)
- update `codecov.yml` to assign [flags](https://docs.codecov.com/docs/flags) to each package so reports can be updated independently
- remove obsolete workflows for unit test and test-all-versions for PRs and push events

## Checklist

Codecov flags feature require a first manual upload to create them so it would be helpful to resume #2214 to at least have a script at the root to testa and upload all reports. There is also some PRs that affect the flag configuration so it would be preferable to have them merged 1st

- [x] merge #2886  or at least have a `test:with-services` script at the root
- [x] merge redis refactor https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2915
- [x] merge #2928
- [ ] manage codecov reports and flags


.github/workflows/pr-test.yml
```diff
@@ -185,54 +183,97 @@ jobs:
...
         with:
           fetch-depth: 0
       - uses: actions/setup-node@v4
         with:
-          node-version: ${{ matrix.node }}
+          node-version: 18
       - name: Install
         run: npm ci
-      - name: Download Build Artifacts
+      # TODO: add the rest of versions (18.19.0 & 20.6.0)? or are these enough??
```
pichlermarc: if easy to do we could also use these for completeness, but just using 18 and 20 is likely more than okay 🙂

.github/workflows/pr-test.yml
```diff
@@ -34,7 +35,7 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
-  unit-test:
+  unit-and-tav-test:
```
pichlermarc: keeping the name `unit-test` would help with repo settings, since unit-test (18) and unit-test (18.19.0) required checks. Repo managment is now automated from another repo, so changing the check's name would require us to open a PR in that other repo before merging this (which would then block other PRs from getting merged)

If we want to change the name we can do so in a smaller follow-up PR. :)

david-luna:
pino instrumentation test are failing with a missing package error although I've added it as peer dependency in `tav.yml`

```
-- required packages ["fast-safe-stringify@2.1.1","pino@5.14.0"]
  -- installing ["fast-safe-stringify@2.1.1","pino@5.14.0"]
  -- running test "npm run test" with pino (env: {})
  
  > @opentelemetry/instrumentation-pino@0.50.0 test
  > nyc --no-clean mocha 'test/**/*.test.ts'
  
  
   Exception during run: Error: Cannot find module 'fast-safe-stringify'
  Require stack:
  - /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-pino/node_modules/pino/lib/tools.js
  - /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-pino/node_modules/pino/lib/levels.js
 ```

the rest seems to work fine.

File: .github/workflows/pr-test.yml (hunk 2 above)
Comment (david-luna) [reply .github/workflows/pr-test.yml]: changed back to `unit-test` in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866/commits/444bee8072e0ca1cdf82c5cbd551fa22a86c8629

**Comment by david-luna** [comment]
For some reason `test-all-versions` is not installing the peer dependency and the test fails when loading because of the missing dependency.

```
  -- required packages ["fast-safe-stringify@2.1.1","pino@6.1.0"]
  -- installing ["fast-safe-stringify@2.1.1","pino@6.1.0"]
  -- running test "npm ls fast-safe-stringify" with pino (env: {})
  opentelemetry-contrib@0.1.0 /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib
  └── (empty)
```
workflow run link: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17122550403/job/48566983564?pr=2866

To add more weirdness it doesn't happen when we run TAV only with pino instrumentation (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2991). I think it might be related to [Nx parallelization](https://nx.dev/ci/concepts/parallelization-distribution#parallelization). Several processes installing dependencies at the same time may affect the dependencies in unexpected ways. 

I'll put the PR ready for review. I'll try to find a fix while we discuss the proposed changes.

File: .github/workflows/pr-test.yml [hunk 3]
```diff
@@ -1,14 +1,14 @@
-name: Unit Tests
+name: Pull Request Tests
 on:
-  push:
-    branches: [main]
   pull_request:
+    # TODO: remove it
+    branches: [main]
```
pichlermarc: this would now not run on `main` anymore, right? is this intentional?

.github/workflows/pr-test.yml
```diff
@@ -185,26 +186,27 @@ jobs:
...
-        if: ${{ !matrix.code-coverage }}
-        run: npm run test:ci:changed
-      - name: Report Coverage
-        if: ${{ matrix.code-coverage && !cancelled()}}
-        uses: codecov/codecov-action@v5
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        run: npm run test:ci:affected
+      - name: Test All Versions (Delta)
+        run: npm run test-all-versions:ci:affected
```
pichlermarc: does `affected` also trigger when `package-lock.json` is updated? Previosuly we've had this auto-labelling of PRs that allowed us to trigger TAV tests for all packges even if their files have not changed, is that still something that we can do with this? 🤔

.github/workflows/pr-test.yml
```diff
@@ -238,3 +240,56 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
+      - name: Install
+        run: npm ci
+      # NOTE: keep this in sync with the node versions from `unit-test` job
+      # TODO: doownload browser coverage
```
pichlermarc: looks like the browser tests don't follow this pattern - is there a significant difference in that would prevent this? 🤔

.github/workflows/pr-test.yml
```diff
@@ -238,3 +240,56 @@ jobs:
...
+      - name: Check coverage
+        run: find . | grep coverage-final
+
+      - name: Report Coverage
+        uses: codecov/codecov-action@v5
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        with:
+          verbose: true
+          dry_run: true
```
pichlermarc: is this still in `dry_run` on purpose? 🤔

scripts/codecov-upload-flags.mjs
```diff
@@ -0,0 +1,26 @@
+import path from 'path';
```
pichlermarc: hmm, could not find usage for this anywhere - is this still used? 🤔

pichlermarc: aah, I should've checked first what it does before asking. 😅
it's to generate the flags for the codecov yaml

david-luna:
Forced to use a single process in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866/commits/2694774feb07fdc447025e5c8a5ba0aa4e20e31d made the tests [pass](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17125111558/job/48575195856?pr=2866). The assumption seems to be correct. But this solution is not what we want because it increases significantly the time consumed for testing :(

.github/workflows/pr-test.yml
david-luna: I think I might got it wrong. The idea was to have this workflow running for all PRs so that's why there is a todo to remove the filter.

Now looking again at this section I think we may want to keep the same behaviour: run on push to main and any PR so we may want to revert to 

```yaml
name: Unit Tests
on:
  push:
    branches: [main]
  pull_request:
    # no filters here
```

.github/workflows/pr-test.yml
david-luna: Nx considers that a change outside the workspaces affects the entire repo so it runs the task for all of them.

.github/workflows/pr-test.yml
david-luna: For browser tests we just to a single run in node 22. That allows us to upload the coverage report right away without having to merge them with other runs in different node versions.

.github/workflows/pr-test.yml
david-luna: yes, I think I need to update a 1st report here before letting the CI do it

.github/workflows/pr-test.yml
pichlermarc: awesome :) 
that simplifies life significantly :D

david-luna:
> Forced to use a single process in [2694774](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/2694774feb07fdc447025e5c8a5ba0aa4e20e31d) made the tests [pass](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17125111558/job/48575195856?pr=2866). The assumption seems to be correct. But this solution is not what we want because it increases significantly the time consumed for testing :(

Comparing to other TAV runs in main branch (which test all packages) this workaround is still faster:
- TAV [run in main](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17127910504) took 1h 23m 40s
- TAV run [in this PR running sequentially](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17125111558) took 51m 33s. It failed in the coverage part but I don't think reporting coverage would consume the 30m diff

.github/workflows/pr-test.yml
david-luna: Sorry I made this change long ago and I forgot the reason of this change. Let me explain why I think it should work only on PRs and not on push.

In the current state when a push is made on main there are 2 workflows running:
- unit tests (the one changed here)
- TAV for push which runs `test-all-versions` for all packages

Now this PR adds TAV to the workflow so we will test for all versions twice. This is a waste of resources.

If we still want to run unit tests on push to main I thin we could modify `.github/workflows/test-all-versions.push.yml`

package.json
```diff
@@ -15,10 +15,14 @@
...
     "compile": "nx run-many -t compile",
+    "compile:ci:affected": "nx affected -t compile --base=origin/main --head=HEAD",
     "test": "nx run-many -t test",
+    "test:ci:affected": "nx affected -t test --base=origin/main --head=HEAD",
     "test:browser": "nx run-many -t test:browser",
-    "test:ci:changed": "nx affected -t test --base=origin/main --head=HEAD",
-    "test-all-versions": "nx run-many -t test-all-versions",
+    "test:browser:ci:affected": "nx affected -t test:browser --base=origin/main --head=HEAD",
+    "test-all-versions": "nx run-many --parallel=false -t test-all-versions",
+    "test-all-versions:ci:affected": "nx affected --parallel=false -t test-all-versions --base=origin/main --head=HEAD",
```
Ugzuzg: I have an idea for a tav patch that might make it work in parallel. As I understand, tav operates on the monorepo, so when it does it in parallel, everything breaks.
We could try installing the packages tav needs to install into a package that isn't under "workspaces", and then linking them into the `node_modules` of the specific instrumentation.

Ugzuzg: Trying it out here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2994
Looks like pino now passes, but some other stuff fails.

david-luna: I'm not opposed to the idea but this PR is quite big already. I'd prefer to focus on the feature (building & testing only what changed) and work on performance in a follow up PR. Running them sequentially it's still faster that what we have right now. See the [comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866#issuecomment-3211409185) above.

Ugzuzg: Not saying we should do it in one PR. And, at the current stage, the tav patch is extremely unready, so we should definitely do these things separately!

.github/workflows/pr-test.yml
```diff
@@ -185,26 +186,27 @@ jobs:
...
+        uses: actions/upload-artifact@v4
         with:
-          verbose: true
+          name: tests-coverage-cache-${{ github.run_number }}-${{ matrix.node }}
+          include-hidden-files: true
+          if-no-files-found: error
+          retention-days: 1
+          path: |
+            !node_modules
+            **/.nyc_output/**
```
Ugzuzg: There's [a bug in glob resolution of actions/toolkit](https://github.com/actions/toolkit/issues/657), which will walk the directory tree even if that path should already be excluded by the exclusion rule.

In the case of this PR, you can reduce the time to upload an artifact from ~15s to ~3s (not critical):
```suggestion
            packages/*/.nyc_output/**
            incubator/*/.nyc_output/**
```

In the modified tav version of this PR, where I used npm's isolated installs feature which produces deep directory structures, it took 10m to walk the directory tree! By changing the glob pattern to the one above, the step was executed in 3s instead.

Ugzuzg: Another thing about upload-artifact action is that it will find the deepest common directory for all paths, and create an archive starting from that common directory. Once [X-Ray Sampler moves out of incubating](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2993), that common directory will change from the root of the repo to `packages/`. This means that the coverage will be extracted to the wrong place.

Might be a good idea to include some file from the root to ensure `.nyc_output` is extracted into the right place.
```suggestion
            packages/*/.nyc_output/**
            incubator/*/.nyc_output/**
            LICENSE
```
or something similar.

.github/workflows/pr-test.yml
pichlermarc: Ah I completely missed that `test-all-versions.push.yaml` still exists. What I'm mostly worried about is that `test-all-versions.push.yaml` does not upload any codecov reports right now.

I might be misunderstanding something, though 😅

IIUC codecov would just carryforward from the PR's parent commit IFF nothing was reported in the latest test run. So without reporting on main the following will happen:
- PR-1 is opened, tests run and report the diff
- PR-1 is squashed and merged, so commit hash changes (the reported coverage from the PR does not apply to this commit anymore)
- PR-2 is opened, tests run and report the diff, but since we did not report on `main` for the squashed commit, we carryforward the coverage reports from before PR-1 was merged

So over time these would go out of sync and changes to coverage will look more extreme than they are, since without reporting on `main` since we're comparing to an old state on PRs. Also if files are added in one of the PRs, we won't have anything to carry forward, and that will be reported as no coverage on that file.

.github/workflows/pr-test.yml
david-luna: @Ugzuzg thanks for the tip. I'll apply your 2nd suggestion :)

david-luna: Looking at `package.json` I see incubator is not listed as part of the workspaces. I've did a quick search in the commit history but couldn't find `incubator` folder in workspaces. Apparently it wasn't even present when we use the old `lerna.json` conifg. You can check it at https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1771

So I'll keep incubator out for now and add it in a follow up PR. Will mention it in today's SIG

david-luna: > So I'll keep incubator out for now and add it in a follow up PR. Will mention it in today's SIG

It might not be needed since that package is about to be moved from incubator to packages
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2993

.github/workflows/test.yml
```diff
@@ -125,7 +140,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       cassandra:
-        image: bitnami/cassandra:3
+        image: cassandra:3
```
david-luna: today 2025-08-28 bitnamis is removing cassandra images from docker hub ([ref](https://hub.docker.com/r/bitnami/cassandra#%EF%B8%8F-important-notice-upcoming-changes-to-the-bitnami-catalog)). So I'm changing to use the official image which is already in `test/docker-compose.yml` file

packages/instrumentation-user-interaction/package.json
```diff
@@ -30,7 +30,8 @@
     "tracing",
     "profiling",
     "metrics",
-    "stats"
+    "stats",
+    "user-interaction"
```
david-luna, line 34: note to reviewer: change made to trigger `test:browser` in this package. This confirms the `affected` command works

.github/workflows/test.yml
david-luna: also got updated in main branch

.github/workflows/pr-test.yml
pichlermarc: this was addressed in the latest revision.

pichlermarc (approved):
Looks great! Looking forward to seeing it in action. 🙂

david-luna:
> Looks great! Looking forward to seeing it in action. 🙂

Thanks for the approval. However I found out that codecov action is not taking in consideration the flags defined in the `codecov.yml` file :_)

I could not find any option in the docs that allow this for multiple flags but just for one. People are using their own script to call upload each report with its corresponding flag. This [stack overflow conversation](https://stackoverflow.com/questions/66561281/optimal-usage-of-codecov-in-a-monorepo-context-with-separate-flags-for-each-pack) is old but I think is still valid.

I assumed that codecov action would read the `codecov.yml` file and apply the flags per package folder but it just upload all coverage files as one bug report. I think if we finally need to do one upload per package we could simplify the `codecov.yml` to use [automatic flag management](https://docs.codecov.com/docs/flags#recommended-automatic-flag-management)

david-luna:
@pichlermarc thanks for the approval.

If you don't mind to do another review. I had to change the codecov action but a custom script since there was no options in the action to update flags for each package and adding a codecov step for each package of the repo feels too much.

Until now I've set the `--dry-run` option so this PR was not messing with the previous reports. Once you have a second look and if all is fine I will proceed to do the actual upload. LAten I will probably need someone with access to codecov to enable flag analytics as described in https://docs.codecov.com/docs/flags#tracking-flag-coverage-over-time

scripts/codecov-upload-flags.mjs
```diff
@@ -0,0 +1,97 @@
...
+  process.exit(-1);
+}
+if (typeof codecovToken !== 'string') {
+  console.log('CODECOV_TOKEN env var missing! Exiting');
+  process.exit(-1);
+}
+
+const ROOT_DIR = process.cwd();
+const readPkg = (dir) => JSON.parse(readFileSync(path.join(dir, 'package.json'), 'utf8'));
+const execCmd = (cmd, opts = {}) => execSync(cmd, {encoding: 'utf-8', stdio: 'inherit', ...opts});
```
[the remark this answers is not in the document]
pichlermarc: I guess this is complaining since we do not use something like `execFileSync` (and arguments as an array here) - so if the ROOT_DIR is something with a space in it, it might try to execute the wrong thing. I don't see an angle for an exploit here, but it might still be worth aligning this (alerts usually recur from time-to-time and produce inbox clutter that makes it more difficult to see when there's an actual problem).

pichlermarc (changes requested):
Overall looks good % this comment here https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866#discussion_r2315833371

david-luna:
> Overall looks good % this comment here [#2866 (comment)](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866#discussion_r2315833371)

@pichlermarc thanks for the review. I fixed it and if you give me the green check I'll remove the `--dry-run` to actually upload the reports with flags. I may need your help to enable the flags in codecov UI.

pichlermarc (approved):
looks good, thank you for working on this 🙂

david-luna:
workflow run on main branch https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/17457685900

PR #2875, opened 2025-06-03
fix(instr-undici): fix user agent extraction and handle of multiple values on headers

david-luna:

Supersedes: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2781


plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -218,25 +254,14 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
-      const idx = request.headers.findIndex(
-        h => h.toLowerCase() === 'user-agent'
-      );
-      if (idx >= 0) {
-        userAgent = request.headers[idx + 1];
+    for (const { key, value } of this.parseRequestHeaders(request)) {
+      if (key.toLowerCase() === 'user-agent') {
+        // user-agent should only appear once per the spec, but the library doesn't
+        // prevent passing it multiple times, so we handle that to be safe.
+        const userAgent = Array.isArray(value) ? value[0] : value;
```
trentm: (non-blocking) FWIW, from the `curl` docs the *last* specified User-Agent wins:

>  If --user-agent is provided several times, the last set value is used.

So we could consider taking the *last* value if it is an array.  I am fine with the current code, though.

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -218,25 +254,14 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     }
 
     // Get user agent from headers
```
trentm, line 264: Note: Not for this PR, but https://opentelemetry.io/docs/specs/semconv/http/http-spans/ says that `user_agent.original` is Opt-In. I.e. collecting it should be off by default. We should do that in a separate (breaking) PR change.

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -330,26 +355,14 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
-        const value = hasSeparator
-          ? h.substring(sepIndex + 1)
-          : rawHeaders[idx + 1];
-
+      for (const { key, value } of this.parseRequestHeaders(request)) {
+        const name = key.toLowerCase();
         if (headersToAttribs.has(name)) {
-          spanAttributes[`http.request.header.${name}`] = value.trim();
+          spanAttributes[`http.request.header.${name}`] = value
+            .toString()
```
trentm: nit: To be extremely spec picky, https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3 says (emphasis mine):

> A recipient MAY combine multiple field lines within a field section that have the same field name into one field line, without changing the semantics of the message, by appending each subsequent field line value to the initial field line value in order, separated by a comma (",") and optional whitespace ([OWS](https://www.rfc-editor.org/rfc/rfc9110.html#whitespace), defined in [Section 5.6.3](https://www.rfc-editor.org/rfc/rfc9110.html#whitespace)). **For consistency, use comma SP.**

So I would say it is slightly preferred to have the joining of an array of values use `', '` rather than the `','` that is implicitly being used via `Array.prototype.toString()`.

plugins/node/instrumentation-undici/src/undici.ts
david-luna: There was a similar conversation in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2781#discussion_r2031017833 but we did not get into a conclusion. IMO is not a bad idea to follow what other well known tools like `curl` are doing. I'll change it

plugins/node/instrumentation-undici/src/undici.ts
david-luna: Let's go with the preferred separator

plugins/node/instrumentation-undici/src/undici.ts
david-luna: done in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2875/commits/9a7f8d72def50fc0048eca5c0cb165f5e335b748

plugins/node/instrumentation-undici/src/undici.ts
david-luna: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2875/commits/9a7f8d72def50fc0048eca5c0cb165f5e335b748

david-luna:
Last [test run](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/15427742221/job/43419275941?pr=2875) revealed the generator function wasn't the best suited for the job. In some scenarios (headers as string) the generator returned multiple values with the same `key` property since it was yielding the result for each line.

Instrumentation needs to process all lines to make sure it collects all header values for a given name. So I've changed the implementation to return a map instead in commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2875/commits/01c1c865fa2c29fe6e0b822a96a47cdd181954f8

david-luna:
Sorry @trentm but I prefer to have a second review since I've changed a bit the way headers are parsed. I'm still able to merge although I pushed a couple of commits after your approval. Is that supposed to work like this?

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -218,24 +262,13 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
-      );
-      userAgent =
-        uaHeader && uaHeader.substring(uaHeader.indexOf(':') + 1).trim();
-    }
+    const headersMap = this.parseRequestHeaders(request);
+    const userAgentValues = headersMap.get('user-agent');
 
-    if (userAgent) {
+    if (userAgentValues) {
+      const userAgent = Array.isArray(userAgentValues)
```
trentm, line 272: nit: Choosing the last user-agent value is an "interesting thing" to have a comment on, I think. Could point to the `curl` example as justification.

plugins/node/instrumentation-undici/src/undici.ts
```diff
@@ -329,27 +362,14 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
-        ).toLowerCase();
-        const value = hasSeparator
-          ? h.substring(sepIndex + 1)
-          : rawHeaders[idx + 1];
-
+      for (const [name, value] of headersMap.entries()) {
         if (headersToAttribs.has(name)) {
-          spanAttributes[`http.request.header.${name}`] = value.trim();
+          const attrValue = Array.isArray(value) ? value.join(', ') : value;
+          spanAttributes[`http.request.header.${name}`] = attrValue.trim();
```
trentm: nit: I don't think the `.trim()` is necessary now because `parseRequestHeaders` does that.

PR #2873, opened 2025-06-03
fix: export KoaLayerType enum correctly

cincauhangus:
## Short description of the changes

KoaLayerType is an enum. It was erroneously exported as a type causing errors on in Koa setups. This breaks the configuration setup for ignoring certain Koa Layers when publishing spans. 

This was working in @opentelemetry/instrumentation-koa v0.46.0, now broken in v0.50.0.

Typescript linter error: 

image

PR #2876, opened 2025-06-05
fix: incorrect export of enums

cincauhangus:
- Some enums are incorrectly exported as a type as a result of commit https://github.com/open-telemetry/opentelemetry-js-contrib/commit/60612f26582f470663512c0cb2aabae9d00d2651 (PR: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2839)

- Follow up from https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2873
- Fixed the exports


cincauhangus:
Hi @david-luna, FYI I created a separate PR as these are packages I don't use but found similar `export` issues through a code search.

Issue #2867, opened 2025-05-29
a plan for `redis` instrumentation


trentm:
Currently we have:
- `@opentelemetry/instrumentation-redis` which instruments redis@2 (last release 2017-08-08) and redis@3 (last release 2021-04-20), and
- `@opentelemetry/instrumentation-redis-4` which instruments redis@4 (active)
- nothing instruments redis@5 (first release 2025-04-30)

# Proposal

**tl;dr: We move to one `@opentelemetry/instrumentation-redis` that instruments redis@4 and redis@5. We drop support for instrumenting older versions of redis.**

Details for the legacy package:

- "plugins/node/opentelemetry-instrumentation-redis" is moved to "archive/instrumentation-redis-legacy/"
- We open an issue specifically about the instrumentation support for redis@2 and redis@3 no longer getting releases. This remains open for a period of time (3 months? 6 months?) where users can give feedback on whether they still use/need this instrumentation -- indicating if they are willing to be or find a maintainer/code-owner for the instrumentation.
- At the end of that period, OTel JS maintainers decide if there is a need to revive the instrumentation. If so, it returns as plugins/node/instrumentation-redis-legacy with the package name `@opentelemetry/instrumentation-redis-legacy`. If not, the "archive/..." code is replaced with just a README.md similar to [the archived "browser-extension-autoinjection"](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/archive/opentelemetry-browser-extension-autoinjection/README.md).

Details for the current `@opentelemetry/instrumentation-redis-4` package:

- "plugins/node/opentelemetry-instrumentation-redis-4" *stays* in the same dir (in anticipation of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223 being revived, and avoiding *two* directory moves).
- However, it changes to being published as `@opentelemetry/instrumentation-redis`.
- `@opentelemetry/auto-instrumentations-node` updates to this new release and removes `@opentelemetry/instrumentation-redis-4` -- effectively dropping support for redis@2 and redis@3.
- The `@opentelemetry/instrumentation-redis-4` package in npm is marked deprecated with a pointer to the new package and issue explaining the move.
- Separately, [support for redis@5](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2830) is added to the new `@opentelemetry/instrumentation-redis` package.
[...]

Issue #2830, opened 2025-05-18
Support `redis` v5


Vovcharaa:
### Is your feature request related to a problem? Please describe

Currently, the latest supported version of `redis` library is 4.
Version 5 was released a couple of weeks ago and has breaking changes that prevent `opentelemetry-instrumentation-redis-4` from applying tracing patches to the library.

### Describe the solution you'd like to see

Update `opentelemetry-instrumentation-redis-4` or create `opentelemetry-instrumentation-redis-5` instrumentation.

PR #2878, opened 2025-06-07
feat(redis): consolidate redis v2,3 and redis v4 instrumentation to one package

blumamir:
related: #2867 

This PR is a preparation for adding support to redis@5 in a followup PR #2830 

## Motivation

- merge `redis` and `redis-4` into one package which users can install without thinking about the redis version they are using.
- align with the rest of the repo which uses one npm package per instrumentation.
- limit the number of packages we need to handle so that we don't grow them when new versions are added.
- the name `redis-4` is a bit confusing since it's not obvious that 4 is the version. this pattern doesn't play well when a single package instrument multiple versions.

## Change

Create a wrapper `RedisInstrumentation` which installs both `v2-v3` and `v4` existing instrumentations. In a followup PR it will be trivial to also add support to `v5` which again had major internal refactor not trivial to support as part of `v4`.

- allows to conveniently manage each target version range as different `InstrumentationBase` as we do now (with potentially specific patches, types, utils, tests.
- This PR only takes care of merging existing implementation into the single `@opentelemetry/instrumentaiton-redis` without any other improvements or refactors. instrumentation logic is remain as is.
- I have tested it when both `redis` (new) and `redis-4` (deprecated) instrumentations are registered together, and the spans are only emitted once. This is due to instrumentation checking [`isWrapped`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/11a94b09111787489643365690591d1e607dd119/plugins/node/opentelemetry-instrumentation-redis-4/src/instrumentation.ts#L99) on the patched function and unwrapping before applying a new patch. While this works, it means that only the last instrumentation to patch will apply. I suggest that after a migration period we will re-publish `redis-4` with empty implementation to signal users to stop using it and migrate to the maintained package.

### DIFF
--- a/plugins/node/opentelemetry-instrumentation-redis-4/README.md
+++ b/plugins/node/opentelemetry-instrumentation-redis-4/README.md
@@ -1,5 +1,7 @@
 # OpenTelemetry redis Instrumentation for Node.js
 
+> ⚠️ **DEPRECATED**: The support for `redis@4` instrumentation is now part of `@opentelemetry/instrumentation-redis`. please use it instead.
+
 [![NPM Published Version][npm-img]][npm-url]
 [![Apache License][license-image]][license-image]
 


plugins/node/opentelemetry-instrumentation-redis/package.json
```diff
@@ -59,6 +63,7 @@
     "cross-env": "7.0.3",
     "nyc": "17.1.0",
     "redis": "3.1.2",
+    "redis-v4": "npm:redis@4.7.1",
```
blumamir, line 66: reviewer note: this is needed for the `v4` tests. 
Since `redis` version is `3` by default, then `npm run compile` would fail when trying to import `RedisClientType` if not pulled into `node_modules` somehow.

When running the tests, `tav redis 4.7.1` would take care of updating the redis to make the test pass, but `compile` is run before that

plugins/node/opentelemetry-instrumentation-redis/package.json
```diff
@@ -1,12 +1,16 @@
...
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
   "repository": "open-telemetry/opentelemetry-js-contrib",
   "scripts": {
-    "test": "nyc mocha 'test/**/*.test.ts'",
+    "test": "npm run test-v2-v3 && npm run test-v4",
+    "test-v2-v3": "tav redis 3.1.2 npm run test-v2-v3-run",
+    "test-v4": "tav redis 4.7.1 npm run test-v4-run",
+    "test-v2-v3-run": "nyc mocha --no-clean --require '@opentelemetry/contrib-test-utils' 'test/v2-v3/*.test.ts'",
+    "test-v4-run": "nyc mocha --no-clean --require '@opentelemetry/contrib-test-utils' 'test/v4/*.test.ts'",
```
blumamir, line 13: reviewer note: 

this is aligned with [`@opentelemetry/instrumentation-mongodb`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/11a94b09111787489643365690591d1e607dd119/plugins/node/opentelemetry-instrumentation-mongodb/package.json#L10)

plugins/node/opentelemetry-instrumentation-redis/test/v2-v3/redis.test.ts
```diff
@@ -40,14 +39,12 @@ import {
   SEMATTRS_NET_PEER_PORT,
 } from '@opentelemetry/semantic-conventions';
 
-const instrumentation = new RedisInstrumentation();
-instrumentation.enable();
-instrumentation.disable();
+const instrumentation = registerInstrumentationTesting(
+  new RedisInstrumentation()
+);
```
blumamir, line 44: reviewer note:

Now that we have more than one ".test.ts" file, without this we get multiple errors in tests because the version that create the "patch" (first one) might be different than the one that the test is using. Migrating to the `contrib-test-utils` framework avoid these issues by registering only a single instrumentation instance and making sure it is used consistently

blumamir: It also remove a lot of boiler plate code and makes the tests file lighter

plugins/node/opentelemetry-instrumentation-redis/test/v2-v3/redis.test.ts
```diff
@@ -239,40 +218,18 @@ describe('redis@2.x', () => {
       });
     });
 
-    describe('Removing instrumentation', () => {
```
blumamir, line 242: reviewer note:

This was tricky to support in test after the refactor, and since "disable" is not a feature that we recommend users, and cannot work in some cases, I removed it from here

plugins/node/opentelemetry-instrumentation-redis/src/types.ts
```diff
@@ -17,16 +17,6 @@
...
 
-// exported from
-// https://github.com/redis/node-redis/blob/v3.1.2/lib/command.js
-export interface RedisCommand {
-  command: string;
-  args: string[];
-  buffer_args: boolean;
-  callback: (err: Error | null, reply: unknown) => void;
-  call_on_write: boolean;
-}
```
blumamir, line 28: reviewer note:

this is technically a breaking change (since it was publicly exported by the package and now it is not). but it's only used internally and there is no reason for any user to consume it for anything.

plugins/node/opentelemetry-instrumentation-redis-4/README.md
```diff
@@ -1,5 +1,7 @@
 # OpenTelemetry redis Instrumentation for Node.js
 
+> ⚠️ **DEPRECATED**: The support for `redis@4` instrumentation is now part of `@opentelemetry/instrumentation-redis`. please use it instead.
```
trentm, line 3:
```suggestion
> ⚠️ **DEPRECATED**: The support for `redis@4` instrumentation is now part of `@opentelemetry/instrumentation-redis` (as of v0.50.0). Please use it instead.
```

plugins/node/opentelemetry-instrumentation-redis/src/types.ts
trentm: How about we put the `!` marker on the PR title, then, and add a section to the PR description that mentions the remote-but-technically-possible breaking change?

plugins/node/opentelemetry-instrumentation-redis/src/v4/instrumentation.ts
```diff
@@ -0,0 +1,490 @@
...
+export class RedisInstrumentationV4 extends InstrumentationBase<RedisInstrumentationConfig> {
+  static readonly COMPONENT = 'redis';
+
+  constructor(config: RedisInstrumentationConfig = {}) {
+    super(PACKAGE_NAME, PACKAGE_VERSION, { ...DEFAULT_CONFIG, ...config });
+  }
+
+  override setConfig(config: RedisInstrumentationConfig = {}) {
+    super.setConfig({ ...DEFAULT_CONFIG, ...config });
+  }
```
trentm, line 64: I believe you can drop the `DEFAULT_CONFIG` handling in this file, as you already did for src/v2-v3/instrumentation.ts:

```diff
diff --git a/plugins/node/opentelemetry-instrumentation-redis/src/v4/instrumentation.ts b/plugins/node/opentelemetry-instrumentation-redis/src/v4/instrumentation.ts
index 5ef319df..38f444ca 100644
--- a/plugins/node/opentelemetry-instrumentation-redis/src/v4/instrumentation.ts
+++ b/plugins/node/opentelemetry-instrumentation-redis/src/v4/instrumentation.ts
@@ -48,19 +48,11 @@ interface MutliCommandInfo {
   commandArgs: Array<string | Buffer>;
 }

-const DEFAULT_CONFIG: RedisInstrumentationConfig = {
-  requireParentSpan: false,
-};
-
 export class RedisInstrumentationV4 extends InstrumentationBase<RedisInstrumentationConfig> {
   static readonly COMPONENT = 'redis';

   constructor(config: RedisInstrumentationConfig = {}) {
-    super(PACKAGE_NAME, PACKAGE_VERSION, { ...DEFAULT_CONFIG, ...config });
-  }
-
-  override setConfig(config: RedisInstrumentationConfig = {}) {
-    super.setConfig({ ...DEFAULT_CONFIG, ...config });
+    super(PACKAGE_NAME, PACKAGE_VERSION, config);
   }

   protected init() {
```

plugins/node/opentelemetry-instrumentation-redis/package.json
trentm: IMO we should be leaving testing of multiple versions of the target library to `test-all-versions` tests -- that's what `test-all-versions` is for. I think the mongodb testing change was misguided.

I have an alternative PR that builds on yours that shows what I mean: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2915
That PR also has a few other changes. PTAL.

blumamir:
replaced with #2915

Issue #2879, opened 2025-06-10
http.response_content_length does not get populated when ignoreNetworkEvents is set to true


howardyoo:
### What version of OpenTelemetry are you using?
Latest (0.19.x)

### What version of Node are you using?
Latest

### What did you do?
When using opentelemetry js on web, setting the `ignoreNetworkEvents` to true does not seem to populate the http.content_response_length attribute, even though that is not the part of the network events (span events).

### What did you expect to see?
Regardless of whether ignoreNetworkEvents are set to true or false, the response content length should be returned to the resource fetch span, because it is not a part of network span events.

### What did you see instead?
looked like http.content_repsonse_length would not get populated when ignoreNetworkEvents is set to true.

### Additional context
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts#L121
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts#L209

The code is checking whether `!this.getConfig().ignoreNetworkEvents` returns false, and then move to run the addSpanNetworkEvents. This is how the function is written:
```
function addSpanNetworkEvents(span, resource, ignoreNetworkEvents = false, ignoreZeros, skipOldSemconvContentLengthAttrs) {
      if (ignoreZeros === undefined) {
        ignoreZeros = resource[PerformanceTimingNames.START_TIME] !== 0;
      }
      if (!ignoreNetworkEvents) {
        addSpanNetworkEvent(span, PerformanceTimingNames.FETCH_START, resource, ignoreZeros);
        addSpanNetworkEvent(span, PerformanceTimingNames.DOMAIN_LOOKUP_START, resource, ignoreZeros);
        addSpanNetworkEvent(span, PerformanceTimingNames.DOMAIN_LOOKUP_END, resource, ignoreZeros);
        addSpanNetworkEvent(span, PerformanceTimingNames.CONNECT_START, resource, ignoreZeros);
        addSpanNetworkEvent(span, PerformanceTimingNames.SECURE_CONNECTION_START, resource, ignoreZeros);
        addSpanNetworkEvent(span, PerformanceTimingNames.CONNECT_END, resource, ignoreZeros);
        addSpanNetworkEvent(span, PerformanceTimingNames.REQUEST_START, resource, ignoreZeros);
[...]

PR #2880, opened 2025-06-10
fix: http.response_content_length does not get populated when ignoreNetworkEvents is set to true

howardyoo:
fix: http.response_content_length does not get populated when ignoreNetworkEvents is set to true

This PR fixes the issue #2879 where when the instrumentation is configured with `ignoreNetworkEvents` set as true, the two attributes: http.response_content_length and http.response_content_length_compressed are also getting omitted, due to a logic flaw that did not call the `addSpanNetworkEvents` when the ignoreNetworkEvents were set to true.

Regardless of whether ignoreNetworkEvents are set to true or false, the response content length should be returned to the resource fetch span, as it is not a part of network span events.

Fixes: #2879

### DIFF
--- a/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
+++ b/plugins/web/opentelemetry-instrumentation-document-load/src/instrumentation.ts
@@ -117,9 +117,7 @@ export class DocumentLoadInstrumentation extends InstrumentationBase<DocumentLoa
         if (fetchSpan) {
           fetchSpan.setAttribute(SEMATTRS_HTTP_URL, location.href);
           context.with(trace.setSpan(context.active(), fetchSpan), () => {
-            if (!this.getConfig().ignoreNetworkEvents) {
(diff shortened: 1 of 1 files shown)


dyladan:
Please add tests

JamieDanielson (commented):
Thanks for fixing this!
As Dan mentioned, just needs a test update. Also please run `npm run lint` to see the change needed to fix the failing status check. In fact you should be able to do `npm run lint:fix` and it will fix it for you.

howardyoo:
> Thanks for fixing this! As Dan mentioned, just needs a test update. Also please run `npm run lint` to see the change needed to fix the failing status check. In fact you should be able to do `npm run lint:fix` and it will fix it for you.

Thanks, yes, ran lint:fix and fixed the formatting issue.

howardyoo:
> Please add tests

Sure, added the test for this. Thanks.

JamieDanielson (approved):

PR #2886, opened 2025-06-16
Add docker compose and test serivces scripts

david-luna:

- TBD

- add docker compose file
- add script to spin up services for testing
- move MySQL log config command from workflow to tests


.github/workflows/test-all-versions.yml
```diff
@@ -163,8 +163,6 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node }}
-      - name: Set MySQL variables
```
david-luna, line 166: note for reviewer: this is required only for testing `@opentelemetry/instrumentation-mysql` and can be executed as a query using the client. Hence is removed from there and added in `plugins/node/opentelemetry-instrumentation-mysql/test/mysql.test.ts`

.github/workflows/unit-test.yml
```diff
@@ -176,8 +176,6 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node }}
-      - name: Set MySQL variables
```
david-luna, line 179: note for reviewer: same here

plugins/node/opentelemetry-instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -64,28 +64,38 @@ interface Result extends RowDataPacket {
...
-    if (testMysqlLocally) {
-      this.timeout(5000);
-      testUtils.cleanUpDocker('mysql');
+describe('mysql2', () => {
+  // assumes local mysql db is already available in CI or
+  // using `npm run test-services:start` script at the root folder
+  const shouldTest = process.env.RUN_MYSQL_TESTS;
+
+  before(async function () {
+    const connection = createConnection({
```
david-luna, line 83: note for reviewer: enabling here the `general_log` setting instead of doing it in the workflows

plugins/node/opentelemetry-instrumentation-pg/test/pg.test.ts
```diff
@@ -64,7 +64,7 @@ const memoryExporter = new InMemorySpanExporter();
 const CONFIG = {
   user: process.env.POSTGRES_USER || 'postgres',
   password: process.env.POSTGRES_PASSWORD || 'postgres',
-  database: process.env.POSTGRES_DB || 'postgres',
+  database: process.env.POSTGRES_DB || 'otel_pg_database',
```
david-luna, line 67: note for reviewer: align the value with the one in workflow services

package.json
```diff
@@ -19,6 +19,9 @@
     "test:browser": "nx run-many -t test:browser",
     "test:ci:changed": "nx affected -t test --base=origin/main --head=HEAD",
     "test-all-versions": "nx run-many -t test-all-versions",
+    "test-services:start": "docker compose -f ./test/docker-compose.yaml up -d --wait",
+    "test-services:stop": "docker compose -f ./test/docker-compose.yaml down",
+    "test:with-services-config": "NODE_OPTIONS='-r dotenv/config' DOTENV_CONFIG_PATH=./test/test-services.env npm test",
```
raphael-theriault-swi: This should probably use `cross-env` or something similar to not bail on Windows.

raphael-theriault-swi (commented):
Love this, small nit tho.

CONTRIBUTING.md
```diff
@@ -135,6 +135,34 @@ The conventional commit type (in PR title) is very important to automatically bu
...
+```sh
+npm run test-services:start       # starts services in Docker
+npm run test:with-services-config # runs 'npm test' with envvars from test/test-services.env
+npm run test-services:stop        # stops services in Docker
+```
+
+If you only want to test a sigle package (e.g. the `instrumentation-mongodb`) you can `cd` into it and run the tests after you started the services.
+
+```sh
+npm run test-services:start                           # starts services in Docker
```
trentm, line 157: This could optionally be `npm run test-services:start  mongodb` but probably don't need to bother mentioning this. Easier to keep it simple.

plugins/node/opentelemetry-instrumentation-mysql/test/index.test.ts
```diff
@@ -54,9 +53,9 @@ describe('mysql@2.x-Tracing', () => {
   let connection: mysqlTypes.Connection;
   let pool: mysqlTypes.Pool;
   let poolCluster: mysqlTypes.PoolCluster;
-  const testMysql = process.env.RUN_MYSQL_TESTS; // For CI: assumes local mysql db is already available
-  const testMysqlLocally = process.env.RUN_MYSQL_TESTS_LOCAL; // For local: spins up local mysql db via docker
-  const shouldTest = testMysql || testMysqlLocally; // Skips these tests if false (default)
+  // assumes local mysql db is already available in CI or
+  // using `npm run test-services:start` script at the root folder
+  const shouldTest = process.env.RUN_MYSQL_TESTS; // For CI: assumes local mysql db is already available
```
trentm:
```suggestion
  const shouldTest = process.env.RUN_MYSQL_TESTS;
```

Don't need to state the assumption twice.

test/docker-compose.yaml
```diff
@@ -0,0 +1,119 @@
...
+      APP_USER_PASSWORD: "secret"
+      ORACLE_PASSWORD: "oracle"
+    ports:
+      - 1521:1521
+    healthcheck:
+      test: ["CMD", "sqlplus" ,"system/oracle@//localhost/FREEPDB1"]
+      interval: 10s
+      timeout: 5s
+      retries: 30
+  postgres:
```
trentm, line 89:
```suggestion

  postgres:
```

test/docker-compose.yaml
```diff
@@ -0,0 +1,119 @@
...
+      test: ["CMD-SHELL", "[ $$(nodetool statusgossip) = running ]"]
+      interval: 1s
+      timeout: 10s
+      retries: 30
+
+  memcached:
+    image: memcached:1.6.37-alpine
+    ports:
+    - 11211:11211
+  mongodb:
```
trentm, line 32:
```suggestion

  mongodb:
```

test/docker-compose.yaml
```diff
@@ -0,0 +1,119 @@
...
+# `@opentelemetry/instrumentation-*` locally.
+#
+# Note: This isn't used in CI. CI uses GitHub Actions' `services: ...` for
+# defining test services.
+#
+# Usage:
+#   npm run test-services:start [services...]
+#   npm run test-services:stop [services...]
+
+name: opentelemetry-nodejs-test-services
```
trentm:
```suggestion
name: opentelemetry-js-contrib-test-services
```

test/test-services.env
trentm, line 1: nit: One thing I'd started doing at https://github.com/trentm/opentelemetry-js-contrib/blob/tm-test-services/test/test-services.env was to group these envvars by service/instrumentation. Might be nice to do that sometime. This is totally a "nit", though, because it is just to self-document a little bit.

CONTRIBUTING.md
```diff
@@ -135,6 +135,34 @@ The conventional commit type (in PR title) is very important to automatically bu
...
+
+```sh
+npm test
+```
+
+However, some instrumentations require test-services to be running (e.g. the `instrumentation-mongodb` package requires a MongoDB server). Use the `test-services`-related npm scripts to start all required services in Docker and then run the tests with the appropriate configuration to use those services:
+
+```sh
+npm run test-services:start       # starts services in Docker
+npm run test:with-services-config # runs 'npm test' with envvars from test/test-services.env
```
trentm, line 150: I like the `:with-services-config` (or perhaps `:with-services-env` or the loooong `:with-test-services-env`) as a better alternative to my earlier `:with-test-services`.

test/test-services.env
david-luna: I like the idea. Will do in a next commit

CONTRIBUTING.md
david-luna: I'd prefer to not repeat `test`. I'll change it to `test:with-services-env`

david-luna: commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2886/commits/f832c977421fb8f826f87013126af2d128ac203c

test/test-services.env
david-luna: commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2886/commits/f832c977421fb8f826f87013126af2d128ac203c

CONTRIBUTING.md
david-luna: I'm planning to do a follow up PR to add scripts for each package which will use this feature. I'll remember to update this description.

CONTRIBUTING.md
```diff
@@ -135,6 +135,34 @@ The conventional commit type (in PR title) is very important to automatically bu
...
+npm run test-services:stop        # stops services in Docker
+```
+
+If you only want to test a sigle package (e.g. the `instrumentation-mongodb`) you can `cd` into it and run the tests after you started the services.
```
pichlermarc:
```suggestion
If you only want to test a single package (e.g. the `instrumentation-mongodb`) you can `cd` into it and run the tests after you started the services.
```

pichlermarc (approved):
thank you for working on this! 🙂

Issue #2891, opened 2025-06-25
coordinate linting of JS files in examples/* and .mjs files in the repo


trentm:
Currently the linting in this repo works like this:

- `npm run lint` at the top-level runs `npm run lint` in each of the packages in the workspace
- these all run eslint using a shared "eslint.config.js" from the top-level (which includes the "prettier" plugin for style, including using the "prettier.config.js" at the top-level)

This issue is about:

1. Including JS  and TS files in `examples/...`, linted via `npm run lint` and using the same shared config files. (Currently there is a `npm run lint:examples`, but it (a) is **not run in CI** and (b) uses **a different eslint config** ("examples/.eslintrc").
2. Linting *JS* files (in addition to the TS files that are already linted) elsewhere in the repo. There aren't many: mostly these are `.../test/fixtures/*.mjs` files for testing ESM instrumentation.

The plan is to:

- [x] Update the eslint.config.js to support linting .js and .mjs files. Currently it just works for TS files.
  DONE in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2567
- [x] Work through each of the `examples/*` packages adding `npm run lint` and `npm run lint:fix` using the top-level shared eslint config file; and getting that lint passing.
    - [x] examples/bunyan (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2567) 
    - [x] examples/connect (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2897)
    - [x] examples/dns (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2898)
    - [x] examples/express (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2894)
    - [x] examples/fastify (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2896)
    - [x] examples/graphql (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2899)
    - [x] examples/grpc-census-prop (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2900)
    - [x] examples/hapi (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2892)
    - [x] examples/host-metrics (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2901)
    - [x] examples/ioredis (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2902)
[...]

PR #2903, opened 2025-06-26
chore(examples): lint examples/koa using shared top-level eslint config

trentm:
Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2891


examples/koa/src/client.ts
```diff
@@ -1,29 +1,55 @@
...
+ * limitations under the License.
+ */
 
-import { setupTracing } from "./tracer";
-const tracer = setupTracing('example-koa-client');
 import * as api from '@opentelemetry/api';
-import { default as axios } from 'axios';
+import * as axios from 'axios';
+// eslint-disable-next-line import/extensions
+import { setupTracing } from './tracer';
```
trentm, line 20: FYI: This need to `eslint-disable-next-line import/extensions` will be dropped once *all* examples/ packages are using the top-level eslint.config.js and we can delete examples/.eslintrc. It is the `"extends": "airbnb-base",` in "examples/.eslintrc" that is making this necessary for now.

(The alternative is to delete "examples/.eslintrc" and possibly break `npm run lint:examples` in the top-level until I complete all the items in https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2891)

maryliag: do you have an issue to track dropping the comments (on this and any other PR that you had to add it) or can you update the issue 2891 to include that as a step?

maryliag (approved):

PR #2912, opened 2025-06-26
chore(examples): lint examples/web using shared top-level eslint config

trentm:
Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2891


examples/web/examples/meta/index.js
```diff
@@ -54,6 +73,7 @@ function btnAddClick() {
 }
 
 function prepareClickEvents() {
+  // eslint-disable-next-line no-plusplus
```
maryliag, line 76: this looks like such a silly rule, it's so common to use ++. Do we really need it? Can we have the config updated to allow it?

maryliag (approved):

examples/web/examples/meta/index.js
trentm: @maryliag Agreed. My *guess* is that this is effectively coming from ["examples/.eslintrc"](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/examples/.eslintrc), which still gets picked up (and I guess *merged* with the top-level eslint.config.js?). This is a similar issue to an `eslint-disable-next-line` comment here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2903#discussion_r2169676392

My intention is to: (a) merge all these, and then (b) delete "examples/.eslintrc" and revisit the dozen or so `eslint-disable-next-line ...` directives I had to add to get `npm run lint` to pass.

maryliag: Sounds good!

Issue #2757, opened 2025-03-14
tests: possible flaky test in mongodb instrumentation package


david-luna:
There's been some failures recently in Unit tests workflow. These failures happen in [some PRs](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/13649886134/job/38647847585?pr=2685)
 and also in [main branch](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/13831362498/job/38696586871?pr=2726)

The issue seems to happen for mongodb v3 tests and in node v20 giving the following output.
```
> @opentelemetry/instrumentation-mongodb@0.52.0 test-v3-run
> nyc --no-clean mocha --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v3.test.ts'



  MongoDBInstrumentation-Tracing-v3
(node:10022) [MONGODB DRIVER] Warning: Current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor.
(Use `node --trace-warnings ...` to show where the warning was created)
    Instrumenting query operations
      ✔ should create a child span for insert
      ✔ should create a child span for update
      ✔ should create a child span for remove
    Instrumenting cursor operations
      ✔ should create a child span for find
      ✔ should create a child span for cursor operations
    Instrumenting command operations
      ✔ should create a child span for create index
      ✔ should create a child span for aggregation
    when using enhanced database reporting without db statementSerializer
      ✔ should properly collect db statement (hide attribute values)
      ✔ should properly collect nested db statement (hide attribute values)
    when specifying a dbStatementSerializer configuration
      with a valid function
        ✔ should properly collect db statement
      with an invalid function
        ✔ should not do any harm when throwing an exception
    when specifying a responseHook configuration
      with a valid function
        ✔ should attach response hook data to the resulting span for insert function
        ✔ should attach response hook data to the resulting span for find function
      with an invalid function
        ✔ should not do any harm when throwing an exception
    Mixed operations with callback
      ✔ should create a span for find after callback insert
    MongoDb useUnifiedTopology enabled
      ✔ should generate correct span attributes
    requireParentSpan
      1) should not create spans without parent span when requireParentSpan is explicitly set to true
      ✔ should create spans without parent span when requireParentSpan is false
    Removing Instrumentation
      ✔ should unpatch plugin
[...]

PR #2915, opened 2025-06-26
feat(instrumentation-redis)!: consolidate redis v2,3 and redis v4 instrumentation to one package

trentm:
This PR builds on Amir's https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2878 (the instr-redis consolidation).

# summary of diffs to Amir's

- I've changed the testing so that `npm test` just tests against the version of redis currently installed and `npm run test-all-versions` handles testing against older versions of redis. (See my argument for this in a section below. commit 8a52338b1859a49b8e284ffe2bf7feebb025a01d)
- I've renamed src/instrumentation.ts to src/redis.ts just to have a name change. This allows `git` to notice that src/instrumentation.ts was *moved* to src/v2-v3/instrumentation.ts. This makes for **a smaller, clearer diff** and **clearer history tracking with `git log --follow ...` later. (commit 9b0fc9e966f17fbdcd43eb43ba73b24942badc00)
- I've **removed** all the sources in the old instrumentation-redis-4 for the same reason: git notices the **move** of src/... and test/... files to the new location. (commit 79cd098219ba9471d0199e7943e5c99c31d71215)
- Drop `DEFAULT_CONFIG` handling in src/v4/instrumentation.ts, which I believe is no longer necessary (commit f46e1664c26b5ab9bd67bff29ac999d0438e064f). Amir's PR had already dropped the `DEFAULT_CONFIG` handling for src/v2-v3/instrumentation.ts.
- Remove `instrumentation-redis-4` from auto-instrumentations-node and release-please: necessary given that I removed the package.  (commit d2fa3a2828b30b58b749128a3dda2196d7628c83) I'm not sure if this is a "breaking" change for auto-instrumentations-node. It certainly is something we should point out in the changelog / PR description.

# why the test changes

I moved all testing against redis v2 and v3 to `npm run test-all-versions`.  Testing against older versions of the target module is what `test-all-versions` is for.

This, then, makes `npm test` only test against the currently installed version of the module being instrumented (redis v4).  IMO this is what `npm test` should be doing in all our packages.  `npm test` should *not* change the versions of things installed, which is what happens in Amir's PR (note the "invalid" version in `npm ls` after running `npm test`):

```
% npm test
...

% npm ls redis
opentelemetry-contrib@0.1.0 /Users/trentm/tm/opentelemetry-js-contrib7
└─┬ @opentelemetry/instrumentation-redis@0.49.0 -> ./plugins/node/opentelemetry-instrumentation-redis
  └── redis@4.7.1 invalid: "3.1.2" from plugins/node/opentelemetry-instrumentation-redis
```

One current exception (the only one?) to this is the instrumentation-mongodb tests. Testing of multiple versions of mongodb was added to `npm test` in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2681 **to help with coverage numbers**. A better answer is to get coverage data from `test-all-version` runs. @david-luna is working on this.  (Note that the testing of mongodb@3 was **dropped** from `npm test` as a workaround for flaky tests a while back: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2757#issuecomment-2733499340)

With my changes in this PR, to do local dev testing against redis@3, one would need to:

```
npm run test:docker:run
npm install --no-save redis@3 # or whatever specific version
RUN_REDIS_TESTS=1 npm run test-v2-v3
```

I think this is reasonable. There will be vanishly little dev on the legacy *v2-v3* instrumentation.

# follow-ups

If we go with this PR and once it is released:

- [ ] `npm deprecate ...` for the old package
- [x] ~slurp in the utils from the `redis-common` package and deprecate / remove it.~  Amir correctly pointed out that instrumentation-ioredis also uses `redis-common`, so there is no driving need to drop it.
- [ ] drop the old issues label (https://github.com/open-telemetry/opentelemetry-js-contrib/labels?q=redis)


trentm:
Moved back to draft, because I need to cope with *redisTypes* usage in test/v2-v3/redis.test.ts now that the currently installed redis is no longer redis@3, so the types cause a `npm run compile` failure.

trentm:
Grrr, why can I not do this?  (This is a rhetorical question.)

```
% npm run compile

> @opentelemetry/instrumentation-redis@0.49.1 compile
> tsc -p .

src/v2-v3/utils.ts:17:34 - error TS6137: Cannot import type declaration files. Consider importing 'redis' instead of '@types/redis'.

17 import type * as redisTypes from '@types/redis';
                                    ~~~~~~~~~~~~~~

test/v2-v3/redis.test.ts:46:44 - error TS6137: Cannot import type declaration files. Consider importing 'redis' instead of '@types/redis'.

46 import type { RedisClient, Callback } from '@types/redis';
                                              ~~~~~~~~~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  src/v2-v3/utils.ts:17
     1  test/v2-v3/redis.test.ts:46
```

plugins/node/opentelemetry-instrumentation-redis/src/v2-v3/utils.ts
```diff
@@ -29,14 +28,14 @@ export const endSpan = (span: Span, err?: Error | null) => {
 };
 
 export const getTracedCreateClient = (original: Function) => {
-  return function createClientTrace(this: redisTypes.RedisClient) {
-    const client: redisTypes.RedisClient = original.apply(this, arguments);
+  return function createClientTrace(this: any) {
```
david-luna, line 31: is not strictly necessary to know the client type here just to bind it to the context. IMO this change is fine :)

plugins/node/opentelemetry-instrumentation-redis/src/v2-v3/instrumentation.ts
```diff
@@ -25,10 +25,10 @@ import {
   getTracedCreateClient,
   getTracedCreateStreamTrace,
 } from './utils';
-import { RedisCommand, RedisInstrumentationConfig } from './types';
+import { RedisInstrumentationConfig } from '../types';
 /** @knipignore */
-import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
-import { RedisPluginClientTypes } from './internal-types';
+import { PACKAGE_NAME, PACKAGE_VERSION } from '../version';
+import { RedisCommand, RedisPluginClientTypes } from './internal-types';
```
david-luna: nit: we can leverage this change to use `import type` which is more accurate and also it may benefit if we want to be bundler friendly. Same on the line importing `RedisInstrumentationConfig`

plugins/node/opentelemetry-instrumentation-redis/test/v2-v3/redis.test.ts
```diff
@@ -21,16 +21,15 @@ import {
...
+  ROOT_CONTEXT,
 } from '@opentelemetry/api';
-import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
-import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
 import * as testUtils from '@opentelemetry/contrib-test-utils';
 import {
-  InMemorySpanExporter,
-  SimpleSpanProcessor,
-} from '@opentelemetry/sdk-trace-base';
+  getTestSpans,
```
david-luna: nit: isn't this already available in `testUtils` object? any particular reason to have an explicit import?

plugins/node/opentelemetry-instrumentation-redis/test/v4/redis.test.ts
```diff
@@ -75,12 +76,12 @@ describe('redis@^4.0.0', () => {
...
   });
 
-  let client: any;
+  let client: RedisClientType;
 
   beforeEach(async () => {
     client = createClient({
       url: redisTestUrl,
-    });
+    }) as unknown as RedisClientType;
```
david-luna: nit: I've checked out the branch and got the proper types from `createClient` function from the `redis` package. This type casting is probably unnecessary

david-luna:
and thanks to @blumamir & @trentm for working on this :)

trentm:
> Could we be more "noisy" about this change?

One of the followup tasks is to set `npm deprecate ...` on the old package name. IIUC, every `npm install ...` done by users will get a warning that it is deprecated. We can set a message that will be shown there to point users to the consolidated package.

plugins/node/opentelemetry-instrumentation-redis/test/v4/redis.test.ts
trentm: Okay, great, thanks. I got those from Amir's branch... which may have been necessary when the default installed redis (and its types) were for the legacy redis v3.

trentm: dropped in commit aef400b4

plugins/node/opentelemetry-instrumentation-redis/test/v2-v3/redis.test.ts
trentm: commit de0d5bf5, thanks.

plugins/node/opentelemetry-instrumentation-redis/src/v2-v3/instrumentation.ts
trentm: commit 3f925be3, thanks

blumamir (approved):
That looks great @trentm  thank you for bringing it to the finish line 💯

blumamir:
we should also remove redis-4 from the [component owner](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e45605d0e70158b0ea868bc3c8acb65095d6d4d1/.github/component_owners.yml#L135) also

blumamir:
and [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e45605d0e70158b0ea868bc3c8acb65095d6d4d1/packages/opentelemetry-redis-common/README.md?plain=1#L7) :)

trentm:
The "follow-ups" are complete.

PR #2923, opened 2025-07-03
fix: Enable union type support in @opentelemetry/instrumentation-graphql (#1506)

### LINKED ISSUE #1506: GraphQL field instrumentation does not work for union return types

@opentelemetry/api v 1.4.1

18.16.0

### What did you do?

Create a GraphQL API (I'm using Apollo but I don't believe it matters) with a query, mutation, or a field on an individual type whose response type is a union, eg:

```graphql
type Foo {
  field: String
}

type Bar {
  anotherField: String
}

union MyUnion = Foo | Bar

type Query {
  fooOrBar: MyUnion!
}
```

Then execute a GraphQL query, eg:

```graphql
{
  fooOrBar {
    ...on Foo {
      myField
    }
    ...on Bar {
      anotherField
    }
  }
}
```

Resolve spans for each resolver that ran (eg. if a `Foo` is returned, a span for `fooOrBar` and `myField`)

### What did you see instead?

A single resolve span for the query field only (`fooOrBar`). This is the case no matter how deeply the query nests.

### Additional context

I'm using the node SDK to instrument automatically. I also made sure to set `ignoreTrivialResponseSpans` to `false` and `depth` to `-1` (I know they are defaulted, but I wanted to be sure it wasn't related) like so:

```typescript
  // other settings...
  instrumentations: [getNodeAutoInstrumentations({
    '@opentelemetry/instrumentation-graphql': {
      ignoreTrivialResolveSpans: false,
      depth: -1,
    },
  })],
});
```

nitrofski:

Fixes #1506.

Enable support for installing field instrumentation on union types.

The "type unwrap" logic was generalized to support unwrapping to a list of types. It is now a recursive function rather than a loop. The new unwrapping has been implemented in a way that is more agreeable to TypeScript.


david-luna:
Hi @nitrofski and thanks for your contribution.

I order to proceed we need a couple of things:
- a [recent PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2928) which changes the folder structure has ben merged. Please sync your PR with the latest changes
- you need to sing the CLA so we can run the CI

Cheers

dyladan:
@obecny can you please take a look at this?

nitrofski:
@david-luna 

I merged with main.

There are issues with CLA approval at my company at the moment, which we are working on getting resolved. I will sign as soon as that's done. Sorry for the delay.

nitrofski:
CLA issues have been resolved.

dyladan:
@obecny can you please take a look at this?

nitrofski:
Good day,

This hasn't seen feedback in over a month. Is there anything I can do to help get attention on this PR?

packages/instrumentation-graphql/src/utils.ts
```diff
@@ -305,11 +305,7 @@ export function wrapFields(
   tracer: api.Tracer,
   getConfig: () => GraphQLInstrumentationParsedConfig
 ): void {
-  if (
-    !type ||
-    typeof type.getFields !== 'function' ||
```
david-luna, line 310: I'm not familiar enough with this library. Are we certain that all possible values of `type` have a function named `getFields`?

packages/instrumentation-graphql/src/utils.ts
```diff
@@ -328,16 +324,47 @@ export function wrapFields(
...
+  type: graphqlTypes.GraphQLOutputType
+): readonly graphqlTypes.GraphQLObjectType[] {
+  // unwrap wrapping types (non-nullable and list types)
+  if ('ofType' in type) {
+    return unwrapType(type.ofType);
+  }
+
+  // unwrap union types
+  if (isGraphQLUnionType(type)) {
+    return type.getTypes();
```
david-luna, line 345: Here I have a similar question. Are we certain that `getTypes()` returns an array of `GraphQLObjectType`. Could be possible to have nested unions?

packages/instrumentation-graphql/src/utils.ts
nitrofski: It is part of `graphqlTypes.GraphQLObjectType`. As long as we maintain type safety and that the type definitions are accurate (to my testing, they were), we should be safe.

The previous implementation used an `any` type before recursing, so they lost type safety and ended up passing non-compliant object. It had to re-assess the object here.

packages/instrumentation-graphql/src/utils.ts
nitrofski: Fair concern. Doing a quick search through the graphql type definitions, `GraphQLUnionType` is the only type with a `getTypes()` function, and it always returns an array of `GraphQLObjectType`. That all being said, is it enough for us or do we want to be extra safe and also validate what the function returns at runtime?

nitrofski: > Could be possible to have nested unions?

Doing a bit of research...

From the Apollo Graphql [Union and Interfaces](https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces) schema help page
> All of a union's included types must be [object types](https://www.apollographql.com/docs/apollo-server/schema/schema/#object-types) (not scalars, input types, etc.). Included types do not need to share any fields.

It sounds like unions strictly require their components to be objects and that nested union are not allowed. Though, to dispel any doubts, I tried launching my graphql server with a dummy union type that uses other union types:

```
Union type MyTestUnion can only include Object types, it cannot include <other union type>.
```

david-luna: thanks for your thorough explanation @nitrofski :)

packages/instrumentation-graphql/src/utils.ts
david-luna: okay then

david-luna:
@nitrofski thank for resolving my doubts. You have my approval but I'll wait a bit in case someone else wants to do a review. @obecny maybe?

david-luna:
@nitrofski I'm merging this today so it will be included in the [upcoming release](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3072)

Could I ping you if there is any issue in this new version? Thanks :)

nitrofski:
> Could I ping you if there is any issue in this new version? Thanks :)

Sure, I can help. FYI I have another PR on the same package coming.

PR #2824, opened 2025-05-13
feat(opentelemetry-sampler-aws-xray): Add Rules Caching and Rules Matching Logic

jj22ee:
## Which problem is this PR solving?

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2736
- This PR will be followed up by 1 more PR to complete the X-Ray Sampler Implementation

## Short description of the changes

This PR is a followup to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2750

- Add Sampling `RuleCache`
  - Caches a list of `SamplingRuleApplier`s, ordered by rule priority then rule name. Each Rule Applier corresponds to the Sampling Rule from GetSamplingRules. Each call to GetSamplingRules will only update the `Rule`s that have changed properties, to preserve the state of unchanged rules. This means when Reservoir and Statistics are implemented (later) in the `Rule`s, they will persist for unchanged rules.
  - The RuleCache will determine which Rule that a span matches (via the set of `{ResourceAttributes,SpanAttributes}`) that has highest priority.
- Update `SamplingRuleApplier` to perform Fixed Rate Sampling, and to include a method to apply matching logic against a set of {ResourceAttributes,SpanAttributes} by using the wild card and attribute matching from Utils
- Initial class for `FallbackSampler`
- Update `X-Ray Remote Sampler` to depend on the SamplingRuleApplier from `RuleCache` and the FallBack Sampler to perform `shouldSample`

Issue #2736, opened 2025-02-27
[sampler-aws-xray] X-Ray Remote Sampler full implementation


jj22ee:
### Is your feature request related to a problem? Please describe

I am creating this issue for tracking the X-Ray Remote Sampler implementation. See "xray" sampler in the [specification here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md). 

### Describe the solution you'd like to see

I am intending on completing the implementation of the X-Ray Remote Sampler here, originally started by its previous owner in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1443, but incomplete in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1443 and https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1466.

I have an implementation that was used internally in aws's ADOT instrumentation for about half a year now, so I intend to mostly copy over the implementation into the [current partial implementation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/incubator/opentelemetry-sampler-aws-xray).

### Additional context

The implementation will be similar to the [.NET](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Sampler.AWS) and [Java](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray) implementations.

pichlermarc:
Sounds good, the addition of the X-Ray remote sampler component was already approved previously. 🙂 

Ideally having a second owner would be beneficial - this way you can review each other's PRs which tends to speed up things (recommended but not required for vendor-components).

PR #2924, opened 2025-07-03
feat(opentelemetry-sampler-aws-xray): Add Rate Limiter and Sampling Targets Poller Logic

jj22ee:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2736
- This PR will complete the X-Ray Sampler Implementation

This PR is a followup to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2824

- Added logic to fetch sampling targets for each `SamplingRuleApplier`.
  - The sampling targets are periodically fetched every 10 seconds by making the [GetSamplingTargets API call to X-Ray](https://docs.aws.amazon.com/xray/latest/api/API_GetSamplingTargets.html).
  - The targets determine the reservoir quota and the rate at which a `SamplingRuleApplier` will sample the requests.
  - Each rule applier keeps and updates a sampling statistics document which is required in `GetSamplingTargets` call to determine the next target
- Added a `RateLimitingSampler` (applied before the fixed rate sampler) to be used in each rule applier.
  - Together these samplers determine how many requests to sample every second and what percentage of additional requests to sample in that second.
  - The `FallbackSampler` is updated to be a combination of above samplers to sample `1 req/sec` and `5%` of additional requests in that second.

This PR's implementation was tested against [X-Ray's Centralized Sampling Integration tests](https://github.com/aws-observability/aws-otel-community/tree/master/centralized-sampling-tests#start-integration-tests).


incubator/opentelemetry-sampler-aws-xray/src/remote-sampler.ts
```diff
@@ -36,11 +36,19 @@ import { FallbackSampler } from './fallback-sampler';
...
+  TargetMap,
 } from './types';
-import { RuleCache } from './rule-cache';
+import {
+  DEFAULT_TARGET_POLLING_INTERVAL_SECONDS,
+  RuleCache,
+} from './rule-cache';
 
 import { SamplingRuleApplier } from './sampling-rule-applier';
+import { PACKAGE_NAME } from './version';
```
lukeina2z, line 51: How this file is generated? I tried 'npm run setup:dev' and 'npm run compile', none of them generate this file and that is causing build failure.

incubator/opentelemetry-sampler-aws-xray/src/fallback-sampler.ts
```diff
@@ -21,16 +21,20 @@
...
+import { RateLimitingSampler } from './rate-limiting-sampler';
 
 // FallbackSampler samples 1 req/sec and additional 5% of requests using TraceIdRatioBasedSampler.
 export class FallbackSampler implements Sampler {
   private fixedRateSampler: TraceIdRatioBasedSampler;
+  private rateLimitingSampler: RateLimitingSampler;
 
   constructor() {
     this.fixedRateSampler = new TraceIdRatioBasedSampler(0.05);
+    this.rateLimitingSampler = new RateLimitingSampler(1);
```
lukeina2z: Consider allowing these values to be passed as params with default value. That will it more reusable.
```suggestion
  constructor(rateLimit: number = 1, ratio: number = 0.05) {
    this.fixedRateSampler = new TraceIdRatioBasedSampler(ratio);
    this.rateLimitingSampler = new RateLimitingSampler(rateLimit);
  }
```

incubator/opentelemetry-sampler-aws-xray/test/fallback-sampler.test.ts
```diff
@@ -18,11 +18,187 @@
...
+      []
+    );
+
+    // 0 seconds passed, 0 quota available
+    let sampled = 0;
+    for (let i = 0; i < 30; i++) {
+      if (
+        sampler.shouldSample(
+          context.active(),
+          '1234',
```
lukeina2z: Is it true that we intentionally use an invalid trace ID to avoid being sampled by the Fixed Rate Sampler? If that is the case, can we use an const variable instead of the magic string? Or maybe a comment here. That will make it easier to understand the purpose of the UT.
```suggestion
          const InvaidTraceIdToAvoidFixedRateSampler = '1234',
```

incubator/opentelemetry-sampler-aws-xray/src/rate-limiter.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public take(cost = 1): boolean {
+    if (this.quota === 0) {
+      return false;
+    }
+
+    const quotaPerMillis: number = this.quota / 1000.0;
+
+    // assume divide by zero not possible
```
lukeina2z, line 45: How about combine line 45 and 48: = (cost * 1000) / quota; so you don't need this comment.

incubator/opentelemetry-sampler-aws-xray/src/rate-limiting-sampler.ts
```diff
@@ -0,0 +1,58 @@
...
+      return {
+        decision: SamplingDecision.RECORD_AND_SAMPLED,
+        attributes: attributes,
+      };
+    }
+    return { decision: SamplingDecision.NOT_RECORD, attributes: attributes };
+  }
+
+  public toString(): string {
+    return `RateLimitingSampler{rate limiting sampling with sampling config of ${this.quota} req/sec and 0% of additional requests}`;
```
lukeina2z, line 56: nit: do we need 'and 0% of additional requests' in the context of rate limiting?

incubator/opentelemetry-sampler-aws-xray/src/aws-xray-sampling-client.ts
```diff
@@ -45,7 +45,7 @@ export class AWSXRaySamplingClient {
     this.makeSamplingRequest<GetSamplingTargetsResponse>(
       this.samplingTargetsEndpoint,
       callback,
-      this.samplerDiag.debug,
+      this.samplerDiag.debug.bind(this.samplerDiag),
```
lukeina2z: nit. lambda is modern, more readable.
(message: string, ...args: unknown[]) => this.samplerDiag.debug(message, ...args)

incubator/opentelemetry-sampler-aws-xray/src/rule-cache.ts
```diff
@@ -89,4 +98,61 @@ export class RuleCache {
...
+    });
+    return statisticsDocuments;
+  }
+
+  // Update ruleAppliers based on the targets fetched from X-Ray service
+  public updateTargets(
+    targetDocuments: TargetMap,
+    lastRuleModification: number
+  ): [boolean, number] {
+    let minPollingInteral: number | undefined = undefined;
```
lukeina2z: typo: do you mean 'interval'?

lukeina2z (approved):

dyladan:
@jj22ee I see [lukeina2z](https://github.com/lukeina2z) is also an AWS employee. Should he be considered a component owner or SME for this PR? I think this will be mergeable if so when the comments are addressed.

jj22ee:
@dyladan Yes, Luke is from AWS, and he can be considered as an SME for this PR. Will be addressing his comments soon.

incubator/opentelemetry-sampler-aws-xray/src/remote-sampler.ts
jj22ee: It is generated by `npm run version:update`. However, while this package is in the `incubator` folder, this command isn't automatically run for this package until it is moved out of this directory.

incubator/opentelemetry-sampler-aws-xray/test/fallback-sampler.test.ts
jj22ee: >Is it true that we intentionally use an invalid trace ID to avoid being sampled by the Fixed Rate Sampler?

No, that is unintended. Will use your [suggested approach](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2924#discussion_r2201587038) to directly set the Fallback sampler's ratio sampling config for this test.

incubator/opentelemetry-sampler-aws-xray/src/rate-limiting-sampler.ts
jj22ee: No, but I kept this text to be aligned with the Python implementation.

incubator/opentelemetry-sampler-aws-xray/src/rate-limiter.ts
jj22ee: Combined the lines, but kept the comment.

incubator/opentelemetry-sampler-aws-xray/test/fallback-sampler.test.ts
jj22ee: Done^

pichlermarc (approved):

PR #2920, opened 2025-06-27
chore: rename workspace dirs to packages/${package-base-name} (extreme edition)

trentm:
This is take 2 of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223
This implements option 3 of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223#issuecomment-2125925648
(i.e. the "extreme" option)

# Status

**DO NOT MERGE**

Let's not merge this yet, even if agreed on doing this. I think we can finish the instr-redis consolidation, and drop the redis-common package first.

PR #2928, opened 2025-07-07
chore: rename workspace dirs to packages/${package-base-name} (extreme edition)

trentm:
This PR is the final take of the plan discussed at #2223.
This implements option 3 of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223#issuecomment-2125925648
(i.e. the "extreme" option).

Before this change the packages published from this repo were spread
across these workspace dirs:

    "packages/*",
    "plugins/node/*",
    "plugins/web/*",
    "propagators/*",
    "detectors/node/*",
    "metapackages/*"

with the directory name *usually* matching the npm package name, and
with or without a leading "opentelemetry-". This PR normalizes all dirs
to be under "packages/*" and to have the base dir match the package
name (excluding the npm `@opentelemetry/` org prefix).

Obsoletes: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2223
Obsoletes: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2920


trentm:
For the record, here is the `allyourdir2.sh` script I used to automate most of this change:
https://gist.github.com/trentm/d95f0b0c1d3873968e16c2ccb1a13279

trentm:
The instr-mongodb TAV tests failed with Node.js v18 only when testing `mongodb@3.7.4`.
This passes for me locally. I suspect this is a flaky test.
I'm re-running.

```
-- installing ["mongodb@3.7.4"]
-- running test "npm run test-v3" with mongodb (env: {})

> @opentelemetry/instrumentation-mongodb@0.55.1 test-v3
> tav mongodb 3.7.4 npm run test-v3-run

-- required packages ["mongodb@3.7.4"]
-- installing ["mongodb@3.7.4"]
-- running test "npm run test-v3-run" with mongodb (env: {})

> @opentelemetry/instrumentation-mongodb@0.55.1 test-v3-run
> nyc --no-clean mocha --require '@opentelemetry/contrib-test-utils' 'test/**/mongodb-v3.test.ts'


  MongoDBInstrumentation-Tracing-v3
...
    requireParentSpan
      1) should not create spans without parent span when requireParentSpan is explicitly set to true
...


  21 passing (409ms)
  1 failing

  1) MongoDBInstrumentation-Tracing-v3
       requireParentSpan
         should not create spans without parent span when requireParentSpan is explicitly set to true:

1 !== 0

      + expected - actual

      -1
      +0
      
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-mongodb/test/mongodb-v3.test.ts:689:20
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

packages/instrumentation-kafkajs/tsconfig.json
david-luna, line 1: Strange that some packages have a change in this file but others, like this one,  appears as a new file when it already existed in the previous location.

david-luna (approved):
Thanks for working on this :)

package.json
david-luna, line 1: nit: now that the `plugins` folder is gone I wonder if we should keep the word in the keywords list. Also there is such word in some instrumentations. I think we could be more specific in the keywords section of each package. I can create an issue with a proposal and discuss it in the next SIG meeting

pichlermarc: > now that the plugins folder is gone I wonder if we should keep the word in the keywords list. Also there is such word in some instrumentations.

Yep, the `plugin` label can be removed for sure - it was the name for `instrumentation` from way back in the day.

> I think we could be more specific in the keywords section of each package. I can create an issue with a proposal and discuss it in the next SIG meeting

Agree, there's the `profiling` keyword everywhere. A proposal to clean this up would be appreciated. :)

maryliag (approved):

packages/instrumentation-kafkajs/tsconfig.json
trentm: I'm guessing that may be due to there being a number of "tsconfig.json" files with the same (or mostly similar) content that `git` couldn't reasonably disambiguate which file moved to where, so it gives up and just categorizes them as independent "deletes" and "creates".

PR #2939, opened 2025-07-16
feat(kafkajs): instrument transaction send and sendBatch

### LINKED ISSUE #2741: @opentelemetry/instrumentation-kafkajs does not instrument `sendBatch` or `send` methods if they are being called during KafkaJs transaction (e.g. `producer.transaction().sendBatch()`)

```
"@opentelemetry/auto-instrumentations-node": "^0.56.1",
"@opentelemetry/sdk-node": "^0.57.2",
```

v22.13.1

### What did you do?

I have code that sends batch messages during the kafkajs' transaction, e.g:

```
// producer is KafkajsProducer

const transaction = await producer.transaction();

transaction.sendBatch(batch)

await transaction.commit()

```

I expected to see span are emitted for each `sendBatch` call that happens during kafkajs' transaction and each message in kafka to have `traceparent` header if it is possible

### What did you see instead?

I didn't see any spans were emitted or `traceparent` header was set

### Additional context

I can confirm that if I run it without transaction I see spans are emitted and `traceparent` header is set

My best guess is `producer.transaction().sendBatch()` is not wrapped currently, only send methods of producer are wrapped (e.g `producer.sendBatch()`)

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-kafkajs/src/instrumentation.ts#L133
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/instrumentation-kafkajs/src/instrumentation.ts#L142

adbista:

- Fixes #2741 

- Instruments transaction.send and sendBatch with proper parent span (transactionSpan), which is ended on commit or abort


packages/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -503,11 +512,114 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
...
-      return function sendBatch(
+    return (original: Producer['transaction']) => {
+      return function transaction(
         this: Producer,
+        ...args: Parameters<Producer['transaction']>
+      ): ReturnType<Producer['transaction']> {
+        const transactionSpan = instrumentation.tracer.startSpan(
+          'transaction',
+          {
+            kind: SpanKind.INTERNAL,
```
seemk: The default span kind is internal, so the extra options can be omitted

packages/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -503,11 +512,114 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
...
+            };
+
+            const originalCommit = transaction.commit;
+            transaction.commit = function commit(
+              this: kafkaJs.Transaction,
+              ...args
+            ) {
+              const originCommitPromise = originalCommit.apply(this, args);
+              return instrumentation._endSpansOnPromise(
+                [transactionSpan],
```
seemk, line 582: I think setting the transaction span status to success on a successful commit makes sense here

seemk (approved):

PR #2946, opened 2025-07-18
fix(instrumentation-runtime-node): fix .isEnabled()

trentm:
This instrumentation overrides the enable() and disable() methods. However,
it was not calling super for those, so the private '_enabled' state on
the super class was not getting updated, which broke 'instr.isEnable()'.
It would always return false.

# example showing isEnabled() not working

```js
> var mod = require('@opentelemetry/instrumentation-runtime-node')
undefined
> var instr = new mod.RuntimeNodeInstrumentation()
undefined
> instr.isEnabled()
false
> instr.enable()
undefined
> instr.isEnabled()
false
```

That last `.isEnabled()` result should be `true`.

### DIFF
--- a/packages/instrumentation-runtime-node/src/instrumentation.ts
+++ b/packages/instrumentation-runtime-node/src/instrumentation.ts
@@ -73,6 +73,7 @@ export class RuntimeNodeInstrumentation extends InstrumentationBase<RuntimeNodeI
   }
 
   override enable() {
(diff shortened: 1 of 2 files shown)


david-luna:
Hopefully codecov will be improved when https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866. is merged.

PR #2947, opened 2025-07-19
fix: enable programmatic config when environment variables are unset

YangJonghun:

- fixes incorrect configuration priority where environment variables were overriding programmatic configuration
- aligns with OpenTelemetry JS SDK convention: **programmatic config > environment variables**
- resolves issue where users cannot override environment variables with explicit code configuration
- enables users to programmatically enable default-excluded instrumentations (like `fs`) regardless of environment variable settings

- **Establish correct configuration precedence**: programmatic config > environment variables > defaults
- **Fix priority order**: `userConfig.enabled` is now checked first, before any environment variable logic
- **Enable flexible overrides**: users can now override any environment variable setting with explicit programmatic configuration
- **Maintain consistency**: follows the same pattern as other OpenTelemetry JS SDK packages

### Configuration Priority (after changes):
1. **Programmatic config** (`userConfig.enabled: true/false`) - highest priority, always respected
2. **Environment variables** (`OTEL_NODE_DISABLED_INSTRUMENTATIONS`, `OTEL_NODE_ENABLED_INSTRUMENTATIONS`) - used when programmatic config is not set
3. **Default exclusions** (`defaultExcludedInstrumentations`) - applied when neither programmatic config nor env vars are set


packages/auto-instrumentations-node/src/utils.ts
```diff
@@ -166,15 +166,35 @@ export function getNodeAutoInstrumentations(
     // Defaults are defined by the instrumentation itself
     const userConfig: any = inputConfigs[name] ?? {};
 
-    if (
-      userConfig.enabled === false ||
-      !enabledInstrumentationsFromEnv.includes(name) ||
-      disabledInstrumentationsFromEnv.includes(name)
-    ) {
-      diag.debug(`Disabling instrumentation for ${name}`);
+    // Configuration priority: Environment variables > programmatic config
```
maryliag: There isn't a clear guideline of what takes precedent (env or programmatic), but in the JS SDK we have been using
programmatic > env
so to keep consistency across the SDK, I would suggest you make the changes accordingly here

YangJonghun: @maryliag 
Thank you for pointing this out! I've updated the PR to follow the **programmatic > env** convention as you suggested. ([e1e4e4b](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2947/commits/e1e4e4b199c66a5840f0b459ff239c281eb5260a))
I'd appreciate it if you could review the changes again. Thanks!

PR #2950, opened 2025-07-22
feat(redis): add support for redis v5

### LINKED ISSUE #2830: Support `redis` v5

adbista:

- Fixes #2830 

- Add support for Redis v5. 
- Replaced deprecated semantic attributes with current semantic conventions
- Updated tests


packages/instrumentation-redis/src/v2-v3/instrumentation.ts
```diff
@@ -144,15 +144,16 @@ export class RedisInstrumentationV2_V3 extends InstrumentationBase<RedisInstrume
...
         if (this.address) {
-          span.setAttribute(
-            SEMATTRS_DB_CONNECTION_STRING,
-            `redis://${this.address}`
-          );
+          // DB_CONNECTION_STRING is deprecated; Replaced by server.address and server.port.
+          // span.setAttribute(
+          //   SEMATTRS_DB_CONNECTION_STRING,
+          //   `redis://${this.address}`
+          // );
```
seemk: These lines (and the condition) can be removed now

seemk (commented):
The version number in CHANGELOG.md needs updating as  well :pray:

moekify:
Would be great to get this merged :)

dyladan:
@blumamir any chance you can get this on your review queue?

dyladan (changes requested):
Looks like this does not follow the guidance at https://github.com/open-telemetry/semantic-conventions/tree/main/docs/database

Specifically, the environment variable opt-in mechanism for new semconv breaking changes

ekarademir:
We are all rooting for this PR to be merged soon 😅

seemk (approved):

adbista:
I think the coverage issue arises because different Redis versions (v2-v3 and v4-v5) use separate code paths and tests for each version are run independently (v2-v3 and v4-v5 have distinct sets of tests) and this can lead to incorrect coverage report,  which is why the  codecov/patch failed.

trentm:
> I think the coverage issue arises because different Redis versions (v2-v3 and v4-v5) use separate code paths ...

Yup, exactly. Don't worry about the codecov failing check. Part of the coming work in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866 will change to getting coverage data for multiple target module versions using the `test-all-versions` tests rather than from just `npm test` for a single version.

adbista:
@dyladan @blumamir, is there anything else I should do for this pull request, or does everything look good?

packages/instrumentation-redis/src/v4-v5/instrumentation.ts
```diff
@@ -48,11 +55,26 @@ interface MutliCommandInfo {
...
+export class RedisInstrumentationV4_V5 extends InstrumentationBase<RedisInstrumentationConfig> {
   static readonly COMPONENT = 'redis';
+  private _semconvStability: SemconvStability;
 
   constructor(config: RedisInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
+    this._semconvStability = semconvStabilityFromStr(
+      'database',
+      process.env.OTEL_SEMCONV_STABILITY_OPT_IN
+    );
```
blumamir: we need to check if this is set in config here as well I think

```suggestion
    this._semconvStability = config.semconvStability
      ? config.semconvStability
      : semconvStabilityFromStr(
          'database',
          process.env.OTEL_SEMCONV_STABILITY_OPT_IN
        );
```

packages/instrumentation-redis/src/v2-v3/instrumentation.ts
```diff
@@ -37,14 +39,34 @@ import {
...
 export class RedisInstrumentationV2_V3 extends InstrumentationBase<RedisInstrumentationConfig> {
   static readonly COMPONENT = 'redis';
+  private _semconvStability: SemconvStability;
 
   constructor(config: RedisInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
+    this._semconvStability = semconvStabilityFromStr(
+      'database',
+      process.env.OTEL_SEMCONV_STABILITY_OPT_IN
+    );
```
blumamir:
```suggestion
    this._semconvStability = config.semconvStability
      ? config.semconvStability
      : semconvStabilityFromStr(
          'database',
          process.env.OTEL_SEMCONV_STABILITY_OPT_IN
        );
```

packages/instrumentation-redis/src/v2-v3/instrumentation.ts
```diff
@@ -127,28 +149,60 @@ export class RedisInstrumentationV2_V3 extends InstrumentationBase<RedisInstrume
 
         const dbStatementSerializer =
           config?.dbStatementSerializer || defaultDbStatementSerializer;
+
+        const attributes: { [key: string]: any } = {};
```
blumamir: Prefer the correct type here

```suggestion
        const attributes: Attributes = {};
```

where

```
import { Attributes } from '@opentelemetry/api';
```

packages/instrumentation-redis/src/v2-v3/instrumentation.ts
```diff
@@ -127,28 +149,60 @@ export class RedisInstrumentationV2_V3 extends InstrumentationBase<RedisInstrume
...
           }
         );
 
         // Set attributes for not explicitly typed RedisPluginClientTypes
         if (this.connection_options) {
-          span.setAttributes({
-            [SEMATTRS_NET_PEER_NAME]: this.connection_options.host,
-            [SEMATTRS_NET_PEER_PORT]: this.connection_options.port,
-          });
+          const connectionAttributes: { [key: string]: any } = {};
```
blumamir:
```suggestion
          const connectionAttributes: Attributes = {};
```

packages/instrumentation-redis/src/types.ts
```diff
@@ -56,4 +59,7 @@ export interface RedisInstrumentationConfig extends InstrumentationConfig {
 
   /** Require parent to create redis span, default when unset is false */
   requireParentSpan?: boolean;
+
+  /** Controls which semantic convention attributes are emitted on spans. */
+  semconvStability?: SemconvStability;
```
blumamir: nit: consider mentioning the default here (OLD?), so it's straight forward for normal users.
also can be nice to mention that this configuration will overwrite any value set via the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable

packages/instrumentation-redis/src/v4-v5/instrumentation.ts
```diff
@@ -379,12 +404,24 @@ export class RedisInstrumentationV4 extends InstrumentationBase<RedisInstrumenta
...
+    const attributes = getClientAttributes(
+      this._diag,
+      clientOptions,
+      this._semconvStability
+    );
 
     try {
+      if (this._semconvStability & SemconvStability.STABLE) {
+        attributes[ATTR_DB_OPERATION_NAME] = commandName;
+      }
```
blumamir: nit: the `try catch` is for the `dbStatementSerializer` call.

Operation Name does not belong here IMO

packages/instrumentation-redis/src/v4-v5/instrumentation.ts
```diff
@@ -447,8 +484,15 @@ export class RedisInstrumentationV4 extends InstrumentationBase<RedisInstrumenta
...
     }
+
     for (let i = 0; i < openSpans.length; i++) {
       const { span, commandName, commandArgs } = openSpans[i];
+      if (
+        openSpans.length >= 2 &&
+        this._semconvStability & SemconvStability.STABLE
+      ) {
+        span.setAttribute(ATTR_DB_OPERATION_BATCH_SIZE, openSpans.length);
+      }
```
blumamir: I wonder if this attribute belongs here:

> The number of queries included in a batch operation.

Each of this span is a single operation, not a batch, so this feels wrong to me.

If we had one parent span for the "MULTI" command than it makes sense to me, but here to my understanding, this is not the correct use

PR #2954, opened 2025-07-23
chore(instr-undici): use tsup for cjs,esm output

david-luna:

OpenTelemetry JS modules do not have valid ESM files in the compilation result. See https://github.com/open-telemetry/opentelemetry-js/issues/3989

This PR is an attempt to fix it with adding a new tool named `tsup` which compiles the code with `esbuild` under the hood.
The scope is reduced to `@opentelemetry/instrumentation-undici` because this implies changes in the toolchain and needs to be discussed in the Javascript SIG

- Added `tsup` dependency
- changed compile script to use `tsup`
- changed the `package.json` to get the proper mapping foe cjs and esm modules
  - add `"exports"` section
  - modify `files` section

NOTES:

- this add a new tool in the repo (esbuild) which is faster but it does not check the types (actually the reason why it's faster)
- test scripts still use `tsc` but IMO it is odd to have a type error it in compilation time but in test
- `tsup` does emit, in this case, a single file with all the instrumentation code instead of emit a single file per `.ts` file. Not sure if this is a good or bad thing

### Notice

This PR is meant to be reviewed along with https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2955 which does the same but using the current toolchain of the repo. The intention is to have a conversation on the path we want to take about tooling. Should we stick to the current tools (for compilation) or change them?

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -22,7 +22,8 @@
         "{projectRoot}/test/**/*.ts"
       ],
       "outputs": [
-        "{projectRoot}/build"
+        "{projectRoot}/build",
+        "{projectRoot}/dist"
       ]
     },
     "watch": {
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ build/Release
 node_modules/
 jspm_packages/
 build/
+dist/
 
 # TypeScript v1 declaration files
 typings/
--- a/package-lock.json
+++ b/package-lock.json
@@ -12530,6 +12530,21 @@
         "node": ">= 0.10.x"
       }
     },
+    "node_modules/bundle-require": {
+      "version": "5.1.0",
+      "resolved": "https://registry.npmjs.org/bundle-require/-/bundle-require-5.1.0.tgz",
+      "integrity": "sha512-3WrrOuZiyaaZPWiEt4G3+IffISVC9HYlWueJEBWED4ZH4aIAC2PnkdnuRrR94M+w6yGWn4AglWtJtBI8YqvgoA==",
+      "dev": true,
+      "dependencies": {
+        "load-tsconfig": "^0.2.3"
+      },
+      "engines": {
+        "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+      },
+      "peerDependencies": {
+        "esbuild": ">=0.18"
+      }
+    },
     "node_modules/bunyan": {
       "version": "1.8.15",
       "resolved": "https://registry.npmjs.org/bunyan/-/bunyan-1.8.15.tgz",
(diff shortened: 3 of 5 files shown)


JamieDanielson:
- Just for easy finding, linking to tracking issue in core https://github.com/open-telemetry/opentelemetry-js/issues/4898

t2t2:
> `tsup` does emit, in this case, a single file with all the instrumentation code instead of emit a single file per `.ts` file. Not sure if this is a good or bad thing

esbuild has a [mentioned-in-documentation-but-never-explained](https://esbuild.github.io/api/#build) config option `bundle` for this, so adding `bundle: false,`  (`--no-bundle` ?) will emit them separately again (if desired - considering the potential upsides for bundling were mentioned in the sig)

...however there's a bug in tsup where it either drops or puts a wrong type of file extension in requires/imports (either .mjs trying to load extensionless .js (commonjs) or (when package.json has type: module) .cjs files trying to load .js (esm)) - [this snippet has worked for me in the second case](https://github.com/egoist/tsup/issues/953#issuecomment-2294998890) but it requires a config file to add a plugin

david-luna:
@t2t2 thanks for the info about `esbuild` bundle option :)

In last JS SIG (2025-07-23) there was a comment about the potential performance benefit of having the code in a single file. So I'll keep the current options for now.

This PR is now ready to compare with https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2955. I'll add the `no-bundle` option if it's required for the discussion on which approach we should take.

nx.json
```diff
@@ -22,7 +22,8 @@
         "{projectRoot}/test/**/*.ts"
       ],
       "outputs": [
-        "{projectRoot}/build"
+        "{projectRoot}/build",
+        "{projectRoot}/dist"
```
dyladan, line 26: Is there a reason we need both build and dist?

dyladan (commented):
LGTM. Looks a lot easier than I expected thankfully

nx.json
david-luna: This is because I couldn't find an option to change the destination dir in tsup. So for now nx needs to cache build and dist. If we move forward with this all packages will have the compilation output to dist so the bukd folder will disappear from this config.

dyladan (approved):
Sorry meant to mark this as approved before not just "comment"

david-luna:
> Sorry meant to mark this as approved before not just "comment"

Does that mean you prefer this solution over https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2955?

pichlermarc:
@david-luna do you think it would be feasible to use `esbuild` and `rollup` directly for this change. That would cut down on one extra dependency, but would likely require a few more config lines.

trentm:
There was some offline discussion on your two PRs, but not hard decision on preference for either of your PRs yet.

trentm:
> do you think it would be feasible to use `esbuild` and `rollup` directly for this change.

A possible example of doing so manually might be (I haven't tested): https://github.com/open-feature/js-sdk/blob/dc1970e717a8cbe7261d52d993c8811059718bed/packages/server/package.json#L21-L23

david-luna:
going to use a more suitable package for this PoC

PR #2955, opened 2025-07-23
chore(instr-undici): add ESM output

david-luna:

This is a different approach of #2954 which uses the current toolchain to generate the ESM files. Basically is another try on a [similar PR targeting express](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1859) which at that time did not have the best tools to work.

- set `"type": "module"` in `package.json`
- add `exports` property in `package.json`
- add `tsconfig.esm.json` for ESM compilation
- modify `tsconfig.json` for CJS compilation
- add the `.js` extension in import statements

2025-07-24 UPDATE: to simplify integration with `mocha` and `ts-node/esm` loader for tests I've changed the compilation step
- the default compilation output is ESM
- there is a specific `tsconfig.cjs.json` file for CJS output
- for now the `ts-node/esm` loader is set in the test script. IF this change is applied to all other packages we could promote this config to `.mocharc.yml`

### Notice

This PR is meant to be reviewed along with https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2954 which does the same but using a new tool for compilation. The intention is to have a conversation on the path we want to take about tooling. Should we stick to the current tools (for compilation) or change them?


JamieDanielson:
- Just for easy finding, linking to tracking issue in core https://github.com/open-telemetry/opentelemetry-js/issues/4898

david-luna:
going to use a more suitable package for this PoC

Issue #2952, opened 2025-07-23
Unexpected ANSI color codes appearing in the severityText field


mznet:
### What version of OpenTelemetry are you using?
@opentelemetry/auto-instrumentations-node@0.62.0

### What version of Node are you using?
v18.20.7

### What did you do?

An example application prints logs at three different levels. In the OpenTelemetry logs emitted by this application, the severityText field contains both the log level and ANSI color codes—for example, \x1B[32minfo\x1B[39m, which represents the word “info” in green.

```
LogRecordImpl {
    hrTime: [ 1753258688, 585000000 ],
    hrTimeObserved: [ 1753258688, 585000000 ],
    spanContext: undefined,
    resource: ResourceImpl {
      _rawAttributes: [Array],
      _asyncAttributesPending: false,
      _memoizedAttributes: undefined
    },
    instrumentationScope: {
      name: '@opentelemetry/winston-transport',
      version: '0.14.0',
      schemaUrl: undefined
    },
    attributes: { timestamp: '2025-07-23T08:18:08.585Z' },
    _severityText: '\x1B[31merror\x1B[39m',
    _severityNumber: undefined,
    _body: 'Test error message',
    _eventName: undefined,
    totalAttributesCount: 1,
    _isReadonly: true,
    _logRecordLimits: { attributeCountLimit: 128, attributeValueLengthLimit: Infinity }
  }
```

<img width="986" height="604" alt="Image" src="https://github.com/user-attachments/assets/33939a32-e5d5-44cf-b13b-84fcf0726d22" />

### What did you expect to see?
- SeverityText should contain only log level.

### What did you see instead?
- SeverityText contains both log level and color code.

### Additional context
https://gist.github.com/mznet/154f00599ae05a08f58a85454e73f90b

trentm:
I can definintely repro. Using the example winston-using script from the gist above and running that instrumented with an SDK that enables instrumentation-winston (and its `logSending` feature), I see this OTLP data sent from the SDK:

```
ExportLogsServiceRequest {
  resourceLogs: [
    ResourceLogs {
      scopeLogs: [
        ScopeLogs {
          logRecords: [
            LogRecord {
              attributes: [ KeyValue { key: 'timestamp', value: AnyValue { stringValue: '2025-07-23T17:30:11.411Z' } } ],
              timeUnixNano: Long { low: -155125056, high: 408220060, unsigned: true },
[...]

PR #2956, opened 2025-07-23
fix(winston-transport, instrumentation-winston): get the correct severityText and severityNumber with winston.format.colorize() is used

trentm:
Before this change the "log sending" feature of instr-winston would
incorrectly include ANSI color codes in the `severityText` for emitted
OTel LogRecords when the winston logger being instrumented was
configured with the `colorize()` formatter, e.g.:

    const logger = winston.createLogger({
      format: winston.format.combine(
        winston.format.timestamp(),
        winston.format.colorize(),
        winston.format.simple()
      ),
      ...

This is because the winston-transport.OpenTelemetryTransportV3 was using
the Winston log info's "level" attribute... which was tweaked by
`colorize()`. Winston log `info` objects have a `Symbol.for("info")`
property which has the unchanged level name:
  https://github.com/winstonjs/winston#streams-objectmode-and-info-objects

This commit *also* changes (fix) what `npm test` runs in the
instrumentation-winston package. Before this, `npm test` was attempting
to test against winston@2 and winston@3 -- presumably for better
coverage. However, it ended up only ever testing with winston@2
(`test-v1-v2` would install winston@2 and test, then `test-v3` would
run the same test again with whatever current version is installed,
which is now winston@2).

The (IMO) correct answer here is:
- `npm test` tests against the latest ver of the target module (winston@3)
- `npm run test-all-versions` handles testing against other versions
- coverage can take a back seat. (Aside: changes coming as part of
  https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2866
  will, IIUC, provide coverage from test-all-versions runs.)

Closes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2952

### DIFF
--- a/packages/winston-transport/src/utils.ts
+++ b/packages/winston-transport/src/utils.ts
@@ -60,19 +60,20 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
 }
 
 export function emitLogRecord(
-  record: Record<string, any>,
+  record: Record<string | symbol, any>,
   logger: Logger
 ): void {
   const { message, level, ...splat } = record;
   const attributes: LogAttributes = {};
+  const levelSym = record[Symbol.for('level')];
   for (const key in splat) {
     if (Object.prototype.hasOwnProperty.call(splat, key)) {
       attributes[key] = splat[key];
     }
   }
   const logRecord: LogRecord = {
-    severityNumber: getSeverityNumber(level),
-    severityText: level,
+    severityNumber: getSeverityNumber(levelSym),
+    severityText: levelSym,
     body: message,
     attributes: attributes,
   };
--- a/packages/winston-transport/src/utils.ts
+++ b/packages/winston-transport/src/utils.ts
@@ -65,6 +65,8 @@ export function emitLogRecord(
 ): void {
   const { message, level, ...splat } = record;
   const attributes: LogAttributes = {};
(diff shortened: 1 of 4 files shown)


trentm:
> Before this, `npm test` was attempting
> to test against winston@2 and winston@3 -- presumably for better
> coverage.

Yup.
My changes here are undoing what was done in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2684

packages/winston-transport/src/utils.ts
```diff
@@ -60,19 +60,20 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
 }
 
 export function emitLogRecord(
-  record: Record<string, any>,
+  record: Record<string | symbol, any>,
   logger: Logger
 ): void {
   const { message, level, ...splat } = record;
   const attributes: LogAttributes = {};
+  const levelSym = record[Symbol.for('level')];
```
hectorhdzg, line 71:
```suggestion
  // Ensures the log level is read from a symbol property, avoiding any accidental inclusion 
  // of ANSI color codes that may be present in the string property
  const levelSym = record[Symbol.for('level')];
```

hectorhdzg (approved):
LGTM, thanks for taking care of this

PR #2957, opened 2025-07-23
feat(instrumentation-koa): Add support for koa 3

tlindhardt:

- During the upgrade of our project from v2 to v3 I noticed that the span name was not correct and the attributes were missing. Upon further inspection I found that the instrumentation was being skipped entirely.

- Switched to ATTR_HTTP_ROUTE from SEMATTRS_HTTP_ROUTE
- Add support to instrumentation-koa for [koa v3](https://github.com/koajs/koa/releases/tag/v3.0.0)
- Updated tests
  - This includes removing a test to validate the generator middleware is skipped as that is no longer supported and mentioned in the [upgrade guide](https://github.com/koajs/koa/blob/master/docs/migration-v2-to-v3.md)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: 18
           cache: 'npm'
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -24,7 +24,7 @@ jobs:
       # - `build-and-cache` job of unit-test workflow
       # - `build-and-cache` job of test-all-versions workflow
       - name: Setup Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: 18
           registry-url: 'https://registry.npmjs.org'
--- a/.github/workflows/test-all-versions.yml
+++ b/.github/workflows/test-all-versions.yml
@@ -29,7 +29,7 @@ jobs:
         with:
           fetch-depth: 0
       # Use the same Node.js version used for `release-please` workflow.
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: 18
       - name: Install
@@ -163,7 +163,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: ${{ matrix.node }}
       - name: Install
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           fetch-depth: 0
       # Use the same Node.js version used for `release-please` workflow.
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: 18
       - name: Install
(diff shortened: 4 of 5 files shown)


tlindhardt:
I wasn't sure of the release process and if I needed to also change the manifest file or not.

packages/instrumentation-koa/test/koa.test.ts
```diff
@@ -502,35 +497,6 @@ describe('Koa Instrumentation', function () {
       );
     });
 
-    it('should not instrument generator middleware functions', async () => {
```
JacksonWeber, line 505: Why is this test no longer necessary?

tlindhardt: [support](https://github.com/koajs/koa/blob/master/docs/migration-v2-to-v3.md) for generator functions have been removed in v3

dyladan:
@seemk @JacksonWeber are either of you interested in becoming component owner here?

packages/instrumentation-koa/src/instrumentation.ts
```diff
@@ -44,7 +44,7 @@ export class KoaInstrumentation extends InstrumentationBase<KoaInstrumentationCo
   protected init() {
     return new InstrumentationNodeModuleDefinition(
       'koa',
-      ['>=2.0.0 <3'],
+      ['>=2.0.0 <=3.0.0'],
```
raphael-theriault-swi: This will only work for 3.0.0, and there's already a 3.0.1 out

```suggestion
'koa',
      ['>=2.0.0 <4'],
```

tlindhardt: Ive addressed this.

tlindhardt:
I went ahead and applied the requested changes.

packages/instrumentation-koa/.tav.yml
```diff
@@ -13,6 +13,6 @@
 koa:
   # Testing ^2.7.0 covers at least 97% of the downloaded koa versions
   versions:
-    include: "^2.7.0"
+    include: "^3.0.1"
```
david-luna: With this change only v3 and up are going to be tested. So we won't know if this instrumentation is still compatible with the `^2.7.0` range because it wont be tested anymore.

```suggestion
    include: ">=2.7.0 <4"
```

packages/instrumentation-koa/test/koa.test.ts
david-luna: See my comment above. We still need to test v2 if we claim to support it. I would skip this test if the version of Koa is v3 or above. Getting the version of the package would be tricky but I think we can check for API changes.

The new version v3 removed the `createAsyncCtxStorageMiddleware` API in https://github.com/koajs/koa/pull/1817. We could check for its presence to verify the version under test is v2.

david-luna:
@tlindhardt 

Thanks for contributing to OpenTelemetry. And sorry for getting here after 3 weeks.

I'll have this PR in my radar from now on. Please resolve the conflicts and read my comments

david-luna:
I sponsor this PR.

tlindhardt:
@david-luna I went ahead and made the requested changes. I wasn't exactly sure the best way to not run the test in 3.0 but let me know what you think.

Thanks!

packages/instrumentation-koa/test/koa.test.ts
```diff
@@ -502,7 +502,11 @@ describe('Koa Instrumentation', function () {
...
+    it('should not instrument generator middleware functions', async function () {
+      if (typeof (app as any).createAsyncCtxStorageMiddleware !== 'function') {
```
david-luna, line 506: Nit: a comment here saying why we skip the test for koa@3 would be helpful.

PR #2959, opened 2025-07-27
feat(instrumentation-cucumber): switch from deprecated attributes to stable

Ugzuzg:

- Move from deprecated to stable semantic attributes in cucumber instrumentation


packages/instrumentation-cucumber/src/instrumentation.ts
```diff
@@ -182,10 +181,9 @@ export class CucumberInstrumentation extends InstrumentationBase<CucumberInstrum
           {
             kind: SpanKind.CLIENT,
             attributes: {
-              [SEMATTRS_CODE_FILEPATH]: gherkinDocument.uri,
-              [SEMATTRS_CODE_LINENO]: scenario.location.line,
-              [SEMATTRS_CODE_FUNCTION]: scenario.name,
-              [SEMATTRS_CODE_NAMESPACE]: feature.name,
+              [ATTR_CODE_FILE_PATH]: gherkinDocument.uri,
+              [ATTR_CODE_LINE_NUMBER]: scenario.location.line,
+              [ATTR_CODE_FUNCTION_NAME]: `${feature.name} ${scenario.name}`,
```
maryliag, line 186: looks like the function attribute is a combination of feature name and scenario name now, so can you also update the readme to reflect that?

PR #2962, opened 2025-07-31
chore(instrumentation-express, instrumentation-fastify): drop unused devDep @types/express

trentm:
For some background, in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1804 the `@types/express` dep was removed from instr-express's *dependencies* after some long discussion. At the time it was moved to a devDep.

However, `@types/express` isn't used at all in these packages. Can we drop them?
Or is there some slight downside for possible IDE completion when developing express examples or similar in this repo by removing these?

Note that I only bothered looking at these because of this bot PR to update these deps: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2925#issuecomment-3137147990


david-luna:
In `@opentelemetry/instrumentation-express` the types are used [internally](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/c0593e65a6d74425a0922e358e4fd235e7112148/packages/instrumentation-express/src/internal-types.ts#L17)

So the compilation step should fail if the types are not present. But if we list where the dependency is
```sh
npm ls @types/express
opentelemetry-contrib@0.1.0 /Users/david/Documents/repos/otel/opentelemetry-js-contrib
├─┬ @opentelemetry/instrumentation-express@0.52.0 -> ./packages/instrumentation-express
│ └── @types/express@4.17.23
├─┬ @opentelemetry/instrumentation-fastify@0.48.0 -> ./packages/instrumentation-fastify
│ └── @types/express@4.17.23 deduped
└─┬ @opentelemetry/instrumentation-koa@0.51.0 -> ./packages/instrumentation-koa
  └─┬ @types/koa@2.15.0
    └─┬ @types/cookies@0.9.1
      └── @types/express@4.17.23 deduped
```

We can see that we get the express types as a dependency from `@types/koa` so the editor and the compiler can see them because they are installed at the root level. That would explain why compilation succeeds.

IMHO removing `@types/express` from `@opentelemetry/instrumentation-fastify` is okay but we should keep them for `@opentelemetry/instrumentation-express`. `@opentelemetry/instrumentation-koa` is unlikely to change but there is the possibility `@types/koa` gets an update that removes the dependency.

trentm:
> In `@opentelemetry/instrumentation-express` the types are used [internally](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/c0593e65a6d74425a0922e358e4fd235e7112148/packages/instrumentation-express/src/internal-types.ts#L17)

Thanks!  Updated.

PR #2969, opened 2025-08-01
chore(lint): switch from 'header' to 'license-header' eslint plugin

trentm:
Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2967
Refs: https://github.com/nikku/eslint-plugin-license-header/pull/26

# checklist

This is still in draft because it is incomplete. The current `license-header` eslint plugin won't suffice until https://github.com/nikku/eslint-plugin-license-header/pull/26 or similar functionality is added to it. This is because the license header for this repo allows for some variance. Currently `license-header` required a single version.

- [ ] get the license-header PR in (hopefully), and a new release
- [ ] update to that release of `license-header`
- [ ] Customize the pattern for instr-oracledb which has another variance. We could technically allow that variance in the top-level regex so a separate custom regex isn't required.
- [ ] discuss whether the strict newline separation enforced by `license-header` is acceptable


trentm:
Obsoleted by the work in #3001 -- we are now using https://github.com/trentm/eslint-plugin-yet-another-license-header, which is effectively the license-header eslint plugin functionality with the added fix that I was trying to upstream.

PR #2970, opened 2025-08-04
feat(instrumentation-aws-lambda): support streaming handlers

### LINKED ISSUE #2827: AWS Lambda Instrumentation causes lambda failures when using streaming response
Latest

Node.js 22.x

### What did you do?
* Create new Lambda function using Node.js 22.x with function body:
```
export const handler = awslambda.streamifyResponse(
  async (event, responseStream, _context) => {
    console.log(event)
    console.log(responseStream)
    console.log(_context)
    responseStream.write('test payload');
    responseStream.end();
  }
)
```
* Test above with helloworld - Success (see [success.log](https://github.com/user-attachments/files/20216847/success.log))
* Checkout https://github.com/open-telemetry/opentelemetry-lambda
* from `nodejs` folder `npm install` & `npm run build`
* Create lambda layer from resultant zip.
* Add lambda layer to function, set env var: `AWS_LAMBDA_EXEC_WRAPPER /opt/otel-handler`
* Test again with helloworld - Failure (see [failure.log](https://github.com/user-attachments/files/20216848/failure.log)):
* `TypeError: responseStream.write is not a function`

Expected instrumentation to work.

### What did you see instead?
Instrumentation caused my lambda to fail.

### Additional context
The problem appears to be [`opentelemetry-instrumentation-aws-lambda`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-lambda). The [instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-lambda/src/instrumentation.ts) basically looks for any `handler` and replaces it with a patched `handler` that has 3 args: event, context, and callback. However, with [streaming response lambdas](https://docs.aws.amazon.com/lambda/latest/dg/config-rs-write-functions.html) the handler is `streamifyReponse` which needs 4 args: event, context, responseStream, and callback. So, `responseStream` is overwritten with callback and becomes unusable.

Two solutions:
1. [Short term] Do not instrument if `handler` is a `streamifyReponse` - Callout, this means streaming response lambdas will not be supported
2. [Long term] Properly instrument `streamifyResponse`, fully supporting both "normal" and streaming response lambdas.

Let me know if I should cut a separate issue for 2, above.

msonnb:

Fixes #2827 

Adds support for AWS Lambda streaming handlers by checking if the handler has the special `aws.lambda.runtime.handler.streaming` symbol.


packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -280,27 +335,98 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+  ): Promise<{}> | undefined {
+    if (typeof maybePromise?.then === 'function') {
+      return maybePromise.then(
+        value => {
+          this._applyResponseHook(span, null, value);
+          return new Promise(resolve =>
+            this._endSpan(span, undefined, () => resolve(value))
+          );
+        },
+        (err: Error | string) => {
```
JacksonWeber, line 368: Is it intentional that the promise fails silently in this scenario?

msonnb: I'm not fully sure what you mean 🤔 we return a new promise that rejects with the error after the span is ended, so it shouldn't fail silently? (Also I didn't change this part, just extracted it into a function)

pichlermarc:
cc @jj22ee (component owner)

jj22ee:
Hey, thanks for the fix. Planning to take a look this week.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -175,12 +193,40 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+        for (const symbol of Object.getOwnPropertySymbols(original)) {
+          (patchedHandler as unknown as Record<symbol, unknown>)[symbol] = (
+            original as unknown as Record<symbol, unknown>
+          )[symbol];
+        }
+
+        return patchedHandler;
+      }
+
       return this._getPatchHandler(original, handlerLoadStartTime);
```
```suggestion
      const patchedHandler = this._getPatchHandler(
        original,
        handlerLoadStartTime
      );
      if (this._isStreamingHandler(original)) {
        // Streaming handlers have special symbols that we need to copy over to the patched handler.
        for (const symbol of Object.getOwnPropertySymbols(original)) {
          (patchedHandler as unknown as Record<symbol, unknown>)[symbol] = (
            original as unknown as Record<symbol, unknown>
          )[symbol];
        }
      }

      return patchedHandler;
```

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -91,6 +96,19 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
       return [];
     }
 
+    // Provide a temporary awslambda polyfill for CommonJS modules during loading
+    // This prevents ReferenceError when modules use awslambda.streamifyResponse at load time
+    if (typeof globalThis.awslambda === 'undefined') {
```
jj22ee: Not familiar with polyfill workarounds. Wondering why this is needed inside the Lambda instrumentation? I don't see this method referenced inside this package apart from unit-tests, as your `patchedStreamingHandler` will just invoke `original.apply` as a generic way to run the streamified handler, without needing to know it is instance of `awslambda.streamifyResponse`. 

- Is this polyfill workaround just for the unit-tests?
- Why will AWS Lambda users will not encounter this `ReferenceError` when not using Lambda Instrumentation, but will encounter `ReferenceError` when using Lambda Instrumentation?

msonnb: You are right, I tested it again and it looks like this isn't needed 👍 
Probably ran into some weird setup situation when I tested it for the first time.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
```diff
@@ -827,4 +831,400 @@ describe('lambda handler', () => {
       );
     });
   });
+
+  describe('streaming handlers', () => {
+    const createMockResponseStream = () => ({
+      write: () => {},
```
jj22ee: It might be valuable to add an explicit test case (or update existing ones) to validate that `write()` receives the expected inputs from your test functions in `../lambda-test/streaming.js`, just to ensure that the lambda streaming functionality is working as normal. Since the existing tests are invoking `write()` & `end()` without error, that kind of at least implies that the `createMockResponseStream` methods are indeed being called though.

jj22ee (approved):
Thx @msonnb for the streaming support, all looks good to me!
Approved with a suggestion for the unit tests.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
msonnb: done, thanks for the review!

pichlermarc:
@msonnb thanks for your contribution. 🙌 Looks like there was a force-push, which makes it difficult to determine if the PR still contains what @jj22ee reviewed. Please only merge `main` in the future instead of rebasing after a review has been posted, that tends to speed things up :) 

@jj22ee could you have another look? Then we can merge this in - thanks. 🙂

pichlermarc (approved):
[cbe5594](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/cbe5594f00141b15cd890976b7a190cee2ba874f)

PR #2971, opened 2025-08-05
feat(instrumentation-aws-sdk): Add Semantic Convention Support for Amazon Secrets Manager Attributes

lukeina2z:
This PR adds the aws.secretsmanager.secret.arn semantic convention attribute for the following AWS resources:

AWS Secrets Manager SDK

The Secret ARN is extracted from both request and response objects, and this behavior is covered by unit tests.

Tests Run:
npm run compile
npm run lint
npm run test

All newly added tests pass, and no regressions were found.

Backward Compatibility:
This change is fully backward compatible. It introduces instrumentation for an additional AWS resource without modifying existing behavior in the auto-instrumentation library.


packages/instrumentation-aws-sdk/test/secretsmanager.test.ts
```diff
@@ -0,0 +1,121 @@
...
+          Name: 'testId',
+        });
+
+      await secretsManager
+        .getSecretValue({
+          SecretId: 'testSecret',
+        })
+        .catch((err: any) => {
+          console.log(err);
+        });
```
jj22ee: For both unit tests, can the `catch` be removed? It may not be necessary.

packages/instrumentation-aws-sdk/src/services/secretsmanager.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  responseHook(
+    response: NormalizedResponse,
+    span: Span,
+    tracer: Tracer,
+    config: AwsSdkInstrumentationConfig
+  ): void {
+    const secretArn = response.data?.ARN;
+    if (secretArn) {
+      span.setAttribute(ATTR_AWS_SECRETSMANAGER_SECRET_ARN, secretArn);
```
jj22ee, line 57: Curious since both request and response hook both sets this `ATTR_AWS_SECRETSMANAGER_SECRET_ARN`.

- In the event where the Client Call does succeed, is it be expected that `response.data?.ARN` would be the exact same content as `request.commandInput?.SecretId`? Meaning that the response hook is repeating the work done in the request hook? Regardless, I am OK with this current approach because `response.data?.ARN` seems like the better source of truth anyways, even if it is expected to be the same value...

lukeina2z: The code was ported from this PR https://github.com/aws-observability/aws-otel-js-instrumentation/pull/121.  My understanding is that the ARN from the request is a guess, while the ARN from the response is the ultimate source of truth. I don't think any SecretsManager SDK APIs will set/require ARN in the request.

packages/instrumentation-aws-sdk/test/secretsmanager.test.ts
lukeina2z: Sure. Deleted them in the new iteration.

jj22ee (approved):

trivikr (approved):

packages/instrumentation-aws-sdk/package.json
```diff
@@ -59,6 +59,7 @@
     "@aws-sdk/client-kinesis": "^3.85.0",
     "@aws-sdk/client-lambda": "^3.85.0",
     "@aws-sdk/client-s3": "^3.85.0",
+    "@aws-sdk/client-secrets-manager": "3.85.0",
```
trentm:
```suggestion
    "@aws-sdk/client-secrets-manager": "^3.85.0",
```

lukeina2z: Great catch. Thank you!

PR #2813, opened 2025-05-05
fix(@opentelemetry/instrumentation-dataloader): fix instrumentation of ESM-imported dataloader

TheDadi:
## Which problem is this PR solving?

- The ESM imported top-level object is a Module Namespace Object per ECMA262 spec. The ".default" holds the equivalent of require('dataloader')

`TypeError: Cannot read properties of undefined (reading 'load')
    at DataloaderInstrumentation._patchLoad `

Contributes to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1942

PR #2973, opened 2025-08-07
fix(instrumentation-dataloader): support ESM imports of dataloader module

brianphillips:
**NB:** code change primarily comes from #2813 and relates to #1942.

- This allows ESM applications to successfully import (vs. require) the dataloader package while instrumenting via otel

- Allows for `default` export to be nested inside the imported object instead of assuming its at the root level

### DIFF
--- a/packages/instrumentation-dataloader/src/instrumentation.ts
+++ b/packages/instrumentation-dataloader/src/instrumentation.ts
@@ -44,6 +44,13 @@ type PrimeFn = (typeof Dataloader.prototype)['prime'];
 type ClearFn = (typeof Dataloader.prototype)['clear'];
 type ClearAllFn = (typeof Dataloader.prototype)['clearAll'];
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function extractModuleExports(module: any) {
+  return module[Symbol.toStringTag] === 'Module'
+    ? module.default // ESM
+    : module; // CommonJS
+}
+
 export class DataloaderInstrumentation extends InstrumentationBase<DataloaderInstrumentationConfig> {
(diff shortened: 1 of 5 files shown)


packages/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -54,7 +61,8 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
       new InstrumentationNodeModuleDefinition(
         MODULE_NAME,
         ['>=2.0.0 <3'],
-        dataloader => {
+        dataloaderPkg => {
+          const dataloader = extractModuleExports(dataloaderPkg);
```
raphael-theriault-swi: Very minor nitpicking but this doesn't match the naming used most other places (`module`/`moduleExports`)
```suggestion
        module => {
          const dataloader = extractModuleExports(module);
```

packages/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -63,14 +71,15 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
 
           return this._getPatchedConstructor(dataloader);
         },
-        dataloader => {
+        dataloaderPkg => {
+          const dataloader = extractModuleExports(dataloaderPkg);
```
raphael-theriault-swi:
```suggestion
        module => {
          const dataloader = extractModuleExports(module);
```

raphael-theriault-swi (approved):
Thank you for picking this up and adding the test !

pichlermarc (approved):
looks good, thanks! :)

PR #2976, opened 2025-08-09
feat(instrumentation-amqp): adds latest semantic conventions

lucas-gregoire:

Currently, the `amqplib` instrumentation only exports legacy attributes. 
However, according to the OpenTelemetry specifications, we will have to migrate to new attributes to ensure better compatibility and consistency across tools and libraries. 
The v1.36.0+ conventions are not yet stable but will become stable in the future, so I think it's important to begin implementing the export of these new attributes, in order to anticipate the deprecation of the legacy ones, improve interoperability with other tools, and align with community best practices.

Adds support for stable semantic conventions in the amqplib instrumentation, controlled by the `OTEL_SEMCONV_STABILITY_OPT_IN=messaging` environment variable.

Key changes:

- Updates the instrumentation to use stable span names and attributes.
- Introduces new tests to validate stable semantic conventions.
- Updates README with migration guide and semantic conventions for both legacy and stable versions.

I used these specifications for implementing new attributes:
- [messaging spec v1.36](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/messaging/messaging-spans.md)
- [RabbitMQ messaging spec v1.36](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/messaging/rabbitmq.md)

## Migration guide

When upgrading to the new semantic conventions, it is recommended to follow this migration path:

1. **Upgrade** `@opentelemetry/instrumentation-amqplib` to the latest version
2. **Enable dual mode**: Set `OTEL_SEMCONV_STABILITY_OPT_IN=messaging/dup` to emit both old and new semantic conventions
3. **Update monitoring**: Modify alerts, dashboards, metrics, and other processes to use the new semantic conventions
4. **Switch to stable**: Set `OTEL_SEMCONV_STABILITY_OPT_IN=messaging` to emit only the new semantic conventions

### DIFF
--- a/packages/instrumentation-amqplib/src/utils.ts
+++ b/packages/instrumentation-amqplib/src/utils.ts
@@ -29,6 +29,7 @@ import {
   ATTR_SERVER_PORT,
 } from '@opentelemetry/semantic-conventions';
 import {
+  ATTR_MESSAGING_OPERATION,
   ATTR_MESSAGING_SYSTEM,
   ATTR_NET_PEER_NAME,
   ATTR_NET_PEER_PORT,
@@ -42,6 +43,7 @@ import {
   ATTR_MESSAGING_RABBITMQ_ROUTING_KEY,
   ATTR_MESSAGING_URL,
   MESSAGING_DESTINATION_KIND_VALUE_TOPIC,
+  MESSAGING_OPERATION_VALUE_PROCESS,
   OLD_ATTR_MESSAGING_MESSAGE_ID,
 } from '../src/semconv-obsolete';
 import type * as amqp from 'amqplib';
@@ -53,6 +55,8 @@ import {
   ATTR_MESSAGING_OPERATION_NAME,
   ATTR_MESSAGING_OPERATION_TYPE,
   ATTR_MESSAGING_RABBITMQ_DESTINATION_ROUTING_KEY,
+  ATTR_MESSAGING_RABBITMQ_MESSAGE_DELIVERY_TAG,
+  MESSAGING_OPERATION_TYPE_VALUE_PROCESS,
   MESSAGING_OPERATION_TYPE_VALUE_SEND,
 } from '@opentelemetry/semantic-conventions/incubating';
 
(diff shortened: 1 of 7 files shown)


pichlermarc:
@lucas-gregoire - looks like the messaging semantic conventions are not *actually* stable yet, so this PR is blocked until is is formally marked as such. 

(Unless this is part of an effort to create prototypes to mark the messaging semconv as stable, if that is the case, please link the corresponding issue from SemConv here, thanks 🙂)

packages/instrumentation-amqplib/src/utils.ts
```diff
@@ -20,16 +20,42 @@ import {
...
+  ATTR_MESSAGING_MESSAGE_CONVERSATION_ID,
+  ATTR_MESSAGING_MESSAGE_ID,
+  ATTR_MESSAGING_OPERATION_NAME,
+  ATTR_MESSAGING_OPERATION_TYPE,
+  ATTR_MESSAGING_RABBITMQ_DESTINATION_ROUTING_KEY,
+  ATTR_MESSAGING_RABBITMQ_MESSAGE_DELIVERY_TAG,
+  ATTR_MESSAGING_SYSTEM,
+  MESSAGING_OPERATION_TYPE_VALUE_PROCESS,
+  MESSAGING_OPERATION_TYPE_VALUE_SEND,
+} from '@opentelemetry/semantic-conventions/incubating';
```
trentm, line 61: Will need to move this to a local "src/semconv.ts" file rather than using the '/incubating' entry-point directly, per https://github.com/open-telemetry/opentelemetry-js/blob/main/semantic-conventions/README.md#unstable-semconv

You can typically used the https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/scripts/gen-semconv-ts.js tool to generate the src/semconv.ts file.

trentm (changes requested):
See comments above.

trentm:
Some context: We discussed this in the OTel JS SIG call today. Doing semconv changes to this instrumentation to get it closer to the current semconv spec state isn't *blocked* until messaging semconv is stabilized. However, because messaging semconv isn't stable yet, we cannot *call these changes "stable"* and we probably shouldn't use the `OTEL_SEMCONV_STABILITY_OPT_IN` envvar.

pichlermarc:
I opened https://github.com/open-telemetry/semantic-conventions/issues/2928 on the semconv repo to find a way how we could accept this change. With that proposal we would be able to use `messaging_latest_experimental` over `messaging` for updating to newer experimental semconv.

JamieDanielson:
Quick update: while we wait for next steps on experimental messaging conventions, I've gone ahead in a separate PR #3306  to update the `net*` semantic conventions. Unfortunately we will have some merge conflicts but I am happy to help with this when the time comes to pick it back up.

Issue #2921, opened 2025-06-29
@opentelemetry-instrumentation-aws-lambda: Support SQS context propagation


wpessers:
### Is your feature request related to a problem? Please describe

Currently the aws-lambda instrumentation does not support context propagation when using an sqs event source mapping. The aws lambda instrumentation offers the possibility of providing a custom context extractor. I've considered using this when initializing the aws-lambda instrumentation. However I believe it is not sufficient.

When a lambda function processes sqs messages, according to semantic conventions we should be linking all the sqs **process** spans back to their **producer** span: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans

SQS Messages can contain batches of messages, so there is no easy way of extracting 1 context to set as parent or link back to, unless your batch only contains 1 message. That's why I believe the custom context extractor can not be sufficient for this usecase.

### Describe the solution you'd like to see

In aws-lambda instrumentation, we should identify whether an input event contains sqs records. If it does we should do something similar to what the aws-sdk instrumentation is already doing, to link processing spans back to the producing spans by extracting context from the message system attributes.

### Describe alternatives you've considered

Initializing the aws lambda instrumentation with a custom context extractor when dealing with sqs input events (in opentelemetry-lambda nodejs layer: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/packages/layer/src/wrapper.ts#L309). However, as explained above I don't see this working as we possibly need to link multiple processing spans (for each message in a batch) back to their respective producer span.

Issue #707, opened 2021-10-19
Discussion(instrumentation-aws-sdk): SQS receive according to semantic conventions


blumamir:
This issue is to document the implementation of aws SQS receive operation according to the current semantic conventions for messaging systems (Oct 2021). there is an active SIG working on messaging systems specification which will probably change the specification and how to handle these situations when it's not possible to accurately extract perfect context.

### receiveMessage

- [Messaging Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#messaging-attributes) are added by this instrumentation according to the spec.
- Additional "processing spans" are created for each message received by the application.  
  If an application invoked `receiveMessage`, and received a 10 messages batch, a single `messaging.operation` = `receive` span will be created for the `receiveMessage` operation, and 10 `messaging.operation` = `process` spans will be created, one for each message.  
  Those processing spans are created by the library. This behavior is partially implemented, [See discussion below](#processing-spans).
- Sets the inter process context correctly, so that additional spans created through the process will be linked to parent spans correctly.  
  This behavior is partially implemented, [See discussion below](#processing-spans).
- Extract trace context from SQS MessageAttributes, and set span's `parent` and `links` correctly according to the spec.

#### Processing Spans

According to OpenTelemetry specification (and to reasonable expectation for trace structure), user of this instrumentation library would expect to see one span for the operation of receiving messages batch from SQS, and then, **for each message**, a span with it's own sub-tree for the processing of this specific message.

For example, if a `receiveMessages` returned 2 messages:

- `msg1` resulting in storing something to a DB.
- `msg2` resulting in calling an external HTTP endpoint.

This will result in a creating a DB span that would be the child of `msg1` process span, and an HTTP span that would be the child of `msg2` process span (in opposed to mixing all those operations under the single `receive` span, or start a new trace for each of them).

Unfortunately, this is not so easy to implement in JS:

1. The SDK is calling a single callback for the messages batch, and it's not straightforward to understand when each individual message processing starts and ends (and set the context correctly for cascading spans).
2. If async/await is used, context can be lost when returning data from async functions, for example:

```js
async function asyncRecv() {
  const data = await sqs.receiveMessage(recvParams).promise();
  // context of receiveMessage is set here
  return data;
}

async function poll() {
  const result = await asyncRecv();
  // context is lost when asyncRecv returns. following spans are created with root context.
  await Promise.all(
[...]

PR #2981, opened 2025-08-12
feat(opentelemetry-instrumentation-aws-lambda): Add sqs context propagation

wpessers:

#2921 No SQS Context propagation in aws lambda.

Used pubsub propagation utils to automatically created processing spans for each sqs record within a lambda sqs event. These spans link to their producer span, as is described in the spec: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans


wpessers:
Converted this back to draft PR. Some stuff I need to figure out first. Used the sqs instrumentation from the aws sdk instrumentation package as an example but there are a bunch of deprecated semconv attributes.

wpessers:
@jj22ee Any thoughts on this PR? There's some use of deprecated semconv attributes for the messaging spans. But these only have alternatives that are still incubating currently. So I believe we would prefer the deprecated ones to avoid possibly breaking on minor version updates for these semconv attributes? Also, there's currently still a bunch of separate commits not adhering to the conventional commits standard. Is it expected that I rebase and squash the commits myself? Or is this done eventually on merge?

wpessers:
Also, I tested this by deploying this version of the aws-lambda instrumentation and actually exporting the spans to grafana. Resulting in traces that look like this:
This is the trace of the lambda that is triggered by sqs, thus containing the regular function invocation span as parent. And then a bunch of process spans for each sqs record in this batch. These each have a span link back to the message producer span as you can see.
Screenshot 2025-08-16 223919

If we follow such a span link, we end up at the message producer span:
Screenshot 2025-08-16 223950

gkaskonas:
Can this be merged, please?

jj22ee:
Hey, planning to take a look this week.

wpessers:
Hi @jj22ee we've discussed this in the FaaS SIG again. The span linking visualization that I showed above for grafana is not something that is generally supported by other observability platforms. Therefore we want to change this PR a bit so that we can make it configurable to either use the current implementation with span linkgs or set the processing spans' parent context to the producer span instead of span linking. Users can then configure themselves how they want the context propagation to work.

Just so people using different observability platforms can also get the benefits of this feature, and the second approach will make this information visible regardless of which platform you export spans to.
Any thoughts?

jj22ee:
Hi @wpessers

> There's some use of deprecated semconv attributes for the messaging spans. But these only have alternatives that are still incubating currently. So I believe we would prefer the deprecated ones to avoid possibly breaking on minor version updates for these semconv attributes?

The workaround that is being done in the contrib repo is to just copy the incubating attributes into the package's semconv.ts file. For example, see this incubating attribute for aws-lambda-instrumentation: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-aws-lambda/src/semconv.ts

jj22ee:
Regarding the original issue: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2921#issue-3186204604
>In aws-lambda instrumentation, we should identify whether an input event contains sqs records. If it does we should do something similar to what the aws-sdk instrumentation is already doing, to link processing spans back to the producing spans by extracting context from the message system attributes.

It was discussed to drop Processing Spans in OTel JS's AWS SDK instrumentation due to implementation limitations and updates to the spec
- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1477#issuecomment-1836903586
  - This related PR was merged to AWS SDK Instrumentation recently, that drops processing spans
    - https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2345

However, there is a separate spec specifically for handling SQS in AWS Lambda, which is [different from the spec you linked](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans) in the description:
- https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs

This one does state that:
>The function invocation span MUST correspond to the SQS event, which is the batch of messages. For each message, an additional span SHOULD be created to correspond with the handling of the SQS message

So this PR does follow the Lambda+SQS Spec to create new process spans for each message.

jj22ee:
I advise you to look at the [Lambda+SQS spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs) because the spec details in there differs from this PR (e.g. required usage of [AWS X-Ray Propagator](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) and using [message system attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-message-system-attributes) instead of message attributes to check for `AWSTraceHeader` to generate the context for span link).

> Therefore we want to change this PR a bit so that we can make it configurable to either use the current implementation with span linkgs or set the processing spans' parent context to the producer span instead of span linking. Users can then configure themselves how they want the context propagation to work.

I'd recommend to just start with the initial support for process spans for each message w/ span links [since that is part of spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs). Do you know if the alternative configuration will also be added to the spec? Sounds like it might be complicated because it still make sense for processing spans' parent to be the Lambda Function Span.

wpessers:
> It was discussed to drop Processing Spans in OTel JS's AWS SDK instrumentation due to implementation limitations and updates to the spec
> 
> * [Tests failing for `@aws-sdk/client-sqs >=3.316`  #1477 (comment)](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1477#issuecomment-1836903586)
>   
>   * This related PR was merged to AWS SDK Instrumentation recently, that drops processing spans
>     
>     * [feat(aws-sdk)!: SQS receive: use span links instead of processing spans #2345](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2345)
>
Oh, I see. This was apparently ongoing right before I opened this PR, but I somehow missed it.

wpessers:
> I advise you to look at the [Lambda+SQS spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs) because the spec details in there differs from this PR (e.g. required usage of [AWS X-Ray Propagator](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) and using [message system attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-message-system-attributes) instead of message attributes to check for `AWSTraceHeader` to generate the context for span link).
> 
That makes sense, will read through it and make the necessary changes.

> I'd recommend to just start with the initial support for process spans for each message w/ span links [since that is part of spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs). Do you know if the alternative configuration will also be added to the spec? Sounds like it might be complicated because it still make sense for processing spans' parent to be the Lambda Function Span.

Sounds good, I think we can indeed just start with the implementation adhering to the spec. I don't know if the alternative configuration will be added to the spec but it's probably worth discussing. Agree, it does indeed make sense for the parent to be the Lambda Function span. It's just a shame that this will not be represented properly in a lot of observability platforms. If we can (after the initial implementation) also add an experimental feature to allow end-users to opt-in to the alternative way of creating the processing spans with the producer span as their parent, then we can properly evaluate whether a spec change would be warranted.

jj22ee:
Commenting again about the processing spans created for each message in the SQS batch.
- AWS SDK instrumentation dropped process spans because of 2 reasons (context from [this comment](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1477#issuecomment-1836903586) & [this issue](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/707)):
  1. limited ability to create processing spans automatically (e.g. requires a specific for-loop setup to create these spans)
  2. [New Batch receiving spec](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#batch-receiving) doesn't mention creating processing spans, but instead mentions to add all producer span links to the singular consumer span that receives all the producer messages.

Reason (i) still applies in Lambda case. However this isn't the exact case for reason (ii), as the [Lambda Spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs) still mentions `"function invocation span MUST correspond to the SQS event, which is the batch of messages. For each message, an additional span SHOULD be created to correspond with the handling of the SQS message"` while still acknowledging the limitations: `"automatic instrumentation mechanisms without code change will often not be able to instrument the processing of the individual messages"`.

So I'm wondering if we should still avoid creating these processing spans (deviate from Lambda spec), and instead just add a link from all producer spans to the singular Lambda Function Span, similar to AWS SDK instr today. @trentm do you think we should still drop process spans here, despite the Lambda spec?

wpessers:
It's clear to me what changed in the normal messaging spec and why the changes were made. I'm merely questioning if the decision to go all-out on the span links made sense from an end-user perspective, having noticed the little support there is on visualizing this information in observability platforms. But I suppose it is up to them to implement the spec, not the other way around. I also agree that maybe it makes more sense to align the lambda spec with the general messaging spec.

wpessers:
@jj22ee regarding your comment about using message system attributes and xray exporter. This is again something that is currently in the spec but I'm not sure about the feasibility. I posed a question to semconv maintainers regarding this as well. But all aws sdk instrumentation libs accross different platforms that I know of, currently have implemented context propagation through message attributes. So if we want to use message  system attributes in lambda instrumentation, that requires changes in aws sdk instrumentation.

pichlermarc:
Looks like this PR has stalled due to inconsistencies in the https://github.com/open-telemetry/semantic-conventions/ repo. @wpessers @jj22ee would you mind opening an issue there to seek guidance on how to resolve the situation here?

trentm:
FWIW, here is the PR that *removed* processing spans and usage of `@opentelemetry/propagation-utils` from the SQS instrumentation in the `instrumentation-aws-sdk` package: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2345

wpessers:
Yes, I know what to do next based on feedback received earlier. Just needed some time to get this going again. Other issues have taken priority

wpessers:
There have been some chats in the semconv slack already, before I continue work on this PR I opened an issue to formalize the conversation and see what will be the right way forward: https://github.com/open-telemetry/semantic-conventions/issues/3217

wpessers:
Okay so stuff is taking a while regarding the spec alignment / change. I'll still keep following up on that since as I stated earlier there seemed to be willingness to properly align. However for now I've added a spec-compliant default behaviour that will use the AWSTraceHeader from `system attributes` for context propagation. The change I introduced earlier, which uses the configured propagator to extract from `message attributes` is now behind an opt-in feature flag. 

I did however keep the span structure / hierarchy and span linking in line with what was described here:

> FWIW, here is the PR that removed processing spans and usage of @opentelemetry/propagation-utils from the SQS instrumentation in the instrumentation-aws-sdk package: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2345

It really doesn't make any sense to keep using the old structure with separate processing spans in my opinion. If making that official in the spec (the alignment of lambda spec with general spec I was talking about) is a prerequisite for getting this PR merged I will go and push harder to get it looked at. But I want to reiterate that this PR would solve a real pain that users have had for quite a long time, and getting this merged would be a huge step forward.

LMK what you think @open-telemetry/javascript-approvers and also particularly @jj22ee

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -92,11 +124,15 @@ function isSupportingCallbacks(): boolean {
...
 export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstrumentationConfig> {
+  private _xrayPropagator: AWSXRayPropagator | undefined;
   declare private _traceForceFlusher?: () => Promise<void>;
   declare private _metricForceFlusher?: () => Promise<void>;
 
   constructor(config: AwsLambdaInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
+    if (!config.useConfiguredPropagatorForSqsExtraction) {
+      this._xrayPropagator = new AWSXRayPropagator();
+    }
```
pichlermarc: Hmm, I'm a bit confused by the need for this. 

**question:** Wouldn't a globally registered propagator properly handle this? 🤔

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +512,154 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+      links,
+    });
+  }
+
+  private static _extractSqsQueueUrl(queueArn: string): string {
+    const queueArnParts = queueArn.split(':');
+    const region = queueArnParts[3];
+    const accountId = queueArnParts[4];
+    const queueName = queueArnParts[5];
+    return `https://sqs.${region}.amazonaws.com/${accountId}/${queueName}`;
```
pichlermarc: I suppose it should never happen, but queueArn could be undefined and fall back to `unknown`?
In that case we'd end up with `https://sqs.undefined.amazonaws.com/undefined/undefined`, right?;

**suggestion:** let's add some defensive code, just in case. 🙂

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -562,6 +761,17 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
     span.end();
+  }
+
+  private _endInvocationSpanAndFlush(
+    invocationSpan: Span,
+    err: string | Error | null | undefined,
+    callback: () => void,
+    additionalSpans?: Span[]
+  ): void {
+    additionalSpans?.forEach(s => this._endSpan(s, err ?? null));
```
pichlermarc: **nit:** this is unlikely to be an issue as it's only called once per lambda invocation but we should still avoid `.forEach()` for performance reasons. might also apply to other parts of this PR.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +512,154 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+    return 'unknown';
+  }
+
+  private _startSqsProcessSpan(event: SQSEvent): Span {
+    const messages = event.Records;
+    const queueArn = messages[0]?.eventSourceARN;
+    const queueName = queueArn?.split(':').pop() ?? 'unknown';
+    const links = this._extractSqsSpanLinks(messages);
+
+    return this.tracer.startSpan(`${queueName} process`, {
```
pichlermarc: **suggestion:** let's align this name directly with the messaging semconv, which states:

>  The span name SHOULD be {messaging.operation.name} {destination} (see below for the exact definition of the [{destination}](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/messaging-spans.md#destination-placeholder) placeholder).


```suggestion
    return this.tracer.startSpan(`process ${queueName}`, {
```

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +512,154 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+      kind: SpanKind.CONSUMER,
+      attributes: {
+        [ATTR_FAAS_TRIGGER]: FAAS_TRIGGER_VALUE_PUBSUB,
+        [ATTR_MESSAGING_OPERATION_TYPE]: MESSAGING_OPERATION_TYPE_VALUE_PROCESS,
+        [ATTR_MESSAGING_SYSTEM]: MESSAGING_SYSTEM_VALUE_AWS_SQS,
+        [ATTR_MESSAGING_DESTINATION_NAME]: queueName,
+        [ATTR_MESSAGING_BATCH_MESSAGE_COUNT]: event.Records.length,
```
pichlermarc: not sure if this applies here: [IIUC should only be set if the system does not support single message batches:](https://github.com/open-telemetry/semantic-conventions/blob/dd90cbeee3d5182907257fe2e935a07601e22829/docs/registry/attributes/messaging.md?plain=1#L40)

> Instrumentations SHOULD NOT set `messaging.batch.message_count` on spans that operate with a single message. When a messaging client library supports both batch and single-message API for the same operation, instrumentations SHOULD use `messaging.batch.message_count` for batching APIs and SHOULD NOT use it for single-message APIs.

**suggestion (if applicable):** let's align with spec and only set it conditionally.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +512,154 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+        [ATTR_URL_FULL]: AwsLambdaInstrumentation._extractSqsQueueUrl(queueArn),
+      },
```
pichlermarc, line 594: **question:** should we use `aws.sqs.queue.url` instead?

https://github.com/open-telemetry/semantic-conventions/blob/dd90cbeee3d5182907257fe2e935a07601e22829/docs/messaging/sqs.md?plain=1#L29

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +512,154 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+      links,
+    });
+  }
```
pichlermarc: [`messaging.operation.name`](https://github.com/open-telemetry/semantic-conventions/blob/dd90cbeee3d5182907257fe2e935a07601e22829/docs/messaging/sqs.md?plain=1#L23) seems to be missing.

**suggestion:** let's add it since it's marked as "REQUIRED"

pichlermarc: [`server.address`](https://github.com/open-telemetry/semantic-conventions/blob/dd90cbeee3d5182907257fe2e935a07601e22829/docs/messaging/sqs.md?plain=1#L27) seems to be conditionally required if it exists, but seems to be missing here.

**question:** is there any way we could get this? 🙂

pichlermarc (commented):
Thanks for working on this 🙌 

My (still relatively narrow) understanding of this issue is that we'd want to align as far as possible with the newer SQS spec, right? (I based my review off of that)

> It really doesn't make any sense to keep using the old structure with separate processing spans in my opinion. If making that official in the spec (the alignment of lambda spec with general spec I was talking about) is a prerequisite for getting this PR merged I will go and push harder to get it looked at. But I want to reiterate that this PR would solve a real pain that users have had for quite a long time, and getting this merged would be a huge step forward.

Yep, I agree. I think we should avoid the old structure wherever possible.

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: I did this because of the earlier feedback I got here: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2981#issuecomment-3251254095

> I advise you to look at the [Lambda+SQS spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs) because the spec details in there differs from this PR (e.g. required usage of [AWS X-Ray Propagator](https://opentelemetry.io/docs/specs/otel/context/api-propagators/) and using [message system attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-message-system-attributes) instead of message attributes to check for AWSTraceHeader to generate the context for span link).

I don't like it but apparently it's required for us to be spec-compliant. The `useConfiguredPropagatorForSqsExtraction` config option I added specifically to give users flexibility to deviate, since it doesn't make sense to limit propagation to only be possible through `AWSTraceHeader` with the xray propagator.

I would think If we want to rely on the globally registered propagator that would mean the end-user is required to configure their propagators to include the xray propagator no?

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: Good idea!

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: Good shout, will double-check the changes and replace.

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: Checking!

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: Nice find, seems like I cut a corner here. In case the lambda receives just 1 record it should not have this attribute indeed.

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: I got my inspiration for this one from the aws-sdk package. But looks like it makes more sense to change this to `aws.sqs.queue.url` yes.

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: I've updated the code to set it to `sqs.<region>.amazonaws.com`, to me that makes sense here given what the span signifies. Let me know what you think!

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: I updated this code to make it more defensive, the sqs queue url is now set as the value of the `aws.sqs.queue.url` attribute as that was also a [requested change](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2981#discussion_r3123490096). Thus some stuff moved around and since `aws.sqs.queue.url` is only a "recommended" attribute according to the spec I've opted to conditionally set the queue url only if it's possible to reconstruct it, instead of falling back to "unknown". Lmk what you think!

packages/instrumentation-aws-lambda/README.md
```diff
@@ -168,17 +169,44 @@ Alternatively, use the `auto-configuration-package` as in example #1 and set the
...
+Additional attributes collected on the SQS process (CONSUMER) span:
+
+| Attribute                        | Short Description                                                  |
+|----------------------------------|--------------------------------------------------------------------|
+| `faas.trigger`                   | Type of the trigger (`pubsub`).                                    |
+| `messaging.operation.type`       | The type of the messaging operation (`process`).                   |
+| `messaging.system`               | The messaging system (`aws_sqs`).                                  |
+| `messaging.destination.name`     | The queue name extracted from the event source ARN.                |
+| `messaging.batch.message_count`  | The number of messages in the batch.                               |
+| `url.full`                       | The reconstructed SQS queue URL.                                   |
```
jj22ee:
```suggestion
| `aws.sqs.queue.url`              | The reconstructed SQS queue URL.                                   |
```

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +517,172 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+
+  private static _extractLinksFromMessageAttributes(
+    messages: SQSRecord[]
+  ): Link[] {
+    const propagationFields = propagation.fields();
+    const links: Link[] = [];
+    for (const message of messages) {
+      const hasPropagationFields = Object.keys(
+        message.messageAttributes || []
```
```suggestion
        message.messageAttributes || {}
```

packages/instrumentation-aws-lambda/test/lambda-test/async.js
```diff
@@ -19,6 +19,10 @@ exports.handler = async function (event, context) {
   return 'ok';
 };
 
+exports.sqshandler = async function (event, context) {
```
jj22ee: This seems to be unused and can be removed?

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +517,172 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+      messages[0]?.eventSourceARN
+    );
+    const queueName = queueArnParts?.queueName ?? 'unknown';
+    const links = this._extractSqsSpanLinks(messages);
+
+    return this.tracer.startSpan(`process ${queueName}`, {
+      kind: SpanKind.CONSUMER,
+      attributes: {
+        [ATTR_FAAS_TRIGGER]: FAAS_TRIGGER_VALUE_PUBSUB,
+        [ATTR_MESSAGING_OPERATION_NAME]: 'ack',
```
jj22ee: This should be `process` instead of `ack` (https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs)

I think `ack; nack; send` from [docs/messaging/sqs.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/sqs.md) were just examples.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
```diff
@@ -1220,4 +1221,195 @@ describe('lambda handler', () => {
       });
     });
   });
+
+  describe('sqs context propagation (default: AWSTraceHeader)', () => {
+    it('creates process span with span link (sync handler)', async () => {
```
jj22ee, line 1230: You could add assertions for the new attributes you are adding here.
```
faas.trigger
messaging.*
aws.sqs.queue.url
server.address
```

packages/instrumentation-aws-lambda/test/lambda-test/async.js
wpessers: Good catch, missed this indeed

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: Oh I assumed they were the only valid values. But indeed the spec says "example values" so seems like we could set it to process. Spec further down in the example for [sqs-lambda-tracing-passive](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs-lambda-tracing-passive) also shows "process" as a value for `messaging.operation.name`. Will update!

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
wpessers: Done!

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -441,31 +517,172 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+  }
+
+  private _extractLinksFromAWSTraceHeader(messages: SQSRecord[]): Link[] {
+    const links: Link[] = [];
+    for (const message of messages) {
+      const awsTraceHeader = message.attributes?.AWSTraceHeader;
+      if (!awsTraceHeader) continue;
+
+      const carrier = { [AWSXRAY_TRACE_ID_HEADER]: awsTraceHeader };
+      const propagatedContext = this._xrayPropagator!.extract(
```
jj22ee: I'd prefer to avoid the non-null assertion here. Can we always set `this._xrayPropagator = new AWSXRayPropagator();` since [`AWSXRayPropagator`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/propagator-aws-xray/src/AWSXRayPropagator.ts) is lightweight?

packages/instrumentation-aws-lambda/README.md
```diff
@@ -168,17 +169,44 @@ Alternatively, use the `auto-configuration-package` as in example #1 and set the
...
+Additional attributes collected on the SQS process (CONSUMER) span:
+
+| Attribute                        | Short Description                                                  |
+|----------------------------------|--------------------------------------------------------------------|
+| `faas.trigger`                   | Type of the trigger (`pubsub`).                                    |
+| `messaging.operation.type`       | The type of the messaging operation (`process`).                   |
+| `messaging.system`               | The messaging system (`aws_sqs`).                                  |
+| `messaging.destination.name`     | The queue name extracted from the event source ARN.                |
+| `messaging.batch.message_count`  | The number of messages in the batch.                               |
+| `aws.sqs.queue.url`              | The reconstructed SQS queue URL.                                   |
```
jj22ee, line 230: Also add `server.address` and `messaging.operation.name`.

jj22ee (commented):
2 remaining nits from me, otherwise LGTM!

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: I should've checked the actual code of the propagator earlier, agree that's essentially free to instantiate it by default.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -551,6 +770,21 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+    callback: () => void,
+    additionalSpans?: Span[]
+  ): void {
+    if (additionalSpans) {
+      for (const s of additionalSpans) {
+        this._endSpan(s, err ?? null);
+      }
+    }
+
+    this._endSpan(invocationSpan, err ?? null);
```
```suggestion
        this._endSpan(s, err);
      }
    }

    this._endSpan(invocationSpan, err);
```

packages/instrumentation-aws-lambda/README.md
```diff
@@ -188,17 +189,46 @@ Alternatively, use the `auto-configuration-package` as in example #1 and set the
...
+
+When your Lambda function is triggered by SQS, the instrumentation creates a CONSUMER span with span links to the producer trace contexts extracted from the SQS messages.
+
+By default, context is extracted from the `AWSTraceHeader` system attribute using the AWS X-Ray propagator, as defined by the [OpenTelemetry semantic conventions for AWS Lambda](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs-event). This works when AWS X-Ray active tracing is enabled on the SQS queue.
+
+If your SQS producers inject trace context into message attributes using the globally configured propagator (as `@opentelemetry/instrumentation-aws-sdk` does by default with W3C traceparent), set `useConfiguredPropagatorForSqsExtraction: true`:
+
+```js
+new AwsLambdaInstrumentation({
+    useConfiguredPropagatorForSqsExtraction: true,
```
jj22ee: nit: maybe `useGlobalPropagatorForSqsExtraction` ?

jj22ee (approved):
Added 2 more comments for your input.  
Also please update `Short description of the changes` in PR description.

Apart from the above, approving this PR. Thanks for your contribution!

packages/instrumentation-aws-lambda/src/instrumentation.ts
wpessers: good catch!

wpessers:
The remaining ci failure is unrelated to this PR, I've opened up #3510 to tackle that though.

packages/instrumentation-aws-lambda/src/instrumentation.ts
pichlermarc: > I would think If we want to rely on the globally registered propagator that would mean the end-user is required to configure their propagators to include the xray propagator no?

Hmm yes. I'm wondering who this spec was initially written for. I think it was SDK authors, but not instrumentation authors.

[This spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#xray-lambda-propagator-functionality) looks like guidance for SDK authors to include a lambda-specific propagator if they decide to provide a aws-lambda instrumentation.

[This spec](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#xray-lambda-propagator-configuration) looks to me like it's additional guidance for [`@opentelemetry/auto-configuration-propagator`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/auto-configuration-propagators/src/utils.ts#L40-L80) to use the `xray-lambda` propagator if both are configured.

IIUC, if that functionality is available, then there's no need for AWS lambda instrumentation to create their own `xray-lambda` propagator. The SDK should simply offer the possibility to configure it, it's then up to the user to set `OTEL_PROPAGATORS=xray-lambda` if they wish to use it.

pichlermarc: I'll go check some other language's lambda instrumentations - maybe I can find some hints. 🕵️

pichlermarc: Looking at the Java implementation, it looks like they document that it's always enabled, but that does not seem to be true to me. [Here, they say](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/178530b706bb25ac988b07cd89b4a8bfab68284d/instrumentation/aws-lambda/aws-lambda-events-3.11/library):

> This instrumentation supports propagating traces using the X-Amzn-Trace-Id format for both normal requests and SQS requests. X-Ray propagation is always enabled, there is no need to configure it explicitly.

The way they do it seems to be by including the X-Ray propagator in the classpath and then auto-loading it when the instrumentation is active. I suppose people could select or deselect it via some mechnism like env var config. But the way context extraction works seems to be via the globally configured propagator.

I also checked python and was not able to find any such mechanism where the propagator is used directly.

In our case, we don't have the auto loading mechanism that Java has, but IMO it is okay if folks need to configure it manually. WDYT @wpessers?

wpessers: > Hmm yes. I'm wondering who this spec was initially written for. I think it was SDK authors, but not instrumentation authors.

This would actually make a lot of sense, hadn't thought about that 💡 

> I also checked python and was not able to find any such mechanism where the propagator is used directly.

Last time I checked I don't think python already supported sqs -> lambda context propagation.

I'm not sure I follow regarding the Java instrumentation, I think they always use the `AWSXrayPropagator` regardless of what the user configures: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda/aws-lambda-events-common-2.2/library/src/main/java/io/opentelemetry/instrumentation/awslambdaevents/common/v2_2/internal/SqsMessageSpanLinksExtractor.java#L34-L38

wpessers: @pichlermarc I think this is now the only remaining open point to get this PR merged, since @jj22ee approved the pr 2 weeks ago. Please let me know if the above is not making sense or if you disagree with it, I'm happy to make further changes if needed!

pichlermarc (approved):

packages/instrumentation-aws-lambda/src/instrumentation.ts
pichlermarc: Yep, let's keep it as is.
If the spec changes in the future, we can still make adjustments, it should not block this PR.

pichlermarc:
one more merge from `main` then this should be good to merge. 
thank you @wpessers for working on this.

wpessers:
@pichlermarc missed this, taking care of it later today!

wpessers:
Done @pichlermarc

PR #2983, opened 2025-08-14
fix(instrumentation-pino): Propagate all args to `mixin`

72636c:

We noticed that our user-defined `mixin` was breaking because it could not access the third `logger` argument. Excerpt from https://getpino.io/#/docs/api?id=mixin-function:

```typescript
const logger = pino({
  mixin (obj, num, logger) {
    return {
      tags: logger.tags
    }
  }
})
```

[Update(by trentm): the 3rd arg to a mixin was added in pino@8.14.0.]

Try to propagate `mixin` args. 

Unfortunately I couldn't get this running on my local; `npm run compile` just hangs on my machine. So this is just some quick guesswork, hopefully it's enough to understand the problem though!

```
 NX   The Nx Daemon is unsupported in WebAssembly environments. Some things may be slower than or not function as expected.

memory allocation of 70 bytes failed
```


packages/instrumentation-pino/src/instrumentation.ts
```diff
@@ -67,10 +67,14 @@ export class PinoInstrumentation extends InstrumentationBase<PinoInstrumentation
...
-            logger[mixinSym] = (ctx: object, level: number) => {
+            logger[mixinSym] = (
+              ctx: object,
+              level: number,
+              ...rest: unknown[]
+            ) => {
               return Object.assign(
                 otelMixin(ctx, level),
-                origMixin(ctx, level)
+                origMixin(ctx, level, ...rest)
```
72636c, line 77: We could also explore an approach similar to https://github.com/DataDog/dd-trace-js/blob/213d56fb37bc7b929ee9cea99d21bb00eeafc412/packages/datadog-instrumentations/src/pino.js#L37-L51

trentm:
> The Nx Daemon is unsupported in WebAssembly environments.

Interesting. What is your environment? I'm just curious.  Are you running some "in-browser dev environment"?

trentm:
Test failures. I'm guessing they are due to the test being dependent on a particular newer version of pino.

The feature where the logger is passed into the mixin as the 3rd arg came in https://github.com/pinojs/pino/pull/1709 which was included in v8.14.0.

```
* d3a4560 - (tag: v8.14.0) Bumped v8.13.0 (2 years, 3 months ago) <Matteo Collina>
* b473f36 - docs: add 'best performance for basic stdout logging' to help.md (#1584) (2 years, 3 months ago) <Janek Lasocki-Biczysko>
* 962bc2b - feat: pass logger/child logger as param to mixin (#1709) (2 years, 3 months ago) <mary marchini>
```

and the current installed pino is `pino@8.3.1`:

```
% npm ls pino
opentelemetry-contrib@0.1.0 /Users/trentm/tm/opentelemetry-js-contrib17
└─┬ @opentelemetry/instrumentation-pino@0.50.0 -> ./packages/instrumentation-pino
  └── pino@8.3.1
```

Still, we'll need to write the test to work with older versions for the "test-all-versions" tests.

trentm (approved):
This is great. Thanks for the fix.

72636c:
Thanks for getting the tests going @trentm! Turns out this was PEBKAC; I recently set up a new laptop and mistakenly had Node.js version defaulted to v24 in the absence of a `.node-version` or `.nvmrc`, which resulted in a borked `npm ci`. LTS works wonders 🙂

PR #2994, opened 2025-08-21
test: trying out nx tav parallel patch

Ugzuzg:

-

-


test-all-versions.js
Ugzuzg, line 1: @david-luna, apart from some flaky mongodb tests which fail under stress (and some undici tests that I've fixed already), this tav patch seems to be working and offers a 2x speedup. I'll check with tav maintainers what they think of this approach, and if it's possible to upstream.

david-luna: @Ugzuzg 

thanks! I think for now we can rely on `npm run --workspaces` and revisit this if upstreamed.

Ugzuzg: The mongodb test failures were caused by the missing callback. Now, everything seems to be working!

david-luna: Nice! is the assignment of `process.env.npm_config_legacy_peer_deps` the only difference from upstream?

Ugzuzg: No, it also isolates the installs from each other by creating synthetic packages outside the scope of the npm monorepo and then linking the installed packages into the `node_modules` of the specific npm "workspace".

https://github.com/watson/test-all-versions/pull/26

PR #3001, opened 2025-08-28
chore(eslint): upgrade to eslint 9 flat config

overbalance:

Upgrade to eslint 9 because 8 is EOL.

- Remove all .eslintignore files in packages
- Remove all `lint` and `lint:fix` commands in packages
- Add lots of rules to warn on invalid ts and js called out by the new eslint 9 default recommendations
- Replace the license plugin
- Replace the node plugin

## Still TODO for this PR

- [ ] Refactor some code that is expecting an expression but instead is a ternary
- [ ] Determine how to handle vendors that have manually added their names to the copyright block of the license in their package


trentm:
> * Determine how to handle contributors that have manually added company names to the copyright block of the license

Let me know if you need help with this one. I've spent some time on the eslint plugin we are using for the license headerr checking.  (https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2967, with hopes to change to a different plugin once/if this PR gets accepted: https://github.com/nikku/eslint-plugin-license-header/pull/26)

overbalance:
@trentm That would be great. I see some companies have added a credit after the full license and some have added their names to the copyright alongside otel at the top of the license. I'm in favor of supporting a single way to do this.
Hopefully your PR is accepted soon so we can apply a regex.

overbalance:
FYI, I'm going to keep rebasing this until someone joins to write code.

eslint.config.mjs
```diff
@@ -0,0 +1,187 @@
...
+      node: {
+        tryExtensions: ['.js', '.json', '.node', '.ts', '.tsx'],
+      },
+    },
+    rules: {
+      // old rules
+      quotes: ['error', 'single', { avoidEscape: true }],
+      eqeqeq: ['error', 'smart'],
+      'prefer-rest-params': 'off',
+      'license-header/header': ['error', license],
```
trentm: Can we not use the existing `header/header` plugin? Or does it perhaps not support the newer eslint config ("flat config" or some name I vaguely recall hearing about)?

eslint.config.mjs
```diff
@@ -0,0 +1,187 @@
...
+  {
+    ignores: [
+      '**/build/**',
+      '**/coverage/**',
+      '**/dist/**',
+      '**/node_modules/**',
+    ],
+  },
+  {
+    files: ['**/*.{js,ts,mjs}', '*.{js,ts,mjs}'],
```
trentm: nit: The `'**/*.{js,ts,mjs}'` entry doesn't capture top-level files?

package.json
```diff
@@ -28,7 +28,7 @@
     "test-all-versions:with-services-env": "cross-env NODE_OPTIONS='-r dotenv/config' DOTENV_CONFIG_PATH=./test/test-services.env npm run test-all-versions",
     "changelog": "lerna-changelog",
     "lint": "nx run-many -t lint && npm run lint:deps && npm run lint:readme && npm run lint:markdown && npm run lint:semconv-deps && npm run lint:examples",
```
trentm: I'm missing something: If the "lint" script was removed from all package.json files (in the workspace dirs), then what is `nx run-many -t lint` actually running?

eslint.config.mjs
overbalance: That's right. It does not have version 9 support.

package.json
trentm: Ah. `nx run[-many]` can have more tasks than what is in package.json#scripts. In this case the `lint` task is from this nx plugin (in nx.json):

```json
  "plugins": [
    {
      "plugin": "@nx/eslint/plugin",
      "options": {
        "targetName": "lint"
      }
    }
```

overbalance: Yeah, sorry missed this comment. It runs whatever you put here in the package context. Which is why the root command has --fix at the end. It's passing the option to the plugin.

overbalance:
> Let me know if you need help with this one. I've spent some time on the eslint plugin we are using for the license headerr checking.  (https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2967, with hopes to change to a different plugin once/if this PR gets accepted: https://github.com/nikku/eslint-plugin-license-header/pull/26)

What do you think about having contributors append their credits after the license in a new comment? We could always change it later but it would unblock this iteration.

trentm:
> What do you think about having contributors append their credits after the license in a new comment? We could always change it later but it would unblock this iteration.

I haven't published it yet (will do tomorrow), but I hope we can use https://github.com/trentm/eslint-plugin-yet-another-license-header -- I took the chance to learn a bit more about eslint plugins in doing this. Mostly it matches what the eslint-plugin-license-header with my unmerge PR would do.

trentm:
@overbalance I've published https://www.npmjs.com/package/eslint-plugin-yet-another-license-header

Here is a commit that changes this branch to use it: https://github.com/trentm/opentelemetry-js-contrib/commit/e8004c0b2b6cc3ddef348c9a9fbcbef14596a69a

The [`licensePattern`](https://github.com/trentm/opentelemetry-js-contrib/commit/e8004c0b2b6cc3ddef348c9a9fbcbef14596a69a#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R41-R48) I've defined there put *some* minimal limits on where the additional `Copyright ...` assignments live in the header block. This required manually changing some of the resource-detector-gcp and instrumentation-oracledb headers to match.

Note that this commit builds on a couple others that I had made while playing around, so **it probably won't apply cleanly to your branch right now**:

```
* e8004c0b - (HEAD -> overbalance/repair-eslint, origin/overbalance/repair-eslint) lint: Use a new yet-another-license-header eslint-plugin to handle license headers (4 minutes ago) <Trent Mick>
* fb3ca3a8 - fix lint error for empty catch block (3 hours ago) <Trent Mick>
* 9dd2d5a4 - lint:fix, with some manual fixes (including whitespace cleanup from the eslint rule removing unneeded 'eslint-disable' comments) (28 hours ago) <Trent Mick>
* 194909d1 - temp disable license-header rule for now to work on other parts (28 hours ago) <Trent Mick>
* c6f80535 - skipLibCheck for aws, openai (2 days ago) <overbalance>
* 42f099f0 - lint dns test (2 days ago) <overbalance>
* d732db80 - this is jsx (2 days ago) <overbalance>
* 9ca1dbfd - upgrade to eslint 9, upgrade prettier and nx (2 days ago) <overbalance>
```

Would you like me to attempt to push my change to your branch?

trentm:
A possible alternative is taht I could do a PR that *separately* changes to this new license-header plugin (basically a PR close to this older draft one of mine: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2969) and then you could merge/rebase yours on top of that. I'm easy either way.

overbalance:
> A possible alternative is taht I could do a PR that _separately_ changes to this new license-header plugin (basically a PR close to this older draft one of mine: #2969) and then you could merge/rebase yours on top of that. I'm easy either way.

Keep it simple and throw it on this branch. I noted it above but I have not run lint:fix, I thought that would be better as another branch off this one to keep the review **possible**.

prettier.config.js
```diff
@@ -1,4 +1,7 @@
+/* eslint-disable license-header/header */
 module.exports = {
+  // the repo is half with parens and half without
+  // which do we want?
   arrowParens: 'avoid',
```
overbalance, line 18: The repo is inconsistent regarding arrowParens. I prefer them but I will leave it to the group to decide. Neither choice is necessarily fewer changes.

trentm:
> Keep it simple and throw it on this branch.

@overbalance  You are on a corporate fork so I cannot push to your branch:

```
% git push embrace-io overbalance/repair-eslint
ERROR: Permission to embrace-io/opentelemetry-js-contrib.git denied to trentm.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

So, instead, I've pushed to another branch on my fork (https://github.com/trentm/opentelemetry-js-contrib/tree/overbalance/repair-eslint-trentm-tweaks). This is the commit:

https://github.com/trentm/opentelemetry-js-contrib/commit/39dc52e2b53d62e30927de99fc69dee00ac30f86

You should be able to cherry-pick that to your feature branch.
After this, the remaining lint issues are straightforward I think: mostly `--fix`-able with a few manual small changes. Only remaining lint errors are in these four pkgs:

```
% npm run lint
   ✖  nx run @opentelemetry/instrumentation-kafkajs:lint
   ✖  nx run @opentelemetry/instrumentation-mongoose:lint
   ✖  nx run opentelemetry-js-contrib-scripts:lint
   ✖  nx run @opentelemetry/propagation-utils:lint
...
```

prettier.config.js
trentm: The current value in opentelemetry-js-contrib and opentelemetry-js.git is `avoid`.

When looking at just the TypeScript files under `packages/*`, i.e. running:

```
npx prettier --write 'packages/**/*.ts'
```

with both values for `arrowParens`, there are *no* changes for this style with `avoid`, but approximately 1344 changes with `always`.

```bash
vi prettier.config.js  # set to "always"
npx prettier --write 'packages/**/*.ts'
git diff | rg '^\+' | rg '=>' | wc -l      # prints 1344 (some are false positives)
```

**I think for now we should stick with `avoid`** and *separately* consider changing the style.

When I look outside of `packages/*`, I see some examples in `scripts/*.js` that are using the other style.

eslint.config.mjs
overbalance: in a previous iteration they did not overlap. good catch

prettier.config.js
overbalance: That's a lot... I was comparing to the number of total changes in this branch 🫠

overbalance:
@trentm I pulled in your changes and updated the header regex because of this alert https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3001/checks?check_run_id=51300982566

Compile and lint are working for me locally. I also attempted to give you direct access to the branch.

trentm:
@overbalance I've lost the plot. Where are we with this PR? No *rush* from me, I'd just be happy to see this get over the line.

overbalance:
> @overbalance I've lost the plot. Where are we with this PR? No _rush_ from me, I'd just be happy to see this get over the line.

I just rebased and backed out a few unnecessary changes. I left notes in the eslint config itself. I tried to match the old rules as best I could but there are new defaults that surfaced a few issues. Those items have been set to warn. This is ready to review now that all checks are passing.

eslint.config.mjs
```diff
@@ -0,0 +1,218 @@
...
+      '**/packages/instrumentation-document-load/**/*',
+      '**/packages/instrumentation-long-task/**/*',
+      '**/packages/plugin-react-load/**/*',
+    ],
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        Zone: 'readonly',
+        Task: 'readonly',
+      },
```
trentm, line 214: How had the `Zone` and `Task` usages in "packages/instrumentation-user-interaction" been passing before this?

package.json
```diff
@@ -27,7 +27,7 @@
...
-    "lint:fix": "nx run-many -t lint:fix && npm run lint:markdown:fix",
+    "lint:fix": "nx run-many -t lint --fix && npm run lint:markdown:fix",
     "lint:deps": "npx --yes knip@5.33.3 --dependencies --production --tags=-knipignore",
     "lint:examples": "./scripts/npm-run-in-examples.js --if-present lint",
     "lint:examples:fix": "./scripts/npm-run-in-examples.js --if-present lint:fix",
```
trentm: I think we can drop `lint:examples` and `lint:examples:fix` now, and them being called by the "lint" target. IIUC files under `examples/*` are being linted by the `lint` target now.

packages/auto-configuration-propagators/package.json
```diff
@@ -22,8 +22,6 @@
     "clean": "rimraf build/*",
     "compile": "tsc -p .",
     "compile:with-dependencies": "nx run-many -t compile -p @opentelemetry/auto-configuration-propagators",
-    "lint": "eslint . --ext=ts,js,mjs",
-    "lint:fix": "eslint . --ext=ts,js,mjs --fix",
```
trentm, line 26: No longer being able to lint (and easily lint) in just a package dir is a slight bummer.
Thoughts on adding something like this to each package.json:

```
    "lint": "nx run-many -p @opentelemetry/instrumentation-bunyan -t lint",
```

as has been done with `"compile:with-dependencies"`?
Could be done separately. I don't want to make work for you.

packages/host-metrics/.eslintrc.js
```diff
@@ -1,22 +0,0 @@
...
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-module.exports = {
-  env: {
-    mocha: true,
```
trentm, line 18: Do you happen to know where `mocha` top-level defines get allowed? Or why using those pass in the new configuration?

This is the effective config for a .ts file (from running `npx eslint --print-config packages/instrumentation-bunyan/src/instrumentation.ts > effective-config-ts.after.txt`):
https://gist.github.com/trentm/08ba8623d58f534dc2983ef15b0250b0

No `env` in the new flat config. No mention of `mocha` or its `describe` global. [/shrug]

tsconfig.base.json
```diff
@@ -21,9 +22,7 @@
     "sourceMap": true,
     "strict": true,
     "strictNullChecks": true,
-    "target": "es2022"
-  },
-  "exclude": [
-    "node_modules"
-  ]
+    "target": "es2022",
+    "types": ["node", "mocha", "webpack-env"]
```
trentm: Changing the tsconfig doesn't belong in this PR, does it?

Oh, is this meant to be the answer for to my "why does lint pass now with `mocha` global usage?" question from above?

This means that runtime files have Mocha types available during the build. Seems odd/wrong.

eslint.config.mjs
trentm: I guess this was from the `env: browser` entries in the .eslintrc.js files for browser-y packages.

packages/host-metrics/.eslintrc.js
trentm: See my next review comment. I guess the mocha globals/types are coming from the "types" added to tsconfig.base.json.

packages/auto-configuration-propagators/package.json
overbalance: We have two options:
1. Add `"lint": "nx lint",` to each package.json
2. Run `npx nx lint` when you're in the package folder

NX automatically picks up the package name (I think from `$npm_package_name` env var)

tsconfig.base.json
overbalance: Definitely wrong. All types are inferred when there is no entry. I only changed it to satisfy a newer version of typescript that we agreed to revert. Removed.

packages/host-metrics/.eslintrc.js
overbalance: You are correct that the types in tsconfig were satisfying the linter, but it turns out removing the tsconfig entry entirely allows the types to be read from node_modules. I commented below: removing the entry works globally.

packages/auto-configuration-propagators/package.json
trentm: > NX automatically picks up the package name (I think from `$npm_package_name` env var)

Nice. I hadn't realized that. Okay, we can decide whether we want `"lint": "nx lint",` back later. I'm content telling people to use `npx nx lint` to start with.

.prettierignore
```diff
@@ -0,0 +1,9 @@
+*.md
+*.json
+*.jsonc
+*.json5
+*.yml
+*.yaml
+!package.json
+
+/.nx
```
trentm, line 9: Should we consider adding this `/.nx` to the top `ignores` section in the eslint config?

```
const baseConfig = tseslint.config(
  {
    ignores: [
      '**/build/**',
      '**/coverage/**',
      '**/dist/**',
      '**/node_modules/**',
    ],
  },
```

Otherwise, I'm not sure if eslint is attempting to process files in .nx/...

lerna.json
```diff
@@ -1,20 +1,6 @@
 {
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "npmClient": "npm",
   "version": "independent",
-  "changelog": {
```
legendecas, line 4: lerna-changelog reads from `lerna.json#changelog` field. I think this is not a necessary change: https://github.com/lerna/lerna-changelog/blob/41059809bac9ee9b12fd4c4241cff07f63d083f7/src/configuration.ts#L87-L92

package.json
```diff
@@ -26,11 +26,9 @@
     "test:with-services-env": "cross-env NODE_OPTIONS='-r dotenv/config' DOTENV_CONFIG_PATH=./test/test-services.env npm test",
     "test-all-versions:with-services-env": "cross-env NODE_OPTIONS='-r dotenv/config' DOTENV_CONFIG_PATH=./test/test-services.env npm run test-all-versions",
     "changelog": "lerna-changelog",
-    "lint": "nx run-many -t lint && npm run lint:deps && npm run lint:readme && npm run lint:markdown && npm run lint:semconv-deps && npm run lint:examples",
-    "lint:fix": "nx run-many -t lint:fix && npm run lint:markdown:fix",
+    "lint": "nx run-many -t lint && npm run lint:deps && npm run lint:readme && npm run lint:markdown && npm run lint:semconv-deps",
```
pichlermarc, line 29: given that the `lint` target now does not exist anymore on packages this never actually calls `eslint`, right? 🤔

eslint.config.mjs
```diff
@@ -0,0 +1,218 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import eslint from '@eslint/js';
+import tseslint from 'typescript-eslint';
+import globals from 'globals';
+import nodePlugin from 'eslint-plugin-n';
+import yalhPlugin from 'eslint-plugin-yet-another-license-header';
```
pichlermarc, line 21: looks like the installed `eslint-plugin-prettier` is never used. is this intentional?

lerna.json
overbalance: @legendecas If you look on line 94 of your link, it's also read from package.json. Look at my diff, `changelog` was already in package.json, I just deduped it from lerna.json. I can revert but we should use one or the other.

package.json
overbalance: Check out nx.json. The new NX plugin is running `eslint` in the context of each package with the "name" lint which is why `npx nx lint` works:
```
"plugins": [
    {
      "plugin": "@nx/eslint/plugin",
      "options": {
        "targetName": "lint",
        "cache": true
      }
    }
  ]
```
⬇️ 
```
NX   Successfully ran target lint for project @opentelemetry/instrumentation-cucumber (412ms)
```

eslint.config.mjs
overbalance: Yes, I was following their best practices for eslint 9: https://prettier.io/docs/integrating-with-linters#notes
Good catch, I will remove them from package.json.
They have a command that can check your rules for conflicts and the two it flagged I chose to ignore since keeping them in sync with the prettier config is trivial:
```
❯ npx eslint-config-prettier eslint.config.mjs
The following rules are enabled but cannot be automatically checked. See:
https://github.com/prettier/eslint-config-prettier#special-rules

- no-unexpected-multiline
- quotes

Other than that, no rules that are unnecessary or conflict with Prettier were found.
```

overbalance:
@pichlermarc 
> Thanks for working on this! 🙌
> 
> Running `eslint` locally, I'm seeing an OOM error......

Thanks! I'm not sure what to do about this one. I don't think you can run `eslint` without nx because it will try to add all matching files to the end of your command and fail (as you saw).

.prettierignore
overbalance: The nx cache does not contain any js or ts files, so it won't find any matches but good idea to be explicit in case they are present in the future.

package.json
pichlermarc: Ah, I missed that - sorry. 🤦 All good then 🙂

eslint.config.mjs
pichlermarc: thanks, sounds good. just wanted to double-check 🙂

lerna.json
trentm: Rabbit hole: I don't think *either* of these blocks of `"changelog": ...` are used.  I don't think `lerna-changelog` is actually used anymore in this repo that I can tell, though I could be wrong.  

The content added to the **top-level "CHANGELOG.md" file used to** use this `npm run changelog` script I think -- its headers show signs of having used those header values. However, that CHANGELOG.md file is unused since 2022.

Now the new changelog content is in "packages/*/CHANGELOG.md" and those are updated by `release-please`, e.g. this part of the last release PR: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3135/files#diff-5edc63adcef2085e5eea10b22f8d8ec1a026edeef5daa20799756bf8fcb165d2
Here is the distribution of those header values in the current set of CHANGELOG.md files:

```
% rg --no-filename '^### ' */CHANGELOG.md | rg -v github.com  | sort | uniq -c |
 sort -n
   3 ### Performance Improvements
  66 ### Miscellaneous Chores
 140 ### ⚠ BREAKING CHANGES
 999 ### Dependencies
1033 ### Bug Fixes
1225 ### Features
```

Those values match the set in release-please here:
- most of the values here: https://github.com/googleapis/release-please/blob/a9b82178ce8040af09e55be509911fa36e0c20e7/src/util/filter-commits.ts#L20-L32
- `### Dependencies` being handled separately here: https://github.com/googleapis/release-please/blob/a9b82178ce8040af09e55be509911fa36e0c20e7/src/plugins/workspace.ts#L486-L495
- `BREAKING CHANGES` being handled separately here by a dep used by release-please: https://github.com/conventional-changelog/conventional-changelog/blob/1552de5434163eb63d3bf73d97bcb3bf34de8586/packages/conventional-changelog-conventionalcommits/src/writer.js#L90

tl;dr: I think we should clean up the changelog vestiges **in a separate PR**. Ideally this PR would not bother changing this part of lerna.json, so that it stays focused. However, I thikn it is fine and I just would like to merge this now.

.prettierignore
trentm: > The nx cache does not contain any js or ts files

```
% npx eslint --print-config .nx/cache/18287085831844831640/packages/instrumentation-router/src/version.ts
{
  "settings": {
    "node": {
      "tryExtensions": [
        ".js",
        ".json",
        ".node",
        ...
```

trentm: Looks like only the "version.ts" files.

```
% find .nx | rg -v build/ | rg '\.(ts|js)$'
.nx/cache/11224883197310830386/packages/instrumentation-mysql2/src/version.ts
.nx/cache/5581609661682288038/packages/instrumentation-document-load/src/version.ts
.nx/cache/16519942250677500878/packages/instrumentation-connect/src/version.ts
...
```

overbalance: My mistake. Just looked manually after linting, but not after the other commands.

PR #3003, opened 2025-08-29
test(instr-mongodb): fix tests mongdb tests for 6.19.0

david-luna:

The new release of `mongodb@6.19.0` changed the internal model of, at least, the insert command response. this does not break the instrumentation but the `responseHook` tests were relying on that to make certain assertions. As a result running `test-all-versions` failed when selecting such version of the package.

- update dependency to v6.19.0
- change test code to consider both models in the response object

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24502,13 +24502,13 @@
       }
     },
     "node_modules/mongodb": {
-      "version": "6.15.0",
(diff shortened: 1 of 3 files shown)


pichlermarc (approved):
thank you for fixing this 🙌

Issue #2933, opened 2025-07-10
Extend `@opentelemetry/resource-detector-gcp` with support for Google Cloud Run


Silthus:
### Is your feature request related to a problem? Please describe

Right now the GCP resource detection for NodeJS does not properly detect running in a Google Cloud Run instance.
For now it is required to set the required attributes manually or use the otel-collector sidecar, which sets the [correct `faas` and `host.type` attributes](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#gcp-metadata) for Google Cloud Run.

### Describe the solution you'd like to see

It would be great if the `@opentelemetry/resource-detector-gcp` would detect and set the correct resource attributes for Google Cloud Run to the same as the [`otel-collector/processor/resource_detection/gcp`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#gcp-metadata).

### Additional context

The environment variables from the [Cloud Run runtime contract](https://cloud.google.com/run/docs/container-contract#metadata-server) can be used to detect and extract the running Cloud Run instance.

[See here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/detectors/gcp/faas.go) for the relevant env variables used by the Go GCP SDK.

Silthus:
How can I move this forward?
Are you open to contributions on the topic?

mmanciop:
This issue would be solved, at least for the `faas.*` part, by the planned upstreaming of the GCP detectors agreed upon in #2818

aabmass:
I'm updating the GCP resource detector in our repo to use SDK 2.x, then will be sending a PR to transfer it to this repo.

Silthus:
Thank you @aabmass and @pichlermarc for tackling this 🎉

Unfortunately the current implementation is lacking the support for [Cloud Run Jobs](https://cloud.google.com/run/docs/container-contract) and only works for Cloud Run Services. For jobs it would need to detect the `CLOUD_RUN_JOB` and `CLOUD_RUN_EXECUTION` env variables and then add those as `faas.*` attributes. And I don't know if there is any kind of attribute that makes it possible to distinguish between the different types of Cloud Run flavors (Job, Service, Function). If so that would be great to add as well.

[...]

PR #3007, opened 2025-08-29
feat(resource-detector-gcp)!: contribute Google's comprehensive resource detector

aabmass:

Moving the "official" GCP resource detector from [GoogleCloudPlatform/opentelemetry-operations-js](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/tree/main/packages/opentelemetry-resource-util) ([@google-cloud/opentelemetry-resource-util](https://www.npmjs.com/package/@google-cloud/opentelemetry-resource-util) on NPM)  to this repo. This package will now be the official one we recommend to customers and support, rather than living in Google's repo 😃. Once this is released, we will delete the detector code out of Google's repo.

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2933

I have kept individual commits for easier review. In summary:
- Copied the detector code and tests from [GoogleCloudPlatform/opentelemetry-operations-js](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/980b7ef4f44408d54afe712c77af037905db56e5/packages/opentelemetry-resource-util/src/detector/detector.ts#L215) here, replacing the existing implementation
- Add OpenTelemetry Authors copyright while retaining Google copyright (both are Apache 2).
- Small tweaks to match repo style, get `GcpDetectorIntegration.test.ts` passing, and `npm run lint:format`.
- Deleted the previous code and unit tests.

I've marked this as a breaking change, since the behavior will be different. Overall this implementation covers more environments (GCE, GKE, Cloud Run, Cloud Run (+Functions), App Engine Standard and Flex) and follows Resource SemConv. Google has an extensive E2E test suite that deploys the detector in all supported environments, which runs in [GoogleCloudPlatform/opentelemetry-operations-js](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/) today. I will be updating those tests to periodically run this resource detector and make any fixes here.

For more context, see discussion on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2818.


aabmass:
@psx95 would you mind giving a review here?

packages/resource-detector-gcp/src/detectors/GcpDetector.ts
```diff
@@ -1,4 +1,5 @@
 /*
+ * Copyright 2022 Google LLC
```
psx95, line 2: The year probably needs to be updated to 2025.

packages/resource-detector-gcp/src/detectors/GcpDetector.ts
```diff
@@ -14,147 +15,216 @@
...
-    } catch {
-      return '';
-    }
+  // Note the order of these if checks is significant with more specific resources coming
+  // first. E.g. Cloud Functions gen2 are executed in Cloud Run so it must be checked first.
+  if (await gke.onGke()) {
+    return await gkeResource();
+  } else if (await faas.onCloudFunctions()) {
+    return await cloudFunctionsResource();
+  } else if (await faas.onCloudRun()) {
```
psx95, line 88: Was  there no Cloud Run Jobs ?

aabmass: No it was already missing. Can add it as a follow up

aabmass:
> I'm assuming the newly added files are copied over from the operations repo. I noticed that probably Cloud Run Jobs are missing ?

Yup 335091fb8eb78bee557f8cfc6fa4c00cba3f158c

packages/resource-detector-gcp/src/detectors/GcpDetector.ts
aabmass: Discussed offline, not needed

david-luna (approved):
I think we are still unable to suppress tracing on this detector. But it's good for now
ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2320

aabmass:
> I think we are still unable to suppress tracing on this detector. But it's good for now.

Ya based on the discussion I don't think this problem will be solved

PR #3010, opened 2025-09-03
feat(instrumentation-ioredis): add Redis Cluster instrumentation support

PavelPashov:

Currently when using the `ioredis` in cluster mode the `instrumentation-ioredis` package wraps only the standalone client methods, which cluster mode still ultimately uses, however this is a bit misleading and does not represent the correct tracing.

- Added `instrumentCluster?: boolean` to enable cluster instrumentation
- Instruments `Cluster.sendCommand()` and `Cluster.connect()` methods
- Added cluster instrumentation tests
- Updated the README

## Additional Information
Cluster instrumentation is gated behind a configuration flag to maintain backward compatibility for users who may not need it. Additionally, the attributes for standalone clients and clusters are not fully compatible.


pichlermarc:
cc @blumamir @naseemkullah (component owners)

PR #3023, opened 2025-09-05
chore(instrumentation-pg): skip test if db not running

overbalance:

- Locally I was unable to pass `npm run test` if the db service was not started.

- I copied the guard from the regular test to the ESM version. I have no prior experience with the db packages so please close this if it's off base.


maryliag (approved):
Great catch and thanks for adding the missing code! 😄

maryliag:
@overbalance for some reason is not showing me the option to update with the main before merging, would you mind updating with the latest from main so I can merge it? thanks

PR #3025, opened 2025-09-08
Return underlying modules from redis instrumentation

DrewCorlin:
Return version-specific module instrumentations from version-agnostic instrumentation, to enable consumers of this library to get access to the instrumentations that are being applied without having to know this instrumentation is really just a wrapper for 2 others.

I'm the author of https://github.com/DrewCorlin/opentelemetry-node-bundler-plugins. This library depends on `getModuleDefinitions()` to know what modules to patch (in it's case to patch them during the bundling of an app). Updating to the latest version to support instrumentation of redis@v5 uncovered this.

Rather than special casing this in my plugin I thought it made sense to have this module behave consistently with the rest. 

Also let me know if I should remove the link to my plugin here. I thought it was useful context, but understand if links to other projects that depend on this aren't desired.


trentm:
> https://github.com/DrewCorlin/opentelemetry-node-bundler-plugins

👀   Nice!

PR #3032, opened 2025-09-10
build: upgrade TypeScript to 5.9.2, fix CommonJS module resolution

overbalance:
# DRAFT
Beyond compile and test, I'm relying on CI to surface issues so I'm keeping this in draft until I see the results.

This PR upgrades TypeScript from 5.0.4 to 5.9.2. The newer version found some type issues that needed fixing, and also required adjusting our module resolution settings since the repo uses CommonJS without file extensions in imports.

Upgraded the monorepo from TypeScript 5.0.4 to 5.9.2 with fixes for stricter type checking.
Hoisted all build dev dependencies to the root.

### 1. devDependency Hoisting
Consolidated all devDependencies to root package.json for consistent versions across packages.

### 2. Root TypeScript Configuration - TypeScript 5.9.2's `node16` resolution requires extensions
- **tsconfig.base.json**: Changed to `module: "commonjs"` and `moduleResolution: "node"` (from "node16")
- **tsconfig.base.esm.json**: Changed `moduleResolution` to `"bundler"`
- **tsconfig.base.esnext.json**: Changed `moduleResolution` to `"bundler"`

### @opentelemetry/instrumentation-socket.io
- **Issue:** Import statement `import * as expect from 'expect'` failed - `expect` ^29 changed to default export
- **Fix:** Changed to `import expect from 'expect'` in test files

### @opentelemetry/instrumentation-dns
- **Issue:** TypeScript couldn't infer types for resolver mock functions in tests
- **Fix:** Added explicit type annotations for mock resolver callbacks

### @opentelemetry/instrumentation-aws-sdk
- **Issues:**
  - Nock mock responses had Content-Length: 143 but actual body was 147 bytes
  - Tests import from `@opentelemetry/semantic-conventions/incubating` subpath which CommonJS can't resolve
  - AWS SDK v3.883.0 changed response body handling (no more `transformToString()` method)
- **Fixes:**
  - Updated all Bedrock mock files to match their Content-Length
  - Added path mapping in tsconfig.json pointing to `node_modules/@opentelemetry/semantic-conventions/build/src/index-incubating.d.ts`

### @opentelemetry/instrumentation-openai
- **Issues:**
  - Mocha tests failed with "Unknown file extension '.ts'" in ESM mode
  - Missing formdata-node dependency for multipart form tests
  - @types/node has Buffer/ArrayBuffer type incompatibilities that TS 5.9.2 catches
- **Fixes:**
  - Created .mocharc.json with `"node-option": ["loader=ts-node/esm"]`
  - Added formdata-node as devDependency
  - Added `skipLibCheck: true` in tsconfig.json

### @opentelemetry/resource-detector-gcp
- **Issues:**
  - Same ESM/TypeScript loader issues as OpenAI
  - Missing json-bigint and google-logging-utils dependencies
- **Fixes:**
  - Created .mocharc.json with ESM/TypeScript configuration
  - Added missing dependencies to package.json

## Why skipLibCheck for OpenAI?

Required due to @types/node Buffer/ArrayBuffer incompatibilities:
```
Interface 'Buffer' incorrectly extends interface 'Uint8Array<ArrayBufferLike>'
Type 'Buffer' does not satisfy the constraint 'ArrayBufferView'
Overload signatures must all be optional or required
```

TypeScript 5.9.2 catches these errors that earlier versions missed. `skipLibCheck: true` bypasses third-party type errors outside our control.

# Major Files for Maintainer Review

## Critical Build Configuration
- `tsconfig.base.json` - Changed module resolution strategy for CommonJS
- `tsconfig.base.esm.json` - Changed module resolution to bundler
- `tsconfig.base.esnext.json` - Changed module resolution to bundler
- `packages/instrumentation-aws-sdk/tsconfig.json` - Added path mapping for semantic-conventions/incubating
- `packages/instrumentation-openai/tsconfig.json` - Added skipLibCheck for @types/node issues

## Test Configuration Changes
- `packages/instrumentation-openai/.mocharc.json` - New file for ESM/TypeScript loader
- `packages/resource-detector-gcp/.mocharc.json` - New file for ESM/TypeScript loader

## Test Code Changes
- `packages/instrumentation-socket.io/test/*.ts` - Changed expect import syntax
- `packages/instrumentation-dns/test/functionals/dns-disable.test.ts` - Added type annotations
- `packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts` - Changed response.body parsing

## Mock Data Changes
- `packages/instrumentation-aws-sdk/test/mock-responses/bedrock-*.json` - Fixed Content-Length headers (7 files)


pichlermarc:
FYI upgrading `typescript` is often a breaking change as the `typescript` project does not follow Semantic Versioning for the output it produces.

overbalance:
> FYI upgrading `typescript` is often a breaking change as the `typescript` project does not follow Semantic Versioning for the output it produces.

For sure. I've been playing with the version number to see how far we can comfortably upgrade. It's whack-a-mole with this much code so I'm thinking we leave it at 5.0.4, and focus on consolidating the devdependencies.

packages/instrumentation-long-task/karma.conf.js
```diff
@@ -18,9 +18,17 @@ const karmaWebpackConfig = require('../../karma.webpack');
 const karmaBaseConfig = require('../../karma.base');
 
 module.exports = config => {
+  {
+    const plugins = (karmaBaseConfig.plugins = []);
+    const rootPackageJson = require('../../package.json');
+    const toAdd = Object.keys(rootPackageJson.devDependencies || {})
```
overbalance, line 24: Needed for hoisting

karma.webpack.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
       // because the `util` package expects there to be a global variable named `process`.
       // Thanks to https://stackoverflow.com/a/65018686/14239942
       // NOTE: I wish there was a better way as this pollutes the tests with a defined 'process' global.
-      process: 'process/browser'
+      process: 'process/browser.js'
```
overbalance, line 39: There was a warning that suggested this change.

tsconfig.base.esnext.json
```diff
@@ -1,9 +1,9 @@
 {
+  "$schema": "https://json.schemastore.org/tsconfig",
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "module": "esnext",
-    // target should be aligned with tsconfig.base.json
```
overbalance, line 5: This seems to contradict the idea of an esnext export

packages/instrumentation-openai/package.json
```diff
@@ -53,19 +53,9 @@
...
     "@opentelemetry/sdk-trace-base": "^2.0.0",
     "@opentelemetry/sdk-trace-node": "^2.0.0",
-    "@types/node": "18.18.14",
-    "@typescript-eslint/eslint-plugin": "5.8.1",
-    "@typescript-eslint/parser": "5.8.1",
-    "cross-env": "7.0.3",
-    "expect": "29.2.0",
-    "nock": "^14.0.0",
-    "nyc": "17.1.0",
+    "form-data-encoder": "^4.1.0",
```
trentm: Is this one needed?

The openai dep brings in this package, though a MUCH older one.

```
opentelemetry-js-contrib@0.1.0 /Users/trentm/src/opentelemetry-js-contrib
└─┬ @opentelemetry/instrumentation-openai@0.4.0 -> ./packages/instrumentation-openai
  └─┬ openai@4.104.0
    └── form-data-encoder@1.7.2
```

packages/instrumentation-openai/tsconfig.json
```diff
@@ -2,10 +2,8 @@
   "extends": "../../tsconfig.base",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build"
+    "outDir": "build",
+    "skipLibCheck": true
```
trentm, line 6: nit: If you have it, it would be nice to see what the compilation error was.
E.g. in instrumentation-aws-sdk skipLibCheck was necessary because some conflict caused by aws-sdk's deps publishing with features from a newer version of TypeScript than we use.

packages/instrumentation-redis/package.json
```diff
@@ -50,22 +50,15 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@opentelemetry/api": "^1.3.0"
+    "@opentelemetry/api": "^1.3.0",
+    "redis": "^5.6.0"
```
trentm: Are we sure we want a change in peer deps here?

packages/resource-detector-gcp/.mocharc.json
trentm, line 1: Curious why this is the odd package out in needing this.

tsconfig.base.esm.json
```diff
@@ -1,7 +1,8 @@
 {
+  "$schema": "https://json.schemastore.org/tsconfig",
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "module": "es2022",
-    "moduleResolution": "node16"
+    "moduleResolution": "bundler"
```
trentm: I don't have a good sense of the potential impact of this. (https://www.typescriptlang.org/tsconfig/#moduleResolution)

packages/instrumentation-aws-sdk/test/mock-responses/bedrock-invokemodel-adds-amazon-nova-model-attributes-to-span.json
```diff
@@ -36,7 +38,7 @@
             "Content-Type",
             "application/json",
             "Content-Length",
-            "68",
+            "181",
```
trentm, line 41: Guessing these changes are from the nock 13 -> 14 upgrade?

trentm (commented):
Early feedback. I think this is looking pretty good. ;)

- I don't know the karma code well.
- I don't have a feel for the impact of the moduleResolution change in tsconfig.json files.
- (not your fault) Eww on the `assert` module (because it shadows the core node.js module). Granted that was already installed at the top-level so this PR doesn't impact that at all.

Some questions/notes inline.

packages/instrumentation-openai/package.json
overbalance: This is required for TAV to pass. Maybe they can review it and confirm ?

packages/instrumentation-openai/tsconfig.json
overbalance: This is an artifact from when I was testing upgrading typescript. OpenAI compiles successfully with 5.0.4 so I will revert this change.

packages/resource-detector-gcp/.mocharc.json
overbalance: In cjs projects, mocha isn't able to process .mjs files without this config.
GCP has this unusual fixture:
`argv: ['fixtures/detect-with-http-instrumentation.mjs'],`

tsconfig.base.esm.json
overbalance: Docs: https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports

The main point to be aware of is that you may only import files explicitly exported by the author. The "exports" key is read first when present in imported packages, and prevents paths like this: 
```js
import Something from "@swc/core/dist/esm/some-internal-folder/types/internal.js"
```

packages/instrumentation-redis/package.json
overbalance: I'm not super familiar with how TAV works but when it installs different versions of redis, it only passes if no single version is specified in dependencies. I would like the owner to review if possible.

packages/instrumentation-openai/package.json
overbalance: I previously saw issues when running TAV but I just tried again and it's passing. I'll remove.

packages/resource-detector-gcp/.mocharc.json
trentm: mocha shouldn't need to process that file, because it is a "fixture": There is a `*.test.ts` file in this package the exec's that fixture in a subprocess.  Similar thing in a number of other packages' test suites.

tsconfig.base.esm.json
trentm: @overbalance We discussed this with other maintainers. Would it be straightforward to take the tsconfig changes **out of this PR** and do them in a separate PR? That would allow this PR to just be about "hoist all devDeps to root" and we could discuss the other change without all the noise of the hoisting.

packages/instrumentation-openai/package.json
trentm: Okay, noted. I'll take a look. I know TAV well and I'm the component-owner of instr-openai.

packages/resource-detector-gcp/.mocharc.json
overbalance: Ok. Let me run it without the config and I'll post the error here.

tsconfig.base.esm.json
overbalance: Yep, reverted.

overbalance:
@trentm If Node 18 support is needed then there's more work to do. I missed this requirement when I started this work. 😕 How do you want to proceed? Try to roll back dependency versions until it's happy?
```bash
❯ npm ci
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: '@azure-rest/core-client@2.5.1',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.8', npm: '10.8.2' }
npm warn EBADENGINE }
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: '@azure/core-auth@1.10.1',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.8', npm: '10.8.2' }
npm warn EBADENGINE }
```

packages/resource-detector-gcp/.mocharc.json
overbalance: Working now. Prob another artifact from TS 5.9

JamieDanielson:
> @trentm If Node 18 support is needed then there's more work to do. I missed this requirement when I started this work. 😕 How do you want to proceed? Try to roll back dependency versions until it's happy?
> 

🤔  We say we [support Active and Maintenance LTS versions of Node](https://github.com/open-telemetry/opentelemetry-js?tab=readme-ov-file#node-support), and [Node 18 is officially end of life](https://nodejs.org/en/about/previous-releases). So it might be fine in Contrib to drop support for 18. That's a bigger discussion though (cc @open-telemetry/javascript-maintainers ) and we'll need to update docs too.

trentm:
> If Node 18 support is needed then there's more work to do. I missed this requirement when I started this work. 😕 How do you want to proceed? Try to roll back dependency versions until it's happy?
> 
> ```shell
> ❯ npm ci
> npm warn EBADENGINE Unsupported engine {
> ```

@overbalance I don't think those EBADENGINE warnings are blocking for this PR. We already have a number of those warnings in the current state.  Unfortunately an EBADENGINE warning will matter for *some* deps, but not for some deps that are just for testing. For example this one:

```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: '@azure/core-auth@1.10.1',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.8', npm: '10.8.2' }
npm warn EBADENGINE }
```

That dep is used by tedious@17:

```
% npm ls @azure/core-auth
opentelemetry-js-contrib@0.1.0 /Users/trentm/src/opentelemetry-js-contrib
└─┬ @opentelemetry/instrumentation-tedious@0.25.0 -> ./packages/instrumentation-tedious
  └─┬ tedious@17.0.0
    ├─┬ @azure/identity@3.4.2
    │ ├── @azure/core-auth@1.10.1
    │ ├─┬ @azure/core-client@1.10.1
    │ │ └── @azure/core-auth@1.10.1 deduped
    │ └─┬ @azure/core-rest-pipeline@1.22.1
    │   └── @azure/core-auth@1.10.1 deduped
    └─┬ @azure/keyvault-keys@4.10.0
      ├─┬ @azure-rest/core-client@2.5.1
      │ └── @azure/core-auth@1.10.1 deduped
      ├── @azure/core-auth@1.10.1 deduped
      └─┬ @azure/keyvault-common@2.0.0
        └── @azure/core-auth@1.10.1 deduped
```

which is a test dep. We intentionally install a recent-ish major version of `tedious` so that unit tests (i.e. `npm test`) for an instrumentation (instrumentation-tedious in this case) tests with a relatively modern version of the target package.  We then rely on our `test-all-versions` tests to handle testing against those versions of `tedious` that work with older Node.js versions (e.g. Node.js v18)

overbalance:
@trentm Those warnings are blocking CI from what I see. What do you recommend?

trentm:
> Those warnings are blocking CI from what I see.

I don't know what is going on. I can `npm ci` this feature branch locally.

This CI step failed with a network error: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18537223726/job/53057847496?pr=3032

```
...
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'walk-up-path@4.0.0',
npm WARN EBADENGINE   required: { node: '20 || >=22' },
npm WARN EBADENGINE   current: { node: 'v18.19.0', npm: '10.2.3' }
npm WARN EBADENGINE }
npm ERR! code ECONNRESET
npm ERR! network aborted
npm ERR! network This is a problem related to network connectivity.
npm ERR! network In most cases you are behind a proxy or have bad network settings.
npm ERR! network 
npm ERR! network If you are behind a proxy, please make sure that the
npm ERR! network 'proxy' config is set properly.  See: 'npm help config'

npm ERR! A complete log of this run can be found in: /home/runner/.npm/_logs/2025-10-17T23_04_44_742Z-debug-0.log
Error: Process completed with exit code 1.
```

The other test jobs are stuck in `npm ci`.   I cancelled the run after 20m.
I think the "network aborted" above is just a fluke. The real issue is the stuck `npm ci` steps.

The `lint` workflow, which does `npm ci --ignore-scripts`, worked fine.

Ideas on what this could be:

- Some install/build/postinstall script in one of the deps is breaking things.
- Some GitHub Actions or npmjs bug/incident that is causing this to break.
- Some limit is being hit in GH Actions? THis is total guess.

overbalance:
@trentm I repaired the lockfile and this is finally 🟢. Thanks for your help.

packages/instrumentation-openai/tsconfig.json
```diff
@@ -2,11 +2,7 @@
   "extends": "../../tsconfig.base",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build",
-    "skipLibCheck": true
```
trentm, line 6: Was `skipLibCheck` accidentally dropped? It was added recently for:

```
commit c352766cc93fc2ebf219392f1a52143a0a951f64
Author: Jared Freeze <overbalance@users.noreply.github.com>
Date:   2025-10-14T11:22:00-07:00 (2 weeks ago)

    chore(eslint): upgrade to eslint 9 (#3001)
```

Perhaps all changes to this file could be reverted. The rest of it is just formatting.

packages/instrumentation-redis/package.json
```diff
@@ -48,22 +48,15 @@
...
```
trentm: Was the move of this from devDependencies to peerDependencies accidental?

packages/instrumentation-socket.io/test/config.test.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { SocketIoInstrumentation, SocketIoInstrumentationConfig } from '../src';
-import * as expect from 'expect';
+import expect from 'expect';
```
trentm, line 17: nit: This is an unrelated change, right? (Ditto for the few below and the `import axios ...` one near the top.)

overbalance: The latest version of `expect` uses a default export, not a namespace. This is intentional.

packages/instrumentation-redis/package.json
overbalance: It is intentional. I think someone else asked about this in this PR. (I know there's too much noise to see it now.) Leaving Redis in devDeps was causing TAV to fail. Would you like me to reconfirm on a new branch?

packages/instrumentation-openai/tsconfig.json
overbalance: I kept the removal because it's no longer throwing a TS error and it seems better to type check as much as possible. I remember changes to the openai package coming in after I started this PR. Maybe they repaired the issue ? 
For simplicity I'll revert.

packages/instrumentation-redis/package.json
trentm: > I think someone else asked

I feel it was likely *me* that asked that before and that I said I'd try to figure out why it is necessary.

trentm: It isn't acceptable to add redis as a peerDep, because it'll result in *installing* that particular redis when a user attempts to install the instrumentation:

```
% cd .../packages/instrumentation-redis
% npm pack
...
opentelemetry-instrumentation-redis-0.56.0.tgz
% cp opentelemetry-instrumentation-redis-0.56.0.tgz ~/tmp/asdf/
```

Then in a play/clean npm package in "~/tmp/asdf":

```bash
% npm install ./opentelemetry-instrumentation-redis-0.56.0.tgz

% cat package.json
{
...
  "dependencies": {
    "@opentelemetry/instrumentation-redis": "file:opentelemetry-instrumentation-redis-0.56.0.tgz"
  }
}

% npm ls -a
asdf@1.0.0 /Users/trentm/tmp/asdf
└─┬ @opentelemetry/instrumentation-redis@0.56.0
  ├── @opentelemetry/api@1.9.0
  ├─┬ @opentelemetry/instrumentation@0.207.0
  │ ├─┬ @opentelemetry/api-logs@0.207.0
  │ │ └── @opentelemetry/api@1.9.0 deduped
  │ ├── @opentelemetry/api@1.9.0 deduped
  │ ├─┬ import-in-the-middle@2.0.0
  │ │ ├─┬ acorn-import-attributes@1.9.5
  │ │ │ └── acorn@8.15.0 deduped
  │ │ ├── acorn@8.15.0
  │ │ ├── cjs-module-lexer@1.4.3
  │ │ └── module-details-from-path@1.0.4
  │ └─┬ require-in-the-middle@8.0.1
  │   ├─┬ debug@4.4.3
  │   │ └── ms@2.1.3
  │   └── module-details-from-path@1.0.4 deduped
  ├── @opentelemetry/redis-common@0.38.2
  ├── @opentelemetry/semantic-conventions@1.37.0
  └─┬ redis@5.9.0
    ├─┬ @redis/bloom@5.9.0
    │ └── @redis/client@5.9.0 deduped
    ├─┬ @redis/client@5.9.0
    │ └── cluster-key-slot@1.1.2
    ├─┬ @redis/json@5.9.0
    │ └── @redis/client@5.9.0 deduped
    ├─┬ @redis/search@5.9.0
    │ └── @redis/client@5.9.0 deduped
    └─┬ @redis/time-series@5.9.0
      └── @redis/client@5.9.0 deduped
```

I'll see if I can find the TAV issue.

trentm: > I'll see if I can find the TAV issue.

TAV passes for me locally (on macOS, node v20.19.1) after I apply this change to move the "redis" dep back to "devDependencies", and run `npm i` to update package-lock.json:

```diff
diff --git a/packages/instrumentation-redis/package.json b/packages/instrumentation-redis/package.json
index 068654d0..a8912441 100644
--- a/packages/instrumentation-redis/package.json
+++ b/packages/instrumentation-redis/package.json
@@ -48,15 +48,15 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@opentelemetry/api": "^1.3.0",
-    "redis": "^5.6.0"
+    "@opentelemetry/api": "^1.3.0"
   },
   "devDependencies": {
     "@opentelemetry/api": "^1.3.0",
     "@opentelemetry/context-async-hooks": "^2.0.0",
     "@opentelemetry/contrib-test-utils": "^0.54.0",
     "@opentelemetry/sdk-trace-base": "^2.0.0",
-    "@opentelemetry/sdk-trace-node": "^2.0.0"
+    "@opentelemetry/sdk-trace-node": "^2.0.0",
+    "redis": "^5.6.0"
   },
   "dependencies": {
     "@opentelemetry/instrumentation": "^0.207.0",
```

Then to run TAV tests I did:

```
cd packages/instrumentation-redis
npm run test-services:start
npm run test-all-versions:with-services-env
```

So I'm hoping we can just make that change. Whatever the thing is that you hit is hopefully gone now.

trentm: Here is my patch to your feature branch: https://gist.github.com/trentm/c415da5938e4ddf81aea9a7670bd4118

Unless I'm missing something, I don't have perms to push to your branch:

```sh
% git push embrace-io overbalance/hoist-all-shared-deps
ERROR: Permission to embrace-io/opentelemetry-js-contrib.git denied to trentm.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

trentm:
<details>
<summary>Unhelpful npm rant</summary>

I want to try to care about reviewing package-lock changes, but ... npm, I don't even. This latest update:

1. updates a bunch of other packages, because the only reasonable way I know of (and probably what you did) was completely regenerate the package-lock file by deleting and re-running `npm install`.
2. Loses all the 'integrity' and 'resolved' fields, e.g.:

```diff
-      "resolved": "https://registry.npmjs.org/uuid/-/uuid-9.0.1.tgz",
-      "integrity": "sha512-b+1eJOlsR9K8HJpow9Ok3fiWOWSIcIzXodvv0rQjVoOVNpWMpxf1wZNpt4y9h10odCNrqnYp1OBzRktckBe3sA==",
```

So basically those fields are just useless noise from npm in lock files. What a waste. The package-lock.json tooling design is ...poor.

</details>

package-lock.json
trentm, line 1: This has the "package-lock has lost all the platform-specific optionalDependencies except the one platform" problem again, described at https://cloud-native.slack.com/archives/C08T7MZTV8W/p1761781178575799?thread_ts=1761762353.429559&cid=C08T7MZTV8W

You'll need to manually re-add them, or do some voodoo process under a full moon something like the following to correctly re-gen the lock file:

```
rm package-lock.json
rm -rf node_modules
npm i
```

trentm (changes requested):

package-lock.json
trentm: Unforunately in the latest update a number of deps in package-lock.json were incorrectly marked with `"peer": true` due to a bug in npm. 

For example, here is a subset of them (this is partial output of running https://github.com/trentm/npm-tools/blob/main/bin/package-lock-diff):

```diff
- node_modules/react: 17.0.2 (dev)
+ node_modules/react: 17.0.2 (dev, peer)
- node_modules/reflect-metadata: 0.2.2 (dev)
+ node_modules/reflect-metadata: 0.2.2 (dev, peer)
- node_modules/rollup: 4.52.5 (dev)
+ node_modules/rollup: 4.52.5 (dev, peer)
- node_modules/rxjs: 7.8.2 (dev)
+ node_modules/rxjs: 7.8.2 (dev, peer)
- node_modules/schema-utils/node_modules/ajv: 8.17.1 (dev)
+ node_modules/schema-utils/node_modules/ajv: 8.17.1 (dev, peer)
- node_modules/ts-node: 10.9.2 (dev)
+ node_modules/ts-node: 10.9.2 (dev, peer)
- node_modules/typescript: 5.0.4 (dev)
+ node_modules/typescript: 5.0.4 (dev, peer)
- node_modules/webpack: 5.102.0 (dev)
+ node_modules/webpack: 5.102.0 (dev, peer)
- node_modules/webpack-cli: 6.0.1 (dev)
+ node_modules/webpack-cli: 6.0.1 (dev, peer)
```


 My guess is that you regenerated the package-lock file using npm@11.6.1 or npm@11.6.2 -- both of which are broken.  The expectation is that this will be fixed by https://github.com/npm/cli/pull/8645 which should be in [the npm 11.6.3 release](https://github.com/npm/cli/pull/8671).
 
 For now, please regenerate the lock file using npm 11.6.0 or earlier.

trentm (changes requested):

package-lock.json
overbalance: ran it again with 11.6.2

trentm:
FYI: `  "rollup": "4.52.5"` was added to devDeps ~at the top-level~ to pin that transitive dep of `@web/dev-server-rollup` (used in instrumentation-document-load), because the current latest rollup 4.x (4.53.0 is broken: https://github.com/rollup/rollup/issues/6168)

trentm:
```
% npm ls rollup --package-lock-only
opentelemetry-js-contrib@0.1.0 /Users/trentm/tm/opentelemetry-js-contrib8
└─┬ @opentelemetry/instrumentation-document-load@0.54.0 -> ./packages/instrumentation-document-load
  ├─┬ @rollup/plugin-commonjs@29.0.0
  │ ├─┬ @rollup/pluginutils@5.3.0
  │ │ └── rollup@4.53.0 deduped
  │ └── rollup@4.53.0
  ├─┬ @rollup/plugin-node-resolve@16.0.3
  │ └── rollup@4.53.0 deduped
  ├─┬ @web/dev-server-rollup@0.6.4
  │ ├─┬ @rollup/plugin-node-resolve@15.3.1
  │ │ └── rollup@4.53.0 deduped
  │ └── rollup@4.53.0 deduped
  └── rollup@4.52.5
```
Still have rollup v4.53.0 lingering in there.



Here is a patch that I believe will fix it:
https://gist.github.com/trentm/466ef202543c862275afdabf2f6e57a4

trentm:
Before:

```
% du -sh .
1.3G	.

% npm ls -a | wc -l
npm error code ELSPROBLEMS
npm error invalid: zod@3.23.8 /Users/trentm/pristine/opentelemetry-js-contrib/node_modules/zod
npm error A complete log of this run can be found in: /Users/trentm/.npm/_logs/2025-11-07T22_23_08_495Z-debug-0.log
    7824
```

After:

```
% du -sh .
1.1G	.

% npm ls -a | wc -l
npm error code ELSPROBLEMS
npm error invalid: ws@8.17.1 /Users/trentm/tm/opentelemetry-js-contrib8/node_modules/ws
npm error invalid: zod@3.23.8 /Users/trentm/tm/opentelemetry-js-contrib8/node_modules/zod
npm error A complete log of this run can be found in: /Users/trentm/.npm/_logs/2025-11-07T22_23_24_355Z-debug-0.log
    7032
```

So 10-20% dev tree size reduction.

I'm not sure what's up with the "npm error invalid"... but FWIW we have this case in "main" now, so I don't necessarily think this should block. I'll poke a bit.

trentm:
Both the `zod` and `ws` npm ls errors are due to `openai` having an optional peer dep on these. From openai/package.json:

```json
  "peerDependencies": {
    "ws": "^8.18.0",
    "zod": "^3.25 || ^4.0"
  },
  "peerDependenciesMeta": {
    "ws": {
      "optional": true
    },
    "zod": {
      "optional": true
    }
  }
```

peer-optional means that the dep is *not* installed by default (i.e. not added to the package-lock tree).

However, because ws and zod *are* transitive deps of other things in the tree we do get versions installed, but not necessarily matching the peer-optional range, e.g.:

```
% npm ls zod
opentelemetry-js-contrib@0.1.0 /Users/trentm/src/opentelemetry-js-contrib
├─┬ @opentelemetry/instrumentation-document-load@0.54.0 -> ./packages/instrumentation-document-load
│ └─┬ @web/test-runner@0.18.3
│   └─┬ @web/test-runner-chrome@0.16.0
│     └─┬ puppeteer-core@22.15.0
│       └─┬ chromium-bidi@0.6.3
│         └── zod@3.23.8 deduped invalid: "^3.25 || ^4.0" from node_modules/openai
└─┬ @opentelemetry/instrumentation-openai@0.6.0 -> ./packages/instrumentation-openai
  └─┬ openai@6.7.0
    └── zod@3.23.8 invalid: "^3.25 || ^4.0" from node_modules/openai

npm error code ELSPROBLEMS
npm error invalid: zod@3.23.8 /Users/trentm/src/opentelemetry-js-contrib/node_modules/zod
```

For now I think we can ignore these. I'm somewhat sure that our openai testing in instrumentation-openai does not use any of the optional zod or ws functionality in `openai`, so we can ignore. If it *did* matter, then we'd want to have devDeps for these in instrumentation-openai.  We *could* do that just to silence the `npm ls` error message if necessary.

trentm:
Another view of the before/after, showing the top few multiple installs of given packages in the tree:

Before:

```
% npm ls -ap | awk -F node_modules/ '{print $NF}' | sort | uniq -c | sort -n | tail -20
  12 rimraf
  12 signal-exit
  12 yallist
  13 hosted-git-info
  13 ms
  13 supports-color
  14 color-convert
  14 color-name
  15 @smithy/node-http-handler
  15 readable-stream
  15 string_decoder
  17 ansi-styles
  20 brace-expansion
  22 glob
  23 debug
  24 semver
  31 minimatch
  38 lru-cache
  69 undici-types
  70 @types/node
```

After:

```
% npm ls -ap | awk -F node_modules/ '{print $NF}' | sort | uniq -c | sort -n | tail -20
  10 @azure/abort-controller
  10 npm-package-arg
  10 path-scurry
  10 type-fest
  10 yargs
  11 ms
  11 rimraf
  11 signal-exit
  12 minipass
  12 yallist
  13 hosted-git-info
  15 @smithy/node-http-handler
  18 string-width
  19 brace-expansion
  19 emoji-regex
  19 semver
  20 glob
  21 debug
  30 minimatch
  37 lru-cache
```

trentm (approved):

PR #3044, opened 2025-09-15
chore(resource-detector-container): update semconv usage to ATTR_ exports

trentm:
Refs: #2377

### DIFF
--- a/packages/instrumentation-document-load/package.json
+++ b/packages/instrumentation-document-load/package.json
@@ -75,8 +75,7 @@
   "dependencies": {
     "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/instrumentation": "^0.205.0",
-    "@opentelemetry/sdk-trace-web": "^2.0.0",
-    "@opentelemetry/semantic-conventions": "^1.27.0"
+    "@opentelemetry/sdk-trace-web": "^2.0.0"
   },
   "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-document-load#readme"
 }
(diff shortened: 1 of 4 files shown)


package-lock.json
trentm, line 1: The large changes here are from re-running `npm install` in the resource-detector-container dir after removing the semconv package dep. I *guess* these were latent package-lock changes from earlier instr-redis and resource-detector-gcp changes. I haven't dug into it.

maryliag: it's a little hard to review if all the changes here make sense. For example, the removal of `gcp-metadata`, because that is still being used in `@opentelemetry/resource-detector-gcp`, so the removal must have come from another place.
I guess as long as tests are passing it should be good? Otherwise I'm not really sure how to review this file

maryliag (approved):
Approving from changes in general, just would be nice to get an extra check in the package-lock file if everything is expected

trentm:
> I guess as long as tests are passing it should be good?

And they are *not* currently. :)

trentm:
test-all-versions tests are failing on redis@3 testing. Given there are redis dep-related changes in the package-lock.json changes here, that seems a likely culprit.

trentm:
I've reverted the package-lock.json changes. FWIW, those changes also come from running `npm install` in the repo (same result when doing this on the "main" branch) **when using npm@11, the npm included with node v24**. These package-lock.json changes are *not* made by npm@10 (the default npm in node v20 and v22).

Poking around a bit, it looks to me like this is a difference in npm's handling of peerDependencies in transitive deps. (It feels like npm@11 is deciding it does *not* need to install some peerDependencies.) There is no mention of peerDependencies behaviour changes in the npm@11.0.0 release notes, so I'm not sure what is going on.

maryliag (approved):

PR #3045, opened 2025-09-15
test(instrumentation-koa): update semconv usage to ATTR_ exports

trentm:
Refs: #2377

### DIFF
--- a/packages/instrumentation-koa/test/koa.test.ts
+++ b/packages/instrumentation-koa/test/koa.test.ts
@@ -27,8 +27,7 @@ import {
 } from '@opentelemetry/sdk-trace-base';
 import {
   ATTR_HTTP_ROUTE,
-  SEMATTRS_EXCEPTION_MESSAGE,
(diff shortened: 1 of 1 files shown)


packages/instrumentation-koa/test/koa.test.ts
```diff
@@ -625,7 +624,7 @@ describe('Koa Instrumentation', function () {
 
       const requestHook = sinon.spy(
         (span: Span, info: KoaRequestInfo<KoaContext, KoaMiddleware>) => {
-          span.setAttribute(SEMATTRS_HTTP_METHOD, info.context.request.method);
+          span.setAttribute('http.method', info.context.request.method);
```
trentm, line 627: Reviewer note: The test case already uses the literal `'http.method'` below this. Using the `ATTR_HTTP_METHOD` export (which is experimental) would require having a local "semconv.ts" copy from the incubating entry-point. I think that is overkill for a single case **in a test file**.

PR #3047, opened 2025-09-16
chore(instrumentation-document-load): update semconv usage to ATTR_ exports

trentm:
Refs: #2377


packages/instrumentation-document-load/src/instrumentation.ts
```diff
@@ -41,10 +41,7 @@ import {
 import { AttributeNames } from './enums/AttributeNames';
 /** @knipignore */
 import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
-import {
-  SEMATTRS_HTTP_URL,
-  SEMATTRS_HTTP_USER_AGENT,
-} from '@opentelemetry/semantic-conventions';
```
maryliag, line 47: looks like you can remove the dependency from package.json for the sem conv package

trentm: Thanks, will do tomorrow.

PR #3049, opened 2025-09-16
chore(instrumentation-aws-lambda): update semconv usage to ATTR_ exports

trentm:
This is the first case of using a src/semconv-obsolete.ts file
(as proposed here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2377#issuecomment-2868036592)
for semconv constants so old they don't have a current value in Semantic
Conventions (e.g. 'faas.id').

Refs: #2377


packages/instrumentation-aws-lambda/src/semconv-obsolete.ts
```diff
@@ -0,0 +1,53 @@
+/*
```
maryliag, line 1: just wondering about the name "obsolete", because it seems to be adding one more nomenclature that could cause confusion "what is the difference between obsolete and deprecated?"
why not use `deprecated`?

trentm: I used `obsolete` to try to find a separate (and slightly stronger) word. `deprecated`, in the context of JS semconv, already has two uses defined here: https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#deprecations

These "obsolete" cases are the rare cases (I think we'll see them in ~4 instrumentations) of values that used to be in Semantic Conventions but were *removed* from the Semantic Conventions schema files -- rather than being *kept, but marked deprecated*. Because they don't exist in the semconv schema files, the `@opentelemetry/semantic-conventions` JS package doesn't export constants for them.

My dictionary defines "deprecated" as:

>  (**be deprecated**) (chiefly of a software feature) be usable but regarded as obsolete and best avoided, typically due to having been superseded: "this feature is deprecated and will be removed in later versions"

I feel there is some small utility in not re-using the same "deprecated" word. Otherwise, I think it is (slightly) confusing to answer why a `@deprecated` constant in "src/semconv.ts" is not in "src/semconv-obsolete.ts", and vice versa.

I originally case across the issue here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2377#issuecomment-2868036592  where I presented other options.

maryliag (approved):

PR #3060, opened 2025-09-16
chore(resource-detector-gcp): update semconv usage to ATTR_ exports

trentm:
Refs: #2377


maryliag (approved):

trentm:
@aabmass There isn't any requirement for keeping this upstreamed resource-detector-gcp somehow in sync with the google repo from which you upstreamed it, is there?

PR #3061, opened 2025-09-16
chore(contrib-test-utils)!: update semconv usage to ATTR_ exports

trentm:
This also drops the `imageTag` option to `assertContainerResource`.
This option is unused by all users of `assertContainerResource` in the
repo. (It dates back to when `@opentelemetry/resources` in the core
repo used contrib-test-utils. It no longer does.)  Dropping this
option allows us to not have to deal with the Semantic Conventions
change of 'container.image.tag' to container.image.tags' (plural)
in semconv v1.22.0.

Refs: #2377

### DIFF
--- a/packages/contrib-test-utils/src/instrumentations/index.ts
+++ b/packages/contrib-test-utils/src/instrumentations/index.ts
@@ -15,7 +15,7 @@
  */
 
 import { resourceFromAttributes } from '@opentelemetry/resources';
-import { SEMRESATTRS_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
+import { ATTR_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
 import { getInstrumentation } from './instrumentation-singleton';
 import { registerInstrumentationTestingProvider } from './otel-default-provider';
 import { resetMemoryExporter } from './otel-provider-api';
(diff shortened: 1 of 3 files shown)


trentm:
The test failure is because of this change in "packages/contrib-test-utils/src/resource-assertions.ts":

```diff
 import * as semconv from '@opentelemetry/semantic-conventions';
...
  * `prefix`. By "known", we mean it is an attribute defined in semconv.
  */
 const assertHasOneLabel = (prefix: string, resource: Resource): void => {
-  const semconvModPrefix = `SEMRESATTRS_${prefix.toUpperCase()}_`;
+  const semconvModPrefix = `ATTR_${prefix.toUpperCase()}_`;
   const knownAttrs: Set<string> = new Set(
     Object.entries(semconv)
       .filter(
```

This test utility function is attempting to look at *all* semantic-conventions constants. The `'@opentelemetry/semantic-conventions` entry point **includes all `SEMRESATTRS_*` constants** but **does NOT include all `ATTR_*`**.  It is the `.../incubating` entry point that includes all the `ATTR_*` constants.

So we want to use the incubating entry point here.
However, that breaks our `npm run lint:semconv-deps`. So we want to add exception for this file: it is allowed to use the incubating entry point. This isn't runtime code -- it is for tests -- so this is fine.

PR #3068, opened 2025-09-18
ci: harden publish workflow by switching to OIDC

pichlermarc:

Similar to https://github.com/open-telemetry/opentelemetry-js/pull/5946

Nothing has happened to us, but with everything that's been happening recently, this PR does a few things to harden our publishing process:

- switches publishing from token-based to OIDC
  - I have already updated all packages to disallow token publish, and set up OIDC for all of the packages published from this repo
 - makes use of a new `npm-publish-environment` environment, to lock down who can actually publish to npm
        until recently access was more broad, but additonal maintainer approval will now be required for publishing here and in contrib
   - requiring the environment and OIDC will also limit a potential blast radius to one repo, we can set up to require multiple more than one maintainer approval before publishing.
- I have moved npm ci to run outside of the deploy environment to further increase complexity of smuggling things in that don't belong here.
- I switched from `opentelemetrybot` to `otelbot` while at it

Tokens that we previously used for publishing are already revoked. ✅

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -8,35 +8,37 @@ permissions:
 
 name: Run Release Please
 jobs:
+  # creates the release PR, and publishes releases
   release-please:
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
     runs-on: ubuntu-latest
-    permissions:
-      contents: write # required for pushing changes
-      id-token: write
     steps:
-      # The logic below handles the npm publication:
       - name: Checkout Repository
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
-      # If you update this version make sure to keep in sync:
-      # - `build-and-cache` job of unit-test workflow
-      # - `build-and-cache` job of test-all-versions workflow
       - name: Setup Node
         uses: actions/setup-node@v5
         with:
-          node-version: 18
+          node-version: 24
           registry-url: 'https://registry.npmjs.org'
 
       - name: Install packages
         run: |
           npm ci
 
+      - uses: actions/create-github-app-token@df432ceedc7162793a195dd1713ff69aefc7379e # v2.0.6
+        id: otelbot-token
+        with:
+          app-id: ${{ vars.OTELBOT_JS_APP_ID }}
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -10100,93 +10100,68 @@
       }
     },
     "node_modules/@redis/bloom": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@redis/bloom/-/bloom-1.2.0.tgz",
-      "integrity": "sha512-HG2DFjYKbpNmVXsa0keLHp/3leGJz1mjh09f2RLGGLQZzSHpkmZWuwJbAvo3QcRY8p80m5+ZdXZdYOSBLlp7Cg==",
+      "version": "5.8.2",
```
pichlermarc: note: this was not on purpose - I'll still need to change it back.

.github/workflows/release-please.yml
```diff
@@ -8,35 +8,37 @@ permissions:
...
           fetch-depth: 0
 
-      # If you update this version make sure to keep in sync:
-      # - `build-and-cache` job of unit-test workflow
-      # - `build-and-cache` job of test-all-versions workflow
       - name: Setup Node
         uses: actions/setup-node@v5
         with:
-          node-version: 18
+          node-version: 24
```
david-luna: Is there a specific reason to this change of version?
The node version should not affect the TS compilation output so I'm fine with this change. I'll update the test workflow to compile with the same version.

pichlermarc: Yes - so the idea is to get the latest NPM. Trusted publishing is available since `npm@11.5.1` so we need to have at least that version. Actually, we could likely also do compilation with `18` and publishing with `24`

pichlermarc: I switched compile back to `18` now but kept publish at `24` 🙂 ([aa21e3f](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3068/commits/aa21e3f19481b3bf448d3f8570441e3cb083a7c9))

package-lock.json
pichlermarc: rolled in back in the latest revision.

PR #3076, opened 2025-09-18
chore(sampler-aws-xray): update semconv usage to modern exports

trentm:
Refs: #2377

### DIFF
--- a/packages/sampler-aws-xray/test/remote-sampler.test.ts
+++ b/packages/sampler-aws-xray/test/remote-sampler.test.ts
@@ -34,10 +34,7 @@ import {
 } from '@opentelemetry/api';
 import { SamplingDecision } from '@opentelemetry/sdk-trace-base';
 import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
-import {
-  SEMRESATTRS_CLOUD_PLATFORM,
-  ATTR_SERVICE_NAME,
-} from '@opentelemetry/semantic-conventions';
+import { ATTR_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
 import { expect } from 'expect';
 import * as nock from 'nock';
 import * as sinon from 'sinon';
@@ -46,6 +43,7 @@ import {
   AWSXRayRemoteSampler,
 } from '../src/remote-sampler';
 import { AWSXRaySamplingClient } from '../src/aws-xray-sampling-client';
+import { ATTR_CLOUD_PLATFORM } from '../src/semconv';
 
 const DATA_DIR_SAMPLING_RULES =
   __dirname + '/data/test-remote-sampler_sampling-rules-response-sample.json';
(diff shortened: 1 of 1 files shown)


packages/sampler-aws-xray/src/sampling-rule-applier.ts
```diff
@@ -39,6 +39,7 @@ import {
   ATTR_URL_FULL,
   ATTR_URL_PATH,
   ATTR_SERVICE_NAME,
+  // XXX can we drop this? (else replace it with semconv-obsolete.ts usage)
   SEMRESATTRS_FAAS_ID,
```
trentm: @jj22ee Hi. I have a question about the `faas.id` usage here.

Background: I've been working on removing all of the usages of the [old/deprecated exports](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions/#migrating-from-semattrs_-semresattrs_-) from the `@opentelemetry/semantic-conventions` package, which includes all of the `SEMRESATTRS_*` exports -- as part of https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2377.
This usage is one of the last ones.

`faas.id` was replaced with `cloud.resource_id` in SemConv 1.19 in 2023: https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md#v1190-2023-03-06

What is an expected case where `getLambdaArn()` (below) will encounter a resource with the `faas.id` attribute?  I'm not challenging, just trying to understand how this sampler would be used.

`instrumentation-aws-lambda` *does* still use the old `faas.id` (https://github.com/open-telemetry/opentelemetry-js-contrib/blob/febf30e66adcceb20dd2fa3c031630c28c61c3fa/packages/instrumentation-aws-lambda/src/instrumentation.ts#L233-L239) though it sets taht as a **span attribute** and not as a resource attribute.

jj22ee: TLDR: `SEMRESATTRS_FAAS_ID` can [be dropped](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e5adbf1d159783d7f969de5085ba4a14040554ef/packages/sampler-aws-xray/src/sampling-rule-applier.ts#L263), doesn't change the experience for users of X-Ray Sampler.

-------------

Today there isn't an expected case where `getLambdaArn()` would encounter a resource with `faas.id` because lambda instrumentation or resource-detector doesn't set it, and even if they did, it doesn't matter for X-Ray Sampler today because:

- One of the requirements for a Sampling Rule to be used in X-Ray Sampler is if that [Sampling Rule's Resource ARN matches matches the ARN extracted from the resource attributes](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e5adbf1d159783d7f969de5085ba4a14040554ef/packages/sampler-aws-xray/src/sampling-rule-applier.ts#L187), which could include `getLambdaArn()`.

  - Today, all Sampling Rule's Resource ARN is hardcoded to be the `*` match-all wildcard, and cannot be configured by the user (except for Sampling Rules used for AWS APIGW).
    https://docs.aws.amazon.com/xray/latest/devguide/xray-console-sampling.html#xray-console-sampling-options

So effectively, the result of `getLambdaArn()` doesn't change anything, as all sampling rules will match its result because of the `*` wildcard match. The check for `faas.id` in `getLambdaArn()` was intended for future-proofing in case X-Ray does want to support letting users match Sampling Rule's Resource ARN to `getLambdaArn()`.

trentm: dropped in commit 9d0f303

PR #3077, opened 2025-09-18
feat(instrumentation-aws-sdk): support migration to stable HTTP semconv

trentm:
This adds support for using the OTEL_SEMCONV_STABILITY_OPT_IN envvar
to control migration to the stable HTTP semantic conventions for
telemetry from this instrumentation. Note that the only
impacted span attribute is the one for the HTTP response status code.


packages/instrumentation-aws-sdk/test/load-instrumentation.ts
```diff
@@ -30,6 +30,11 @@ import {
 } from '@opentelemetry/sdk-metrics';
 import { AwsInstrumentation } from '../src';
 
+// This is a meagre testing of just a single value of
+// OTEL_SEMCONV_STABILITY_OPT_IN, because testing multiple configurations of
+// `AwsInstrumentation` in this all-in-one-process test setup is difficult.
+process.env.OTEL_SEMCONV_STABILITY_OPT_IN = 'http/dup';
+
```
trentm, line 37: Reviewer note: I realize I am not testing all the values of `OTEL_SEMCONV_STABILITY_OPT_IN` in the tests here. My plea is that it is not worth the effort:

- Currently the tests run all in one process, and all the test cases share one instance of AwsInstrumentation (from "test/load-instrumentation.ts"). Loading a second instrumentation (to test separate configurations of it) is generally fraught.
- This instrumentation makes very minimal use of the `...OPT_IN` envvar: just for a single HTTP-related attribute.
- After ~6 months we will consider the HTTP migration to be complete and will be able to remove this `...OPT_IN` handling.

maryliag: I see you chose the option to send both, so should be good enough to check if both ways are being generated, but you didn't update the test to confirm the stable values are being sent, like you did on the other file. Is there nothing here to check to confirm the stable value is also being send?

trentm: > but you didn't update the test to confirm the stable values are being sent

@maryliag I *did* add this check: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3077/files#diff-72b8a42053cbffbe6d2b66205d55265dcabd2d25ac3bb0bafee1e8fe7ce9e155R81-R85
So that's one place.

There *are* a number of other test locations that check `ATTR_HTTP_STATUS_CODE` where I could also add a check for `ATTR_HTTP_RESPONSE_STATUS_CODE`. Is that what you mean?

maryliag: Yes, precisely because I saw you added the check on the other file, but I didn't see any check in this particular file, so I was wondering if there are other scenarios worth checking here that the check you added in the other file would not cover.
If you think that check is enough, then it's okay, I was mostly surprise with the addition of the flag here but no other file changes

trentm: This file (load-instrumentation.ts) is the utility file that is used by all the other actual test files, this one doesn't have assertions.

However, I can add to the other places:

```
aws-sdk-v3-s3.test.ts
39:  ATTR_HTTP_STATUS_CODE,
79:      expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(200);
105:        expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(200);
134:      expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(200);
168:        expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(403);

aws-sdk-v3-sqs.test.ts
35:  ATTR_HTTP_STATUS_CODE,
97:    expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(200);
147:    expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(200);
180:      expect(span.attributes[ATTR_HTTP_STATUS_CODE]).toEqual(200);
```

trentm: Commit 71a26a20.
Good call. I had missed one place.

maryliag (approved):

PR #3081, opened 2025-09-22
ci: ensure EasyCLA check passes on release PR

pichlermarc:

See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3072 - the js-contrib otelbot does not have easycla signed, so the check is failing. However, we do push a follow-up commit as otelbot already, which works. This PR makes it so that all changes are made in one go via the otelbot instead. 

That should result in 
- only one commit being in the PR
- that commit having the EasyCLA agreement signed


.github/workflows/release-please.yml
```diff
@@ -51,17 +51,19 @@ jobs:
...
         run: |
+          git reset HEAD~1
           npm install --ignore-scripts --package-lock-only
-          git add package-lock.json
+          git add .
           git config user.name otelbot
           git config user.email 197425009+otelbot@users.noreply.github.com
-          git commit -m "chore: sync package-lock.json"
-          git push
+          git commit -m "chore: release main"
```
legendecas: Maybe use `git commit --amend --reset-author` to avoid `git reset`?

pichlermarc: yep, that's better - thanks 🙂  applied in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3081/commits/a760bda2b590664b9e1dc103c6465584c5244207

.github/workflows/release-please.yml
```diff
@@ -60,8 +61,8 @@ jobs:
           git add package-lock.json
           git config user.name otelbot
           git config user.email 197425009+otelbot@users.noreply.github.com
-          git commit -m "chore: sync package-lock.json"
-          git push
+          git commit --amend --reset-author
```
legendecas: Sorry, I should have suggested with editor suppressed.

```suggestion
          git commit --amend --reset-author --no-edit
```

pichlermarc: All good, thanks for the suggestion 🙂 
I usually use my git commands all complicated so my initial approach was kinda awkward. 😅 This is better.

trentm (approved):

PR #3095, opened 2025-09-24
fix: maximum call stack exceeded in `instrumentation-mongodb`

### LINKED ISSUE #2281: Circular reference in _scrubStatement causes maximum call stack exceeded in instrumentation-mongodb

rndD:
see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3131

Fixes #2281

* Added unit tests for `_defaultDbStatementSerializer` to reproduce the problem
* Replaced `_scrubStatement` with a code provided by @masontikhonov, https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2281#issuecomment-3325811121 - instead of implementing recursive scruber, we use `JSON.stringify` with `replacer` function to get serialized DbStatement and drop circular refs.

### DIFF
--- a/packages/instrumentation-mongodb/src/instrumentation.ts
+++ b/packages/instrumentation-mongodb/src/instrumentation.ts
@@ -950,6 +950,19 @@ export class MongoDBInstrumentation extends InstrumentationBase<MongoDBInstrumen
     );
   }
 
+  private _getDefaultDbStatementReplacer(): Replacer {
+    const seen = new WeakSet();
+    return (_key, value) => {
+      // undefined, boolean, number, bigint, string, symbol, function || null
+      if (typeof value !== 'object' || !value) return '?';
+
+      // objects (including arrays)
+      if (seen.has(value)) return '[Circular]';
+      seen.add(value);
+      return value;
+    };
+  }
+
   private _defaultDbStatementSerializer(commandObj: Record<string, unknown>) {
     const { enhancedDatabaseReporting } = this.getConfig();
 
@@ -957,21 +970,7 @@ export class MongoDBInstrumentation extends InstrumentationBase<MongoDBInstrumen
       return JSON.stringify(commandObj);
     }
 
-    const getReplacer = (): Replacer => {
-      // WeakSet is used for tracking seen objects to avoid memory leaks
-      const seen = new WeakSet();
-      return (_key, value) => {
-        // undefined, boolean, number, bigint, string, symbol, function || null
-        if (typeof value !== 'object' || !value) return '?';
-
-        // objects (including arrays)
-        if (seen.has(value)) return '[Circular]';
-        seen.add(value);
-        return value;
-      };
-    };
-
-    return JSON.stringify(commandObj, getReplacer());
+    return JSON.stringify(commandObj, this._getDefaultDbStatementReplacer());
   }
 
   /**


packages/instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -952,28 +952,26 @@ export class MongoDBInstrumentation extends InstrumentationBase<MongoDBInstrumen
...
 
-    if (typeof value === 'object' && value !== null) {
-      return Object.fromEntries(
-        Object.entries(value).map(([key, element]) => [
-          key,
-          this._scrubStatement(element),
-        ])
-      );
-    }
+    const getReplacer = (): Replacer => {
```
masontikhonov: Micro-optimization: I would hoist it as a private method of current class, so that it is not re-declared on each call to `_defaultDbStatementSerializer()`

rndD: yeah, let's go

masontikhonov (approved):
I'm not the right person to approve, but it looks good to me.
Thanks for adding extensive tests! 🙌

AdivA-PMG (approved):

rndD:
Hi, @blumamir! Maybe you can take a look if you have time.

packages/instrumentation-mongodb/package.json
```diff
@@ -19,7 +19,8 @@
     "compile": "tsc -p .",
     "prepublishOnly": "npm run compile",
     "tdd": "npm run test-v5-v6-run -- --watch-extensions ts --watch",
-    "test": "npm run test-v5-v6",
+    "test": "npm run test-v5-v6 && npm run test-unit",
+    "test-unit": "nyc --no-clean mocha --require '@opentelemetry/contrib-test-utils' 'test/**/unit/*.test.ts'",
```
pichlermarc, line 23: nit: why not add this to the `test-v5-v6` script's filter instead? 🤔

packages/instrumentation-mongodb/src/types.ts
```diff
@@ -74,3 +74,8 @@ export enum MongodbCommandType {
   COUNT = 'count',
   UNKNOWN = 'unknown',
 }
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer
+ */
+export type Replacer = (key: string, value: unknown) => unknown;
```
pichlermarc, line 81: looks like this is not used in the public API, so this can live in `internal-type.ts` instead. 🙂

pichlermarc (commented):
Looks solid - thanks for working on this 🙂 
Just one nit around the extra test script, and one on the location of the `Replacer` type.

Also pining @onurtemizkan for review (who is the component owner for this package) 🙂 My knowledge about this package is quite superficial.

packages/instrumentation-mongodb/src/types.ts
rndD: sure

packages/instrumentation-mongodb/package.json
rndD: I can, but `test-v5-v6` are not really unit tests, we start real mongodb there and run international tests with different driver version.
What I added feels like a different type of tests, also this code is diver version independent.

rndD:
Sorry, removed branch in my fork by accident.

rndD:
Reopened in #3131

PR #3096, opened 2025-09-24
ci(release-please): use releases_created instead of release_created

pichlermarc:

The job to actually publish the release to npm did not run due to a missing `s` in the output. 
- `releases_created` -> at least one was created. 
- `release_created` -> top-level release was created (which is just the top-level package.json, for us which is set to `private`)


pichlermarc:
Due to this typo, GitHub releases were created - but there were no packages actually published to npm.

I guess updating the tag to a different commit would be mighty sketchy, so I'll do this:
- I'll put a commit override message that bumps the patch versions of each of the non-released packages once again in my fix PR .
- Once actually published: I'll update the broken GitHub releases to state all changes are included in the next version

Issue #3099, opened 2025-09-24
[ci] GitHub releases published, but no packages published to npm


pichlermarc:
We have recently updated our publish workflows to OIDC. However, due to a typo in the new workflow (resolved in #3096), we've created GitHub releases without actually publishing to `npmjs.org` 

I have opened a PR to force a new release #3098 - once that is merged, we should be able to re-run the workflow and publish new releases that should then **actually** be published to `npmjs.org`. 

I'll update the unintentionally created GitHub releases to state that the next patch release is the actual version on `npmjs.org`

pichlermarc:
After a few workflow changes, we attempted to publish again, but we had another problem that ended up in the same situation - when downloading the artifact, which is a .zip file, file permissions on some scripts got lost and that prompted `lerna` to refuse to publish to npm.

https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3107#issuecomment-3335566262

pichlermarc:
We were able to publish the packages, but there's been a **mismatch in the most recent GitHub releases and packages published to `npmjs.com`.** That's because the GitHub releases were already published, and versions in `package.json` were already incremented. However, they were not published on their commit due to the recent changes in the publish workflow.

A GitHub release exists for each of the published packages, but the commit on the releases may not match the provenance statement on the package in npm.

**Commit only matches for three packages:**
- `@opentelemetry/auto-instrumentations-node` (published via OIDC, correct provenance statement)
- `@opentelemetry/instrumentation-bunyan` (published via OIDC, correct provenance statement)
- `@opentelemetry/sampler-aws-xray` (this is a new package and was not published via OIDC. It does not have a provenance statement. I had to publish it manually from my own machine using 2FA for the first time. It's built from https://github.com/open-telemetry/opentelemetry-js-contrib/commit/931c7b34f53ea625da900726b1f57c5c934b5b28. Subsequent releases will contain a provenance statement and will be published via OIDC)

**All other packages** have provenance statements pointing to https://github.com/open-telemetry/opentelemetry-js-contrib/commit/931c7b34f53ea625da900726b1f57c5c934b5b28, however the release points to https://github.com/open-telemetry/opentelemetry-js-contrib/commit/c7f4e6c08131d6ac97eaa4cac5dcf20905ccb6a2

[...]

PR #3103, opened 2025-09-25
ci(release-please): pass EasyCLA, correct if-logic for releases_created var

trentm:
- The release-please PR can include *multiple* commits, so we need to amend
  the author on all of them to pass EasyCLA.
- The logic for guarding on releases_created was wrong: I believe because
  it is a *string* true or false, rather than a boolean that works the
  way it was previously.

Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3099

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -64,7 +64,7 @@ jobs:
           npm install --ignore-scripts --package-lock-only
           git add package-lock.json
           git commit -m "chore: sync package-lock.json 'dependencies' key"
-          git push
+          git push --force
 
   install-and-compile:
     needs: release-please
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -52,23 +52,24 @@ jobs:
           token: ${{ steps.otelbot-token.outputs.token }}
           fetch-depth: 0
 
-      # previous commiter does not have CLA signed, and the commit is missing updates to package-lock.json,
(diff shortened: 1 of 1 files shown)


trentm:
^^ check failures are all `docker pull ...`, I assume it is because of https://www.dockerstatus.com/pages/incident/533c6539221ae15e3f000031/68d47a2f93c09e05486d93a9

.github/workflows/release-please.yml
```diff
@@ -52,23 +52,24 @@ jobs:
...
-          git add package-lock.json
           git config user.name otelbot
           git config user.email 197425009+otelbot@users.noreply.github.com
-          git commit --amend --reset-author --no-edit
-          git push --force
+          git rebase main -x "git commit --amend --reset-author --no-edit"
+          npm install --ignore-scripts --package-lock-only
+          git add package-lock.json
+          git commit -m "chore: sync package-lock.json 'dependencies' key"
+          git push
```
pichlermarc: since we still re-write the history.
```suggestion
          git push --force
```

pichlermarc (approved):
Thanks 🙂 

For posterity: https://github.com/googleapis/release-please-action/issues/912 - is the behavior that requires the updates to the conditions

.github/workflows/release-please.yml
trentm: ah, right, thanks.

PR #3121, opened 2025-09-29
ci: use --ignore-scripts wherever possible

pichlermarc:

I have checked some postinstall scripts, but they seem to mostly be for for warning users of improper usage and can therefore be safely ignored in CI. Ignoring the scripts also reduces attack surface a bit further.


pichlermarc:
Follow-up: #3122 - looks like `knex` actually needs scripts to generate bindings on install. That does only affect runtime behavior though, so I allowed scripts for the test run only.

PR #3125, opened 2025-09-29
chore(instrumentation-undici): update metric names to semconv `METRIC_*` exports

trentm:
Refs: https://github.com/open-telemetry/opentelemetry-js/issues/5956


trentm:
FWIW, I opted *not* to update the usage of that export name in the test files. I'm easy either way.

Issue #2953, opened 2025-07-23
Implement stable DB Semantic Conventions for all DB Plugins


maryliag:
Add environment variable option OTEL_SEMCONV_STABILITY_OPT_IN that includes database and database/dup options as recommended in the [Database Migration Plan](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md).

database - emit the stable database conventions, and stop emitting the old database conventions that the instrumentation emitted previously.
database/dup - emit both the old and the stable database conventions, allowing for a phased rollout of the stable semantic conventions.
The default behavior (in the absence of one of these values) is to continue emitting whatever version of the old database conventions the instrumentation was emitting previously.

Existing packages to be implemented:

- [ ] Postgres
- [ ] MySQL
- [ ] MySQL2
- [ ] MongoDB
- [ ] Redis
- [ ] Oracle
- [ ] Cassandra
- [ ] Knex
- [ ] Memcached
- [ ] Tedious
- [ ] TypeORM

trentm:
I believe all the DB-y instrumentations support OTEL_SEMCONV_STABILITY_OPT_IN.
See the tracking issue for some final follow-ups: https://github.com/open-telemetry/opentelemetry-js/issues/4572

PR #3137, opened 2025-10-02
feat(instrumentation-mysql2): support `net.*` and database semconv migration

trentm:

---

Note: I've also started a user doc for this migration here: https://github.com/open-telemetry/opentelemetry-js/pull/5982

### DIFF
--- a/packages/instrumentation-mysql2/README.md
+++ b/packages/instrumentation-mysql2/README.md
@@ -54,19 +54,28 @@ You can set the following instrumentation options:
 
 ## Semantic Conventions
 
-This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+This instrumentation implements Semantic Conventions (semconv) v1.7.0. Since then, networking (in semconv v1.23.1) and database (in semconv v1.33.0) semantic conventions were stabilized. As of `@opentelemetry/instrumentation-mysql2@0.53.0` support has been added for migrating to the stable semantic conventions using the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable as follows:
+
+1. Upgrade to the latest version of this instrumentation package.
+2. Set `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup,database/dup` to emit both old and stable semantic conventions. (The `http` token is used to control the `net.*` attributes, the `database` token to control to `db.*` attributes.)
+3. Modify alerts, dashboards, metrics, and other processes in your Observability system to use the stable semantic conventions.
+4. Set `OTEL_SEMCONV_STABILITY_OPT_IN=http,database` to emit only the stable semantic conventions.
+
+See [the HTTP migration guide](https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/) and the [database migration guide](https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/) for details.
 
 Attributes collected:
 
-| Attribute               | Short Description                                                              |
-| ----------------------- | ------------------------------------------------------------------------------ |
-| `db.connection_string`  | The connection string used to connect to the database.                         |
-| `db.name`               | This attribute is used to report the name of the database being accessed.      |
-| `db.statement`          | The database statement being executed.                                         |
-| `db.system`             | An identifier for the database management system (DBMS) product being used.    |
-| `db.user`               | Username for accessing the database.                                           |
-| `net.peer.name`         | Remote hostname or similar.                                                    |
-| `net.peer.port`         | Remote port number.                                                            |
+| Old semconv            | Stable semconv   | Description |
+| ---------------------- | ---------------- | ----------- |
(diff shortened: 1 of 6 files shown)


packages/instrumentation-mysql2/test/mysql.test.ts
```diff
@@ -36,6 +36,8 @@ import {
 import * as assert from 'assert';
 import { MySQL2Instrumentation, MySQL2InstrumentationConfig } from '../src';
 
+process.env.OTEL_SEMCONV_STABILITY_OPT_IN = 'http/dup,database/dup';
+
```
raphael-theriault-swi, line 40: Would it be possible to add test cases for non-dup cases ?

raphael-theriault-swi (commented):
Thank you for picking this one up ! FWIW I'm not against merging this as is given the dup attribute should exercise both paths, but the `pg` PR did test both and I've been referring to that one for the review. Don't have anything else to nitpick on apart from that.

raphael-theriault-swi:
cc @maryliag

packages/instrumentation-mysql2/test/mysql.test.ts
maryliag: +1 I think is important, so we know we're not sending extra telemetry the user doesn't want, therefore costing more (resource/money)

An example on how I did for pg:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-pg/test/pg-pool.test.ts#L994

packages/instrumentation-mysql2/README.md
```diff
@@ -54,19 +54,28 @@ You can set the following instrumentation options:
...
 ## Semantic Conventions
 
-This package uses `@opentelemetry/semantic-conventions` version `1.22+`, which implements Semantic Convention [Version 1.7.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/semantic_conventions/README.md)
+This instrumentation implements Semantic Conventions (semconv) v1.7.0. Since then, networking (in semconv v1.23.1) and database (in semconv v1.33.0) semantic conventions were stabilized. As of `@opentelemetry/instrumentation-mysql2@0.53.0` support has been added for migrating to the stable semantic conventions using the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable as follows:
+
+1. Upgrade to the latest version of this instrumentation package.
+2. Set `OTEL_SEMCONV_STABILITY_OPT_IN=http/dup,database/dup` to emit both old and stable semantic conventions. (The `http` token is used to control the `net.*` attributes, the `database` token to control to `db.*` attributes.)
+3. Modify alerts, dashboards, metrics, and other processes in your Observability system to use the stable semantic conventions.
+4. Set `OTEL_SEMCONV_STABILITY_OPT_IN=http,database` to emit only the stable semantic conventions.
+
```
maryliag, line 63:
```suggestion

By default, if `OTEL_SEMCONV_STABILITY_OPT_IN` includes neither of the above tokens, the old v1.7.0 semconv is used.
The intent is to provide an approximate 6 month time window for users of this instrumentation to migrate to the new database and networking semconv, after which a new minor version will use the new semconv by default and drop support for the old semconv.
```

I added this to the PG one, to make things clear there is a timeframe for the update and what to expect if you don't set the env variable

packages/instrumentation-mysql2/README.md
```diff
@@ -54,19 +54,28 @@ You can set the following instrumentation options:
...
-| `db.system`             | An identifier for the database management system (DBMS) product being used.    |
-| `db.user`               | Username for accessing the database.                                           |
-| `net.peer.name`         | Remote hostname or similar.                                                    |
-| `net.peer.port`         | Remote port number.                                                            |
+| Old semconv            | Stable semconv   | Description |
+| ---------------------- | ---------------- | ----------- |
+| `db.connection_string` | Removed          | The connection string used to connect to the database. |
+| `db.name`              | Removed, integrated into the new `db.namespace` | The name of the database. |
+| `db.system`            | `db.system.name` | 'mysql' |
+| `db.statement`         | `db.query.text`  | The database query. |
```
maryliag:
```suggestion
| `db.statement`         | `db.query.text`  | The database query being executed. |
```

maryliag (commented):
Changes look good. Just comments on the README and suggestion for the test

packages/instrumentation-mysql2/test/mysql.test.ts
trentm: commit 5f1bbe0e

packages/instrumentation-mysql2/README.md
trentm: Actually on a re-read of the migration guide https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/ (same for the http guide):

> May drop the environment variable in their next major version and emit only the stable database conventions.

I think we should consider only dropping these in OTel JS SDK v3... which we'll do to drop Node.js 18 support (and likely Node.js 20 support) about Jun 2026.  I guess, given all these instrumentations are 0.x it *will* be a "minor" version, but *breaking* change, so [shrug/].

trentm:
Sorry for all the noise. I screwed up a git merge and ... temporarily made this branch include re-written changes of a dozen commits that had already been merged to main.

Issue #3082, opened 2025-09-22
I use pipeline cannot right show


seth-shi:
### What happened?

'''
const pipe =redis.multi()
pipe.get
pipe.get
pipe.exec()
'''

show two get. not exec

### OpenTelemetry Setup Code

```JavaScript

```

### package.json

```JSON

```

### Relevant log output

```shell

```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. [Learn more here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

pichlermarc:
Transferred to contrib since this seems to be about the redis instrumentation. 
Details on this ticket are.. sparse, though.

seth-shi:
ok, wait me a second, i will show the detail

seth-shi:
<details>

<summary>click here expand the code</summary>

### code

```typescript
import { NodeSDK } from '@opentelemetry/sdk-node';
import { ConsoleSpanExporter, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-node';
import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node';



async function main() {
    const sdk = new NodeSDK({
        spanProcessors: [new SimpleSpanProcessor(new ConsoleSpanExporter())],
        instrumentations: [
            getNodeAutoInstrumentations({
                '@opentelemetry/instrumentation-ioredis': { enabled: true },
                '@opentelemetry/instrumentation-redis': { enabled: true },
            }),
        ],
    });
    await sdk.start();

    const { default: IORedis } = await import('ioredis');
    const { createClient } = await import('redis');

    // --- ioredis ---
    const redis1 = new IORedis({ host: '127.0.0.1', port: 6379 });
    await redis1.set('key1', 'value-from-ioredis');

    // pipeline with ioredis
    const pipe1 = redis1.pipeline();
    pipe1.get('key1');
    pipe1.set('key2', 'v2');
    pipe1.get('key2');
    const res1 = await pipe1.exec();
    console.log('ioredis pipeline result:', res1);

    // --- node-redis ---
    const redis2 = createClient({ url: 'redis://127.0.0.1:6379' });
    await redis2.connect();
    await redis2.set('keyA', 'value-from-node-redis');

[...]

PR #3140, opened 2025-10-03
fix(redis): correct span handling for multi commands

aryamohanan:

This PR fixes an issue in the Redis instrumentation for `MULTI` (transaction) commands where spans were not correctly handled for mixed command sets.

Previously, the `MULTI` span did not follow the semantic conventions and could not accurately represent it in the span data

According to [this part of the semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/4af2ac1934023e25c319895a7c865e4f6fca601f/docs/database/redis.md?plain=1#L47-L48):

> **`db.operation.name`**: It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization. For transactions and pipelined calls, if the individual operations are known to have the same command then that command SHOULD be used prepended by `MULTI` or `PIPELINE`. Otherwise `db.operation.name` SHOULD be `MULTI` or `PIPELINE`.

* Updated the `db.operation.name` on the `MULTI` span:

  * Set to **`MULTI GET`** when all operations are `GET`.
  * Set to **`MULTI`** when commands are mixed, in compliance with the semantic conventions.
* Added support for **`execAsPipeline`**, so that when `MULTI` is executed with `execAsPipeline`, it is treated as a pipeline in Redis instrumentation.
* Expanded tests to cover multiple command scenarios.

**Related issue:** [#3082](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3082)


aryamohanan:
@pichlermarc Could you please take a look at this PR?

I’ve updated the Redis instrumentation to handle MULTI and pipeline spans according to the semantic conventions. Specifically:

- Updated the db.operation.name on the MULTI span:

  -   MULTI GET when all operations are GET.
  
  -   MULTI when commands are mixed.

- Added support for execAsPipeline, so that when MULTI is executed with execAsPipeline, it is treated as a pipeline in Redis instrumentation.

aryamohanan:
@blumamir Could you please take a look at this PR?

packages/instrumentation-redis/src/v4-v5/instrumentation.ts
```diff
@@ -472,27 +480,40 @@ export class RedisInstrumentationV4_V5 extends InstrumentationBase<RedisInstrume
...
     for (let i = 0; i < openSpans.length; i++) {
-      const { span, commandName, commandArgs } = openSpans[i];
+      const { span, commandArgs } = openSpans[i];
       const currCommandRes = replies[i];
       const [res, err] =
         currCommandRes instanceof Error
           ? [null, currCommandRes]
           : [currCommandRes, undefined];
-      this._endSpanWithResponse(span, commandName, commandArgs, res, err);
+      span.setAttribute(ATTR_DB_OPERATION_NAME, operationName);
```
pichlermarc: Q: hmm, should this only be on the span when `this._semconvStability & SemconvStability.STABLE`? It looks like this is part of the "new" stable semconv :)

aryamohanan: Thank you for pointing that out. I’ve included the check.

PR #3141, opened 2025-10-03
feat(tedious): context propagation to SQL Server

mhennoch:
Calls `set context_info` before methods that actually execute user's SQL(omitting prepare and execBulkLoad). Relevant [specification](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/sql-server.md#set-context_info).  This is opt in as it causes an extra round trip per request.

### DIFF
--- a/packages/instrumentation-tedious/src/instrumentation.ts
+++ b/packages/instrumentation-tedious/src/instrumentation.ts
@@ -40,6 +40,11 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 const CURRENT_DATABASE = Symbol(
   'opentelemetry.instrumentation-tedious.current-database'
 );
+
+export const INJECTED_CTX = Symbol(
+  'opentelemetry.instrumentation-tedious.context-info-injected'
+);
+
 const PATCHED_METHODS = [
   'callProcedure',
   'execSql',
@@ -89,7 +94,7 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
             this._wrap(
               ConnectionPrototype,
               method,
-              this._patchQuery(method) as any
+              this._patchQuery(method, moduleExports) as any
             );
           }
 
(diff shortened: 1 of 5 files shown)


packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -207,12 +266,28 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
...
+            ...arguments
+          );
+        };
+
+        const cfg = thisPlugin.getConfig();
+        const shouldInject =
+          cfg.enableTraceContextPropagation &&
+          thisPlugin._shouldInjectFor(operation);
+
+        if (!shouldInject) return runUserRequest();
```
trentm, line 279: The previous code and this code path will `return` the return value of the `originalMethod()` call, which is typically what is wanted when wrapping.  The `_injectContextInfo()` code path below this will *not*.  I gather that won't matter for this instrumentation because, at least currently, every patched method's return type is `void`.

packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -207,12 +266,28 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
...
+
+        const traceparent = thisPlugin._buildTraceparent(span);
+        thisPlugin
+          ._injectContextInfo(this, tediousModule, traceparent)
+          .finally(runUserRequest);
```
trentm: There is a lint error to deal with here:

```
  286:9   error    Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
```

I *think* you could do this (untested):

```
        void thisPlugin
          ._injectContextInfo(this, tediousModule, traceparent)
          .finally(runUserRequest);
```

packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -207,12 +266,28 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
...
+
```
trentm: Separate question on the async behaviour here:
- Without calling `set context_info`, a patched `conn.execSql()` will synchronously call the original `execSql` method before returning.
- With the changes here and when `set context_info` is enabled, that is no longer true. Now, the original `execSql` will be called *sometime later*.

If I have JS code that does something like:

```js
conn.execSql(sql1)
conn.execBulkLoad(sql2)
```

**isn't there a sequencing problem here**?  I don't know tedious internals at all (e.g. whether and how it is queuing things).  But with the example above, is it possible that the actual sequence of SQL executed is the following?

1. `set context_info`
2. `sql2`
3. `sql1`

packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -127,11 +132,65 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
...
   }
 
-  private _patchQuery(operation: string) {
+  private _buildTraceparent(span: api.Span): string {
+    const sc = span.spanContext();
+    const sampled =
+      (sc.traceFlags & api.TraceFlags.SAMPLED) === api.TraceFlags.SAMPLED
+        ? '01'
+        : '00';
+    return `00-${sc.traceId}-${sc.spanId}-${sampled}`;
```
trentm: To be an absolutely correct `traceparent`, with Trace Context Level 2, there are possibly other trace-flags to consider: https://www.w3.org/TR/trace-context-2/#trace-flags

I don't know if true, but I would hope there is some utility for creating the traceparent string from a SpanContext.   Also, it looks like OTel JS doesn't yet implement the `random-trace-id` flag.   So this code is fine *now*, but could get surprised later.

trentm: I didn't find a utility, but this building of the traceparent string is a little more future proof: https://github.com/open-telemetry/opentelemetry-js/blob/a2e3f4542f9a6a61c569d4f70fa49220da18cda0/packages/opentelemetry-core/src/trace/W3CTraceContextPropagator.ts#L84-L86

Perhaps use that handling of the traceFlags.

packages/instrumentation-tedious/src/instrumentation.ts
mhennoch: This throws an error in Tedious.  [Docs](http://tediousjs.github.io/tedious/api-connection.html) say: Only one request at a time may be executed on a connection. Once a Request has been initiated (with callProcedure, execSql, or execSqlBatch), another should not be initiated until the Request's completion callback is called.

trentm: Okay, after playing a bit, I get it now.

packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -127,11 +132,61 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
...
+   * connection. Marks the request with INJECTED_CTX so our patch skips it.
+   */
+  private _injectContextInfo(
+    connection: any,
+    tediousModule: typeof tedious,
+    traceparent: string
+  ): Promise<void> {
+    return new Promise(resolve => {
+      try {
+        const sql = 'set context_info @opentelemetry_traceparent';
```
trentm, line 151: The spec (https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/sql-server.md#set-context_info) suggests using `@traceparent` for the parameter name. Does that matter?

(If so, then I gather the .NET impl should change as well.)

mhennoch: I don't think it does

trentm: Okay, cool.

Issue #3136, opened 2025-10-02
Review on latest CI changes and how to move on


david-luna:
- [x] This only affects the JavaScript OpenTelemetry library
- [ ] This may affect other libraries, but I would like to get opinions here first

This is a follow up issue of the short analysis shared in the JS SIG. Recap:

- we got similar compilation times since most of the times we have to compile and test `auto-instrumentations-node`
  - this package depends on all node instrumentations so it needs to compile all of them
- we got faster builds when only code is changed
  - CI runs the unit and TAV test only for the affected package and auto-instrumentations
- when updating dependencies we usually got slower times
  - because `package-lock.json` is modified it runs unit and TAV for all

Diff with prev CI
- new CI uploads a report per package tested (more time)
- prev CI was doing TAV based on `pkg:instrumentation-*` labels so it was testing. That's where the mos time is saved.

Possible improvements
- create a build cache of last commit in `main`. Will help to reduce compilation time.
- go back to TAV testing based on labels
  - [@pichlermarc] and trigger TAV for all if a certain label is set

I've created this issue so anyone can discuss about the items on the list an add any idea that may help to speed up the review process.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

david-luna:
some discussion about compile time https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3162#issuecomment-3380098877

Options:
- compile cache in main to keep compiling affected packages
- remove `auto-instrumentations-node` and `auto-instrumentations-web` from the affected list

pichlermarc:
> - compile cache in main to keep compiling affected packages

I think this is the option that I'd prefer. Given that `auto-instrumentaions-*` public API can change with instrumentation package changes, this seems like the safest option to me. 🙂

david-luna:
Update on work in compilation speed. Picking the right subset of packages to compile and test too complex (at least for me) because it depends on a couple of vars: package dependencies and task dependencies. I'll revisit the cache because there is a way to clear cache entries although the action does not offer it. See https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3208#issuecomment-3617257330

david-luna:
[...]

PR #3162, opened 2025-10-07
test: separate the unit-test and test-all-versions jobs in the main Test workflow

trentm:
IMHO the "unit tests" (i.e. running `npm test` for each affected package) and the test-all-versions ("TAV") tests (running tav, `npm run test-all-versions`, for affected packages) have slightly different requirements:

- We *require* that unit tests pass on a PR to merge it (excepting special circumstances, e.g. when there are multiple failures being handled by separate PRs).
- We *aspire* to have TAV tests pass for all PRs. (An example of an exception to this is if a new version of a module being tested is published that breaks an instrumentation. It isn't necessarily the responsibility of the current PR to fix that.)

Some open questions:
- I've reduced the node version matrix for the TAV tests. Are we okay with this? (Note that it will then match the node versions in the matrix in test-all-versions.yml -- the daily TAV test run.)
- I'm using the fairly recent YAML anchor/alias support (https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases) to reduce duplication. I'm not sure this'll work as I have it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,6 +54,7 @@ jobs:
 
   unit-test:
     needs: compile
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -64,8 +65,9 @@ jobs:
           - "20"
           - "22"
           - "24"
-    runs-on: ubuntu-latest
-    services:
+    # `&foo` syntax is a YAML anchor. See:
+    # https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases
+    services: &test_services
       memcached:
         image: memcached:1.6.39-alpine
         ports:
@@ -147,7 +149,7 @@ jobs:
         env:
           RABBITMQ_DEFAULT_USER: username
           RABBITMQ_DEFAULT_PASS: password
-    env:
+    env: &test_env
       RUN_CASSANDRA_TESTS: 1
       RUN_MEMCACHED_TESTS: 1
       RUN_MONGODB_TESTS: 1
@@ -218,6 +220,46 @@ jobs:
         run: npm run test:browser:ci:affected
       - name: Unit tests (Nodejs - Delta)
         run: npm run test:ci:affected
+
+  test-all-versions:
+    needs: unit-test
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node:
+          - "18"
+          - "20"
+          - "22"
+          - "24"
+    services: *test_services
+    #XXX why lint error on env: here with the YAML alias
+    env: *test_env
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v5
+        with:
+          node-version: ${{ matrix.node }}
+      - run: npm ci
+      - name: Download Build Artifacts
+        uses: actions/download-artifact@v5
+        with:
+          name: compile-cache-${{ github.run_number }}
+          path: .nx
+      # Note: see comment in the compile job
+      - name: Set base and head commits
+        run: |
+          if [ "${{github.event_name}}" == "push" ]; then
+            echo "NX_BASE=origin/main~1" >> "$GITHUB_ENV"
+            echo "NX_HEAD=origin/main" >> "$GITHUB_ENV"
+          else
+            echo "NX_BASE=origin/main" >> "$GITHUB_ENV"
+            echo "NX_HEAD=HEAD" >> "$GITHUB_ENV"
+          fi
+      - name: Compile (Delta)
+        run: npm run compile:ci:affected
       - name: Test All Versions (Delta)
         run: npm run test-all-versions:ci:affected
       - name: Upload Test Artifacts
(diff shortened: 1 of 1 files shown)


trentm:
(My assumption is that workflows run with the .yml on *main* rather than my changes here, so I assume I won't be able to test if my workflow changes work here.)

trentm:
> (My assumption is that workflows run with the .yml on _main_ rather than my changes here, so I assume I won't be able to test if my workflow changes work here.)

Nope. Looks like it does run with my workflow changes in the PR. Nicer for testing.

trentm:
First run seemed to work much as expected: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18327018063/job/52193894528?pr=3162

However, there were no packages "affected" per nx's `nx affected`, so we spent [28m of compute](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18327018063/usage?pr=3162) basically doing a no-op. All time in (a) initialize containers and (b) `npm ci`.

(Aside: Perhaps interestingly, `npm ci` is taking 2-4x as long with Node.js 24 as with earlier versions.)

I'll artificially add some package changes to see that working.

trentm:
@david-luna FYI. See the "Compile (Delta)" step of the "compile" job here: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18327259544/job/52194634063?pr=3162

```
> opentelemetry-js-contrib@0.1.0 compile:ci:affected
> nx affected -t compile
 NX   No explicit --base argument provided, but found environment variable NX_BASE so using its value as the affected base: origin/main
 NX   No explicit --head argument provided, but found environment variable NX_HEAD so using its value as the affected head: HEAD
 NX   Running target compile for 3 projects and 87 tasks they depend on:
- @opentelemetry/instrumentation-bunyan
- @opentelemetry/auto-instrumentations-node
- @opentelemetry/instrumentation-express
...
```

I think that because we have `@opentelemetry/auto-instrumentations-node`, any change in any package included in auto-instrumentations-node means that well end up doing a compile of almost all packages.   Just mentioning it because I noticed it. I'm not asking for a fix or workaround.

david-luna:
> I've reduced the node version matrix for the TAV tests.

It's okay for me. I think the coverage will not decrease since most of the code paths depend on the version of the instrumented package.

> I think that because we have `@opentelemetry/auto-instrumentations-node`, any change in any package included in auto-instrumentations-node means that well end up doing a compile of almost all packages. Just mentioning it because I noticed it. I'm not asking for a fix or workaround.

Yes. This was mentioned in the SIG and I've created https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3136 to discuss improvements. This is slightly better than the previous version of the workflow which was compiling all packages no matter what.

Removing `auto-intrumentations-node` (and web) from the set of affected packages on a PR would speed up the CI but we will ignore if a change in one instrumentation breaks the auto instrumentation process. Another option is to have a conmpile cache of the last commit in `main`.

.github/workflows/test.yml
```diff
@@ -218,6 +220,45 @@ jobs:
         run: npm run test:browser:ci:affected
       - name: Unit tests (Nodejs - Delta)
         run: npm run test:ci:affected
+
+  test-all-versions:
```
david-luna, line 224: If the goals is just to be able to merge if unit tests passed and we are having concerns about speed maybe we could consider the option of continue the build if `test-all-versions` step fails. ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error
(If I knew it before I'd use it)

I think a separate job looks more organised and you can easily spot where the failure was but comes with the cost of spinning up 4 CI nodes and start the tests services in each one of them. This job works in parallel with `unit-test` but the gain is not much. In https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18314859661/job/52152712110?pr=3141 you can see the unit tests for tedious instrumentation and auto instrumentations took 6sec

david-luna: I think of an use case that I would prefer a separate job for TAV. Updates to `package-lock.json` file are triggering tests for all packages but probably it's okay to merge once the unit tests pass (not sure if possible to merge a PR that has workflows running, even if it has all the required checks).

trentm: > (not sure if possible to merge a PR that has workflows running, even if it has all the required checks).

Yes, it is possible. The only thing stopping a merge is the required checks and require reviews.

david-luna: > > (not sure if possible to merge a PR that has workflows running, even if it has all the required checks).
> 
> Yes, it is possible. The only thing stopping a merge is the required checks and require reviews.

That's good and not depending on the context. The `test-all-versions` job is the one that uploads the coverage reports and fail if the coverage drops over the configured thresholds.

- in the case we update a single dependency of a package and we want to avoid waiting 1h for TAV to pass i think it's fine
- but if it could be abused for other PRs and in worst case scenario we won't be uploading any coverage reports at all (though I think is very unlikely)

I'm okay with this change as long as we do not abuse of skipping TAV. It would be good if we could have some notification if TAV fails on `main`. By looking at https://github.com/open-telemetry/opentelemetry-js-contrib/actions/workflows/test.yml we ran a lot of tests without flakyness.

trentm: > * but if it could be abused for other PRs and in worst case scenario we won't be uploading any coverage reports at all (though I think is very unlikely)

FWIW, my main use case (and the reason [we disabled the "required checks" in the first place](https://github.com/open-telemetry/admin/pull/276)) was to not have to wait for 1h+ for every small change we were making when attempting to fix the release process.

If we do merge a PR before `test-all-versions` jobs have completed: AFAIK **the jobs keep running** so the coverage will still get updated, I *think*. I haven't verified this.

Issue #3157, opened 2025-10-07
add instrumentation-openai to auto-instrumentations-node


trentm:
instrumentation-openai was added in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2941
Consider adding this to the default set in auto-instrumentations-node.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

aryamohanan:
Hi @trentm, I’m interested in working on this case. Could you please assign it to me?

PR #3163, opened 2025-10-08
add instrumentation-openai to auto-instrumentations-node

fsd-niraj:
#3157 
-

imported the package and added its instance to the `InstrumentationMap` in `utils.ts`
-


david-luna (changes requested):
The compilation fail reveals that you need to add `@opentelemetry/instrumentation-openai` in the dependencies list of the `package.json` file

maryliag:
Hi @fsd-niraj , thank you for your contribution.
On the issue itself someone else has asked to be assigned and also opened a PR #3164 for it, so I want to be respectful of the process, therefore I will close this PR
But please, keep an eye on any other issue you might be interested on and we would be happy to review it 😄

PR #3169, opened 2025-10-09
feat(instrumentation-memcached): support net.* and db.* semconv migra…

JamieDanielson:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12125,11 +12125,25 @@
       }
     },
     "node_modules/@types/chai": {
-      "version": "4.3.20",
-      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-4.3.20.tgz",
-      "integrity": "sha512-/pC9HAB5I/xMlc5FP77qjCnI16ChlJfW0tGa0IUcFn38VJrTV6DeZ60NU5KZBtaOZqjdpwTWohz5HU1RrhiYxQ==",
+      "version": "5.2.3",
+      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-5.2.3.tgz",
+      "integrity": "sha512-Mw558oeA9fFbv65/y4mHtXDs9bPnFMZAL/jxdPFUpOHHIXX91mcgEHbS5Lahr+pwZFR8A7GQleRWeI6cGFC2UA==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "dependencies": {
+        "@types/deep-eql": "*",
+        "assertion-error": "^2.0.1"
+      }
+    },
+    "node_modules/@types/chai/node_modules/assertion-error": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/assertion-error/-/assertion-error-2.0.1.tgz",
+      "integrity": "sha512-Izi8RQcffqCeNVgFigKli1ssklIbpHnCYc6AknXGYoB6grJqyeby7jv12JUQgmTAnIDnbck1uxksT4dzN3PWBA==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=12"
+      }
     },
     "node_modules/@types/co-body": {
       "version": "6.1.3",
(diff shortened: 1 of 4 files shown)


packages/instrumentation-memcached/src/instrumentation.ts
```diff
@@ -90,15 +111,24 @@ export class MemcachedInstrumentation extends InstrumentationBase<Instrumentatio
...
+
+      const attributes: api.Attributes = {
+        'memcached.version': moduleVersion,
+      };
+
+      if (instrumentation._dbSemconvStability & SemconvStability.OLD) {
+        attributes[ATTR_DB_SYSTEM] = DB_SYSTEM_VALUE_MEMCACHED;
+      }
+      if (instrumentation._dbSemconvStability & SemconvStability.STABLE) {
+        attributes[ATTR_DB_SYSTEM_NAME] = DB_SYSTEM_VALUE_MEMCACHED;
```
trentm: nit: use `DB_SYSTEM_NAME_VALUE_MEMCACHED` for this one

packages/instrumentation-memcached/test/index.test.ts
```diff
@@ -50,9 +60,14 @@ const CONFIG = {
 };
 
 const DEFAULT_ATTRIBUTES: Attributes = {
```
maryliag: the name might be misleading, because the default behaviour is to send only old, but I'm assuming is the default on your test after you use the `dup` env variables, but might be worth making the name a little more clear

packages/instrumentation-memcached/src/instrumentation.ts
```diff
@@ -134,16 +164,35 @@ export class MemcachedInstrumentation extends InstrumentationBase<Instrumentatio
...
         'db.memcached.key': query.key,
         'db.memcached.lifetime': query.lifetime,
-        [ATTR_DB_OPERATION]: query.type,
-        [ATTR_DB_STATEMENT]: instrumentation.getConfig()
-          .enhancedDatabaseReporting
-          ? query.command
-          : undefined,
-        ...utils.getPeerAttributes(client, server, query),
-      });
+        ...utils.getPeerAttributes(
```
maryliag, line 172: did you forgot to add your changes to the utils file? You have a compile errors because it's expecting 3 parameters, not 4 here

packages/instrumentation-memcached/src/instrumentation.ts
JamieDanielson: I started to do that but realized `DB_SYSTEM_NAME_VALUE_MEMCACHED` is still experimental 🤔

packages/instrumentation-memcached/src/instrumentation.ts
JamieDanielson: 🤦  yes

JamieDanielson: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3169/commits/f9ce510bf1dbbad5db7199787ccd89718ff3614a

packages/instrumentation-memcached/test/index.test.ts
JamieDanielson: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3169/commits/9743a7ac1f9ce360548150134aed764515cb340e

JamieDanielson:
@maryliag I made the changes if you can have another look when you get a chance 🙏

maryliag (approved):
🚀

packages/instrumentation-memcached/src/instrumentation.ts
trentm: Yes, it would mean adding a src/semconv.ts wiht a local copy of that var.

trentm: Oh, I see what you mean. Meh, I think it still counts as the right thing to do for the "stable" semconv.

`DB_SYSTEM_VALUE_MEMCACHED` is also "Stability: Development" and arguably is deprecated because it is an enum for `db.system` which is deprecated: https://github.com/open-telemetry/semantic-conventions/blob/41df6e41371948bb924cec6fe4468f0a68431852/model/database/deprecated/registry-deprecated.yaml#L370-L523

JamieDanielson: updated!
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3169/commits/cf291bd1d8538d63addd7d98db03d1bc427fc2e2

maryliag (approved):

PR #3182, opened 2025-10-18
feat(pg-instrumentation): record exceptions as span events

### LINKED ISSUE #3172: Record Postgres instrumentation errors

The Postgres instrumentation package, in various places, handles errors by marking the span as `ERROR` and setting the error message: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/d519d4bdd85db5954adb42dee038f3933f676f5f/packages/instrumentation-pg/src/utils.ts#L430-L435

However is it not clear _when_ the exception actually occurred. This makes it more challenging to debug issues since there is no clear timeline of events.

The OTEL Spec [says](https://opentelemetry.io/docs/specs/otel/trace/exceptions/):

> An exception SHOULD be recorded as an Event on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR.

So I propose that in the places where the spans are being set to `ERROR` and the error re-raised, we also record the exception via `span.recordException(err)`

N/A

aryamohanan:

This PR addresses the lack of recorded exceptions in PostgreSQL spans. Previously, when a query failed, the span only set `SpanStatusCode.ERROR` without capturing exception details.

Fixes #3172

* Added exceptions event  to record unhandled query errors as span events.
* Added tests.


maryliag:
Hi, thank you for working on this!

I have just one concern: this is also recording the error message. Currently we're sending error messages in a few places, but there is a concern that it could be sending private information.
For example, imagine you're trying to store a private information, but you got an error, so you might have errors like `value 'super-secret-info' is not type X`, so now you have something confidential being sent as is.
We do have a plan on removing the messages from a few places and just using the error code, so you can follow the same logic here.

`span.recordException(e);` also accepts a string as a parameter, so you can create a function that gets the error code and create a nice message that doesn't contain private info, and pass only that as your argument.

Let me know if you have questions, and I will be happy to provide more guidance!

aryamohanan:
Hi @maryliag 

Thanks for the feedback ❤️ 

I’ve updated it to use `sanitizedErrorMessage`, which records only the error type and PostgreSQL code in span.recordException, omitting any sensitive details.

Let me know if any further tweaks are needed.

maryliag (approved):
Great! Thank you for making the changes!
🚀

PR #3183, opened 2025-10-19
build(instrumentation-winston): fix phantom deps

jjangga0214:

`types.ts` depends on  `@opentelemetry/winston-transport`, causing a [phantom dependency](https://rushjs.io/pages/advanced/phantom_deps/).

It should be in **`dependencies`**, not `devDependencies`.

**Why `dependencies` is required here:**
- `WinstonInstrumentationConfig extends InstrumentationConfig` - the interface is extended and exported
- Any consumer importing this type needs to resolve `InstrumentationConfig` at compile time
- The type is part of the public API surface

**When `devDependencies` would be acceptable:**
- If using `import type` for types that are ONLY used internally and never re-exported
- Example: `import type { InstrumentationConfig } from '...'` for local type constraints only

**Rule of thumb:** If you extend or re-export a type from a package, that package must be in `dependencies`.

### DIFF
--- a/packages/instrumentation-winston/package.json
+++ b/packages/instrumentation-winston/package.json
@@ -49,7 +49,6 @@
     "@opentelemetry/context-async-hooks": "^2.0.0",
     "@opentelemetry/sdk-trace-base": "^2.0.0",
     "@opentelemetry/sdk-trace-node": "^2.0.0",
-    "@opentelemetry/winston-transport": "^0.17.0",
     "@types/mocha": "10.0.10",
     "@types/node": "18.18.14",
     "@types/sinon": "17.0.4",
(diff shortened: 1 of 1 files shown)


pichlermarc:
We actually don't re-export the types as stated in the PR title. We do however dynamically require it (we catch any errors if it does not exist).

jjangga0214:
@pichlermarc 

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/23c5f4e8884ba87ab2c697bf34b34e106ed3abc5/packages/instrumentation-winston/src/types.ts#L19-L24

On dependent(of this package)'s side, eslint complains `@opentelemetry/winston-transport` is missing. The reason is what I've written above.

jjangga0214:
Oops, I am sorry. As you said, `@opentelemetry/instrumentation` is used, not `opentelemetry/winston-transport`.. My mistake.

PR #3189, opened 2025-10-21
chore(instrumentation-aws-lambda): move aws lambda types dependency to dev dependencies

wpessers:

- The `@types/aws-lambda` dependency is currently a dependency in the instrumentation-aws-lambda package. This is causing some version conflicts in https://github.com/open-telemetry/opentelemetry-lambda where we use the types package as a devDependency. Since the types package is only useful at compile time I don't see why it should not be a dev dependency here either.

- Moved the `@types/aws-lambda` dependency to dev dependencies

### DIFF
--- a/packages/instrumentation-aws-lambda/package.json
+++ b/packages/instrumentation-aws-lambda/package.json
@@ -51,6 +51,7 @@
     "@opentelemetry/sdk-metrics": "^2.0.0",
     "@opentelemetry/sdk-trace-base": "^2.0.0",
     "@opentelemetry/sdk-trace-node": "^2.0.0",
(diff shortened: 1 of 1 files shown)


pichlermarc (changes requested):
Hi @wpessers. Unfortunately, we can't do that as these types are used in the public API of that package. Omitting it will cause transpile errors for *all* of our users.

To move it do `devDependencies` we have to do what I suggested [here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3168#issuecomment-3397302125) and move copy the type that we need to this package.

wpessers:
Ok, got it. Will look at another fix on our end then.

PR #3196, opened 2025-10-27
fix(pg-instrumentation): capture query values when passed as second argument for prepared statements

### LINKED ISSUE #3193: Query values not captured when passed as second argument (prepared statements)

- `@opentelemetry/instrumentation-pg`: `0.59.0`
- `@opentelemetry/auto-instrumentations-node`: `0.65.0`
- `@opentelemetry/sdk-node`: `0.206.0`

`v22.21.0`

### What did you do?

I'm using Drizzle ORM with PostgreSQL with prepared statements, which internally calls `pg.Client.query()` with the following signature:

```typescript
client.query(
  { name: 'statement_name', text: 'SELECT * FROM users WHERE id = $1' },
  [123] // values passed as second argument
)
```

I enabled `enhancedDatabaseReporting: true` in the pg instrumentation config:

```typescript
'@opentelemetry/instrumentation-pg': {
  enabled: true,
  enhancedDatabaseReporting: true,
}
```

I expected to see the `db.postgresql.values` attribute in my spans containing the query parameter values (e.g., `["123"]`).

### What did you see instead?

The `db.postgresql.values` attribute is not present in the spans. The span only contains:
- `db.statement`: The SQL query text
- `db.postgresql.plan`: The prepared statement name
- Other connection attributes

But no parameter values.

### Root Cause

The bug is in `[instrumentation.ts](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-pg/src/instrumentation.ts#L336-L350)`:

```typescript
const queryConfig = firstArgIsString
  ? {
      text: arg0 as string,
      values: Array.isArray(args[1]) ? args[1] : undefined,  // ✅ Checks args[1]
    }
  : firstArgIsQueryObjectWithText
  ? (arg0 as utils.ObjectWithText)  // ❌ Just casts arg0, ignores args[1]!
  : undefined;
```

When the first argument is a **query config object** (common with prepared statements), the code just casts `arg0` without checking if values are passed as `args[1]`. This means `queryConfig.values` remains `undefined` even when values are provided.

### Proposed Fix

```typescript
const queryConfig = firstArgIsString
  ? {
      text: arg0 as string,
      values: Array.isArray(args[1]) ? args[1] : undefined,
    }
  : firstArgIsQueryObjectWithText
  ? {
      ...(arg0 as utils.ObjectWithText),
      // Merge values from args[1] if not already in the config object
      values: (arg0 as any).values ?? (Array.isArray(args[1]) ? args[1] : undefined),
    }
  : undefined;
```

This would check for values in `args[1]` when using a query config object, just like it already does for string queries.

### Workaround

As a temporary workaround, I'm patching `pg.Client.prototype.query` and `pg.Pool.prototype.query` before OpenTelemetry instruments them, to merge `args[1]` into `args[0].values` when this pattern is detected.

### Additional context

This affects any ORM or library that uses prepared statements with the signature `query(config, values)` where values are passed separately from the config object. This includes:
- Drizzle ORM (when using `.prepare()`)
- Potentially other ORMs that use a similar pattern

The `requestHook` cannot work around this issue because by the time it's called, `queryConfig.values` is already `undefined`.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

aryamohanan:

Fixes #3193

When using prepared statements with the following syntax, query parameter values were not captured even when `enhancedDatabaseReporting: true` was enabled:

```js
client.query(
  { name: 'statement_name', text: 'SELECT * FROM pg_tables WHERE schemaname = $1' },
  ['public']
);
```

* Fixed missing parameter values when executing prepared statements.
* Updated instrumentation to correctly read values from the second argument (`args[1]`) when the first argument is a query config object.
* Ensured that `db.postgresql.values` is only recorded when `enhancedDatabaseReporting:true`.
* Maintained query text sanitization, placeholders (`$1`, `$2`, etc.) remain in `db.statement`.


aryamohanan:
@maryliag Could you please take a look at this PR? TIA

packages/instrumentation-pg/src/instrumentation.ts
```diff
@@ -360,12 +360,21 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
 
         const instrumentationConfig = plugin.getConfig();
 
+        // Normalize the query configuration so that prepared statements with a separate
```
maryliag: I like the suggestion on the issue itself, of just updating the line 342 to 
`values: (arg0 as any).values ?? (Array.isArray(args[1]) ? args[1] : undefined),`, this way it avoids doing some repetition on checks, which is what you're doing here.

Would you mind making that solutions instead and checking if the tests will work as expected?

maryliag (approved):
Awesome! Thank you!

PR #3197, opened 2025-10-29
chore(ci): limit installs to 10 minutes

overbalance:
I have a PR that has been running for three hours (and I don't have permission to cancel it).
https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/18913134967/job/53989890872?pr=3032

Added a timeout to all install CI commands. I modified every install in test and test-all-versions, please advise if that's not correct.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
           node-version: 18
       - name: Install
         run: npm ci --ignore-scripts
+        timeout-minutes: 10
       # Note: when pushing to main we want to test only what changed in
       # last commit. Otherwise we want to test all changes from origin/main.
       # So we set the right values for base and head commits depending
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -141,11 +142,11 @@ jobs:
       cassandra:
         image: cassandra:3
         ports:
-        - 9042:9042
+          - 9042:9042
       rabbitmq:
         image: rabbitmq:3
         ports:
-        - 22221:5672
+          - 22221:5672
```
trentm, line 149: nit: Is this from some local yaml lint tool you have?

trentm (approved):

.github/workflows/test.yml
overbalance: Ah yes, Prettier runs in Zed on save, by default.

PR #3198, opened 2025-10-29
chore(ci): add compilation cache of main branch

david-luna:

Improve compilation times by adding a compilation cache which contains the artefacts form the last commit in main. Ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3136#issuecomment-3380429812

- add a global compilation cache with the cache action https://github.com/actions/cache so its available for all workflow runs
- compilation step restores that cache and uses it for compilation.
- if build happens in main branch (push event) workflow updates the global cache
- on PRs workflow uploads specific cache for test jobs as it was done before

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,10 +35,14 @@ jobs:
         with:
           key: compile-cache-main
           path: .nx
+      # Cache was build on a different machine so nx will complain.
+      # ref: https://nx.dev/docs/troubleshooting/unknown-local-cache
+      # The cache is only created in this workflow so its okay to skip the check.
       - name: Compile packages
-        run: npm run compile
+        run: NX_REJECT_UNKNOWN_LOCAL_CACHE=0 npm run compile
       # Note: update the compilation cache for main branch
       - name: Compile Cache Update (Push)
+        if: ${{ github.event_name == 'push' }}
         uses: actions/cache/save@v4
         with:
           key: compile-cache-main


.github/workflows/test.yml
```diff
@@ -27,25 +27,27 @@ jobs:
...
-            echo "NX_BASE=origin/main" >> "$GITHUB_ENV"
-            echo "NX_HEAD=HEAD" >> "$GITHUB_ENV"
-          fi
-      - name: Compile (Delta)
-        run: npm run compile:ci:affected
-      - name: Upload Build Artifacts
-        uses: actions/upload-artifact@v4
+      # Note: we restore the last compilation form main branch
+      # so NX will only compile what has changed since then.
+      # Cache action is used because can be restores from dfferen workfow runs
```
trentm:
```suggestion
      # Restore the nx compilation cache for the latest commit to main.
      # The subsequent 'npm run compile' should only need to compile changed packages.
```

.github/workflows/test.yml
```diff
@@ -27,25 +27,27 @@ jobs:
...
+      - name: Compile Cache Lookup
+        uses: actions/cache/restore@v4
```
trentm, line 34: Do we want to consider *not* using the cache for pushes to "main"?
IIUC, the way it is now, that `compile-cache-main` will never start fresh. After a year, it will have a year of possible cruft in it, right?  I realize that would mean doing a full clean `npm run compile` for pushes to main, but we'd still get the benefit of the cache for the more common updates to PRs.

david-luna: We can consider it :)

> the way it is now, that compile-cache-main will never start fresh.

you're right in the sense that the whole cache will never be refreshed at once unless a push changes all the packages at once. But even in that scenario we will refresh contents for each package keeping the cache structure. We could call it [Theseus' cache](https://en.wikipedia.org/wiki/Ship_of_Theseus). 

I think the gains we get for PRs are more than enough and it is possible if we keep the cache for long time we may get som issues form `nx` if this changes the way it manages caching. Obviously after an update.

I'm going to add a new commit so:
- compiles everything in `main` and sets the cache
- uses the cache for PRs only

david-luna: commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3198/commits/71859dcfe01f5b06d4c37b1089c25e4fbb202ee2

however in the [last run](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/19130605517/job/54670196499?pr=3198) I got the following mewssage

```
 NX   Unrecognized Cache Artifacts
Nx found unrecognized artifacts in the cache directory and will not be able to use them.
Nx can only restore artifacts it has metadata about.
Read about this warning and how to address it here: https://nx.dev/troubleshooting/unknown-local-cache
```
need to review the troubleshooting guide and test a bit more in this PR

david-luna: The upload/download action works fine but not the cache. If we contain the cache management in this workflow I think we could tell nx to trust the cache with `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` ([ref](https://nx.dev/docs/troubleshooting/unknown-local-cache#you-share-cache-with-another-machine-using-a-network-drive)).

If we discard that option (which is fine since nx discourages to disable cache checking) I think the alternative is to explicitly tell nx when to compile and test `auto-instrumentations-*` packages which are the ones that forces to compile the rest of them.

david-luna: skipping the check did not work in https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/19134336471/job/54682522648?pr=3198 so I guess this is a dead end :( 

message
```
NX_REJECT_UNKNOWN_LOCAL_CACHE=0 is not supported with the new database cache. Read more at https://nx.dev/deprecated/legacy-cache#nxrejectunknownlocalcache.
```

@trentm  I'll close this PR and open a new one with the alternative mentioned in previous comment. Thanks for your review :)

PR #3206, opened 2025-11-04
fix(knex-instrumentation): use correct db system attribute

pantajoe:

[MikroORM](https://mikro-orm.io) overrides the default postgres client in knex to a [`PostreSqlKnexDialect`](https://mikro-orm.io/api/knex/class/PostgreSqlKnexDialect), so `Knex.client.config.client` is not a string, but an instance to that dialect class.

Therefore, for users of an ORM that does this with knex, the knex instrumentation triggers a bunch of `Invalid attribute set for database.system.name` warnings.

Use `knex.Client.config.driverName` instead of `knex.Client.config.client` to derive the DB system attribute: `database.system.name` (or the deprecated one `database.system`)


raphael-theriault-swi:
Could you add a test case that illustrates the issue being fixed, and to make sure there isn't a regression in the future ?

pantajoe:
> Could you add a test case that illustrates the issue being fixed, and to make sure there isn't a regression in the future ?

Done ✅

david-luna:
@pantajoe 

I've started the workflows but I'm not sure if this change will be compatible with all the versions supported `>=0.10.0 <4`. 

Update: unit tests pass but we only test with the version set in `package.json`. There is no other version tested

```sh
Run npm run test-all-versions:ci:affected

> opentelemetry-js-contrib@0.1.0 test-all-versions:ci:affected
> npm --loglevel=silent run list:affected | xargs -I {} -t npm run --if-present test-all-versions -w {}
```

david-luna:
Tested down to v1.0.0 locally with the changes in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3221 and no errors :)

Let's merge #3221 first so we see the tests results in CI.

pantajoe:
Nice, happy to hear that 👍🏽

PR #3205, opened 2025-11-03
fix(example-express): update to Express v5 and fix deprecated APIs

overbalance:
## Which problem is this PR solving?

The Express example was using Express v4 which has been superseded by Express v5. Additionally, the example was using deprecated OpenTelemetry APIs (`new Resource()`) and incorrect error logging patterns.

## Short description of the changes

- Upgraded Express from v4.17.1 to v5.1.0 and updated types
- Updated OpenTelemetry packages to latest versions (`@opentelemetry/instrumentation-express` from v0.44.0 to v0.56.0, `@opentelemetry/resources` from v1.27.0 to v2.2.0)
- Replaced deprecated `new Resource()` constructor with `resourceFromAttributes()` function
- Fixed error logging to use `console.error()` instead of `console.log()`
- Added proper type import for `Resource` in server.ts
- Updated imports to use `SamplingDecision` from correct package location

### Test Plan

- [x] Updated dependencies
- [x] Verified Express v5 compatibility
- [x] Confirmed deprecated API replacements follow current OpenTelemetry standards

PR #3207, opened 2025-11-05
fix(express-example): minor tweaks after #3205

trentm:
That `resources` is unrelated to OTel `Resource` I believe.


raphael-theriault-swi (approved):
My bad on not catching that

PR #3208, opened 2025-11-06
chore(ci): pick better which packages should compile and test

david-luna:
ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3136

This is an alternative to [using a cache](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3198) for speeding up the CI time spent in PRs.

This approach uses a different way to resolve which packages need to be compiled and tested. The result is we get only the list of packages which sources are modified and this is what we pass to the `nx run-many` command.

**Why `nx affected -t compile` compiles so many packages?**

The reason is because internally the command takes in consideration `nx.json` dependencies on tasks. The script iterates over each package and runs the "compile" task if:
- the package itself is modified
- the compile task depends on another compile task in another package that has been modified

So `nx.json` is expanding the dependency graph. Now we know a bit more about the behaviour of the command.

**Which solution is proposed?**

The command `nx affected` if called without params gives a list with only the changed packages (which is what we want). Then we can pass that list to `nx run-many` command to narrow the list of packages that should run the given task (compile, test). Here is an example

```sh
npx nx run-many -t compile -p "@opentelemetry/instrumentation-aws-sdk" "@opentelemetry/instrumentation-undici"
   ✔  nx run @opentelemetry/instrumentation-aws-sdk:version:update (923ms)
   ✔  nx run @opentelemetry/instrumentation-undici:version:update (926ms)
   ✔  nx run @opentelemetry/contrib-test-utils:compile (3s)
   ✔  nx run @opentelemetry/instrumentation-undici:compile (3s)
   ✔  nx run @opentelemetry/instrumentation-aws-sdk:compile (3s)
```
you can see that the command takes care of dependencies like `@opentelemetry/contrib-test-utils`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,8 +42,6 @@ jobs:
             echo "NX_BASE=origin/main" >> "$GITHUB_ENV"
             echo "NX_HEAD=HEAD" >> "$GITHUB_ENV"
           fi
-      - name: List affected packages
-        run: npm run list:affected
       - name: Compile (Delta)
         run: npm run compile:ci:affected
       - name: Upload Build Artifacts


david-luna:
In https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/19137279065/job/54692601282?pr=3208 some of the test fails because network issues in `npm ci`. But the list of packages on compilation and test are what I think we're looking for.

Commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3208/commits/20e7409eed9c424b4ef932b8a0069ca973685396 added a change in pg instrumentation and the `contrib-test-utils` package has been compiled since its a dependency of the instrumentation.

packages/instrumentation-undici/src/undici.ts
```diff
@@ -88,7 +88,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
   }
 
   // No need to instrument files/modules
-  protected override init() {
+  override init() {
```
trentm, line 91: nit: mismerge?  (Same thing below)

david-luna: Did this change on purpose to validate the workflow was picking correctly. Will remove it before merging. Thanks for the heads up

david-luna:
In a conversation with @trentm we agreed that this approach is not what we want in some situations. An example is if `@opentelemetry/contrib-test-utils` package is changed and no other then we need to run the tests for all packages that are  depending on it because that change may break tests.

After that discussion I think **the only packages** we may want to skip compiling/testing if not changed is `@opentelemetry/auto-instrumentations-node` and `@opentelemetry/auto-instrumentations-web` so I think we could tell `nx` to ignore them if not modified

david-luna:
Since the package graph changes depending on the input task of the `nx affected` command it's quite complex to reason about what's the best option. This workflows does more than one thing (compile, test & TAV) and each task can generate a different dependency graph. Also the resolution process is done internally in `nx` so we're not sure it's going to stay that way.

So I'll close this PR and give a second opportunity to [caching](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3198) in a new PR. The principal concern was about keeping a cache for a long period of time. There is a way to clean the cache using GH CLI ([example](https://github.com/playframework/playframework/blob/3a78ae68acb10e614e8b1f1411ff4f7b93cd909e/.github/workflows/delete-caches.yml)).

PR #3224, opened 2025-11-14
chore(deps): remove rollup override workaround

overbalance:

Removes the `rollup@4.52.5` override that was previously added as a workaround for the postinstall script issue in rollup 4.53.0.

Ref: https://github.com/rollup/rollup/issues/6168

- Removed the `overrides` section from `package.json` that pinned rollup to version 4.52.5
- Updated `package-lock.json` to reflect the latest compatible rollup version and its dependencies

The rollup postinstall script issue has been resolved in newer versions, so the override is no longer needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -282,59 +282,58 @@
       }
     },
     "node_modules/@aws-sdk/client-bedrock-runtime": {
-      "version": "3.926.0",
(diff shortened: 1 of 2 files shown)


raphael-theriault-swi:
cc @trentm since I think you have the most context on this

PR #3226, opened 2025-11-17
feat(oracledb): propagate context using V$SESSION.ACTION

mhennoch:
Adds context propagation via V$SESSION.ACTION. [Spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/db/oracledb.md#vsessionaction). It seems that requestHook doesn't have access to the connection so I couldn't use that.

### DIFF
--- a/packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
+++ b/packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
@@ -65,9 +65,7 @@ function getTraceHandlerBaseClass(
   }
 }
 
-export function buildTraceparent(
-  spanContext: SpanContext
-): string | undefined {
+export function buildTraceparent(spanContext: SpanContext): string | undefined {
   return `00-${spanContext.traceId}-${spanContext.spanId}-0${Number(
     spanContext.traceFlags || TraceFlags.NONE
   ).toString(16)}`;
--- a/packages/instrumentation-oracledb/test/oracle.test.ts
+++ b/packages/instrumentation-oracledb/test/oracle.test.ts
@@ -36,7 +36,7 @@ import {
 import * as assert from 'assert';
 import { OracleInstrumentation } from '../src';
 import { SpanNames } from '../src/constants';
-import { buildTraceparent} from '../src/OracleTelemetryTraceHandler';
+import { buildTraceparent } from '../src/OracleTelemetryTraceHandler';
 
 import {
   ATTR_DB_NAMESPACE,
@@ -985,7 +985,7 @@ describe('oracledb', () => {
         { outFormat: oracledb.OUT_FORMAT_OBJECT }
       );
       const row = result.rows?.[0] as Record<string, string> | undefined;
-      const actionValue = row?.ACTION;        
+      const actionValue = row?.ACTION;
       const spans = memoryExporter.getFinishedSpans();
       const executeSpan = spans[spans.length - 1];
       assert.ok(executeSpan, 'expected span to verify trace propagation');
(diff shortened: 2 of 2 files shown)


pichlermarc:
cc @sudarshan12s @sharadraju (component owners) please review 🙂

trentm:
Correct link for the "Spec", I think: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/db/oracledb.md#vsessionaction
Not sure if it moved recently.

mhennoch:
@sudarshan12s @sharadraju any chance you could take a look?

packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -63,6 +65,14 @@ function getTraceHandlerBaseClass(
   }
 }
 
+export function buildTraceparent(
+  spanContext: SpanContext
+): string | undefined {
+  return `00-${spanContext.traceId}-${spanContext.spanId}-0${Number(
```
sudarshan12s, line 69: traceFlags is already a number , we do not need to use Number right?

I think we can use below?
```

(spanContext.traceFlags ?? TraceFlags.NONE)
    .toString(16)
    .padStart(2, '0');
```

packages/instrumentation-oracledb/src/types.ts
```diff
@@ -91,4 +91,11 @@ export interface OracleInstrumentationConfig extends InstrumentationConfig {
    * @default false
    */
   requireParentSpan?: boolean;
+
+  /**
+   * Automatic propagation of trace context using V$SESSION.ACTION.
+   *
+   * @default false
+   */
+  traceContextPropagation?: boolean;
```
sudarshan12s: The database wire protocol latest version supports passing the traceContext from client to server. The DB server can then start a child span of the corresponding client span. 

I have plans to add it in this module after few `node-oracledb` driver changes..

The `ACTION` field could be used by application and hence the native change of wire protocol would resolve that issue (also the limit of 64 bytes with `ACTION` field)

Hence can we rename this attribute to 
`propagateTraceContextToSessionAction` , so that when a wire protocol change is added, we can use this flag, `traceContextPropagation` ?


Also can we add the same in README.md

| `propagateTraceContextToSessionAction` | `boolean` | `true` | If true, injects the W3C Trace Context into the Oracle `V$SESSION.ACTION` field. This allows the OpenTelemetry Collector to correlate application traces with database server spans. |

sudarshan12s:
> @sudarshan12s @sharadraju any chance you could take a look?

Sorry for delay in checking this.  Thanks for the changes.

packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
mhennoch: I think we could but I used the same one that is used in [tedious](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-tedious/src/instrumentation.ts#L161) which is from [here](https://github.com/open-telemetry/opentelemetry-js/blob/a2e3f4542f9a6a61c569d4f70fa49220da18cda0/packages/opentelemetry-core/src/trace/W3CTraceContextPropagator.ts#L84-L86). Probably should extract it to somewhere and use the same one. I'd say let this be and I'll unify it in some other PR.

packages/instrumentation-oracledb/src/types.ts
mhennoch: Tedious uses [enableTraceContextPropagation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-tedious/README.md#trace-context-propagation) so to be honest I am not sure why didn't use the same one here. Spec doesn't specify it so we can use whatever. Would make sense for all DBs to use the same one tho. I guess oracledb only one who will use action for this so I'll use `propagateTraceContextToSessionAction`.

mhennoch:
@pichlermarc

PR #3232, opened 2025-11-20
chore: assign Browser SIG Maintainers Approver-permissions for browser packages

pichlermarc:
Same as https://github.com/open-telemetry/opentelemetry-js/pull/6093, but for contrib.
Assigns Approver permissions for @open-telemetry/browser-maintainers, scoped to packages that target browsers.

cc @open-telemetry/javascript-maintainers


JamieDanielson (approved):
I believe the error in the file is because we need the admin PR to be merged first. Approving for the change but let's make sure the error goes away before merging 😄

pichlermarc:
errors are now gone since the admin PR has merged. 🙂

PR #3249, opened 2025-11-26
fix(pg-instrumentation): capture query props when passed as class instance

GochoMugo:

When using sql-templating libraries such  as [sql-template-strings](https://www.npmjs.com/package/sql-template-strings), the query object is usually not a plain object, and instead a class instance.

An [extracted](https://github.com/felixfbecker/node-sql-template-strings/blob/master/index.js) snippet:

```js
class SQLStatement {
  constructor(strings, values) {
    this.strings = strings
    this.values = values
  }

  /** Returns the SQL Statement for node-postgres */
  get text() {
    return this.strings.reduce((prev, curr, i) => prev + '$' + i + curr)
  }

  setName(name) {
    this.name = name
    return this
  }
}

function SQL(strings) {
  return new SQLStatement(strings.slice(0), Array.from(arguments).slice(1))
}
```

This class would be instantiated, for example, when using template string:

```js
client.query(SQL`select ...`);
```

With the changes added in PR #3196, the `text` property from the `Statement` class above would not be captured when using the spread operator i.e. `{ ...arg0 }`.

For example,

```js
class Klass {
  constructor() {
    this.foo = 1;
  }
  get bar() {
    return 2;
  }
}
const instance = new Klass();
const obj = { ...instance };

assert(obj.foo === 1);
assert(obj.bar === undefined); // `bar` is not set!
```

The current fix is to explicitly capture `name` and `text` props while still using the spread operator.

### DIFF
--- a/packages/instrumentation-pg/src/instrumentation.ts
+++ b/packages/instrumentation-pg/src/instrumentation.ts
@@ -333,24 +333,21 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
         // TODO: remove the `as ...` casts below when the TS version is upgraded.
         // Newer TS versions will use the result of firstArgIsQueryObjectWithText
         // to properly narrow arg0, but TS 4.3.5 does not.
-        let queryConfig: any;
-
-        if (firstArgIsString) {
-          queryConfig = {
-            text: arg0 as string,
-            values: Array.isArray(args[1]) ? args[1] : undefined,
-          };
-        } else if (firstArgIsQueryObjectWithText) {
-          const q = arg0 as any;
-
-          if (q.values === undefined && Array.isArray(args[1])) {
-            q.values = args[1];
-          }
-
-          queryConfig = q;
-        } else {
-          queryConfig = undefined;
-        }
+        const queryConfig = firstArgIsString
+          ? {
+              text: arg0 as string,
+              values: Array.isArray(args[1]) ? args[1] : undefined,
+            }
+          : firstArgIsQueryObjectWithText
+            ? {
+                ...(arg0 as any),
+                name: arg0.name,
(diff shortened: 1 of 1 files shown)


packages/instrumentation-pg/test/pg.test.ts
```diff
@@ -583,10 +583,15 @@ describe('pg', () => {
       });
 
       it('should record query and values for prepared statements', done => {
```
maryliag, line 585: instead of updating this test, can you add a new one, this way we can check changes work with both scenarios?

GochoMugo: @maryliag Addressed in commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3249/commits/b8524b1e4b88d253350740e49fabd392c93d8d08.

raphael-theriault-swi:
Hey @GochoMugo, I reviewed #3259 without seeing that you opened this before. However I do prefer the approach you're taking here by still creating a new object as opposed to updating the existing one. Would you be able to rebase your changes on the current version ? There are already new tests covering the non-enumerable property cases so only the instrumentation changes would be needed.

GochoMugo:
> Hey @GochoMugo, I reviewed #3259 without seeing that you opened this before. However I do prefer the approach you're taking here by still creating a new object as opposed to updating the existing one. Would you be able to rebase your changes on the current version ? There are already new tests covering the non-enumerable property cases so only the instrumentation changes would be needed.

@raphael-theriault-swi Sure, I have rebased the PR.

maryliag:
@GochoMugo for future reference, don't force push your changes, because then we miss the history and we can't see what exactly changed since latest review :) 
thanks for making the changes

PR #3242, opened 2025-11-24
fix(instrumentation-openai)!: add missing access modifiers to private methods

pichlermarc:
Fixes #3240 
Introduced via #3164 -> #3157

Errors stem from OpenAI types being used in the instrumentation's public interface. Instrumentations are supposed to no-op without causing errors even if the instrumented package is not installed.

This PR fixes this by adding `private` access modifiers the `OpenAIInstrumentation` intended private properties. This will rease the types in the compile output and removes the imports.

You can test this by following  these steps:
- check out `main`
- run `npm ci && npm run compile`
- inspect `./packages/instrumentation-openai/build/src/instrumentation.d.ts`
  - you'll see type-imports from `openai`
- check out this PRs branch
- run `npm ci && npm run compile`
- inspect `./packages/instrumentation-openai/build/src/instrumentation.d.ts`
  - type imports from `openai` should be gone

**Before** this change the compiled output looks like this:
```ts
// instrumentation.d.ts
import type { Attributes, Context, Span } from '@opentelemetry/api';
import { InstrumentationBase, InstrumentationNodeModuleDefinition } from '@opentelemetry/instrumentation';
import type { ChatCompletion, ChatCompletionCreateParams, ChatCompletionChunk } from 'openai/resources/chat/completions';
import type { CreateEmbeddingResponse, EmbeddingCreateParams } from 'openai/resources/embeddings';
import { OpenAIInstrumentationConfig } from './types';
export declare class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumentationConfig> {
    private _genaiClientOperationDuration;
    private _genaiClientTokenUsage;
    constructor(config?: OpenAIInstrumentationConfig);
    setConfig(config?: OpenAIInstrumentationConfig): void;
    protected init(): InstrumentationNodeModuleDefinition[];
    _updateMetricInstruments(): void;
    _getPatchedChatCompletionsCreate(): any;
    /**
     * Start a span for this chat-completion API call. This also emits log events
     * as appropriate for the request params.
     */
    _startChatCompletionsSpan(params: ChatCompletionCreateParams, config: OpenAIInstrumentationConfig, baseURL: string | undefined): {
        span: Span;
        ctx: Context;
        commonAttrs: Attributes;
    };
    /**
     * This wraps an instance of a `openai/streaming.Stream.iterator()`, an
     * async iterator. It should yield the chunks unchanged, and gather telemetry
     * data from those chunks, then end the span.
     */
[...]

PR #3250, opened 2025-11-26
feat: apply rule @typescript-eslint/explicit-member-accessibility

legendecas:

- Identifying pitfalls like https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3242 earlier.

Given that applying the `private` access modifier could be a potential breaking change, this does not apply the rule to every package to reduce the burden of reviewers.

- Apply rule `@typescript-eslint/explicit-member-accessibility` to (non-instrumentation) packages (less changes).

### DIFF
--- a/packages/resource-detector-container/src/detectors/ContainerDetector.ts
+++ b/packages/resource-detector-container/src/detectors/ContainerDetector.ts
@@ -24,12 +24,12 @@ import { extractContainerIdFromLine } from './utils';
 import { ATTR_CONTAINER_ID } from '../semconv';
 
 export class ContainerDetector implements ResourceDetector {
-  private readonly CONTAINER_ID_LENGTH = 64;
-  private readonly DEFAULT_CGROUP_V1_PATH = '/proc/self/cgroup';
-  private readonly DEFAULT_CGROUP_V2_PATH = '/proc/self/mountinfo';
-  private readonly UTF8_UNICODE = 'utf8';
-  private readonly HOSTNAME = 'hostname';
-  private readonly MARKING_PREFIX = ['containers', 'overlay-containers'];
+  public readonly CONTAINER_ID_LENGTH = 64;
+  public readonly DEFAULT_CGROUP_V1_PATH = '/proc/self/cgroup';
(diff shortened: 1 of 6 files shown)


packages/resource-detector-container/src/detectors/ContainerDetector.ts
```diff
@@ -24,20 +24,16 @@ import { extractContainerIdFromLine } from './utils';
...
-  readonly CRIO = 'crio-';
-  readonly CRI_CONTAINERD = 'cri-containerd-';
-  readonly DOCKER = 'docker-';
-  readonly HEX_STRING_REGEX: RegExp = /^[a-f0-9]+$/i;
+  private readonly CONTAINER_ID_LENGTH = 64;
+  private readonly DEFAULT_CGROUP_V1_PATH = '/proc/self/cgroup';
+  private readonly DEFAULT_CGROUP_V2_PATH = '/proc/self/mountinfo';
+  private readonly UTF8_UNICODE = 'utf8';
+  private readonly HOSTNAME = 'hostname';
+  private readonly MARKING_PREFIX = ['containers', 'overlay-containers'];
```
JamieDanielson: Nice cleanup, looks like those constants aren't used here and are already being exported in utils.

JamieDanielson (approved):
Should this be marked as breaking since there are some private modifiers being added in where they weren't previously?

pichlermarc (changes requested):
Looks good overall, but I think we may have to undo the changes from `public` to `private` in `packages/resource-detector-aws`, as the package is already stable.

legendecas:
There are `_` prefixed methods in `packages/resource-detector-aws` with implicit `public` accessibility. Should we mark these methods as `private`?

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/efc2c8da6f36d9620a58fa44e1d9ec6e8d478d27/packages/resource-detector-aws/src/detectors/AwsBeanstalkDetector.ts#L92

legendecas:
Actually, I think it makes sense to revert the `private` on implicit fields. But I kept the `_method` as `private`.

JamieDanielson (approved):

Issue #3251, opened 2025-11-26
@opentelemetry/instrumentation-aws-lambda - Breaks Lambda Node 24 runtime


RaphaelManke:
### What version of OpenTelemetry are you using?
- latest

### What version of Node are you using?
NodeJs 24 - AWS Lambda

### What did you do?

- Got to lambda console and create a new empty lambda with Nodesjs24

- add the layer eg.

- `arn:aws:lambda:eu-central-1:184161586896:layer:opentelemetry-nodejs-0_17_0:1`
invoke it with the test button

--> see the error

### What did you expect to see?

No error
### What did you see instead?

```
{
  "errorType": "Runtime.CallbackHandlerDeprecated",
  "errorMessage": "ERROR: AWS Lambda has removed support for callback-based function handlers starting with Node.js 24. You need to modify this function to use a supported handler signature to use Node.js 24 or later. For more information see https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html.",
  "trace": [
    "Runtime.CallbackHandlerDeprecated: ERROR: AWS Lambda has removed support for callback-based function handlers starting with Node.js 24. You need to modify this function to use a supported handler signature to use Node.js 24 or later. For more information see https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html.",
    "    at errorOnDeprecatedCallback (file:///var/runtime/index.mjs:617:11)",
    "    at createRuntime (file:///var/runtime/index.mjs:1222:5)",
    "    at async ignition (file:///var/runtime/index.mjs:1637:21)"
  ]
}
```

### Additional context

The issue is that AWS removed the callback support from the handler.
https://aws.amazon.com/blogs/compute/node-js-24-runtime-now-available-in-aws-lambda/

In particular, callback-based function handlers are no longer supported.
https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-aws-lambda

also reported here 
https://github.com/open-telemetry/opentelemetry-lambda/issues/2034
but since the instrumentation is part of this repo I recreated an issue here. 

This affects also the AWS layer for appsignal and ADOT

```
{
  "errorType": "Runtime.CallbackHandlerDeprecated",
[...]

PR #3255, opened 2025-11-26
feat(instrumentation-aws-lambda): add runtime-aware handler support for Node.js 24 compatibility

RaphaelManke:
## Description

This PR fixes issue #3251 by adding runtime-aware handler support that automatically detects the Node.js version from `AWS_EXECUTION_ENV` and adapts handler signatures accordingly.

**Fixes:** #3251

## Problem

AWS Lambda deprecated callback-based handlers in Node.js 24 runtime, causing the instrumentation to fail with:
```
Runtime.CallbackHandlerDeprecated: ERROR: AWS Lambda has removed support for callback-based function handlers starting with Node.js 24
```

## Solution

The instrumentation now automatically detects the Node.js runtime version and adapts:

- **Node.js 24+**: Handler signature is `(event, context)` - no callback parameter (respects AWS deprecation)
- **Node.js 22 and lower**: Handler signature is `(event, context, callback?)` - supports both callback and Promise-based handlers for backward compatibility

**I tested it by building a lambda layer with this code version.**

## Changes

- **Runtime Detection**: Added `getNodeRuntimeVersion()` function to detect Node.js version from `AWS_EXECUTION_ENV` environment variable
- **Runtime-Aware Handlers**: 
  - Node.js 24+: Only Promise-based handlers (callbacks not allowed)
  - Node.js 22 and lower: Both callback and Promise-based handlers supported
- **Backward Compatibility**: Re-added `_wrapCallback()` method for callback support on older runtimes
- **Tests**: Added callback-based handler tests that run when simulating Node.js 22 runtime

## Testing

- ✅ All 53 tests passing (including 3 new callback-based handler tests)
- ✅ Code coverage: 93.28% statements, 83.11% branches, 97.07% functions, 96.87% lines
- ✅ No linting errors
- ✅ Verified backward compatibility on Node.js 22 and lower
- ✅ Verified Node.js 24+ compliance (no callback parameter)

## Backward Compatibility

This change maintains full backward compatibility:
- Existing callback-based handlers continue to work on Node.js 22 and lower
- Promise-based handlers work on all Node.js versions
- Node.js 24+ enforces Promise-based handlers only (as per AWS Lambda deprecation)

## Related Links

- AWS Lambda Node.js 24 announcement: https://aws.amazon.com/blogs/compute/node-js-24-runtime-now-available-in-aws-lambda/
- AWS Lambda handler documentation: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html
- Related issue in opentelemetry-lambda: open-telemetry/opentelemetry-lambda#2034


packages/instrumentation-aws-lambda/README.md
```diff
@@ -23,6 +23,32 @@ npm install --save @opentelemetry/instrumentation-aws-lambda
...
+
+Example handlers:
+```js
+// Callback-based handler (Node.js 22 and lower only)
+exports.handler = function(event, context, callback) {
+  callback(null, 'ok');
+};
+
+// Promise-based handler (all Node.js versions)
+exports.handler = async function(event, context) {
```
maxday: We recommend using the ESM syntax, can you please update it:
```nodejs
export const handler = async (event, context) => {
```

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -272,47 +291,105 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
-
-      const span = plugin._createSpanForRequest(
-        event,
-        context,
-        requestIsColdStart,
-        parent
-      );
+    // Determine runtime version to decide whether to support callbacks
+    const nodeVersion = getNodeRuntimeVersion();
+    const supportsCallbacks = nodeVersion === null || nodeVersion < 24;
```
maxday: since we don't use `getNodeRuntimeVersion` elsewhere, what do you think about removing the fact that it can return null and rename it to "isSupportingCallbacks()"
something like
```nodejs
const isSupportingCallbacks = (): boolean {
  const executionEnv = process.env.AWS_EXECUTION_ENV;
  if (!executionEnv) {
    return true;
  }

  // AWS_EXECUTION_ENV format: AWS_Lambda_nodejs24.x, AWS_Lambda_nodejs22.x, etc.
  const match = executionEnv.match(/AWS_Lambda_nodejs(\d+)\./);
  if (match && match[1]) {
    return parseInt(match[1], 10) < 24
  }

  return true;
}
```

packages/instrumentation-aws-lambda/CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
 # Changelog
 
+## [Unreleased]
+
+### Features
+
+* **instrumentation-aws-lambda:** Added runtime-aware handler support. The instrumentation automatically detects the Node.js runtime version from `AWS_EXECUTION_ENV` and adapts handler signatures accordingly:
+  - **Node.js 24+**: Only Promise-based handlers are supported (callbacks deprecated by AWS Lambda)
+  - **Node.js 22 and lower**: Both callback-based and Promise-based handlers are supported for backward compatibility
+  This ensures seamless operation across different Node.js runtime versions while respecting AWS Lambda's deprecation of callbacks in Node.js 24+.
```
trentm: nit: 
```suggestion
  This ensures seamless operation across different Node.js runtime versions while respecting AWS Lambda's removal of callbacks in Node.js 24+.
```

maxday (commented):
Thanks @RaphaelManke 
I left minor comments, feel free to ignore

packages/instrumentation-aws-lambda/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@opentelemetry/instrumentation-aws-lambda",
-  "version": "0.60.1",
+  "version": "0.61.0",
```
trentm: Don't update the version in this PR. The updating of package versions is handled by the release-please tooling that does releases.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -71,6 +71,25 @@ export const AWS_HANDLER_STREAMING_SYMBOL = Symbol.for(
 );
 export const AWS_HANDLER_STREAMING_RESPONSE = 'response';
 
+/**
+ * Detects the Node.js runtime version from AWS_EXECUTION_ENV environment variable.
+ * Returns the major version number (e.g., 24, 22, 20) or null if not detected.
+ */
+function getNodeRuntimeVersion(): number | null {
```
trentm: Would an alternative be to check `process.version`? This always runs in the same process as the lambda.  I'm not sure which is "more" correct.

Ah, at least this won't work with the test case below.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -443,11 +520,11 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
   private _wrapCallback(original: Callback, span: Span): Callback {
     const plugin = this;
     return function wrappedCallback(this: never, err, res) {
-      diag.debug('executing wrapped lookup callback function');
+      diag.debug('AWS Lambda instrumentation: Executing wrapped callback function');
```
trentm: nit:

```suggestion
      plugin._diag.debug('executing wrapped callback function');
```

Instrumentations (they used to be called Plugins, hence the `plugin` var name here), have a namespaced diag logger that includes the instrumentation name in log output (https://github.com/open-telemetry/opentelemetry-js/blob/167185afbabd20c3bf60ce4827a73dbc88695bec/experimental/packages/opentelemetry-instrumentation/src/instrumentation.ts#L58-L60).

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -443,11 +520,11 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
       plugin._applyResponseHook(span, err, res);
 
       plugin._endSpan(span, err, () => {
-        diag.debug('executing original lookup callback function');
+        diag.debug('AWS Lambda instrumentation: Executing original callback function');
```
trentm: nit:

```suggestion
        plugin._diag.debug('executing original callback function');
```

packages/instrumentation-aws-lambda/test/lambda-test/sync.js
```diff
@@ -15,26 +15,40 @@
...
+exports.callbackerror = async function (event, context) {
+  throw new Error('handler error');
+};
+
+exports.stringerror = async function (event, context) {
+  throw 'handler error';
 };
 
-exports.stringerror = function (event, context, callback) {
+exports.callbackstringerror = async function (event, context) {
```
trentm: nit: I suppose the "callback" name in this test fixture is a bit out of place now.

maxday:
Confirming it's working fine by 
1. building this package
2. doing some magic with `npm pack`
3. modifying https://github.com/open-telemetry/opentelemetry-lambda to use the packed node_modules
4. running `npm run build` from the `nodejs` folder in https://github.com/open-telemetry/opentelemetry-lambda
5. uploading the layer to my test AWS account
6. creating a simple hello world function using nodejs 24
7. attaching the layer
8. setting `AWS_LAMBDA_EXEC_WRAPPER="/opt/otel-handler"`

was failing before, succeeding with this change

maxday (approved):
Thanks @RaphaelManke 💯

packages/instrumentation-aws-lambda/test/lambda-test/sync.js
RaphaelManke: yeah this is a leftover from an intermediate state. Deleted it as its covered in another test

packages/instrumentation-aws-lambda/test/lambda-test/sync.js
```diff
@@ -15,26 +15,36 @@
...
 };
 
-exports.error = function (event, context, callback) {
+exports.error = async function (event, context) {
   throw new Error('handler error');
 };
 
-exports.callbackerror = function (event, context, callback) {
-  callback(new Error('handler error'));
+exports.callbackerror = async function (event, context) {
```
trentm, line 27: nit: have exports.callbackerror and exports.callbackError :)    This one here is currently a bit misleading. Minor though.

Issue #3258, opened 2025-11-28
Postgres instrumentation crashes when sql-template-strings is used


jaysonsantos:
### What version of OpenTelemetry are you using?
ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-nodejs:0.67.2
instrumentation-pg as of 0.61.1 (grep version /otel-auto-instrumentation-nodejs/node_modules/@opentelemetry/instrumentation-pg/package.json)

### What version of Node are you using?
v22.7.0

### What did you do?
The original error is from postgres-migrations but if you use sql-template-strings as the examples show, this can be triggered

```javascript
import { Client } from "pg";
import SQL from "sql-template-strings";

const client = new Client();
await client.connect();

const name = "any";
client
  .query(
    SQL`SELECT * FROM information_schema.tables WHERE table_name = ${name}`,
  )
  .then((result) => {
    console.log(result.rowCount);
    process.exit(0);
  });

```

### What did you expect to see?
```
node index.js
0
```

### What did you see instead?

```
/private/tmp/bug-report/autoinstrumentation/node_modules/@opentelemetry/instrumentation-pg/build/src/utils.js:80
  const indexOfFirstSpace = queryText.indexOf(" ");
                                      ^

TypeError: Cannot read properties of undefined (reading 'indexOf')
    at parseNormalizedOperationName (/private/tmp/bug-report/autoinstrumentation/node_modules/@opentelemetry/instrumentation-pg/build/src/utils.js:80:39)
    at getQuerySpanName (/private/tmp/bug-report/autoinstrumentation/node_modules/@opentelemetry/instrumentation-pg/build/src/utils.js:73:9)
    at Client.handleConfigQuery (/private/tmp/bug-report/autoinstrumentation/node_modules/@opentelemetry/instrumentation-pg/build/src/utils.js:209:20)
    at Client.query (/private/tmp/bug-report/autoinstrumentation/node_modules/@opentelemetry/instrumentation-pg/build/src/instrumentation.js:288:46)
    at file:///private/tmp/bug-report/index.js:9:4
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```
### Additional context

Here are the types that it returns

```
ARGS ARE: [
  SQLStatement {
    strings: [
      'SELECT * FROM information_schema.tables WHERE table_name = ',
      ''
    ],
    values: [ 'any' ]
  }
]
Query Config: {
[...]

PR #3259, opened 2025-11-29
fix(instrumentation-pg): preserve SQL template text property

aryamohanan:
closes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3258

Fixes crash in the pg instrumentation when using SQL template literal library ([sql-template-strings](https://www.npmjs.com/package/sql-template-strings)). 

`TypeError: Cannot read properties of undefined (reading 'indexOf')
    at parseNormalizedOperationName`
    
The issue was introduced by #3196 , where spreading query objects caused SQL template objects to lose their `text` getter, leading to `parseNormalizedOperationName()` receiving `undefined` and throwing the error.

- Preserve the original SQL template object when building `queryConfig` instead of spreading it
- Add a check to assign `values` from the second argument only if missing
- Add a test mocking SQL template objects to verify that `client.query(SQL\`...\`)` works
  correctly and spans are recorded

### DIFF
--- a/packages/instrumentation-pg/test/pg.test.ts
+++ b/packages/instrumentation-pg/test/pg.test.ts
@@ -1040,7 +1040,7 @@ describe('pg', () => {
       });
     });
 
-    it('should handle SQL template literal objects', async () => {
+    it('should generate traces for SQL template literal objects ', async () => {
       const span = tracer.startSpan('test span');
       await context.with(trace.setSpan(context.active(), span), async () => {
         // Mock SQLStatement from sql-template-strings library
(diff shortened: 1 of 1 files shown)


packages/instrumentation-pg/src/instrumentation.ts
```diff
@@ -339,13 +339,15 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
...
+          ? (() => {
+              const query = arg0 as any;
+              // If the query object has no values yet, use the second argument as values
+              if (query.values === undefined && Array.isArray(args[1])) {
+                query.values = args[1];
               }
-            : undefined;
+              return query;
+            })()
+          : undefined;
```
raphael-theriault-swi: I already did not like this ternary soup, so if you don't mind I'd like to use this opportunity to refactor it into something more readable rather than adding an IIFE into the mix 😅

raphael-theriault-swi (changes requested):

packages/instrumentation-pg/src/instrumentation.ts
aryamohanan: Absolutely, a clean refactor sounds like the right move 👍

packages/instrumentation-pg/src/instrumentation.ts
raphael-theriault-swi, line 390: Another thing, there is another spread happening here. Could you add a test checking SQL template literals with added comments too ?

aryamohanan: Definitely, I’ll expand the test to cover SQL template literals with added comments too. Appreciate you catching the spread.

PR #3268, opened 2025-12-03
feat(instrumentation-net): support `net.*` semconv migration

trentm:
This adds support for using `OTEL_SEMCONV_STABILITY_OPT_IN` for controlled migration to stable `net.*` semconv. The `net.*` attributes are controlled by the `http[/dup]` token in `OTEL_SEMCONV_STABILITY_OPT_IN` (as [discussed here](https://github.com/open-telemetry/opentelemetry-js/issues/5663#issuecomment-3314043915)).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38531,7 +38531,8 @@
       "version": "0.52.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@opentelemetry/instrumentation": "^0.208.0"
+        "@opentelemetry/instrumentation": "^0.208.0",
+        "@opentelemetry/semantic-conventions": "^1.33.0"
       },
       "devDependencies": {
         "@opentelemetry/api": "^1.3.0",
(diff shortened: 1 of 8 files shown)


trentm:
Note: The `tls.*` attributes added to "tls-connect" spans are all waay out of date. It was added while `tls.*` semconv was being discussed. The currently [TLS semconv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/tls/) has somewhat different values. However, [as similar to this PR to instr-amqplib](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2976#issuecomment-3415202969) the `tls.*` semconv are not stable so we cannot use `OTEL_SEMCONV_STABILITY_OPT_IN=http` to handle the migration of those attributes.

These should still be updated, IMO. I have some notes on that. It is a TODO on me to open an issue for that.

david-luna (approved):

PR #3274, opened 2025-12-05
chore(lint): restore running prettier as part of linting

trentm:
This re-adds running prettier as part of the 'lint' step. There are separate
`lint:prettier` and `lint:prettier:fix` scripts if one wants to run just style checking.

Refs: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3001

---

AFAICT prettier is no longer being run. My theory is that #3001 accidentally caused that. 

Update: The accident is that we didn't follow-up after 3001 to add running `prettier` *separately*. Prettier's [recommendation is to run prettier separate from linters](https://prettier.io/docs/integrating-with-linters#notes), which I can get behind.

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,10 +1,6 @@
 {
-  "extension": [
-    ".ts"
-  ],
-  "include": [
-    "src/**"
-  ],
+  "extension": [".ts"],
+  "include": ["src/**"],
   "reporter": ["text", "json"],
   "exclude": [
     "**/*.d.ts",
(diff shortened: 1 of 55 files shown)


trentm:
Slack discussion: https://cloud-native.slack.com/archives/C08T7MZTV8W/p1764893850195369

david-luna (approved):
TIL `declare private` & `private declare` is allowed by the compiler.

PR #3287, opened 2025-12-09
feat(instrumentation-graphql): add parent name in attributes of resolver span

antogyn:

I am trying to derive performance metrics from GraphQL traces.

Currently, I cannot consistently measure the performance of a specific, logical resolver, identified by the combination of its parent type name and field name.

While the existing `graphql.field.path` attribute is available, it is query-path dependent. Consider this simplified schema and two queries hitting the same `name` resolver:

```graphql
type Account {
  name: String! # <-- Logical Resolver: Account.name
}
type Query {
  account(id: ID!): Account # Query 1
  user(id: ID!): User
}
type User {
  accounts: [Account!]! # Query 2
}
```

Execution | Field Name (graphql.field.name) | Field Path (graphql.field.path)
-- | -- | --
Query 1: query { account { name } } | name | Query.account.name
Query 2: query { user { accounts { name } } } | name | Query.user.accounts.name

Both queries execute the same `Account.name` resolver, but the existing `graphql.field.path` differs. This makes it impossible to aggregate all performance data for `Account.name` using the existing attributes.

This PR adds the parent type name to a new attribute on resolver spans

1.  I've added `graphql.parent.name` to the `AttributeNames` enum.
2. The instrumentation records the parent type name using `info.parentType.name` in this new attribute.


obecny (approved):
lgtm, thx for contribution

pichlermarc:
huh - looks like the workflows did not run so this did not get merged. I merged `main` and approved the workflow run so that we can see if tests pass; It should auto-merge then.

PR #3288, opened 2025-12-10
feat(instrumentations-mysql): support `net.*` and database semconv migration

trentm:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,12 +16,14 @@
 .github/ @open-telemetry/javascript-maintainers
 
 # Any browser-specific code is co-owned by Browser SIG and JS SIG (list more packages here as needed):
+examples/web/ @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
+packages/auto-instrumentations-web @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
+packages/instrumentation-browser-navigation @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
 packages/instrumentation-document-load/ @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
 packages/instrumentation-long-task/ @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
 packages/instrumentation-user-interaction/ @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
 packages/instrumentation-web-exception/ @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
 packages/plugin-react-load/ @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
-packages/auto-instrumentations-web @open-telemetry/browser-maintainers @open-telemetry/javascript-approvers
 
 # All other files
 * @open-telemetry/javascript-approvers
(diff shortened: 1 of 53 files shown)


trentm:
^^ Took this into draft because after some discussion it was pointed out that we should not add the not-yet-stable `db.client.connection.count` metric under the "gimme the stable database semconv" banner (aka `OTEL_SEMCONV_STABILITY_OPT_IN=database`)

The plan: (tl;dr: not sure)
- Remove the addition of the `db.client.connection.count` metric in this PR.
- Q: Should the existing `db.client.connections.usage` metric -- which is not current semconv, and has an alternative in [the db-migration](https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/#database-client-connection-count) -- be **removed** if `OTEL_SEMCONV_STABILITY_OPT_IN=database` is specified? A: The migration docs *do* say "`database` - emit the stable database conventions, and stop emitting the old database conventions that the instrumentation emitted previously", so I would say *yes*.
- Likely re-add `db.client.connection.count`, likely in a separate PR. I'm not sure if that should be opt-in, because semconv marks this metric (among others) as "status:development" but also *"Required"*. Sigh.   Are we meant to read that "Required" as (a) "not opt-in, i.e. on by default" or (b) "ignore the 'Required' until this is marked as *stable*"?
- Consider making both of the metrics optional... whether via specific separate programmatic config options to the instrumentation, or a single generic config option, and/or an envvar that might be generic to Node.js instrs (say `OTEL_NODE_I_CAN_HAZ_THESE_EXPERIMENTAL_SEMCONV_THINGS=db.client.connections.usage,db.client.connection.count`). Unsure.


Update: Moved the followup parts of this "plan" to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3291

packages/instrumentation-mysql/test/index.metrics.test.ts
```diff
@@ -93,20 +105,14 @@ describe('mysql@2.x-Metrics', () => {
...
   describe('#Pool - metrics', () => {
-    let pool: mysqlTypes.Pool;
+    let pool: Pool;
 
     beforeEach(() => {
       initMeterProvider();
       instrumentation.disable();
       instrumentation.enable();
       inMemoryMetricsExporter.reset();
-      // credentials to connect to pool if you run docker locally using 'npm run docker:start' from 'examples/mysql' folder
```
trentm, line 103: Reviewer note: The `docker:start` script was removed a long while back. This comment is obsolete.

packages/instrumentation-mysql/test/index.metrics.test.ts
```diff
@@ -122,158 +128,78 @@ describe('mysql@2.x-Metrics', () => {
...
+      conn.release();
+
+      const exportedMetrics = await waitForNumberOfExports(
+        inMemoryMetricsExporter,
+        2
+      );
+      assert.strictEqual(exportedMetrics.length, 2);
+      assertMetrics(exportedMetrics[1].scopeMetrics[0].metrics, {
+        expectedConnCountIdle: 1,
+        expectedConnCountUsed: 0,
```
trentm, line 144: Reviewer note: Two main structural changes in this file:
1. Each test was changed to use a `const thing = await promisify(...` style. Before this change a *passing* test would be fine, but if one of the assertions *failed* the AssertionError would not get caught (because it was thrown in a deep callback function) and the test case would *timeout*. Finding the actual failed assertion could be a pain.
2. I moved a bunch of assertion code that was mostly common between all the test cases into `assertMetrics`.

trentm:
(It is frustrating the continued unrelated CI failures on this one.)

Issue #3286, opened 2025-12-09
@opentelemetry/instrumentation-openai *crashes* the app when `client.chat.completions.create({stream: true, ...}).withResponse()` is used


lucasvieirasilva:
### What version of OpenTelemetry are you using?

- `@opentelemetry/auto-instrumentations-node`: `0.67.2`
- `@opentelemetry/instrumentation-openai`: `0.7.0`

### What version of Node are you using?

v24.2.0

### What did you do?

When applying the `getNodeAutoInstrumentations()` instrumentation and using the OpenAI SDK as follows:

```typescript
const response = await client.chat.completions
        .create(requestBody)
        .withResponse();
```
I got the following error:

`client.chat.completions.create(...).withResponse is not a function`

### What did you expect to see?

The `@opentelemetry/instrumentation-openai` should also patch the `withResponse` function.

### What did you see instead?

I got the following error:

`client.chat.completions.create(...).withResponse is not a function`

### Additional context

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. [Learn more here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

trentm:
@lucasvieirasilva I cannot reproduce this. Here is what I tried:

#### package.json

```json
{
  "name": "asdf.20251210t122838",
  "version": "1.0.0",
  "license": "Apache-2.0",
  "dependencies": {
    "@opentelemetry/auto-instrumentations-node": "^0.67.2",
    "openai": "^6.10.0"
  }
}
```

#### use-openai.js

```js
const {execSync} = require('child_process');
const {OpenAI} = require('openai');

console.log('node version:', process.version);
execSync('npm ls openai', {stdio: 'inherit'});

async function main() {
  const client = new OpenAI();
  const requestBody = {
    model: 'gpt-5.1',
    messages: [
      {
        role: 'user',
        content: 'Hi',
      },
    ]
  };
  const retval = await client.chat.completions
    .create(requestBody)
    .withResponse();
  console.log('retval: ', retval);
}

main();
```

#### telemetry.mjs

```js
import os from 'os';
import path from 'path';
import {NodeSDK} from '@opentelemetry/sdk-node';
[...]

PR #3295, opened 2025-12-12
fix(instrumentation-openai): fix crash when user uses .withResponse() with streaming chat completion

trentm:
Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3286

### DIFF
--- a/packages/instrumentation-openai/src/instrumentation.ts
+++ b/packages/instrumentation-openai/src/instrumentation.ts
@@ -226,7 +226,11 @@ export class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumenta
               return stream;
             })
             .catch(
-              self._createAPIPromiseRejectionHandler(startNow, span, commonAttrs)
+              self._createAPIPromiseRejectionHandler(
+                startNow,
+                span,
+                commonAttrs
+              )
             );
           // We have to return the original `APIPromise` instance, because it is
           // a customized subclass of `Promise`.


packages/instrumentation-openai/src/instrumentation.ts
```diff
@@ -208,22 +208,29 @@ export class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumenta
...
+                };
+              });
+              return stream;
+            })
+            .catch(
+              self._createAPIPromiseRejectionHandler(startNow, span, commonAttrs)
+            );
+          // We have to return the original `APIPromise` instance, because it is
+          // a customized subclass of `Promise`.
+          return apiPromise;
```
trentm, line 237: Review pro-tip: Look at the diff ignoring whitespace. Most of this is an indentation change.

packages/instrumentation-openai/test/mock-responses/openai-streaming-chat-completions-copes-with-withresponse-usage.json
trentm, line 1: This file is a copy of the existing "openai-streaming-chat-completions-adds-genai-conventions.json" mock-responses file. The way the test suite works requires a separate mock response file per test case (based on the test case name).

packages/instrumentation-openai/test/openai.test.ts
```diff
@@ -96,7 +96,7 @@ describe('OpenAI', function () {
     const filename = `${this.currentTest
       ?.fullTitle()
       .toLowerCase()
-      .replace(/\s/g, '-')}.json`;
+      .replace(/[^a-z0-9]+/g, '-')}.json`;
```
trentm, line 99: Reviewer note: this improves the slug algorithm so that the test/mock-responses/ filename for test name `'copes with .withResponse() usage'` is
`openai-streaming-chat-completions-copes-with-withresponse-usage.json`
rather than the painful-for-the-file-system:
`openai-streaming-chat-completions-copes-with-.withresponse()-usage.json`

PR #3302, opened 2025-12-16
feat(instr-cassandra-driver): support net.* and db semconv migration

JamieDanielson:

- Refs: https://github.com/open-telemetry/opentelemetry-js/issues/5663

### DIFF
--- a/packages/instrumentation-cassandra-driver/test/cassandra-driver.test.ts
+++ b/packages/instrumentation-cassandra-driver/test/cassandra-driver.test.ts
@@ -81,12 +81,12 @@ const DEFAULT_STABLE_ATTRIBUTES = {
   [ATTR_DB_SYSTEM_NAME]: DB_SYSTEM_NAME_VALUE_CASSANDRA,
 };
 
-const ADDRESS_OLD_ATTRIBUTES = {
+const NET_OLD_ATTRIBUTES = {
   [ATTR_NET_PEER_NAME]: cassandraContactPoint,
   [ATTR_NET_PEER_PORT]: 9042,
 };
 
-const ADDRESS_STABLE_ATTRIBUTES = {
+const NET_STABLE_ATTRIBUTES = {
   [ATTR_SERVER_ADDRESS]: cassandraContactPoint,
   [ATTR_SERVER_PORT]: 9042,
 };
@@ -102,8 +102,8 @@ function assertSpan(
   const attributes: Attributes = {
     ...DEFAULT_OLD_ATTRIBUTES,
     ...DEFAULT_STABLE_ATTRIBUTES,
-    ...(includeAddressAttrs ? ADDRESS_OLD_ATTRIBUTES : {}),
-    ...(includeAddressAttrs ? ADDRESS_STABLE_ATTRIBUTES : {}),
+    ...(includeAddressAttrs ? NET_OLD_ATTRIBUTES : {}),
+    ...(includeAddressAttrs ? NET_STABLE_ATTRIBUTES : {}),
     ...customAttributes,
   };
 
@@ -155,8 +155,8 @@ function assertErrorSpan(
   const attributes: Attributes = {
     ...DEFAULT_OLD_ATTRIBUTES,
     ...DEFAULT_STABLE_ATTRIBUTES,
-    ...(includeAddressAttrs ? ADDRESS_OLD_ATTRIBUTES : {}),
-    ...(includeAddressAttrs ? ADDRESS_STABLE_ATTRIBUTES : {}),
+    ...(includeAddressAttrs ? NET_OLD_ATTRIBUTES : {}),
+    ...(includeAddressAttrs ? NET_STABLE_ATTRIBUTES : {}),
     ...customAttributes,
   };
 
@@ -461,7 +461,7 @@ describe('CassandraDriverInstrumentation', () => {
         SpanKind.CLIENT,
         {
           ...DEFAULT_OLD_ATTRIBUTES,
-          ...ADDRESS_OLD_ATTRIBUTES,
+          ...NET_OLD_ATTRIBUTES,
         },
         [],
         { code: SpanStatusCode.UNSET }
@@ -482,7 +482,7 @@ describe('CassandraDriverInstrumentation', () => {
         SpanKind.CLIENT,
         {
           ...DEFAULT_STABLE_ATTRIBUTES,
-          ...ADDRESS_STABLE_ATTRIBUTES,
+          ...NET_STABLE_ATTRIBUTES,
         },
         [],
         { code: SpanStatusCode.UNSET }


packages/instrumentation-cassandra-driver/src/instrumentation.ts
```diff
@@ -304,24 +340,39 @@ export class CassandraDriverInstrumentation extends InstrumentationBase<Cassandr
...
-    const attributes: Attributes = {
-      [ATTR_DB_SYSTEM]: DB_SYSTEM_VALUE_CASSANDRA,
-    };
+    const attributes: Attributes = {};
+
+    if (this._dbSemconvStability & SemconvStability.OLD) {
+      attributes[ATTR_DB_SYSTEM] = DB_SYSTEM_VALUE_CASSANDRA;
+    }
+    if (this._dbSemconvStability & SemconvStability.STABLE) {
+      attributes[ATTR_DB_SYSTEM_NAME] = DB_SYSTEM_VALUE_CASSANDRA;
```
trentm: nit:

```suggestion
      attributes[ATTR_DB_SYSTEM_NAME] = DB_SYSTEM_NAME_VALUE_CASSANDRA;
```

would have to add it to src/semconv.ts and import it as well.

packages/instrumentation-cassandra-driver/src/instrumentation.ts
```diff
@@ -234,10 +266,12 @@ export class CassandraDriverInstrumentation extends InstrumentationBase<Cassandr
 
         const batchPromise = safeExecuteInTheMiddle(
           () => {
-            return original.apply(
-              this,
-              args
-            ) as Promise<CassandraDriver.types.ResultSet>;
+            return context.with(batchContext, () => {
```
trentm, line 270: Did tests show that the context wasn't correct? Curious how you noticed this.

packages/instrumentation-cassandra-driver/test/cassandra-driver.test.ts
```diff
@@ -63,21 +71,44 @@ const cassandraContactPoint = process.env.CASSANDRA_HOST
...
+const DEFAULT_OLD_ATTRIBUTES = {
+  [ATTR_DB_SYSTEM]: DB_SYSTEM_VALUE_CASSANDRA,
+  [ATTR_DB_USER]: 'cassandra',
+};
+
+const DEFAULT_STABLE_ATTRIBUTES = {
+  [ATTR_DB_SYSTEM_NAME]: DB_SYSTEM_VALUE_CASSANDRA,
+};
+
+const ADDRESS_OLD_ATTRIBUTES = {
```
trentm: super pedant nit: I'd use "NET_" rather than "ADDRESS_" for these constants.

packages/instrumentation-cassandra-driver/src/instrumentation.ts
```diff
@@ -304,24 +340,39 @@ export class CassandraDriverInstrumentation extends InstrumentationBase<Cassandr
...
-      attributes[ATTR_DB_NAME] = client.keyspace;
+      if (this._dbSemconvStability & SemconvStability.OLD) {
+        attributes[ATTR_DB_NAME] = client.keyspace;
+      }
+      if (this._dbSemconvStability & SemconvStability.STABLE) {
+        attributes[ATTR_DB_NAMESPACE] = client.keyspace;
+      }
     }
 
     return this.tracer.startSpan(`cassandra-driver.${op}`, {
```
trentm, line 379: A digression. I had started on the semconv update for instr-cassandra-driver as well -- we should coordinate ;), all good here though because I hadn't gotten that far. I got pulled into thinking about the span **name**. In this or a separate PR I think we should consider redoing the span name when the stable semconv is selected. (Same for some other instrumentations.)

The stable spec for DB span names (https://opentelemetry.io/docs/specs/semconv/database/database-spans/#name) is a little head spinning -- a little bit of jumping around to a few places in the spec. At least I found it so. I *think* the unfortunate result for cassandra will be that all span names will be `${client.keyspace}`.

It is *possible* a quickish and defensive parse of the CQL syntax (https://cassandra.apache.org/doc/4.0/cassandra/cql/definitions.html) could be done and the `db.operation.name` extracted from the first 1-3 tokens.
- The simplistic "first space-delim token" used by instr-mysql2 (and, I think instr-pg too) could be misleading: fine for common `SELECT`, `INSERT`, etc. (https://cassandra.apache.org/doc/4.0/cassandra/cql/dml.html), but misleading for `DROP`, etc. See https://cassandra.apache.org/doc/4.0/cassandra/cql/ddl.html which use two tokens.
- Batch ops may use *3* tokens: `BEGIN [ UNLOGGED | COUNTER ] BATCH`
- *warning*: parse out comments! (three possible forms)

Thoughts?

trentm (approved):

packages/instrumentation-cassandra-driver/src/instrumentation.ts
trentm: > I _think_ the unfortunate result for cassandra will be that all span names will be `${client.keyspace}`.

Actually, I think I'm wrong. The keyspace should only be the `{target}` when we know the query is acting directly on the keyspace. E.g. `DROP KEYSPACE my_keyspace` or whatever.  We aren't parsing the query CQL (yet) so we don't know the appropriate `{target}` per https://opentelemetry.io/docs/specs/semconv/database/database-spans/#name

That means the fallback span.name will be just `cassandra`.

packages/instrumentation-cassandra-driver/src/instrumentation.ts
JamieDanielson: Yeah tests were failing without it

packages/instrumentation-cassandra-driver/src/instrumentation.ts
JamieDanielson: Not sure it's a nit, this is technically incorrect! Updated

packages/instrumentation-cassandra-driver/test/cassandra-driver.test.ts
JamieDanielson: Makes sense, updated

packages/instrumentation-cassandra-driver/src/instrumentation.ts
JamieDanielson: Trent and I discussed this a bit and think it will take quite a bit of digging in each of the databases to get this right for all of them. We will update that in a separate PR.

JamieDanielson: I created an issue for follow-up https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3309

package-lock.json
```diff
@@ -37822,15 +37822,15 @@
       "version": "0.54.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@opentelemetry/instrumentation": "^0.208.0"
+        "@opentelemetry/instrumentation": "^0.208.0",
+        "@opentelemetry/semantic-conventions": "^1.37.0"
```
trentm, line 37826: For this monumentus change, thou shalt pay with 4h of CI compute time.

JamieDanielson: 😂 😭

PR #3307, opened 2025-12-16
feat(instr-mongoose): support net.* and db semconv migration

JamieDanielson:

- Refs: https://github.com/open-telemetry/opentelemetry-js/issues/5663

### DIFF
--- a/packages/instrumentation-mongoose/test/mongoose-common.test.ts
+++ b/packages/instrumentation-mongoose/test/mongoose-common.test.ts
@@ -989,9 +989,9 @@ describe('mongoose instrumentation [common]', () => {
       checkCollector: (collector: testUtils.TestCollector) => {
         const spans = collector.sortedSpans;
         assert.strictEqual(spans[0].name, 'manual');
-        assert.strictEqual(spans[1].name, 'mongoose.Test.save');
+        assert.strictEqual(spans[1].name, 'save tests');
         assert.strictEqual(spans[1].parentSpanId, spans[0].spanId);
-        assert.strictEqual(spans[2].name, 'mongoose.Test.findOne');
+        assert.strictEqual(spans[2].name, 'findOne tests');
         assert.strictEqual(spans[2].parentSpanId, spans[0].spanId);
       },
     });


packages/instrumentation-mongoose/README.md
```diff
@@ -40,20 +40,29 @@ registerInstrumentations({
...
-| `db.name`               | This attribute is used to report the name of the database being accessed.   |
-| `db.operation`          | The name of the operation being executed, or the SQL keyword.               |
-| `db.statement`          | The database statement being executed.                                      |
-| `db.system`             | An identifier for the database management system (DBMS) product being used. |
-| `db.user`               | Username for accessing the database.                                        |
-| `net.peer.name`         | Remote hostname or similar.                                                 |
-| `net.peer.port`         | Remote port number.                                                         |
+| Old semconv             | Stable semconv        | Description                                                                 |
+| ----------------------- | --------------------- | --------------------------------------------------------------------------- |
+| `db.system`             | `db.system.name`      | An identifier for the database management system (DBMS) product being used. Value: 'mongoose' |
```
JamieDanielson: Should this be value mongodb? It is currently mongoose but I think it should be mongodb based on my read of the [database semconv docs](https://opentelemetry.io/docs/specs/semconv/database/database-spans/).

packages/instrumentation-mongoose/src/mongoose.ts
```diff
@@ -426,16 +467,29 @@ export class MongooseInstrumentation extends InstrumentationBase<MongooseInstrum
...
+      ),
+    };
+
+    if (this._dbSemconvStability & SemconvStability.OLD) {
+      finalAttributes[ATTR_DB_OPERATION] = operation;
+      finalAttributes[ATTR_DB_SYSTEM] = 'mongoose';
+    }
+    if (this._dbSemconvStability & SemconvStability.STABLE) {
+      finalAttributes[ATTR_DB_OPERATION_NAME] = operation;
+      finalAttributes[ATTR_DB_SYSTEM_NAME] = 'mongoose';
```
JamieDanielson: Same question here as noted in the readme above. Should this be value mongodb? It is currently mongoose but I think it should be mongodb based on my read of the [database semconv docs](https://opentelemetry.io/docs/specs/semconv/database/database-spans/).

packages/instrumentation-mongoose/README.md
trentm: Yes, I think it should be mongodb as well.

packages/instrumentation-mongoose/src/mongoose.ts
trentm: Ditto. Should be `mongodb` IMO.

packages/instrumentation-mongoose/README.md
trentm: Hrm... or maybe not. Perhaps arguably this isn't really a DB. It is an ORM.  I don't know if there is any prior art for semconv for ORMs in other languages. I suppose we do have instrumentation-typeorm. It is using the underlying "type" (per this list: https://typeorm.io/docs/data-source/data-source-options#common-data-source-options) -- which is close to the semconv-defined values for the `db.system.name` enum.

So this is a gray area. I think we just do our best.

The "operations" here are Mongoose-y names, I thnk, e.g. https://mongoosejs.com/docs/queries.html
which differ from mongodb-y operation names. So, what's best for users here? If `mongoose.` is taken out of the span name -- as I think it should be -- and if `db.system.name: 'mongodb'`, then the only indicator that this is "Mongoose" is (a) recognizing the operation names and (b) the instrumentationScope.name (which is indirectly `@opentelemetry/instrumentation-mongoose`.

I'm still inclined to use 'mongodb'. :)

packages/instrumentation-mongoose/src/mongoose.ts
```diff
@@ -426,16 +467,29 @@ export class MongooseInstrumentation extends InstrumentationBase<MongooseInstrum
...
+    }
+
     return this.tracer.startSpan(
       `mongoose.${modelName}.${operation}`,
```
trentm: I was going to make some suggestion for changing span.name... but when I was trying to look at what `db.collection.name` is set to I see *uuids* in that value. I.e. definitely not low cardinality. I'm not really sure what is going on. I don't know Mongoose well. Could instr-mongoose be *wrong* in getting `db.collection.name` the way it currently is?

From the `instrumenting save operation with promise` test case in mongoose-common.test.ts, this mongoose-using code:

```ts
      const document = {
        firstName: 'Test first name',
        lastName: 'Test last name',
        email: 'test@example.com',
      };
      const user: IUser = new User(document);

      await user.save();
```

Results in a span like this:

```json
    attributes: {
      'db.statement': '{"document":{"email":"test@example.com","firstName":"Test first name","lastName":"Test last name","_id":"69436c39e1817a5978cf5796"}}',
      'db.query.text': '{"document":{"email":"test@example.com","firstName":"Test first name","lastName":"Test last name","_id":"69436c39e1817a5978cf5796"}}',
      'db.mongodb.collection': 'db5f0a98-dfb5-45fc-9bd5-712b1b56299fusers',
      'db.name': 'opentelemetry-tests',
      'db.collection.name': 'db5f0a98-dfb5-45fc-9bd5-712b1b56299fusers',
      'db.namespace': 'opentelemetry-tests',
      'net.peer.name': '127.0.0.1',
      'net.peer.port': 27017,
      'server.address': '127.0.0.1',
      'server.port': 27017,
      'db.operation': 'save',
      'db.system': 'mongoose',
      'db.operation.name': 'save',
      'db.system.name': 'mongoose'
    },
    name: 'mongoose.db5f0a98-dfb5-45fc-9bd5-712b1b56299fUser.save',
```

I think both `name` and `db.collection.name` should be looked into.
I don't know where the UUID is coming from.

Ultimately I think we'd (perhaps, I quibble because I don't use Mongoose) want:
- span.name: `save User` (following the `$operation $target` spec from https://opentelemetry.io/docs/specs/semconv/database/database-spans/#name)
- db.collection.name: `User`

Or that `User` could be `user` or `users`... whatever is consistently available from the Mongoose data.

packages/instrumentation-mongoose/README.md
JamieDanielson: 🤔  Hm. This is funky. It is breaking if I change the old value to 'mongodb' from 'mongoose', so I was going to leave that as-is for backwards compatibility... but if someone selects `database/dup` then which should win?

JamieDanielson: I'm inclined to say the new value (mongodb) should override the old (mongoose) in this particular case, if database/dup is selected. But the default will be to continue emitting the old (mongoose)

JamieDanielson: 🤦  Ignore me actually. They are different keys so if someone selects database/dup they will get `db.system: mongoose` and `db.system.name: mongodb`.

packages/instrumentation-mongoose/src/mongoose.ts
JamieDanielson: Huh, interesting. That name for User in this test really does have a UUID, but it's intentional. It's defined in [test/user.ts](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/a928555c4d144e94602a1eff6d4e87bc3476550b/packages/instrumentation-mongoose/test/user.ts#L35) to include a UUID, I guess to avoid collisions in tests. So the current name if it didn't have that would simply say `mongoose.User.save`, and so now that might be replaced with `save User`. And same with collection, that right now is really just `users` but the UUID is added in the test case.

```
const User = mongoose.model<IUser>(crypto.randomUUID() + 'User', UserSchema);
```

JamieDanielson: I'm removing the UUID from the test because it doesn't seem to have any benefit and only confuses.

trentm: > const User = mongoose.model<IUser>(crypto.randomUUID() + 'User', UserSchema);

Phew, thanks. I totally missed that.

My guess is the UUID was added to avoid collisions for multiple test runs after a broken run... or possibly between concurrent test runs. We shall see.

PR #3312, opened 2025-12-17
feat(instr-aws-sdk): support db semconv migration for dynamodb

JamieDanielson:
This adds support for using `OTEL_SEMCONV_STABILITY_OPT_IN` for controlled migration to stable `db.*` semconv. The `db.*` attributes are controlled by the `database[dup]` token.


trentm (approved):
Nice.

PR #3313, opened 2025-12-18
feat(instrumentation-mongodb): support `net.*` and database semconv migration

trentm:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38376,7 +38376,8 @@
       "version": "0.62.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@opentelemetry/instrumentation": "^0.208.0"
+        "@opentelemetry/instrumentation": "^0.208.0",
+        "@opentelemetry/semantic-conventions": "^1.33.0"
       },
       "devDependencies": {
         "@opentelemetry/api": "^1.3.0",


onurtemizkan (approved):

maryliag (approved):

packages/instrumentation-mongodb/src/instrumentation.ts
```diff
@@ -55,34 +69,60 @@ import {
...
   override _updateMetricInstruments() {
-    this._connectionsUsage = this.meter.createUpDownCounter(
-      METRIC_DB_CLIENT_CONNECTIONS_USAGE,
-      {
-        description:
-          'The number of connections that are currently in state described by the state attribute.',
-        unit: '{connection}',
-      }
-    );
+    if (this._dbSemconvStability & SemconvStability.OLD) {
```
JamieDanielson: I think we decided we always want this metric right now, regardless of semconv stability selected. That's what I suggested in this issue requesting guidance in semconv spec https://github.com/open-telemetry/semantic-conventions/issues/3200

> I am trying to understand the appropriate attributes/metrics to emit for experimental attributes/metrics that have changed over time. For example, we are currently emitting db.client.connections.usage, and the latest (still experimental) name is db.client.connection.count. This is my understanding of what attributes/metrics we should emit:
> 
> default: db.client.connections.usage
> database, database/dup: db.client.connections.usage

trentm: [Some DM Slackin' with Jamie.]

Thanks. I'll update (impl and the README), to not remove the `db.client.connections.usage` metric. Also I'll go back to instr-mysql and do the same.

trentm: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3313/commits/3eeb7c483d98166219d783ff906f811a198a31ff

Issue #3297, opened 2025-12-12
`unit-test` job can fail with "Unable to download artifact(s): Artifact not found for name: compile-cache-{PRNUM}"


trentm:
I have a `unit-test` job can failing with

> Unable to download artifact(s): Artifact not found for name: compile-cache-${PRNUM}

For example https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/20117049153/job/57946234366?pr=3288 is failing for me:

```
- unit-test (22)
  - Download Build Artifacts

Run actions/download-artifact@v5
  with:
    name: compile-cache-1928
    path: .nx
    merge-multiple: false
    repository: open-telemetry/opentelemetry-js-contrib
    run-id: 20117049153
...

Downloading single artifact
Error: Unable to download artifact(s): Artifact not found for name: compile-cache-1928
        Please ensure that your artifact is not expired and the artifact was uploaded using a compatible version of toolkit/upload-artifact.
        For more information, visit the GitHub Artifacts FAQ: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md
```

This is for PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3288
I had been getting a number of flaky test failures, so the workflow run above was the third run of "Re-run failed jobs".

My **guess** is that the `compile-cache-1928` artifact from the "compile" job of this workflow has now timed out / gone stale (the original run was 2d ago), so now it is impossible to get a step re-run to work.

This is not a *critical* bug. If my guess is correct, this is only something that will happen when re-running jobs on a workflow run from over 2d earlier.

My guess is that doing a full re-run of all jobs will workaround the issue.

/cc @david-luna

david-luna:
>My guess is that the compile-cache-1928 artifact from the "compile" job of this workflow has now timed out / gone stale (the original run was 2d ago), so now it is impossible to get a step re-run to work.

The cache is configured with `retention-days: 1` and it was created in Dec 10th. I think you're right. Compile step should be run again to have the cache entry in place.

```
Wed, 10 Dec 2025 23:57:36 GMT Run actions/upload-artifact@v4
Wed, 10 Dec 2025 23:57:36 GMT   with:
Wed, 10 Dec 2025 23:57:36 GMT     name: compile-cache-1928
Wed, 10 Dec 2025 23:57:36 GMT     path: .nx
Wed, 10 Dec 2025 23:57:36 GMT     include-hidden-files: true
[...]

PR #3318, opened 2025-12-18
chore: start and stop test serivces with npm scripts in workflows

david-luna:

The goal of this PR is make CI faster by only starting the services that are needed for testing. The current state is the workflow start all services (mysql, mongo, redis, ...) before testing anything. The problem is most of the times we do not need all of them to work. Also having the services defined in the workflow requires multiple maintenance of the configs (docker file and each workflow that needs services running)

- removed serivces section from the workflow files
- make use of `test-services:start` & `test-services:stop` script available in each package
- put a new script at the root `pacakge.json` the runs the scripts above for the changed packages

## Comparison

- Starting all services in CI takes around `1min and 10-30secs`. See: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/21135177260/job/60775875236?pr=3347
- For instrumentation scoped PRs (starting one service) the time drops to seconds (`10-14secs` for mongodb). See https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/20345473151/job/58456967241


david-luna:
this was auto assigned to the owners of the packages I've modified temporarily to test the workflow. I've removed them (and the labels) now that the changes in the packages are rolled back

.github/workflows/test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           path: .nx
           include-hidden-files: true
           if-no-files-found: error
-          retention-days: 1
+          retention-days: 7
```
david-luna, line 54: there is [an issue](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3297) which highlights this value is probably short. A flaky test may fail and then the PR author may want to run the failing job after a day. It will fail since the artifact has expired.

In the JS SIG we looked at the caches of this repo and we found we were already using all its capacity. But looking at [the docs](https://docs.github.com/en/actions/reference/limits#storage-limits-for-all-github-hosted-runners) it seems artifacts have a different limits.

So IMHO we can leet the artifact live for a week. After that is better to run the entire workflow again

.github/workflows/test.yml
```diff
@@ -66,129 +66,6 @@ jobs:
           - "20"
           - "22"
           - "24"
-    # `&foo` syntax is a YAML anchor. See:
```
david-luna, line 69: note for reviewer: now the source of truth is `/test/docker-compose.yaml`

.github/workflows/test-all-versions.yml
```diff
@@ -54,112 +54,6 @@ jobs:
       matrix:
         node: ["18", "20", "22", "24"]
     runs-on: ubuntu-latest
-    services:
```
david-luna, line 57: note for reviewer: now the source of truth is /test/docker-compose.yaml

PR #3322, opened 2025-12-19
fix(instrumentation-web-exception): properly add bounded error listener

BenoitZugmeyer:

The [README.md](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-web-exception#usage) advises to use the instrumentation like this:

```ts
registerInstrumentations({
  instrumentations: [
    new ExceptionInstrumentation({ ... }),
  ],
});
```

But this does not work because the function passed to `addEventListener('error', ...)` is the unbounded method, so accesses to `this.getConfig()` fails.

I know this sounds surprising, because the `onError` method was properly reaffected with the bounded method in the constructor. But actually, `enable()` is called by the base class, during `super()`, which is called *before the `onError` is bound*!

```ts
export class ExceptionInstrumentation extends InstrumentationBase<GlobalErrorsInstrumentationConfig> {
  constructor(config: GlobalErrorsInstrumentationConfig = {}) {
    super(PACKAGE_NAME, PACKAGE_VERSION, config); // calls this.enable()
    this.onError = this.onError.bind(this); // this is done too late!
  }
```

This wasn't caught in tests because in test, `enable()` is called two times, once when instantiating the instrumentation, and once explicitely when calling `instr.enable()`. This second call *do* register the correctly bounded method.

- To work around this, this PR make sure the `onError` method is proprely bound in `enable()`.
- Adjust the tests to be closer to the documented setup instructions.

### DIFF
--- a/packages/instrumentation-web-exception/src/instrumentation.ts
+++ b/packages/instrumentation-web-exception/src/instrumentation.ts
@@ -47,6 +47,10 @@ export interface GlobalErrorsInstrumentationConfig
 }
 
 export class ExceptionInstrumentation extends InstrumentationBase<GlobalErrorsInstrumentationConfig> {
+  declare private _onErrorHandler?: (
+    event: ErrorEvent | PromiseRejectionEvent
+  ) => void;
+
   constructor(config: GlobalErrorsInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
   }
@@ -95,13 +99,17 @@ export class ExceptionInstrumentation extends InstrumentationBase<GlobalErrorsIn
   }
 
   override disable(): void {
-    window.removeEventListener('error', this.onError);
-    window.removeEventListener('unhandledrejection', this.onError);
+    if (this._onErrorHandler) {
+      window.removeEventListener('error', this._onErrorHandler);
+      window.removeEventListener('unhandledrejection', this._onErrorHandler);
+    }
   }
 
   override enable(): void {
-    this.onError = this.onError.bind(this);
-    window.addEventListener('error', this.onError);
-    window.addEventListener('unhandledrejection', this.onError);
+    if (!this._onErrorHandler) {
+      this._onErrorHandler = this.onError.bind(this);
+    }
+    window.addEventListener('error', this._onErrorHandler);
+    window.addEventListener('unhandledrejection', this._onErrorHandler);
   }
 }


wolfgangcodes (approved):
Thanks @BenoitZugmeyer

packages/instrumentation-web-exception/src/instrumentation.ts
```diff
@@ -101,6 +100,7 @@ export class ExceptionInstrumentation extends InstrumentationBase<GlobalErrorsIn
   }
 
   override enable(): void {
+    this.onError = this.onError.bind(this);
```
david-luna: If we enable multiple times we may end up with several callbacks registered with no way to remove them. I think we should add a guard agains it.

packages/instrumentation-web-exception/test/instrumentation.test.ts
```diff
@@ -98,17 +98,15 @@ describe('ExceptionInstrumentation', () => {
   });
 
   describe('throwing an error', () => {
-    const instr = new ExceptionInstrumentation();
+    let disableInstrumentations: () => void;
     beforeEach(() => {
-      registerInstrumentations({
-        instrumentations: [instr],
+      disableInstrumentations = registerInstrumentations({
```
david-luna, line 103: TIL `registerInstrumentations` returns a tear down function. Neat!

david-luna (changes requested):
LGTM. Once we guard agains multiple calls to enable I think is good to go :)

packages/instrumentation-web-exception/src/instrumentation.ts
BenoitZugmeyer: I managed to do it but it was surprisingly challenging! Some context for anyone interested:

Using class properties in instrumentation classes is challenging because the `enable()` method is called in the base class constructor, before the class is fully initialized by the JS runtime.

See [this](https://www.typescriptlang.org/play/?useDefineForClassFields=true&target=11#code/MYGwhgzhAEBCkFNoG8BQ0PWAewHYQBcAnAV2AOyIAoBKFdTRggCwEsIA6BXMAIxAS0GGAL6ph0bnwG0UYseNCQYAMWzZJADwLcAJjHgQkaRiHY7cCIgH4AXNFkBeAHzQAbtla7xjHPmJkFNR0JoyYECQADlZCYZh+ENgCHCDYAOZUAEQABmaE3FbZ7mAgJEhgAGY6RNAR0cG2mQA00CzsKeYFRDQSCoxS-IIhEkxsnHkWVtCODnQuciPxeInJqRk5E11FbiVl0LoIFayWupk9jH2YuuzSQ-RxGKwVDgCEbeOdlt1Yy0kIKeksgB5dSRGC5T6FaDsaAkXAHI4nM69VAKVAJAjQCrqabQSwAd2gamwQmx2A41wgt1oQA) minimal example to illustrate the issue.

When instantiating the class, this happens chronologically:

1. the class constructor calls `super()`
2. the base class constructor  calls `this.enable()`
3. `enable()` defines the property
4. Once `super()` is finished, the JS runtime sets the property to `undefined`
5. So when running `disable()`, the property is `undefined` and cannot be reset.

This is very subtle and other instrumentations suffer from this issue. Unit tests are either not covering the `removeEventListener` case properly or the instrumentation is instantiated with `enable: false` so `enable()` is not called during the instantiation.

For example, the `instrumentation-browser-navigation` uses a [bunch of class properties to store event listeners/handlers](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/062fb29c3fb7c99c03096fe61dcf41743f17ca64/packages/instrumentation-browser-navigation/src/instrumentation.ts#L43-L45). Those are effectively set to `undefined` after `enable()` is called in the base class constructor and are not removed when calling `disable()`. Tests are not covering this (contrary to `instrumentation-web-exception` tests).

The trick is to use the TypeScript `declare` keyword, so the class property is not declared in JavaScript, and so the JavaScript runtime doesn't initialize it to `undefined`. I have seen this trick [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/5981dde6d77dbfe443f1ece963a37c3035b4690d/packages/instrumentation-user-interaction/src/instrumentation.ts#L57) and in a bunch of other instrumentations.

wolfgangcodes: This is really excellent context! I'm wondering for subtle things like this if there's a "common gotchas for web instrumentations" type FAQ we might want to put in a readme in the repo? 🤔

Issue #3310, opened 2025-12-17
[instr-browser-navigation] flaky test "Should redact sensitive query params: expected false to be truthy"


trentm:
I have seen this test fail a number of times in unrelated PRs.

E.g.: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/20290123086/job/58334216574?pr=3308 (in job `unit-test (22)`):

```
   > @opentelemetry/instrumentation-browser-navigation@0.1.0 test:browser
  > wtr --coverage
  
  
  Chrome: |██████████████████████████████| 0/2 test files | 0 passed, 0 failed
  
  Running tests...
  
  Running 2 test files...
  
  
  test/navigation.test.ts:
  
   🚧 Browser logs:
        {
          message: 'expected false to equal true',
          actual: false,
          expected: true,
          showDiff: true,
          operator: 'strictEqual',
          name: 'AssertionError',
          ok: false,
          stack: 'AssertionError: expected false to equal true\n' +
            '    at http://localhost:8000/test/navigation.test.ts?wtr-session-id=s04KgINfxj3MfuSeQSgUC:346:22'
        }
        {
          message: 'Should redact sensitive query params: expected false to be truthy',
          actual: false,
          showDiff: false,
          operator: 'strictEqual',
          name: 'AssertionError',
          ok: false,
          stack: 'AssertionError: Should redact sensitive query params: expected false to be truthy\n' +
            '    at http://localhost:8000/test/navigation.test.ts?wtr-session-id=s04KgINfxj3MfuSeQSgUC:478:18',
          expected: undefined
        }
  
   ❌ Browser Navigation Instrumentation > export navigation LogRecord > should sanitize URLs with credentials using default sanitizer
        Error: Uncaught expected false to equal true (http://localhost:8000/__wds-outside-root__/2/node_modules/chai/index.js:1515)
          at D.onerror (../../node_modules/@web/test-runner-mocha/dist/autorun.js:1:258618)
  
   ❌ Browser Navigation Instrumentation > export navigation LogRecord > should use custom sanitizeUrl callback when provided
        Error: Uncaught Should redact sensitive query params: expected false to be truthy (http://localhost:8000/__wds-outside-root__/2/node_modules/chai/index.js:1515)
          at D.onerror (../../node_modules/@web/test-runner-mocha/dist/autorun.js:1:258618)
  
  Chrome: |██████████████████████████████| 2/2 test files | 35 passed, 2 failed
[...]

PR #3323, opened 2025-12-19
fix(instrumentation-browser-navigation): improve test stability with …

Abinet18:
…retry logic and increased timeouts for CI environments

## Which problem is this PR solving?

-

## Short description of the changes

-

PR #3326, opened 2025-12-20
fix(instrumentation-browser-navigation): use 'declare' to avoid JS class field initialization surprise

trentm:
Before this change these class fields were being set to undefined after the
super() call, which blows away values set in the .enable() called by the
super constructor.

This also reverts the following change that was attempting to work around test failures:
   "fix(instrumentation-browser-navigation): improve test stability with … (#3323)"

This also adds a guard on .enable() being called multiple times because it
isn't idempotent.

Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3310

---

The best description of the `declare` class fields issue is here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3310#issuecomment-3677040183

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,129 +66,129 @@ jobs:
           - "20"
           - "22"
           - "24"
-    # # `&foo` syntax is a YAML anchor. See:
-    # # https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases
-    # services: &test_services
-    #   memcached:
-    #     image: memcached:1.6.40-alpine
-    #     ports:
-    #       - 11211:11211
-    #   mongo:
-    #     image: mongo
-    #     ports:
-    #       - 27017:27017
-    #   mssql:
-    #     image: mcr.microsoft.com/mssql/server:2022-latest
-    #     env:
-    #       MSSQL_SA_PASSWORD: mssql_passw0rd
-    #       ACCEPT_EULA: Y
-    #     ports:
-    #       - 1433:1433
-    #     options: >-
-    #       --health-cmd "/opt/mssql-tools18/bin/sqlcmd -S localhost -U sa -P $MSSQL_SA_PASSWORD -C -Q 'select 1' -b -o /dev/null"
-    #       --health-interval 1s
-    #       --health-timeout 30s
-    #       --health-start-period 10s
-    #       --health-retries 20
-    #   mysql:
-    #     image: mysql:5.7
-    #     env:
-    #       MYSQL_USER: otel
-    #       MYSQL_PASSWORD: secret
-    #       MYSQL_DATABASE: otel_mysql_database
-    #       MYSQL_ROOT_PASSWORD: rootpw
-    #     ports:
-    #       - 3306:3306
-    #     options: >-
-    #       --health-cmd="mysqladmin ping"
-    #       --health-interval 10s
-    #       --health-timeout 5s
-    #       --health-retries 5
-    #   oracledb:
-    #     image: gvenzl/oracle-free:slim
-    #     env:
-    #       APP_USER: otel
-    #       APP_USER_PASSWORD: secret
-    #       ORACLE_PASSWORD: oracle
-    #     ports:
-    #       - 1521:1521
-    #     options: >-
-    #       --health-cmd "sqlplus system/oracle@//localhost/FREEPDB1"
-    #       --health-interval 10s
-    #       --health-timeout 5s
-    #       --health-retries 5
-    #   postgres:
-    #     image: postgres:16-alpine
-    #     env:
-    #       POSTGRES_USER: postgres
-    #       POSTGRES_DB: otel_pg_database
-    #       POSTGRES_PASSWORD: postgres
-    #     ports:
-    #       - 5432:5432
-    #     options: >-
-    #       --health-cmd pg_isready
-    #       --health-interval 10s
-    #       --health-timeout 5s
-    #       --health-retries 5
-    #   redis:
-    #     image: redis
-    #     ports:
-    #       - 6379:6379
-    #     options: >-
-    #       --health-cmd "redis-cli ping"
-    #       --health-interval 10s
-    #       --health-timeout 5s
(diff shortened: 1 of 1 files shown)


trentm:
The initial state (the first two commits) should fix the testing issue. However, I propose to also drop the explicit `instrumentation.enable()` calls in some of the tests that is resulting in double-calling `.enable()`.

trentm:
Back in draft state, there is more going on here. 

- BrowserNavigationInstrumentation does not safely handle being `.enable()`d twice. For example `_patchHistoryApi()` is doing its wrapping twice, so we are getting duplicate/multiple `_onSoftNavigation` events.
-  The `setTimeout(..., 0)` in the isntrumentation for timing for *the tests* smells to me. In fairness I haven't fully grokked this yet.
- `_waitForPageLoad()` is called for each `.enable()` meaning we get duplicate/multiple `_onHardNavigation` events.
- The tests should assert on the exact number of expected navigation events. Or at least I hope that is possible. I'm a newb with browser dev so, I don't have experience on caveats in browser events. E.g. I've been playing with this test case:

```ts
it.only('should export LogRecord with type traverse when history.back() triggers a popstate', done => {
```

packages/instrumentation-browser-navigation/src/instrumentation.ts
```diff
@@ -133,8 +138,7 @@ export class BrowserNavigationInstrumentation extends InstrumentationBase<Browse
     // Check if document has already loaded completely
     if (document.readyState === 'complete' && !this._hasProcessedInitialLoad) {
       this._hasProcessedInitialLoad = true;
-      // Use setTimeout to allow tests to reset exporter before this fires
-      setTimeout(() => this._onHardNavigation(), 0);
+      this._onHardNavigation();
```
trentm, line 141: Reviewer note: I don't think this was necessary. Tests are passing without it. It *does* mean that one can a navigation event synchronously on enabling the instrumentation... instead of in a later tick. If there is a usage issue with that, we could restore this.

packages/instrumentation-browser-navigation/src/instrumentation.ts
```diff
@@ -155,6 +159,11 @@ export class BrowserNavigationInstrumentation extends InstrumentationBase<Browse
    * implements enable function
    */
   override enable() {
+    if (this._isEnabled) {
+      return;
+    }
+    this._isEnabled = true;
```
trentm, line 165: Reviewer note: This seemed the simplest way to avoid double-enabling. Some of the `enable()` code is not idempotent.

packages/instrumentation-browser-navigation/src/instrumentation.ts
```diff
@@ -243,8 +257,10 @@ export class BrowserNavigationInstrumentation extends InstrumentationBase<Browse
   }
 
   private _patchHistoryApi(): void {
-    // unpatching here disables other instrumentation that use the same api to wrap history, commenting it out
-    // this._unpatchHistoryApi();
+    // Unpatching here disables other instrumentation that use the same api to
+    // wrap history, commenting it out.
+    //    this._unpatchHistoryApi();
+    // TODO: For example, instrumentation-user-interaction also patches `history`. Do the two instrumentations work at the same time?
```
trentm, line 263: I think this is worthwhile testing. With double-wrapping (by two separate instrumentations), their attempts to *unwrap* will be surprising if they do it in a separate order.

@Abinet18 was `instrumentation-user-interaction` a particular "other instrumentation" you meant? Are there others you know of?

packages/instrumentation-browser-navigation/test/navigation.test.ts
```diff
@@ -194,10 +193,10 @@ describe('Browser Navigation Instrumentation', () => {
 
       history.replaceState({}, '', '/dummy2.html');
 
-      assert.strictEqual(exporter.getFinishedLogRecords().length, 1);
+      const recs = exporter.getFinishedLogRecords();
+      assert.strictEqual(recs.length, 2);
```
trentm, line 197: Reviewer note: This and other assertion changes in this test file that expect *2* now are due to the `setTimeout` removal in instrumentation.ts. The expectation is now: 1. the hard nav event, 2. the soft nav event.

packages/instrumentation-browser-navigation/test/navigation.test.ts
trentm, line 1: Reviewer note: This file is a mix of reverting the earlier change adding retries and changing timeouts, plus my changes that drop unneeded `instrumentation.enable()` calls, and updating some assertions for changes due to the `setTimeout` removal in instrumentation.ts

packages/instrumentation-browser-navigation/src/instrumentation.ts
martinkuba: The user-interaction instrumentation is the only other one. I wonder if it would make sense to move patching these common APIs to a common package and have instrumentations subscribe to an internal event instead.

martinkuba:
@trentm This is great investigation! I have been looking at this as well and would have fixed just the test. Well done on finding the issue with the private fields and events not getting unsubscribed. All of this makes sense to me, and I will approve once the github action is reverted.

packages/instrumentation-browser-navigation/src/instrumentation.ts
Abinet18: It is user interaction for now but I assume there could be a new instrumentation like soft navigation which can do the same.

Abinet18:
@trentm thanks for looking deep into this. I didn't know about the usage of declare in the fields. I was facing test issues only in CI while they pass in local  and focused on fixing that.

martinkuba:
fwiw, I have been running the tests continuously over the last couple of days with these changes, and I have not seen it fail again.
https://github.com/martinkuba/opentelemetry-js-contrib/actions/workflows/navigation-flake.yml

trentm:
> All of this makes sense to me, and I will approve once the github action is reverted.

I just pushed that revert of test.yml now (commit `d8703d93..7540de86`). I'd had that sitting in my working copy over the xmas break and I guess had forgotten to push. Sorry about the delay.

martinkuba (approved):

PR #3337, opened 2026-01-12
fix(mongoose): patch updateOne/deleteOne on model level

### LINKED ISSUE #3336: [instrumentation-mongoose] TAV tests failing for `mongoose@8.21.0`
TAV tests are failing for `mongoose@8.21.0`, the error indicates that telemetry (statement info) is missing. I think it may be related to the fact that `mongoose` is now calling different methods internally (this only seems to happen when operating directly on the `Model`). 

Refs: 
- Failing tests (search `-- fatal` in raw logs): https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/20827438153/job/59833026791?pr=3332 
- `8.21.0` release: https://github.com/Automattic/mongoose/releases/tag/8.21.0
- PR I suspect caused this: https://github.com/Automattic/mongoose/pull/15908/changes (I don't really understand the mechanism why)

pichlermarc:

Fixes #3336. 
Unblocks #3332

Mongoose tests are failing due to a change in `mongoose@8.21.0` where we cannot rely on `Query.exec()` generating correct spans for `Model.updateOne()`/`Model.deleteOne()` anymore, since the `Query` instance is modified while `exec()` is running through the `pre` hook.

Unfortunately, after `Query.exec()` is finished, the options are left modified by the library, so just delaying
attaching the attributes after the span is done is not an option. Therefore, I'm proposing we patch Model methods and grab the data directly where the user provides it.


pichlermarc:
I think the `nyc`'s `--no-clean` was missing, so coverage checks failed. Let's try again with that. https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3337/commits/bc533aab7ee762da1337185976e29393870c3fca

packages/instrumentation-mongoose/WHY_MODEL_PATCH_NEEDED.md
david-luna, line 1: do you plan to add som content on this file?

pichlermarc: Oops. I have been playing around with coding agents on topics that I have insight already to see what works and what does not. I this case I was tying to see if it could come up with a good explanation for my PR before I tell it exactly what my patch is for (the idea being that we could use it during review to figure out "why" on PRs that have little context)

It's output was lackluster as I could not get the agent to read code in `node_modules` and match it to GitHub lines (was trying to get it to replicate my usual workflow). The file is a remnant of that.

Removed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3337/changes/44686d40d3bf2fc1f3166ddb24ad94425a12126b

david-luna (approved):

PR #3356, opened 2026-01-24
fix(instrumentation-pino): Allow control over logged fields

jnloureiro:

In some environments, automatic log processing expects certain field names to be present (or absent) in the log record or for the values to have specific complex formats.

The `logKeys` option now accepts partial custom keys to use the default values, and explicitly undefined values to remove the keys from the correlation record.

To prevent any breaking changes from being introduced, a new `passLogCorrelationToMixin` option has been added, which allows total control over the logged values through the mixin function. This option defaults to false.

### DIFF
--- a/packages/instrumentation-pino/src/instrumentation.ts
+++ b/packages/instrumentation-pino/src/instrumentation.ts
@@ -72,10 +72,12 @@ export class PinoInstrumentation extends InstrumentationBase<PinoInstrumentation
               level: number,
               ...rest: unknown[]
             ) => {
-              return Object.assign(
-                otelMixin(ctx, level),
-                origMixin(ctx, level, ...rest)
-              );
+              const otelMixinResult = otelMixin(ctx, level);
+              if (config.passLogCorrelationToMixin) {
+                return origMixin(otelMixinResult, level, ...rest);
+              } else {
+                return Object.assign(otelMixinResult, origMixin(ctx, level, ...rest));
+              }
             };
           }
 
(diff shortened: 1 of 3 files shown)


packages/instrumentation-pino/src/types.ts
```diff
@@ -50,8 +57,8 @@ export interface PinoInstrumentationConfig extends InstrumentationConfig {
    * available.
    */
   logKeys?: {
-    traceId: string;
-    spanId: string;
-    traceFlags: string;
+    traceId?: string;
+    spanId?: string;
+    traceFlags?: string;
```
dyladan: I think using `string | null` would be much more clear about the intent here

packages/instrumentation-pino/src/instrumentation.ts
```diff
@@ -72,10 +72,12 @@ export class PinoInstrumentation extends InstrumentationBase<PinoInstrumentation
...
-              return Object.assign(
-                otelMixin(ctx, level),
-                origMixin(ctx, level, ...rest)
-              );
+              const otelMixinResult = otelMixin(ctx, level);
+              if (config.passLogCorrelationToMixin) {
+                return origMixin(otelMixinResult, level, ...rest);
+              } else {
+                return Object.assign(otelMixinResult, origMixin(ctx, level, ...rest));
+              }
```
dyladan, line 80: Is this feature related? this seems like it might be 2 separate features which may justify 2 separate PRs

packages/instrumentation-pino/src/types.ts
jnloureiro:
```suggestion
    traceId: string | null;
    spanId: string | null;
    traceFlags: string | null;
```

I agree. Using `string | null`, all three fields are required in the configuration object just as they were before, but we can explicitly exclude any of them.

packages/instrumentation-pino/src/instrumentation.ts
jnloureiro: I'd argue it's part of the same feature. This change allows full control over the mixin result, while the change in the configuration object serves as a succinct version of this behavior in case all we need is to ignore a field.

PR #3360, opened 2026-01-27
fix(pino): avoid mutating user options when reused

### LINKED ISSUE #2024: pino instrumentation might modify user object
### What did you do?
reviewed code of pino instrumentation

Expected the instrumentation to have no visible side effects on the objects passed as parameters

### What did you see instead?
There might be visible changes to the passed options object that if called with an additional constructor might result in multiple wrappers of the mixin function, in extreme situations very deep call stacks and possible crashes

If we consider following example:
`const options = {};`
`const firstLogger = pino(options);` // this will alter the options object => options.mixin = [see here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts#L166)
`const secondLogger = pino(options);` // this will again alter it but at a different location => options.mixin = [see here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-pino/src/instrumentation.ts#L173)
`const thirdLogger = pino(options);` // this will again alter the object at the previous location

`thirdLogger.error("some Error");` // this will have mixin => mixin => mixin on the call stack before it's done

MUSKANNISHAD:
This change avoids mutating user-provided options when creating multiple Pino
logger instances with the same configuration object.

Previously, reusing the same options object across multiple `pino(options)`
calls could lead to mixin wrappers being stacked, since the options reference
was shared. This patch defensively clones the options argument before passing
it to `pino()`, preventing unintended side effects.

A regression test has been added to ensure that reusing the same options object
does not cause errors or recursive mixin behavior.

Fixes #2024 

Note: On Windows, `nx test instrumentation-pino` fails locally due to Mocha glob
quoting (`'test/**/*.test.ts'`). Tests pass in CI (Linux), which is the supported
environment for this repository.

### DIFF
--- a/packages/instrumentation-pino/src/instrumentation.ts
+++ b/packages/instrumentation-pino/src/instrumentation.ts
@@ -55,7 +55,12 @@ export class PinoInstrumentation extends InstrumentationBase<PinoInstrumentation
           const config = instrumentation.getConfig();
           const isEnabled = instrumentation.isEnabled();
 
-          const logger = moduleExports(...args);
+         const patchedArgs = [...args];
+
+          if (patchedArgs.length > 0 && typeof patchedArgs[0] === 'object') {
+           patchedArgs[0] = { ...(patchedArgs[0] as object) };
+           }
+           const logger = moduleExports(...patchedArgs);
 
           // Setup "log correlation" -- injection of `trace_id` et al fields.
           // Note: If the Pino logger is configured with `nestedKey`, then
--- a/packages/instrumentation-pino/src/instrumentation.ts
+++ b/packages/instrumentation-pino/src/instrumentation.ts
@@ -55,12 +55,12 @@ export class PinoInstrumentation extends InstrumentationBase<PinoInstrumentation
           const config = instrumentation.getConfig();
           const isEnabled = instrumentation.isEnabled();
 
-         const patchedArgs = [...args];
+          const patchedArgs = [...args];
 
           if (patchedArgs.length > 0 && typeof patchedArgs[0] === 'object') {
-           patchedArgs[0] = { ...(patchedArgs[0] as object) };
-           }
-           const logger = moduleExports(...patchedArgs);
+            patchedArgs[0] = { ...(patchedArgs[0] as object) };
+          }
+          const logger = moduleExports(...patchedArgs);
 
           // Setup "log correlation" -- injection of `trace_id` et al fields.
           // Note: If the Pino logger is configured with `nestedKey`, then
--- a/packages/instrumentation-pino/test/pino.test.ts
+++ b/packages/instrumentation-pino/test/pino.test.ts
@@ -321,16 +321,16 @@ describe('PinoInstrumentation', () => {
     });
 
     it('does not stack mixins when reusing the same options object', () => {
-  const options: any = {};
+      const options: any = {};
 
-  const logger1 = pino(options);
-  const logger2 = pino(options);
-  const logger3 = pino(options);
+      pino(options);
+      pino(options);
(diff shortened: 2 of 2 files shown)


packages/instrumentation-pino/test/pino.test.ts
```diff
@@ -320,6 +320,18 @@ describe('PinoInstrumentation', () => {
...
+      const options: any = {};
+
+      pino(options);
+      pino(options);
+      const logger3 = pino(options);
+
+      assert.doesNotThrow(() => {
+        logger3.info('test message');
+      });
+    });
```
dyladan, line 333: The assertion here doesn't seem to match the description. The assertion is just that it doesn't throw, but the description talks about "stacking mixins."

packages/instrumentation-pino/src/instrumentation.ts
```diff
@@ -55,7 +55,12 @@ export class PinoInstrumentation extends InstrumentationBase<PinoInstrumentation
           const config = instrumentation.getConfig();
           const isEnabled = instrumentation.isEnabled();
 
-          const logger = moduleExports(...args);
+          const patchedArgs = [...args];
+
+          if (patchedArgs.length > 0 && typeof patchedArgs[0] === 'object') {
+            patchedArgs[0] = { ...(patchedArgs[0] as object) };
+          }
+          const logger = moduleExports(...patchedArgs);
```
dyladan, line 63: It seems like the mutation is happening somewhere else, and this is just passing a copy to then be mutated. Would it not be better to prevent the mutation in the first place? Is the mutation happening in our code or pino itself?

trentm:
I tried to repro the original issue, and could not with the *current* instrumentation-pino:

```
% npm ls
asdf.20260409t113544@1.0.0 /Users/trentm/tmp/asdf.20260409T113544
├── @opentelemetry/auto-instrumentations-node@0.72.0
└── pino@10.3.1

% npm ls @opentelemetry/instrumentation-pino
asdf.20260409t113544@1.0.0 /Users/trentm/tmp/asdf.20260409T113544
└─┬ @opentelemetry/auto-instrumentations-node@0.72.0
  └── @opentelemetry/instrumentation-pino@0.60.0
```

My test script **use-pino.js**:

```js
const pino = require('pino');

const opts = {}
const log = pino(opts);
const log2 = pino(opts);
const log3 = pino(opts);
console.log('opts (after 3): ', opts);

log3.info({foo: 'bar'}, 'hi');
```

Running that without and with instrumentation:

```
% node use-pino.js
opts (after 3):  {}
{"level":30,"time":1776291178321,"pid":74919,"hostname":"peach.local","foo":"bar","msg":"hi"}

% node --require '@opentelemetry/auto-instrumentations-node/register' use-pino.js
opts (after 3):  {}
{"level":30,"time":1776291186819,"pid":74921,"hostname":"peach.local","foo":"bar","msg":"hi"}
```

## This is already fixed

I believe this was fixed way back in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2249 (which was merged a few weeks after the initial bug report: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2024).

Specifically, the instrumentation-pino code before #2249 *used* to modify the given `options` argument:
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2249/changes#diff-d9d63b1d85594618c93de58adfdf835349464699239c4c8b3e5b54e24dda41c8L65-L67

Now it no longer does that.
Please re-open this if I've missed something.

PR #3361, opened 2026-01-27
feat(instrumentation-runtime-node): add active resource gauge

sfozz:

Add a gauge to monitor the Active Resources that are keeping the event loop alive. This metric is recommended to be monitored to avoid saturation and starvation issues.

- allows detection of resource leaks (sockets and file descriptors for example)

- Add gauge metric `v8js.resource.active`
- Add metric attribute `v8js.resource.type` 

## Reference

- https://nodejs.org/api/process.html#processgetactiveresourcesinfo

### DIFF
--- a/packages/instrumentation-runtime-node/src/instrumentation.ts
+++ b/packages/instrumentation-runtime-node/src/instrumentation.ts
@@ -11,6 +11,7 @@ import { EventLoopDelayCollector } from './metrics/eventLoopDelayCollector';
 import { GCCollector } from './metrics/gcCollector';
 import { HeapSpacesSizeAndUsedCollector } from './metrics/heapSpacesSizeAndUsedCollector';
 import { EventLoopTimeCollector } from './metrics/eventLoopTimeCollector';
+import { ActiveResourcesCollector } from './metrics/activeResourcesCollector';
 /** @knipignore */
 import { PACKAGE_VERSION, PACKAGE_NAME } from './version';
 
@@ -33,6 +34,7 @@ export class RuntimeNodeInstrumentation extends InstrumentationBase<RuntimeNodeI
       new EventLoopDelayCollector(this._config),
       new GCCollector(this._config),
       new HeapSpacesSizeAndUsedCollector(this._config),
+      new ActiveResourcesCollector(this._config),
     ];
     if (this._config.enabled) {
       for (const collector of this._collectors) {
--- a/packages/instrumentation-runtime-node/src/metrics/activeResourcesCollector.ts
+++ b/packages/instrumentation-runtime-node/src/metrics/activeResourcesCollector.ts
@@ -1,0 +1,77 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


trentm:
@pichlermarc @maryliag Thoughts on whether the *requirement* here should be that we update semconv first?  IIUC, the original definition of v8 semconv metrics was perhaps whatever this instrumentation already just happened to be emitting? Do you recall, Marylia?

maryliag:
> Thoughts on whether the requirement here should be that we update semconv first? 

I think the best approach is indeed updating the semconv first. @sfozz can you make the changes [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/v8js-metrics.md#semantic-conventions-for-v8-js-engine-runtime-metrics) first?

> IIUC, the original definition of v8 semconv metrics was perhaps whatever this instrumentation already just happened to be emitting? Do you recall, Marylia?

First we did the conventions, then the instrumentation was added, was a little back and forward, but the implementation was based on the sem conv, and at that time we chose the most requested metrics.


Also, @sfozz please add tests to this PR

sfozz:
@maryliag https://github.com/open-telemetry/semantic-conventions/pull/3445 for the update to semconv.

dyladan:
Closed and reopened to trigger CI. Should this be held until the linked semconv PR merges?

sfozz:
> Closed and reopened to trigger CI. Should this be held until the linked semconv PR merges?

Yes please... :crossed_fingers:

Hronom:
As far as I understand this blocks this changes
https://github.com/open-telemetry/semantic-conventions/pull/3481
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3405

Very important changes to maintain properly infra with node based microservices.

sfozz:
@Hronom 
> As far as I understand this blocks this changes [open-telemetry/semantic-conventions#3481](https://github.com/open-telemetry/semantic-conventions/pull/3481) #3405

They're independent of each other.

open-telemetry/opentelemetry-js-contrib/pull/3361 is pending the merge of open-telemetry/semantic-conventions/pull/3445


It looks like open-telemetry/opentelemetry-js-contrib/pull/3405 is pending the merge of open-telemetry/semantic-conventions/pull/3481

Hronom:
@sfozz oh sorry, I just wanna things get done, for us it's important metric e.g. and looks like a bug...
I got comment here https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3405#issuecomment-4113620005 that soemthing needed to be accepted in this pr.

I don't have time to get deep in the processes of this project, but it seems like very slow.
I just wanna merge that change.

sfozz:
@dyladan ready for re-review :bow:

dyladan (commented):
This looks good except for one small caveat. Values of `0` will never be reported because they're not returned by `getActiveResourcesInfo`. This means if you have some resource type like `Immediate` keeping the event loop active, and that drops to `0`, the drop will never be reported. The last known non-zero value will be the last value in your monitoring tool until another non-zero value comes in. I think we should emit `0` explicitly for any of the known resource types even if they are not returned by `getActiveResourcesInfo`.

sfozz:
Have updated as suggested, is the tcpserver test going to cause problems in CI?

danyalahmed1995:
I reran the focused active-resource test locally after installing deps, and the zero-emission behavior looks correct.

One small test-hardening note: the TCP server test currently uses fixed port `4000`, fixed sleeps, and does not await `listen` / `close`. It passed locally for me, so I wouldn’t call it a blocker, but switching to port `0` and awaiting the server lifecycle could make it less timing-sensitive if CI flakes.

sfozz:
@dyladan is there anything left on this?

sfozz:
@dyladan @d4nyll is there anything left on this?

pichlermarc (approved):

PR #3367, opened 2026-01-29
fix(instrumentation-pg): correct `db.client.connection.count` metric counting to handle multiple pg pools

### LINKED ISSUE #3354: @opentelemetry/instrumentation-pg: multiple pools cause db.client.connection.count to be recorded incorrectly

```
"@opentelemetry/api": "^1.9.0"
"@opentelemetry/auto-instrumentations-node": "^0.67.0"
"@opentelemetry/core": "^2.2.0"
"@opentelemetry/sdk-metrics": "^2.2.0"
"@opentelemetry/sdk-node": "^0.208.0"
"@opentelemetry/sdk-trace-node": "^2.2.0"
```

22.17.1

### What did you do?

We are using instrumentation-pg via auto instrumentation to monitor the connections to a db. One of our services has 2 connection pools (different dbs soon to be different schema in the same db) but is looks like the counters are not working as expected. One of the pools connection counters just seems to increment where as the other just decrements.

```
db.client.connection.count{pool=poolA,state=used} 47
db.client.connection.count{pool=poolB,state=used} -47 
```

The pool only has a maximum size of 10, so it should never get beyond that.
Other services with only one pool are not impacted and are reporting fine

We are using FoalTS v5

#### OTEL Config

```
OTEL_SERVICE_NAME=our-api
OTEL_TRACES_SAMPLER=parentbased_always_off
OTEL_EXPORTER_OTLP_ENDPOINT=http://opentelemetry-kube-stack-collector:4318
OTEL_TRACES_EXPORTER=otlp
OTEL_METRICS_EXPORTER=otlp
OTEL_SEMCONV_STABILITY_OPT_IN=http
OTEL_NODE_RESOURCE_DETECTORS=env,host,os
OTEL_LOGS_EXPORTER=none
OTEL_RESOURCE_ATTRIBUTES=deployment.environment=prod,service.version=production
```

for the default pool size I would expect the used count to go up to at most 10 and reduce as used connections are returned the the pool

### What did you see instead?

That one pool always increments and the other pool always decrements

Image

### Additional context

Will supply example code

trentm:
Before this change the instrumentation was assuming there was only ever
one active PG pool.

Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3354

# Checklist

- [ ] Add tests
- [ ] Discuss whether getPoolName and `db.namespace` could include the PG schema, if any. Though really this would be a separate PR.

### DIFF
--- a/packages/instrumentation-pg/src/utils.ts
+++ b/packages/instrumentation-pg/src/utils.ts
@@ -381,7 +381,7 @@ export function updateCounter(
   pool: PgPoolExtended,
   connectionCount: UpDownCounter,
   connectionPendingRequests: UpDownCounter,
-  connCountFromPoolName: {[name: string]: poolConnectionsCounter},
+  connCountFromPoolName: { [name: string]: poolConnectionsCounter }
 ) {
   const all = pool.totalCount;
   const pending = pool.waitingCount;
@@ -390,7 +390,7 @@ export function updateCounter(
 
   let connCount = connCountFromPoolName[poolName];
   if (!connCount) {
-    connCount = {used: 0, idle: 0, pending: 0};
+    connCount = { used: 0, idle: 0, pending: 0 };
     connCountFromPoolName[poolName] = connCount;
   }
 
--- a/packages/instrumentation-pg/src/utils.ts
+++ b/packages/instrumentation-pg/src/utils.ts
@@ -381,28 +381,36 @@ export function updateCounter(
   pool: PgPoolExtended,
   connectionCount: UpDownCounter,
   connectionPendingRequests: UpDownCounter,
-  latestCounter: poolConnectionsCounter
-): poolConnectionsCounter {
+  connCountFromPoolName: {[name: string]: poolConnectionsCounter},
+) {
   const all = pool.totalCount;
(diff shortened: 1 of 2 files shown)


packages/instrumentation-pg/src/utils.ts
```diff
@@ -381,28 +381,36 @@ export function updateCounter(
...
   const all = pool.totalCount;
   const pending = pool.waitingCount;
   const idle = pool.idleCount;
   const used = all - idle;
 
-  connectionCount.add(used - latestCounter.used, {
+  let connCount = connCountFromPoolName[poolName];
+  if (!connCount) {
+    connCount = {used: 0, idle: 0, pending: 0};
+    connCountFromPoolName[poolName] = connCount;
```
trentm, line 394: A concern here is that `connCountFromPoolName` will grow for every new PG `poolName`. Entries are never removed from this mapping.

In normal usage I would think an app would only ever have one or a few PG pools.  If this is a concern, we could use something like an LRU cache  (e.g. https://github.com/isaacs/node-lru-cache, or https://github.com/isaacs/ttlcache).

PR #3377, opened 2026-02-09
feat(resource-detector-aws): read cloud.account.id from symlink in Lambda detector

RaphaelManke:
## Summary

- Reads `cloud.account.id` from the symlink at `/tmp/.otel-aws-account-id` created by the OTel Lambda Extension
- Uses `fs.readlinkSync()` to read the symlink target (the AWS account ID)
- Logs at debug level if the symlink doesn't exist — no impact on other resource detection

## Depends on

- open-telemetry/opentelemetry-lambda#2127 (extension creates the symlink)

## Changes

- `AwsLambdaDetector.ts` — Added readlink logic with `ACCOUNT_ID_SYMLINK_PATH` constant
- Tests: happy path, leading zeros preserved, missing symlink handled gracefully

## Test plan

- [x] Unit test: symlink exists → `cloud.account.id` set correctly
- [x] Unit test: leading zeros preserved as string
- [x] Unit test: missing symlink → attribute absent, no error

### DIFF
--- a/packages/resource-detector-aws/src/detectors/AwsLambdaDetector.ts
+++ b/packages/resource-detector-aws/src/detectors/AwsLambdaDetector.ts
@@ -15,6 +15,7 @@
  */
 
 import * as fs from 'fs';
+import { diag } from '@opentelemetry/api';
 import {
   ResourceDetector,
   DetectedResource,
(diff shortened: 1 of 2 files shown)


dyladan:
Is there any sort of spec or documentation for this feature? Can we be certain this will always continue to work into the future?

RaphaelManke:
> Is there any sort of spec or documentation for this feature? Can we be certain this will always continue to work into the future?

I will update the lambda extension docs to state this feature. 

The capability that is used is documented on aws in the lambda extension init. https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-extensions-registration-api

image

PR #2726, opened 2025-02-20
chore: release main

dyladan:
:robot: I have created a release *beep* *boop*
---

<details><summary>auto-configuration-propagators: 0.4.0</summary>

### ⚠ BREAKING CHANGES

* update to JS SDK 2.x ([#2738](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2738))

### Miscellaneous Chores

* update to JS SDK 2.x ([#2738](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2738)) ([7fb4ba3](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/7fb4ba3bc36dc616bd86375cfd225722b850d0d5))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/propagator-aws-xray bumped from ^1.26.2 to ^2.0.0
    * @opentelemetry/propagator-aws-xray-lambda bumped from ^0.53.2 to ^0.54.0
    * @opentelemetry/propagator-ot-trace bumped from ^0.27.3 to ^0.28.0
</details>

<details><summary>auto-instrumentations-node: 0.57.0</summary>

### ⚠ BREAKING CHANGES

* update to JS SDK 2.x ([#2738](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2738))

### Bug Fixes

* **deps:** update otel core experimental to ^0.57.2 ([#2716](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2716)) ([d2a9a20](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/d2a9a20f1cd8c46c842e18490a4eba36fd71c2da))

### Miscellaneous Chores

* update to JS SDK 2.x ([#2738](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2738)) ([7fb4ba3](https://github.com/open-telemetry/opentelemetry-js-contrib/commit/7fb4ba3bc36dc616bd86375cfd225722b850d0d5))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @opentelemetry/instrumentation-amqplib bumped from ^0.46.1 to ^0.47.0
    * @opentelemetry/instrumentation-aws-lambda bumped from ^0.50.3 to ^0.51.0
    * @opentelemetry/instrumentation-aws-sdk bumped from ^0.49.1 to ^0.50.0
[...]

PR #3378, opened 2026-02-10
fix(instrumentation-undici)!: fix header capture to match spec

trentm:
- When headers are captured (to `http.request.header.<name>` or
  `http.response.header.<name` span attributes) according to the
  `headersToSpanAttributes` option, the value is now an *array of
  string*, rather than a string.
  https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/
- The `http.response.header.content-length` is no longer specially
  handled. Before this change it would (a) always be captured if
  the response included a 'Content-Length' header and (b) would be
  converted to a number. Now, one must use `headersToSpanAttributes`
  to capture this header and it is not converted to a number.

(The spec includes a separate, unstable, `http.response.body.size` that
could be added in a separate PR.)

### DIFF
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -384,7 +384,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
 
       for (const [name, value] of headersMap.entries()) {
         if (headersToAttribs.has(name)) {
-          const attrValue = Array.isArray(value) ? value.join(', ') : value;
+          const attrValue = Array.isArray(value) ? value : [value];
           spanAttributes[`http.request.header.${name}`] = attrValue;
         }
       }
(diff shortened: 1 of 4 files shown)


trentm:
I should add a changelog specific entry that documents the effective change for users. I have to look up the mechanism to do that again.

trentm:
> I should add a changelog specific entry that documents the effective change for users. I have to look up the mechanism to do that again.

There is https://github.com/googleapis/release-please?tab=readme-ov-file#how-can-i-fix-release-notes but that isn't quite what is wanted here. I'm not sure if there is a way to get release-please to include some more prose for the generated changelog beyond the commit title. Wanh wanh.

The only example in the past I know of had Marc manually updating changelog files in a commit on the "chore: release main" release-please PR: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2726/commits (see last commit there)

PR #3379, opened 2026-02-10
feat(instrumentation-runtime-node): emit logs for uncaught exceptions

### LINKED ISSUE #3376: Capture uncaught exceptions and emit a log message
### Is your feature request related to a problem? Please describe.

Backend services frequently fail in ways that never become telemetry: uncaught exceptions and unhandled promise rejections. Unless users manually register process-level handlers or add a third-party error tracker, these failures produce no OpenTelemetry logs, which means:

- crashes and fatal errors are missing from observability backends
- incident timelines are incomplete (no “last error” record)
- teams lose the stacktrace context needed to triage production issues quickly
- error tracking experiences are inconsistent across services and languages.

### Describe the solution you'd like

- When an application throws an uncaught exception or has an unhandled rejection, the SDK should emit a `LogRecord` that captures the exception details.
- The log should include the exception semantic convention attributes (`exception.type`, `exception.message`, `exception.stacktrace`) and a meaningful severity (e.g., FATAL for uncaught exceptions, ERROR for unhandled rejections). This can use the new `Exceptions` API https://github.com/open-telemetry/opentelemetry-js/issues/6379
- The resulting log should be exported through the normal logs pipeline so it appears in the same backend alongside other logs.

### Additional context

Capturing these exceptions as logs provides immediate value for error tracking, incident response, and troubleshooting by ensuring the last failure event is recorded in the OpenTelemetry pipeline. It also reduces the gap between tracing/logs and error tracking, giving a consistent experience across services without forcing each team to add bespoke handlers.

I will send a PR implementing this feature.

iblancasa:

Fixes https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3376

- Introduces a way to capture unhandled exceptions. Those are captured as logs.
- The logs contain the attributes defined by semantic conventions for exceptions https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/


package-lock.json
```diff
@@ -38115,7 +38030,8 @@
       "version": "0.15.1",
       "resolved": "https://registry.npmjs.org/zone.js/-/zone.js-0.15.1.tgz",
       "integrity": "sha512-XE96n56IQpJM7NAoXswY3XRLcWFW83xe0BiAOeMD7K5k5xecOeul3Qcpx6GqEeeHNkW5DWL5zOyTbEfB4eti8w==",
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
```
pichlermarc: hmm, these `"peer": true`'s are a npm bug, please:
- reset package-lock.json to the one from `main`
- update your version of `npm` to latest
- run `npm install --package-lock-only` from repo root
- commit the results

packages/instrumentation-runtime-node/src/instrumentation.ts
```diff
@@ -72,12 +91,136 @@ export class RuntimeNodeInstrumentation extends InstrumentationBase<RuntimeNodeI
...
+      process.on('uncaughtExceptionMonitor', this._onUncaughtExceptionHandler);
+    }
+
+    if (
+      config.captureUnhandledRejection &&
+      !this._onUnhandledRejectionHandler
+    ) {
+      this._onUnhandledRejectionHandler =
+        this._handleUnhandledRejection.bind(this);
+      process.on('unhandledRejection', this._onUnhandledRejectionHandler);
```
trentm: This (likely) changes the behaviour of the application, which I think is a no-go. If I run this program:

```js
console.log('hi');
Promise.reject(new Error('boom'));
console.log('bye');

setInterval(() => {
  console.log('still here')
}, 1000)
```

The result is it crashes out:

```
% node uhr.js
hi
bye
/Users/trentm/src/opentelemetry-js-contrib/uhr.js:8
Promise.reject(new Error('boom'));
               ^

Error: boom
    at Object.<anonymous> (/Users/trentm/src/opentelemetry-js-contrib/uhr.js:8:16)
    at Module._compile (node:internal/modules/cjs/loader:1761:14)
    at Object..js (node:internal/modules/cjs/loader:1893:10)
    at Module.load (node:internal/modules/cjs/loader:1481:32)
    at Module._load (node:internal/modules/cjs/loader:1300:12)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
    at node:internal/main/run_main_module:33:47

Node.js v24.13.0
```

If I change it to have an unhandledRejection handler, 

```js
process.on('unhandledRejection', (reason) => {
  console.error('got an unhandledRejection: reason=%s', reason);
});

console.log('hi');
Promise.reject(new Error('boom'));
console.log('bye');

setInterval(() => {
  console.log('still here')
}, 1000)
```

then it no longer crashes out:

```
% node uhr.js
hi
bye
got an unhandledRejection: reason=Error: boom
    at Object.<anonymous> (/Users/trentm/src/opentelemetry-js-contrib/uhr.js:6:16)
    at Module._compile (node:internal/modules/cjs/loader:1761:14)
    at Object..js (node:internal/modules/cjs/loader:1893:10)
    at Module.load (node:internal/modules/cjs/loader:1481:32)
    at Module._load (node:internal/modules/cjs/loader:1300:12)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
    at node:internal/main/run_main_module:33:47
still here
still here
still here
...
```

Node.js's help output for this option shows that for some values (including the default), the presence of a `unhandledRejection` handler changes app behaviour:

```
  --unhandled-rejections=...             define unhandled rejections behavior. Options are
                                         'strict' (always raise an error), 'throw' (raise an
                                         error unless 'unhandledRejection' hook is set), 'warn'
                                         (log a warning), 'none' (silence warnings),
                                         'warn-with-error-code' (log a warning and set exit code
                                         1 unless 'unhandledRejection' hook is set). (default:
                                         throw)
```

IIUC, this is the reason the `uncaughtExceptionMonitor` event was added: to have a way to monitor uncaught exceptions without changing app behaviour.

packages/instrumentation-runtime-node/test/instrumentation.test.ts
```diff
@@ -15,7 +15,7 @@
  */
 import { MeterProvider } from '@opentelemetry/sdk-metrics';
 
-import { RuntimeNodeInstrumentation } from '../src';
+import { RuntimeNodeInstrumentation } from '../src/index';
```
trentm, line 7: I'm curious why these changes adding "/index"?

packages/instrumentation-runtime-node/src/instrumentation.ts
```diff
@@ -49,6 +67,7 @@ export class RuntimeNodeInstrumentation extends InstrumentationBase<RuntimeNodeI
       for (const collector of this._collectors) {
         collector.enable();
       }
+      this._registerExceptionHandlers();
```
trentm, line 54: I think there is no need for this one when it will be registered in `.enable()`.
(Granted the collector.enable() calls that are already here are in the same boat.)

trentm (changes requested):
Thanks for the PR. I have some reservations about unhandledRejection in comments below.   

I haven't reviewed the particular log record details yet. Please take a look at #6385 merged  just recently (after you created this PR). The Logs SDK spec now includes a mechanism for including an exception. I think this wiould be an obvious candidate to use that.

packages/instrumentation-runtime-node/src/instrumentation.ts
iblancasa: You are totally right. I will revert the `unhandled rejection` stuff.

iblancasa:
> I haven't reviewed the particular log record details yet. Please take a look at #6385 merged just recently (after you created this PR). The Logs SDK spec now includes a mechanism for including an exception. I think this wiould be an obvious candidate to use that.

Yes. I implemented that PR. I guess I have to wait for a new release, right? Maybe we can start working on this one in the meantime. Even create a follow up PR to migrate to that mechanism if we are able to make this one happen before.

packages/instrumentation-runtime-node/README.md
```diff
@@ -4,7 +4,7 @@
 [![Apache License][license-image]][license-image]
 
 This module provides automatic metric instrumentation that exposes measurements from the [Performance measurement APIs](https://nodejs.org/api/perf_hooks.html) (i.e. `perf_hooks`).
-While currently it is limited to metrics, it may be modified to produce other signals in the future.
+It can also emit OpenTelemetry logs for uncaught exceptions and unhandled promise rejections.
```
JacksonWeber: Looks like the PR has been updated to remove the handling for uncaught exceptions. Should update the README.

JacksonWeber (approved):
LGTM with small nit

packages/instrumentation-runtime-node/README.md
iblancasa: Oh.. you are right. Fixing.

packages/instrumentation-runtime-node/src/instrumentation.ts
```diff
@@ -72,12 +81,75 @@ export class RuntimeNodeInstrumentation extends InstrumentationBase<RuntimeNodeI
...
+    const errorLog: LogRecord = {
+      body: 'exception',
+      exception: error,
+      severityNumber,
+      attributes: customAttributes,
+      timestamp,
+      observedTimestamp: timestamp,
+    };
+
+    this.logger.emit(errorLog);
```
JacksonWeber, line 153: The new handler only calls `this.logger.emit(errorLog)` inside `uncaughtExceptionMonitor`. That works with the test's `SimpleLogRecordProcessor`, but the normal logs pipeline commonly uses `BatchLogRecordProcessor`, which buffers records on a timer and exports asynchronously. After an uncaught exception, Node terminates before that timer-driven export can run, so the new FATAL log might be dropped in production. Either provide a crash-safe/synchronous flush path, explicitly scope the feature to synchronous processors, or add an end-to-end process-crash test that proves the record is actually exported.

packages/instrumentation-runtime-node/src/instrumentation.ts
```diff
@@ -27,10 +34,16 @@ import { PACKAGE_VERSION, PACKAGE_NAME } from './version';
 
 const DEFAULT_CONFIG: RuntimeNodeInstrumentationConfig = {
   monitoringPrecision: 10,
+  captureUncaughtException: true,
```
raphael-theriault-swi:
```suggestion
  captureUncaughtException: false,
```

I would feel much more comfortable with this as an opt-in feature, given a buggy `uncaughtExceptionMonitor` handler can cause a number of major issues with the Node process, and the default handler is not trivial enough to confidently know there may not be some subtle issue with it that could arise in the future.

raphael-theriault-swi (changes requested):

packages/instrumentation-runtime-node/src/instrumentation.ts
iblancasa: Ok, it makes sense to me. I will set this as `false` by default.

iblancasa:
Is there are anything else I should address as part of the PR?

iblancasa:
Ping

packages/instrumentation-runtime-node/src/types.ts
```diff
@@ -2,9 +2,24 @@
...
+   */
+  captureUncaughtException?: boolean;
+  /**
+   * Add custom attributes to the emitted exception log records.
+   * @experimental
+   */
+  applyCustomAttributes?: (
+    error: unknown,
+    eventType: 'uncaughtException'
+  ) => Attributes;
```
raphael-theriault-swi, line 24: Sorry to nitpick again but maybe this should be called `applyCustomEventAttributes`/`applyCustomExceptionAttributes` or similar ? Given the main feature of the package is the metrics the name could cause confusion.

raphael-theriault-swi (changes requested):

packages/instrumentation-runtime-node/src/types.ts
iblancasa: Makes sense. Naming is not easy! Applying...

raphael-theriault-swi (approved):
Thank you !

PR #3380, opened 2026-02-10
feat: add neo4j instrumentation

t2t2:

Add neo4j instrumentation, [originally created by aspecto](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/packages/instrumentation-neo4j). Compared to the original, the code has been updated to support newer versions of neo4j-driver package (versions 5.x and 6.x) and to match the latest semantic conventions.


packages/instrumentation-neo4j/src/neo4j.ts
```diff
@@ -0,0 +1,228 @@
...
+        return function (this: unknown, ...args: RunArgs) {
+          if (self.getConfig().ignoreOrphanedSpans) {
+            if (!trace.getSpan(context.active())) {
+              return originalRun.apply(this, args);
+            }
+          }
+
+          const connectionAttributes = getAttributesFromNeo4jSession(this);
+          const query = args[0] as string;
+          const operation = query.trim().split(/\s+/)[0];
```
thompson-tomo: Semconv docs state:

> The operation name SHOULD NOT be extracted from db.query.text, when the database system supports query text with multiple operations in non-batch operations.

packages/instrumentation-neo4j/src/neo4j.ts
```diff
@@ -0,0 +1,228 @@
...
+      (originalRun: neo4j.Session['run']) => {
+        return function (this: unknown, ...args: RunArgs) {
+          if (self.getConfig().ignoreOrphanedSpans) {
+            if (!trace.getSpan(context.active())) {
+              return originalRun.apply(this, args);
+            }
+          }
+
+          const connectionAttributes = getAttributesFromNeo4jSession(this);
+          const query = args[0] as string;
```
thompson-tomo: Has the query been sanitised of sensitive data as mentioned in semconv

> Non-parameterized query text SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data, e.g. by redacting all literal values present in the query text. See Sanitization of db.query.text. Parameterized query text SHOULD be collected by default (the query parameter values themselves are opt-in, see db.query.parameter.<key>).

packages/instrumentation-neo4j/src/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+    [ATTR_DB_NAMESPACE]: session._database ? session._database : 'neo4j',
+  };
+  if (address) {
+    attributes[ATTR_NETWORK_PEER_ADDRESS] = address._host;
+    attributes[ATTR_NETWORK_PEER_PORT] = address._port;
+  }
+
+  if (auth?.principal) {
+    attributes[ATTR_DB_USER] = auth.principal;
+  }
```
thompson-tomo:
```suggestion
```
This attribute is deprecated so I wouldn't add it.

packages/instrumentation-neo4j/src/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+    (session._mode === 'WRITE'
+      ? session._writeConnectionHolder
+      : session._readConnectionHolder) ??
+    session._connectionHolder ??
+    {};
+  const connectionProvider = connectionHolder._connectionProvider ?? {};
+
+  // seedRouter is used when connecting to a url that starts with "neo4j", usually aura
+  const address = connectionProvider._address ?? connectionProvider._seedRouter;
+  const auth = getAuth(connectionProvider);
```
thompson-tomo:
```suggestion
```

packages/instrumentation-neo4j/src/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const attributes: Record<string, any> = {
```
thompson-tomo, line 38: Can we add `server.address`/`server.port`

packages/instrumentation-neo4j/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Installation
+
+```bash
+npm install --save @opentelemetry/instrumentation-mysql2
+```
+
+## Supported Versions
+
+- [`neo4j-driver`](https://www.npmjs.com/package/neo4j-driver) versions `>=4.0.0 <6`
```
seemk: The latest version is 6.x, can we make sure the instrumentation supports it?

packages/instrumentation-neo4j/README.md
```diff
@@ -0,0 +1,65 @@
...
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for the [`neo4j-driver`](https://github.com/neo4j/neo4j-javascript-driver) module, which may be loaded using the [`@opentelemetry/sdk-trace-node`](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-sdk-trace-node) package.
+
+Compatible with OpenTelemetry JS API and SDK `1.0+`.
+
+## Installation
+
+```bash
+npm install --save @opentelemetry/instrumentation-mysql2
```
seemk:
```suggestion
npm install --save @opentelemetry/instrumentation-neo4j
```

packages/instrumentation-neo4j/README.md
```diff
@@ -0,0 +1,65 @@
+# OpenTelemetry Neo4j Instrumentation for Node.js
+
+[![NPM Published Version][npm-img]][npm-url]
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for the [`neo4j-driver`](https://github.com/neo4j/neo4j-javascript-driver) module, which may be loaded using the [`@opentelemetry/sdk-trace-node`](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-sdk-trace-node) package.
+
+Compatible with OpenTelemetry JS API and SDK `1.0+`.
```
seemk, line 8: The SDK is 2.x now, is this valid?

packages/instrumentation-neo4j/src/neo4j.ts
t2t2: The argument is a parameterised query, ref [neo4j usage doc](https://neo4j.com/docs/javascript-manual/current/query-simple/), [api doc](https://neo4j.com/docs/api/javascript-driver/6.0/class/lib6/session.js~Session.html#instance-method-run)

The orig code did not reflect for possible `{text, parameters}` value, updated to also reflect that

packages/instrumentation-neo4j/README.md
t2t2: Confirmed by manually re-running tav, just readme was left oudated

packages/instrumentation-neo4j/README.md
t2t2: [More of a question for >half of the packages in this repo](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-js-contrib+%22Compatible+with+OpenTelemetry+JS+API+and+SDK+%601.0%2B%60.%22&type=code) since most use the same readme text

packages/instrumentation-neo4j/src/neo4j.ts
```diff
@@ -0,0 +1,237 @@
...
+          if (typeof rawQuery === 'string') {
+            query = rawQuery;
+          } else if (rawQuery instanceof String) {
+            query = rawQuery.toString();
+          } else if (typeof rawQuery === 'object' && rawQuery?.text != null) {
+            query = rawQuery.text;
+          }
+
+          const span = self.tracer.startSpan(
+            `${connectionAttributes[ATTR_DB_NAMESPACE]}`,
```
seemk, line 124: https://opentelemetry.io/docs/specs/semconv/db/database-spans/#generating-a-summary-of-the-query

I think it is ok to get  the operation name from the query text. The [earlier comment] was valid only for `db.operation.name` attributes.

thompson-tomo: Span name is constructed with the attribute if it is available https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name

PR #3381, opened 2026-02-12
chore(instrumentation-web-exception): avoid wildcard exports

trentm:
Refs: https://github.com/open-telemetry/opentelemetry-js/issues/4186


overbalance:
And we get rid of barrel files soon...

Issue #3391, opened 2026-02-19
[host-metrics] tooling reports CVE-2026-26318 through `systeminformation` (unused code-path)


monwolf:
Hi,

Our SCA tools are alerting that the package @opentelemetry/host-metrics is impacted by CVE-2026-26318. Could you update the version of hostsmetrics? 

Regards

pichlermarc:
Hi @monwolf - it looks like the vulnerable code-path is not used by our package. We can still update though.

trentm:
Updated in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3392
I guess this will be "fixed" once we have a new release out.

PR #3392, opened 2026-02-19
fix(instrumentation-host-metrics): unpin and update to systeminformation@^5.31.1

pichlermarc:

See #3391, a vulnerability in an unused code-path is triggering alerts for end-users. This also unpins the package, so that future patches are pulled in without us needing to publish a new version.


Krishan27:
Thank you for merging it and releasing it it was causing us issue.

PR #3403, opened 2026-02-25
test: avoid a race in test-services:start:ci

trentm:
In PRs with a number of affected packages, there was a race in
test-services:start:ci where two packages could be attempting to
start the same docker compose service at the same time.
E.g. instr-redis and instr-ioredis both use the 'redis' service.

This PR changes to run a single 'docker compose up ...' command,
by first getting the set of docker services via a test-services:ls
run on each affected package.


trentm:
See also https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3402 that does similar for `test-services:stop:ci`.

PR #3404, opened 2026-02-25
feat(hapi-instrumentation): extend span name with ext method name

afharo:

- Hapi `ext` spans are hard to pinpoint to the underlying registered extension

image

- When the function has a name, it is appended to the end of the span name.

image

### DIFF
--- a/packages/instrumentation-hapi/src/utils.ts
+++ b/packages/instrumentation-hapi/src/utils.ts
@@ -117,7 +117,8 @@ export const getExtMetadata = (
   name: string;
 } => {
   let baseName = `ext - ${extPoint}`;
-  if (methodName && methodName !== 'method') { // method is the default name for the extension in the ServerExtEventsObject format.
+  if (methodName && methodName !== 'method') {
+    // method is the default name for the extension in the ServerExtEventsObject format.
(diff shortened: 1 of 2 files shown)


packages/instrumentation-hapi/src/utils.ts
```diff
@@ -110,27 +110,33 @@ export const getRouteMetadata = (
...
+  pluginName?: string,
+  methodName?: string
 ): {
   attributes: Attributes;
   name: string;
 } => {
+  let baseName = `ext - ${extPoint}`;
+  if (methodName && methodName !== 'method') {
+    // method is the default name for the extension in the ServerExtEventsObject format.
+    baseName = `ext - ${extPoint} - ${methodName}`;
```
trentm, line 122: I like it. This is much better.

FWIW, there is also some prior art in the express instrumentation. Doing this:

```js
const app = express();
app.use(async function myMiddleware(req, res, next) {
  // ...
  next();
});
```

will result in an Express span named `middleware - myMiddleware`.

trentm (approved):
Assuming tests pass.
LGTM.
Sorry for the delay in reviewing.

PR #3422, opened 2026-03-09
fix(instrumentation-aws-lambda): handle non-configurable exports from esbuild CJS bundles

### LINKED ISSUE #1781: Node.js layer: "Cannot redefine property: handler" when using esbuild
**Describe the bug**
The Node.js layer is working correctly when using a handwritten CommonJS handler like this:

```js
module.exports = {
  handler: () => 200
}
```

My real code is written in ES Modules and is bundled via esbuild. Compile the following code with esbuild and then deploy it to Lambda.

```js
export const handler = () => 200;
```

Now, the Lambda fails with the error

```
{
  "errorType": "TypeError",
  "errorMessage": "Cannot redefine property: handler",
  "stack": [
    "TypeError: Cannot redefine property: handler",
    "    at Function.defineProperty (<anonymous>)",
    "    at n (/opt/wrapper.js:1:242970)",
    "    at i (/opt/wrapper.js:1:243369)",
    "    at o._wrap (/opt/wrapper.js:1:252372)",
    "    at x.patch (/opt/wrapper.js:1:204436)",
    "    at /opt/wrapper.js:1:254659",
    "    at Array.reduce (<anonymous>)",
    "    at t._onRequire (/opt/wrapper.js:1:254445)",
    "    at t (/opt/wrapper.js:1:255685)",
    "    at Module.T (/opt/wrapper.js:1:92076)"
  ]
}
```

**Steps to reproduce**

1. Clone the opentelemetry-lambda repo and execute `npm run build` in the `nodejs` directory. This produces a `layer.zip` file.
2. Upload `layer.zip` as a layer via the AWS Console.
3. Create a new Lambda function (Node.js 22.x) and add the layer to it.
4. In the environment variables for the Lambda, set `AWS_LAMBDA_EXEC_WRAPPER` to `/opt/otel-handler`.
5. Locally, create `index.js` as 
    ```js
    export const handler = () => 200;
    ```
6. Bundle the code using esbuild by running `npx esbuild index.js --platform=node --bundle --outfile=compiled.js`.  [View the compiled bundle here](https://github.com/smagura-jm/otel-repro/blob/main/compiled.js)
7. Paste the compiled code into the code editor for the Lambda function in the AWS Console. Click "Deploy".
8. In the AWS Console, test the Lambda function with an arbitrary payload.

**What did you expect to see?**
The Lambda succeeds.

**What did you see instead?**
The Lambda fails with the error

```
{
  "errorType": "TypeError",
  "errorMessage": "Cannot redefine property: handler",
  "stack": [
    "TypeError: Cannot redefine property: handler",
    "    at Function.defineProperty (<anonymous>)",
    "    at n (/opt/wrapper.js:1:242970)",
    "    at i (/opt/wrapper.js:1:243369)",
    "    at o._wrap (/opt/wrapper.js:1:252372)",
    "    at x.patch (/opt/wrapper.js:1:204436)",
    "    at /opt/wrapper.js:1:254659",
    "    at Array.reduce (<anonymous>)",
    "    at t._onRequire (/opt/wrapper.js:1:254445)",
    "    at t (/opt/wrapper.js:1:255685)",
    "    at Module.T (/opt/wrapper.js:1:92076)"
  ]
}
```

**What version of collector/language SDK version did you use?**
I cloned the opentelemetry-lambda repo, and the latest commit on main was 939cecd57868731b816e85131e86e791ada44bde.

**What language layer did you use?**
Node.js

RaphaelManke:
## Summary
- When esbuild bundles ESM to CJS format, it defines exports using non-configurable accessor descriptors. OpenTelemetry's shimmer then fails with `Cannot redefine property` when trying to wrap the handler.
- Adds a `makeExportsConfigurable()` helper in the Lambda instrumentation's patch callback that replaces such exports with a new object having configurable data descriptors before shimmer attempts patching.

Fixes: open-telemetry/opentelemetry-lambda#1781
See: https://github.com/evanw/esbuild/issues/2199

## Steps to reproduce

Create a handler that uses the same non-configurable export pattern that esbuild emits when bundling ESM to CJS:

```js
// index.js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
Object.defineProperty(exports, "handler", {
  enumerable: true,
  get: function () { return handler; }
  // intentionally no configurable: true — this is what esbuild produces
});

async function handler(event) {
  return { statusCode: 200, body: JSON.stringify({ message: "OK" }) };
}
```

Deploy as a Lambda with the OTel layer and `AWS_LAMBDA_EXEC_WRAPPER=/opt/otel-handler`:

```sh
zip handler.zip index.js

aws lambda create-function \
  --function-name otel-cjs-repro \
  --runtime nodejs20.x \
  --role <execution-role-arn> \
  --handler index.handler \
  --zip-file fileb://handler.zip \
  --layers arn:aws:lambda:<region>:184161586896:layer:opentelemetry-nodejs-0_20_0:1 \
  --environment "Variables={AWS_LAMBDA_EXEC_WRAPPER=/opt/otel-handler}"

aws lambda invoke --function-name otel-cjs-repro \
  --cli-binary-format raw-in-base64-out \
  --payload '{}' response.json && cat response.json
```

**Error (before fix):**
```json
{
  "errorType": "TypeError",
  "errorMessage": "Cannot redefine property: handler",
  "trace": [
    "TypeError: Cannot redefine property: handler",
    "    at Function.defineProperty (<anonymous>)",
    "    at T._wrap (...)",
    "    at Z.patch (...)",
    "    at T._onRequire (...)"
  ]
}
```

**Root cause:** esbuild emits exports as non-configurable accessor descriptors (see [esbuild#2199](https://github.com/evanw/esbuild/issues/2199)), so `Object.defineProperty` in shimmer fails when trying to wrap the handler.

## Test plan
- [x] Added `test/lambda-test/esbuild-cjs.js` simulating esbuild's non-configurable export pattern
- [x] Added integration tests verifying both success and error spans work with esbuild-style exports
- [x] Verified tests fail without the fix (`TypeError: Cannot redefine property: handler`)
- [x] Verified all 55 existing tests continue to pass
- [x] Normal CJS exports (already configurable) return the same object without creating a copy

### DIFF
--- a/packages/instrumentation-aws-lambda/src/instrumentation.ts
+++ b/packages/instrumentation-aws-lambda/src/instrumentation.ts
@@ -91,6 +91,34 @@ function isSupportingCallbacks(): boolean {
   return false;
 }
 
+// When esbuild bundles ESM to CJS format, it defines exports using non-configurable
+// accessor descriptors. OpenTelemetry's shimmer then fails with "Cannot redefine property"
+// when trying to wrap the handler. This replaces such exports with a new object that has
+// configurable data descriptors so shimmer can wrap them.
+// See: https://github.com/evanw/esbuild/issues/2199
+function makeExportsConfigurable<T>(moduleExports: T): T {
+  if (typeof moduleExports !== 'object' || moduleExports === null) {
+    return moduleExports;
+  }
+  const keys = Object.getOwnPropertyNames(moduleExports);
+  const descriptors = Object.getOwnPropertyDescriptors(moduleExports);
+  if (keys.every(key => descriptors[key].configurable)) {
+    return moduleExports;
+  }
+  const fixed = Object.create(
+    Object.getPrototypeOf(moduleExports)
+  ) as T;
+  for (const key of keys) {
+    Object.defineProperty(fixed, key, {
+      value: (moduleExports as Record<string, unknown>)[key],
+      writable: true,
+      enumerable: descriptors[key].enumerable,
+      configurable: true,
+    });
+  }
+  return fixed;
+}
+
 export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstrumentationConfig> {
   declare private _traceForceFlusher?: () => Promise<void>;
   declare private _metricForceFlusher?: () => Promise<void>;
@@ -175,6 +203,7 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
             module,
             ['*'],
             (moduleExports: LambdaModule) => {
+              moduleExports = makeExportsConfigurable(moduleExports);
               if (isWrapped(moduleExports[functionName])) {
                 this._unwrap(moduleExports, functionName);
               }
--- a/packages/instrumentation-aws-lambda/src/instrumentation.ts
+++ b/packages/instrumentation-aws-lambda/src/instrumentation.ts
@@ -105,9 +105,7 @@ function makeExportsConfigurable<T>(moduleExports: T): T {
   if (keys.every(key => descriptors[key].configurable)) {
     return moduleExports;
   }
-  const fixed = Object.create(
-    Object.getPrototypeOf(moduleExports)
-  ) as T;
+  const fixed = Object.create(Object.getPrototypeOf(moduleExports)) as T;
   for (const key of keys) {
     Object.defineProperty(fixed, key, {
       value: (moduleExports as Record<string, unknown>)[key],
--- a/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
+++ b/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.test.ts
@@ -736,6 +736,38 @@ describe('lambda handler', () => {
     });
   });
 
+  describe('esbuild CJS bundle with non-configurable exports', () => {
+    it('should export a valid span', async () => {
+      initializeHandler('lambda-test/esbuild-cjs.handler');
+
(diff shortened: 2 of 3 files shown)


raphael-theriault-swi (commented):
Solution looks good, but I'm wondering whether we might want to tackle this from the instrumentation package given it's not technically unique to Lambda

RaphaelManke:
Can you point me in a direction? I am not sure which instrumentation you are referring to? Like in the core repo?

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -91,6 +91,32 @@ function isSupportingCallbacks(): boolean {
...
+// See: https://github.com/evanw/esbuild/issues/2199
+function makeExportsConfigurable<T>(moduleExports: T): T {
+  if (typeof moduleExports !== 'object' || moduleExports === null) {
+    return moduleExports;
+  }
+  const keys = Object.getOwnPropertyNames(moduleExports);
+  const descriptors = Object.getOwnPropertyDescriptors(moduleExports);
+  if (keys.every(key => descriptors[key].configurable)) {
+    return moduleExports;
+  }
```
jj22ee, line 107: Can this just check if `functionName` is configurable instead of every key?
```
const desc = Object.getOwnPropertyDescriptor(moduleExports, functionName);
if (desc && !desc.configurable) {
  // your logic to clone the module with configurable descriptors
}
```

jj22ee: This would also make this change a bit more unique to Lambda.

Issue #3425, opened 2026-03-10
feat(instrumentation-pino): use Logs API exception field when translating pino error records


iblancasa:
`@opentelemetry/instrumentation-pino` currently translates pino log records without using the new Logs API `exception` field introduced in https://github.com/open-telemetry/opentelemetry-js/pull/6385.

As a result, pino error payloads are treated as regular attributes instead of being processed via the SDK exception mapping (`exception.type`, `exception.message`, `exception.stacktrace`).

raphael-theriault-swi:
Closed by #3426

PR #3426, opened 2026-03-10
feat(instrumentation-pino): use Logs API exception field when translating pino error records

iblancasa:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3425

- Call the exceptions API added in https://github.com/open-telemetry/opentelemetry-js/pull/6385 from `pino` auto instrumentation.


seemk (approved):

iblancasa:
@open-telemetry/javascript-approvers can I get some review or merge for this PR? Thank you

packages/instrumentation-pino/src/log-sending-utils.ts
```diff
@@ -223,9 +227,26 @@ export class OTelPinoStream extends Writable {
...
+  if (!exception || typeof exception !== 'object' || Array.isArray(exception)) {
+    return exception;
+  }
+
+  const exceptionObject = exception as Record<string, unknown>;
+  if (
+    typeof exceptionObject['type'] === 'string' &&
+    typeof exceptionObject['name'] !== 'string'
+  ) {
+    return { ...exceptionObject, name: exceptionObject['type'] };
```
pichlermarc: **question:** the object we're copying here is a new one we created anyway through `JSON.parse(s)` above - wouldn't it be more efficient to just define `name` on the existing object instead of allocating a new one?

iblancasa: Good point. I will push a new commit with the change.

pichlermarc (approved):

Issue #3418, opened 2026-03-06
"test-all-versions" (TAV) tests for instrumentation-sequelize are failing, only in CI


trentm:
https://github.com/open-telemetry/opentelemetry-js-contrib/actions/workflows/test-all-versions.yml started failing on Feb 25:

- last working run: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/22383067651
- first failing run: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/22428638731

Failures on in the " Run test-all-versions (24)" step, running the tests against sequelize@6.37.7 (the current latest release of sequelize, and the version also installed in our package-lock, meaning the same version that the tests run against for non-TAV tests).

A snippet of the failure:

```
-- required packages ["sequelize@6.37.7"]
-- installing ["sequelize@6.37.7"]
-- running test "npm run test" with sequelize (env: {})

> @opentelemetry/instrumentation-sequelize@0.6.0 test
> nyc --no-clean mocha --require '@opentelemetry/contrib-test-utils' 'test/**/*.test.ts'


 Exception during run: Error: Directory import '/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-sequelize/src' is not supported resolving ES modules imported from /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-sequelize/test/sequelize.test.ts
```

<details>
<summary>More complete excerpt of the test failure:</summary>

```
> @opentelemetry/instrumentation-sequelize@0.6.0 test-all-versions
> tav

-- required packages ["sequelize@6.37.7"]
-- installing ["sequelize@6.37.7"]
-- running test "npm run test" with sequelize (env: {})

> @opentelemetry/instrumentation-sequelize@0.6.0 test
> nyc --no-clean mocha --require '@opentelemetry/contrib-test-utils' 'test/**/*.test.ts'


 Exception during run: Error: Directory import '/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-sequelize/src' is not supported resolving ES modules imported from /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-sequelize/test/sequelize.test.ts
    at finalizeResolution (node:internal/modules/esm/resolve:263:11)
    at moduleResolve (node:internal/modules/esm/resolve:865:10)
[...]

PR #3132, opened 2025-10-01
feat(instrumentation-langchain): add initial package skeleton for instrumentation-langchain

yiyuan-he:
## Which problem is this PR solving?
This PR adds the initial project skeleton to kickstart a new `instrumentation-langchain` package. 

Original PR raised by our intern: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3024

## Short description of the changes
Mostly package metadata and licensing information. @mxiamxia and I (@yiyuan-he) also volunteer to be component owners for this new package.

PR #3429, opened 2026-03-13
test(instrumentation-sequelize): ensure have devDeps necessary for tests

trentm:
Closes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3418


package-lock.json
```diff
@@ -6423,42 +6423,6 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/@langchain/langgraph/node_modules/react": {
```
trentm, line 6426: There is some churn here in package-lock related to #3132, presumably. I don't *think* that is an issue, but I'm not positive.

Issue #3430, opened 2026-03-13
feat(instrumentation-bunyan): use Logs API exception field when translating bunyan error records


iblancasa:
`@opentelemetry/instrumentation-bunyan` currently translates bunyan log records without using the new Logs API `exception` field introduced in https://github.com/open-telemetry/opentelemetry-js/pull/6385.

Implementing this feature, bunyan errors can be processed using the SDK exception mapping (`exception.type`, `exception.message`, `exception.stacktrace`).

raphael-theriault-swi:
Closed by #3431

PR #3431, opened 2026-03-13
feat(instrumentation-bunyan): use Logs API exception field when translating bunyan error records

iblancasa:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3430

- Call the exceptions API added in https://github.com/open-telemetry/opentelemetry-js/pull/6385 from `bunyan` auto instrumentation.


raphael-theriault-swi (approved):
LGTM

Issue #2647, opened 2025-01-15
Ownership Transition for Fastify Instrumentation


metcoder95:
- [x] This only affects the JavaScript OpenTelemetry library
- [ ] This may affect other libraries, but I would like to get opinions here first

Hello! 👋 
I'm part of the Fastify Core Collaborators, and I'm opening this issue to continue the efforts towards the ownership signup for the `@opentelemetry/instrumentation-fastify` instrumentation.

After we received the following issue https://github.com/fastify/fastify/issues/5862 to sign up as component owners of the instrumentation library previously mentioned, we decided internally to adopt and issue within Fastify organization an official instrumentation for Fastify: https://github.com/fastify/otel

As a further step, we would like to know your thoughts and ask you if mark the `@opentelemetry/instrumentation-fastify` as a deprecated library in favor of `@fastify/otel`.

We will be maintaining the instrumentation under our umbrella from now on, but we would like to communicate this to further users so they are able to transition into the new instrumentation.

Thanks!

pichlermarc:
Hi @metcoder95 - this is great news! We're super happy to see this being officially instrumented by fastify authors 🎉
I'll sync with the other maintainers today how we go about deprecating the component - I'll document what we talked about here. 

I expect that we're going to:
- update the README.md that the package is now deprecated, pointing to https://github.com/fastify/otel as the replacement
- add a `@deprecated` annotation to point towards `@fastify/otel` as a replacement
- deprecate `@opentelemetry/instrumentation-fastify` in npm and point towards `@fastify/otel` as a replacment
- remove the code here so that only the `README.md` remains for people that are looking for `@opentelemetry/instrumentation-fastify`

While we're doing that, I think adding the instrumentation to the [OTel Registry](https://opentelemetry.io/ecosystem/registry/?s=&component=&language=&flag=all) (see [instructions](https://opentelemetry.io/ecosystem/registry/adding/)) would help a lot with discoverability of the new instrumentation. 🙂

pichlermarc:
> I'll sync with the other maintainers today how we go about deprecating the component - I'll document what we talked about here.

Alright, this is the full plan:

- As a first step, we'll modify the `@opentelemetry/auto-instrumentations-node` package and make `@opentelemetry/instrumentation-fastify` default-disabled, we'll document how users can move forward with `@fastify/otel`. 
[...]

PR #3435, opened 2026-03-13
feat(nestjs): add instrumentation for NestJS microservice

neilime:

- NestJS instrumentation currently only covers the core HTTP/controller lifecycle through `@opentelemetry/instrumentation-nestjs-core`, so Nest microservice applications and hybrid apps do not produce dedicated spans for microservice bootstrapping or message/event handlers.
- This change adds first-class NestJS microservice instrumentation while keeping the existing core-only package available as a compatibility path for current users.

- Add a new `@opentelemetry/instrumentation-nestjs` package that instruments both Nest HTTP controllers and Nest microservice handlers.
- Create spans for standalone and hybrid microservice startup (`NestFactory.createMicroservice` and `app.connectMicroservice`) and for message/event handler execution.
- Add Nest-specific attributes for microservice tracing, including handler callback, pattern, and transport information.
- Keep `@opentelemetry/instrumentation-nestjs-core` as a deprecated compatibility wrapper that delegates to the new package with microservice instrumentation disabled.
- Register the new package in `@opentelemetry/auto-instrumentations-node` and update the related README/docs.
- Add tests covering HTTP behavior, standalone microservices, and hybrid microservices.

### DIFF
--- a/.github/component-label-map.yml
+++ b/.github/component-label-map.yml
@@ -161,6 +161,10 @@ pkg:instrumentation-nestjs-core:
   - changed-files:
       - any-glob-to-any-file:
           - packages/instrumentation-nestjs-core/**
+pkg:instrumentation-nestjs:
+  - changed-files:
+      - any-glob-to-any-file:
+          - packages/instrumentation-nestjs/**
 pkg:instrumentation-net:
   - changed-files:
       - any-glob-to-any-file:
--- a/.github/component_owners.yml
+++ b/.github/component_owners.yml
@@ -127,7 +127,9 @@ components:
     # Unmaintained
   packages/instrumentation-mysql2:
     - raphael-theriault-swi
-  packages/instrumentation-nestjs-core:
+  packages/instrumentation-nestjs-core: []    
+    # Unmaintained
+  packages/instrumentation-nestjs:
     - neilime
   packages/instrumentation-net:
     - seemk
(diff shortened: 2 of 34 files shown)


neilime:
Hi @blumamir, @dyladan, @[legendecas, @pichlermarc,

I will need you opinion on this PR.

I'm convinced that having instrumentation for just nestjs-core it to restrictive. Having an instrumation by nestjs package is bloated.

IMO the best solution is to deprecated the original nestjs-core instrumentation and forge a new nestjs (only) instrumentation.

I have made the code, it's working good, battle tested and backward compatible.

But I dont how to deal with de preciation and the new package. I've made what I supposed to be ok, but I'll appreciate your recommandation to mathc with repository practices.

Thanks

david-luna:
Hi @neilime 

Thanks for working on this. I'm okay with having a new and more generic instrumentation (let's wait for more feedback). IMHO I would split this PR into smaller ones with this order

- [ ] PR to add the new instrumentation (this one)
- [ ] wait for the publication of the new instrumentation
- [ ] PR to add the deprecation notice into `@opentelemetry/instrumentation-nestjs-core`
- [ ] PR to replace it in auto instrumentations
- [ ] after a period of time remove the deprecated instrumentation. Leaving a README for people looking at the instrumentation

`@opentelemetry/instrumentation-fastify` had a similar process. Ref: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2647

If you agree with this process I think this PR should only contain the new instrumentation. And I'd prefer to not refactor the current instrumentation so we avoid maintenance on it.

david-luna:
@neilime any feedback from my previous comment?

PR #3436, opened 2026-03-13
fix(instrumentation-sequelize): do not include 'server.address' for SQLite DB spans

trentm:
I've tweaked the instr to not include server.address / server.port
for SQLite spans, because it doesn't make sense. Before this change
it was using the SQLite *filename* as the `server.address`.
This is not discussed in DB semantic conventions.
I've asked about this in semconv Slack.

I've also corrected the connection string for a SQLite in-memory
database in the tests. The previous usage was creating an actual local
*file* named 'memory'.

### DIFF
--- a/packages/instrumentation-sequelize/test/sequelize.test.ts
+++ b/packages/instrumentation-sequelize/test/sequelize.test.ts
@@ -326,7 +326,8 @@ describe('instrumentation-sequelize', () => {
   });
 
   describe('sqlite', () => {
-    const instance = new sequelize.Sequelize('sqlite:memory', {
+    // https://sequelize.org/docs/v6/getting-started/#connecting-to-a-database
+    const instance = new sequelize.Sequelize('sqlite::memory:', {
       logging: false,
     });
     instance.define('User', {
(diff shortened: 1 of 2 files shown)


packages/instrumentation-sequelize/test/sequelize.test.ts
```diff
@@ -341,7 +342,10 @@ describe('instrumentation-sequelize', () => {
       assert.strictEqual(spans.length, 1);
       const attributes = spans[0].attributes;
       assert.strictEqual(attributes[ATTR_DB_SYSTEM_NAME], 'sqlite');
-      assert.strictEqual(attributes[ATTR_SERVER_ADDRESS], 'memory');
+      // db.namespace: I think we'd expect ":memory:" here, with the trailing
```
maryliag, line 345: is this a comment just for reviewers or was to plan to keep this?
asking because the "I think" might not be the best comment 😅

trentm:
> I've asked about this in semconv Slack.

https://cloud-native.slack.com/archives/C041APFBYQP/p1773437220978069

packages/instrumentation-sequelize/test/sequelize.test.ts
```diff
@@ -341,7 +342,10 @@ describe('instrumentation-sequelize', () => {
...
+      // colon as at https://www.sqlite.org/inmemorydb.html. However sequelize
+      // Sequelize#config.database drops the trailing ':'.
+      assert.strictEqual(attributes[ATTR_DB_NAMESPACE], ':memory');
```
seemk, line 348: This is indeed bizarre, but seems to be a bug in sequelize, where the sqlite path string is parsed as an URL: https://github.com/sequelize/sequelize/blob/v6.37.8/src/sequelize.js#L198

Since we already have a check for the database dialect, perhaps we can fix the name there as well? I.e. `:memory` would translate to `:memory:`

Issue #3357, opened 2026-01-25
[instrumentation-express] possible bug in baggage propagation


antechrestos:
### What version of OpenTelemetry are you using?

2.2.0

### What version of Node are you using?

v22.21.1

### What did you do?

Initiated opentelemtry 

```
const traceExporter = new OTLPTraceExporter();
const metricReader = new PeriodicExportingMetricReader({
    // exporter: new ConsoleMetricExporter()
    exporter: new OTLPMetricExporter()
});

const resource = resourceFromAttributes({
    [ATTR_SERVICE_NAME]: global.APPLICATION,
    [ATTR_SERVICE_VERSION]: global.VERSION
});
const sdk = new opentelemetry.NodeSDK({
    resource: resource,
    textMapPropagator: new CompositePropagator({
        propagators: [new W3CTraceContextPropagator(), new W3CBaggagePropagator()]
    }),
    traceExporter,
    metricReaders: [metricReader],
    contextManager: new AsyncLocalStorageContextManager(),
    instrumentations: [
        getNodeAutoInstrumentations({
            '@opentelemetry/instrumentation-express': {},
            '@opentelemetry/instrumentation-http': {
                startIncomingSpanHook: (request) => {
                    return {
                        ["http.method"]: request.method,
                    }
                }
            },
            '@opentelemetry/instrumentation-winston': {
                logHook: (span, record) => {
                    record['resource.service.name'] = resource.attributes[ATTR_SERVICE_NAME];
                    record['resource.service.version'] = resource.attributes[ATTR_SERVICE_VERSION];
                    const currentBaggage = propagation.getActiveBaggage()

                    if (currentBaggage !== undefined) {
                        for(const key of ['client.id', 'user.id', 'graphql.operation', 'leav.attribute']) {
                            const baggageValue = currentBaggage.getEntry(key);
                            if(baggageValue !== undefined) {
                                record[key] = baggageValue.value;
                            }
                        }
                    }

                }
            }
        }) // liste des instrumentations supportées https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node
    ]
});
```

Also, I defined the middleware in express

```
          app.use((req, res, next) => {
            req.baggage = (propagation.getActiveBaggage() || propagation.createBaggage())
                .setEntry('client.id', {value: 'client-id'})
                .setEntry('user.id', {value: 'some-user'});

            context.with(propagation.setBaggage(context.active(), req.baggage), () => {
[...]

PR #3437, opened 2026-03-14
fix(instrumentation-express): propagate baggage in middleware

tgriesser:

#3357 - baggage propagation broken in Express middleware

Currently the `next()` callback is bound to the `parentContext` (the context from before the middleware ran), which discards any context modifications the middleware made.

This was introduced in #2638 to prevent sibling middleware spans from accidentally nesting, since if each middleware creates a span, the next middleware's span would be created as a child of the previous one without some sort of rebinding. The fix was correct for span nesting but has the side effect of stripping baggage that might get created in the middleware. Since it's a popular pattern to add baggage via things that you pick up in the middleware (auth, params, etc.) this is an issue.

So now instead of discarding the entire active context, we take the current active context after the middleware was executed which includes the baggage and only replaces the span with the parent span, which should give us what we're after.

Spent several hours trying to figure out why this was broken, and then came across #3357 once I figured it out, so thought I'd take a shot at fixing it even though it was already assigned to @pratik50

### DIFF
--- a/packages/instrumentation-express/test/express.test.ts
+++ b/packages/instrumentation-express/test/express.test.ts
@@ -39,12 +39,7 @@ instrumentation.enable();
 instrumentation.disable();
 
 import * as express from 'express';
-import {
-  RPCMetadata,
-  RPCType,
-  getRPCMetadata,
-  setRPCMetadata,
-} from '@opentelemetry/core';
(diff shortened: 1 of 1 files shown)


tgriesser:
@raphael-theriault-swi thanks for the review! I know how it goes, I have also accidentally broken OSS code quite a few times code over the years 😄

Issue #3440, opened 2026-03-16
feat(instrumentation-wiston): use Logs API exception field when translating bunyan error records


iblancasa:
`@opentelemetry/instrumentation-wiston ` currently translates wiston log records without using the new Logs API `exception` field introduced in https://github.com/open-telemetry/opentelemetry-js/pull/6385.

Implementing this feature, wiston errors can be processed using the SDK exception mapping (`exception.type`, `exception.message`, `exception.stacktrace`).

PR #3441, opened 2026-03-16
feat(instrumentation-winston): use Logs API exception field when translating winston error records

iblancasa:

- https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3440

- Call the exceptions API added in https://github.com/open-telemetry/opentelemetry-js/pull/6385 from `bunyan` auto instrumentation.

### DIFF
--- a/packages/winston-transport/src/utils.ts
+++ b/packages/winston-transport/src/utils.ts
@@ -15,6 +15,7 @@
  */
 
 import {
+  AnyValue,
   LogAttributes,
   LogRecord,
   Logger,
@@ -59,7 +60,11 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
   return npmLevels[level] ?? sysLoglevels[level] ?? cliLevels[level];
 }
 
-function isExceptionCandidate(value: unknown): boolean {
+/**
+ * Mirrors the exception shapes handled by sdk-logs `LogRecordImpl._setException`.
+ * We cannot reuse that internal method directly from this package.
+ */
+function isSdkLogsExceptionCandidate(value: unknown): boolean {
   if (typeof value === 'string' || typeof value === 'number') {
     return true;
   }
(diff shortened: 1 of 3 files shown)


packages/winston-transport/src/utils.ts
```diff
@@ -59,6 +59,86 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
   return npmLevels[level] ?? sysLoglevels[level] ?? cliLevels[level];
 }
 
+function isExceptionCandidate(value: unknown): boolean {
```
hectorhdzg: Is there similar code in OTel SDK that we can reuse for this?

packages/winston-transport/src/utils.ts
```diff
@@ -69,8 +149,15 @@ export function emitLogRecord(
   // accidental inclusion of ANSI color codes that may be present in the string
   // property.
   const levelSym = record[Symbol.for('level')];
+  const exceptionPayload = getExceptionPayload(record);
+  const excludedAttributes = new Set(
```
hectorhdzg, line 227: When winston throws an internal error will include extra attributes like process, os, trace, etc. Do we want to include those as well?, these are not being excluded currently
https://github.com/winstonjs/winston/blob/master/lib/winston/exception-handler.js#L75

packages/winston-transport/src/utils.ts
```diff
@@ -59,6 +59,86 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
...
+  return (
+    typeof exception.code === 'string' ||
+    typeof exception.code === 'number' ||
+    typeof exception.name === 'string' ||
+    typeof exception.message === 'string' ||
+    typeof exception.stack === 'string'
+  );
+}
+
+function getExceptionPayload(record: Record<string | symbol, any>): {
```
hectorhdzg, line 159: Do we want to support error,cause? do you know if that is the specification?
https://nodejs.org/api/errors.html#errorcause

packages/winston-transport/src/utils.ts
```diff
@@ -59,6 +59,86 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
...
+
+  return (
+    typeof exception.code === 'string' ||
+    typeof exception.code === 'number' ||
+    typeof exception.name === 'string' ||
+    typeof exception.message === 'string' ||
+    typeof exception.stack === 'string'
+  );
+}
+
```
hectorhdzg, line 90:
```suggestion

/**
 * Attempts to extract an exception/error from a Winston log record.
 *
 * Winston records can carry error information in several ways depending on
 * how the user logged and which formats are configured. This function checks
 * three sources in priority order:
 *
 * 1. Named fields ('err', 'error') — e.g. from Winston's ExceptionHandler
 *    which sets record.error for uncaught exceptions, or from user code that
 *    manually attaches an error to the record metadata.
 *
 * 2. Splat args (Symbol.for('splat')) — positional arguments passed via
 *    logger.log('error', 'message', errorObj). Winston stores extra args
 *    under the splat symbol.
 *
 * 3. Flattened stack field — when winston.format.errors({ stack: true }) is
 *    used, the Error is dissolved into top-level record properties (message,
 *    stack, name, code). We reconstruct an exception object from those fields
 *    and parse the error type from the stack header if 'name' isn't available.
 *
 * @returns The extracted exception and a list of record keys that should be
 *          excluded from OTel log attributes (to avoid duplication), or null
 *          if no error was found.
 */
```

packages/winston-transport/src/utils.ts
iblancasa: It is an internal package. Will add a comment here.

packages/winston-transport/src/utils.ts
iblancasa: Good idea.

packages/winston-transport/src/utils.ts
iblancasa: These are the semantic conventions https://opentelemetry.io/docs/specs/otel/trace/exceptions/
I will try to support `error.cause` in the best way.

iblancasa:
Is there anything else I should address as part of the PR? Thank you.

iblancasa:
Ping.

hectorhdzg (approved):
It would be good in the future to refactor and add some of the functionalities here is some common package, so all LogRecord Instrumentations can reuse

iblancasa:
Anything else as part of this PR? I will address @hectorhdzg comment as part of a different PR.

Issue #3444, opened 2026-03-18
Console Instrumentation


hectorhdzg:
### Which library do you want to instrument?
Node.js runtime console

### Code Owners

hectorhdzg
jacksonweber

We have support for Console auto generation for LogRecords in Azure for several years now, having support for this upstream in OpenTelemetry would be a great thing to have.

I believe there was an issue created a while ago about having support for this but I was not able to find it. so creating new instrumentation request instead

hectorhdzg:
Instrumentation is available now https://www.npmjs.com/package/@opentelemetry/instrumentation-console

PR #3445, opened 2026-03-18
feat(instrumentation-console): Add console intrumentation

hectorhdzg:
Adding console instrumentation, following similar configurations and functionality in other Log instrumentations in this project like Bunyan, Winston and Pino

#3444

### DIFF
--- a/packages/instrumentation-console/README.md
+++ b/packages/instrumentation-console/README.md
@@ -1,0 +1,76 @@
+# OpenTelemetry Instrumentation for Node.js Console
+
+[![NPM Published Version][npm-img]][npm-url]
+[![Apache License][license-image]][license-image]
+
+This module provides automatic instrumentation for the Node.js [`console`](https://nodejs.org/api/console.html) module, generating OpenTelemetry LogRecords for console method calls (`console.log`, `console.error`, `console.warn`, etc.).
+
+## Installation
+
+```bash
+npm install @opentelemetry/instrumentation-console
+```
+
+## Supported Versions
+
+- Node.js `^18.19.0 || >=20.6.0`
+
+## Usage
+
+```javascript
+const { NodeSDK } = require('@opentelemetry/sdk-node');
+const { ConsoleInstrumentation } = require('@opentelemetry/instrumentation-console');
+
+const sdk = new NodeSDK({
+  instrumentations: [new ConsoleInstrumentation()],
+});
+sdk.start();
+
+// Now console calls will generate LogRecords
+console.log('Hello, world!');    // severity: INFO
+console.warn('Watch out!');      // severity: WARN
+console.error('Something bad');  // severity: ERROR
+```
+
+## Console Methods Instrumented
+
+| Console Method   | Severity Number | Severity Text |
+| ---------------- | --------------- | ------------- |
+| `console.trace`  | TRACE           | TRACE         |
+| `console.debug`  | DEBUG           | DEBUG         |
+| `console.log`    | INFO            | INFO          |
+| `console.info`   | INFO            | INFO          |
+| `console.warn`   | WARN            | WARN          |
+| `console.error`  | ERROR           | ERROR         |
+| `console.dir`    | INFO            | INFO          |
+
+## Configuration
+
+| Option                  | Type                  | Default | Description                                                        |
+| ----------------------- | --------------------- | ------- | ------------------------------------------------------------------ |
+| `disableLogSending`     | `boolean`             | `false` | Disable sending log records to the OTel Logs SDK                   |
+| `disableLogCorrelation` | `boolean`             | `false` | Disable injecting trace context fields into log record attributes  |
+| `logSeverity`           | `SeverityNumber`      | —       | Minimum severity level; only logs at or above this level are sent  |
+| `logHook`               | `LogHookFunction`     | —       | Hook to inject additional fields into log records                  |
+
+### logHook Example
+
+```javascript
+new ConsoleInstrumentation({
+  logHook: (span, record) => {
+    record['custom.field'] = 'value';
+  },
+});
+```
+
+## Trace Context Correlation
+
+When a console method is called within an active span context, the resulting LogRecord will automatically include `trace_id`, `span_id`, and `trace_flags` attributes for log correlation.
+
+## License
+
+Apache 2.0 - See [LICENSE](./LICENSE) for more information.
+
+[npm-url]: https://www.npmjs.com/package/@opentelemetry/instrumentation-console
+[npm-img]: https://badge.fury.io/js/%40opentelemetry%2Finstrumentation-console.svg
+[license-image]: https://img.shields.io/badge/license-Apache_2.0-green.svg?style=flat
--- a/packages/instrumentation-console/README.md
+++ b/packages/instrumentation-console/README.md
@@ -67,6 +67,14 @@ new ConsoleInstrumentation({
 
 When a console method is called within an active span context, the resulting LogRecord will automatically include `trace_id`, `span_id`, and `trace_flags` attributes for log correlation.
 
+## Important: Infinite Loop Prevention
+
+This instrumentation requires `@opentelemetry/api >= 1.9.1`. Starting from that version, the `DiagConsoleLogger` saves references to the original `console` methods at module load time, so internal OTel diagnostic logging bypasses the instrumentation and avoids infinite loops.
+
+Additionally, this instrumentation includes a re-entrancy guard to prevent loops from exporters that write to the console (e.g., `ConsoleLogRecordExporter`).
+
+> **Note:** Avoid using `ConsoleLogRecordExporter` or `ConsoleSpanExporter` together with this instrumentation in production. Use network-based exporters (e.g., OTLP) instead.
+
 ## Useful links
 
 - For more information on OpenTelemetry, visit: <https://opentelemetry.io/>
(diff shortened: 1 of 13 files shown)


packages/instrumentation-console/test/console.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      console.debug('debug message');
+      const logRecords = memExporter.getFinishedLogRecords();
+      assert.strictEqual(logRecords.length, 1);
+      assert.strictEqual(logRecords[0].body, 'debug message');
+      assert.strictEqual(logRecords[0].severityNumber, SeverityNumber.DEBUG);
+      assert.strictEqual(logRecords[0].severityText, 'DEBUG');
+    });
+
+    it('emits LogRecord for console.trace', () => {
+      console.trace('trace message');
```
JacksonWeber, line 88: Looks like we're ensuring that `console.trace()` contains the message field here, but trace should also print a stack trace. We should ensure that the stack is also captured with this instrumentation.

JacksonWeber (approved):
Minor comment, but overall LGTM.

pichlermarc:
Might be worth syncing with the browser SIG, they also currently have a PR open, but they'll likely integrate it into their single instrumentation package:
https://github.com/open-telemetry/opentelemetry-browser/pull/98

hectorhdzg:
@pichlermarc  this is specific to Node.js console not browser, we should ensure there are no name conflicts and customer confusion, I can rename this package to @opentelemetry/instrumentation-console-nodejs or something like that, what you think?

packages/instrumentation-console/src/instrumentation.ts
```diff
@@ -0,0 +1,204 @@
...
+    super.disable();
+    this._unpatchConsole();
+  }
+
+  private _patchConsole(): void {
+    if (!this._originals) {
+      this._originals = new Map();
+    }
+    for (const method of Object.keys(CONSOLE_METHODS)) {
+      const original = (console as any)[method];
```
hectorhdzg, line 79: This is not using require/import in the middle, that would work if people import manually the console from their code, that is usually not the case and people rely on global console.

trentm:
> can rename this package to @opentelemetry/instrumentation-console-nodejs or something like that, what you think?

+1.   FWIW, we've used "-node" suffix in a couple cases:

```
% ls ~/src/opentelemetry-js-contrib/packages | rg node
auto-instrumentations-node
instrumentation-runtime-node
```

So perhaps `instrumentation-console-node`.

pichlermarc:
> > can rename this package to @opentelemetry/instrumentation-console-nodejs or something like that, what you think?
> 
> +1. FWIW, we've used "-node" suffix in a couple cases:
> 
> ```
> % ls ~/src/opentelemetry-js-contrib/packages | rg node
> auto-instrumentations-node
> instrumentation-runtime-node
> ```
> 
> So perhaps `instrumentation-console-node`.

`-node` suffix sounds good to me. 👍

JacksonWeber (approved):

packages/instrumentation-console/package.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "devDependencies": {
+    "@opentelemetry/api": "^1.9.1",
+    "@opentelemetry/sdk-logs": "^0.213.0",
+    "@opentelemetry/sdk-trace-base": "^2.0.0",
+    "@opentelemetry/sdk-trace-node": "^2.0.0"
+  },
+  "dependencies": {
+    "@opentelemetry/api-logs": "^0.213.0",
+    "@opentelemetry/instrumentation": "^0.213.0"
```
trentm:
```suggestion
    "@opentelemetry/sdk-logs": "^0.219.0",
    "@opentelemetry/sdk-trace-base": "^2.0.0",
    "@opentelemetry/sdk-trace-node": "^2.0.0"
  },
  "dependencies": {
    "@opentelemetry/api-logs": "^0.219.0",
    "@opentelemetry/instrumentation": "^0.219.0"
```

packages/instrumentation-console/src/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * A function that allows injecting additional fields in log records. It is
+   * called, as `logHook(span, record)`, for each log record emitted in a valid
+   * span context. It requires `disableLogCorrelation` to be false.
+   */
+  logHook?: LogHookFunction;
+}
+
+/** Console methods to instrument and their severity mappings. */
+export const CONSOLE_METHODS: Record<string, SeverityNumber> = {
```
trentm: nit: This and `CONSOLE_SEVERITY_TEXT` aren't *types* so would belong better in instrumentation.ts.

packages/instrumentation-console/src/types.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** Severity text labels for console methods. */
+export const CONSOLE_SEVERITY_TEXT: Record<string, string> = {
+  trace: 'TRACE',
+  debug: 'DEBUG',
+  log: 'INFO',
+  info: 'INFO',
+  warn: 'WARN',
+  error: 'ERROR',
+  dir: 'INFO',
```
trentm: nit: Could perhaps have severityText be *lower case*. FWIW `@opentelemetry/browser-instrumentation/console` is using lowercase names (it is using the console method name for `severityText`, but doesn't instrument `console.dir` so doesn't have an awkward `severityText === 'dir'` value).

packages/instrumentation-console/src/instrumentation.ts
```diff
@@ -0,0 +1,193 @@
...
+function formatArgs(args: unknown[]): string {
+  if (args.length === 0) return '';
+  return args
+    .map(arg => {
+      if (typeof arg === 'string') return arg;
+      try {
+        return JSON.stringify(arg);
+      } catch {
+        return String(arg);
+      }
```
trentm: This is effectively the same as in the browser console instr (https://github.com/open-telemetry/opentelemetry-browser/blob/8a873bc0441219081611f9de54d795a04c17a72b/packages/instrumentation/src/console/instrumentation.ts#L33), so all good.

However, comparing to Node.js' `console` methods, this formatting:
1. doesn't handle `%s`-codes in the first argument, e.g. `console.log('this is my num: %d', 42);`
2. doesn't do `util.inspect`-like representations

Would you want to consider supporting those? If so:
1. See https://github.com/nodejs/node/blob/f95bedd9634601fe8e16997f6f3479979139e0bb/lib/internal/util/inspect.js#L2526-L2650
2. See https://github.com/open-telemetry/opentelemetry-js/blob/d70e7dffa830ef54f6ccf2e9223d95db3ceae7b9/packages/sdk-trace/src/inspect.ts#L44-L71

Interestingly, the browser instr allows a custom `messageSerializer` function to be configured. I'm not suggesting that be added here. 

These could be done separately, if at all.

packages/instrumentation-console/src/instrumentation.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      const message = formatArgs(args);
+
+      const attributes: LogAttributes = {};
+
+      // Add trace context to attributes for log correlation
+      // The OTel SDK also auto-populates spanContext on the LogRecord
+      // from the active context, so these attributes are for additional
+      // correlation in the log output itself.
+      if (!config.disableLogCorrelation) {
```
trentm: I think this might be a misunderstanding. "Log correlation" is about adding `trace_id` and `span_id` fields *to the existing log output*. 

For example, for a Bunyan `log.info('hi there')`, "log correlation" would change *the Bunyan log output, e.g. to stdout by default*, from this:

```
{"name":"my-app","hostname":"peach.local","pid":60697,"level":30,"msg":"hi there","time":"2026-06-17T19:20:32.799Z","v":0}
```

to this:

```
{"name":"my-app","hostname":"peach.local","pid":60697,"level":30,"msg":"hi there","time":"2026-06-17T19:20:32.799Z","v":0,"trace_id":"...","span_id":"..."}
```

Log correlation is *not* about adding `trace_id` and `span_id` **attributes to the OTLP Log Record**.
The OTel Logs API has a `context` field for that.

---

What about this?

1. **no having log correlation** at all in this instrumentation, and
2. pass `context: context.active()` in the `otelLogger.emit(...)` call below.

packages/instrumentation-console/README.md
```diff
@@ -0,0 +1,83 @@
...
+| `console.info`   | INFO            | info          |
+| `console.warn`   | WARN            | warn          |
+| `console.error`  | ERROR           | error         |
+| `console.dir`    | INFO            | info          |
+
+## Configuration
+
+| Option                  | Type                  | Default | Description                                                        |
+| ----------------------- | --------------------- | ------- | ------------------------------------------------------------------ |
+| `disableLogSending`     | `boolean`             | `false` | Disable sending log records to the OTel Logs SDK                   |
```
trentm: Okay, excellent. This is much smaller now.

Next: let's *remove* the `disableLogSending`. Now that log-sending is the only thing this instrumentation does, one can disable log sending by disabling the instrumentation entirely, or not using it.
(The only reason `disableLogSending` and `disableLogCorrelation` were added to the other logging framework instrumentations was because they are doing the two independent things. A user wanting just one of those things needs to have those boolean config options.)

Sorry to not have caught this in my review yesterday.

packages/instrumentation-console/src/instrumentation.ts
hectorhdzg: added util.format, we can add more functionalities in following PRs if needed

JacksonWeber (approved):

pichlermarc (approved):

hectorhdzg:
Thanks for reviewing @trentm, @pichlermarc  and @JacksonWeber, I will go ahead and merge this one and we can address anything else in following PRs

PR #3454, opened 2026-03-26
feat(instrumentation-pg): propagate context using set application_name

mhennoch:
## Changes 
Adds SET application_name as a context propagation mechanism, following the pattern established for SQL Server ([SET CONTEXT_INFO](https://github.com/open-telemetry/semantic-conventions/pull/2363)) and Oracle ([V$SESSION.ACTION](https://github.com/open-telemetry/semantic-conventions/pull/2610)).

### Why SET application_name?
PostgreSQL instrumentation already supports SQL Commenter, but it has limitations:

- Prepared statements -- SQL Commenter cannot be used with named/prepared queries, as the comment is baked into the statement text and every subsequent execution carries stale trace context.
- Stored procedures -- SQL comments on the outer CALL statement are not visible to queries executed inside the procedure body.

SET application_name operates at the session level and avoids all of these issues.

There is ongoing [discussion in the npgsql community](https://github.com/npgsql/npgsql/issues/6255#issuecomment-3794199069) about this not being the ideal long-term solution, but no better alternative currently exists for PostgreSQL, and getting to one will take a long time. I will also create a semconv issue about it. This feature is disabled by default, as in the other instrumentations.

### Existing and related implementations

- Node.js (tedious/SQL Server): [opentelemetry-js-contrib#3141](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3141)
- Node.js (oracledb): [opentelemetry-js-contrib#3226](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3226)
- Java (PostgreSQL): [splunk-otel-java#2540](https://github.com/signalfx/splunk-otel-java/pull/2540)
- Collector: [opentelemetry-collector-contrib#44868](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44868)


maryliag:
I haven't checked the code yet, but can you add more tests? I can see that the codecov is failing, so this will need to increase the coverage

packages/instrumentation-pg/src/instrumentation.ts
```diff
@@ -466,6 +482,31 @@ export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConf
...
+              const setResult: unknown = original.apply(this, [
+                setQuery,
+              ] as never);
+              if (setResult instanceof Promise) {
+                setResult.catch(() => {});
+              }
+            } catch {
+              // Silently ignore SET failures to avoid breaking user queries
+            }
+          }
```
raphael-theriault-swi, line 515: I'd like there to be some diag warnings here on failure so it doesn't fail completely silently. Would make things a lot easier to debug for people if they're not getting correlation.

packages/sql-common/src/index.ts
raphael-theriault-swi: Nitpicking but this could be removed to use the newly added one at the module scope.

raphael-theriault-swi (changes requested):
LGTM apart from the silent failure. Sorry it took so long to get a review.

raphael-theriault-swi (approved):
Thank you !

PR #3459, opened 2026-03-30
fix(instr-user-interaction): avoid wrapping/unwrapping browser APIs multiple times

david-luna:

Unpatching history APIs may lead to breaking other instrumentations like `@opentelemetry/instrumentation-browser-navigation`. Ref: https://github.com/open-telemetry/opentelemetry-browser/issues/204

This was discussed in the Browser SIG and decided that instrumentations should rely on enabled state instead to create telemetry instead of disconnection (unpatching) of the browser APIs.

- instrumentation now patches only once and never unpatches APIs from browser and zone.js is available
- added `_isEnabled` private property to control the creation of spans from the instrumentation
- remove/updates tests that were just checking if patches are applied. Instead in it checks is spans are created or not.

### DIFF
--- a/packages/instrumentation-user-interaction/src/instrumentation.ts
+++ b/packages/instrumentation-user-interaction/src/instrumentation.ts
@@ -398,9 +398,12 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
    * Patches the certain history api method
    */
   _patchHistoryMethod() {
-  if (!plugin._isEnabled) { return original.apply(this, args); }
+    const plugin = this;
     return (original: any) => {
       return function patchHistoryMethod(this: History, ...args: unknown[]) {
+        if (!plugin._isEnabled) {
+          return original.apply(this, args);
+        }
         const url = `${location.pathname}${location.hash}${location.search}`;
         const result = original.apply(this, args);
         const urlAfter = `${location.pathname}${location.hash}${location.search}`;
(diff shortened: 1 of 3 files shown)


packages/instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -378,8 +386,6 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
    * Patches the history api
    */
   _patchHistoryApi() {
```
overbalance, line 388: Since this is public, I would add a guard in case it's called more than once:
```suggestion
  _patchHistoryApi() {
    if (!this._isEnabled) return;
```

packages/instrumentation-user-interaction/src/instrumentation.ts
overbalance: Also needs a guard for disabled state. As is, `_updateInteractionName` executes on all navigation. If disabled but already patched, this will corrupt an active span name from another instrumentation
```suggestion
  if (!plugin._isEnabled) { return original.apply(this, args); }
```

overbalance (changes requested):
I would consider halting `_patchZoneScheduleTask` and `_patchZoneCancelTask` with a guard when disabled too

david-luna:
@overbalance thanks for the feedback. I've applied your suggestions

overbalance (approved):

PR #3460, opened 2026-03-30
feat(detector-aws): detect Lambda availability zone metadata

garysassano:

- `@opentelemetry/resource-detector-aws` currently detects standard AWS Lambda `cloud` and `faas` attributes, but it does not populate `cloud.availability_zone`.
- AWS Lambda now exposes execution-environment AZ metadata through a local metadata endpoint, which makes this attribute available without requiring an external service call.
- The OpenTelemetry semantic conventions define `cloud.availability_zone`, so the Lambda detector can now populate it consistently with the other AWS resource detectors.

- Add opt-in Lambda metadata endpoint support to the AWS Lambda detector and map `AvailabilityZoneID` to `cloud.availability_zone` when enabled.
- Keep the default detector path environment-variable-only, preserving the previous cold-start behavior unless consumers opt in with `new AwsLambdaDetector({ fetchAvailabilityZone: true })`.
- Suppress internal tracing for the new localhost HTTP request, matching the package’s existing detector behavior.
- Add unit coverage for successful metadata lookup and failure fallback behavior.
- Extend the internal tracing integration test to exercise the Lambda metadata path.
- Document the new attribute in the package README.

## Reasoning

This change stays close to the new Lambda platform feature and the existing detector design. The detector already reads Lambda runtime metadata from environment variables when it is running inside Lambda. The Lambda-provided metadata endpoint is used only when `fetchAvailabilityZone` is enabled, keeping the previous environment-variable-only default while allowing consumers to opt into AZ enrichment.

I chose a direct HTTP call instead of using Powertools. That keeps the detector self-contained and avoids introducing a new runtime dependency on `@aws-lambda-powertools/commons`, or assuming that Powertools is already present through a Lambda layer. For a generic OpenTelemetry detector library, the direct endpoint call is the safer default.

The implementation stores the Lambda-provided AZ identifier as-is. Lambda returns an AZ ID such as `use1-az1`, not an AZ name such as `us-east-1a`. That value is still useful because AZ IDs are consistent across AWS accounts, while AZ names are not.

## Possible optimizations

- Powertools path: if this package ever accepts an explicit Powertools dependency or an optional integration layer, the detector could use the Powertools metadata utility instead of making its own HTTP request. That would delegate response caching and SnapStart-aware refresh behavior to AWS-maintained code.
- Optional translation to AZ name: if a consumer explicitly wants an AZ name instead of the Lambda AZ ID, a separate opt-in path could translate the ID through `DescribeAvailabilityZones`. I did not add that here because it would require additional IAM permissions, another API dependency, and would move the detector away from the value Lambda directly exposes.
- Shared metadata helper: if more Lambda-specific metadata fields are added later, the package could factor the HTTP request code into a reusable helper. Right now that would be premature because only one new field uses the endpoint.

## Testing

- Added unit tests for successful opt-in AZ metadata detection, disabled-by-default behavior, and graceful fallback when the metadata endpoint returns an error.
- Updated the internal tracing integration fixture to opt into the Lambda metadata path and verify it does not emit internal spans.
- Ran `npm run compile --workspace @opentelemetry/resource-detector-aws`.
- Ran `npm test --workspace @opentelemetry/resource-detector-aws` (46 passing).

## References

1. OpenTelemetry semantic conventions, `cloud.availability_zone`: <https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-availability-zone>
2. AWS Lambda metadata endpoint developer guide: <https://docs.aws.amazon.com/lambda/latest/dg/configuration-metadata-endpoint.html>
3. AWS Lambda availability zone metadata announcement, posted March 19, 2026: <https://aws.amazon.com/about-aws/whats-new/2026/03/lambda-availability-zone-metadata/>
4. Powertools for AWS Lambda (TypeScript) Metadata utility: <https://docs.aws.amazon.com/powertools/typescript/latest/features/metadata/>

### DIFF
--- a/packages/resource-detector-aws/README.md
+++ b/packages/resource-detector-aws/README.md
@@ -107,6 +107,7 @@ Populates `faas` and `cloud` for functions running on [AWS Lambda](https://aws.a
 
 | Resource Attribute | Description                                                         |
 |--------------------|---------------------------------------------------------------------|
+| cloud.availability_zone | Value of `AvailabilityZoneID` from the Lambda metadata endpoint (AZ ID, for example `use1-az1`) |
 | cloud.platform     | The cloud platform. In this context, it's always "aws_lambda"       |
 | cloud.provider     | The cloud provider. In this context, it's always "aws"              |
 | cloud.region       | Value of Process Environment Variable `AWS_REGION`                  |
--- a/packages/resource-detector-aws/README.md
+++ b/packages/resource-detector-aws/README.md
@@ -105,14 +105,29 @@ Populates `container` and `k8s.cluster_name` for containers running on [Amazon E
 Populates `faas` and `cloud` for functions running on [AWS Lambda](https://aws.amazon.com/lambda/).
 `faas.id` is currently not populated as it is not provided by the runtime at startup.
 
-| Resource Attribute | Description                                                         |
-|--------------------|---------------------------------------------------------------------|
-| cloud.availability_zone | Value of `AvailabilityZoneID` from the Lambda metadata endpoint (AZ ID, for example `use1-az1`) |
-| cloud.platform     | The cloud platform. In this context, it's always "aws_lambda"       |
-| cloud.provider     | The cloud provider. In this context, it's always "aws"              |
-| cloud.region       | Value of Process Environment Variable `AWS_REGION`                  |
-| faas.name          | Value of Process Environment Variable `AWS_LAMBDA_FUNCTION_NAME`    |
-| faas.version       | Value of Process Environment Variable `AWS_LAMBDA_FUNCTION_VERSION` |
+| Resource Attribute      | Description                                                                |
+|-------------------------|----------------------------------------------------------------------------|
+| cloud.availability_zone | Value of `AvailabilityZoneID` when `fetchAvailabilityZone` is enabled      |
+| cloud.platform          | The cloud platform. In this context, it's always "aws_lambda"              |
+| cloud.provider          | The cloud provider. In this context, it's always "aws"                     |
+| cloud.region            | Value of Process Environment Variable `AWS_REGION`                         |
+| faas.name               | Value of Process Environment Variable `AWS_LAMBDA_FUNCTION_NAME`           |
+| faas.version            | Value of Process Environment Variable `AWS_LAMBDA_FUNCTION_VERSION`        |
+
+#### Availability zone (opt-in)
+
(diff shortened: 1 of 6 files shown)


jj22ee (commented):
Hey @garysassano, I was interested in the cold-start impact since this detector previously only read environment variables and this adds an HTTP request on the init path.

I compared the current detector vs. this PR, on average (10 runs) the one-time cold-start cost was `+53 ms` at `128 MB` and `+5.5 ms` at `512 MB` of memory. So the performance impact seems okay, wondering if you had similar results or other thoughts?

Regarding populating `cloud.availability_zone` with the AZ ID instead of the AZ name, it technically still fits the spec, but it will be inconsistent with the other AWS EC2 and ECS detectors in this package, which set the value to be the AZ name instead. So I suppose if consumers tries to ever query resources grouped by `cloud.availability_zone`, they'd need to account for this discrepancy. Given the limitation of Lambda only exposing the ID, I think this is okay for now, as you also mentioned in the description. I'm thinking that if the AZ name is ever exposed in the future (if that even makes sense for Lambda), we should switch to it as a breaking change. For now, could you leave a comment under `_fetchAvailabilityZone along` these lines?
```
// Lambda's metadata endpoint only exposes the AZ ID (e.g. `use1-az1`), not the AZ name.
// This is recorded best-effort. Other AWS resource detectors (EC2 and ECS) in this package
// populate `cloud.availability_zone` with the AZ name (e.g. `us-east-1a`), so this value
// is not directly comparable across detectors. Translating to a name would require
// `ec2:DescribeAvailabilityZones` and is intentionally not done here.
```

garysassano:
Thanks for checking this and for the detailed numbers.

To be frank, I do not remember running a comparable cold-start benchmark when I first opened this, so your measurements are useful. The `+53 ms` at 128 MB is small, but this detector previously had an environment-variable-only init path, so I agree it is better not to add the metadata HTTP request by default for every Lambda user.

I pushed an update that makes the availability-zone lookup opt-in via:

```ts
new AwsLambdaDetector({ fetchAvailabilityZone: true })
```

The default `awsLambdaDetector` now keeps the previous behavior and only reads environment variables. Consumers that want `cloud.availability_zone` can opt into the one-time metadata endpoint request explicitly.

I also added the comment you suggested under `_fetchAvailabilityZone`. I agree with the AZ ID vs. AZ name concern: the value still fits the semantic convention, but it is not directly comparable with the EC2/ECS detector values today. If Lambda ever exposes the AZ name directly, switching to that would make sense as a breaking change. I intentionally did not add an AZ ID to AZ name translation path because that would require `ec2:DescribeAvailabilityZones`, extra IAM permissions, and an AWS API dependency.

jj22ee (approved):
lgtm!

PR #3470, opened 2026-04-08
fix(instr-browser-navigation): do not unpatch history API

david-luna:

Follow up of https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3459. This time is for browser navigation instrumentation.

- instrumentation now patches only once and never unpatches APIs from browser
- add/remove event listeners is kept
- add `_isHistoryPatched` private property to avoid multiple patching

### DIFF
--- a/packages/instrumentation-browser-navigation/src/instrumentation.ts
+++ b/packages/instrumentation-browser-navigation/src/instrumentation.ts
@@ -172,7 +172,7 @@ export class BrowserNavigationInstrumentation extends InstrumentationBase<Browse
       ((window as any).navigation as EventTarget);
 
     // Only patch history API if Navigation API is not available
-    if (!navigationApi && ! this._isHistoryPatched) {
+    if (!navigationApi && !this._isHistoryPatched) {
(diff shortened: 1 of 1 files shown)


packages/instrumentation-browser-navigation/src/instrumentation.ts
```diff
@@ -171,8 +172,9 @@ export class BrowserNavigationInstrumentation extends InstrumentationBase<Browse
       ((window as any).navigation as EventTarget);
 
     // Only patch history API if Navigation API is not available
-    if (!navigationApi) {
+    if (!navigationApi && ! this._isHistoryPatched) {
```
overbalance: Does this work with a space?

overbalance (approved):

packages/instrumentation-browser-navigation/src/instrumentation.ts
david-luna: it's odd `npm run lint` passed locally but not in CI. I'll check my env.

Issue #3469, opened 2026-04-08
Improve sanitization for Redis ACL and CONFIG commands


mznet:
### What version of OpenTelemetry are you using?
@opentelemetry/redis-common 0.38.2

### What version of Node are you using?
v25.1.0

### What did you do?
Used Redis commands that include sensitive information as arguments:

```
ACL SETUSER alice on >MySecretPass ~user:alice:* +@read +@write
CONFIG SET requirepass MyNewPassword123
```

I expected these commands to be sanitized with redis command sanitizer, but it's not working as i expected.

### What did you expect to see?
                                                                                                                                                           
ACL:                                                                                                                                                     
```
ACL SETUSER [3 other arguments]
```
                                                                                                                                                          
CONFIG:
```
CONFIG SET [2 other arguments]
```

### What did you see instead?

ACL:                                                                                                                                                     
```
ACL SETUSER alice on >MySecretPass ~user:alice:* +@read +@write
```
                                                                                                                                                           
CONFIG:                                                                                                                                                  
```
CONFIG SET requirepass MyNewPassword123
```

### Additional context

I think it would work if moving ACL and CONFIG from the args: -1 group to args: 1 in the serializationSubsets definition in `packages/redis-common/src/index.ts`.  

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. [Learn more here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

pichlermarc:
Hi @mznet - thanks for reporting this. You're right, this should not be included in the reported telemetry like that. I'm working on a fix.

In case you ever run into security-relevant issues in the future, please report them through the [Security Tab](https://github.com/open-telemetry/opentelemetry-js-contrib/security) in accordance with the project's [Security Policy](https://github.com/open-telemetry/opentelemetry-js-contrib?tab=security-ov-file#security-policy). 🙂 This ensures the problem gets the attention it deserves (jumping ahead of the normal bug triage process) and provides maintainers with the tools to coordinate disclosure.

JamieDanielson:
@pichlermarc  Can we close as completed with your fixed in #3472 ?

pichlermarc:
yes, that's already done. :)

PR #3472, opened 2026-04-09
fix(redis-common): expand redaction to include ACL, CONFIG, PSETEX, GETSET

pichlermarc:

We don't sanitize quite some commands that can contain sensitive data, as described in #3469.
In addition to `ACL` and `CONFIG`, there's also `PSETEX` (set something for some time) and `GETSET` (atomic get + set) that we should redact in the same way as `SET`

- add `ACL`, `CONFIG`, `PSETEX` and `GETSET` to the list of redacted things (all should only include the first arg)

### DIFF
--- a/packages/instrumentation-redis/test/v2-v3/redis.test.ts
+++ b/packages/instrumentation-redis/test/v2-v3/redis.test.ts
@@ -335,6 +335,80 @@ describe('redis v2-v3', () => {
       });
     });
 
+    describe('sensitive command sanitization', () => {
+      it('should redact CONFIG SET arguments in db.statement', function (done) {
+        const span = tracer.startSpan('test span');
+        context.with(trace.setSpan(context.active(), span), () => {
+          client.config('set', 'hz', '15', (err: unknown) => {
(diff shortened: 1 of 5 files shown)


packages/instrumentation-redis/test/v2-v3/redis.test.ts
```diff
@@ -335,6 +335,80 @@ describe('redis v2-v3', () => {
       });
     });
 
+    describe('sensitive command sanitization', () => {
```
maryliag, line 338: I see you aded "should redact ACL SETUSER arguments in db.statement" to the other files, but no to this one, does v2 and v3 don't have that command?

maryliag (approved):
Just small question, otherwise lgtm

packages/instrumentation-redis/test/v2-v3/redis.test.ts
pichlermarc: Yep, exactly that. As far as I could tell it's not there in these versions yet.

PR #3474, opened 2026-04-13
chore: update codecov script

maryliag:
`commitSha` and `branchName` are read from environment variables and joined into a shell command string passed to `execSync`. If these contain shell metacharacters (e.g. ; rm -rf /), they execute arbitrary code. This runs in CI but is still bad practice.

Fix: pass args as an array via `spawnSync` instead


david-luna:
In https://github.com/open-telemetry/opentelemetry-js/pull/6586 you changed from `execSync` to `execFileSync` bun in here you're switching to `spawnSync`. Any particular reason?

maryliag:
No particular reason, I was trying out different things, they work very similar, but to keep consistency, I'm changing this one to also use `execFileSync`

david-luna (approved):

PR #3484, opened 2026-04-21
feat(auto-instrumentations-node): add opt-in host metrics via OTEL_NODE_HOST_METRICS_ENABLED

### LINKED ISSUE #3483: feat(auto-instrumentations-node): add opt-in host metrics via OTEL_NODE_HOST_METRICS_ENABLED
## Summary

When using `@opentelemetry/auto-instrumentations-node` in the OpenTelemetry Operator's Node.js autoinstrumentation image, host-level metrics (CPU, memory, network) are not available without manually adding a separate init container. This is a common enough need that it should be supported opt-in directly from the `register` entry point.

## Proposed change

Add an opt-in code path in `src/register.ts` that starts `HostMetrics` when `OTEL_NODE_HOST_METRICS_ENABLED=true` is set. The feature is off by default so there is no behaviour change for existing users.

```shell
export OTEL_NODE_HOST_METRICS_ENABLED="true"
node --require @opentelemetry/auto-instrumentations-node/register app.js
```

`@opentelemetry/host-metrics` is added as a runtime dependency of `@opentelemetry/auto-instrumentations-node`. `HostMetrics` calls `sdk.start()` first and then picks up the global `MeterProvider` automatically — no explicit provider wiring needed (confirmed by #3239).

## Motivation

- The OpenTelemetry Operator's [`autoinstrumentation-nodejs`](https://github.com/open-telemetry/opentelemetry-operator/tree/main/autoinstrumentation/nodejs) image uses this package as its `--require` script. Adding host metrics there today requires a second init container.
- Already validated end-to-end on a GKE cluster: `system.cpu.time`, `system.memory.usage`, `process.cpu.time`, and network metrics all flowing through Alloy to Grafana Cloud.

bradleypettit:
## Summary

Adds opt-in host metrics (CPU, memory, network) to the `register` entry point,
enabled by setting `OTEL_NODE_HOST_METRICS_ENABLED=true`.

- `@opentelemetry/host-metrics` is added as a runtime dependency
- The env var is checked at startup; if unset (the default) nothing changes
- `HostMetrics` is started after `sdk.start()` so it picks up the global
  MeterProvider automatically — no explicit wiring needed

## Motivation

The OpenTelemetry Operator's `autoinstrumentation-nodejs` image uses this
package as its `--require` script. Getting host metrics today requires
manually adding a second init container. This makes it available opt-in
with a single env var.

Validated end-to-end on GKE: `system.cpu.time`, `system.memory.usage`,
`process.cpu.time`, and network metrics all flowing correctly.

Closes #3483

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36310,6 +36310,7 @@
       "version": "0.73.0",
       "license": "Apache-2.0",
       "dependencies": {
+        "@opentelemetry/host-metrics": "^0.38.0",
         "@opentelemetry/instrumentation": "^0.215.0",
         "@opentelemetry/instrumentation-amqplib": "^0.62.0",
         "@opentelemetry/instrumentation-aws-lambda": "^0.67.0",
--- a/packages/auto-instrumentations-node/README.md
+++ b/packages/auto-instrumentations-node/README.md
@@ -124,6 +124,16 @@ Notes:
 - Logs are always sent to console, no matter the environment, or debug level.
 - Debug logs are extremely verbose. Enable debug logging only when needed. Debug logging negatively impacts the performance of your application.
 
+### Host Metrics
+
+Host metrics (CPU, memory, network) are not collected by default. To enable them, set `OTEL_NODE_HOST_METRICS_ENABLED=true`:
+
+```shell
+export OTEL_NODE_HOST_METRICS_ENABLED="true"
+```
+
+This uses [`@opentelemetry/host-metrics`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/host-metrics) and automatically exports to the same metrics pipeline configured for your application.
+
 ## Usage: Instrumentation Initialization
 
 OpenTelemetry Meta Packages for Node automatically loads instrumentations for Node builtin modules and common packages.
--- a/packages/auto-instrumentations-node/package.json
+++ b/packages/auto-instrumentations-node/package.json
@@ -43,6 +43,7 @@
     "@opentelemetry/core": "^2.0.0"
   },
   "dependencies": {
+    "@opentelemetry/host-metrics": "^0.38.0",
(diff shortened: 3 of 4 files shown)


maryliag:
Hey @bradleypettit, we're not creating new environment variables anymore, but there is an existing one that makes sense for you to use in this case: `OTEL_NODE_ENABLED_INSTRUMENTATIONS`.

What you can do to make it work:
- remove the changes you have here in the register file
- add the import of host metrics [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/auto-instrumentations-node/src/utils.ts#L24)
- add it to the mapping [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/auto-instrumentations-node/src/utils.ts#L102)
- update the readme accordingly 

And that should be enough to have those metrics generated by default or by passing host in the list of `OTEL_NODE_ENABLED_INSTRUMENTATIONS`.

I would like to get the take from maintainers if that one should be off by default, and if so, then you would need to add it to the list [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/auto-instrumentations-node/src/utils.ts#L129)

trentm:
> What you can do to make it work:

I think host-metrics would need to be implemented as an Instrumentation to be used in that way in auto-instrumentations-node. Currently it isn't implemented as an Instrumentation, though I think it should be: the package is effectively an "instrumentation" in that it is a separate component that uses the OTel API to generate telemetry.

I've added an agenda item to tomorrow's OTel JS SIG call to discuss.

trentm:
I moved my follow-up to the issue: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3483#issuecomment-4331071520

bradleypettit:
Thanks @trentm - I think best I close this PR for now, and I'll open another PR to address the comments you left in the issue.

PR #3488, opened 2026-04-26
fix(instrumentation-undici): do not record aborted requests as errors

### LINKED ISSUE #2482: instrumentation-undici records exception when calling Response.body.cancel.
## Steps to Reproduce
```ts
const response = await fetch('https://www.google.com', { method: 'get' });
await response.body?.cancel('HELLO WORLD');
```
## Expected Result

Here the response has a 200 status code. I do not intend to use the response body and since I am using Undici which comes  packaged with Node 20, I explicitly cancel the response body stream as directed by the Undici [docs](https://undici.nodejs.org/#/?id=garbage-collection):

>  The [Fetch Standard](https://fetch.spec.whatwg.org/) allows users to skip consuming the response body by relying on [garbage collection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#garbage_collection) to release connection resources. Undici does not do the same. Therefore, it is important to always either consume or cancel the response body.

Because this cancellation is an intentional performance optimization that is promoted by the undici maintainers and is not meant to indicate some sort of issue with the request, I expected the generated span status to be OK.

## Actual Result

The span status is set to ERROR and an exception is recorded. See the relevant log output below for details.

### OpenTelemetry Setup Code

```JavaScript
import { UndiciInstrumentation } from '@opentelemetry/instrumentation-undici';

import { NodeSDK } from '@opentelemetry/sdk-node';
import {
  ConsoleSpanExporter,
  SimpleSpanProcessor,
} from '@opentelemetry/sdk-trace-base';

const exporter = new ConsoleSpanExporter();
const sdk = new NodeSDK({
  autoDetectResources: true,
  traceExporter: exporter,
  spanProcessors: [new SimpleSpanProcessor(exporter)],
  instrumentations: [
    new UndiciInstrumentation(),
  ],
});

sdk.start();

process.on('SIGTERM', () => {
  sdk
    .shutdown()
    .then(() => console.log('Tracing terminated'))
    .catch((error) => console.log('Error terminating tracing', error))
    .finally(() => process.exit(0));
});
```

### package.json

```JSON
{
  "dependencies": {
    "@opentelemetry/instrumentation-undici": "0.6.0",
    "@opentelemetry/sdk-node": "^0.53.0",
    "undici": "^6.20.0"
  },
  "devDependencies": {
    "typescript": "^5.3.2"
  }
}
```

### Relevant log output

```shell
{
  resource: {
    attributes: {
      'service.name': 'unknown_service:node',
      'telemetry.sdk.language': 'nodejs',
      'telemetry.sdk.name': 'opentelemetry',
      'telemetry.sdk.version': '1.26.0',
      'process.pid': 65973,
      'process.executable.name': 'node',
      'process.runtime.version': '20.14.0',
      'process.runtime.name': 'nodejs',
      'process.runtime.description': 'Node.js',
       ...
      'host.arch': 'arm64'
    }
  },
  instrumentationScope: {
    name: '@opentelemetry/instrumentation-undici',
    version: '0.6.0',
    schemaUrl: undefined
  },
  traceId: 'f4372db8a37858d737a008c167d579e6',
  parentId: undefined,
  traceState: undefined,
  name: 'GET',
  id: '14bd8da53cfd3848',
  kind: 2,
  timestamp: 1728675289178000,
  duration: 335598.666,
  attributes: {
    'http.request.method': 'GET',
    'http.request.method_original': 'GET',
    'url.full': 'https://www.google.com/',
    'url.path': '/',
    'url.query': '',
    'url.scheme': 'https',
    'server.address': 'www.google.com',
    'server.port': 443,
    'user_agent.original': 'undici',
    'network.peer.address': '142.250.65.228',
    'network.peer.port': 443,
    'http.response.status_code': 200
  },
  status: { code: 2, message: undefined },
  events: [
    {
      name: 'exception',
      attributes: { 'exception.message': 'HELLO WORLD' },
      time: [ 1728675289, 513519708 ],
      droppedAttributesCount: 0
    }
  ],
  links: []
}
```

### LINKED ISSUE #3480: opentelemetry-instrumentation-fetch records aborted fetches as errors
## Steps to Reproduce
  

I'm seeing in my traces exceptions being reported for fetch calls, however, these are happening because we're passing an signal to the fetch and deliberately aborting it.

The exception recorded is:
```
{
  "exception.message": "The operation was aborted.",
  "exception.stacktrace": "AbortError: The operation was aborted.\n    at new DOMException (node:internal/per_context/domexception:76:18)\n    at Object.onConnect (node:internal/deps/undici/undici:13633:23)\n    at Request.onConnect (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/core/request.js:231:29)\n    at writeH1 (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/dispatcher/client-h1.js:931:13)\n    at Object.write (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/dispatcher/client-h1.js:757:14)\n    at _resume (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/dispatcher/client.js:614:50)\n    at resume (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/dispatcher/client.js:536:3)\n    at Client.<computed> (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/dispatcher/client.js:261:31)\n    at connect (/app/node_modules/.pnpm/undici@6.24.1/node_modules/undici/lib/dispatcher/client.js:521:18)",
  "exception.type": "20",
  "monocle.exception.fingerprint": "ce7bcd8725d9",
  "monocle.exception.normalized_message": "The operation was aborted."
}
```

## Expected Result

Aborted fetches do not appear as exceptions in traces, since usually this is the case of something deliberately aborting the fetch (a timeout, or a race between two requests)

## Actual Result

Aborted fetches result in exceptions in traces.

## Additional Details
 
I'm using the otel instrumentation via automatic instrumentation, using the `@adonisjs/otel` package via `@monocle.sh/adonisjs-agent`.

The reason we're aborting fetches is because we do a race between a DNS resolve and a fetch call, if DNS resolves first, we abort the fetch, since we don't need it's results (this is how handle resolution in AT Protocol works)

### OpenTelemetry Setup Code

```JavaScript

```

### package.json

```JSON

```

### Relevant log output

```shell

```

### Operating System and Version

Debian Trixie

### Runtime and Version

Node.js 24

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. [Learn more here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Renegade2345:

Fixes #3480
Fixes #2482

Aborted fetch/undici requests were being recorded as span errors with
`SpanStatusCode.ERROR`, an exception event, and `error.type` set. This
happens when callers deliberately cancel requests via `AbortController.signal`
(e.g. racing two requests and cancelling the loser, or implementing timeouts).

Per the [OTel HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client):

> If the HTTP client instrumentation can detect that the request was cancelled
> intentionally by the caller, the cancellation SHOULD NOT be treated as an
> error: the span status SHOULD be left unset and `error.type` SHOULD NOT be set.

In `onError()`, abort errors are now detected by checking:
- `error.name === 'AbortError'`, or
- `error instanceof DOMException && error.code === DOMException.ABORT_ERR`
  (handles undici v6.3.0+ which switched to built-in `DOMException`)

For aborts: span ends with status UNSET, no exception recorded, no `error.type`.
For all other errors: existing behaviour is preserved.

Tests updated in both `fetch.test.ts` and `undici.test.ts`.

### DIFF
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -447,7 +447,6 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     if (!record) {
       return;
     }
-
     const { span, attributes, startTime } = record;
 
     // End the span
@@ -473,22 +472,30 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
 
     const { span, attributes, startTime } = record;
 
-    // NOTE: in `undici@6.3.0` when request aborted the error type changes from
-    // a custom error (`RequestAbortedError`) to a built-in `DOMException` carrying
-    // some differences:
-    // - `code` is from DOMEXception (ABORT_ERR: 20)
-    // - `message` changes
-    // - stacktrace is smaller and contains node internal frames
-    span.recordException(error);
-    span.setStatus({
-      code: SpanStatusCode.ERROR,
-      message: error.message,
-    });
-    span.end();
-    this._recordFromReq.delete(request);
+    // Per OTel HTTP spec: if the request was intentionally cancelled via an
+    // AbortController signal, it SHOULD NOT be treated as an error.
+    // Span status should be left unset and error.type should not be set.
+    // https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client
+    const isAbort =
+      error.name === 'AbortError' ||
+      (typeof DOMException !== 'undefined' &&
+        error instanceof DOMException &&
+        error.code === DOMException.ABORT_ERR);
 
-    // Record metrics (with the error)
-    attributes[ATTR_ERROR_TYPE] = error.message;
+    if (isAbort) {
+      span.end();
+    } else {
+      span.recordException(error);
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: error.message,
+      });
+      span.end();
+      // Record metrics (with the error)
+      attributes[ATTR_ERROR_TYPE] = error.message;
+    }
+
+    this._recordFromReq.delete(request);
     this.recordRequestDuration(attributes, startTime);
   }
 
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -491,7 +491,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
         message: error.message,
       });
       span.end();
-      // Record metrics (with the error)
+      
       attributes[ATTR_ERROR_TYPE] = error.message;
     }
 
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -491,7 +491,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
         message: error.message,
       });
       span.end();
-      
+
       attributes[ATTR_ERROR_TYPE] = error.message;
     }
 
--- a/packages/instrumentation-undici/test/undici.test.ts
+++ b/packages/instrumentation-undici/test/undici.test.ts
@@ -789,7 +789,6 @@ describe('UndiciInstrumentation `undici` tests', function () {
       assert.ok(span, 'a span is present');
       assert.strictEqual(spans.length, 1);
 
-      
       assert.strictEqual(
         span.status.code,
         SpanStatusCode.UNSET,
--- a/packages/instrumentation-undici/test/undici.test.ts
+++ b/packages/instrumentation-undici/test/undici.test.ts
@@ -789,7 +789,7 @@ describe('UndiciInstrumentation `undici` tests', function () {
       assert.ok(span, 'a span is present');
       assert.strictEqual(spans.length, 1);
 
-      // 🚨 Core fix validation
+      
       assert.strictEqual(
         span.status.code,
         SpanStatusCode.UNSET,
--- a/packages/instrumentation-undici/test/undici.test.ts
+++ b/packages/instrumentation-undici/test/undici.test.ts
@@ -762,52 +762,52 @@ describe('UndiciInstrumentation `undici` tests', function () {
     });
 
     it('should not record abort as an error if undici request is aborted', async function () {
-  let spans = memoryExporter.getFinishedSpans();
-  assert.strictEqual(spans.length, 0);
+      let spans = memoryExporter.getFinishedSpans();
+      assert.strictEqual(spans.length, 0);
 
-  const controller = new AbortController();
+      const controller = new AbortController();
 
-  const requestPromise = undici.request(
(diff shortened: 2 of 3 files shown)


packages/instrumentation-undici/src/undici.ts
```diff
@@ -473,22 +472,30 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
+    if (isAbort) {
+      span.end();
+    } else {
+      span.recordException(error);
+      span.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: error.message,
+      });
+      span.end();
+      // Record metrics (with the error)
```
trentm:
```suggestion
```

nit: I think we can drop this comment. It made a little more sense when it was for the block with this attributes update and `this.recordRequestDuration`. Now it could be slightly misleading.

packages/instrumentation-undici/test/undici.test.ts
```diff
@@ -761,52 +761,53 @@ describe('UndiciInstrumentation `undici` tests', function () {
...
-        error: requestError,
-        noNetPeer: true, // do not check network attribs
-        forceStatus: {
-          code: SpanStatusCode.ERROR,
-          message: requestError?.message,
-        },
-      });
-    });
 
+      // 🚨 Core fix validation
```
trentm:
```suggestion
```

nit: Does this comment add context?

trentm (approved):
Thanks! This looks great.
I have a could of nits, but they are just nits.

david-luna:
@Renegade2345 thanks for your contribution :)

The semantic conventions were updated recently regarding the cancelation of request https://github.com/open-telemetry/semantic-conventions/pull/3494 so this PR will align with the most recent version. Good

I think this PR also will close https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2482 where instead of using and `AbortController` the user was cancelling the body.

Could you add another test case where `response.body.cancel()` is used?

Thanks!

david-luna (approved):

david-luna:
I think we should follow up with a similar issue & PR for http instrumentation.

Renegade2345:
> I think we should follow up with a similar issue & PR for http instrumentation.

Sure! @david-luna  Meanwhile I'll fix the issue here and get it done.
Thanks for taking your time out!   @david-luna and @trentm

packages/instrumentation-undici/test/undici.test.ts
Renegade2345: I have made the changes. Thank you for suggesting. @trentm

PR #3490, opened 2026-04-27
fix(ci): run tests with services env

mhennoch:
The `test:ci:affected` and `test-all-versions:ci:affected` scripts ran without sourcing `test/test-services.env`, so `RUN_*_TESTS` and DB connection vars were unset in CI.


package.json
```diff
@@ -16,11 +16,11 @@
     "compile:ci:affected": "nx affected -t compile",
     "list:affected": "nx affected -t test-all-versions --graph=stdout | jq -c '.tasks.tasks[].target.project'",
     "test": "nx run-many -t test",
-    "test:ci:affected": "nx affected -t test",
+    "test:ci:affected": "cross-env NODE_OPTIONS='-r dotenv/config' DOTENV_CONFIG_PATH=./test/test-services.env nx affected -t test",
     "test:browser": "nx run-many -t test:browser",
     "test:browser:ci:affected": "nx affected -t test:browser",
     "test-all-versions": "npm run --if-present test-all-versions --workspaces",
-    "test-all-versions:ci:affected": "npm --loglevel=silent run list:affected | xargs -I {} -t npm run --if-present test-all-versions -w {}",
+    "test-all-versions:ci:affected": "cross-env NODE_OPTIONS='-r dotenv/config' DOTENV_CONFIG_PATH=./test/test-services.env sh -c 'npm --loglevel=silent run list:affected | xargs -I {} -t npm run --if-present test-all-versions -w {}'",
```
trentm, line 23: Unfortunate that it needs to be this way instead of `s/test-all-versions/test-all-versions:with-services-env/g`. Doing the latter won't work because most instrumentations don't have a `test-all-versions:with-services-env`.   (We should improve all this, but not for this PR.)

packages/instrumentation-pg/test/pg.test.ts
trentm, line 1: @mhennoch Did you intend to have instr-pg changes in the final version of this? I'm guessing you initially did to sanity check that the CI run your changes as expected.

mhennoch: This is unrelated to the pg change this just fails if you actually run the tests.

trentm (approved):

Issue #3458, opened 2026-03-27
Memory leak - possible regression


lamuertepeluda:
### What version of OpenTelemetry are you using?

@opentelemetry/auto-instrumentations-node: 0.72.0

### What version of Node are you using?

Node 24.14.1

### What did you do?

Launch my NestJS application with

```bash
export NODE_OPTIONS="--enable-source-maps --require @opentelemetry/auto-instrumentations-node/register"
node dist/src/main
```

### What did you expect to see?

No errors

### What did you see instead?

```
(node:7) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 finish listeners added to [ServerResponse]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
```

repeated a lot of times

### Additional context

Looks like https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2479 and https://github.com/open-telemetry/opentelemetry-js-contrib/pull/278

Probably a regression of it

I can't tell for sure if downgrading to a previous version would solve the issue

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. [Learn more here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

trentm:
@lamuertepeluda Are you able to provide some more details:

1. Can you show a full stack trace? For example see the full stack trace a similar MaxListenersExceededWarning on this old issue https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2479   The stacktrace might help track down where the issue is.
2. Are you able to provide a small reproduction repo? That would help anyone working on this. Thanks.

dhaakchg:
@opentelemetry/auto-instrumentations-node: 0.72.0
Node: 24.14.0
Express: 5.2.1

Downgrading to 0.71.0 stops the repeated error output for me with Express 5.2.1.
Notably this doesn't seem to be happening with express 4.22.1

```
(node:1) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 finish listeners added to [ServerResponse]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
    at genericNodeError (node:internal/errors:985:15)
    at wrappedFn (node:internal/errors:539:14)
    at _addListener (node:events:582:17)
[...]

PR #3495, opened 2026-04-29
fix(instrumentation-router): fix MaxListenersExceededWarning

mhennoch:
Fixes router instrumentation span cleanup to avoid MaxListenersExceededWarning for the common case where many synchronous middleware call next() in a single request. #3458 

The instrumentation now delays attaching the response close fallback listener until after the router layer returns, and skips attaching it entirely when synchronous next() already ended the span. If a layer is async or ends the response itself, the fallback still ensures the span is closed.

Note: requests with a large number of concurrently pending async middleware can still temporarily attach many response close listeners before those middleware call next() but this shouldn't be a common case. Also this somewhat matches what express instrumentation is doing.

### DIFF
--- a/packages/instrumentation-router/src/instrumentation.ts
+++ b/packages/instrumentation-router/src/instrumentation.ts
@@ -108,13 +108,13 @@ export class RouterInstrumentation extends InstrumentationBase {
       ) {
         return original.call(this, req, res, next);
       }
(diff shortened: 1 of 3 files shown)


seemk (approved):
Tested it locally, didn't see any issues.

pichlermarc (approved):
thanks!

PR #3505, opened 2026-05-05
feat(auto-instrumentations-node): opt-in crash flush via OTEL_NODE_FLUSH_ON_CRASH

ElfoLiNk:
## Summary

- Adds opt-in handlers in `packages/auto-instrumentations-node/src/register.ts` that flush the SDK on `uncaughtException`, `unhandledRejection`, and `SIGINT` before exiting.
- Gated behind `OTEL_NODE_FLUSH_ON_CRASH=true`; default behavior is unchanged.
- Optional `OTEL_NODE_FLUSH_ON_CRASH_TIMEOUT_MS` (default 5000ms) bounds shutdown so a wedged exporter cannot keep the process alive.

## Motivation

`register.ts` currently wires `SIGTERM` and `beforeExit` to `sdk.shutdown()`. Per the [Node.js docs](https://nodejs.org/api/process.html#event-beforeexit), `beforeExit` does **not** fire for `process.exit()`, uncaught exceptions, or unhandled promise rejections. With the default `BatchLogRecordProcessor` / `BatchSpanProcessor` (queue-and-flush every N ms), any records still in the buffer when the process is killed by a crash are dropped — this is the well-known "logs are in stdout but never reach the collector" symptom around crash timestamps.

In a real incident this looked like:

- An uncaught `KafkaJSProtocolError` (TOPIC_AUTHORIZATION_FAILED) propagated from a producer.
- The error log emitted ~milliseconds before termination was visible in stdout but never arrived in the OTEL backend.
- `BatchLogRecordProcessor.scheduledDelayMillis` (default 1000ms) is the worst-case loss window.

## Behavior

When `OTEL_NODE_FLUSH_ON_CRASH=true`:

1. The original error is written to stderr (preserving the existing crash output for humans/log scrapers).
2. `sdk.shutdown()` is raced against a bounded timer; whichever finishes first wins.
3. The process exits with the canonical Node status code (`1` for crashes, `130` for `SIGINT` per `128 + signum`) so shells, supervisors, and CI see the same status they would have without the handler.
4. A re-entrancy guard protects against `shutdown()` itself throwing.

The flag is opt-in because registering an `uncaughtException` handler suppresses Node's default termination — users should make that trade-off consciously, and existing applications that rely on the current behavior are unaffected.

## Test plan

- [ ] `npm run compile` in `packages/auto-instrumentations-node` (TypeScript)
- [ ] Existing `register.test.ts` cases still pass (no behavior change with the flag unset)
- [ ] Manual: run a script that throws after emitting a log; with `OTEL_NODE_FLUSH_ON_CRASH=true` and a console exporter, confirm the trailing record is exported before exit
- [ ] Manual: confirm process exits with 1 on uncaught exception, 130 on SIGINT
- [ ] Manual: with a deliberately broken exporter endpoint, confirm the timeout fires and the process still exits within `OTEL_NODE_FLUSH_ON_CRASH_TIMEOUT_MS`

Happy to add an automated test for the crash-flush path if maintainers want it — it would mirror the existing SIGTERM test in `register.test.ts` (spawn child, send signal / cause throw, assert shutdown message in stdout).

### DIFF
--- a/packages/auto-instrumentations-node/README.md
+++ b/packages/auto-instrumentations-node/README.md
@@ -124,6 +124,20 @@ Notes:
 - Logs are always sent to console, no matter the environment, or debug level.
 - Debug logs are extremely verbose. Enable debug logging only when needed. Debug logging negatively impacts the performance of your application.
 
+### Flush telemetry on crash
+
+By default, the SDK only shuts down on `SIGTERM` or normal `beforeExit`. Uncaught exceptions, unhandled promise rejections, and `SIGINT` bypass `beforeExit`, so any spans/metrics/logs still buffered in a Batch processor are dropped — this is the typical "logs visible in stdout but never reach the collector" symptom.
+
+Set `OTEL_NODE_FLUSH_ON_CRASH=true` to register handlers for `uncaughtException`, `unhandledRejection`, and `SIGINT` that race `sdk.shutdown()` against a bounded timeout before exiting. The original error is written to `stderr` first, and the process still exits non-zero so the crash semantics are preserved.
+
+```shell
+export OTEL_NODE_FLUSH_ON_CRASH=true
+# Optional, defaults to 5000ms:
+export OTEL_NODE_FLUSH_ON_CRASH_TIMEOUT_MS=3000
+```
+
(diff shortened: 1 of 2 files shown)


raphael-theriault-swi:
ref #3379

PR #3511, opened 2026-05-08
ci(instrumentation-mongoose): increase minimum node version to match what mongoose 9 requires

wpessers:

The mongoose instrumentation causes the `test-all-versions` job for node 18 to hang until github default timeout (6h). The mongodb driver that is bundled by mongoose calls `crypto.randomUUID()` to create lsids. `crypto` however is marked unstable and behind an experimental feature flag for node 18: https://nodejs.org/docs/latest-v18.x/api/globals.html#crypto_1

Run that's hanging on a PR of mine that is completely unrelated to mongoose: https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/25541671848/job/74969459010?pr=2981
You can see a bunch of "...Error: crypto is not defined" errors at the end there. I think it's hanging indefinitely due to the cleanup hooks throwing before they can close the underlying mongo client, like this one: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/051a93abfdefd2d2530dfe9df11418ccffa906b7/packages/instrumentation-mongoose/test/mongoose-v7-v9.test.ts#L53-L57
Tries to call `mongoose.connection.close()` which underneath tries to use the `crypto` global as described above. So the call throws and the underlying mongo driver doesn't get torn down. Making it so the process doesn't exit and runs until the timeout.

Increased required node version for mongoose 9 tests.

### DIFF
--- a/packages/instrumentation-mongoose/.tav.yml
+++ b/packages/instrumentation-mongoose/.tav.yml
@@ -16,5 +16,5 @@ mongoose:
   - versions:
       include: ">=9 <10"
       mode: max-7
-    node: '>=18.0.0'
+    node: '>=20.19.0'
     commands: npm run test-v7-v9


david-luna (approved):
the "engines" section was updated in https://github.com/Automattic/mongoose/pull/15728/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

PR #3515, opened 2026-05-09
test(instrumentation-express): fix aborted request test on Windows

### LINKED ISSUE #3461: instrumentation-express: finish listener never called on aborted requests
The `finish` listener on middleware handlers is never called on aborted requests, leaving incomplete/orphaned spans.

In this block:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/c9d62be989802534c01373e8ab41e13747d7ee3e/packages/instrumentation-express/src/instrumentation.ts#L326-L334

The `finish` listener isn't always guaranteed to be called, if there is a client abort or some other cause of termination. Since we're just using this as a signal to end the span, we want to make sure it's is always called. From the Node docs:

### finish

> Emitted when the response has been sent. More specifically, this event is emitted when the last segment of the response headers and body have been handed off to the operating system for transmission over the network. It does not imply that the client has received anything yet.

### close

> Indicates that the response is completed, or its underlying connection was terminated prematurely (before the response completion).

The `http` client listens on the `close` event, the express instrumentation should probably do the same.

https://github.com/open-telemetry/opentelemetry-js/blob/786eefd3834fc36684874b1856fe42491d0b0b30/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L691-L703

Luffy-nani:

After the fix in #3462, the test 'should end spans when the client aborts the request' still fails on Windows (Node.js v22.17.0).

On Windows, clientReq.destroy() does not always emit an 'error' event unlike Linux/Mac, causing the assertion to fail before client events fire.

-

- Added a 'close' event listener on clientReq as a fallback to catch the abort event on Windows
- Added a small setTimeout delay after responseClosedPromise to allow client events to fire before the assertion runs

Fixes #3461
-


packages/instrumentation-express/test/express.test.ts
```diff
@@ -629,14 +629,24 @@ describe('ExpressInstrumentation', () => {
...
+          if (!clientError) {
+            clientError = new Error('connection closed');
+          }
+        });
+        
           await requestReceivedPromise;
           clientReq.destroy();
           await responseClosedPromise;
+          // wait for client events to fire (needed on Windows)
+          await new Promise(resolve => setTimeout(resolve, 100));
```
raphael-theriault-swi: Would this work with `setImmediate` instead ?

Luffy-nani: I tested with setImmediate but the test still fails on Windows. 
The setTimeout(100ms) seems necessary to give enough time for the client events to fire before the assertion runs on Windows.

Happy to investigate further if you have any other suggestions!

Luffy-nani:
Updated this to use an event-driven wait instead of a flat 100ms delay: a promise resolves as soon as clientReq emits 'error' (typical on Linux/Mac) or 'close' (can happen on Windows when 'error' isn't emitted), raced against a 200ms timeout as a safety net. This avoids polling entirely and keeps the wait minimal on platforms where the event fires immediately, while still giving Windows enough time.
Tested locally on Windows and it's passing.

@raphael-theriault-swi would love for you to take a look whenever you get a chance :)

raphael-theriault-swi (approved):
Thank you for the revisit ! Sorry for the lack of eyes on this, turns out none of the Express reviewers have a Windows machine to test on. I'm happy with this new approach.

PR #3521, opened 2026-05-12
fix(instrumentation-express): prevent post-handler middleware from overwriting rpcMetadata.route

Luffy-nani:

When a middleware runs after the request handler, it resets `rpcMetadata.route` back to `/`, overwriting the correct route already set by the request handler

Only update `rpcMetadata.route` if the new route is longer than or equal to the current one. This ensures that once a specific route like `/users` is set, a generic catch-all middleware with route `/` cannot overwrite it.
A regression test was added to verify that post-handler middlewares do not overwrite the correctly matched route.


raphael-theriault-swi:
Note to self & other reviewers: need to look at the history and PRs related to that piece of code because I'm pretty sure there have been very similar attempted fixes before.

PR #3528, opened 2026-05-15
fix(instrumentation-router): preserve baggage across next

### LINKED ISSUE #3527: @opentelemetry/instrumentation-router drops baggage added by middleware before next() on Express 5

  - `@opentelemetry/api` `1.9.1`
  - `@opentelemetry/sdk-node` `0.216.0`
  - `@opentelemetry/auto-instrumentations-node` `0.75.0`
  - `@opentelemetry/instrumentation-express` `0.65.0`
  - `@opentelemetry/instrumentation-router` `0.61.0`

v22.21.1

### What did you do?

I built a minimal Express 5 repro with this middleware pattern:

  1. middleware A adds baggage with `context.with(nextContext, next)`
  2. middleware B expects to see that baggage, adds more baggage, and calls `next()`
  3. the route handler expects to see both baggage entries

```ts
  app.use((req, res, next) => {
    const baggage = (
      propagation.getBaggage(context.active()) ?? propagation.createBaggage()
    ).setEntry('foo', { value: 'bar' });

    const nextContext = propagation.setBaggage(context.active(), baggage);
    context.with(nextContext, next);
  });

  app.use((req, res, next) => {
    console.log(
      'middleware B entry',
      propagation.getBaggage(context.active())?.getAllEntries()
    );

    const baggage = (
      propagation.getBaggage(context.active()) ?? propagation.createBaggage()
    ).setEntry('baz', { value: 'qux' });

    const nextContext = propagation.setBaggage(context.active(), baggage);
    context.with(nextContext, next);
  });

  app.get('/x', (req, res) => {
    console.log(
      'handler entry',
      propagation.getBaggage(context.active())?.getAllEntries()
    );
    res.end('ok');
  });
```
 compared Express 4 and Express 5 with the same OTel versions:

  - Express 4.21.2: baggage survives across middleware boundaries
  - Express 5.2.1: baggage is lost at the first next() handoff

Additionally, disabling @opentelemetry/instrumentation-router in the express v5 scenario will fix this issue. Express 5 uses a standalone router package as a dependency which may explain the discrepancy

Baggage propagation on both express 4 and 5

### What did you see instead?

Baggage loss on express 5 with @opentelemetry/instrumentation-router enabled

### Additional context

Came across this change, however, I believe a similar issue exists within @opentelemetry/instrumentation-router

https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3437/changes

ColinThompson1:
## Summary

  Fixes baggage loss across `next()` boundaries in `@opentelemetry/instrumentation-router`.

  This updates `wrappedNext` to preserve the current active context while restoring only the parent span before calling `next()`. That keeps middleware-added baggage intact without causing sibling middleware spans to nest under each
  other.

  Fixes #3527

  ## Problem

  `instrumentation-router` currently captures `parent = api.context.active()` before middleware execution and later calls `next()` with that full parent context.

  That drops baggage added by middleware before `next()`, because the newer active context is replaced with the older pre-middleware context.

  This affects Express 5 because it uses the standalone `router` package, so `@opentelemetry/instrumentation-router` participates in middleware handoff there.

  ## Changes

  - update `wrappedNext` in `packages/instrumentation-router/src/instrumentation.ts`
  - add a regression test covering baggage propagation across middleware and handler boundaries

  ## Notes

  The fix mirrors the baggage-preservation approach already used in @opentelemetry/instrumentation-express, but applies it to the router-layer next() wrapper used by Express 5.

### DIFF
--- a/packages/instrumentation-router/src/instrumentation.ts
+++ b/packages/instrumentation-router/src/instrumentation.ts
@@ -222,10 +222,13 @@ export class RouterInstrumentation extends InstrumentationBase {
         }
         span.end();
       }
-      if (parent) {
-        return api.context.with(parent, next, undefined, err);
-      }
-      return next(err);
+
+      const activeContext = api.context.active();
+      const nextContext = parentSpan
+        ? api.trace.setSpan(activeContext, parentSpan)
+        : activeContext;
+
+      return api.context.with(nextContext, next, undefined, err);
     };
 
     const ensureFallbackListener = () => {
--- a/packages/instrumentation-router/test/index.test.ts
+++ b/packages/instrumentation-router/test/index.test.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { context, trace } from '@opentelemetry/api';
+import { context, propagation, trace } from '@opentelemetry/api';
 import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
 import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
 import {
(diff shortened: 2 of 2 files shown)


trentm (approved):
Nice.

trentm:
@ColinThompson1 You'll need to sign the CLA (see the "EasyCLA" comment above) before we can accept this.

PR #3531, opened 2026-05-20
chore: zizmor checks

maryliag:
Adding workflow for zizmor checks

### DIFF
--- a/.github/workflows/zizmor-checks.yml
+++ b/.github/workflows/zizmor-checks.yml
@@ -1,0 +1,23 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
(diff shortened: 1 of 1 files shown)


trentm (approved):
Thanks!

trentm:
The failure in the zizmor action: I'll follow-up in a separate PR. It is related to a limitation in the YAML parser used by zizmor:

```
fatal: no audit was performed
error: failed to load file://./.github/workflows/close-stale.yml as workflow
  |
  = help: this typically indicates a bug in zizmor; please report it
  = help: https://github.com/zizmorcore/zizmor/issues/new?template=bug-report.yml

Caused by:
    0: failed to load file://./.github/workflows/close-stale.yml as workflow
    1: failed to load internal pathing document
    2: input is not valid YAML
```

PR #3536, opened 2026-05-21
chore: zizmor improvements

maryliag:
Checkout don't need the full history, so making it more efficient.
Adding option for workflow dispatch


maryliag:
test failure unrelated

david-luna (approved):

pichlermarc (approved):

PR #3537, opened 2026-05-21
fix(instrumentation-runtime-node): use finer GC duration buckets

mcollina:

- `v8js.gc.duration` used very coarse bucket boundaries (`[0.01, 0.1, 1, 10]`), which caused sub-second GC pauses like ~200ms to be reported in the `<=1s` bucket.

- switch the GC duration histogram to latency-style second buckets
- add a unit test covering the configured bucket boundaries

## Validation

- `npm run version:update -w @opentelemetry/instrumentation-runtime-node`
- `npm run compile -w @opentelemetry/instrumentation-runtime-node`
- `npm test -w @opentelemetry/instrumentation-runtime-node`

### DIFF
--- a/packages/instrumentation-runtime-node/src/metrics/gcCollector.ts
+++ b/packages/instrumentation-runtime-node/src/metrics/gcCollector.ts
@@ -11,8 +11,6 @@ import { Histogram, ValueType } from '@opentelemetry/api';
 import { BaseCollector } from './baseCollector';
 import { ATTR_V8JS_GC_TYPE, METRIC_V8JS_GC_DURATION } from '../semconv';
 
-// Use latency-style second buckets so sub-second GC pauses do not all collapse
-// into the `<=1s` bucket in histogram-based backends.
 const DEFAULT_GC_DURATION_BUCKETS = [
   0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10,
 ];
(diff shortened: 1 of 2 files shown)


pichlermarc:
@maryliag @trentm I seem to remember one (or both) of you being involved on [the Semantic Conventions side](https://github.com/open-telemetry/semantic-conventions/blob/cbb10b68a58c91a330ce8c9c66dd1a5195c6a886/docs/runtime/v8js-metrics.md#metric-v8jsgcduration) for this metric definition. WDYT?

I think having more granular buckets could be helpful for this metric. Introducing more buckets is usually not a breaking change.

maryliag:
that was indeed me! I see no issue with this change, but I would ask you @mcollina to update the semantic convention first [here](https://github.com/open-telemetry/semantic-conventions/blob/cbb10b68a58c91a330ce8c9c66dd1a5195c6a886/docs/runtime/v8js-metrics.md?plain=1#L29), then we can go ahead with this PR!
Feel free to tag me there once the PR is up

packages/instrumentation-runtime-node/src/metrics/gcCollector.ts
```diff
@@ -11,7 +11,11 @@ import { Histogram, ValueType } from '@opentelemetry/api';
 import { BaseCollector } from './baseCollector';
 import { ATTR_V8JS_GC_TYPE, METRIC_V8JS_GC_DURATION } from '../semconv';
 
-const DEFAULT_GC_DURATION_BUCKETS = [0.01, 0.1, 1, 10];
+// Use latency-style second buckets so sub-second GC pauses do not all collapse
+// into the `<=1s` bucket in histogram-based backends.
```
maryliag:
```suggestion
```

No need to add this comment, since after you make the semantic conventions change, it would be just following it

mcollina:
> I think having more granular buckets could be helpful for this metric. Introducing more buckets is usually not a breaking change.

It's not just helpful, the current buckets leads to bad decisions. 

A major gc break stop the event loop. There is a massive of difference between a 100ms major gc break and a 1s major gc break.
One is ok, the other one cause a measurable production incident and should raise alerts.

raphael-theriault-swi (approved):
Pending on semconv update

joaopgrassi:
Took the liberty to fix missing stuff on the semconv PR. It is merged now.

pichlermarc (approved):

PR #3194, opened 2025-10-24
feat(instrumentation-openai): support Responses API

btomaj:
## Which problem is this PR solving?

 - The existing OpenAI instrumentation covers Chat Completions and Embeddings but emits no telemetry for the newer Responses API. This PR adds spans, logs, and metrics for the Responses API.

## Short description of the changes

  - Adopt Semantic Conventions 1.38.0 for Responses API.
  - Replace hard-coded values with constants from Semantic Conventions 1.38.0.
  - Add instrumentation for `OpenAI.Responses.create`:
      1. Spans
        - Calls to `Responses.create` and `Responses.stream` (as a result of internally calling `Reponses.create`) start a client span, and records the request metadata (`model`, `max_output_tokens`, `temperature`, `top_p`, base URL), and if provided, system instructions.
        - Records the response ID, finish reason, and if provided, conversation ID.
      2. Logs
        - Emits input messages as an `gen_ai.client.inference.operation.details` event at request time.
        - For streaming responses, emits individual output messages as they arrive as an `gen_ai.client.inference.operation.details` event.
        - For non-streaming responses, emits the complete output messages on completion as an `gen_ai.client.inference.operation.details` event.
        - Tool/file-search/MCP/computer/code-interpreter/image-generation items are normalized into the [GenAI Output Message JSON schema](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-output-messages.json).
        - Message bodies are only included when `captureMessageContent` is enabled.
      4. Metrics – streaming and non-streaming flows record two histograms:
        - `gen_ai.client.operation.duration` (one sample per call, tagged with error type on failures)
        - `gen_ai.client.token.usage` (two samples per call—input vs output tokens—tagged with the response model and token type).
  - Add tests for `OpenAI.Responses.create` instrumentation:
    - Streaming and non-streaming Responses API scenarios with and without content capture.
    - Nock mocks for text, tool, file-search, MCP, computer, code-interpreter, and image-generation scenarios.

PR #3552, opened 2026-05-29
fix(deps): Add missing dev dependency breaking the build

hectorhdzg:
❌ > nx run @opentelemetry/resource-detector-alibaba-cloud:compile
  
  
  > @opentelemetry/resource-detector-alibaba-cloud@0.33.8 compile
  > tsc --build tsconfig.json tsconfig.esm.json
  
  Error: test/detectors/AlibabaCloudEcsDetectorIntegration.test.ts(8,37): error TS2307: Cannot find module '@opentelemetry/instrumentation-http' or its corresponding type declarations.
  npm error Lifecycle script `compile` failed with error:
  npm error code 2
  npm error path /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/resource-detector-alibaba-cloud
  npm error workspace @opentelemetry/resource-detector-alibaba-cloud@0.33.8
  npm error location /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/resource-detector-alibaba-cloud
  npm error command failed
  npm error command sh -c tsc --build tsconfig.json tsconfig.esm.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22364,9 +22364,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -22381,9 +22378,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -22398,9 +22392,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -22415,9 +22406,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -22364,9 +22364,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
```
trentm, line 22369: Not sure about dropping these "libc" entries. Not sure it matters to us. I've seen these added and removed in various PRs of late.

trentm (approved):
@hectorhdzg Thanks. I agree this should be in devDeps here.


However, how did you repro this, or where did you see this?
I tried to repro but could not.

(Aside: We use `knip ...` to attempt to find missing deps. However, we only use it for production dependency checking because of too many false positives in the past. I quickly tried and was unable to get `knip` to notice this missing instrumentation-http devDep. shrug.)

package-lock.json
pichlermarc: mhm, probably npm version differences.

pichlermarc:
> However, how did you repro this, or where did you see this? I tried to repro but could not.

I think the failure happened at #3194

pichlermarc (approved):

PR #3553, opened 2026-05-31
feat(instrumentation-aws-lambda)!: replace obsolete faas attributes w…

wpessers:
…ith corresponding incubating ones

Some obsolete resource attributes were still in use, while other language's instrumentation libs have already moved on. When upgrading semconv for the opentelemetry collector layer, that meant we need a temporary fallback until the js instrumentation also uses the latest ones. See: https://github.com/open-telemetry/opentelemetry-lambda/issues/2325

Swapped the obsolete attributes for their incubating counterparts.


pichlermarc (approved):

pichlermarc:
(changed the title to `feat!` because it's breaking for folks that rely on the old semconv, this way it's properly reflected in the changelog when we release)

PR #3555, opened 2026-06-01
fix(instrumentation-openai): guard stream iterator against chunks missing a choices array

### LINKED ISSUE #3554: instrumentation-openai: stream iterator crashes on chunks without a `choices` array (Cannot read properties of undefined (reading '0'))
### Component
`@opentelemetry/instrumentation-openai`

`@opentelemetry/instrumentation-openai@0.16.0` (latest). The defect is also present on `main` (latest commit to the file: `36a030c76`).

### What happened?

The streaming chat-completions instrumentation accesses `chunk.choices[0]` with optional chaining applied **after** the index:

```ts
const idx = chunk.choices[0]?.index ?? 0;
```

The `?.` guards `choices[0]` being undefined, but not `chunk.choices` itself being undefined. When a streamed chunk arrives without a `choices` array, `chunk.choices[0]` throws `TypeError: Cannot read properties of undefined (reading '0')`.

Because the chunk is `yield`-ed before the telemetry code runs, the throw surfaces inside the **consumer's** `for await` loop when it requests the next chunk — so an instrumentation-only defect breaks the application's stream consumption.

This is not hypothetical: OpenAI-compatible gateways/providers (in our case Gemini 2.5 Flash behind an OpenAI-compatible gateway) legitimately stream chunks where `choices` is absent (e.g. a usage-only or metadata chunk). The OpenAI SDK itself tolerates these; the instrumentation does not.

### Affected lines

On `main` (`packages/instrumentation-openai/src/instrumentation.ts`), all inside `_onChatCompletionsStreamIterator`:

- L473 `const idx = chunk.choices[0]?.index ?? 0;`
- L481 `const contentPart = chunk.choices[0]?.delta?.content;`
- L490 `const toolCallPart = chunk.choices[0]?.delta?.tool_calls?.[0];`
- L527 `const finishReason = chunk.choices[0]?.finish_reason;`

(Same four sites in `0.16.0` build output at lines 337/342/351/389.)

### Reproduce

Point the instrumented `OpenAI` client at any OpenAI-compatible endpoint that emits a streamed chunk without a `choices` array (or mock the stream to yield `{ id, model, usage }` with no `choices`). With `instrumentation-openai` enabled and streaming chat completions, the consumer's iteration throws `Cannot read properties of undefined (reading '0')`.

### Expected
Instrumentation must never throw into the consumer. A chunk lacking `choices` should be tolerated (skipped for telemetry purposes) exactly as the OpenAI SDK tolerates it.

### Actual
The instrumentation throws, propagating into the consumer's stream iteration and failing the request.

### Suggested fix
Move the optional chain before the index at all four sites:

```diff
- const idx = chunk.choices[0]?.index ?? 0;
+ const idx = chunk.choices?.[0]?.index ?? 0;
```
```diff
- const contentPart = chunk.choices[0]?.delta?.content;
+ const contentPart = chunk.choices?.[0]?.delta?.content;
```
```diff
- const toolCallPart = chunk.choices[0]?.delta?.tool_calls?.[0];
+ const toolCallPart = chunk.choices?.[0]?.delta?.tool_calls?.[0];
```
```diff
- const finishReason = chunk.choices[0]?.finish_reason;
+ const finishReason = chunk.choices?.[0]?.finish_reason;
```

Happy to open a PR with the change plus a regression test (a streamed chunk with no `choices` array).

micael-mbagira-parloa:

Fixes #3554.

The streaming chat-completions instrumentation gathers telemetry from each
stream chunk with an unguarded `chunk.choices[0]`:

```ts
const idx = chunk.choices[0]?.index ?? 0;
```

The optional chaining is applied **after** the `[0]` index, so it guards
`choices[0]` being `undefined` but not `chunk.choices` itself being
`undefined`. Some OpenAI-compatible backends legitimately stream a chunk with
no `choices` array at all (e.g. **Gemini 2.5** behind an OpenAI-compatible
gateway, which emits metadata/usage-only chunks). On such a chunk,
`chunk.choices[0]` throws `TypeError: Cannot read properties of undefined
(reading '0')`.

Because the iterator `yield`s the chunk **before** gathering telemetry, the
throw surfaces inside the consumer's `for await` loop on the next iteration —
so an instrumentation-only defect breaks the application's stream consumption.
The OpenAI SDK itself tolerates these chunks; the instrumentation should too.

- Guard all four `chunk.choices[0]` accesses in
  `_onChatCompletionsStreamIterator` with `chunk.choices?.[0]` so a chunk
  lacking a `choices` array is skipped for telemetry instead of throwing.
- Add a regression test (`copes with chunks missing a choices array`) plus its
  nock fixture. The fixture is hand-authored rather than recorded, because the
  triggering behavior comes from non-OpenAI backends and cannot be reproduced
  against `api.openai.com`; the stream contains a content chunk, a chunk with
  **no** `choices` key, and a terminating `finish_reason: "stop"` chunk. The
  test asserts the stream is consumed without throwing and the span is still
  produced.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- `npm test` for `@opentelemetry/instrumentation-openai` (added regression test
  passes; existing tests unaffected).
- Verified the new test fails without the source change (throws
  `Cannot read properties of undefined (reading '0')`) and passes with it.

## Checklist:

- [x] Followed the [style guidelines](https://github.com/open-telemetry/opentelemetry-js/blob/main/CONTRIBUTING.md) of this project
- [x] Unit tests have been added
- [ ] Documentation has been updated (n/a)

### DIFF
--- a/packages/instrumentation-openai/test/openai.test.ts
+++ b/packages/instrumentation-openai/test/openai.test.ts
@@ -1773,6 +1773,43 @@ describe('OpenAI', function () {
       });
     });
 
+    it('copes with chunks missing a choices array', async () => {
+      // Regression test for #3554. Some OpenAI-compatible backends (e.g.
+      // Gemini 2.5 via a gateway) stream a chunk that omits the `choices`
+      // array entirely. The stream iterator must not throw on such a chunk;
+      // previously `chunk.choices[0]` threw "Cannot read properties of
+      // undefined (reading '0')", which propagated into the caller's
+      // `for await` loop and failed the request.
+      const messages: OpenAI.Chat.Completions.ChatCompletionMessageParam[] = [
+        {
+          role: 'user',
+          content: input,
+        },
+      ];
+      const stream = await client.chat.completions.create({
+        model,
+        messages,
+        stream: true,
+      });
+      let content = '';
+      for await (const part of stream) {
+        content += part.choices?.[0]?.delta?.content || '';
+      }
+      expect(content).toEqual('Atlantic Ocean.');
+
+      // The span is still produced and finished cleanly despite the
+      // choices-less chunk.
+      const spans = getTestSpans();
+      expect(spans.length).toBe(1);
+      expect(spans[0].attributes).toMatchObject({
+        [ATTR_GEN_AI_OPERATION_NAME]: 'chat',
+        [ATTR_GEN_AI_REQUEST_MODEL]: model,
+        [ATTR_GEN_AI_SYSTEM]: 'openai',
+        [ATTR_GEN_AI_RESPONSE_FINISH_REASONS]: ['stop'],
+        [ATTR_GEN_AI_RESPONSE_MODEL]: 'gpt-4o-mini-2024-07-18',
+      });
+    });
+
     it('copes with .withResponse() usage', async () => {
       const messages: OpenAI.Chat.Completions.ChatCompletionMessageParam[] = [
         {
--- a/packages/instrumentation-openai/src/instrumentation.ts
+++ b/packages/instrumentation-openai/src/instrumentation.ts
@@ -470,7 +470,7 @@ export class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumenta
         'OpenAI.Chat.Completions.create stream chunk: %O',
         chunk
       );
-      const idx = chunk.choices[0]?.index ?? 0;
+      const idx = chunk.choices?.[0]?.index ?? 0;
       if (!choices[idx]) {
         choices[idx] = {} as {
           content: string;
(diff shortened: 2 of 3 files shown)


packages/instrumentation-openai/test/openai.test.ts
```diff
@@ -1773,6 +1773,43 @@ describe('OpenAI', function () {
       });
     });
 
+    it('copes with chunks missing a choices array', async () => {
+      // Regression test for #3554. Some OpenAI-compatible backends (e.g.
```
hectorhdzg, line 1777: No need to add details about the actual resolved issues in the code, including Github issue numbers, we have changelog for that

hectorhdzg (approved):
LGTM

PR #3557, opened 2026-06-04
fix(instrumentation-express): filter Express v5 wildcard paths in get…

### LINKED ISSUE #3551: `@opentelemetry/instrumentation-express`: Doesn't filter out Express v5 dynamic path fragments

`@opentelemetry/instrumentation-express` @ 0.65.0

`20.20.2`

### What did you do?

Used with Express v5, I'm seeing dynamic path fragments not getting filtered in spans.

Cleaned paths. 

### What did you see instead?

Image

### Additional context

It seems to me that line 214–216 in [`src/utils`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-express/src/utils.ts), the function `getConstructedRoute`, should include a filter for Express v5 paths, probably something like: 

```ts
const meaningfulPaths = layersStore.filter(
  path =>
    path !== '/' &&
    path !== '/*' &&
    !/^\/\{\*[^}]*\}$/.test(path) // Express v5: /{*path}, /{*}, etc.
);
```

Luffy-nani:

Express v5 changed how wildcard paths work, instead of `/*` it now uses syntax like `/{*path}` or `/{*}`. The existing filter in getConstructedRoute only knew about the old Express v4 style, so these new wildcards were leaking into span names and making http.route messy

Fixes #3551 

Express v5 uses a new wildcard syntax like `/{*path}` for catch-all routes. These fragments were leaking into http.route span attributes because the filter in getConstructedRoute only stripped out `/` and `/*` from v4. This PR extends the filter with a regex to catch the v5 style too. Added unit tests in utils.test.ts for a few different patterns and an integration test in express.test.ts that actually spins up an Express v5 server with wildcard middlewares and checks that none of the spans end up with a wildcard fragment as their route, not just the request handler span.

### DIFF
--- a/packages/instrumentation-express/src/utils.ts
+++ b/packages/instrumentation-express/src/utils.ts
@@ -212,10 +212,7 @@ export function getConstructedRoute(req: {
     : [];
 
   const meaningfulPaths = layersStore.filter(
-    path =>
-      path !== '/' &&
-      path !== '/*' &&
-      !/^\/\{[*][^}]*\}$/.test(path)
+    path => path !== '/' && path !== '/*' && !/^\/\{[*][^}]*\}$/.test(path)
   );
 
   if (meaningfulPaths.length === 1 && meaningfulPaths[0] === '*') {
--- a/packages/instrumentation-express/src/utils.ts
+++ b/packages/instrumentation-express/src/utils.ts
@@ -212,6 +212,7 @@ export function getConstructedRoute(req: {
     : [];
 
   const meaningfulPaths = layersStore.filter(
+    // Exclude root path, wildcard '/*', and Express v5 wildcard segments like /{*param} or /{**param}
     path => path !== '/' && path !== '/*' && !/^\/\{[*][^}]*\}$/.test(path)
   );
 
--- a/packages/instrumentation-express/src/utils.ts
+++ b/packages/instrumentation-express/src/utils.ts
@@ -212,7 +212,10 @@ export function getConstructedRoute(req: {
     : [];
 
   const meaningfulPaths = layersStore.filter(
-    path => path !== '/' && path !== '/*'
+    path =>
+      path !== '/' &&
+      path !== '/*' &&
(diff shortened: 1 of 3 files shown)


O-BERNARDOFOEGBU:
Hi! I independently investigated the same issue and opened #3560 before noticing this PR.

One thing I found while working through the fix was that adding an integration test in `express.test.ts` helped verify that Express v5 wildcard fragments do not appear in the generated `http.route` attribute, in addition to the utility-level tests.

Feel free to cherry-pick any relevant test coverage from #3560 if it's useful.

Luffy-nani:
Thanks! Took a look at your PR and went ahead and added an integration test here too. It covers a couple of different wildcard patterns and loops over all spans to make sure none of them end up with a wildcard fragment as their route, not just the request handler. Also noticed the regex is slightly different here. [*][^}]* properly negates }, while [\*[^}]] inside a single character class doesn't quite work the way it looks like it should

O-BERNARDOFOEGBU:
Glad I could help! Thanks for taking a look at my PR and incorporating the integration test coverage. I especially like the approach of validating all spans rather than only the request handler span.

Thanks for pointing out the regex nuance as well. Happy to see the issue getting fixed and the test coverage strengthened.

packages/instrumentation-express/src/utils.ts
```diff
@@ -212,7 +212,7 @@ export function getConstructedRoute(req: {
     : [];
 
   const meaningfulPaths = layersStore.filter(
-    path => path !== '/' && path !== '/*'
+    path => path !== '/' && path !== '/*' && !/^\/\{[*][^}]*\}$/.test(path)
   );
```
raphael-theriault-swi, line 217: It would be nice to have a commend describing what the regex does given it's not immediately obvious.

raphael-theriault-swi (commented):
Thank you for the fix ! Minor request but otherwise this looks good :)

raphael-theriault-swi (approved):
LGTM

PR #3560, opened 2026-06-06
fix(instrumentation-express): filter Express v5 wildcard path fragmen…

### LINKED ISSUE #3551: `@opentelemetry/instrumentation-express`: Doesn't filter out Express v5 dynamic path fragments

O-BERNARDOFOEGBU:

- Express v5 (via path-to-regexp v8) introduced a new wildcard route syntax: /{*param} (for example, app.use('/{*path}', handler)).

- The existing route construction logic in getConstructedRoute() already filtered Express v4 wildcard fragments such as '/*', but did not filter the new Express v5 wildcard syntax. As a result, wildcard fragments could appear directly in the constructed route and http.route span attribute.

- This PR extends the existing filtering logic to treat Express v5 wildcard fragments the same way Express v4 wildcard fragments are treated.

- **Fixes #3551.**

- Extended the `getConstructedRoute` filter in `src/utils.ts` to also strip
  Express v5 wildcard path segments matching `/^\/\{\*[^}]*\}$/`
- Added 3 unit tests in `utils.test.ts`
- Added 1 integration test in `express.test.ts` (v5-gated)

- Extend route filtering in getConstructedRoute() to exclude Express v5 wildcard fragments matching /{*param}.
- Added unit tests covering:
  - standalone wildcard fragments
  - wildcard fragments mixed with meaningful route segments
  - arbitrary wildcard parameter names
- Added an integration test verifying that Express v5 wildcard fragments do not appear in the generated http.route attribute.
- Preserve existing Express v4 wildcard behavior.

### DIFF
--- a/packages/instrumentation-express/src/utils.ts
+++ b/packages/instrumentation-express/src/utils.ts
@@ -212,7 +212,13 @@ export function getConstructedRoute(req: {
     : [];
 
   const meaningfulPaths = layersStore.filter(
-    path => path !== '/' && path !== '/*'
+    path =>
+      path !== '/' &&
+      path !== '/*' &&
+      // Express v5 uses /{*param} wildcard syntax (path-to-regexp v8).
+      // These catch-all fragments are meaningless for route identification,
+      // the same as '/*' was in Express v4.
+      !/^\/\{\*[^}]*\}$/.test(path)
   );
 
   if (meaningfulPaths.length === 1 && meaningfulPaths[0] === '*') {
--- a/packages/instrumentation-express/test/express.test.ts
+++ b/packages/instrumentation-express/test/express.test.ts
@@ -499,6 +499,61 @@ describe('ExpressInstrumentation', () => {
       );
     });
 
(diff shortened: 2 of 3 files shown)


raphael-theriault-swi:
This should be fixed with #3557. Still, thank you for the contribution !

PR #3565, opened 2026-06-10
fix(ci): resolve Codecov binary verification failure - fix GPG key pipe and keyring

Luffy-nani:

The "Report Coverage with Flags" CI step was consistently failing with `gpg: no valid OpenPGP data found` and exiting with code 1, meaning coverage reports were never getting uploaded to Codecov.

Dug into the failure and found three things wrong in the verification script:

1. **The GPG key import was silently broken** — wrapping curl in `echo "$(curl ...)"` corrupts the PGP data before it even reaches GPG (subshell substitution strips trailing newlines). Switched to piping curl directly instead.

2. **Wrong Keybase account** — the script was pointing at `codecovsecurity` which is the old legacy uploader account. Since we're using the CLI binary from `cli.codecov.io`, the correct account is `codecovsecops`. Codecov's own GitHub Action made the same switch in v7.

3. **Missing keyring flag on import** `--keyring trustedkeys.gpg` was present on the verify step but not the import, so the key was never actually accessible when verification ran. Added it to the import step too.

**References:**
- [Codecov CLI integrity checking docs](https://docs.codecov.com/docs/codecov-uploader)
- [Codecov GitHub Action v7 migration to `codecovsecops`](https://github.com/codecov/codecov-action)


pichlermarc (approved):
thanks!

pichlermarc (approved):

Issue #3236, opened 2025-11-21
@opentelemetry/instrumentation-express - Bare middleware can't resolve value for `http.route` attribute


vitorvasc:
### What version of OpenTelemetry are you using?

```json
"dependencies": {
    "@opentelemetry/api": "^1.9.0",
    "@opentelemetry/instrumentation-express": "^0.57.0",
    "@opentelemetry/instrumentation-http": "^0.208.0",
    "@opentelemetry/sdk-metrics": "^2.2.0",
    "@opentelemetry/sdk-node": "^0.208.0",
    "@opentelemetry/sdk-trace-node": "^2.2.0",
    "express": "^5.1.0"
}
```

### What version of Node are you using?

v24.11.0

This behavior was also reproduced in other applications using v18 and above.

### What did you do?

When an Express application includes a bare middleware that returns a response before the route handler runs — such as an ETag cache check returning `304 Not Modified` — the `http.route` attribute isn't populated with the full matched route.

The snippet below is a minimal example. Full code is available at: https://github.com/vitorvasc/js-otel-instrumentation-express.

**main.js**

```js 
// Bare Middleware that may return early (e.g., cache hit with 304)
apiRouter.use([(req, res, next) => {
  const etag = req.headers['if-none-match'];
  const currentETag = '"12345-abcdef"';

  if (etag && etag === currentETag) {
    res.status(304).end();  // Early return
    return;
  }
  next();
}]);

// Define routes
apiRouter.get('/users', handler);
apiRouter.get('/products', handler);

// Mount router
app.use('/', apiRouter);
``` 

**cURL**
```bash
curl http://localhost:${port}/api/users -H 'If-None-Match: "12345-abcdef"'
```

### What did you expect to see?

Even if a middleware responds early, the `http.route` attribute should reflect the full matched route.

Expected behavior:

- Normal request (200): `http.route = "/api/users"`
- Bare middleware (304): `http.route = "/api/users"`

### What did you see instead?

When a bare middleware finishes the request, the instrumentation captures only the router's mount path and loses the specific route information:

Actual behavior:

- Normal request (200): `http.route = "/api/users"` 
- Bare middleware (304): `http.route = "/api"`

In this scenario, only `/api` is recorded instead of `/api/users`.

<details>
<summary>Metric output: `http.server.duration`</summary>

```javascript
{
  descriptor: {
    name: 'http.server.duration',
    type: 'HISTOGRAM',
[...]

Issue #2884, opened 2025-06-14
Express instrumentation: rpcMetadata.route overwritten by middlewares after request handler


Pratyush0812:
### What version of OpenTelemetry are you using?
"@opentelemetry/api": "1.9.0",
"@opentelemetry/auto-instrumentations-node": "0.58.0",
"@opentelemetry/sdk-node": "0.200.0",
"@opentelemetry/sdk-trace-base" : "2.0.0",
"@opentelemetry/sdk-trace-base" : "2.0.0",
"@opentelemetry/instrumentation-express": "0.48.1"

### What version of Node are you using?
v20.18.1

### What did you do?

I instrumented my Express application using @opentelemetry/instrumentation-express alongside @opentelemetry/instrumentation-http. While tracing incoming HTTP requests, I noticed an inconsistency in how the http.route attribute is set in spans, depending on the order and position of middleware layers relative to the request handler.

In some cases, middleware layers executed after the request handler were resetting the RPCMetadata.route value to '/', which caused the http.route attribute in the SERVER span to be incorrectly reported as '/' instead of the actual matched route.

### What did you expect to see?
The route stored in RPCMetadata.route should remain unchanged once set by the request handler layer, and subsequent middleware layers should not overwrite it back to '/'.

### What did you see instead?
In some cases, middleware layers that execute after the request handler are resetting the RPCMetadata.route value back to '/'.
As a result, the http.route attribute on the SERVER span incorrectly gets overwritten to '/' during these middleware executions, even though the correct route  had already been set by the request handler.

### Suggested Fix
Suggested fix
Update the Express instrumentation’s _applyPatch logic to ensure that RPCMetadata.route is only set once per request, ideally during the first route handler layer (or when it’s not already set).
Subsequent middleware layers, especially those with layer paths like '/', should not override the existing value in RPCMetadata.route if it’s already populated.

One possible way:
```
if (rpcMetadata?.type === RPCType.HTTP) {
  if (!rpcMetadata.route || rpcMetadata.route === '/') {
    rpcMetadata.route = route || '/';
  }
}
```

I am fairly new to this instrumentation code and my understanding or logic might be flawed. Please kindly help with a fix or guidance on how to properly handle setting the route in middleware and handler layers. I appreciate your support!

Pratyush0812:
If I’m thinking along the right lines here, I would be more than happy to raise a PR addressing this issue. Please let me know if that would be helpful!

[...]

PR #3571, opened 2026-06-12
fix(instrumentation-express): fix missing http.route when path-less middleware is ignored

### LINKED ISSUE #3570: `@opentelemetry/instrumentation-express` http.route undefined on non-parameterized routes when `ignoreLayersType: ['middleware']` is used

- `@opentelemetry/instrumentation-express`: `0.66.0`
- `@opentelemetry/instrumentation`: `0.218.0`
- `@opentelemetry/api`: `1.9.1`

`v22.22.0`. Tested on Express `4.22.1` and `5.2.1` — both reproduce identically.

### What did you do?

Set up an Express app with a nested sub-router that has one path-less `router.use(middleware)` and two route handlers — one non-parameterized, one parameterized. Configured `ExpressInstrumentation` with `ignoreLayersType: ['middleware']`. Reproduction script included below.

A `http.route` attribute in all server spans. This was behavior in 0.48.0, which was a version our service was using before upgrade to 0.66.0:

```
/api/leaf   ->  http.route = /leaf
/api/p/123  ->  http.route = /p/:id
```

### What did you see instead?

For non-parameterised routes, the `http.route` value is missing. This broke our dashboards that used queries dependent on `http.route`:
```
/api/leaf   ->  http.route = <MISSING>
/api/p/123  ->  http.route = /p/:id
```

The same thing happens to all requests that exit through a path-less `router.use(...)` middleware before reaching a route handler (e.g. middleware that responds with `401`/`403`/`410`/`503`).

This appears to be a regression introduced by #2843. We discovered it when our dashboards started working after upgrade from 0.48.0 to 0.66.0.

#### Reproduction script 

```js
const { registerInstrumentations } = require('@opentelemetry/instrumentation');
const { HttpInstrumentation } = require('@opentelemetry/instrumentation-http');
const { NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
const { SimpleSpanProcessor, InMemorySpanExporter } = require('@opentelemetry/sdk-trace-base');

const exporter = new InMemorySpanExporter();
new NodeTracerProvider({ spanProcessors: [new SimpleSpanProcessor(exporter)] }).register();
registerInstrumentations({
  instrumentations: [
    new HttpInstrumentation(),
    new ExpressInstrumentation({ ignoreLayersType: ['middleware'] }),
  ],
});

const http = require('http');

const app = express();
const api = express.Router();

api.use((req, res, next) => next());             // path-less middleware (the trigger)
api.get('/leaf', (req, res) => res.end('ok'));   // non-parameterized route
api.get('/p/:id', (req, res) => res.end('ok'));  // parameterized route (control)

app.use('/api', api);

const server = http.createServer(app).listen(0, async () => {
  const port = server.address().port;
  for (const p of ['/api/leaf', '/api/p/123']) {
    await new Promise(resolve => {
      http.get(`http://127.0.0.1:${port}${p}`, res => {
        res.on('data', () => {});
        res.on('end', resolve);
      });
    });
  }
  setTimeout(() => {
    for (const span of exporter.getFinishedSpans().filter(s => s.kind === 1)) {
      const route = span.attributes['http.route'] ?? '<MISSING>';
      console.log(`${span.attributes['http.target']}  ->  http.route = ${route}`);
    }
    server.close();
  }, 100);
});
```

#### Behavior without middleware

When the `api.use(...)` line with middleware is deleted, the returns are

```
/api/leaf   ->  http.route = /api/leaf
/api/p/123  ->  http.route = /api/p/:id
```

Note the additional effect of having middleware included: all routes, including parameterised, lose their `/api/` prefix with middlesare (`/p/:id` instead of `/api/p/:id`)

#### Related issues

- #1947 (closed).
- #2884 (open) — same probable root cause, different symptom: route overwritten to `/` by post-handler middleware.
- #3236 (open) — same probable root cause, different symptom: `http.route` reports the enclosing router's mount path when a middleware returns early.

#### Root cause research

Based on Claude research, there are two interacting issues:

1. **Long-standing**: when a middleware layer is registered without a path (`router.use(fn)`), the `isLayerIgnored` block in `instrumentation.ts` still pops one entry off `req[_LAYERS_STORE_PROPERTY]` for any MIDDLEWARE-type layer. Because the middleware never pushed anything, the pop instead strips the mount path contributed by the enclosing router. Each ignored path-less middleware therefore silently erases one mount-path entry.

2. **Introduced by #2843**: by the time the route handler's local path is pushed, the store has been popped down to just that local path (e.g. `['/leaf']`). `getActualMatchedRoute` then validates:

   ```ts
   const isValidRoute = normalizedRoute.length > 0 &&
       (req.originalUrl === normalizedRoute ||
        req.originalUrl.startsWith(normalizedRoute) ||
        isRoutePattern(normalizedRoute));
   ```

   `req.originalUrl` does **not** start with `normalizedRoute` because parts of the URL have been removed, and `isRoutePattern` only matches strings containing `:` or `*`. So `actualMatchedRoute` returns `undefined`. The HTTP instrumentation omits `http.route` from the SERVER span as a result.

Pre-#2843 the same popping behavior existed, but the route attribute was set via `route || '/'` (a simple filter + join), so the leaf-only path survived as a usable (if truncated) value. The new validator removes that safety net.

Luffy-nani:

When ignoreLayersType includes middleware and a path-less middleware (router.use(fn) with no path argument) is present, http.route ends up missing entirely for non-parameterized routes, and has its mount prefix stripped for parameterized routes

Traced this down to the ignored-layer branch in _applyPatch unconditionally popping from the layers store even when nothing was pushed which happens when getLayerPath returns undefined for a path-less middleware. This silently 
eats the mount path pushed by the enclosing router. The stricter validator added in #2843 is what turns this silent truncation into a fully missing or prefix-stripped http.route

Fixes #3570
Related: #2884, #3236

Gated the pop in the ignored-middleware branch on isLayerPathStored, so we only pop what was actually pushed. One condition added in instrumentation.ts:

if (type === ExpressLayerType.MIDDLEWARE && isLayerPathStored) {

Also added a regression test in custom-config.test.ts covering the exact scenario from the issue nested router with a path-less middleware alongside non-parameterized and parameterized leaf routes, with ignoreLayersType: 
['middleware']. Verified /api/leaf now correctly gets http.route = /api/leaf and /api/p/123 gets /api/p/:id instead of /p/:id.

## Breaking Change
Teams relying on the previous (incorrect) http.route values in dashboards or alerts will see the full prefixed route after this fix (e.g. /api/v2/sign-in instead of /sign-in). This is the correct behavior but may require dashboard updates on upgrade.

## Verification
Tested by issue reporter (@myrosia) against a real app (0.48.0 → 0.66.0 + this PR):

| Endpoint | Pre-upgrade (0.48.0) | Post-upgrade vanilla (0.66.0) | Post-upgrade + this PR |
|---|---|---|---|
| POST /api/v2/sign-in | /sign-in | MISSING | /api/v2/sign-in |
| GET /api/v2/profiles | /profiles | MISSING | /api/v2/profiles |
| GET /api/v2/assessments | /assessments | MISSING | /api/v2/assessments |
| GET /api/v2/me | /me | MISSING | /api/v2/me |
| GET /api/v2/home-cards | /home-cards | MISSING | /api/v2/home-cards |
| GET /api/v2/features | /features | MISSING | /api/v2/features |
| GET /api/v2/health-profiles/:profileKey | /health-profiles/:profileKey | /health-profiles/:profileKey | /api/v2/health-profiles/:profileKey |
| GET /api/v2/assessments/:assessmentId/pdf | /assessments/:assessmentId/pdf | /assessments/:assessmentId/pdf | /api/v2/assessments/:assessmentId/pdf |
| GET /api/v2/assessments/:assessmentId | /assessments/:assessmentId | /assessments/:assessmentId | /api/v2/assessments/:assessmentId |


packages/instrumentation-express/test/custom-config.test.ts
```diff
@@ -61,6 +61,9 @@ describe('ExpressInstrumentation', () => {
 
     afterEach(() => {
       server.close();
+      instrumentation.setConfig({
+        ignoreLayersType: [ExpressLayerType.MIDDLEWARE],
+      });
```
raphael-theriault-swi: This should only be enabled for tests where it is relevant

packages/instrumentation-express/test/custom-config.test.ts
```diff
@@ -142,19 +146,19 @@ describe('ExpressInstrumentation', () => {
...
         }
       );
     });
 
     it('should correctly name http root path when its /', async () => {
       instrumentation.setConfig({
-        ignoreLayerTypes: [
-          ExpressLayerType.MIDDLEWARE,
-          ExpressLayerType.REQUEST_HANDLER,
-        ],
```
raphael-theriault-swi, line 155: This changes the behaviour of an existing test, I'd prefer the assertion to be changed if the result of the existing test was corrected by this change.

raphael-theriault-swi (changes requested):

Luffy-nani:
so, I've addressed both of your comments!
For the first one, I removed the setConfig from afterEach and moved it inside the regression test instead, since that's the test where it is actually relevant
For the second one, I kept the original config (MIDDLEWARE + REQUEST_HANDLER) in the existing test and updated the assertion instead. Since REQUEST_HANDLER is ignored, no request handler span should be emitted, but rpcMetadata.route is still correctly set to '/' because the route assignment happens before the ignore check in the instrumentation

Please take a look and let me know if this was the intended fix, happy to adjust if needed! @raphael-theriault-swi

raphael-theriault-swi (approved):
Thank you !

PR #3576, opened 2026-06-20
fix(instrumentation-user-interaction): use WeakMap for per-element listener bookkeeping

### LINKED ISSUE #3539: UserInteractionInstrumentation leaks detached DOM elements via plain Maps in _wrappedListeners
**What version of OpenTelemetry are you using?**
@opentelemetry/instrumentation-user-interaction@0.56.0 (latest at time of filing). Pulled in transitively via @maersk-global/telemetry-web-sdk@5.0.1, but reproducible directly against the OTel package — the SDK wrapper is not required.

**What version of Node are you using?**
Node 20.x (also reproduced on 18.x). The leak is in the instrumentation's bookkeeping data structure and is platform-independent — it reproduces under jest/jsdom and in real Chromium tabs (Chrome 124+).

**What did you do?**
UserInteractionInstrumentation keeps a private bookkeeping structure of shape WeakMap<listener, Map<eventType, Map<DOMElement, wrappedListener>>>. The outer WeakMap self-cleans, but the two inner Maps are plain Maps — and the innermost one uses the DOM element itself as a key. As long as the user-supplied listener stays reachable (which is the normal case for Vue's createFnInvoker wrapper, React's bound handlers held by an unmounted parent fiber, etc.), every DOM element that listener was ever attached to is strongly retained, even after the element is removed from the document.

Source (verified in node_modules/@opentelemetry/instrumentation-user-interaction@0.56.0/build/esm/instrumentation.js):

_wrappedListeners = new WeakMap();         // line 40  — outer, OK
...
addPatchedListener(on, type, listener, wrappedListener) {
  let listener2Type = this._wrappedListeners.get(listener);
  if (!listener2Type) {
    listener2Type = new Map();             // line 167 — LEAK: plain Map
    this._wrappedListeners.set(listener, listener2Type);
  }
  let element2patched = listener2Type.get(type);
  if (!element2patched) {
    element2patched = new Map();           // line 172 — LEAK: plain Map
    listener2Type.set(type, element2patched);
  }
  ...
  element2patched.set(on, wrappedListener);// line 178 — DOM element as key
}
Minimal deterministic reproduction (Jest + jsdom, no framework required):

type Leaky = UserInteractionInstrumentation & {
  _wrappedListeners: WeakMap<
    EventListenerOrEventListenerObject,
    Map<string, Map<EventTarget, EventListenerOrEventListenerObject>>
  >;
};

test("detached DOM is retained by inner Map", () => {
  const inst = new UserInteractionInstrumentation() as Leaky;
  inst.enable();

  const sharedListener = () => {};
  const container = document.createElement("div");
  document.body.appendChild(container);

  for (let i = 0; i < 200; i++) {
    const button = document.createElement("button");
    container.appendChild(button);
    button.addEventListener("click", sharedListener);
  }

  // Simulate framework re-render / unmount. Framework does NOT call
  // removeEventListener with the exact same function reference — it relies
  // on GC of the element + wrapper.
  container.innerHTML = "";
  document.body.removeChild(container);

  const byType = inst._wrappedListeners.get(sharedListener)!;
  const byElement = byType.get("click")!;

  expect(byElement).toBeInstanceOf(Map);     // bug: should be WeakMap (or gone)
  expect(byElement.size).toBe(200);          // all 200 detached <button>s still here
  for (const el of byElement.keys() as Iterable<HTMLElement>) {
    expect(el.parentNode).toBeNull();        // they really are detached
  }
});
A complementary GC-backed test (run under node --expose-gc) attaches the listener to one element, drops every test-scope reference, forces GC, and confirms a WeakRef to the element still resolves — i.e. the only remaining retainer is the inner Map.

Both assertions pass against @opentelemetry/instrumentation-user-interaction@0.56.0.

**What did you expect to see?**
After the elements are removed from the DOM and the consumer drops their strong references, the instrumentation's internal bookkeeping should not prevent the elements from being garbage-collected. A correct implementation would either:

Replace the two inner Maps with WeakMaps (the innermost is keyed by a DOM element, so this is the natural fix), or
Drop the per-element bookkeeping entirely and instrument event dispatch via delegation, or
Use WeakRef + a periodic sweep.
**What did you see instead?**
Every DOM element that the patched addEventListener was ever called with is strongly retained inside element2patched (instrumentation.js:172), indefinitely, until the user-supplied listener function itself is collected. In any modern SPA framework, the listener wrapper outlives the element it was bound to, so this retention is effectively permanent for the lifetime of the tab.

Real-world impact (Maersk Customs Integration Platform, ~2,000 daily users, 8+ hour shifts):

Heap snapshot showed a single Map with 6,073 entries of detached DOM elements (≈115 kB just for the Map; the retained DOM + Vue component instances were the dominant cost), retainer chain terminating in the OTel instrumentation code path.
Tab heap grew from 290 MB baseline to 2.6–4 GB within minutes of normal use, ending in Aw, Snap! OOM crashes.
With interactions: false, heap stays under 2 GB even after heavy navigation and returns near baseline on idle.
Additional context
No framework involvement is required to reproduce. The Jest test above runs against the OTel package directly in jsdom. Vue/React only make the leak fatal faster by churning the DOM at scale; they are not the cause.
Every consumer is exposed by default. The Faro/@maersk-global/telemetry-web-sdk family enable interactions: true by default, and that's the recommended OTel browser default as well. Apps only escape the leak by explicitly opting out.
**Why hasn't this been reported before?** Our best guess is that most consumer apps aren't long-lived enough for the leak to be fatal. Ours are (8+ hour shifts with heavy DOM churn over the same tab), which is why it surfaces here.
removePatchedListener does not save us: it only fires when the consumer calls removeEventListener with the exact same function reference. Vue (createFnInvoker) and React (bound handlers) generate fresh wrappers per render and rely on GC for cleanup — they never call removeEventListener on unmount.
**Suggested minimum fix at the source level:** change new Map() at lines 167 and 172 to new WeakMap() (the innermost is keyed by a DOM element, so this is type-safe), or remove the per-element bookkeeping entirely in favor of delegated event capture.

Renegade2345:
## What

`_wrappedListeners` tracks patched event listeners via `WeakMap<listener, Map<type, Map<HTMLElement, Function>>>`. The innermost map was a plain `Map` keyed by the DOM element itself. That strongly retains every element a listener was ever attached to until the listener function itself becomes unreachable.

`removePatchedListener` only cleans up an entry when the consumer calls `removeEventListener` with the exact same listener reference. SPA frameworks often skip this on unmount. Vue's `createFnInvoker` and React's bound handlers generate fresh wrapper functions each render and rely on garbage collection of the element instead. Under that pattern every element a shared listener was attached to stays referenced for the life of the page, even after it's removed from the DOM.

In long-lived tabs with heavy DOM churn this leads to unbounded heap growth and eventual OOM crashes.

## Fix

Changes the innermost per-element map from `Map<HTMLElement, Function>` to `WeakMap<HTMLElement, Function>`. This is type-safe since the key is always an `HTMLElement`. The `type`-keyed `Map` above it stays a `Map`, since `string` keys aren't valid for `WeakMap`.

One side effect: `removePatchedListener` used to check `element2patched.size === 0` to prune the now-empty `type` entry from `listener2Type`. `WeakMap` has no `.size`, so that pruning is gone. A small empty `WeakMap` shell is left per `(listener, type)` pair after its last element is removed instead of being deleted outright. It doesn't grow with element count and clears once `listener` becomes unreachable, since the outer `_wrappedListeners` is already a `WeakMap`.

## Testing

Added a regression test in `userInteraction.nozone.test.ts` that attaches a shared listener to 50 elements with no `removeEventListener` call, simulating the SPA pattern above, and asserts the per-element bookkeeping is a `WeakMap`. Fails against the old `Map`-based code, passes with this fix. Full existing suite (42 tests) still passes.

Fixes #3539

### DIFF
--- a/packages/instrumentation-user-interaction/src/instrumentation.ts
+++ b/packages/instrumentation-user-interaction/src/instrumentation.ts
@@ -52,7 +52,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
   // for addEventListener/removeEventListener state
   private _wrappedListeners = new WeakMap<
     Function | EventListenerObject,
-    Map<string, Map<HTMLElement, Function>>
+    Map<string, WeakMap<HTMLElement, Function>>
   >();
   // for event bubbling
   private _eventsSpanMap: WeakMap<Event, api.Span> = new WeakMap<
@@ -216,7 +216,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
     }
     let element2patched = listener2Type.get(type);
     if (!element2patched) {
-      element2patched = new Map();
+      element2patched = new WeakMap();
       listener2Type.set(type, element2patched);
     }
     if (element2patched.has(on)) {
@@ -245,12 +245,6 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
     const patched = element2patched.get(on);
     if (patched) {
       element2patched.delete(on);
-      if (element2patched.size === 0) {
-        listener2Type.delete(type);
-        if (listener2Type.size === 0) {
-          this._wrappedListeners.delete(listener);
-        }
-      }
     }
     return patched;
   }
--- a/packages/instrumentation-user-interaction/test/userInteraction.nozone.test.ts
+++ b/packages/instrumentation-user-interaction/test/userInteraction.nozone.test.ts
@@ -133,6 +133,42 @@ describe('UserInteractionInstrumentation', () => {
       assert.strictEqual(called, false);
     });
 
+    it('should not retain detached elements in listener bookkeeping when removeEventListener is never called', () => {
+      // Simulates the SPA framework pattern (Vue's createFnInvoker, React's
+      // bound handlers) where the same listener function is reused across
+      // many elements, and removeEventListener is never called on unmount -
+      // cleanup is expected to happen via garbage collection of the element
+      // instead. See open-telemetry/opentelemetry-js-contrib#3539.
+      const sharedListener = function () {};
+      const elements: HTMLElement[] = [];
+      for (let i = 0; i < 50; i++) {
+        const element = createButton();
+        elements.push(element);
+        element.addEventListener('click', sharedListener);
+      }
+
+      const instrumentationAny = userInteractionInstrumentation as any;
+      const wrappedListeners = instrumentationAny._wrappedListeners;
+
+      const listener2Type = wrappedListeners.get(sharedListener);
+      assert.ok(listener2Type, 'listener should be tracked');
+      const element2patched = listener2Type.get('click');
+      assert.ok(element2patched, 'click bookkeeping should exist');
+
+      // The bug: the per-element bookkeeping is a plain Map keyed by the
+      // HTMLElement itself, so every element added above is strongly
+      // retained here even though none of them were ever attached to the
+      // document and nothing else in the test references them anymore.
+      assert.ok(
+        element2patched instanceof WeakMap,
+        'per-element listener bookkeeping should be a WeakMap so detached ' +
+          'elements are not strongly retained (#3539)'
+      );
+
+      // Drop the only other reference we held.
+      elements.length = 0;
+    });
+
     it('should not double-register a listener', () => {
       let callCount = 0;
       const listener = function () {


david-luna (approved):

david-luna:
Thanks for working on this :)

PR #3577, opened 2026-06-22
feat(instrumentation-kafkajs): add messaging.kafka.cluster.id span attribute

### LINKED ISSUE #3622: feat(kafkajs): add messaging.kafka.cluster.id span attribute
## Problem

KafkaJS producer and consumer spans do not include `messaging.kafka.cluster.id`. When the same topic name exists across multiple clusters or environments, there is no way to distinguish which cluster a span came from.

## Expected behavior

`messaging.kafka.cluster.id` is now part of the [Kafka messaging semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md) with a **`Recommended`** requirement level (Development stability). Recommended attributes are emitted by default, so this attribute should be added to KafkaJS producer and consumer spans **by default** — no opt-in flag.

The cluster id is resolved from broker metadata via the Kafka client / AdminClient API (`admin().describeCluster()` in KafkaJS), fetched once per client and cached, so there is no per-message overhead. Resolution is best-effort: if the broker is unreachable or returns no id, the attribute is omitted.

## Context

Part of a cross-language effort to add this attribute consistently across OTel instrumentations. See [opentelemetry-java-instrumentation#18977](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/18977) for the cross-language tracking issue.

shashank-reddy-nr:

Kafka spans carry no information about which cluster a message came from. When the same topic name exists across environments, or when an application talks to more than one cluster, there is no way to tell which cluster a given span belongs to. `messaging.kafka.cluster.id` disambiguates them.

Adds the `messaging.kafka.cluster.id` attribute to KafkaJS producer and consumer spans.

This attribute is part of the [Kafka messaging semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/kafka.md) with a **`Recommended`** requirement level (Development stability). Per the requirement-level semantics, Recommended attributes are emitted **by default**, so this attribute is added automatically — there is no opt-in flag.

Implementation notes:

- The cluster id is read from the kafkajs-internal `cluster.brokerPool.metadata.clusterId` field. The `Cluster` instance is captured by hooking kafkajs's internal `producer/index.js` and `consumer/index.js` factory modules (via `InstrumentationNodeModuleFile`) and reading `params.cluster` off the call — those factories already receive the `Cluster` as a plain argument, so no `Kafka` instance is patched and there is no `AdminClient`, no extra connection, and no ACL requirement.
- Resolution is **best-effort and synchronous at span time**, read from a `WeakMap` keyed by the Producer/Consumer instance. For a consumer this is effectively guaranteed by the time `eachMessage`/`eachBatch` runs, since kafkajs refreshes metadata on `connect()` and at the top of every `fetch()` cycle before either callback can fire. For a producer it is not guaranteed on the very first `send()`/`sendBatch()` of a freshly created producer, since that metadata is fetched lazily inside kafkajs's own send path; the attribute is simply omitted for that span and appears on subsequent sends once metadata has arrived. There is no per-message async overhead either way.
- Tested directly: no mutation of the `Kafka` instance while capturing, correct pass-through for a non-function module export, and the full capture → span-attribute path, in addition to producer/consumer/sendBatch/eachBatch coverage.

Fixes #3622


shashank-reddy-nr:
Hi @seemk Can you please review these code changes?

packages/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -151,6 +154,34 @@ function prepareDurationHistogram<T extends Attributes>(
 const HISTOGRAM_BUCKET_BOUNDARIES = [
   0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10,
 ];
+
+// Maps a Producer/Consumer to its kafkajs Cluster instance.
+const _clusterWeakMap = new WeakMap<object, unknown>();
```
david-luna: nit: If you explicitly set `cluster` object in this WeakMap if wonder why not give it a type instead of `unknown`. Is there any situation where the instrumentation could call `_clusterWeakMap.set(...)` with a value that is not a cluster from the params?

If possible, having a type here would simpliy the logic int he function below

```ts
interface ClusterType {
  brokerPool?: {
    metadata?: { clusterId?: string }
  }
}

const _clusterWeakMap = new WeakMap<object, ClusterType>();

function _readClusterId(client: unknown): string | undefined {
  if (client === null || client === undefined) return undefined;
  const cluster = _clusterWeakMap.get(client as object);

  // We could use optional chaining to get the property safely
  // ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
  const clusterId = cluster?.brokerPool?.metadata?.clusterId;

  return typeof clusterId === 'string' && clusterId !== ''
      ? clusterId
      : undefined;
}
```

packages/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -227,7 +292,7 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
         instrumentation._wrap(
           newConsumer,
           'run',
-          instrumentation._getConsumerRunPatch()
+          instrumentation._getConsumerRunPatch(newConsumer)
```
david-luna: question: IIUC the scope of `newConsumer.run` function is the consumer itself. So `this` should refer to the the `newConsumer` within the `run` function and in this scenario it is redundant to pass the consumer to the patcher function.

david-luna: Same happens below with the producer methods (sendbatch, send & transaction)

packages/instrumentation-kafkajs/src/instrumentation.ts
shashank-reddy-nr: Done — introduced a ClusterType interface (brokerPool?: { metadata?: { clusterId?: string } }) and retyped the WeakMap as WeakMap<object, ClusterType>. The read path now uses optional chaining (cluster?.brokerPool?.metadata?.clusterId) wrapped in a try/catch to guard against getters that throw (there's a test covering that case).

packages/instrumentation-kafkajs/src/instrumentation.ts
shashank-reddy-nr: Done — removed the kafkaClient parameter from `_getConsumerRunPatch`. The wrapper function already has this: Consumer, so `_getConsumerEachMessagePatch(this)` and `_getConsumerEachBatchPatch(this) `now receive the consumer directly via this. And also removed kafkaClient from `_getSendPatch`, `_getSendBatchPatch`, and `_getProducerTransactionPatch`. For the direct send/sendBatch case this is the Producer, which is already registered in `_clusterWeakMap`. For the transaction case, this inside the patched transaction.send/sendBatch is the Transaction object, not the Producer — so `_getProducerTransactionPatch` now pre-registers the transaction in `_clusterWeakMap` (copying the producer's cluster entry) before patching those methods, making `_readClusterId(this)` work uniformly in both cases.

shashank-reddy-nr:
Hi @david-luna I have addressed all your comments. Can you please take a look?

david-luna (approved):

david-luna:
I'll wait a bit before merge. Let's give @seemk more time to have the chance to review

seemk (approved):

seemk:
LGTM! Summer, so was away for a bit

PR #3581, opened 2026-06-27
chore(eslint): use canonical n/ prefix for eslint-plugin-n in flat config

bhuvan-somisetty:

The `eslint-plugin-n` package (the successor to the now-deprecated `eslint-plugin-node`) uses `n` as its canonical rule namespace in flat config. The current `eslint.config.mjs` still registers the plugin under the legacy `node:` key and references rules with the `node/` prefix, which is inconsistent with the package's actual name and its flat-config conventions.

- Rename the plugin registration from `node: nodePlugin` to `n: nodePlugin` so the namespace matches the package name
- Update `settings.node` to `settings.n` to align with the plugin's settings key
- Rename the `node/no-deprecated-api` rule reference to `n/no-deprecated-api`
- Update the commented-out `node/no-missing-import` placeholder to `n/no-missing-import` for consistency
- Remove the stale `// nodePlugin.configs['flat/recommended']` comment (the individual rule overrides in the `rules` block already express the intended configuration explicitly)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -43,9 +43,7 @@ const baseConfig = tseslint.config(
       'yet-another-license-header': yalhPlugin,
       n: nodePlugin,
     },
-    extends: [
-      eslint.configs.recommended,
-    ],
+    extends: [eslint.configs.recommended],
     languageOptions: {
       ecmaVersion: 2022,
       globals: {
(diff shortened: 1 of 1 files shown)


bhuvan-somisetty:
Hey @pichlermarc @trentm — would appreciate a review when you get a chance. This finishes the ESLint flat config migration by switching the `eslint-plugin-n` registration to the canonical `n:` namespace (the package was renamed from `eslint-plugin-node`). Should be a small, low-risk change.

bhuvan-somisetty:
Hey @david-luna - the unit tests are waiting on workflow approval before they can run. Could you approve the workflows when you get a chance? Thanks!

trentm (approved):

PR #3582, opened 2026-06-28
fix(instrumentation-koa): use fallback name for anonymous middleware spans

### LINKED ISSUE #2220: [Koa] Some koa middleware spans have weird data
We are seeing some spans emitted by the koa instrumentation that look like this:

```js
{
  attributes: {
      'koa.name': '',
      'koa.type': 'middleware',
    },
    name: 'middleware - ',
}
```

Which I guess comes from here: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-koa/src/utils.ts#L43

I guess there needs to be some more checking there, and/or some fallback behavior if `layer.name` is empty 🤔 

The app that triggers this span is basically:

```js
const Koa = require('koa');
const Router = require('@koa/router');

const app1 = new Koa();

router1.get('/test-transaction', ctx => {
  // do something here...

  ctx.body = {};
});

app1.use(router1.routes()).use(router1.allowedMethods());

app1.listen(port1);
```

And then visiting `/test-transaction`.

Note that the router span seems correct:

```js
{
  attributes: {
    'http.route': '/test-transaction',
    'koa.name': '/test-transaction',
    'koa.type': 'router',
  },
  name: 'router - /test-transaction',
}

### PR DESCRIPTION
**Pull request by bhuvan-somisetty**

Fixes #2220

When a Koa middleware is an anonymous function — the common case of passing an arrow function directly to `app.use()` — its `.name` property is an empty string in JavaScript (name inference only applies to variable assignments, not function arguments). This caused the instrumentation to emit spans with empty names:

```js
{
  name: 'middleware - ',
  attributes: { 'koa.name': '', 'koa.type': 'middleware' },
}
```

The root cause was two-fold in `utils.ts`:

1. The `koa.name` attribute used `layer.name ?? 'middleware'`. The nullish coalescing operator (`??`) only falls back for `null` / `undefined`, not for empty strings, so `''` was passed through unchanged.
2. The span name was built with `` `middleware - ${layer.name}` ``, producing a trailing space and nothing when the name is empty.

- In `getMiddlewareMetadata`, extract `layer.name || 'middleware'` into a local variable `middlewareName`. The logical OR operator treats empty strings as falsy, so anonymous middlewares receive the `'middleware'` fallback.
- Use `middlewareName` for both the `koa.name` attribute and the span name, giving `middleware - middleware` for anonymous middlewares.
- Add unit tests in `utils.test.ts` covering the empty-name fallback and the named-middleware path.
- Add an integration test in `koa.test.ts` verifying that an inline anonymous arrow function (which has `.name === ''`) produces a span named `middleware - middleware`.

### DIFF
--- a/packages/instrumentation-koa/test/koa.test.ts
+++ b/packages/instrumentation-koa/test/koa.test.ts
@@ -469,15 +469,15 @@ describe('Koa Instrumentation', function () {
           // rather than an empty 'middleware - ' span name.
           const anonymousSpan = memoryExporter
             .getFinishedSpans()
-            .find(span => span.name === 'middleware - middleware');
+            .find(span => span.name === 'middleware - anonymous');
           assert.notStrictEqual(
             anonymousSpan,
             undefined,
-            'Expected a span named "middleware - middleware" for the anonymous middleware'
+            'Expected a span named "middleware - anonymous" for the anonymous middleware'
           );
           assert.strictEqual(
             anonymousSpan?.attributes[AttributeNames.KOA_NAME],
-            'middleware'
+            'anonymous'
           );
           assert.strictEqual(
             anonymousSpan?.attributes[AttributeNames.KOA_TYPE],
--- a/packages/instrumentation-koa/test/koa.test.ts
+++ b/packages/instrumentation-koa/test/koa.test.ts
@@ -452,11 +452,17 @@ describe('Koa Instrumentation', function () {
 
     it('should use fallback name for anonymous middleware', async () => {
       const rootSpan = tracer.startSpan('rootSpan');
-      // Arrow functions passed as arguments have .name === '' in JS (no name inference).
-      // The context-setting middleware below is one such case.
-      app.use((_ctx, next) =>
+      // Sets rootSpan as active context for subsequent middlewares.
+      // This middleware itself runs with no parent so it won't get a span.
+      app.use((ctx, next) =>
         context.with(trace.setSpan(context.active(), rootSpan), next)
       );
+      // Anonymous arrow function (no name inference for function arguments in JS).
+      // Runs after context is set, so the instrumentation will have a parent span
+      // and will create a span for it with the 'anonymous' fallback name.
+      app.use(async (_ctx, next) => {
+        await next();
+      });
       app.use(simpleResponse());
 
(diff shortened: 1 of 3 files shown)


bhuvan-somisetty:
This is a bug fix for clearly broken behavior (empty span names for anonymous middleware), not a new feature addition. Would appreciate a sponsor from @open-telemetry/javascript-approvers to review and merge. Happy to make any adjustments needed.

bhuvan-somisetty:
Hey @pichlermarc @trentm — would appreciate a review when you get a chance. This fixes a real bug where anonymous Koa middleware (arrow functions passed directly to `app.use()`) produce span names like `middleware - ` and `koa.name: ''` because `layer.name` is an empty string and `??` doesn't catch that. The fix switches to `||` with a `'middleware'` fallback.

packages/instrumentation-koa/test/koa.test.ts
```diff
@@ -450,6 +450,43 @@ describe('Koa Instrumentation', function () {
...
+        trace.setSpan(context.active(), rootSpan),
+        async () => {
+          await httpRequest.get(`http://localhost:${port}`);
+          rootSpan.end();
+
+          // With the fix, anonymous middlewares must produce the fallback name
+          // rather than an empty 'middleware - ' span name.
+          const anonymousSpan = memoryExporter
+            .getFinishedSpans()
+            .find(span => span.name === 'middleware - middleware');
```
david-luna: this is just a preference but why not `middleware - anonymous`?

bhuvan-somisetty: Good call - updated to use `'anonymous'` as the fallback. Span name is now `'middleware - anonymous'` and `koa.name` attribute is `'anonymous'`. Updated in e8ad68ef.

bhuvan-somisetty:
Hey @david-luna @pichlermarc @trentm - addressed the feedback, updated the fallback to `'anonymous'` as suggested. Ready for another look when you get a chance.

bhuvan-somisetty:
Hey @david-luna @pichlermarc @trentm — fixed the test structure and prettier formatting issues. Ready for another look.

david-luna (approved):

Issue #1705, opened 2023-09-30
pg: Prepared statements cannot be used with sqlCommenter instrumentation


Subv:
### What version of OpenTelemetry are you using?
```
"@opentelemetry/auto-instrumentations-node": "0.39.2",
"@opentelemetry/sdk-node": "0.43.0",
"pg": "8.11.3",
```

### What version of Node are you using?
18.16.0

### What did you do?

NestJS project with https://github.com/madeindjs/nestjs-graphile-worker v0.3.1 which under the hood uses graphile-worker 0.13.0 (https://github.com/madeindjs/nestjs-graphile-worker/blob/c956c3c2850b3cde2fd2976af3e8a8c0b314b1d1/package.json#L64) with the following otelSDK configuration:

```ts
const otelSDK = new NodeSDK({
  spanProcessor: new BatchSpanProcessor(new ConsoleSpanExporter()),
  textMapPropagator: new CompositePropagator({
    propagators: [
      new JaegerPropagator(),
      new W3CTraceContextPropagator(),
      new W3CBaggagePropagator(),
      new B3Propagator(),
      new B3Propagator({
        injectEncoding: B3InjectEncoding.MULTI_HEADER,
      }),
    ],
  }),
  instrumentations: [
    getNodeAutoInstrumentations({
      '@opentelemetry/instrumentation-fs': {
        enabled: false,
      },
      '@opentelemetry/instrumentation-pg': {
        addSqlCommenterCommentToQueries: true,
      },
    }),
  ],
});
```

### What did you expect to see?
The graphile-worker library is a postgres-backed job queue library which by default uses prepared statements under the hood to obtain the next job to execute ( see https://github.com/graphile/worker/blob/a6a52b4948e1f85d71211ff516b4f061a71f2167/src/worker.ts#L144 ).
I expected the library to just work after enabling the sqlCommenter instrumentation.

### What did you see instead?
The call to get_job fails with the following error message:
`Prepared statements must be unique - 'get_job/graphile_worker' was used for a different statement`

See trace:
```
{
   traceId: '17c870761a1bc2a549e1bc88dccf451e',
   parentId: undefined,
   traceState: undefined,
   name: 'pg.query:get_job/graphile_worker <SNIP>',
   id: '73b38f695f764bef',
   kind: 2,
   timestamp: 1696110796199000,
   duration: 753.917,
   attributes: {
     'db.system': 'postgresql',
     'db.name': '<SNIP>',
[...]

PR #3589, opened 2026-07-02
fix(instrumentation-pg): Only treat "name" as relevant if it actually is a value

blast-hardcheese:
Related: https://github.com/drizzle-team/drizzle-orm/pull/5974 and https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1705

Thank you tremendously for your work curating and maintaining the OTEL ecosystem, it has been a joy to use!

- `instrumentation-pg` gates adding sql tracing comments on the presence of the `name` field, not the `name` field actually being defined. This means tooling that mechanically and indiscriminately populate that field make automatic traceparent comment injection impossible.

- Simply change `!('name' in arg0)` into `!('name' in arg0) || arg0.name === undefined`, thus permitting the semantic intent even if the implementation of some libraries is a little fast and loose.


maryliag (approved):

blast-hardcheese:
Thank you!

Issue #3590, opened 2026-07-02
remove metric `db.client.connections.usage`


maryliag:
When updating to only send stable DB semconv, `db.client.connections.usage` should be removed. Today the emission is not gated on semconv stability.
Ideally the replacement for it (`db.client.connection.count`) should be sent instead (on places that an experimental metric can be sent, otherwise don't sent it yet, until there is a way to opt-in for experimental)

Dependency on https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3585

Luffy-nani:
Hey @maryliag! 
I took a shot at fixing this issue and opened PR #3592.

I'd love to get your feedback! Please take a look at the PR when you have a chance and let me know if there are any changes or improvements you'd like to see. Happy to iterate!

maryliag:
hey @Luffy-nani, thank you for your interest! I added some comments to the PR, that was not quite what I was looking for with this issue. I added the dependency PR to the description, so hopefully is more clear.

For now, let's put this one on hold, until we decide clearly the next steps.

Luffy-nani:
@maryliag  Thanks for clarifying. I'd implemented it based on how the original issue described the fix (gating the metric behind the semconv opt-in), so that's where the approach came from. I'll check the updated description for the dependency PR link now that it's added.

maryliag:
Hey @Luffy-nani , that PR is now merged, so let me know if you still want to take this issue, which would be the removal of `db.client.connections.usage` while also adding `db.client.connection.count`

Luffy-nani:
@maryliag, Yes, Id love to take this on
I will start working on it

Luffy-nani:
Hey @maryliag 
before I start, want to confirm one thing. Looking at instrumentation-pg, it emits db.client.connection.count unconditionally (no gating behind OTEL_SEMCONV_STABILITY_OPT_IN). Should Mongo/MySQL follow that same unconditional pattern, or is there gating logic you have in mind that I should implement instead? Just want to make sure I match the right approach before diving in.

maryliag:
You can follow the same pattern as PG and always send them

Luffy-nani:
Hey @maryliag
opened PR : #3608 . Replaced db.client.connections.usage with db.client.connection.count in both Mongo and MySQL, sent unconditionally to match instrumentation-pg's pattern 
Let me know if anything needs adjusting!

pichlermarc:
[...]

PR #3592, opened 2026-07-02
fix: only emit db.client.connections.usage under old database semconv

Luffy-nani:

The db.client.connections.usage metric is part of the old (pre-stable) OpenTelemetry database semantic conventions and has been deprecated in favor of db.client.connection.count. However, the MongoDB and MySQL instrumentations currently emit this deprecated metric regardless of the OTEL_SEMCONV_STABILITY_OPT_IN configuration. This means that even when users explicitly opt into the stable database semantic conventions by setting OTEL_SEMCONV_STABILITY_OPT_IN=database, they still receive a deprecated metric, which is inconsistent with the expected behavior. This PR fixes that issue by ensuring the deprecated metric is only emitted when the old database semantic conventions are active.

Solves #3590 

This PR updates both the MongoDB and MySQL instrumentations so that db.client.connections.usage is created only when the old database semantic conventions are enabled. As a result, the metric is emitted when OTEL_SEMCONV_STABILITY_OPT_IN does not include database or includes database/dup, and it is no longer emitted when users opt into only the stable database semantic conventions (OTEL_SEMCONV_STABILITY_OPT_IN=database). The _updateMetricInstruments() method was updated in both instrumentations to implement this behavior. In addition, the README files were updated to clearly document when the metric is emitted, and new tests were added for the MongoDB instrumentation to verify that the deprecated metric is not emitted when only the stable database semantic conventions are enabled.


maryliag:
@Luffy-nani, as you noticed on the readme that you updated, it's on purpose that we're sending this metric always, since there is no stable replacement, so your suggestion to add behind the flag is not what we want.
The issue was created as a follow up from this [PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3585/changes#diff-b092d92207f24a23754646898885510628582f73fb356a719e81f14441b3bf75R211), which is still open. So only after that PR is merged, that we should check if we should just remove it or already replaced with the experimental option.

A comment in general for future PRs: you should not be adding a new file to test a specific case like you did for mongo, it should be added to already existing files. And you didn't add tests for the mysql package, and we always expect tests to be added or updated with code changes!

So for this one, we can go ahead and close it!

Luffy-nani:
Thanks for the context, @maryliag..that makes sense, I wasn't aware of #3585 or that the always-on behavior was intentional pending a stable replacement. I'll close this out and revisit if/when there's a follow-up issue after that PR merges

PR #3598, opened 2026-07-03
feat(instrumentation-redis): support redis@6

trentm:
This does the minimum: bump current monkey-patching to patch v6.
I haven't dug into whether there are architectural changes in v6
that could/should be adapted to.

For example, https://github.com/redis/node-redis/pull/3195 added
diagnostics channel support to redis, such that instrumentation
could be done without monkeypatching. This is worth considering.


david-luna (approved):
Although TAV already covers the new version. Have you considered updating the redis dependency to 6?

PR #3600, opened 2026-07-04
chore: remove examples using ancient sdk-trace-base@0.x

trentm:
Many (most) examples are not well maintained. As an easy start, this
removes any examples that are so old they are using an sdk-trace-base
package version that is *two* major versions behind.

I'm advocating removing these with the argument that
incorrect docs/examples are worse that no docs/examples.


pichlermarc (approved):
thanks!

maryliag (approved):

PR #3601, opened 2026-07-04
fix(auto-instrumentations-node): avoid duplicate diag logger warning in register.js

### LINKED ISSUE #2742: How to use suppressOverrideMessage flag with @opentelemetry/auto-instrumentations-node
I noticed the addition of the `suppressOverrideMessage` flag in [#3366](https://github.com/open-telemetry/opentelemetry-js/pull/3366) but I'm confused as to how to use it with `@opentelemetry/auto-instrumentations-node` and pino.

Can someone please share an example of how to use it?

- [x] This may affect other libraries, but I would like to get opinions here first

bhuvan-somisetty:

- When using `@opentelemetry/auto-instrumentations-node/register` together with a manually configured diag log level (`OTEL_LOG_LEVEL`), the process prints a `Current logger will be overwritten` warning (with a full stack trace) on every single start, as reported in #2742.

  This happens because `register.js` registers a `DiagConsoleLogger` based on `OTEL_LOG_LEVEL` and then constructs `NodeSDK`, which registers an equivalent `DiagConsoleLogger` again internally based on the same env var. The second registration doesn't pass `suppressOverrideMessage`, so `diag.setLogger()` logs the override warning even though nothing is actually going wrong.

- Remove the redundant diag logger setup in `register.js`. `NodeSDK` already reads `OTEL_LOG_LEVEL` and registers a `DiagConsoleLogger` on its own, so `register.js` no longer needs to do it a second time.
- Add a regression test asserting that starting via `register.js` with `OTEL_LOG_LEVEL` set does not log the `Current logger will be overwritten` warning.

Fixes #2742

### DIFF
--- a/packages/auto-instrumentations-node/src/register.ts
+++ b/packages/auto-instrumentations-node/src/register.ts
@@ -3,20 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import * as opentelemetry from '@opentelemetry/sdk-node';
-import { diag, DiagConsoleLogger } from '@opentelemetry/api';
-import { getStringFromEnv, diagLogLevelFromString } from '@opentelemetry/core';
+import { diag } from '@opentelemetry/api';
 import {
   getNodeAutoInstrumentations,
   getResourceDetectorsFromEnv,
 } from './utils';
 
-const logLevel = getStringFromEnv('OTEL_LOG_LEVEL');
-if (logLevel != null) {
-  diag.setLogger(new DiagConsoleLogger(), {
-    logLevel: diagLogLevelFromString(logLevel),
-  });
-}
-
+// `NodeSDK` already registers a `DiagConsoleLogger` based on `OTEL_LOG_LEVEL`
+// internally, so doing it here as well only results in the "Current logger
+// will be overwritten" warning being logged on every start.
 const sdk = new opentelemetry.NodeSDK({
   instrumentations: getNodeAutoInstrumentations(),
   resourceDetectors: getResourceDetectorsFromEnv(),
--- a/packages/auto-instrumentations-node/src/register.ts
+++ b/packages/auto-instrumentations-node/src/register.ts
@@ -3,18 +3,33 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import * as opentelemetry from '@opentelemetry/sdk-node';
-import { diag } from '@opentelemetry/api';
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Hi @pichlermarc @david-luna @trentm, all checks are passing on this one - could you take a look when you have a moment? Would appreciate a review.

david-luna:
@bhuvan-somisetty 

The auto instrumentations register executes some logic before creating a new `NodeSDK` instance in https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/auto-instrumentations-node/src/register.ts#L20-L23 by calling `getNodeAutoInstrumentations` and `getResourceDetectorsFromEnv`

These functions make use of the logger so if we remove the logger setup they will not be honouring the configuration set by the user. I think what we can do is to disable the logger before creating the SDK instance. Something like

```js
const logLevel = getStringFromEnv('OTEL_LOG_LEVEL');
if (logLevel != null) {
  diag.setLogger(new DiagConsoleLogger(), {
    logLevel: diagLogLevelFromString(logLevel),
  });
}

const instrumentations = getNodeAutoInstrumentations(),
const resourceDetectors = getResourceDetectorsFromEnv(),
if (logLevel != null) {
  diag.disable();
}

  instrumentations,
  resourceDetectors,
});
```

WDYT?

bhuvan-somisetty:
Closing and reopening to retrigger CI, one of the checks appears stuck.

bhuvan-somisetty:
@david-luna good catch, thanks! You're right, removing the logger setup entirely would have meant `getNodeAutoInstrumentations()` and `getResourceDetectorsFromEnv()` run without honoring `OTEL_LOG_LEVEL`.

Pushed a new commit that follows your suggestion: set up the `DiagConsoleLogger` first, run those two calls with it active, then call `diag.disable()` right before constructing `NodeSDK` so its internal `diag.setLogger()` doesn't have anything to warn about overwriting.

CI is green on the latest commit, let me know if this looks good.

cc @pichlermarc @trentm as well, would appreciate a look whenever you have time.

david-luna (approved):
LGTM. Thanks for working on this :)

bhuvan-somisetty:
Thanks for the review @david-luna, appreciate the help getting this right!

bhuvan-somisetty:
@pichlermarc @trentm this has an approval from david-luna now, all checks are green - would appreciate a review from either of you when you get a chance.

pichlermarc (approved):

PR #3602, opened 2026-07-04
fix(instrumentation-connect): avoid unconditional close listener causing MaxListenersExceededWarning

Luffy-nani:

`instrumentation-connect` attaches a `close` listener to the response on **every middleware layer**, even when the layer calls `next()` synchronously and the listener is never actually needed. In apps with many chained middlewares, this stacks up listeners faster than Node's default `MaxListeners` (10), triggering `MaxListenersExceededWarning` on every request.

- Only attach the fallback `close` listener *after* the middleware returns, and only if the span hasn't already finished via a synchronous `next()`. This mirrors the same fix already applied to `instrumentation-router` (#3495) and `instrumentation-express`.
- For the common case (synchronous middleware), zero listeners get attached at all no more warning, no behavior change for tracing.
- Added a test that chains 15 synchronous middleware layers and asserts no `close` listeners accumulate.


david-luna (approved):

david-luna:
Thanks for your contribution @Luffy-nani :)

PR #3603, opened 2026-07-06
test(instrumentation-aws-sdk): fix assertions that can never fail

### LINKED ISSUE #3529: instrumentation-aws-sdk: three test cases silently pass on the failure paths they intend to verify

`@opentelemetry/instrumentation-aws-sdk@0.73.0` (current `main`, observed at commit `15ef7506553f631ea4181391e0c5725a56f0d082`).

Node 18.x and 20.x (the package's declared engines range; the issues are static and node-version-independent).

### What did you do?

Three test cases in `packages/instrumentation-aws-sdk/test/` are written so that the assertion they appear to make can never fail — the test will report success even when the behavior under test is broken:

**1. `test/stepfunctions.test.ts` — `expect()` calls with no matcher chained**

Two assertions check whether a semantic-convention key is present on a span's attributes, but the `expect(...)` call evaluates the boolean and is then discarded — no `.toBe(true)` or equivalent is chained, so it asserts nothing.

```ts
// L56–L58 of stepfunctions.test.ts
expect(
  ATTR_AWS_STEP_FUNCTIONS_STATE_MACHINE_ARN in
    stateMachineAttributeSpan.attributes
);
```

```ts
// L92–L94 of stepfunctions.test.ts
expect(
  ATTR_AWS_STEP_FUNCTIONS_ACTIVITY_ARN in activityAttributeSpan.attributes
);
```

If the attribute is removed from the instrumentation, the test continues to pass. (The follow-up assertion on the value also throws a `TypeError` rather than the expected matcher failure if the key is missing, but the presence check itself is a no-op.)

**2. `test/aws-sdk-v3-s3.test.ts` — error-path assertions live only inside `catch`**

The `'aws error'` test wraps an `s3Client.putObject(...)` it expects to fail in a `try { … } catch { … }` with all assertions inside the `catch` block:

```ts
// aws-sdk-v3-s3.test.ts, "aws error"
try {
  await s3Client.putObject(params);
} catch {
  expect(getTestSpans().length).toBe(1);
  // …several more error-path expectations…
}
```

If a future change causes `putObject` to succeed (regression of error-handling, nock fixture drift, etc.), the `catch` block is never entered and the test still passes without making any assertion. A second occurrence with the same pattern appears later in the same file.

**3. `test/bedrock-runtime.test.ts` — silent early return when the streaming body is missing**

Six tests guard streaming-body consumption with `if (!response.body) return;`, which exits the test function — Mocha reads that as a successful completion:

```ts
// bedrock-runtime.test.ts, repeated at L661, L720, L773, L823, L873, L920
const response = await client.send(command);
let collectedText = '';
if (!response.body) return;
for await (const chunk of response.body) { … }
```

If the SDK starts returning an empty/undefined body (which is exactly the regression a streaming test should catch), every one of these tests silently passes.

Each test should fail when the behavior it describes is broken — either by replacing the no-op `expect(...)` with `.toBe(true)` (or just deleting the redundant presence check since the next assertion covers it), by asserting that the awaited call threw (e.g. `await assert.rejects(...)` or capturing the error and asserting in `finally`), and by turning `if (!response.body) return;` into `assert.ok(response.body);` (or equivalent matcher).

### What did you see instead?

All three tests report success on the failure modes they claim to verify.

### Additional context

Source links at the current `main`:

- [`stepfunctions.test.ts` L56–L58](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-aws-sdk/test/stepfunctions.test.ts#L56-L58)
- [`stepfunctions.test.ts` L92–L94](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-aws-sdk/test/stepfunctions.test.ts#L92-L94)
- [`aws-sdk-v3-s3.test.ts` `'aws error'` block](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-aws-sdk/test/aws-sdk-v3-s3.test.ts)
- [`bedrock-runtime.test.ts` L661, L720, L773, L823, L873, L920](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts)

bhuvan-somisetty:

- Fixes #3529. Three test cases in `packages/instrumentation-aws-sdk/test/` are written so that the assertion they appear to make can never fail, so the test reports success even when the behavior under test is broken:
  - `stepfunctions.test.ts`: two `expect(...)` calls with no matcher chained (`expect(KEY in attributes);`), so nothing is actually asserted.
  - `aws-sdk-v3-s3.test.ts`: the `'aws error'` and `'handle aws sdk exception'` tests put all their error-path assertions inside a `catch { ... }` block with nothing verifying the call actually threw. If a regression makes the call resolve instead of reject, the `catch` is never entered and the test still passes.
  - `bedrock-runtime.test.ts`: six streaming tests use `if (!response.body) return;`, which Mocha treats as a successful early return instead of a failure.

- `stepfunctions.test.ts`: removed the two no-op `expect(...)` calls, since the very next `expect(...).toBe(value)` in each test already covers both presence and value of the attribute.
- `aws-sdk-v3-s3.test.ts`: capture the thrown error in a variable and assert it's defined before checking the span attributes, so the test fails if the call unexpectedly succeeds.
- `bedrock-runtime.test.ts`: replaced the silent early return with `assert.ok(response.body)`, which both fails the test on a missing body and narrows the type for the `for await` loop that follows.

### DIFF
--- a/packages/instrumentation-aws-sdk/test/aws-sdk-v3-s3.test.ts
+++ b/packages/instrumentation-aws-sdk/test/aws-sdk-v3-s3.test.ts
@@ -136,31 +136,35 @@ describe('instrumentation-aws-sdk-v3 (client-s3)', () => {
         Key: 'aws-ot-s3-test-object.txt',
       };
 
+      let thrownError: any;
       try {
         await s3Client.putObject(params);
-      } catch {
-        expect(getTestSpans().length).toBe(1);
-        const [span] = getTestSpans();
-
-        // expect error attributes
-        expect(span.status.code).toEqual(SpanStatusCode.ERROR);
-        expect(span.status.message).toEqual('Access Denied');
-        expect(span.events.length).toBe(1);
-        expect(span.events[0].name).toEqual('exception');
-
-        expect(span.attributes[ATTR_RPC_SYSTEM]).toEqual('aws-api');
-        expect(span.attributes[ATTR_RPC_METHOD]).toEqual('PutObject');
-        expect(span.attributes[ATTR_RPC_SERVICE]).toEqual('S3');
-        expect(span.attributes[AttributeNames.AWS_S3_BUCKET]).toEqual(
-          'invalid-bucket-name'
-        );
-        expect(span.attributes[ATTR_HTTP_RESPONSE_STATUS_CODE]).toEqual(403);
-        expect(span.attributes[AttributeNames.CLOUD_REGION]).toEqual(region);
-        expect(span.attributes[AttributeNames.AWS_REQUEST_ID]).toEqual(
-          'MS95GTS7KXQ34X2S'
-        );
-        expect(span.name).toEqual('S3.PutObject');
+      } catch (err) {
+        thrownError = err;
       }
+      expect(thrownError).toBeDefined();
+
+      expect(getTestSpans().length).toBe(1);
+      const [span] = getTestSpans();
+
+      // expect error attributes
+      expect(span.status.code).toEqual(SpanStatusCode.ERROR);
+      expect(span.status.message).toEqual('Access Denied');
+      expect(span.events.length).toBe(1);
+      expect(span.events[0].name).toEqual('exception');
+
+      expect(span.attributes[ATTR_RPC_SYSTEM]).toEqual('aws-api');
+      expect(span.attributes[ATTR_RPC_METHOD]).toEqual('PutObject');
+      expect(span.attributes[ATTR_RPC_SERVICE]).toEqual('S3');
+      expect(span.attributes[AttributeNames.AWS_S3_BUCKET]).toEqual(
+        'invalid-bucket-name'
+      );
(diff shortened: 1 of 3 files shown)


bhuvan-somisetty:
@pichlermarc @david-luna @trentm all checks are green on this one, could one of you take a look when you get a chance? Would appreciate a review.

bhuvan-somisetty:
@david-luna  @trentm - friendly bump on this one whenever you have a moment - all checks are still green.

bhuvan-somisetty:
@blumamir @jj22ee @trivikr @david-luna @pichlermarc @trentm

Following up on this one - it's a small test-only change (43 lines) in instrumentation-aws-sdk, fixing assertions that could never fail. All checks are green.

No rush at all, I know review queues are long - just flagging it in case it slipped by. Happy to make any changes if something needs adjusting.

jj22ee (approved):

PR #3604, opened 2026-07-07
fix(instrumentation-mongodb): fix db.client.connections.usage metric drift

### LINKED ISSUE #3482: [instrumentation-mongodb] db.client.connections.usage gauge drifts upward over time vs real socket count
### What happened

`@opentelemetry/instrumentation-mongodb@0.66.0` reports `db.client.connections.usage{state="idle"}` values that diverge from the actual number of TCP connections to MongoDB over the lifetime of a long-running process. On a pod that had been up ~2 days, the metric reported ~900 idle connections for a single pool, while `/proc/net/tcp` on the same pod showed ~32 established sockets to the MongoDB endpoint.

### Environment
- `@opentelemetry/instrumentation-mongodb`: `0.66.0`
- `mongodb` driver: `6.x`
- Node.js: `22.x`
- Single `MongoClient` per process (via TypeORM's `mongodb` driver)
- OTel SDK (nodejs): `2.6.0`
- Long-running process (days)

### Expected
`db.client.connections.usage{state="idle"} + state="used"` reflects the current connection pool size for the pool, and decreases when the driver closes/prunes idle connections.

### Actual
The gauge only decreases via `checkOut`/`checkIn` paths. When the driver internally closes connections (idle timeout, pool prune, topology change, network error, `serverClosed`, `connectionPoolCleared`), the `idle` bucket is not decremented. The delta accumulates, so the gauge drifts well above the real socket count over time.

### Suggested cause

In `packages/instrumentation-mongodb/src/instrumentation.ts`, the pool metric is updated on patches around `ConnectionPool.checkOut` and session end/checkIn:

```
checkOut (new)  : +1 used
checkOut (reuse): -1 idle, +1 used
checkIn         : -1 used, +1 idle
```

There is no subscription to the driver's pool-monitoring events for `connectionClosed` / `connectionPoolCleared` / `connectionCheckOutFailed`, so connections reclaimed by the driver aren't reflected.

### Repro sketch
1. Run a Node service using `mongodb@6.x` with `@opentelemetry/instrumentation-mongodb@0.66.0` registered.
2. Configure the pool so `maxIdleTimeMS` is non-zero (or rely on driver pool maintenance) so the driver actively closes connections.
3. Run moderate steady traffic for several hours / a few days.
4. Compare `db.client.connections.usage{state="idle"}` to the real socket count from the pod (`ss -tn '( dport = :27017 )' | wc -l` or `/proc/net/tcp`).

### Why it matters

Operators use this metric to size pools and detect connection leaks. A drifting gauge triggers false-positive alerts and drives real configuration changes in the wrong direction.

Happy to contribute a PR that subscribes to `connectionClosed`, `connectionPoolCleared`, and `connectionCheckOutFailed` events from the MongoDB driver's CMAP event emitter.

Adii-45:
Fixes #3482

The `db.client.connections.usage` metric was derived from logical session lifecycle events instead of the MongoDB driver's actual connection pool (CMAP) events. As a result, the reported connection counts could drift upward over time and diverge from the real socket count, particularly for long-running processes.

This PR updates the metric implementation to track the actual connection pool lifecycle, ensuring the reported values stay aligned with the driver's real connection state.

- Replace session-based connection usage tracking with CMAP connection lifecycle events.
- Remove obsolete session-based metric tracking and related internal types.
- Update MongoDB metrics integration tests to validate the new behavior.
- Remove the obsolete unit test that only covered the removed implementation.

### Validation

- ✅ Unit tests pass
- ✅ MongoDB metrics integration tests pass
- ✅ TypeScript checks pass
- ✅ ESLint passes

### DIFF
--- a/packages/instrumentation-mongodb/src/instrumentation.ts
+++ b/packages/instrumentation-mongodb/src/instrumentation.ts
@@ -34,7 +34,6 @@ import {
 import { MongoDBInstrumentationConfig, CommandResult } from './types';
 import {
   CursorState,
-  ServerSession,
   MongodbCommandType,
   MongoInternalCommand,
   MongodbNamespace,
@@ -42,9 +41,10 @@ import {
   WireProtocolInternal,
   V4Connection,
   V4ConnectionPool,
+  V4ConnectionPoolConstructor,
+  V4ConnectionPoolEvent,
   Replacer,
 } from './internal-types';
-import { V4Connect, V4Session } from './internal-types';
 /** @knipignore */
 import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 
@@ -55,7 +55,10 @@ const DEFAULT_CONFIG: MongoDBInstrumentationConfig = {
 /** mongodb instrumentation plugin for OpenTelemetry */
 export class MongoDBInstrumentation extends InstrumentationBase<MongoDBInstrumentationConfig> {
   declare private _connectionsUsage: UpDownCounter;
-  declare private _poolName: string;
+  // Cleanup callbacks for CMAP listeners attached to live ConnectionPool
+  // instances, invoked on unpatch so disable() doesn't leave dangling
+  // listeners on pools that outlive the instrumentation.
+  private _connectionPoolCleanups = new Set<() => void>();
 
   constructor(config: MongoDBInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, { ...DEFAULT_CONFIG, ...config });
@@ -91,7 +94,6 @@ export class MongoDBInstrumentation extends InstrumentationBase<MongoDBInstrumen
       v3UnpatchConnection: v3UnpatchConnection,
     } = this._getV3ConnectionPatches();
 
-    const { v4PatchConnect, v4UnpatchConnect } = this._getV4ConnectPatches();
     const {
       v4PatchConnectionCallback,
       v4PatchConnectionPromise,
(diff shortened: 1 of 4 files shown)


Adii-45:
Hey @pichlermarc @david-luna @trentm, 

This PR fixes the `db.client.connections.usage` metric drift reported in #3482 by moving the metric from logical session lifecycle events to the MongoDB driver's actual CMAP connection lifecycle. The goal is to keep the reported connection counts aligned with the real connection pool state, especially for long-running processes!

This is my first contribution to the `opentelemetry-js-contrib` repository, and I've really enjoyed working through the codebase while putting this fix together!

I'm not entirely sure who the right reviewers are for this package, so I hope I haven't tagged the wrong people. Whenever you get a chance, I'd really appreciate a review and any feedback or suggestions you might have!

Looking forward to contributing more to the repository.. Thanks!

maryliag:
Hi @Adii-45 , thank you for your contribution!
About the issue: `db.client.connections.usage` is actually deprecated and will be removed completely, so it doesn't make sense to fix it now. It's being replaced with `db.client.connection.count`, per issue https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3590
When in doubt or if need more clarification, you can always ask on the issue itself to be assigned before starting to work on it, which will give maintainers an opportunity to check if the issue still makes sense.

I closed the issue, since is no longer valid. If the person that wanted to work on https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3590 doesn't reply in a few days, feel free to work on that one instead. For now I'll close this PR.

To answer your question about tagging the right people. There is no need to tag anyone, if you noticed, the approvers team and code owners are automatically tagged as reviewers, so you don't need to tag anyone. If you don't get any reviews after a few days, that you can tag people like you did in the comments, but give us a little time to get to it 😄

Adii-45:
Thanks for the clarification, @maryliag! That makes sense. I wasn't aware that this metric was already deprecated and being replaced, so I appreciate the explanation.

Thanks as well for the guidance about checking on issues beforehand and for explaining how the review assignment works. I'll keep that in mind for future contributions.

I'll keep an eye on #3590, and if it becomes available I'd be happy to work on it! Looking forward to contributing more!

Issue #3545, opened 2026-05-25
[instrumentation-typeorm] support typeorm 1.x


trentm:
typeorm@1.0.0 was recently released: https://github.com/typeorm/typeorm/releases/tag/1.0.0

```
'1.0.0': '2026-05-19T11:14:59.730Z',
```

Support for instrumenting it will have to be explicitly added.

---

Note that test-all-versions (TAV) tests (https://github.com/open-telemetry/opentelemetry-js-contrib/actions/workflows/test-all-versions.yml) have been failing since the typeorm@1.0.0 release (e.g. search for "fatal" in https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/26385848413/job/77664711317) because (a) the instrumentation, by design, does not instrument 1.x version, but (b) the ".tav.yml" configuration *does*.

PR #3612, opened 2026-07-08
feat(instrumentation-typeorm): add support for typeorm@1.x

katoztmy:
## Summary

- Extend `supportedVersions` from `<1` to `<2` in all `InstrumentationNodeModuleFile` / `InstrumentationNodeModuleDefinition` definitions so the instrumentation activates for typeorm 1.x
- Guard EntityManager method wrapping with an existence check to handle methods removed in typeorm 1.x (`findByIds` was deleted in 1.0); other methods and the `connection` getter (now a deprecated alias for `dataSource`) remain compatible
- Update `.tav.yml` to split into two jobs: 0.3.x range unchanged, 1.x range with `node: '>=20'` constraint (typeorm 1.0 requires Node 20+) and `better-sqlite3` as a peer dependency (typeorm 1.x removed the `sqlite` driver)
- Update test helpers to auto-detect the available SQLite driver (`sqlite` for 0.3.x, `better-sqlite3` for 1.x) so the same test suite runs against both supported version ranges without changes to individual test files

## Root cause of the TAV failure (Refs #3545)

typeorm 1.0 removed `EntityManager.prototype.findByIds`. When the instrumentation attempted to wrap it via `shimmer.wrap`, it threw `"Attempt to wrap undefined property findByIds as function"`, crashing the patch phase entirely. The fix is a one-line existence guard before the wrap call.

The `connection` property (also listed as a breaking change) was kept as a deprecated getter returning `this.dataSource` in typeorm 1.x, so no changes to attribute collection code were necessary.

## Test plan

- [x] `npm test` passes (13/13) with typeorm@0.3.x locally
- [ ] TAV CI (`test-all-versions` job in `test.yml`) passes for both the `>=0.3.0 <1` and `>=1.0.0 <2` ranges on Ubuntu — verifying typeorm@1.x with `better-sqlite3`

## Compatibility

- typeorm 0.3.x: no behaviour change; all existing tests pass
- typeorm 1.x: `findByIds` is silently skipped (method no longer exists); all other EntityManager, SelectQueryBuilder, and DataSource patches apply normally

Refs #3545

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/instrumentation-typeorm/.tav.yml
+++ b/packages/instrumentation-typeorm/.tav.yml
@@ -1,5 +1,12 @@
 typeorm:
-  versions:
-    include: ">=0.3.0 <1"
-    mode: max-7
-  commands: npm run test
+  jobs:
+    - versions:
+        include: ">=0.3.0 <1"
+        mode: max-7
+      commands: npm run test
+    - versions:
+        include: ">=1.0.0 <2"
+        mode: max-7
+      node: '>=20'
+      peerDependencies: "better-sqlite3@^11.0.0"
+      commands: npm run test
--- a/packages/instrumentation-typeorm/src/typeorm.ts
+++ b/packages/instrumentation-typeorm/src/typeorm.ts
@@ -70,14 +70,16 @@ const functionsUsingQueryBuilder: EntityManagerMethods[] = [
   'count',
   'find',
   'findAndCount',
-  'findByIds',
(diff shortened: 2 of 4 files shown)


katoztmy:
Closing — submitted prematurely by mistake. Will reopen when ready.

PR #3616, opened 2026-07-09
fix(lint:deps): fix knip failure

### LINKED ISSUE #3615: lint:deps (knip) fails with "ts.getDefaultLibFilePath is not a function" — knip's unbounded typescript peer dependency resolves to TS 7.0.2

this is a dev-tooling issue in the repo itself (knip, via the lint:deps script), not a bug in a published OpenTelemetry package.

v24.18.0

### What did you do?

Ran `npm run lint:deps` (underlying command: `npx --yes knip@5.33.3 --dependencies --production --tags=-knipignore`).

The dependency-checking script (knip) to run successfully and report any unused/unlisted dependencies, or exit cleanly if none are found.

### What did you see instead?

TypeError: ts.getDefaultLibFilePath is not a function
at .../knip/dist/typescript/create-hosts.js:7:37
at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
at async node:internal/modules/esm/loader:643:26
at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

### Additional context

This isn't isolated to one PR. I've seen it fail on recent, unrelated PRs as well.

**Root cause:** `knip@5.33.3`'s `package.json` declares `"typescript": ">=5.0.4"` as a peer dependency with no upper bound. Since `lint:deps` runs `npx --yes knip@5.33.3 ...`, npm resolves the newest available `typescript` satisfying that range on every fresh invocation currently `typescript@7.0.2`, a very recent release that appears to have changed or removed the `ts.getDefaultLibFilePath` API knip's internals depend on.

This is unrelated to the repo's own pinned `typescript@5.0.4` devDependency  the crash happens during npx's own isolated, unpinned resolution, separate from the project's lockfile.

I confirmed the same crash occurs with older knip versions too (tested 5.30.0), so this isn't fixable by just bumping/pinning a different knip version  any version with this open-ended peer range will eventually hit the same problem as new TypeScript releases ship.

**What I tested:**

- `npx --yes -p typescript@5.0.4 -p knip@5.33.3 knip ...` (explicitly pinning typescript alongside knip in the same npx call) — worked in one environment (Node 22), but still crashed in another (Node 24, this machine). Not reliable enough to recommend as-is.
- Installing knip + typescript@5.0.4 as real, lockfile-pinned devDependencies (not fetched fresh via npx --yes each time)  this reliably avoided the crash and ran a correct analysis end-to-end (verified it correctly flags an actually-unused dependency in a test project, exit code 1 as expected).

**Possible fix:** Add knip as a real devDependency in the root package.json, and change the script from `npx --yes knip@5.33.3 ...` to just `knip ...`, so resolution is locked via package-lock.json instead of re-resolved fresh (and unpinned) on every run.

 happy to submit the fix if this is the preferred direction, or if there's a different approach preferred.

trentm:
This is also an attempt to mitigate this lint failure being hit in CI,
but not locally on macOS:

```
> opentelemetry-js-contrib@0.1.0 lint:deps
> npx --yes knip@5.33.3 --dependencies --production --tags=-knipignore

npm warn deprecated @snyk/github-codeowners@1.1.0: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
file:///home/runner/.npm/_npx/ac0927abb2277a89/node_modules/knip/dist/typescript/create-hosts.js:7
const libLocation = path.dirname(ts.getDefaultLibFilePath({}));
                                    ^

TypeError: ts.getDefaultLibFilePath is not a function
    at file:///home/runner/.npm/_npx/ac0927abb2277a89/node_modules/knip/dist/typescript/create-hosts.js:7:37
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:106:12)

Node.js v18.20.8
Error: Process completed with exit code 1.
```

Fixes: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3615


packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -55,7 +55,7 @@ import {
   MESSAGING_OPERATION_TYPE_VALUE_PROCESS,
   MESSAGING_SYSTEM_VALUE_AWS_SQS,
 } from './semconv';
-import {
+import type {
```
trentm, line 58: The new knip@6 complained about this `aws-lambda` dep. Using explicit `import type ...` here lets knip know that this is using the `@types/aws-lambda` dep.

pichlermarc (approved):

PR #3617, opened 2026-07-09
chore: update to use shared workflow

maryliag:
Update the workflow to point to the [Shared Workflow for Survey](https://github.com/open-telemetry/shared-workflows/tree/main/survey-on-merged-pr)


pichlermarc:
lint step is failing due to #3615; #3616 is already merged, merging `main` should help make the workflow pass.

pichlermarc (approved):

maryliag:
The similar one got merged on core, but it made post a message when it shouldn't. I'm looking into a fix, then I'll update with the latest version here before merging. Will also merge with main before that

PR #3619, opened 2026-07-11
feat(instrumentation-typeorm): add support for typeorm@1.x

### LINKED ISSUE #3545: [instrumentation-typeorm] support typeorm 1.x

katoztmy:

Fixes #3545

typeorm@1.0.0 removed `EntityManager#findByIds`, and the instrumentation tries to wrap it unconditionally, so patching crashes with "Attempt to wrap undefined property findByIds as function". Also `supportedVersions` is `>=0.3.0 <1`, so 1.x is not instrumented at all.

- Extended the version range to `>=0.3.0 <2`. I kept the upper bound to avoid claiming support for major versions that are not verified yet, which is also why the existing code was capped at `<1`.
- Skip wrapping EntityManager methods that don't exist in the loaded version. In 1.x only `findByIds` is affected.
- I kept the `connection` getter as-is. In 1.x it is deprecated but still works (it just returns `dataSource`), and 0.3.x doesn't have `dataSource` on EntityManager, so `connection` is the only property that works on both.
- typeorm 1.x dropped the `sqlite` driver in favor of `better-sqlite3`, so the test helper now picks whichever driver the installed typeorm supports. The tests themselves are unchanged.
- `.tav.yml` runs 1.x in a separate job with `node: ">=20"` (typeorm 1.0 requires Node 20.19+) and `better-sqlite3@^12.0.0`, same range as typeorm's own peer dependency.

Tested with typeorm 0.3.x and 1.0.0, 13/13 passing on both. TAV run on my fork: https://github.com/katoztmy/opentelemetry-js-contrib/actions/runs/29142553352

I used AI assistance (Claude) for this change. I reviewed and tested everything myself.

This is my first contribution to this repo, so please let me know if I missed anything in the process.

### DIFF
--- a/packages/instrumentation-typeorm/.tav.yml
+++ b/packages/instrumentation-typeorm/.tav.yml
@@ -1,5 +1,13 @@
 typeorm:
-  versions:
-    include: ">=0.3.0 <1"
-    mode: max-7
-  commands: npm run test
+  - versions:
+      include: ">=0.3.0 <1"
+      mode: max-7
+    commands: npm run test
+  # typeorm@1.0.0 requires Node.js ^20.19.0 || ^22.13.0 || >=24.11.0 and
+  # replaced the `sqlite` (sqlite3) driver with `better-sqlite3`.
+  - versions:
+      include: ">=1.0.0 <2"
+      mode: max-7
+    node: ">=20"
+    peerDependencies: better-sqlite3@^12.0.0
+    commands: npm run test
--- a/packages/instrumentation-typeorm/README.md
+++ b/packages/instrumentation-typeorm/README.md
@@ -17,7 +17,7 @@ npm install --save @opentelemetry/instrumentation-typeorm
 
 ## Supported versions
 
-- [`typeorm`](https://www.npmjs.com/package/typeorm) versions `>=0.3.0 <1`
+- [`typeorm`](https://www.npmjs.com/package/typeorm) versions `>=0.3.0 <2`
 
 ## Usage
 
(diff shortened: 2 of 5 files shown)


katoztmy:
Hi @mhennoch @seemk @t2t2 @weyert — just a gentle ping on this PR, it's been about two weeks since it was opened and CI is green (tested against typeorm 0.3.x and 1.0.0). This is my first contribution here, so happy to adjust anything that doesn't fit the repo's conventions. Thanks for taking a look when you get a chance!

weyert (approved):
Changes make sense to me

nickrttn:
Would be great to have this merged. TypeORM 1 has been out for quite some time and it would be good to upgrade to it, but not without telemetry 🙈

pichlermarc (approved):

PR #3620, opened 2026-07-11
docs(instrumentation-pg): clarify db.collection.name is not collected (#3614)

### LINKED ISSUE #3614: Add db.collection.name attribute to instrumentation-pg

Currently with default instrumentation, spans are missing collection name on postgres

add name of collection `db.collection.name` to default collected attributes on pg-instrumentation https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-pg#attributes-collected

### Additional context

tejaswiverma121-byte:
### Which problem is this PR solving?

`@opentelemetry/instrumentation-pg` does not emit `db.collection.name`. This PR updates the README to explicitly clarify why `db.collection.name` is not collected by this instrumentation, in accordance with the OpenTelemetry specification.

Fixes #3614

### Short description of the changes

- Updated `packages/instrumentation-pg/README.md` to document that `db.collection.name` is not collected.
- Clarified that the `pg` driver does not expose table names separately, and the OpenTelemetry specification advises against parsing `db.query.text` when the database supports queries touching multiple collections in non-batch operations (which is the case for PostgreSQL).

### DIFF
--- a/packages/instrumentation-pg/README.md
+++ b/packages/instrumentation-pg/README.md
@@ -80,6 +80,10 @@ The `@opentelemetry/instrumentation-pg` versions 0.72.0 and later emit the stabl
 | `server.port`         | Remote port number.                                                                        |
 | `error.type`          | Describes a class of error the operation ended with.                                       |
 
+> [!NOTE]
+> `db.collection.name` is not collected by this instrumentation.
+> The database client does not make the table/collection name easily available, and parsing the SQL query text is not compliant with the OpenTelemetry specification.
+
 Metrics Exported:
 
 - [`db.client.operation.duration`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientoperationduration)
--- a/packages/instrumentation-pg/README.md
+++ b/packages/instrumentation-pg/README.md
@@ -81,8 +81,10 @@ The `@opentelemetry/instrumentation-pg` versions 0.72.0 and later emit the stabl
 | `error.type`          | Describes a class of error the operation ended with.                                       |
 
 > [!NOTE]
-> `db.collection.name` is not collected by this instrumentation.
-> The database client does not make the table/collection name easily available, and parsing the SQL query text is not compliant with the OpenTelemetry specification.
+> `db.collection.name` is not collected. The `pg` driver does not expose the table
+> name separately, and the OpenTelemetry specification advises against parsing
+> `db.query.text` when the database supports queries touching multiple
+> collections in non-batch operations, which is the case for PostgreSQL.
 
 Metrics Exported:
 


.changeset/add-db-collection-name-pg.md
```diff
@@ -0,0 +1,5 @@
+---
```
maryliag: no need for this file

maryliag (commented):
thank you for working on this! You do have a lot of failed tests, so make sure those are fixed. I also added a few more test cases that should be added

.changeset/add-db-collection-name-pg.md
tejaswiverma121-byte: Done — removed in [e8aae18](https://github.com/tejaswiverma121-byte/opentelemetry-js-contrib/commit/e8aae18). Also fixed the prefer-const lint error in the same commit.

packages/instrumentation-pg/src/utils.ts
```diff
@@ -95,6 +95,28 @@ export function parseNormalizedOperationName(queryText: string) {
   return sqlCommand.endsWith(';') ? sqlCommand.slice(0, -1) : sqlCommand;
 }
 
+/**
```
maryliag:
```suggestion
// Match common SQL clauses that precede a table name: FROM, INTO, UPDATE, JOIN
// The table name can be a simple identifier, or a quoted identifier (which can contain spaces),
// or a schema-qualified identifier (e.g. "my-schema"."my table")
const TABLE_NAME_REGEX =
  /\b(?:FROM|INTO|UPDATE|JOIN)\s+((?:[^\s();,"'`]+|["'`][^"'`]+["'`])+)/i;

/**
```

packages/instrumentation-pg/src/utils.ts
```diff
@@ -95,6 +95,28 @@ export function parseNormalizedOperationName(queryText: string) {
...
+  const identifierRegex = /((?:[^\s();,"'`]+|["'`][^"'`]+["'`])+)/;
+  const match = queryText.match(new RegExp(`\\b(?:FROM|INTO|UPDATE|JOIN)\\s+${identifierRegex.source}`, 'i'));
+
+  if (match && match[1]) {
+    // Remove optional quotes (single, double, or backticks) that might surround the table name or schema name
+    const tableName = match[1].replace(/["'`]/g, '');
+    return tableName;
+  }
+  return undefined;
+}
```
maryliag:
```suggestion
export function parseTableName(queryText: string): string | undefined {
  const match = queryText.match(TABLE_NAME_REGEX);

  if (match?.[1]) {
    // Remove optional quotes (single, double, or backticks) that might surround the table name or schema name
    return match[1].replace(/["'`]/g, '');
  }
  return undefined;
}
```

maryliag (commented):
I added a few suggestion to help performance, but I'm still debating if this is the right approach for this feature.

In the spec we have: "The collection name SHOULD NOT be extracted from db.query.text, when the database system supports query text with multiple collections in non-batch operations.", which is the case for PG.

Part of it is because the query can have several tables, so it can be hard to decide which one to select, but also because parsing itself can be very costly. If we have a lot of queries being executed, having to match the regex can be an issue for the performance, so I'm thinking there could be an opt-in somehow, at least initially.
I'll discuss this with the maintainers on our weekly (wednesday), and get back to you.

tejaswiverma121-byte:
Hi @maryliag, thanks for the review and the suggestions!

I've gone ahead and applied your suggested regex changes to help improve the parsing performance. I also updated one of the unit tests since the new regex intentionally drops support for parsing single quotes (which makes sense since they are used for string literals, not table identifiers).

Additionally, I've fixed the failing lint checks, so the CI should be all green now.

maryliag:
@tejaswiverma121-byte we won't be able to merge this PR, because it doesn't follow the spec. 
A few other languages have the value of the collection easily available, but JS + PG doesn't, which is why you had to parse the query text, which is something that the spec tells you're not suppose to do.
For this reason, this feature can't be implemented here.

I'm going to ask you to update the README instead, clarifying that this attribute is not being collected and that is expected. You can make the update in this PR (after reverting all the other changes) or if it's easier, just close this PR and open a new one with just the readme update.

I appreciate you working on this either way!

tejaswiverma121-byte:
@maryliag is there any issue in this pr ?let me know

maryliag:
@tejaswiverma121-byte have you seen the comment I added about 2 weeks ago? I haven't seen it been addressed, which is what is missing to get this merged

tejaswiverma121-byte:
Hi @maryliag,

I have reverted all the SQL parsing code changes and updated the `README.md` as requested to clarify that `db.collection.name` is not collected by `@opentelemetry/instrumentation-pg` (due to OpenTelemetry spec compliance).
The PR now only contains the documentation update. Please take a look when you get a chance!

/dashboard route:reviewers

packages/instrumentation-pg/README.md
```diff
@@ -78,6 +80,10 @@ The `@opentelemetry/instrumentation-pg` versions 0.72.0 and later emit the stabl
 | `server.port`         | Remote port number.                                                                        |
 | `error.type`          | Describes a class of error the operation ended with.                                       |
 
+> [!NOTE]
+> `db.collection.name` is not collected by this instrumentation.
+> The database client does not make the table/collection name easily available, and parsing the SQL query text is not compliant with the OpenTelemetry specification.
```
maryliag:
```suggestion
> `db.collection.name` is not collected. The `pg` driver does not expose the table
> name separately, and the OpenTelemetry specification advises against parsing
> `db.query.text` when the database supports queries touching multiple
> collections in non-batch operations, which is the case for PostgreSQL.
```

maryliag (commented):
I updated the title to reflect the current changes, but you need to also update the PR description to reflect the latest state of the actual changes on this PR.
I also added a suggesting for the wording

tejaswiverma121-byte:
Applied the suggested wording and updated the PR description. Thanks @maryliag! Does it look good to go now?

tejaswiverma121-byte:
@maryliag why the test packages test didnt work ...can u re-run the failed CI job since it's a doc-only change.

tejaswiverma121-byte:
@maryliag is this fix , okay to be merge now ?

maryliag (approved):

PR #3621, opened 2026-07-12
build: test all versions in parallel

Ugzuzg:

- Slow tav runs.
- `npm` issues with installing different versions of dependencies into the entire monorepo.

- Patch test-all-versions to use isolated installs. This allows the tests to be executed in parallel, as test-all-versions no longer operates on the entire monorepo `node_modules` structure. Since it doesn't have to solve for dependencies of the entire monorepo, this also improves the correctness of the installed versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "opentelemetry-js-contrib",
       "version": "0.1.0",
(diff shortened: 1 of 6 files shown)


Ugzuzg:
Sad.

PR #3624, opened 2026-07-17
feat(resource-detector-github)!: use semconv cicd.* and vcs.* attributes

### LINKED ISSUE #2927: Move github detector to use semconv

I want the detector to be following semconv attribute namimg

The github detector leverages attributes defined in semconv rather than custom attributes 

N/a

### Additional context

https://opentelemetry.io/docs/specs/semconv/registry/attributes/cicd/

https://opentelemetry.io/docs/specs/semconv/registry/attributes/vcs/

bhuvan-somisetty:

Fixes #2927

`gitHubDetector` emits custom `github.*` resource attributes for data that the
[CI/CD](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cicd/) and
[VCS](https://opentelemetry.io/docs/specs/semconv/registry/attributes/vcs/) semantic
conventions already define, so the detector's output doesn't line up with what other
CI/CD and VCS instrumentation produces.

Map the attributes that have a semconv equivalent onto it:

| Before | After |
| ------ | ----- |
| `github.workflow` | `cicd.pipeline.name` |
| `github.run_id` | `cicd.pipeline.run.id` |
| `github.sha` | `vcs.ref.head.revision` |
| `github.base_ref` | `vcs.ref.base.name` |
| `github.ref` / `github.head_ref` | `vcs.ref.head.name` |

`github.actor` and `github.run_number` are left as-is, since semconv doesn't currently
define an equivalent for either.

`GITHUB_REF` and `GITHUB_HEAD_REF` both map onto `vcs.ref.head.name`, so the detector
prefers `GITHUB_HEAD_REF` and falls back to `GITHUB_REF`. `GITHUB_HEAD_REF` is only set
for pull request events, and for those events `GITHUB_REF` holds the synthetic merge ref
(e.g. `refs/pull/1/merge`) rather than the branch being merged. There's a test covering
the non-PR fallback.

The unstable semconv definitions are copied into `src/semconv.ts` using
`scripts/gen-semconv-ts.js`, per the [unstable semconv guidance](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv).
No dependency on `@opentelemetry/semantic-conventions` is added, matching the other
detectors that only use copied definitions (gcp, alibaba-cloud, container, instana).

README updated; existing tests updated and one added.

## Note for reviewers

No rush on this — a couple of points worth your input whenever you get to it:

- This renames the emitted attributes outright, which is breaking for anyone matching on
  the old `github.*` names. I went with a straight rename since the package is pre-1.0 and
  that's what #2927 asks for, but I'm happy to emit both old and new for a transition period
  instead if that's preferred.
- Let me know if `github.actor` / `github.run_number` should be dropped rather than kept.
- I saw the bot's note that this package is in feature-freeze without a component owner. If
  the semconv alignment here is something a JS approver would be willing to sponsor, I'd
  appreciate it, but I completely understand if the package's status means that's not
  feasible. Either way I'm happy to keep contributing toward being able to help maintain
  this package longer term.
  
  
  
  
  
  
  
  
  
  
  
  @blumamir @jj22ee @trivikr @david-luna @pichlermarc @trentm
   it looks like it may be flagged for auto-close due to inactivity. Would appreciate a review whenever you get a chance. No         rush!

### DIFF
--- a/packages/resource-detector-github/src/detectors/GitHubDetector.ts
+++ b/packages/resource-detector-github/src/detectors/GitHubDetector.ts
@@ -5,6 +5,14 @@
 
 import { ResourceDetector, DetectedResource } from '@opentelemetry/resources';
 
+import {
+  ATTR_CICD_PIPELINE_NAME,
+  ATTR_CICD_PIPELINE_RUN_ID,
+  ATTR_VCS_REF_BASE_NAME,
+  ATTR_VCS_REF_HEAD_NAME,
+  ATTR_VCS_REF_HEAD_REVISION,
+} from '../semconv';
+
 /**
  * The GitHubDetector can be used to detect GitHub Actions environment variables
  * and returns resource attributes with GitHub-specific metadata that exists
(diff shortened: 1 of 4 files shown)


bhuvan-somisetty:
Friendly bump on this one - it looks like it may be flagged for auto-close due to inactivity. Would appreciate a review whenever you get a chance. No rush!

pichlermarc:
@open-telemetry/javascript-approvers WDYT on updating semconv on this detector without an opt-in period? IIUC there is no opt-in period defined for this part of semconv, so I'd be inclined to go ahead without it.

bhuvan-somisetty:
Thanks @pichlermarc! Sounds good to me. Appreciate you checking with the team.

packages/resource-detector-github/src/detectors/GitHubDetector.ts
```diff
@@ -15,15 +23,19 @@ import { ResourceDetector, DetectedResource } from '@opentelemetry/resources';
  */
 class GitHubDetector implements ResourceDetector {
   public detect(): DetectedResource {
+    // GITHUB_HEAD_REF is only set for pull request events, where GITHUB_REF
+    // holds the synthetic merge ref instead of the branch being merged.
+    const headRef = process.env.GITHUB_HEAD_REF || process.env.GITHUB_REF;
```
maryliag:
```suggestion
    const headRef = process.env.GITHUB_HEAD_REF || process.env.GITHUB_REF_NAME || process.env.GITHUB_REF;
```

`GITHUB_HEAD_REF` is a short branch name (my-feature-branch), but `GITHUB_REF` is a full ref path (refs/heads/main, refs/tags/v1.0.0). `GITHUB_REF_NAME` gives the short-form name and would produce more semconv-consistent values

maryliag (commented):
I would keep `github.actor`, `github.run_number`, if we ever get complain we can remove in the future.

I don't think there is a need for an opt-in here.

I added a small nit

packages/resource-detector-github/src/detectors/GitHubDetector.ts
bhuvan-somisetty: Thanks @maryliag! Updated headRef to check GITHUB_REF_NAME as well (GITHUB_HEAD_REF || GITHUB_REF_NAME || GITHUB_REF) and added test coverage for it.

maryliag (approved):

bhuvan-somisetty:
@trentm, @pichlermarc PTAL whenever u get chance

maryliag:
friendly ping @pichlermarc if you're okay with this solution

PR #3626, opened 2026-07-19
fix(scripts): build glob patterns with forward slashes for Windows

### LINKED ISSUE #3625: Repo scripts using path.join() for glob patterns silently match nothing on Windows

Reproduced with Node v24.14.0 on Windows 11, using `glob@13.0.6` (the version resolved in the repo lockfile). Not Node-version specific — it reproduces on any Windows checkout.

### What did you do?

Ran the repo's own tooling scripts on Windows. Several of them enumerate workspace packages (and source files) with `glob`, but build the glob pattern using `path.join(...)`. On Windows `path.join` uses `\` as the separator, and `glob` treats `\` as an escape character rather than a path separator, so the pattern matches nothing.

The clearest case is `scripts/gen-semconv-ts.js`, which is the documented way to regenerate a package's local `src/semconv.ts`:

```
node scripts/gen-semconv-ts.js packages/resource-detector-github
```

On Windows this prints:

```
Did not find any usage of unstable semconv exports in "packages/resource-detector-github/{src,test}/**/*.ts".
No changes made.
```

even though the package clearly imports unstable semconv definitions. The same root cause affects every script that enumerates workspaces this way.

Minimal reproduction of the underlying behavior:

```js
const { globSync } = require('glob');
const path = require('path');

globSync(path.join('packages', '*', 'package.json')).length;
// Windows -> 0   (pattern becomes "packages\\*\\package.json")
// POSIX   -> 72

globSync('packages/*/package.json').length;
// Windows -> 72  (forward slashes work everywhere)
```

The scripts should discover workspace packages and source files on Windows the same way they do on Linux/macOS, and `gen-semconv-ts.js` should regenerate `src/semconv.ts`.

### What did you see instead?

Every affected script silently finds **zero** matches on Windows and no-ops. There's no error — the workspace-enumerating scripts just behave as if the monorepo were empty, and `gen-semconv-ts.js` reports "No changes made." This is easy to mistake for "nothing to do" rather than a broken glob.

### Affected call sites

All of these pass a `path.join(...)` result straight into `globSync(...)`:

| File | Line | Pattern built | Effect on Windows |
| ---- | ---- | ------------- | ----------------- |
| `scripts/gen-semconv-ts.js` | 54 | `path.join(wsGlob, 'package.json')` | no workspaces found |
| `scripts/gen-semconv-ts.js` | 71 | `path.join(wsDir, '{src,test}', '**', '*.ts')` | no source files found → regeneration no-ops |
| `scripts/lint-semconv-deps.mjs` | 44 | `path.join(wsGlob, 'package.json')` | no workspaces linted |
| `scripts/lint-semconv-deps.mjs` | 75 | `path.join(wsDir, 'src', '**', '*.ts')` | no source files checked |
| `scripts/peer-api-check.js` | 18 | `path.join(wsGlob, 'package.json')` | no workspaces found |
| `scripts/update-otel-deps.js` | 39 | `path.join(wsGlob, 'package.json')` | no workspaces found |
| `scripts/codecov-upload-flags.mjs` | 21 | `path.join(exp, 'package.json')` | no workspaces found |

(`scripts/check-release-please.mjs` and `scripts/gen-component-label-map.mjs` use the same shape via `node:fs` `globSync` / a `join` import and are worth checking under the same fix.)

Because `lint-semconv-deps` is part of `npm run lint`, its checks are effectively skipped for a Windows contributor — it passes by matching nothing rather than by finding no problems.

### Root cause

```mermaid
flowchart TD
    A["Script builds glob pattern<br/>with path.join(...)"] --> B{Platform?}
    B -->|"POSIX (Linux / macOS)"| C["Pattern uses '/'<br/>e.g. packages/*/package.json"]
    B -->|Windows| D["Pattern uses '\\'<br/>e.g. packages\\*\\package.json"]
    C --> E["glob treats '/' as separator"]
    D --> F["glob treats '\\' as ESCAPE char"]
    E --> G["Matches all packages"]
    F --> H["Matches nothing → 0 results"]
    H --> I["Script silently no-ops<br/>(no error, looks like 'nothing to do')"]
```

`glob` intentionally uses forward slashes on every platform and reserves `\` as an escape character (see the [glob "Windows" docs](https://github.com/isaacs/node-glob#windows)). `path.join` is the wrong tool for assembling glob patterns.

### Proposed fix

Stop feeding `path.join` output to `glob`. Two supported options:

1. Build patterns with forward slashes directly (e.g. `` `${wsDir}/{src,test}/**/*.ts` ``), normalizing any `path.sep` in the base to `/` first. This keeps patterns readable and works on all platforms.
2. Pass `{ windowsPathsNoEscape: true }` to the `glob` calls (supported in `glob@13`), which makes `\` behave as a separator.

Option 1 is the more explicit fix and matches glob's own guidance. Either way the change is isolated to the `scripts/` directory and has no runtime/package impact.

I'd like to pick this up and open a PR.

### Additional context

Found while regenerating `src/semconv.ts` for a package on Windows; the generator reported "No changes made" despite the package using unstable semconv exports, which traced back to the glob pattern. No existing open or closed issue tracks this.

bhuvan-somisetty:

Fixes #3625

Several repo tooling scripts enumerate workspace packages (and source files) with `glob`, but build the glob pattern using `path.join(...)`. On Windows, `path.join` uses `\` as the separator, and `glob` treats `\` as an escape character rather than a path separator, so the pattern silently matches nothing.

The clearest case is `scripts/gen-semconv-ts.js`, the documented way to regenerate a package's local `src/semconv.ts`:

```
node scripts/gen-semconv-ts.js packages/resource-detector-github
```

On Windows this printed `Did not find any usage of unstable semconv exports...` and no-op'd, even though the package imports unstable semconv definitions. Every affected script has the same failure mode: it finds zero matches and silently behaves as if the monorepo were empty, with no error.

Because `lint-semconv-deps` is part of `npm run lint`, its checks were effectively skipped for Windows contributors - it passed by matching nothing rather than by finding no problems.

- `scripts/gen-semconv-ts.js`, `scripts/lint-semconv-deps.mjs`, `scripts/peer-api-check.js`, `scripts/update-otel-deps.js`, `scripts/codecov-upload-flags.mjs`, `scripts/gen-component-label-map.mjs`: build glob patterns with forward slashes instead of `path.join(...)`, and pass `posix: true` to `globSync` so the returned paths are also forward-slash (needed since some of these scripts compare returned paths against forward-slash strings from `package.json`/YAML files, e.g. the `lint-semconv-deps` exclude list and `component_owners.yml` lookups).
- `scripts/check-release-please.mjs` uses `node:fs`'s `globSync`, which already handles `path.join`-built patterns fine on Windows. However, it builds its manifest/config lookup keys with `path.relative(...)`, which returns platform-native separators - producing backslash keys that never matched the forward-slash keys already in `.release-please-manifest.json` and `release-please-config.json`. Normalized those to forward slashes too, since this failed outright on Windows before this fix (every package was reported as a missing manifest/config entry).

No dependency changes; the change is isolated to `scripts/` and has no runtime/package impact.

## Test plan

Verified interactively on a Windows checkout (all previously reported zero matches, now find the expected workspaces/files):

- `node scripts/gen-semconv-ts.js packages/contrib-test-utils` finds the package's unstable semconv imports and regenerates `src/semconv.ts` (previously reported "Did not find any usage...").
- `node scripts/lint-semconv-deps.mjs -x packages/contrib-test-utils/src/resource-assertions.ts` (the exact `npm run lint:semconv-deps` invocation) now runs its checks across all workspaces and exits 0 (previously silently checked nothing).
- `node scripts/peer-api-check.js` now checks all workspace packages instead of none.
- `node scripts/check-release-please.mjs` now exits `OK` instead of reporting every package as missing from the manifest/config.
- `node scripts/gen-component-label-map.mjs` now correctly enumerates workspace dirs.

Ran `eslint` and `prettier --check` on all changed files.

### DIFF
--- a/scripts/bitrot.mjs
+++ b/scripts/bitrot.mjs
@@ -27,6 +27,15 @@ const TOP = path.resolve(fileURLToPath(new URL('.', import.meta.url)), '..');
 const USE_COLOR = process.stdout.isTTY && !process.env.NO_COLOR?.length > 0;
 const BUILD_DIR = path.join(TOP, 'build', 'bitrot');
 
+// `glob` always wants forward slashes in its patterns, even on Windows,
+// where `path.join` produces backslashes.
+function toGlobPattern(...parts) {
+  return path
+    .join(...parts)
+    .split(path.sep)
+    .join('/');
+}
+
 let numProbs = 0;
 function problem(...args) {
   numProbs += 1;
@@ -93,7 +102,7 @@ function bitrotRenovateCoreExperimental() {
   }
 
   const pkgNames = globSync(
-    path.join(ojDir, 'experimental/packages/*/package.json')
+    toGlobPattern(ojDir, 'experimental/packages/*/package.json')
   )
     .map(packageJson => JSON.parse(fs.readFileSync(packageJson)))
     .filter(pj => isPublicPackage(pj))
@@ -168,7 +177,7 @@ function getNpmInfo(name) {
  */
 function bitrotInstrumentations() {
   const instrReadmes = globSync(
-    path.join(TOP, 'packages/instrumentation-*/README.md')
+    toGlobPattern(TOP, 'packages/instrumentation-*/README.md')
   );
 
   // Match examples:
--- a/scripts/check-release-please.mjs
+++ b/scripts/check-release-please.mjs
@@ -31,12 +31,17 @@ const getPackages = () => {
   const TOP = process.cwd();
   const pj = readJson(path.join(TOP, 'package.json'));
   return pj.workspaces
-    .map(wsGlob => globSync(path.join(wsGlob, 'package.json')))
+    .map(wsGlob => globSync(`${wsGlob}/package.json`))
     .flat()
     .map(p => {
       const pkgInfo = readJson(p);
       pkgInfo.location = path.dirname(p);
-      pkgInfo.relativeLocation = path.relative(PROJECT_ROOT, pkgInfo.location);
+      // release-please's manifest/config files always use forward slashes
+      // for their package location keys, regardless of platform.
+      pkgInfo.relativeLocation = path
+        .relative(PROJECT_ROOT, pkgInfo.location)
+        .split(path.sep)
+        .join('/');
       return pkgInfo;
     });
 };
--- a/scripts/codecov-upload-flags.mjs
+++ b/scripts/codecov-upload-flags.mjs
@@ -18,7 +18,7 @@ const readPkg = dir =>
   JSON.parse(readFileSync(path.join(dir, 'package.json'), 'utf8'));
 const pkgInfo = readPkg('.');
 const pkgFiles = pkgInfo.workspaces.map(exp =>
-  globSync(path.join(exp, 'package.json'))
+  globSync(`${exp}/package.json`, { posix: true })
 );
 const pkgsWithFlag = pkgFiles.flat().map(f => {
   const path = f.replace('package.json', '');
--- a/scripts/gen-component-label-map.mjs
+++ b/scripts/gen-component-label-map.mjs
@@ -41,7 +41,7 @@ function warn(...args) {
 function getAllWorkspaceDirs() {
   const pj = JSON.parse(fs.readFileSync(join(TOP, 'package.json'), 'utf8'));
   return pj.workspaces
-    .map(wsGlob => globSync(join(wsGlob, 'package.json')))
+    .map(wsGlob => globSync(`${wsGlob}/package.json`, { posix: true }))
     .flat()
     .map(dirname);
 }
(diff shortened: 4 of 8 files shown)


hectorhdzg:
Same pattern is used in https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/scripts/bitrot.mjs it would be good to cover that one as well

bhuvan-somisetty:
Good catch — pushed a fix for the same pattern in `scripts/bitrot.mjs` (two `globSync(path.join(...))` calls) using the same `toGlobPattern` helper approach used in `gen-semconv-ts.js`/`lint-semconv-deps.mjs`. Verified locally on Windows that the glob now correctly matches files (previously it silently returned zero matches due to backslashes in the pattern).

hectorhdzg (approved):
LGTM, I validated on my side, fix is working in Windows machines

bhuvan-somisetty:
Thanks for validating on Windows and for the review!

trentm (approved):

bhuvan-somisetty:
@trentm Thanks for the review! All checks are green and the PR is approved nd ready to merge whenever you get a chance.

PR #3627, opened 2026-07-20
feat(instrumentation-tedious): add missing stable SemConv attributes

### LINKED ISSUE #3290: improve on instrumentation-tedious stable semconv
In https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3273, instr-tedious is being updated to support migration to stable `net.*` and `db.*` semantic conventions. This initial pass focused on migrating existing span attributes (https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/#database-client-span-attributes) and on stable span attributes marked as "Required" in https://opentelemetry.io/docs/specs/semconv/database/sql-server/

There are a number of "Recommended" and "Opt-In" database semconv span attributes that could be added to this instrumentation. In addition there are database *metrics* that could be added to this instrumentation. Please refer to:

- https://opentelemetry.io/docs/specs/semconv/database/sql-server/
- https://opentelemetry.io/docs/specs/semconv/database/database-metrics/

Note: there are some "TODO" comments in the instr-tedious sources with short notes on *some* span attributes that could/should be considered.

aryamohanan:

This PR completes the remaining stable semantic convention support for `instrumentation-tedious`. A few database attributes introduced as part of the stable semconv migration were still missing, and `db.namespace` did not handle SQL Server named instances correctly.

* Added `db.operation.name`.
* Added `db.stored_procedure.name` for `callProcedure`.
* Added `db.response.status_code` and `error.type` on failed requests.
* Fixed `db.namespace` for SQL Server named instances (`{instanceName}|{databaseName}`).
* Updated span names to use the resolved operation name.
* Updated unit tests for the new attributes and span names.
* Updated the README with the new attributes and span naming.

closes #3290

### DIFF
--- a/packages/instrumentation-tedious/src/instrumentation.ts
+++ b/packages/instrumentation-tedious/src/instrumentation.ts
@@ -13,15 +13,19 @@ import {
 import {
   ATTR_DB_COLLECTION_NAME,
   ATTR_DB_NAMESPACE,
+  ATTR_DB_OPERATION_NAME,
   ATTR_DB_QUERY_TEXT,
+  ATTR_DB_RESPONSE_STATUS_CODE,
+  ATTR_DB_STORED_PROCEDURE_NAME,
   ATTR_DB_SYSTEM_NAME,
+  ATTR_ERROR_TYPE,
   ATTR_SERVER_ADDRESS,
   ATTR_SERVER_PORT,
   DB_SYSTEM_NAME_VALUE_MICROSOFT_SQL_SERVER,
 } from '@opentelemetry/semantic-conventions';
 import type * as tedious from 'tedious';
 import { TediousInstrumentationConfig } from './types';
-import { getSpanName, once } from './utils';
+import { getOperationName, getSpanName, once } from './utils';
 /** @knipignore */
 import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 
@@ -199,24 +203,37 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
 
         const attributes: api.Attributes = {};
 
-        // The OTel spec for "db.namespace" discusses handling for connection
-        // to MSSQL "named instances". This isn't currently supported.
-        //    https://opentelemetry.io/docs/specs/semconv/database/sql-server/#:~:text=%5B1%5D%20db%2Enamespace
-        attributes[ATTR_DB_NAMESPACE] = databaseName;
+        // db.namespace: for named instances include the instance name as a
+        // prefix separated by "|" per the SQL Server semconv spec.
+        // https://opentelemetry.io/docs/specs/semconv/database/sql-server/#:~:text=%5B1%5D%20db%2Enamespace
+        const instanceName = this.config?.options?.instanceName;
+        const dbNamespace = instanceName
+          ? `${instanceName}|${databaseName}`
+          : databaseName;
+        attributes[ATTR_DB_NAMESPACE] = dbNamespace;
         attributes[ATTR_DB_SYSTEM_NAME] =
           DB_SYSTEM_NAME_VALUE_MICROSOFT_SQL_SERVER;
         attributes[ATTR_DB_QUERY_TEXT] = sql;
         attributes[ATTR_DB_COLLECTION_NAME] = request.table;
-        // See https://opentelemetry.io/docs/specs/semconv/database/sql-server/#spans
-        // TODO(3290): can `db.response.status_code` be added?
-        // TODO(3290): is `operation` correct for `db.operation.name`
-        // TODO(3290): can `db.query.summary` reliably be calculated?
-        // TODO(3290): `db.stored_procedure.name`
+
+        const operationName = getOperationName(operation, sql);
+        if (operationName !== undefined) {
+          attributes[ATTR_DB_OPERATION_NAME] = operationName;
+        }
+
+        // db.stored_procedure.name: available directly from sqlTextOrProcedure
+        // when callProcedure is used.
+        if (operation === 'callProcedure' && sql) {
+          attributes[ATTR_DB_STORED_PROCEDURE_NAME] = sql;
+        }
 
         attributes[ATTR_SERVER_ADDRESS] = this.config?.server;
         attributes[ATTR_SERVER_PORT] = this.config?.options?.port;
+
+        const spanCollection =
+          operation === 'callProcedure' ? sql : request.table;
         const span = thisPlugin.tracer.startSpan(
-          getSpanName(operation, databaseName, sql, request.table),
+          getSpanName(operationName, dbNamespace, spanCollection),
           {
             kind: api.SpanKind.CLIENT,
             attributes,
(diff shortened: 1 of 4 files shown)


packages/instrumentation-tedious/src/utils.ts
```diff
@@ -4,32 +4,53 @@
...
+  sql: string | undefined
+): string | undefined {
+  if (tediousMethod === 'callProcedure') return 'EXECUTE';
+  if (tediousMethod === 'execBulkLoad') return 'BULK INSERT';
+  if (!sql) return undefined;
+  const trimmed = sql.trimStart();
+  const idx = trimmed.search(/\s/);
+  const verb = idx === -1 ? trimmed : trimmed.slice(0, idx);
+  const normalized = verb.replace(/;$/, '').toUpperCase();
+  return normalized || undefined;
```
pichlermarc: The [specification for MSSQL clients](https://opentelemetry.io/docs/specs/semconv/db/sql-server/) mentions this recommendation:

> [10] db.operation.name: The operation name SHOULD NOT be extracted from db.query.text.

also the [general DB client spec](https://opentelemetry.io/docs/specs/semconv/db/database-spans/) mentions something similar, with additional reasoning:

> [5] db.operation.name: If readily available and if there is a single operation name that describes the database call.
> 
> [6] db.operation.name: It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
>
> The operation name SHOULD NOT be extracted from db.query.text, when the database system supports query text with multiple operations in non-batch operations.
>
> If spaces can occur in the operation name, multiple consecutive spaces SHOULD be normalized to a single space.
>
> For batch operations, if the individual operations are known to have the same operation name then that operation name SHOULD be used prepended by BATCH , otherwise db.operation.name SHOULD be BATCH or some other database system specific term if more applicable.

IIUC, then this part:

> when the database system supports query text with multiple operations in non-batch operations

does mention exactly what this would be doing.

**suggestion:** let's not do this unless we have a good reason to do it this way.

packages/instrumentation-tedious/src/utils.ts
```diff
@@ -4,32 +4,53 @@
...
+  collection: string | undefined
 ): string {
-  if (operation === 'execBulkLoad' && bulkLoadTable && db) {
-    return `${operation} ${bulkLoadTable} ${db}`;
+  if (!operationName) return 'db';
+
+  // For stored-procedure calls and bulk loads the "collection" is the
+  // procedure / table name, which gives a useful low-cardinality name.
+  if (collection && db) {
+    return `${operationName} ${collection} ${db}`;
```
pichlermarc: unrelated (feel free to ignore for now, but might be an interesting follow-up): from what I understand, span names should always follow something like

`{db.operation.name} {target}` with a single target chosen by preference (collection, stored procedure, namespace)

ref: https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name

packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -237,7 +254,17 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
               code: api.SpanStatusCode.ERROR,
               message: err.message,
             });
-            // TODO(3290): set `error.type` attribute?
+
+            const errorType = err.constructor.name;
```
pichlermarc: **suggestion:** just in case the emitter does receive an object that does not have a constructor. it's very unlikely, but I think the user owns the request and can emit whatever they like (including null prototype objects)

```suggestion
            const errorType = err?.constructor?.name ?? 'Error';
```

packages/instrumentation-tedious/src/utils.ts
```diff
@@ -4,32 +4,53 @@
...
-  operation: string,
+  operationName: string | undefined,
   db: string | undefined,
-  sql: string | undefined,
-  bulkLoadTable: string | undefined
+  collection: string | undefined
 ): string {
-  if (operation === 'execBulkLoad' && bulkLoadTable && db) {
-    return `${operation} ${bulkLoadTable} ${db}`;
+  if (!operationName) return 'db';
```
pichlermarc: I think this is the fallback, right?

From the [DB span name spec](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#name):

> If neither {db.operation.name} nor {target} are available, span name SHOULD be {db.system.name}.

```suggestion
  if (!operationName) return DB_SYSTEM_NAME_VALUE_MICROSOFT_SQL_SERVER;
```

actual logic might be a bit more complex as it also should fall back to `{target}` first, which has its own set of rules: https://opentelemetry.io/docs/specs/semconv/db/database-spans/#target-placeholder

packages/instrumentation-tedious/src/instrumentation.ts
```diff
@@ -199,24 +203,37 @@ export class TediousInstrumentation extends InstrumentationBase<TediousInstrumen
...
 
-        // The OTel spec for "db.namespace" discusses handling for connection
-        // to MSSQL "named instances". This isn't currently supported.
-        //    https://opentelemetry.io/docs/specs/semconv/database/sql-server/#:~:text=%5B1%5D%20db%2Enamespace
-        attributes[ATTR_DB_NAMESPACE] = databaseName;
+        // db.namespace: for named instances include the instance name as a
+        // prefix separated by "|" per the SQL Server semconv spec.
+        // https://opentelemetry.io/docs/specs/semconv/database/sql-server/#:~:text=%5B1%5D%20db%2Enamespace
+        const instanceName = this.config?.options?.instanceName;
+        const dbNamespace = instanceName
```
pichlermarc: **question:** I'm not too familiar with the order of things, but is it it possible that `instanceName` is defined but `databaseName` is not and we end up with a value like `myinstance|undefined`

pichlermarc (commented):
thank you for working on this 🙌

packages/instrumentation-tedious/src/utils.ts
aryamohanan: Good catch, thank you! ❤️  You're right, the MSSQL semconv explicitly recommends against extracting db.operation.name from db.query.text. I will update that

packages/instrumentation-tedious/src/utils.ts
aryamohanan: Thanks for the pointer to the spec! I will update this too.

packages/instrumentation-tedious/src/instrumentation.ts
aryamohanan: Great spot! if instanceName is set but the user hasn't configured a database, databaseName would be undefined and we'd produce "myinstance|undefined". I will fix this

packages/instrumentation-tedious/src/utils.ts
aryamohanan: updated

maryliag (commented):
you will need to update the tests that are now breaking after your latest commit, otherwise looks good

maryliag (approved):

pichlermarc (approved):
thanks!

PR #3634, opened 2026-07-22
feat(instrumentation-pg): implement db.client.connection.max

sethrylan:

- Implements the remaining `db.client.connection.max` PostgreSQL pool metric from the database metrics prototype tracked in #2348.

- Adds the local `METRIC_DB_CLIENT_CONNECTION_MAX` semantic-convention constant without importing incubating definitions or changing the semantic-conventions dependency.
- Creates `db.client.connection.max` with unit `{connection}` and description `The maximum number of open connections allowed.`.
- Tracks active pools across MeterProvider replacement by clearing the old instrument and replaying current pool values into the new provider.
- Wraps pool shutdown to subtract the pool's maximum once the pool enters its ending state, preventing stale capacity values.


packages/instrumentation-pg/src/instrumentation.ts
```diff
@@ -73,7 +75,12 @@ const INTERNAL_SET_QUERY = Symbol(
 export class PgInstrumentation extends InstrumentationBase<PgInstrumentationConfig> {
   declare private _operationDuration: Histogram;
   declare private _connectionsCount: UpDownCounter;
+  declare private _connectionMax: UpDownCounter;
```
sethrylan, line 78: The current [Development specification](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/#metric-dbclientconnectionmax) has `UpDownCounter`, but this seems like it could be an ObservableGauge, which would require less record-keeping to prevent double counting and subtract when a connection is closed.

maryliag: For this PR keep following the spec, but you can also open an issue within the spec repo with this topic, and if that is changed there we can also change here :)

maryliag (commented):
Can you also update the readme for this package, to include the new metric

packages/instrumentation-pg/README.md
```diff
@@ -82,6 +82,7 @@ Metrics Exported:
 
 - [`db.client.operation.duration`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientoperationduration)
 - [`db.client.connection.count`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientconnectioncount)
+- [`db.client.connection.max`](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/#metric-dbclientconnectionmax)
```
maryliag, line 85: all the others are using the github link for the spec, so you can either update this one to also use the github one or update all the other to use the website one, as long as they're all consistent

maryliag:
@sethrylan are you still planning on working on this PR?

packages/instrumentation-pg/README.md
sethrylan: Good point! Updated.

sethrylan:
Hi @maryliag, thanks for the ping. Yes, I think this is pretty close to done, so I will see it through. If I can't, I'll be sure to handover.

Let me know if you have any other feedback I need to address on this PR.

maryliag (approved):
Great, thank you for working on this!

PR #3636, opened 2026-07-23
fix(instrumentation-user-interaction): compute the XPath inside the span creation try/catch

rycont:
Hi! I ran into this in my own app.

1. I was getting a ton of `Permission denied to access property "localName"` errors.

2. I'm on Firefox (Zen browser) with Bitwarden. If I disable Bitwarden, the error
   doesn't happen.

3. In my case the console gets flooded while Bitwarden's save-password modal is
   showing on the page.

4. I took a look at the code that throws. The block right after it already handles
   errors: it logs through `_diag` and just skips creating the span instead of
   letting the error out. If that's the intent, it seemed to me the XPath
   computation itself should be inside the try too.

5. So I moved it. Please take a look. I also saw similar reports in other
   frontend projects, so it's not just me, though it does look like a rare
   combination.

It was mostly console noise for me, but as far as I can tell the patched
listener throws before it reaches `_invokeListener()`, so the app's own
listener can get skipped.

### DIFF
--- a/packages/instrumentation-user-interaction/src/instrumentation.ts
+++ b/packages/instrumentation-user-interaction/src/instrumentation.ts
@@ -128,8 +128,8 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
     if (!this._allowEventName(eventName)) {
       return undefined;
     }
-    const xpath = getElementXPath(element, true);
     try {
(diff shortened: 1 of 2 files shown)


overbalance (approved):
Nice catch! Thanks

PR #3642, opened 2026-07-28
docs(instrumentation-runtime-node): fix README inaccuracies

aryamohanan:

- Update supported Node.js version from >=14.10 to ^18.19.0 || >=20.6.0
  to match the engines field in package.json
- Remove stale "two ELU readings" note, outdated since the _lastValue
  initialization fix in PR #3118
- Fix wrong line number links in the options table

### DIFF
--- a/packages/instrumentation-runtime-node/README.md
+++ b/packages/instrumentation-runtime-node/README.md
@@ -12,9 +12,7 @@ process termination.
 
 ## Supported Versions
 
-- Node.js `>=14.10`
-
-<!-- - 14.6.0 - this package uses _private properties_ -->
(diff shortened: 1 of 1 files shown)


david-luna (commented):
LGTM. I wonder if we could remove the line numbers from the links so any future change in `types.ts` does not break this doc.

aryamohanan:
@david-luna Good idea. All three links in now point to types.ts without line anchors, they'll always navigate to the file regardless of future line shifts.

maryliag (approved):

aryamohanan:
Could someone please help merge this when you have a chance? Thanks!

PR #3643, opened 2026-07-29
fix(examples/mysql): initialize tracing before requiring instrumented modules

### LINKED ISSUE #1795: [instrumentation-mysql] example not working
Looks like the example for instrumentation-mysql is not working correctly. I've tried it out but I couldn't get traces to show up in Jaeger. 
________________________
> I tired it out but couldn't get traces to show up in jaeger. I don't think this is related to updating grafana though, more likely that there's some misconfiguration in the example. :thinking:

_Originally posted by @pichlermarc in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1674#pullrequestreview-1634978200_

emmanuel-adu:

Fixes #1795.

The MySQL example initializes tracing after importing `mysql` and `http`. Because `ts-node` compiles the files to CommonJS, those modules are loaded before the instrumentation hooks are registered. By that point, the modules are already cached and cannot be patched.

As a result, the server prints `traceid: undefined`, and no traces are sent to Zipkin.

Moved the `setupTracing` import and call above the `mysql` and `http` imports in:

- `src/server.ts`
- `src/client.ts`

This makes sure tracing is initialized before the modules it needs to instrument are loaded.

## Testing

I tested the change locally using the Docker Compose stac with MySQL, Zipkin, and the OTel Collector. 

I then ran:

- `npm run zipkin:server`
- `npm run zipkin:client`

Before the change, the server printed `traceid: undefined`, and no trace was available in Zipkin. 

After the change, the server printed valid trace IDs. Zipkin showed the full trace across the client, server, and MySQL calls for all five example endpoints. 

## Screenshots

**Before:** No trace

before: No trace found

**After:** full trace with client, server, and mysql spans

after: full trace waterfall

### DIFF
--- a/examples/mysql/src/client.ts
+++ b/examples/mysql/src/client.ts
@@ -3,11 +3,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import * as api from '@opentelemetry/api';
-import * as http from 'http';
 import { setupTracing } from './tracer';
 
 const tracer = setupTracing('example-mysql-client');
+
+import * as api from '@opentelemetry/api';
+import * as http from 'http';
 
 /** A function which makes requests and handles response. */
 function makeRequest() {
--- a/examples/mysql/src/server.ts
+++ b/examples/mysql/src/server.ts
@@ -3,13 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { setupTracing } from './tracer';
+
+setupTracing('example-mysql-server');
+
 import * as api from '@opentelemetry/api';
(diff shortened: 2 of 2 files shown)


david-luna (approved):
Thanks for the fix @emmanuel-adu :)

emmanuel-adu:
@david-luna no prob. All tests are passing, I don't see button a merge. Is that typically handled by a maintainer?

david-luna:
@emmanuel-adu indeed the button is available for specific roles. If in a few hours nobody requests changes I'll merge it myself

PR #3648, opened 2026-07-30
fix(instrumentation-runtime-node): prevent unbounded PerformanceObserver buffer in GCCollector

js62789:

`GCCollector` calls `this._observer.observe({ entryTypes: ['gc'] })` without `buffered: false`. When `buffered` is omitted it defaults to `true`, so Node.js replays all GC entries accumulated since the last observation start into the observer's internal `#buffer`. This buffer is never drained, causing unbounded heap growth proportional to the number of GC cycles — a positive feedback loop under load (more heap pressure → more GC → more buffer entries → more heap pressure).

- Pass `{ entryTypes: ['gc'], buffered: false }` to `PerformanceObserver.observe()` in `GCCollector.internalEnable()` so that historical entries are not buffered; the observer callback processes each entry immediately as it arrives.
- Add a unit test that stubs `PerformanceObserver` and asserts the `observe` call includes `buffered: false`.

## Notes

- The observer callback already processes only `list.getEntries()[0]` per invocation, so there is no dependency on historical replay.
- Setting `buffered: false` on `'gc'` entry types is safe: GC events are fire-and-forget metrics and do not require backfill on observation start.

### DIFF
--- a/packages/instrumentation-runtime-node/src/metrics/gcCollector.ts
+++ b/packages/instrumentation-runtime-node/src/metrics/gcCollector.ts
@@ -59,7 +59,7 @@ export class GCCollector extends BaseCollector {
   }
 
   internalEnable(): void {
-    this._observer.observe({ entryTypes: ['gc'] });
+    this._observer.observe({ entryTypes: ['gc'], buffered: false });
   }
 
   internalDisable(): void {
--- a/packages/instrumentation-runtime-node/test/gc.test.ts
+++ b/packages/instrumentation-runtime-node/test/gc.test.ts
@@ -4,12 +4,39 @@
  */
 import * as assert from 'assert';
 import * as sinon from 'sinon';
+import * as perf_hooks from 'node:perf_hooks';
 import { Meter } from '@opentelemetry/api';
 
 import { GCCollector } from '../src/metrics/gcCollector';
 import { METRIC_V8JS_GC_DURATION } from '../src/semconv';
 
 describe('GCCollector', function () {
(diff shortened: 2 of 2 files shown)


pacocartones:
Thanks for the detailed write-up. Could you share the reproduction or a heap profile showing the unbounded growth? `PerformanceObserverInit.buffered` defaults to false per the Performance Timeline spec, and on Node v24.14.1 an observer started with `observe({entryTypes: ['gc']})` receives no historical entries — I checked with 50 forced GC cycles before `observe()` and only the post-observe cycle was delivered. If that holds, the omitted `buffered` in gcCollector.ts:62 is not buffering anything and the leak mechanism may be elsewhere. Also: does the new test fail on main, or does it only assert that the option is passed?

(Analysis verified locally; comment drafted with AI assistance.)

js62789:
Thank you for checking this, you're right on both counts.

I verified on Node v20.18.3:

- buffered:true  — pre:3 GC cycles, post:1, delivered: 1
- buffered:false — pre:3 GC cycles, post:1, delivered: 1

`buffered` defaults to `false` per the Performance Timeline spec, and Node.js honors that for `gc` entries. Pre-observe entries are never replayed regardless. The change is a no-op and the buffer accumulation mechanism I described is incorrect.

The test also only asserts that the option is passed through — it would not fail on `main`. That's not a meaningful regression guard.

I'll close this PR. The heap growth I was investigating is more likely due to a separate AbortController leak in the application layer (not in this package). Sorry for the noise.

PR #3653, opened 2026-08-04
fix(instrumentation-user-interaction): preserve bare event listener calls

### LINKED ISSUE #3639: `@opentelemetry/auto-instrumentations-web` Crashing with "Invalid value used as weak map key"

2.10.0

24.18.0

### What did you do?

After upgrading `@opentelemetry/auto-instrumentations-web` from `0.65.0` to `0.66.0`, our instrumentation crashed our React application with the error "Invalid value used as weak map key".

We're using the `@honeycombio/opentelemetry-web` package and calling `getWebAutoInstrumentations` from `@opentelemetry/auto-instrumentations-web`, with `@opentelemetry/instrumentation-user-interaction` enabled.

When Honeycomb initializes, it uses the `web-vitals` package to register handlers, like `onCLS`, which then eventually call down to code that calls `addEventListener`. However, the `@opentelemetry/instrumentation-user-interaction` code is doing a patch of `addEventListener` with its own implementation. That patch is now using a `WeakMap` as of version `0.65.0` (see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3576).

When I added debugging logs to the code, the `_patchAddEventListener` function in `instrumentation.ts` was using `this` as the `on` parameter to `addPatchedListener`, which is typed as an `HTMLElement`. What I found was that at runtime, this value was set to `undefined` instead, which caused the `WeakMap` to throw the error about an invalid key.

Our React app rendering

### What did you see instead?

A blank page with console errors saying "Invalid value used as weak map key".

### Additional context

I'm using `@honeycombio/opentelemetry-web` version 1.3.1 and it's using `web-vitals` 5.3.0. I tried overriding to use `web-vitals` 6.0.0 but it didn't have any effect.

pacocartones:

Fixes #3639.

`@opentelemetry/instrumentation-user-interaction` patches `EventTarget.prototype.addEventListener` in the no-Zone path. A bare global call such as the one used by `web-vitals` has an `undefined` receiver in the strict-mode wrapper. The wrapper then tried to use that receiver as a `WeakMap` key before delegating to the native method, producing `TypeError: Invalid value used as weak map key` and preventing the application from loading.

- Delegate immediately to the original listener method when the receiver cannot be a `WeakMap` key, preserving native behavior for bare calls and other invalid receivers.
- Keep support for non-registered `Symbol` keys on runtimes that implement them; feature-detect that capability so engines that do not yet accept Symbol keys also fall back safely.
- Add browser regressions for the real bare-global `addEventListener` path and for the receiver boundary.

## Validation

- Reproduced red in Chrome Headless 150: the new bare-global regression failed with `Invalid value used as weak map key` at `WeakMap.set`.
- Green after the change: `npm run test:browser -w @opentelemetry/instrumentation-user-interaction` — **45/45** in Chrome Headless 150.
- `npm run compile -w @opentelemetry/instrumentation-user-interaction`
- `npx prettier --check` on both touched TypeScript files
- `npx eslint` on both touched files (no new errors; existing package warnings remain)
- `git diff --check`

The issue maintainer also independently reproduced the original error in a browser and requested the Symbol detail handled here.

## AI disclosure

Added after the fact, for consistency with the OpenTelemetry Generative AI policy and with my other PR in this repository (#3661), which carries the same note. It was an omission, not a claim that no assistance was used.

I used an AI coding assistant (Claude) while working on this change: to trace how the patched `addEventListener` reaches the bare-global path, to draft the implementation and the browser regressions, and to help draft this description. I ran every command listed under Validation myself, reviewed every line, and the reasoning about the receiver boundary is my own. This text is written in my own words.

### DIFF
--- a/packages/instrumentation-user-interaction/src/instrumentation.ts
+++ b/packages/instrumentation-user-interaction/src/instrumentation.ts
@@ -29,6 +29,25 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 const ZONE_CONTEXT_KEY = 'OT_ZONE_CONTEXT';
 const EVENT_NAVIGATION_NAME = 'Navigation:';
 const DEFAULT_EVENT_NAMES: EventName[] = ['click'];
+const SUPPORTS_UNREGISTERED_SYMBOL_WEAK_MAP_KEYS = (() => {
+  try {
+    const key = Symbol();
+    new WeakMap().set(key as unknown as object, true);
+    return true;
+  } catch {
+    return false;
+  }
+})();
+
+function isWeakMapKey(value: unknown): boolean {
+  return (
+    (typeof value === 'object' && value !== null) ||
+    typeof value === 'function' ||
+    (typeof value === 'symbol' &&
+      Symbol.keyFor(value) === undefined &&
+      SUPPORTS_UNREGISTERED_SYMBOL_WEAK_MAP_KEYS)
+  );
+}
 
 function defaultShouldPreventSpanCreation() {
   return false;
@@ -277,7 +296,7 @@ export class UserInteractionInstrumentation extends InstrumentationBase<UserInte
         useCapture?: boolean | AddEventListenerOptions
       ) {
         // Forward calls with listener = null
-        if (!listener) {
+        if (!listener || !isWeakMapKey(this)) {
           return original.call(this, type, listener, useCapture);
         }
 
--- a/packages/instrumentation-user-interaction/test/userInteraction.nozone.test.ts
+++ b/packages/instrumentation-user-interaction/test/userInteraction.nozone.test.ts
@@ -716,6 +716,52 @@ describe('UserInteractionInstrumentation', () => {
       document.removeEventListener('click', listener, null);
     });
 
+    it('should preserve bare global addEventListener calls', () => {
+      // web-vitals calls the global event-listener function without a receiver.
+      // The native browser implementation accepts that form, so instrumentation
+      // must not use the undefined receiver as a WeakMap key first.
+      const { addEventListener, removeEventListener } = window;
+      let calls = 0;
+      const listener = () => {
+        calls++;
+      };
+
+      addEventListener('open-telemetry-bare-listener', listener);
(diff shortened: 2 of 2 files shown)


packages/instrumentation-user-interaction/src/instrumentation.ts
```diff
@@ -29,6 +29,25 @@ import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 const ZONE_CONTEXT_KEY = 'OT_ZONE_CONTEXT';
 const EVENT_NAVIGATION_NAME = 'Navigation:';
 const DEFAULT_EVENT_NAMES: EventName[] = ['click'];
+const SUPPORTS_UNREGISTERED_SYMBOL_WEAK_MAP_KEYS = (() => {
```
trentm, line 32: Interesting. TIL that Node.js <20.1 did not support symbols as keys to WeakMap.

For Browsers: https://caniuse.com/mdn-javascript_builtins_weakset_symbol_as_keys
I don't know whether that means that means "Baseline Widely available" or not yet.
In any case, this LGTM. Perhaps eventually this guard can be removed to slightly simplify the code.

trentm (approved):
LGTM, though I'm inclined to have a Browser reviewer (@david-luna ?) take a look as well.

packages/instrumentation-user-interaction/src/instrumentation.ts
pacocartones: Keeping the guard is intentional for the current support matrix: this package still declares Node `^18.19.0 || >=20.6.0`, and Node 18 does not accept Symbols as WeakMap keys. The runtime probe lets newer runtimes accept valid unregistered Symbols without breaking the Node 18 path.

Once the supported runtimes all provide Symbol WeakMap keys, this can be simplified by removing the probe and the Symbol-specific branch.

david-luna: > In any case, this LGTM. Perhaps eventually this guard can be removed to slightly simplify the code.

The instrumentation is meant to work only in browser but nothing prevent a user to install int in a nodejs env. So fr now we need it.

david-luna (approved):

PR #3657, opened 2026-08-06
feat(instrumentation-oracledb): add database client metrics

prajalg:

Adds metrics support for `@opentelemetry/instrumentation-oracledb`.

This introduces Oracle DB client metrics aligned with OpenTelemetry database semantic conventions:

- `db.client.connection.count`
- `db.client.operation.duration`
- `db.client.connection.pending_requests`
- `db.client.connection.timeouts`

This PR adds metric collection to `instrumentation-oracledb` using node-oracledb trace handler callbacks.

Main changes:

- Adds `OracleTelemetryTraceMetricHandler`, extending the existing trace handler behavior with metric callbacks.
- Records pool connection metrics from node-oracledb pool lifecycle hooks such as acquire, release, wait, timeout, shrink, expand, and close.
- Adds metric helpers for tracking pool state deltas and operation duration.
- Uses `connectionsCounterState` map which stores counts as {idle, pending, used, timeouts} for each `poolName` so that it also supports instrumented application to have multiple pools (each with unique pool alias). Maintaining this map also reconciles if any pool changes that were not recorded, such as the changes made while instrumentation was disabled, ensuring metric values remain accurate after instrumentation is enabled again.
- Adds semantic convention constants for Oracle DB client metrics.
- Adds tests covering pool connection count, pending requests, timeouts, operation duration, and instrumentation enable/disable behavior.
- Updates `oracledb` test dependency to `^6.7.0` and `@types/oracledb` to `6.10.4`.

The pool metrics are recorded by reading the latest pool statistics and publishing deltas through OpenTelemetry metric instruments.


package-lock.json
```diff
@@ -11919,9 +11919,9 @@
       "license": "MIT"
     },
     "node_modules/@types/oracledb": {
-      "version": "6.5.2",
-      "resolved": "https://registry.npmjs.org/@types/oracledb/-/oracledb-6.5.2.tgz",
-      "integrity": "sha512-kK1eBS/Adeyis+3OlBDMeQQuasIDLUYXsi2T15ccNJ0iyUpQ4xDF7svFu3+bGVrI0CMBUclPciz+lsQR3JX3TQ==",
+      "version": "6.10.4",
```
sharadraju, line 11922: Should this be 6.10.4 or the 7.0.x version

packages/instrumentation-oracledb/package.json
```diff
@@ -51,19 +51,20 @@
...
     "@opentelemetry/contrib-test-utils": "^0.68.0",
     "@opentelemetry/sdk-trace": "^2.9.0",
     "oracledb": "^6.7.0"
   },
   "dependencies": {
+    "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/instrumentation": "^0.221.0",
     "@opentelemetry/semantic-conventions": "^1.34.0",
-    "@types/oracledb": "6.5.2"
+    "@types/oracledb": "^6.10.4"
```
sharadraju, line 67: Should this version be greater than 7.0.x?

package-lock.json
prajalg: I tested with @types/oracledb 7.0.2, but encountered an error:
`Cannot find lib definition for 'esnext.disposable'`, seems it is not in TS 5.0.4 which the root repository uses.
Looks like it needs some typescript update, so decided to avoid updating to latest types.

packages/instrumentation-oracledb/package.json
prajalg: same reason of use of `esnext.disposable` in newer types.
Repository TS version should be updated first to support 7.0.2 types oracledb version

packages/instrumentation-oracledb/src/metricUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+let connectionsCount!: UpDownCounter;
+let connectionPendingRequests!: UpDownCounter;
+let connectionsTimeouts!: Counter;
+
+// Pool properties provide absolute values, but synchronous counters can only be
+// updated by adding deltas. Store the last recorded values for each pool to
+// calculate the correct delta on every pool event. On the next pool event, this
+// also reconciles any pool changes that were not recorded, such as changes made
+// while instrumentation was disabled, ensuring metric values remain accurate.
+const connectionsCounterState: Record<string, PoolConnectionsCounter> = {};
```
sudarshan12s: a weak map probably may be better to automatically cleanup as the pools are closed/freed without receiving a callback (like when tracing is disabled after enable)  into this module.

`const connectionsCounterState = new WeakMap<oracleDBTypes.Pool, PoolConnectionsCounter>();`

prajalg: Done. Thanks!

packages/instrumentation-oracledb/test/oracle.test.ts
```diff
@@ -866,7 +863,8 @@ describe('oracledb', () => {
       instrumentation.enable();
       await context.with(trace.setSpan(context.active(), span), async () => {
         await new Promise<void>(resolve => {
-          pool.getConnection((err, conn) => {
+          pool.getConnection((...args) => {
+            const [err, conn] = args;
```
prajalg, line 867: `@types/oracledb@6.10.4` defines `ResultCallback<T>` as `(...args: [DBError] | [null, T]) => void`. A callback therefore receives either one error argument or two arguments for success. Our tests previously used `(err, result)`, which assumes there are always two arguments. Using `(...args)`  and destructuring the values works for both cases.

sudarshan12s (approved):
LGTM

sharadraju:
@maryliag Can you help us with the review here?

packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -352,14 +386,48 @@ export function getOracleTelemetryTraceHandlerClass(
           message: traceContext.error.message,
         });
       }
+
+      // Builds the attribute set used for execute duration metrics.
+      const isBatch = traceContext.operation === SpanNames.EXECUTE_MANY;
+      const metricsAttributes: Attributes = {
+        [ATTR_DB_SYSTEM_NAME]: DB_SYSTEM_NAME_VALUE_ORACLE_DB,
+        [ATTR_DB_NAMESPACE]: connAttrs[ATTR_DB_NAMESPACE],
```
maryliag: `connAttrs` can be `{}` from your change above, making some of these `undefined`. I think is cleaner to add them just when there is an actual value to add, so worth adding a check and then adding

packages/instrumentation-oracledb/test/oracle.metrics.test.ts
```diff
@@ -0,0 +1,997 @@
...
+  let metricReader: MetricReader;
+  let meterProvider: MeterProvider;
+  let metricsExporter: InMemoryMetricExporter;
+  let queueTimeout: number;
+  const testOracleDB = process.env.RUN_ORACLEDB_TESTS; // For CI: assumes local oracledb is already available
+  const testOracleDBLocally = process.env.RUN_ORACLEDB_TESTS_LOCAL; // For local: spins up local oracledb via docker
+  const shouldTest = testOracleDB || testOracleDBLocally; // Skips these tests if false (default)
+
+  async function initMeterProvider() {
+    metricsExporter = new InMemoryMetricExporter(
```
maryliag: this is not being used anywhere

maryliag (commented):
Thank you for working on this, I added just small nits

packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
prajalg: Sure, thanks. I added check for it now.

packages/instrumentation-oracledb/test/oracle.metrics.test.ts
prajalg: Removed. Thanks!

packages/instrumentation-oracledb/src/metricUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+  type UpDownCounter,
+  ValueType,
+} from '@opentelemetry/api';
+import {
+  hrTime,
+  hrTimeDuration,
+  hrTimeToMilliseconds,
+} from '@opentelemetry/core';
+import { METRIC_DB_CLIENT_OPERATION_DURATION } from '@opentelemetry/semantic-conventions';
+import * as oracleDBTypes from 'oracledb';
```
JacksonWeber: Referencing `POOL_STATUS_OPEN` here makes `oracledb` a runtime dependency. Since it’s only a dev dependency, I think this could break for users without the driver installed. Could we get the constant from the dynamically loaded module instead?

prajalg: Thanks! Yes we can get it from the dynamically loaded module. I did the change now

maryliag (approved):
as long as tests are passing, LGTM

prajalg:
Thanks @maryliag for reviewing!
Can you trigger the workflows again? Last run failed due to some PR unrelated error (failed to setup container for mysql tests)

maryliag:
cc @JacksonWeber who had also added comments, just checking if there are any other things still need addressing

JacksonWeber:
> cc @JacksonWeber who had also added comments, just checking if there are any other things still need addressing

LGTM, thanks for the heads up!

JacksonWeber (approved):

packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts
```diff
@@ -58,15 +66,36 @@ function getTraceHandlerBaseClass(
...
+  const trimmed = statement.trim();
+  let end = trimmed.length;
+  for (let i = 0; i < trimmed.length; i++) {
+    const code = trimmed.charCodeAt(i);
+    // Checks for space (32), tab (9), LF (10), VT (11), FF (12), CR (13)
+    if (code === 32 || (code >= 9 && code <= 13)) {
+      end = i;
+      break;
+    }
+  }
```
maryliag, line 97: if you wanna try something shorter, you could also try

```
function parseNormalizedOperationName(statement: string): string {
  const sqlCommand = (queryText.trim().match(/^\S+/)?.[0] ?? '').toUpperCase();
}

prajalg: Thanks! 
I just wanted to avoid regex here to minimize overhead and prevent any performance impact.

sudarshan12s:
Thanks @maryliag and @JacksonWeber for your time, feedback and for merging the PR!

Issue #3660, opened 2026-08-06
feat(instrumentation-net, instrumentation-dns): add `requireParentSpan` config option


segevfiner:
> [!NOTE]
> This issue was researched and written by Claude (an AI agent) on behalf of @segevfiner, who reviewed it before filing. The measurements below come from a real four-service Node.js deployment on `@opentelemetry/auto-instrumentations-node@0.79.0`.

## Is your feature request related to a problem?

`instrumentation-net` and `instrumentation-dns` record `tcp.connect` and `dns.lookup` unconditionally, including when no span is active. Anything that opens or refreshes a connection outside a request then becomes its own **single-span trace**, with nothing useful on it.

The most common source is a database driver's background monitoring. A MongoDB SDAM heartbeat refreshing a connection produces one `dns.lookup` trace and one `tcp.connect` trace every heartbeat interval, forever, on every service that talks to MongoDB. Sampled from our backend's traces:

```
count  root span name
   61  GET /api/codebase/:codebaseId
    6  dns.lookup            ← standalone, no parent
    4  GET /api/codebases
    4  GET /api/user
    3  tcp.connect           ← standalone, no parent
    2  GET /api/status
```

Inspecting one of those `dns.lookup` traces: a single root span from `@opentelemetry/instrumentation-dns`, with **no attributes at all**. Nothing identifies what was resolved, so the trace has no diagnostic value on its own — it is pure volume in the trace store and noise in any "recent traces" or service-graph view.

To be clear, these spans are valuable *nested* — DNS resolution and TCP setup latency under an outbound call is exactly what you want when a request is mysteriously slow. The problem is only the parentless case.

## Describe the solution you'd like

The `requireParentSpan` config option that several other instrumentations already have, so parentless connection spans can be skipped while nested ones are kept.

There is a lot of precedent for exactly this option, and it has been accepted repeatedly:

| instrumentation | issue / PR |
|---|---|
| mongodb | #2658 |
| pg | #1199, #1211 |
| knex | #2288 |
| ioredis | #372, #1151 |
| redis | #478 |

`instrumentation-mongodb` goes further and defaults it to `true`:

```js
const DEFAULT_CONFIG = { requireParentSpan: true };
```

Which is arguably the right default for connection-level instrumentations too, though defaulting to `false` and letting users opt in would already solve this.

## Describe alternatives you've considered

Both work, and both are blunter than the config option would be:

[...]

PR #3661, opened 2026-08-07
feat(instrumentation-net): add requireParentSpan config option

pacocartones:
# feat(instrumentation-net): add requireParentSpan config option

Part of #3660.

**Rescoped to `instrumentation-net` only** (force-push, 2026-08-07). This PR originally also touched
`instrumentation-dns`, which has no component owner in `.github/component_owners.yml`. That is what
matched `pkg-status:unmaintained` in `component-label-map.yml`, which in turn started the 14-day
autoclose timer in `close-stale.yml`. The `-dns` half is held back and will go in its own PR.

Nothing was rewritten: the removed hunks are exactly the `instrumentation-dns` ones. I verified it
mechanically — applying both halves onto `main` reproduces the previous tree byte for byte, and the
line counts add up (276+160 = the previous 436, 6+3 = the previous 9).

@seemk you are already assigned here, and `instrumentation-net` is yours, so this should now be
reviewable on its own merits.

**One thing I cannot do myself:** the three labels are stale — `pkg:instrumentation-dns`,
`pkg-status:unmaintained` and `pkg-status:unmaintained:autoclose-scheduled` no longer match the
diff, but `actions/labeler` runs without `sync-labels`, so it only ever adds. Could a maintainer
remove them? Otherwise the timer keeps running against a PR that no longer touches an unmaintained
package. `remove-stale-when-updated: false` means neither this comment nor the push resets it.

## Summary

`instrumentation-net` records its spans unconditionally, including when no span is active. A
connection opened outside of a request therefore becomes its own single-span trace, forever, on
every interval. The most common source is a database driver's background monitoring: a MongoDB SDAM
heartbeat produces one `tcp.connect` trace per heartbeat.

Nested under an outbound call these spans are genuinely useful - TCP setup latency is exactly what
you want when a request is slow. Only the parentless case is noise.

This adds the `requireParentSpan` option. When enabled and no span is active, the original function
is called directly and no span is created.

## Mechanism

The option already exists in many instrumentations here, so this follows the established shape: a
`DEFAULT_CONFIG` merged in the constructor and in `setConfig`, and a
`trace.getSpan(context.active()) === undefined` check (`mongodb`, `pg`, `knex`, `ioredis`, `redis`,
`dataloader`, `fs`, `oracledb`).

`instrumentation-net` had no config interface at all - it took a bare `InstrumentationConfig` - so
this adds `NetInstrumentationConfig` and exports it as a type from `src/index.ts`.
`auto-instrumentations-node` infers config types from the constructor's first argument, so it picks
the new option up with no change there.

### Why the check lives in `patchedConnect`

`instrumentation-net` creates four spans: `connect`, `ipc.connect`, `tcp.connect` and `tls.connect`.
The check sits once at the top of `patchedConnect` rather than at each `startSpan` call site,
because all four originate from that function and, on the TLS path, `_startSpan` runs inside
`context.with(trace.setSpan(active, tlsSpan))`:

```ts
const tlsSpan = this.tracer.startSpan('tls.connect');
const netSpan = context.with(
  trace.setSpan(context.active(), tlsSpan),
  () => this._startSpan(options, socket)
);
```

Skipping only the `tls.connect` span would leave a non-recording span in context,
`trace.getSpan` would return it, and the nested `tcp.connect` span would still be created
parentless. There is a test covering exactly this case.

### Why the default is `false`

The default is `false`, so nothing changes for existing users and the new behaviour is opt-in.

`instrumentation-mongodb` defaults it to `true`, but the majority of instrumentations here default
to `false` (`knex`, `redis`, `pg`, `dataloader`, `fs`, `oracledb`), and `false` seems right for this
package specifically:

- Flipping the default would silently remove spans that people receive today. That is a behaviour
  change, and it would need to be a `feat!`.
- The package ships in `auto-instrumentations-node`, so a `true` default would reach a large number
  of users who never opted into it.
- #3660 notes that opt-in already solves the reported problem.

Happy to change it to `true` if maintainers prefer that trade-off - it is a one-line change.

## Tests

New file `packages/instrumentation-net/test/require-parent-span.test.ts`, 7 tests. It covers all
four span creation sites with the option enabled and no parent (`connect`, `ipc.connect`,
`tcp.connect`, `tls.connect`), the nested cases with a parent active (asserting `tcp.connect` is
still parented to `tls.connect`), and the default configuration still producing a parentless
`tcp.connect` span.

Environment for every run below: Ubuntu 24.04, **Node 18.20.8** (one of the versions in the CI
matrix), clean `npm ci` from the repository root.

### Red - the test file alone, applied on top of unmodified `main`

```
$ npm test -w @opentelemetry/instrumentation-net

  NetInstrumentation requireParentSpan
    when requireParentSpan is true
      1) should not create a tcp.connect span without a parent span
      2) should not create an ipc.connect span without a parent span
      3) should not create a generic connect span without a parent span
      4) should not create tls.connect or tcp.connect spans without a parent span
      ✔ should create a tcp.connect span when a parent span is active
      ✔ should create tls.connect and tcp.connect spans when a parent span is active
    when requireParentSpan is not set
      ✔ should create a tcp.connect span without a parent span

  23 passing (136ms)
  4 failing

  1) NetInstrumentation requireParentSpan
       when requireParentSpan is true
         should not create a tcp.connect span without a parent span:

      Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
      + actual - expected

      + [
      +   'tcp.connect'
      + ]
      - []
```

The three tests that pass in the red run are the ones that assert existing behaviour - the two
parented cases and the default configuration - which is what you would expect from them.

### Green - this branch

```
$ npm test -w @opentelemetry/instrumentation-net

  NetInstrumentation requireParentSpan
    when requireParentSpan is true
      ✔ should not create a tcp.connect span without a parent span
      ✔ should not create an ipc.connect span without a parent span
      ✔ should not create a generic connect span without a parent span
      ✔ should not create tls.connect or tcp.connect spans without a parent span
      ✔ should create a tcp.connect span when a parent span is active
      ✔ should create tls.connect and tcp.connect spans when a parent span is active
    when requireParentSpan is not set
      ✔ should create a tcp.connect span without a parent span

  27 passing (102ms)
```

27 = 20 pre-existing + 7 new, 0 failing.

### Other CI gates, run locally on this branch

| Gate | Command | Result |
| --- | --- | --- |
| `test.yml` → compile | `npm run compile` | `NX Successfully ran target compile for 72 projects and 55 tasks they depend on` |
| `lint.yml` | `npm run lint` (Node 24) | exit 0 - `Successfully ran target lint for 84 projects`, `All matched files use Prettier code style!`, `lint:deps` (knip), `Successfully ran target lint:readme for 37 projects`, `lint:markdown`, `lint:semconv-deps` all clean |
| `peer-api.yaml` | `node ./scripts/peer-api-check.js` | exit 0, 70 packages `OK` |
| `test.yml` → unit-test, other affected project | `npm test -w @opentelemetry/auto-instrumentations-node` | `17 passing`, 0 failing |

`npm run compile` is `tsc -p .` per package and each package's `tsconfig.json` includes
`test/**/*.ts`, so the new test file is type-checked under `strict` by that gate too.

`nx affected -t test` for this branch resolves to `@opentelemetry/instrumentation-net` and
`@opentelemetry/auto-instrumentations-node`; both were run. Neither package has a `test:browser` or
a `test-all-versions` script, so those jobs have nothing to do here.

## What I did not verify

- I have not run the repository's full CI matrix. Tests were run on Node 18.20.8 only, not on
  20.6.0, 20, 22 or 24.
- I have no Docker on the machine I used, so I did not run anything that needs
  `test/docker-compose.yaml`. Neither of the two affected packages needs it.
- `zizmor-checks.yml` was not run - it needs an external binary and a GitHub token. This PR does not
  touch `.github/`.
- I did not reproduce the original MongoDB-heartbeat scenario end to end. The tests here are at the
  unit level.
- On **Node 24** the package test scripts fail before running anything, with
  `Directory import ... is not supported resolving ES modules`. That happens on unmodified `main`
  too, so it is unrelated to this change - Node 24's native type stripping loads the `.ts` files as
  ESM and the `ts-node` CJS hook never runs. `NODE_OPTIONS=--no-experimental-strip-types` works
  around it locally. I mention it only in case it is useful to others.

## AI disclosure

I used an AI coding assistant (Claude) while working on this change: to survey how
`requireParentSpan` is implemented across the other instrumentations in this repository, to draft
the implementation and the tests, and to help draft this description. I reviewed every line,
verified the behaviour myself with the runs described above, and the reasoning about the TLS nesting
and about the default value is my own. This text is written in my own words.

### DIFF
--- a/packages/instrumentation-net/README.md
+++ b/packages/instrumentation-net/README.md
@@ -29,12 +29,22 @@ const { NetInstrumentation } = require('@opentelemetry/instrumentation-net');
 
 const sdk = new NodeSDK({
   instrumentations: [
-    new NetInstrumentation(),
+    new NetInstrumentation({
+      // see below for available configuration
+    }),
   ],
 });
 sdk.start();
 process.once('beforeExit', async () => { await sdk.shutdown(); });
 ```
+
+### Net Instrumentation Options
+
+You can set the following:
+
+| Options | Type | Description |
+| ------- | ---- | ----------- |
+| `requireParentSpan` | `boolean` | Require parent to create net span, default when unset is `false`. Connections opened outside of a request - a database driver's background heartbeat, for example - otherwise each become a standalone single-span trace. |
 
 ## Semantic Conventions
 
--- a/packages/instrumentation-net/src/index.ts
+++ b/packages/instrumentation-net/src/index.ts
@@ -5,3 +5,4 @@
 
 export { NetInstrumentation } from './instrumentation';
 export { TLSAttributes } from './types';
+export type { NetInstrumentationConfig } from './types';
--- a/packages/instrumentation-net/src/instrumentation.ts
+++ b/packages/instrumentation-net/src/instrumentation.ts
@@ -12,7 +12,6 @@ import {
 } from '@opentelemetry/api';
 import {
   InstrumentationBase,
-  InstrumentationConfig,
   InstrumentationNodeModuleDefinition,
   isWrapped,
   safeExecuteInTheMiddle,
@@ -26,7 +25,7 @@ import {
   ATTR_SERVER_PORT,
   NETWORK_TRANSPORT_VALUE_TCP,
 } from '@opentelemetry/semantic-conventions';
-import { TLSAttributes } from './types';
+import { NetInstrumentationConfig, TLSAttributes } from './types';
 import { NormalizedOptions, SocketEvent } from './internal-types';
 import { getNormalizedArgs, STABLE_IPC_TRANSPORT_VALUE } from './utils';
 /** @knipignore */
@@ -35,9 +34,17 @@ import { Socket } from 'net';
 import { TLSSocket } from 'tls';
 import type * as net from 'net';
 
-export class NetInstrumentation extends InstrumentationBase {
-  constructor(config: InstrumentationConfig = {}) {
-    super(PACKAGE_NAME, PACKAGE_VERSION, config);
+const DEFAULT_CONFIG: NetInstrumentationConfig = {
+  requireParentSpan: false,
+};
+
+export class NetInstrumentation extends InstrumentationBase<NetInstrumentationConfig> {
+  constructor(config: NetInstrumentationConfig = {}) {
+    super(PACKAGE_NAME, PACKAGE_VERSION, { ...DEFAULT_CONFIG, ...config });
+  }
+
+  override setConfig(config: NetInstrumentationConfig = {}) {
+    super.setConfig({ ...DEFAULT_CONFIG, ...config });
   }
 
   init(): InstrumentationNodeModuleDefinition[] {
@@ -69,6 +76,18 @@ export class NetInstrumentation extends InstrumentationBase {
     return (original: (...args: unknown[]) => void) => {
       const plugin = this;
       return function patchedConnect(this: Socket, ...args: unknown[]) {
+        // Checked once here rather than at each `startSpan` call site: every
+        // span this instrumentation creates (`connect`, `ipc.connect`,
+        // `tcp.connect` and `tls.connect`) originates from this function, and
+        // `tcp.connect` is parented to `tls.connect` on the TLS path, so a
+        // per-call-site check would still let the nested span through.
+        if (
+          plugin.getConfig().requireParentSpan &&
+          trace.getSpan(context.active()) === undefined
+        ) {
+          return original.apply(this, args);
+        }
+
         const options = getNormalizedArgs(args);
 
         const span =
--- a/packages/instrumentation-net/src/types.ts
+++ b/packages/instrumentation-net/src/types.ts
@@ -3,6 +3,20 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { InstrumentationConfig } from '@opentelemetry/instrumentation';
+
+export interface NetInstrumentationConfig extends InstrumentationConfig {
+  /**
+   * Require a parent span in order to create net spans, default when unset is
+   * `false`.
+   *
+   * Connections opened outside of a request - a database driver's background
+   * heartbeat, for example - would otherwise each produce a standalone
+   * single-span trace.
+   */
+  requireParentSpan?: boolean;
+}
+
 /* The following attributes are not official, see open-telemetry/opentelemetry-specification#1652 */
 export enum TLSAttributes {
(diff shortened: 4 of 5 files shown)


pacocartones:
Rescoped to `instrumentation-net` only, in `e398744a`. Details in the updated description.

Short version: this PR also touched `instrumentation-dns`, which has no entry in `component_owners.yml`. That is what matched `pkg-status:unmaintained` and started the 14-day autoclose timer. The `-dns` half is held back for a separate PR.

Verified the split loses nothing: applying both halves onto `main` reproduces the previous tree byte for byte, and 276+160 = the previous 436 lines added.

Re-ran the gates locally on the reduced branch (Linux, clean clone): `npm run compile` across 72 projects, `npm run lint` across 84, `peer-api-check` 70/70, and `npm test -w @opentelemetry/instrumentation-net` **27 passing** on Node 18.20.8. `nx affected` flags `instrumentation-net` and `auto-instrumentations-node`; both green. Against `main` with only the test file applied, the four parentless cases fail as they should.

**Ask:** the three labels are now stale and I cannot remove them myself. `actions/labeler` runs without `sync-labels`, so it only adds. Could someone drop `pkg:instrumentation-dns`, `pkg-status:unmaintained` and `pkg-status:unmaintained:autoclose-scheduled`? `close-stale.yml` has `remove-stale-when-updated: false`, so nothing I do resets the clock.

**Not verified:** the heavy CI has never run on this PR — `gh pr checks` shows only labeler, auto-assign, EasyCLA, title validation and welcome. It is still waiting on first-time-contributor approval, so everything above is local.

pacocartones:
Friendly reminder before the scheduled autoclose: this PR has been `instrumentation-net` only since `e398744a`, but the stale `pkg:instrumentation-dns`, `pkg-status:unmaintained`, and `pkg-status:unmaintained:autoclose-scheduled` labels are still attached.

Could a maintainer remove those three labels or confirm what else is needed to keep the net-only PR open for review? There have been no code changes since the rescope; the request is only to align the labels and timer with the current diff.

pacocartones:
@pichlermarc sorry for the direct ping, and this is only a label request, not a request for review priority.

seemk (approved):
Looks reasonable, I can't approve the workflows to run though.

@pichlermarc Need your help with it

PR #3664, opened 2026-08-09
feat(instrumentation-anthropic): add basic messages instrumentation.

eternalcuriouslearner:

- The Anthropic JavaScript SDK does not currently have built-in OpenTelemetry instrumentation in this repository. This PR introduces an initial instrumentation based on the OpenInference donation.

- Add the donated @opentelemetry/instrumentation-anthropic package as a focused initial implementation.
- Instrument anthropic.messages.create(), including streaming requests and the anthropic.messages.stream() helper.
- Create basic GenAI client spans with provider, operation, and requested-model attributes.
- Add integration tests using recorded HTTP cassettes.
- Include Anthropic instrumentation in @opentelemetry/auto-instrumentations-node.

## Thoughts

If maintainers feel that this pr is bulky, I can reduce the verbosity by cutting this into smaller PRs.

P.S: LLM help was taken while writing this code.


JacksonWeber:
Could we wrap the stream’s internal iterator method instead of `Symbol.asyncIterator`? `stream.tee()` bypasses this wrapper, so the span may never end. The OpenAI instrumentation handles this the same way and has a `tee()` test.

packages/instrumentation-anthropic/test/mock-responses/.gitkeep
psx95, line 1: nit: delete file

packages/instrumentation-anthropic/src/instrumentation.ts
```diff
@@ -0,0 +1,176 @@
...
+
+        let result: Promise<unknown>;
+        try {
+          result = context.with(ctx, () => original.apply(this, args));
+        } catch (error) {
+          instrumentation._endSpanWithError(state, error);
+          throw error;
+        }
+
+        result.then(
```
JacksonWeber, line 107: Does calling `.then()` here consume the response body before callers can use `.asResponse()`? If so, could we avoid eagerly parsing it and add a test for that case?

packages/instrumentation-anthropic/src/instrumentation.ts
```diff
@@ -0,0 +1,391 @@
...
+        try {
+          next = await iterator.next();
+        } finally {
+          state.awaitingNext = false;
+        }
+        if (next.done) {
+          exhausted = true;
+          break;
+        }
+        yield next.value;
```
JacksonWeber, line 329: Could we check `signal.aborted` before yielding here? An abort during a pending `next()` can still deliver a buffered event. If the caller stops reading afterward, the span never ends.

packages/instrumentation-anthropic/src/instrumentation.ts
```diff
@@ -0,0 +1,415 @@
...
+      }
+      // `Stream.fromSSEResponse` swallows abort errors and simply stops
+      // yielding, so an aborted generation is indistinguishable from a
+      // completed one without checking the signal.
+      if (signal?.aborted) {
+        this._endSpanWithAbort(state);
+      } else {
+        this._endSpan(state);
```
JacksonWeber, line 361: Could we also capture errors from `messages.stream()` itself? If the response ends before `message_stop`, `finalMessage()` rejects, but this span has already ended without an error.

packages/instrumentation-anthropic/src/instrumentation.ts
eternalcuriouslearner: Yep it consumes the response body. good catch. I removed it.

packages/instrumentation-anthropic/src/instrumentation.ts
eternalcuriouslearner: good call, done!!

JacksonWeber (approved):

PR #3667, opened 2026-08-11
fix(instrumentation-openai): avoid consuming APIPromise response twice

### LINKED ISSUE #3586: instrumentation-openai:  breaks OpenAI parse and LangChain withStructuredOutput
* @opentelemetry/auto-instrumentations-node:  0.77.0
* @opentelemetry/instrumentation-openai:  0.17.0

*  v24.14.1

### What did you do?
Try to use instrumentation with a [JSON structured output](https://developers.openai.com/api/docs/guides/structured-outputs).

Success

### What did you see instead?
Instrumentation throws an error and parsing fails.
```
Run failed: TypeError: Body is unusable: Body has already been read
    at consumeBody (node:internal/deps/undici/undici:6961:31)
    at _Response.json (node:internal/deps/undici/undici:6912:18)
    at /home/hien/slim_dev/parse-error/node_modules/openai/internal/parse.js:36:41
    at APIPromise.defaultParseResponse [as parseResponse] (/home/hien/slim_dev/parse-error/node_modules/openai/internal/parse.js:41:7)
    at APIPromise.parseResponse (/home/hien/slim_dev/parse-error/node_modules/openai/core/api-promise.js:26:188)
    at /home/hien/slim_dev/parse-error/node_modules/openai/core/api-promise.js:60:75
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async runParse (/home/hien/slim_dev/parse-error/main.cjs:13:3)
    at async main (/home/hien/slim_dev/parse-error/main.cjs:44:3)
```

### Additional context
Test case attached, example usage:
```
OPENAI_API_KEY=... npm run no-otel
OPENAI_API_KEY=... npm run otel
```

[parse-error-repro.tar.gz](https://github.com/user-attachments/files/29529212/parse-error-repro.tar.gz)

pacocartones:

Fixes #3586.

The instrumentation observed non-streaming OpenAI `APIPromise` objects through their public `.then()` path. That could trigger the SDK's lazy response parsing before helpers such as `chat.completions.parse()` created their derived promise, causing the same response body to be read again.

- Preserve and return the SDK's original `APIPromise` without subscribing through its public promise methods.
- Coordinate the SDK's parser and raw-response paths so structured parsing, `runTools`, `withResponse()`, ordinary `create()`, raw responses, and errors finalize telemetry once without consuming the caller-owned response.
- Bound raw-response telemetry parsing to declared JSON bodies of at most 1 MiB; responses without a usable length or above that limit finish with request-level telemetry instead.
- Add regression coverage for structured output, repeated observers, fire-and-forget calls, HTTP errors, bounded raw responses, and defensive private-shape handling.

The supported OpenAI range remains `>=4.19.0 <7`. The implementation defensively checks the SDK internals used by that range. A call whose response settles before a delayed consumer starts parsing can take the bounded raw-response telemetry path first; that preserves application behavior but may decode a response of at most 1 MiB again when the caller later parses the original body.

### Validation

- `@opentelemetry/instrumentation-openai` test suite: 72 passing
- TypeScript compile: passing
- lint: passing (two pre-existing warnings in untouched `src/responses.ts`)
- focused compatibility controls passed on representative OpenAI 4.x and 6.x versions

The repository's native `test-all-versions` launcher could not start locally on Node 24/Windows because its `npm.cmd` child process exited with `EINVAL`; the PR CI matrix remains authoritative.

AI assistance was used during investigation and drafting. I reviewed the implementation and verified the reported local results.


packages/instrumentation-openai/src/instrumentation.ts
```diff
@@ -266,26 +269,323 @@ export class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumenta
...
+      }
+    };
+    const safeError = (message: string, err: unknown) => {
+      try {
+        this._diag.error(message, err);
+      } catch {
+        // Diagnostics must not affect the caller.
+      }
+    };
+    const replaceProperty = (name: string, replacement: unknown): boolean => {
```
david-luna, line 347: question: I think this function is basically doing the same `shimmer` does because its used later to replace the parse with a wrapper function. Is there any specific reason `instrumentation._wrap()` was not used?

packages/instrumentation-openai/src/instrumentation.ts
```diff
@@ -266,26 +269,323 @@ export class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumenta
...
+      } catch (err) {
+        safeError('unexpected error recording OpenAI telemetry:', err);
+      }
+    };
+    const finishWithError = (err: unknown) => {
+      finish(() => onError(err as Error));
+    };
+
+    let originalParseResponse: unknown;
+    try {
```
david-luna, line 404: I think the `if` below already covers the case where we access an undefined property so maybe this try is not necessary. Unless you expect undefined values for the `apiPromise` param. If so you should add the indefined type and this expression could be written into

```js
const originalParseResponse = apiPromise?.parseResponse;

// undefined if apiPromise is undefined or property doesn't exist
if (typeof originalParseResponse === 'function') {}
```

packages/instrumentation-openai/src/instrumentation.ts
```diff
@@ -266,26 +269,323 @@ export class OpenAIInstrumentation extends InstrumentationBase<OpenAIInstrumenta
...
+      };
+
+      if (!replaceProperty('parseResponse', wrappedParseResponse)) {
+        safeWarn('unable to patch OpenAI APIPromise parseResponse');
+      }
+    } else {
+      safeWarn('OpenAI APIPromise parseResponse is unavailable');
+    }
+
+    const onRawResponse = async (response: unknown): Promise<void> => {
```
david-luna: This function does a lot of casts of the `response` param and its properties to the [Response type](https://developer.mozilla.org/en-US/docs/Web/API/Response). I wonder if possible just to set the signature to.

```js
// NOTE: adding also `null` type since all the type casting is doing it (but maybe is not necessary)
const onRawResponse = async (response: Response | null): Promise<void> => {}
```

Then you can leverage the type and not use Reflect to call methods of the response object.

david-luna:
@pacocartones there are also some test failures. could you check them?

packages/instrumentation-openai/src/instrumentation.ts
pacocartones: The happy path is similar, but `_wrap()` delegates to shimmer, which assumes the target is a writable function property. It reads the property multiple times and `Object.defineProperty` can throw for an accessor, read-only property, or proxy.

Here we are patching a private, per-request SDK field, and instrumentation must degrade without changing `create()` settlement. `replaceProperty` preserves the existing descriptor and turns inspection or replacement failures into a no-op. Commit 585eeb3 documents that distinction and adds a throwing-getter regression.

packages/instrumentation-openai/src/instrumentation.ts
pacocartones: The `try` is for a throwing accessor or proxy rather than an undefined property. `apiPromise` is non-null here, and optional chaining would still evaluate `parseResponse` and propagate a throwing getter. The `typeof` check below already handles an absent or non-function value.

Commit 585eeb3 adds an explanatory comment and a regression proving that a throwing `parseResponse` getter does not escape from `create()`.

packages/instrumentation-openai/src/instrumentation.ts
pacocartones: Implemented in 585eeb3. The private-boundary value is now narrowed once to a minimal `OpenAIRawResponse` shape, and the implementation uses direct `headers.get()`, `clone()`, and `json()` calls while retaining the failure-isolation blocks.

I avoided `instanceof Response` because the supported OpenAI versions can use different Fetch implementations; the structural boundary preserves that compatibility.

pacocartones:
I checked the failing jobs and addressed them in 585eeb3.

The regular unit matrix was already passing on Node 18, 20, 22, and 24. The TAV failures were limited to OpenAI 4.104 and came from two version-specific test assumptions: the clone spy targeted the global Undici `Response` while OpenAI v4 uses a different Fetch response, and invalid JSON is reported as `FetchError` rather than `SyntaxError` there.

The tests now verify the runtime-independent contracts: the integration path still exports complete fire-and-forget telemetry, a deterministic structural probe verifies exactly one bounded clone, the original SDK parse error reaches the caller, and span `error.type` matches that runtime error. OpenAI 4.104 focused controls pass 4/4; OpenAI 6.49 full package passes 74/74; compile, ESLint, Prettier, and diff checks pass.

PR #401, opened 2021-03-31
chore: 0.15.0 release proposal

dyladan:
#### :rocket: Enhancement
* Other
  * [#366](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/366) Add automated release workflows ([@willarmiros](https://github.com/willarmiros))
  * [#391](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/391) chore: adding meta package for web ([@obecny](https://github.com/obecny))
  * [#393](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/393) feat: mysql instrumentation ([@dyladan](https://github.com/dyladan))
  * [#327](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/327) feat: enable root span to contain route ([@DinaYakovlev](https://github.com/DinaYakovlev))
  * [#389](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/389) feat: net module instrumentation ([@seemk](https://github.com/seemk))
  * [#379](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/379) chore: creating meta package for default auto instrumentations for node ([@obecny](https://github.com/obecny))
  * [#380](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/380) Moving Hapi Plugin to Instrumentation ([@obecny](https://github.com/obecny))
* `opentelemetry-host-metrics`
  * [#395](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/395) chore: fixing broken links, updating to correct base url, replacing gitter with github discussions ([@obecny](https://github.com/obecny))

#### :house: Internal
* [#392](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/392) chore: fixing broken links ([@obecny](https://github.com/obecny))

#### :memo: Documentation
* `opentelemetry-host-metrics`
  * [#373](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/373) Update README.md ([@z1c0](https://github.com/z1c0))

#### Committers: 7
- Bartlomiej Obecny ([@obecny](https://github.com/obecny))
- Daniel Dyla ([@dyladan](https://github.com/dyladan))
- Dina.Yakovlev ([@DinaYakovlev](https://github.com/DinaYakovlev))
- Siim Kallas ([@seemk](https://github.com/seemk))
- Valentin Marchaud ([@vmarchaud](https://github.com/vmarchaud))
[...]

PR #3669, opened 2026-08-12
fix(instrumentation-long-task): deprecate package

SyedTashfin:

- Deprecates `@opentelemetry/instrumentation-long-task` because the Long Tasks API is deprecated.
- Relates to open-telemetry/opentelemetry-browser#395, which was created after the Browser SIG decided not to move this instrumentation to the Browser repository.

- Adds a prominent README deprecation notice and points users toward instrumentation based on the Long Animation Frames API when available.
- Emits a diagnostic warning when `LongTaskInstrumentation` is instantiated.
- Adds a focused browser regression test that verifies the warning is emitted exactly once.
- Retains the implementation for a deprecation window rather than removing the package immediately.

## Validation

- `npm run compile --workspace @opentelemetry/instrumentation-long-task`
- `npm run test:browser --workspace @opentelemetry/instrumentation-long-task` (7 passing)
- `npm run lint`
- `npm run compile`
- `npm pack --workspace @opentelemetry/instrumentation-long-task --dry-run --json`
- `git diff --check`

The full repository test command also ran. Its only failure was the unrelated, unchanged `@opentelemetry/resource-detector-aws` integration test `AwsSuppressTracing.test.ts` timing out after 10 seconds; the focused long-task suite remained green.

## Maintainer follow-up

Per the component lifecycle policy, this deprecation needs confirmation from the long-task component owners (`@mhennoch`, `@t2t2`). Once merged/released, a maintainer will also need to mark the published npm package as deprecated; npm registry deprecation is not represented by a source-controlled `package.json` field.

## AI assistance disclosure

I used Hermes Agent (GPT-5.6) to assist with repository research, code and test drafting, and validation. I reviewed the final change and take responsibility for it.

### DIFF
--- a/packages/instrumentation-long-task/README.md
+++ b/packages/instrumentation-long-task/README.md
@@ -1,5 +1,7 @@
 # OpenTelemetry Long Task Instrumentation for web
 
+> **Deprecated:** This package is deprecated because the Long Tasks API is deprecated. Use instrumentation based on the [Long Animation Frames API][mdn-long-animation-frames] when available.
+
 [![NPM Published Version][npm-img]][npm-url]
 [![Apache License][license-image]][license-image]
 
@@ -81,3 +83,4 @@ Apache 2.0 - See [LICENSE][license-url] for more information.
 [npm-img]: https://badge.fury.io/js/%40opentelemetry%2Finstrumentation-long-task.svg
 [mdn-long-task]: https://developer.mozilla.org/en-US/docs/Web/API/Long_Tasks_API
 [mdn-performance-long-task-timing]: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceLongTaskTiming
+[mdn-long-animation-frames]: https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Long_animation_frame_timing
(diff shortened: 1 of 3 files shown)


overbalance (commented):
Thanks for the contribution! Before we can merge, we will need to create the instrumentation recommended here so that the user isn't confused about what to do next

SyedTashfin:
Understood, thank you. I’ll leave this PR unchanged for now. I couldn’t find a dedicated issue for the PerformanceLongAnimationFrameTiming replacement—should I open one, or is one already planned? I’d be interested in helping implement it once the scope is confirmed.

SyedTashfin:
Update on the condition you raised here: the replacement now exists as open-telemetry/opentelemetry-browser#429 — Long Animation Frames instrumentation emitted as log records, targeting #401. I first opened it in contrib (#3751) and david-luna asked me to move it to the browser repository, since that is where the browser instrumentations live now.

Once #429 is merged, the README notice here can point to a released path instead of "when available". I can reword it at that point if you prefer a different phrasing.

If you were already planning the LoAF port yourself, tell me and I will drop mine rather than duplicate the work.

The rest of this PR is unchanged, and the branch is rebased on current main.

SyedTashfin:
Putting the state on the record so nothing here reads as work still owed by me.

This PR is finished from my side. It deprecates the long-task instrumentation package and points readers at the replacement, and the replacement now exists as open-telemetry/opentelemetry-browser#429, the LoAF instrumentation, which is in review in the browser repository. So the condition raised here is satisfied by work already in the pipeline, not by another change on this branch.

I am not planning any further edits here. The README notice is accurate as written, and if you want different wording, or you would rather not reference an instrumentation path that has not shipped yet, adjust it to whatever you prefer at merge time. I am not holding this PR open for an edit from me.

All checks on the current head pass.

Issue #3668, opened 2026-08-12
Tracking: Migrate OpenInference JS Instrumentations into opentelemetry-js-contrib


psx95:
## Overview

This issue tracks the community effort to migrate the donated JavaScript/TypeScript ([original donation issue](https://github.com/open-telemetry/community/issues/3467)) auto-instrumentation packages from the [`donation-openinference`](https://github.com/open-telemetry/donation-openinference) staging repository into `opentelemetry-js-contrib` as part of the broader OpenInference Donation effort.

Unlike Python (which hosts GenAI packages in a separate `opentelemetry-python-genai` repository), the strategy for JavaScript is to integrate all migrated instrumentations directly into the `opentelemetry-js-contrib` monorepo.

>[!IMPORTANT]
>  - Use the instrumentations from https://github.com/open-telemetry/donation-openinference/ as the source
(once seeded). DO NOT use https://github.com/Arize-ai/openinference/.
> - Before starting work on a specific library, please ask to be assigned in the table —
for example, by leaving a comment on this issue.
> - Prioritize packages with more downloads and the ones your organization is interested in.
> - Make sure to link this issue from the migration PRs.

---

## Target Packages to Migrate

The following 14 packages from the OpenInference donation are queued for migration, ordered by monthly downloads (Stats as of August 4th, 2026):

| Target Package Name | Source Package | Monthly Downloads | Description / Notes | Assignee | Status / PR Link | Notes |
| :---- | :---- | :---- | :---- | :---- | :---- | :---- |
| `genai` | `@arizeai/openinference-genai` | 429,731 | Support for GenAI conventions |  |  |  |
| `vercel` | `@arizeai/openinference-vercel` | 424,970 | Support for Vercel AI SDK |  |  |  |
| `instrumentation-langchain` | `@arizeai/openinference-instrumentation-langchain` | 269,100 | Instrumentation for LangChain.js (Reconcile with existing `instrumentation-langchain` in repo) | @JacksonWeber  |  | A variant already exists. Reconcile if necessary. |
| `instrumentation-anthropic` | `@arizeai/openinference-instrumentation-anthropic` | 117,192 | Instrumentation for Anthropic SDK |  @eternalcuriouslearner | https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3664 | |
[...]

PR #3671, opened 2026-08-13
docs: add OpenInference migration skill

JacksonWeber:

- The JavaScript OpenInference migration effort in #3668 recommends adapting the Python migration skill, but JS Contrib does not have a repository-specific workflow for these ports.

- This is adapted from the Python https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/.github/skills/migrate-from-openinference/SKILL.md skill. It retains its greenfield/augment modes, native-instrumentation gate, conformance checks, and migration reporting. Updates replace Python-specific guidance with JS COntrib's Typescript, npm workspace, Nx, instrumentation, testing, ownership, and release conventions.

- Adds `.github/skills/migrate-from-openinference/SKILL.md`, adapted for JS Contrib's TypeScript, npm workspaces, Nx, release-please, instrumentation APIs, test tooling, component ownership, and generated files.
- Covers greenfield and augment migrations, native-instrumentation and shared-utility gates, semantic conventions, privacy-safe content capture, conformance, test-corpus migration, workspace integration, and migration reporting.
- Documents the skill in a new top-level `Skills` section in `CONTRIBUTING.md`.

## Validation

- `markdownlint-cli2 CONTRIBUTING.md .github/skills/migrate-from-openinference/SKILL.md`
- `prettier --check CONTRIBUTING.md .github/skills/migrate-from-openinference/SKILL.md`
- Agent Skills metadata and `git diff --check`

Related to #3668.


hectorhdzg:
Can you include some reference of the Python skill and more context here?

JacksonWeber:
> Can you include some reference of the Python skill and more context here?

Provided further context and a link to the original this was based off of in the description.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
+---
+name: migrate-from-openinference
+description: Migrate a donated JavaScript or TypeScript OpenInference package from open-telemetry/donation-openinference into opentelemetry-js-contrib. Creates a new package or augments an existing package with missing OpenInference coverage while converting it to official OpenTelemetry GenAI semantic conventions. Use when a user asks to migrate or port a donated OpenInference JS package.
```
psx95: Perhaps using original URL of the source donation repo might target the agent better: 
`... or TypeScript OpenInference package from https://github.com/open-telemetry/donation-openinference`

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+---
+
+# Migrate a donated OpenInference JavaScript package
+
+Migrate a package from
```
psx95: Starting with 
```
Migrate an `openinference-instrumentation-<source>` package from ...
```
Could prevent agent from accidentally targeting utilities packages in the donation repo such as `openinference-core` or `openinference-genai`

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+official OpenTelemetry GenAI semantic conventions, and contain no OpenInference
+dependencies.
+
+This skill covers two modes:
+
+- **Greenfield migration**: no equivalent package exists in this repository.
+  Create a new package under `packages/`.
+- **Augment an existing package**: this repository already instruments the
+  library. Inventory both implementations and add only the missing operations,
+  request or response shapes, tests, and semantic convention coverage.
```
psx95: Adding a link to the Augment Mode section specified later in the doc might help guide the agent better (Done similarly in the original Python skill).

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+
+The JavaScript migration effort is tracked in
```
psx95: For the agent skill, is mentioning this required? This might send the agent looking through the issue and all the comments on it that could act as noise when migrating a single instrumentation package.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+
+Derive the target name by removing `openinference-` and changing the npm scope:
+
+```text
+@arizeai/openinference-instrumentation-anthropic
+  -> @opentelemetry/instrumentation-anthropic
+  -> packages/instrumentation-anthropic
+```
+
+Not every donated package is a Node.js auto-instrumentation:
```
psx95: nit: This *might* not be required if we do something [like this](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3671/changes#r3785347429).

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+  support decision before implementation.
+
+## Before starting
+
+### 1. Claim the package
+
+Check the assignment table in
+[`opentelemetry-js-contrib#3668`](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3668).
+Confirm that the tracking issue covers the target as an accepted instrumentation
+request. Before writing code, ask to be assigned to an unclaimed package and
```
psx95: I'm wondering how would an agent address this task of finding potential maintainers?

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+identify at least two contributors who agree to serve as code owners, as
+required by `CONTRIBUTING.md`. Link the tracking issue from the migration pull
+request.
+
+Also confirm that the component belongs in this repository. `CONTRIBUTING.md`
```
psx95: It seems a little unclear as to what `Also confirm that the component belongs in this repository.` means here. 
Looking at the text following this, it seems like this is instructing the agent to check for native instrumentations, but then there's a dedicated section for that later.

Let me know if I'm missing something here.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+prefers native instrumentation, a repository close to the instrumented
+library, or a dedicated public repository before JS Contrib. For vendor
+components, document installation, configuration, exposed APIs, and relevant
+design; meet the repository's current test coverage; and establish an ongoing
+maintainer who can review changes and respond to failures and issues.
+
+### 2. Check for an existing target
+
```
psx95, line 83: nit: Some instructional text on the intended effect or purpose of the shell command might be better.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+```sh
+test -d packages/<target> && echo existing
```
psx95: nit: should `<target>` be changed to `<library>` to keep consistent with the shell command below.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+If an equivalent package exists, use
+[augment mode](#augment-mode-the-package-already-exists). Do not create a
+competing package.
+
+Known examples include:
+
+- `instrumentation-openai`
+- `instrumentation-langchain`
+- AWS Bedrock coverage in `instrumentation-aws-sdk`
+
```
psx95, line 108: Adding:
```
If the package does not exist, continue below. 
```
might be a more explicit instruction.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+   prove that the SDK emits conformant telemetry.
+
+If the SDK already emits equivalent OpenTelemetry telemetry, do not build
+competing monkey-patching instrumentation. Instead:
+
+- add conformance and regression tests for the native integration when that
+  belongs in this repository;
+- document missing operations or semantic convention gaps;
+- propose upstream fixes or an adapter only when maintainers agree it is
+  necessary.
```
psx95: Do you think it might be better to simply copy over the 4 points from [the python skill](http://github.com/open-telemetry/opentelemetry-python-genai/blob/main/.github/skills/migrate-from-openinference/SKILL.md#before-you-start-check-for-native-otel-instrumentation) ? 

It relies on a "write-conformance-test" skill, but we can add that as a fast follow in another PR ?
For now we could update that particular step to state "if write-conformance-tests skill is not present in this repo, skip writing the tests"

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+
```
psx95, line 145: IMO, its important to mention at this point: 

"Only when the library has no native OTel instrumentation do you continue with the migration flow below."

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+### 4. Check for shared GenAI utilities
```
psx95, line 146: FYI: I am working on this package.

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+Only create files the package needs. Middleware-based integrations may not
+need `instrumentation.ts` or `.tav.yml`.
+
+Do not copy the donated README, changelog entries, examples, workspace
+configuration, generated build output, lockfile, or agent configuration. For a
+new package, create the repository-standard `CHANGELOG.md` heading when recent
+package additions do so; release-please owns subsequent changelog entries.
+Copy reusable test fixtures only after sanitizing and reviewing them.
+
+For a new package:
```
psx95, line 349: I thought that in the current form of this skill, an agent will only reach this section if the library is in augment mode, if that's the case then these instructions could act as noise? 
Let me know if I'm missing something here.

In the [python skill's migration flow section](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/.github/skills/migrate-from-openinference/SKILL.md#migration-flow), there's subtext that makes it clear that the Migration flow applies to both paths and which steps should be skipped for a particular path (if any) - perhaps adding that subtext could make it more clear?

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,667 @@
...
+
+- label them clearly as generated test data;
+- base them on the provider's public API schema;
+- mention them in the pull request;
+- create a follow-up issue to replace them with real sanitized recordings.
+
+Compare source and target test inventories before declaring the migration
+complete. A large unexplained reduction indicates missing coverage.
+
+### 9. Add semantic convention conformance tests
```
psx95, line 559: Once we have the conformance tests skill we could re-word this. I think conformance test skill could be good to have as it could apply to all the instrumentations.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Sounds good, I updated the skill description to use the full source repository URL, so the agent has a clearer reference when matching migration requests.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: I narrowed the skill to openinference-instrumentation-* packages and added an explicit guard against selecting utility or other non-instrumentation packages. That should keep packages like open inference-core and openinference-genai from being targeted accidentally.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: I added a direct link from the mode summary to the Augment mode section so the agent can jump to the correct workflow when the target package already exists.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: I removed the tracking issue from the introduction, so it doesn’t distract the agent during migration. I kept the links in the assignment and PR steps, where they’re needed for the contributor workflow.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Updated as a part of an above comment.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Yeah, Python flow is clearer here. I adapted its four steps for JavaScript and added the fallback you suggested. If the write-conformance-tests skill isn’t present, the agent records that limitation in the migration report and skips authoring those tests for now.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Agreed, updated as a part of addressing a previous comment.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Added a note clarifying that it applies to both paths, including exactly which steps augment mode replaces, skips, or limits to the computed delta.

.github/skills/migrate-from-openinference/SKILL.md
hectorhdzg: This claiming of the package is not part of Python skill and we should remove it; agent would be able to address this, we don't want agents to start assigning people without their knowledge

JacksonWeber: I removed the package-claiming and contributor-assignment instructions, including the related checklist items. The skill now only verifies that the instrumentation request is accepted and appropriate for this repository

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: I meant the broader question of whether the instrumentation meets JS Contrib’s acceptance and maintenance requirements, not the native instrumentation check. I’ll reword this paragraph to make that distinction explicit.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Added explanatory text describing the command’s purpose and output.

hectorhdzg (approved):
LGTM, thanks for adding this

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Added

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,690 @@
...
+3. Confirm behavior with a real `TracerProvider`; a dependency alone does not
+   prove that the SDK emits conformant telemetry.
+
+If the library is self-instrumented, do not migrate the OpenInference package.
+Pivot the work:
+
+1. **Ignore the OpenInference instrumentation entirely.** The vendor owns the
+   spans, so there is no instrumentor or patcher to reimplement.
+2. **Write conformance tests against the native instrumentation.** If
+   `.github/skills/write-conformance-tests/SKILL.md` exists, follow it, but
```
aabmass: Why would it conditionally exist? It's not in the repo

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,690 @@
+---
+name: migrate-from-openinference
+description: Migrate a donated JavaScript or TypeScript openinference-instrumentation-* package from https://github.com/open-telemetry/donation-openinference into opentelemetry-js-contrib. Creates a new package or augments an existing package with missing OpenInference coverage while converting it to official OpenTelemetry GenAI semantic conventions. Use when a user asks to migrate or port a donated OpenInference JS instrumentation package.
```
aabmass: Are there actually JavaScript instrumentations to migrate?

.github/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,690 @@
...
+---
+
+# Migrate a donated OpenInference JavaScript instrumentation package
```
aabmass: I haven't looked closely at the OpenInference instrumentations, but do they use the same monkey patching techniques as here in contrib?

aabmass (approved):
Thanks

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Yeah, I’ll remove that reference and describe the expected test approach directly instead.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Updated to make it clear the instrumentations are TS.

.github/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Most of them do. OpenAI, and LangChain extend `InstrumentationBase`, use `InstrumentationNodeModuleDefinition`, and wrap SDK methods much like contrib instrumentations. Others need different hooks, so the skill should verify the package's approach first before making assumptions.

.agents/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,688 @@
...
+The tracking issue calls for shared JavaScript GenAI utilities analogous to
+Python's `opentelemetry-util-genai`. Check the current branch rather than
+assuming they exist:
+
+```sh
+find packages -maxdepth 2 -type d -iname '*genai*'
+rg -l 'gen_ai\.input\.messages|gen_ai\.output\.messages' packages
+```
+
+If a shared package exists, use only its public API. Do not import private
```
psx95, line 157: In the last SIG meeting, we decided that initially, the utils library will not be exporting any utility functions. And the instrumentation authors - as they add instrumentation libraries, will export the functions from the library as required for their instrumentation.
This is done to prevent exporting utility functions that might not be used by instrumentations and would help in stabilizing the library later on.

The instructions left via comments in [index.ts of the library](https://github.com/psx95/opentelemetry-js-contrib/blob/e1c124f994390cd94e5e3921fdce9af591907379/packages/genai-util/src/index.ts) make this intent clear.

As such, we should add the following: 

`If a shared package exists, use only its public API. Do not import private modules or duplicate its models inside an instrumentation. Some of the public functions required may not be exported by the shared util library and might need to be exported. Read any comments in index.ts file of the library to check if exporting functions is allowed.`

.agents/skills/migrate-from-openinference/SKILL.md
```diff
@@ -0,0 +1,688 @@
...
+  packages/<target>/src packages/<target>/test packages/<target>/package.json
+```
+
+The output must be empty.
+
+## Augment mode: the package already exists
+
+OpenInference is a second coverage reference, not permission to rewrite an
+existing package.
+
```
psx95, line 252: Nit: Since the agent can jump to augment mode directly, I think mentioning here that "All [Non-negotiable rules](#non-negotiable-rules) still apply in this mode."  would make behavior expectations explicit for the agent.

Note: the [python skill](https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/.github/skills/migrate-from-openinference/SKILL.md#augment-mode-the-package-already-exists) also does this.

psx95 (commented):
Thanks for addressing the feedback! Just left a couple of comments!

.agents/skills/migrate-from-openinference/SKILL.md
JacksonWeber: Updated to include this language.

.agents/skills/migrate-from-openinference/SKILL.md
JacksonWeber: That makes sense, and I've made that explicit now.

psx95 (approved):
Thanks for this contribution! Excited to try this skill 👍🏻

.agents/skills/migrate-from-openinference/SKILL.md
trentm, line 1: I don't want to block, because I know you are mimicking the python repo, but why `.github/skills` as the dir rather than `.agents/skills`?

IIUC, Codex and Copilot will look in `.agents/skills`.
Claude Code (as is typical of Anthropic I guess?) will only look in `.claude/skills`. It looks like the Python repo has a symlink from `.claude/skills` to ../.github/skills.

trentm (approved):
Approving to not block people. I don't see how this can cause harm.
I haven't *carefully* reviewed the (long) content.

JamieDanielson (approved):
Same note as Trent re: location of the skill. Otherwise looking forward to trying this out and we can iterate as needed. Thanks for doing this

.agents/skills/migrate-from-openinference/SKILL.md
JacksonWeber: To be honest, I just replicated the .github folder location from the python skill. Looks like _only_ copilot recognizes that path, so I'm happy to move this to the more generic .agents folder + a sym link for Anthropic compatibility.

PR #3672, opened 2026-08-14
fix(winston-transport): serialize Error attributes

### LINKED ISSUE #2443: [winston-transport] - Transform instances of `Error` so we don't "lose" them
From <https://github.com/open-telemetry/opentelemetry-js/issues/4828>.

We often log errors with winston, like:

```ts
try {
  // Code that throws
} catch (e) {
  logger.error('Unhandled error', {
    error: e
  })
}
```

`Error` instances are not valid attributes for log records, and they're not transformed by the winston transport, so we get:

1. A console warning, "Invalid attribute value set for key: error".
2. And most importantly, most of the time we don't get the error details in the resulting telemetry.

I'd love it if the winston transport could apply some heuristics to make sure the attributes for the log records are valid values.

As a note, `JSON.stringify`ing objects may not be sufficient since many properties from `Error` are not enumerable.

```node
> JSON.stringify(new Error('Boom'))
'{}'
```

I've noticed that the Azure Monitor exporter [has a trick](https://github.com/Azure/azure-sdk-for-js/blob/eb31ab6e782c6d7c40f32ee705630b893d0867f9/sdk/monitor/monitor-opentelemetry-exporter/src/utils/common.ts#L259-L275) to get more info from an `Error` instance:

```node
> const error = new Error('Boom')
undefined
> JSON.stringify(error, Object.getOwnPropertyNames(error))
'{"stack":"Error: Boom\\n    at REPL11:1:15\\n    at ContextifyScript.runInThisContext (node:vm:136:12)\\n    at REPLServer.defaultEval (node:repl:598:22)\\n    at bound (node:domain:432:15)\\n    at REPLServer.runBound [as eval] (node:domain:443:12)\\n    at REPLServer.onLine (node:repl:927:10)\\n    at REPLServer.emit (node:events:531:35)\\n    at REPLServer.emit (node:domain:488:12)\\n    at [_onLine] [as _onLine] (node:internal/readline/interface:417:12)\\n    at [_line] [as _line] (node:internal/readline/interface:888:18)","message":"Boom"}'
```

This is better, however I'm not sure whether it takes all cases into account.

A quick look on npm seems to suggest that the [serialize-error package](https://github.com/sindresorhus/serialize-error#readme) is the most comprehensive way to get all information one needs about an `Error`. The main issue seems to be that it's an ESM-only package, and even if it wasn't, I'm not sure whether you'd be open to take a dependency on it. Maybe we could vendor the subset of the code in this package?

At this stage, none, it feels like this transport is the "point of entry" into the OTel world, so it should ideally respect the OTel contract when it comes to attributes.

JacksonWeber:

Fixes #2443.

Winston metadata can contain `Error` instances, but those are not valid OpenTelemetry log attribute values. Conventional `error` and `err` fields are already promoted to the Logs API exception field; errors in other or nested metadata fields were still rejected and dropped by the SDK.

- Recursively convert remaining `Error` instances to structured AnyValue maps.
- Preserve `name`, `message`, `stack`, `code`, `cause`, aggregate `errors`, and custom own properties, including non-enumerable properties.
- Represent circular references in serialized error metadata as `[Circular]` without mutating the Winston record.
- Add transport-level edge-case coverage, a real Winston instrumentation integration test, and README documentation.

## Testing

- Compiled `@opentelemetry/winston-transport` and `@opentelemetry/instrumentation-winston`.
- Ran the full Winston transport suite: 26 passing.
- Ran the full Winston instrumentation suite: 24 passing.
- Packed the modified transport and exercised it with Winston 3.19.0 through an OTLP/HTTP pipeline backed by `otel/opentelemetry-collector-contrib:0.158.0` (`sha256:c5918f78992ee73b0d6f0e599423ac5ec52dd5d9726733114d6eca53d5a32ed5`). Confirmed primary exception fields, structured top-level and nested errors, aggregate causes, custom non-enumerable fields, circular markers, and no invalid-attribute diagnostic.

### DIFF
--- a/packages/winston-transport/README.md
+++ b/packages/winston-transport/README.md
@@ -53,6 +53,15 @@ const logger = winston.createLogger({
 > [!IMPORTANT]
 > Logs will be duplicated if `@opentelemetry/winston-transport` is added as a transport in `winston` and `@opentelemetry/instrumentation-winston` is configured with `disableLogSending: false`.
 
+### Error handling
+
+`Error` instances assigned to the conventional `error` or `err` fields are
+emitted through the OpenTelemetry Logs API exception field. Instances assigned
+to other metadata fields, including errors nested in arrays or objects, are
+converted to structured attributes containing their `name`, `message`, `stack`,
+and other own properties. Circular references in serialized error metadata are
+represented as `[Circular]`.
+
 ### Supported versions
 
 - [`winston`](https://www.npmjs.com/package/winston) versions `>=3.0.0 <4`
--- a/packages/winston-transport/src/utils.ts
+++ b/packages/winston-transport/src/utils.ts
@@ -5,6 +5,7 @@
 
 import {
   AnyValue,
+  AnyValueMap,
   LogAttributes,
   LogRecord,
   Logger,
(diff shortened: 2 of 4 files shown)


packages/winston-transport/README.md
```diff
@@ -53,6 +53,15 @@ const logger = winston.createLogger({
 > [!IMPORTANT]
 > Logs will be duplicated if `@opentelemetry/winston-transport` is added as a transport in `winston` and `@opentelemetry/instrumentation-winston` is configured with `disableLogSending: false`.
 
+### Error handling
+
+`Error` instances assigned to the conventional `error` or `err` fields are
```
hectorhdzg, line 58: Is this kind of error handling shared across instrumentation here or only winston?

packages/winston-transport/src/utils.ts
```diff
@@ -50,6 +51,101 @@ function getSeverityNumber(level: string): SeverityNumber | undefined {
...
+  'cause',
+  'errors',
+];
+
+function isPlainObject(value: object): value is Record<string, unknown> {
+  const constructor = (value as { constructor?: unknown }).constructor;
+  return constructor === Object || constructor === undefined;
+}
+
+function serializeError(error: Error, ancestors: Set<object>): AnyValueMap {
```
hectorhdzg, line 69: Is there any kind error serialization we can reuse here? are we serializing errors somewhere else in main or contrib repos?

packages/winston-transport/README.md
JacksonWeber: Recursive serialization of the `Error` values is Winston specific. We could add something like this to bunyan in a future PR, but don't think it makes sense for Pino.

packages/winston-transport/src/utils.ts
JacksonWeber: I checked both repos and couldn’t find anything reusable here. The existing helpers either stringify errors for diagnostics or only handle the standard exception fields, not nested error metadata.

hectorhdzg (approved):

PR #3675, opened 2026-08-15
docs: fix issue template contributing links

rajanpanth:
## Summary
- fix issue-template links that pointed to `.github/blob/main/CONTRIBUTING.md`
- use relative paths back to the repository root `CONTRIBUTING.md`

## Testing
- `git diff --check`

### DIFF
--- a/.github/ISSUE_TEMPLATE/instrumentation_request.md
+++ b/.github/ISSUE_TEMPLATE/instrumentation_request.md
@@ -5,7 +5,7 @@ labels: instrumentation-request
 ---
 
 <!--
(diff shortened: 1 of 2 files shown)


pichlermarc (approved):
This is actually supposed to point to the correct place from the issue/[issue-creation screen](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/new?template=instrumentation_request.md). This PR would break that.

**suggestion:** we can change the link to be a absolute link if the goal is to make it work everywhere. :)

pichlermarc (changes requested):
meant to request changes - see previous comment (https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3675#pullrequestreview-4952099022)

pichlermarc (approved):

PR #3677, opened 2026-08-17
feat: add @opentelemetry/genai-util package

### LINKED ISSUE #3681: feat: Add @opentelemetry/genai-util package for Generative AI instrumentations

As Generative AI libraries and frameworks (OpenAI, Anthropic, Google GenAI, AWS Bedrock, LangChain.js, LlamaIndex.TS, etc.) grow in the Node.js/TypeScript ecosystem, building and maintaining separate instrumentation packages requires solving the same complex telemetry challenges repeatedly:

1. **Streaming & TTFT tracking**: Measuring Time-To-First-Token (TTFT) and managing span lifecycles across `AsyncIterable` streams requires intricate proxying and error handling.
2. **Code duplication & spec drift**: Manually implementing [OpenTelemetry GenAI Semantic Conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/README.md) across different instrumentations leads to duplicated boilerplate and inconsistent attributes, events, and metric histogram boundaries.
3. **Lifecycle variety**: GenAI workflows now span chat inference, local/remote agents, tool executions, vector RAG retrieval, and multi-step workflows.
4. **Privacy controls**: Managing message content capturing (e.g. `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`) should be enforced uniformly.

---

Introduce a shared utility package, `@opentelemetry/genai-util` (under `packages/genai-util`), similar to [`opentelemetry-util-genai`](https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/util/opentelemetry-util-genai) in the Python repository (`open-telemetry/opentelemetry-python-genai`).

The package will provide:
- **Semantic Conventions**: Type-safe constants and enums conforming to OpenTelemetry GenAI Semantic Conventions (`gen_ai.*` attributes, operations, providers, finish reasons, token types, and metrics).
- **`TelemetryHandler` & Typed Invocations**: Standardized lifecycle handlers for AI operations:
  - `InferenceInvocation` (chat, completion, multimodal)
  - `AgentInvocation` (in-process internal agents and remote client agents)
  - `ToolInvocation` (function / tool executions)
  - `RetrievalInvocation` (RAG / vector database lookups)
  - `EmbeddingInvocation` & `WorkflowInvocation`
- **Stream Wrapper (`wrapAsyncStream`)**: Transparent `AsyncIterable` proxy that automatically records TTFT (`gen_ai.client.operation.time_to_first_chunk`), manages chunk processing, and safely ends spans on completion, error, or early loop exit (`return()`).
- **Standardized Metrics Helpers**: Pre-configured histograms with spec-defined explicit bucket boundaries (`duration`, `token.usage`, `time_to_first_chunk`).
- **Content Capture & Hooks**: Standard support for `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` and pluggable `CompletionHookManager` for evaluations or external logging.

---

1. **Re-implementing logic in each instrumentation**: High maintenance burden, duplicate logic, and risks of inconsistent telemetry across packages.
2. **External non-contrib package**: Lacks official alignment with OpenTelemetry JS release cycles and semantic convention updates.

---

### Additional context

- **Python Reference**: [`opentelemetry-python-genai / opentelemetry-util-genai`](https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/util/opentelemetry-util-genai) provides the exact equivalent in the Python ecosystem. Bringing this to `opentelemetry-js-contrib` ensures language parity and makes writing new JS/TS GenAI instrumentations straightforward.
- **Specification**: [OpenTelemetry GenAI Semantic Conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/README.md).

psx95:
> [!NOTE]
> Planning to split this PR into smaller PRs to facilitate faster reviews. Reviewers are requested to review the completed PRs from the list below for easier review.

### Planned split for this PR (tentative)

- [x] **PR 1**: Package scaffolding, GenAI semconv, types, env vars, and core utils (#3709)
- [x] **PR 2**: Metrics helpers & token usage histograms (#3712)
- [x] **PR 3**: Base invocation lifecycle & `TelemetryHandler` foundation (#3757)
- [ ] **PR 4**: Core invocations (`InferenceInvocation`, `EmbeddingInvocation`, `ToolInvocation`)
- [ ] **PR 5**: Agent & pipeline invocations (`AgentInvocation`, `WorkflowInvocation`, `RetrievalInvocation`, `FetchResponseInvocation`)
- [ ] **PR 6**: Streaming telemetry proxy & TTFT calculation (`wrapAsyncStream`)

Fixes #3681

Introduces `@opentelemetry/genai-util` (the TypeScript/JavaScript equivalent of Python's [`opentelemetry-util-genai`](https://github.com/open-telemetry/opentelemetry-python-genai/tree/main/util/opentelemetry-util-genai)) to eliminate boilerplate across GenAI instrumentations (OpenAI, Anthropic, Bedrock, LangChain, Google GenAI, etc.) for semantic conventions, span lifecycles, streaming responses, metrics, content capture, and completion hooks.

**See the draft PR #3698 that highlights how this library can potentially be used.**

---

Introduces the `@opentelemetry/genai-util` package with:

- **Semantic Conventions (`semconv.ts`)**: Stable constants for attributes, operation names, providers, events, finish reasons, token types, and metrics aligned with the latest OpenTelemetry GenAI Semantic Conventions.
- **`TelemetryHandler` & Invocations**: Central façade managing structured invocation lifecycles, span attributes, and metric recording for all GenAI operations (`inference`, `agent`, `embedding`, `tool`, `workflow`, `retrieval`, and `fetch_response`).
- **Stream Wrapper (`wrapAsyncStream`)**: ES6 proxy wrapper for async iterables that tracks streaming lifecycle, captures TTFT (`gen_ai.client.operation.time_to_first_chunk`), handles stream events/errors, and auto-finalizes spans.
- **Metrics Helpers (`metrics.ts`)**: Standard histograms with recommended bucket boundaries (`duration`, `token.usage`, `time_to_first_chunk`, and `server.time_to_first_token`).
- **Privacy & Content Capture (`environment-variables.ts`)**: Configurable content capture modes (`none`, `span_only`, `event_only`, `span_and_event`), event emission, and `CompletionHookManager`.
- **Incremental Export Policy**: APIs marked `@experimental` and kept internal to allow controlled, incremental exposure as instrumentation packages adopt them.
- **Tests & Documentation**: Comprehensive test suite (>95% code coverage), package `README.md`, and configuration for `release-please` and `component_owners.yml`.

---

## Type of change

- [x] New feature (non-breaking change which adds functionality)

---

## Checklist:

- [x] Followed the style guidelines of this project
- [x] Added unit tests for new functionality
- [x] All unit tests pass locally (`npm test`)
- [x] Updated package `README.md` and documentation


packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,183 @@
...
+  };
+
+  if (typeof port === 'number' && !isNaN(port)) {
+    attrs[ATTR_SERVER_PORT] = port;
+  }
+
+  return attrs;
+}
+
+/**
```
DylanRussell, line 59: I think we should not do this.. Instrumentations should coerce types to something that can go in attributes..

DylanRussell: Here and below, input / output messages, content, sys instructions should be strongly types

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,133 @@
...
+    `invalid boolean value for environment variable: ${name}=${val}; ignoring`
+  );
+  return undefined;
+}
+
+/**
+ * Parse an arbitrary input value into a valid ContentCaptureMode.
+ */
+export function parseContentCaptureMode(
+  value?: string | boolean | ContentCaptureMode
```
DylanRussell, line 26: we shouldnt accept a bool here. that is legacy code, new code should require this to be an enum

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
DylanRussell, line 5: following up on our discussion a bit.. I think the only things that should be "public" (available to instrumentations in this repo or outside this repo that have installed this package) is the stuff in handler.ts, types.ts, completion-hook.ts, the classes in stream.ts, not sure about environment-vairables.ts but I think probably shouldn't be public ?.. Not totally sure about my analysis here, but i think it's a good thing to think through carefully..

packages/genai-util/src/environment-variables.ts
psx95: I noticed references of boolean values being used for `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`  in a lot of places like this https://opentelemetry.io/blog/2024/otel-generative-ai/. So I added boolean as a means for backwards compatibility. 

What is the authoritative source that indicates that boolean values are legacy now?

packages/genai-util/src/utils.ts
psx95: Yeah, I agree - I was confused about this as well. IMO, there are pros & cons:
 - Pros:
    - It doesn't really hurt or cause confusion - these functions are stateless, and can act as true utilities for other instrumentations. For instance `formatInputMessages` can be used to convert strings to JSON string.
   
- Cons:
  - If these are public, any changes to their signature or behavior would be considered breaking changes increasing the maintenance burden.
  
 I'd prefer if these can be just used within the utils package. Perhaps we can control the public export by not exporting anything tom `utils.ts` in `index.ts`.

packages/genai-util/src/environment-variables.ts
DylanRussell: the purpose of the env var is to determine if message content (and other sensitive stuff like tool descriptions) should be included in spans  and/or events, initially it was just true/false and that wasn't granular enough, we wanted users to decide between the 4 possible combinations...

it doesn't appear to be officially documented anywhere in the sem conv repo, but that's the history of this env var.. if we are starting from scratch here I don't see why we would accept a `bool` instead of the enum

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Complete the invocation successfully.
+   */
+  public stop(endTime?: HrTime | number): void {
+    if (this._isEnded) {
+      return;
+    }
+    this._isEnded = true;
+
+    const endHr = Array.isArray(endTime) ? endTime : process.hrtime();
```
JacksonWeber, line 74: I think there’s a clock mismatch here. `process.hrtime()` returns monotonic time relative to an arbitrary origin, but `span.end()` interprets the tuple as an absolute OTel `HrTime`. Since the span started with an epoch-based timestamp, the calculated duration is negative, and the SDK clamps it to zero. The same value is also used for event timestamps.

Could we use OTel’s `hrTime()` for span and event timestamps, while keeping a separate monotonic reading for the duration metric? It would also be helpful to add a test asserting that the exported span has a non-zero duration and an end time after its start time.

packages/genai-util/src/invocations/tool.ts
```diff
@@ -0,0 +1,102 @@
...
+      attrs[ATTR_GEN_AI_TOOL_DESCRIPTION] = options.toolDescription;
+    }
+    if (options.toolCallId) {
+      attrs[ATTR_GEN_AI_TOOL_CALL_ID] = options.toolCallId;
+    }
+    if (options.toolType) {
+      attrs[ATTR_GEN_AI_TOOL_TYPE] = options.toolType;
+    }
+    if (options.toolArguments !== undefined) {
+      attrs[ATTR_GEN_AI_TOOL_CALL_ARGUMENTS] = serializeContent(
```
JacksonWeber: It looks like tool arguments and results are always added to the span, regardless of the configured content capture mode. These attributes could contain sensitive user data and are marked opt-in in the semantic conventions, so I’d expect them to be omitted when capture mode is `none` or `event_only`.

packages/genai-util/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "@opentelemetry/api": "^1.1.0"
+  },
+  "dependencies": {
+    "@opentelemetry/core": "^2.0.0",
+    "@opentelemetry/semantic-conventions": "^1.27.0"
+  },
+  "devDependencies": {
+    "@opentelemetry/api": "^1.1.0",
+    "@opentelemetry/sdk-metrics": "^1.25.0",
+    "@opentelemetry/sdk-trace-base": "^1.25.0"
```
pichlermarc: these dependencies are ancient - can we update them to at least `^2.0.0`?
We should also use `@opentelemetry/sdk-trace` over `@opentelemetry/sdk-trace-base`

packages/genai-util/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "license": "Apache-2.0",
+  "engines": {
+    "node": "^18.19.0 || >=20.6.0"
+  },
+  "bugs": {
+    "url": "https://github.com/open-telemetry/opentelemetry-js-contrib/issues"
+  },
+  "homepage": "https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/genai-util#readme",
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.1.0"
```
pichlermarc: Public API is using metrics types, which was introduced in `1.3.0`
```suggestion
    "@opentelemetry/api": "^1.3.0"
```

packages/genai-util/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+export * from './types';
+export * from './semconv';
+export * from './environment-variables';
+export * from './metrics';
+export * from './completion-hook';
+export * from './invocations';
+export * from './stream';
+export * from './handler';
```
pichlermarc: With `export *` we've ended up with excessive exports that were not necessary in the past.

**suggestion:** let's use explicit exports over wildcards so that public API surface is a conscious choice and obvious to reviewers.

packages/genai-util/src/invocations/tool.ts
psx95: Good catch, I missed this one. I will fix this and add the test.

packages/genai-util/src/invocations/base.ts
psx95: Makes sense, let me take a closer look at this.

psx95: I agree that we should avoid the mismatch and so I changed the usages to use `HrTime` from `@opentelemetry/api`. 

Regarding duration metric computation - I didn't fully understand the comment. What is meant by `separate monotonic reading for duration metric` ?

The computation of duration metric across **all** supported invocations relies on the utility method - `calculateDurationSeconds`.

packages/genai-util/src/invocations/inference.ts
```diff
@@ -0,0 +1,436 @@
...
+    }
+    if (this._systemInstructions) {
+      const formatted = formatSystemInstructions(this._systemInstructions);
+      if (formatted) {
+        eventAttrs[ATTR_GEN_AI_SYSTEM_INSTRUCTIONS] = formatted;
+      }
+    }
+
+    if (Object.keys(eventAttrs).length > 0) {
+      this._span.addEvent(
```
lmolkova: These are span events, but `gen_ai.client.inference.operation.details` is defined in semconv as a log-based events, there is no point in reporting these as span events, if Js does not have stable logs API, let's just not emit events for now

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,511 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * Semantic conventions in this file are defined locally rather than imported from
+ * `@opentelemetry/semantic-conventions/incubating`.
+ *
+ * Rationale:
+ * 1. Per OpenTelemetry JavaScript Contrib repository guidelines (GUIDELINES.md),
+ *    packages must not depend on `@opentelemetry/semantic-conventions/incubating`
```
lmolkova, line 12: The rationale for avoiding `/incubating` makes sense, but hand-writing the names is not a viable substitute - 112 literals maintained by hand will rot on the first convention change. We should have genai-specific codegen right into this library. I think it's ok as a follow up

psx95: Created https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3694 to track this. I'll tackle this as a follow-up to this PR.

packages/genai-util/src/invocations/inference.ts
psx95: Oh, I just noticed - https://opentelemetry.io/blog/2026/deprecating-span-events/. I wasn't aware of this. 

Should I just completely remove the support for 'events' for now from this library - meaning `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` will only support `span_only` mode? 

(*Till logs API is stabilized*)

psx95: Created commit https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677/commits/b56ca198d4fd7e45d8c6f3288f550409f411e87e to address this. 

I removed the support for `event_only` and `span_and_event` for now from the codebase. I left the placeholder no-op functions for now in the individual invocation types that are supported by the library.

packages/genai-util/src/semconv.ts
psx95: There was already an issue to track this work in https://github.com/open-telemetry/opentelemetry-js/issues/6783. This proposes addition of genai specific semantic conventions package similar to python.

.github/component_owners.yml
```diff
@@ -42,6 +42,8 @@ components:
     - legendecas
   packages/id-generator-aws-xray:
     - jj22ee
+  packages/genai-util:
+    - psx95
```
aabmass, line 46: Would anyone else like to volunteer to be a component owner?

packages/genai-util/src/invocations/agent.ts
```diff
@@ -0,0 +1,467 @@
...
+      options as RemoteAgentInvocationOptions
+    )?.serverAddress;
+    this._serverPort = (options as RemoteAgentInvocationOptions)?.serverPort;
+    this._agentName = options?.agentName;
+    this._agentId = options?.agentId;
+    this._agentDescription = options?.agentDescription;
+    this._agentVersion = options?.agentVersion;
+    this._conversationId = options?.conversationId;
+    this._dataSourceId = options?.dataSourceId;
+    this._outputType = options?.outputType;
```
aabmass: Are all of these truly optional or some are required?

packages/genai-util/src/invocations/agent.ts
```diff
@@ -0,0 +1,467 @@
...
+  private _agentVersion?: string;
+  private _conversationId?: string;
+  private _dataSourceId?: string;
+  private _outputType?: string;
+  private _usage?: TokenUsage;
+  private _finishReasons?: string[];
+  private _inputMessages?: InputMessages;
+  private _outputMessages?: OutputMessages;
+  private _systemInstructions?: SystemInstruction;
+  private _firstChunkTime?: HrTime;
```
aabmass: Im a little surprised these are all optional, can some be required?

packages/genai-util/src/invocations/base.ts
JacksonWeber: I meant using a monotonic clock for the duration metric, so clock adjustments don’t affect it. But since OTel’s `hrTime()` already uses `performance.now()`, your change covers that too. LGTM!

packages/genai-util/src/invocations/agent.ts
```diff
@@ -0,0 +1,467 @@
...
+
+  constructor(
```
aabmass, line 81: We can drop all these `if` checks. In OTel JS, `span.setAttributes()` ignores keys with `undefined` values anyway (checked in `SpanImpl.setAttribute`), and `Attributes` allows `undefined`.

Also if we default `options = {}` in the constructor args, we can drop all the `options?.` chaining:

```ts
constructor(
  span: Span,
  handler?: TelemetryHandler,
  options: AgentInvocationOptions | RemoteAgentInvocationOptions = {},
  startTime: TimeInput = hrTime()
) {
  super(span, handler, startTime);
  this._providerName = (options as RemoteAgentInvocationOptions).providerName;
  this._requestModel = options.requestModel;
  this._serverAddress = (options as RemoteAgentInvocationOptions).serverAddress;
  this._serverPort = (options as RemoteAgentInvocationOptions).serverPort;
  this._agentName = options.agentName;
  this._agentId = options.agentId;
  this._agentDescription = options.agentDescription;
  this._agentVersion = options.agentVersion;
  this._conversationId = options.conversationId;
  this._dataSourceId = options.dataSourceId;
  this._outputType = options.outputType;

  const reqOpts = options.requestOptions;
  const attrs: Attributes = {
    [ATTR_GEN_AI_OPERATION_NAME]: GEN_AI_OPERATION_NAME_VALUE_INVOKE_AGENT,
    [ATTR_GEN_AI_PROVIDER_NAME]: (options as RemoteAgentInvocationOptions).providerName,
    [ATTR_GEN_AI_REQUEST_MODEL]: options.requestModel,
    [ATTR_GEN_AI_AGENT_ID]: options.agentId,
    [ATTR_GEN_AI_AGENT_NAME]: options.agentName,
    [ATTR_GEN_AI_AGENT_DESCRIPTION]: options.agentDescription,
    [ATTR_GEN_AI_AGENT_VERSION]: options.agentVersion,
    [ATTR_GEN_AI_CONVERSATION_ID]: options.conversationId,
    [ATTR_GEN_AI_DATA_SOURCE_ID]: options.dataSourceId,
    [ATTR_GEN_AI_OUTPUT_TYPE]: options.outputType,
    [ATTR_SERVER_ADDRESS]: (options as RemoteAgentInvocationOptions).serverAddress,
    [ATTR_SERVER_PORT]: (options as RemoteAgentInvocationOptions).serverPort,
    [ATTR_GEN_AI_REQUEST_TEMPERATURE]: reqOpts?.temperature,
    [ATTR_GEN_AI_REQUEST_TOP_P]: reqOpts?.topP,
    [ATTR_GEN_AI_REQUEST_TOP_K]: reqOpts?.topK,
    [ATTR_GEN_AI_REQUEST_MAX_TOKENS]: reqOpts?.maxTokens,
    [ATTR_GEN_AI_REQUEST_STOP_SEQUENCES]: reqOpts?.stopSequences?.length ? reqOpts.stopSequences : undefined,
    [ATTR_GEN_AI_REQUEST_FREQUENCY_PENALTY]: reqOpts?.frequencyPenalty,
    [ATTR_GEN_AI_REQUEST_PRESENCE_PENALTY]: reqOpts?.presencePenalty,
    [ATTR_GEN_AI_REQUEST_CHOICE_COUNT]: reqOpts?.choiceCount,
    [ATTR_GEN_AI_REQUEST_SEED]: reqOpts?.seed,
    [ATTR_GEN_AI_REQUEST_ENCODING_FORMATS]: reqOpts?.encodingFormats?.length ? reqOpts.encodingFormats : undefined,
    ...options.attributes,
  };

  if (options.systemInstructions) {
    this.setSystemInstructions(options.systemInstructions);
  }
  if (options.inputMessages && options.inputMessages.length > 0) {
    this.addInputMessages(options.inputMessages);
  }

  this._span.setAttributes(attrs);
}
```

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+}
+
+/**
+ * Format input messages into a JSON string for span attribute storage.
+ *
+ * @param messages - Input messages payload to format.
+ * @returns JSON string representation of input messages, or `undefined` if empty/invalid.
+ */
+export function formatInputMessages(messages: unknown): string | undefined {
```
aabmass: Should we strongly type `messages` here to `InputMessages | string`?

If we accept `unknown`, instrumentations might pass raw vendor SDK payload objects instead of normalizing them into OTel's `ChatMessage` structure first.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * Calculate duration in seconds between startTime and endTime.
+ * Accepts any OpenTelemetry TimeInput (HrTime tuple, millisecond epoch timestamp, or Date).
+ *
+ * @param startTime - Start timestamp as `HrTime`, millisecond timestamp, or `Date`.
+ * @param endTime - Optional end timestamp as `HrTime`, millisecond timestamp, or `Date` (defaults to current time).
+ * @returns Non-negative duration in fractional seconds.
+ */
+export function calculateDurationSeconds(
```
aabmass: `@opentelemetry/core` already exports `hrTimeDuration` and `hrTimeToMilliseconds`. We shouldn't be reimplementing custom duration math (`calculateDurationSeconds` and `hrTimeToSeconds`) in contrib packages.

Please remove `calculateDurationSeconds` and `hrTimeToSeconds` from `utils.ts` and use `@opentelemetry/core` directly:
```ts
import { hrTimeDuration, hrTimeToMilliseconds } from '@opentelemetry/core';

const durationSec = Math.max(0, hrTimeToMilliseconds(hrTimeDuration(start, end)) / 1000);
```

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,244 @@
...
+        explicitBucketBoundaries: GENAI_SERVER_TIME_TO_FIRST_TOKEN_BUCKETS,
+      },
+    }
+  );
+}
+
+/**
+ * Internal helper to validate and record a duration value in seconds.
+ */
+function _recordDuration(
```
aabmass, line 167: Heads up: unlike traces, `sdk-metrics` doesn't drop `undefined` keys during dimension hashing (`hashAttributes`), so passing `{ [ATTR]: undefined }` creates `["attr_name", null]` dimensions. `@opentelemetry/core` has `sanitizeAttributes()`, but it preserves `undefined` values.

We could add a small helper in `utils.ts` to strip undefined keys out:
```ts
function compactAttributes(attrs: Attributes): Attributes {
  return Object.fromEntries(Object.entries(attrs).filter(([_, v]) => v != null && v !== ''));
}
```
And call it inside `_recordDuration` / `recordTokenUsage` so callers in `InferenceInvocation`, `AgentInvocation`, etc. don't need manual `if` checks every time they record metrics.

packages/genai-util/src/invocations/inference.ts
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    if (options.serverAddress) {
+      attrs[ATTR_SERVER_ADDRESS] = options.serverAddress;
+    }
+    if (options.serverPort) {
+      attrs[ATTR_SERVER_PORT] = options.serverPort;
+    }
+
+    const reqOpts = options.requestOptions;
```
aabmass: `requestOptions` unpacking (temperature, topP, maxTokens, etc.) is duplicated across `InferenceInvocation` and `AgentInvocation`. We could pull that into a small helper like `getRequestOptionsAttributes(options?: GenAIRequestOptions): Attributes` in `utils.ts`.

packages/genai-util/src/metrics.ts
aabmass: I'm surprised there's no util for this already. I suppose it might be good practice to keep all the attributes anyway, even if they are undefined. I'm supposing they become empty strings in OTLP anyway which is how most prometheus exporters work iirc

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+import type { Attributes, Span, Context } from '@opentelemetry/api';
+
+/**
+ * Mode of capturing message content (prompts, completions, tool calls).
+ *
+ * NOTE: OpenTelemetry Semantic Conventions define GenAI events (such as
+ * `gen_ai.client.inference.operation.details`) as Log-based Events (LogRecords
+ * with an event name). Because the Logs and Events API in OpenTelemetry
+ * JavaScript is not yet stable, event-based capture modes (`event_only`,
+ * `span_and_event`) are not yet supported in this library. Content capture is
```
aabmass: FWIW it's not stable in python anyway. i think it's reasonable to just to not implement yet though

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+ */
+
+import type { Attributes, Span, Context } from '@opentelemetry/api';
+
+/**
+ * Mode of capturing message content (prompts, completions, tool calls).
+ *
+ * NOTE: OpenTelemetry Semantic Conventions define GenAI events (such as
+ * `gen_ai.client.inference.operation.details`) as Log-based Events (LogRecords
+ * with an event name). Because the Logs and Events API in OpenTelemetry
```
aabmass:
```suggestion
 * with an event name). Because the Logs API in OpenTelemetry
```

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+export interface GenericPart {
+  /** The type of content captured in this part. */
+  type: string;
+  [key: string]: unknown;
+}
+
+/**
+ * Text content part sent to or received from a model.
+ */
+export interface TextPart extends GenericPart {
```
aabmass: nit but throughout here, it feels wrong to extend `GenericPart` since it's an escape hatch type and not the base type. plus the subinterfaces have to override `type`

WDYT of just inlining the `[key: string]: unknown` into each to make it clear?

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,511 @@
...
+export const ATTR_GEN_AI_USAGE_CACHE_READ_INPUT_TOKENS =
+  'gen_ai.usage.cache_read.input_tokens' as const;
+
+/**
+ * The number of tokens written to create the prompt cache.
+ *
+ * @example 128
+ */
+export const ATTR_GEN_AI_USAGE_CACHE_CREATION_INPUT_TOKENS =
+  'gen_ai.usage.cache_creation.input_tokens' as const;
```
JacksonWeber: I think this should be `gen_ai.usage.cache_write.input_tokens`. Looks like it was renamed here: https://github.com/open-telemetry/semantic-conventions-genai/pull/440

packages/genai-util/src/invocations/agent.ts
```diff
@@ -0,0 +1,467 @@
...
+    this._span.setAttribute(ATTR_GEN_AI_AGENT_ID, agentId);
+    return this;
+  }
+
+  public getAgentId(): string | undefined {
+    return this._agentId;
+  }
+
+  public setAgentName(agentName: string): this {
+    this._span.setAttribute(ATTR_GEN_AI_AGENT_NAME, agentName);
```
JacksonWeber, line 141: `setAgentName()` updates the span, but not `_agentName`, so `getAgentName()` still returns the old value.

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+export type OutputMessages = OutputMessage[];
+
+/**
+ * System instructions provided to guide the model.
+ */
+export type SystemInstruction =
+  | string
+  | {
+      parts: MessagePart[];
+    };
```
aabmass: Since we're starting from scratch, maybe we can simplify here and just match the semconv datamodel directly?

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+  /** Duration of the operation in seconds. */
+  durationSeconds?: number;
+  /** Input messages if captured. */
+  inputMessages?: InputMessages;
+  /** Output messages if captured. */
+  outputMessages?: OutputMessages;
+  /** System instructions if captured. */
+  systemInstructions?: SystemInstruction;
+  /** Error if the invocation failed. */
+  error?: Error | unknown;
```
aabmass: do we need to allow `unknown`?

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+  operationName?: string;
+  /** Requested model name. */
+  requestModel?: string;
+  /** Response model name. */
+  responseModel?: string;
+  /** Response ID. */
+  responseId?: string;
+  /** Lifecycle response status. */
+  responseStatus?: string;
+  /** Finish reasons. */
```
aabmass, line 468: nit but a lot of these comments seem unnecessary. were they needed for some link check or OK to remove?

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,503 @@
...
+
+/**
+ * Hook executed when a GenAI operation finishes (success or failure).
+ */
+export interface CompletionHook {
+  /**
+   * Callback invoked when a GenAI operation completes.
+   * If an exception is thrown, it will be safely caught and logged without affecting user flow.
+   */
+  onCompletion(result: CompletionResult): void | Promise<void>;
```
aabmass, line 494: I'm wondering about the Promise variant. These hooks are intended to not block the instrumentation but I can see how returning promises anyway would be handy for tests.

.github/component_owners.yml
JacksonWeber: I'm happy to work on this as well.

psx95: I'll add you to the list of owners! Thanks @JacksonWeber 👍🏻

packages/genai-util/src/utils.ts
psx95: Yeah,  we could do that. I will update this.

packages/genai-util/src/types.ts
psx95: Yeah, I had the documentation generated for most public types. I agree they don't seem too helpful, I can either remove them or have it be regenerated to be more useful.

packages/genai-util/src/types.ts
psx95: I let it be `Error | unknown` because catch by default mostly expects unknown, but thinking about it again this type will just default to `unknown` and so its not a valid reason.

 I think it's ok to restrict type here.

packages/genai-util/src/types.ts
psx95: Right, was a little confused about this one. Is [this](https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/docs/gen-ai/non-normative/models.py#L286) the correct source for data models?

packages/genai-util/src/types.ts
psx95: I don't have a strong opinion about this, but if its not really a base type then I agree that semantically does not make sense to extend it, even if it allows a bit of code reuse.

I will update this.

packages/genai-util/src/semconv.ts
psx95: Good catch, these concerns will further be mitigated when I rebase this to rely on the genai-semconv package. But I will update this for now.

packages/genai-util/src/invocations/agent.ts
psx95: I think the `_agentName` was accidentally marked readonly. Thanks, I will update.

packages/genai-util/src/metrics.ts
psx95: Hmm, so I am a little confused here - why would it be a good practice to keep undefined metrics? 

What's the behavior of Python in this regard?

packages/genai-util/src/invocations/agent.ts
psx95: Sounds good, thanks!

PR #3680, opened 2026-08-18
feat(instrumentation-amqplib): add RabbitMQ vhost and cluster name span attributes

### LINKED ISSUE #3678: `amqplib`: add `messaging.rabbitmq.cluster.name` / `messaging.rabbitmq.vhost.name` span attributes
Without `messaging.rabbitmq.cluster.name`/`messaging.rabbitmq.vhost.name`, there's no way to
tell which RabbitMQ vhost or cluster a message was published to/consumed from. This matters for
anyone running multiple vhosts on a shared broker (multi-tenant) or multiple clusters — without
it, traces from different logical RabbitMQ environments are indistinguishable.

Gate the two span attributes `amqplib` already populates, from data already on the connection
object:

- `messaging.rabbitmq.vhost.name` ← `connectOptions.vhost` (or the URL's path component)
- `messaging.rabbitmq.cluster.name` ← `conn.serverProperties['cluster_name']`

There isn't much room for alternatives to the feature

### Additional context

emantahir05:

Fixes #3678

and `server.port`. Two consequences:

- Messages published to different RabbitMQ virtual hosts on the same broker are
  indistinguishable — same address, same port, same destination name. Anyone running
  multiple vhosts on a shared broker (multi-tenant) can't separate them in traces.
- There is no way to tell one cluster from another, since `server.address` is
  whatever the client dialed (often a load balancer) rather than the broker's identity.

Adds two connection attributes, both from data already present on the connection at
patch time:

- `messaging.rabbitmq.vhost.name` — from `connectOptions.vhost`, or the url path
  component. Resolved the way amqplib itself resolves it: unescaped (so `%2f` becomes
  `/`) and defaulting to `/` when absent.
- `messaging.rabbitmq.cluster.name` — from the `cluster_name` server property, omitted
  when the broker doesn't report one.

### DIFF
--- a/packages/instrumentation-amqplib/src/semconv.ts
+++ b/packages/instrumentation-amqplib/src/semconv.ts
@@ -30,3 +30,25 @@ export const ATTR_MESSAGING_OPERATION = 'messaging.operation' as const;
  * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
  */
 export const ATTR_MESSAGING_SYSTEM = 'messaging.system' as const;
+
+/**
+ * The name of the RabbitMQ virtual host the connection is bound to.
+ *
+ * @example /
+ * @example tenant-a
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_MESSAGING_RABBITMQ_VHOST_NAME =
+  'messaging.rabbitmq.vhost.name' as const;
+
+/**
+ * The name of the RabbitMQ cluster the connection is established with, as
+ * reported by the broker in the `cluster_name` server property.
+ *
+ * @example rabbit@my-broker
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_MESSAGING_RABBITMQ_CLUSTER_NAME =
+  'messaging.rabbitmq.cluster.name' as const;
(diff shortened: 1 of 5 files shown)


packages/instrumentation-amqplib/src/semconv.ts
```diff
@@ -30,3 +30,25 @@ export const ATTR_MESSAGING_OPERATION = 'messaging.operation' as const;
...
+/**
+ * The name of the RabbitMQ virtual host the connection is bound to.
+ *
+ * @example /
+ * @example tenant-a
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_MESSAGING_RABBITMQ_VHOST_NAME =
+  'messaging.rabbitmq.vhost.name' as const;
```
thompson-tomo, line 43: As attribute doesn't already exist should we follow updated guidance on naming & have it as
```suggestion
  'rabbitmq.vhost.name' as const;
```
This way it shouldn't need tobe renamed in the future.

packages/instrumentation-amqplib/src/semconv.ts
```diff
@@ -30,3 +30,25 @@ export const ATTR_MESSAGING_OPERATION = 'messaging.operation' as const;
...
+/**
+ * The name of the RabbitMQ cluster the connection is established with, as
+ * reported by the broker in the `cluster_name` server property.
+ *
+ * @example rabbit@my-broker
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_MESSAGING_RABBITMQ_CLUSTER_NAME =
+  'messaging.rabbitmq.cluster.name' as const;
```
thompson-tomo, line 54: As above
```suggestion
  'rabbitmq.cluster.name' as const;
```

PR #3684, opened 2026-08-19
fix(instrumentation-dataloader): handle synchronous errors and prevent unended span leaks

### LINKED ISSUE #3683: @opentelemetry/instrumentation-dataloader: Synchronous errors in batch function and dataloader operations leak unended spans and omit error status

Latest `@opentelemetry/instrumentation-dataloader` (main / v0.37.0+)

v18+, v20+, v22+, v24+

### What did you do?

Executed DataLoader operations (`batchLoadFn`, `load`, `loadMany`, `prime`, `clear`, `clearAll`) where a synchronous exception occurs (e.g. synchronous batch function error, argument validation error, or a custom `cacheMap` method throwing synchronously).

Example minimal reproduction:

```javascript
const { DataloaderInstrumentation } = require('@opentelemetry/instrumentation-dataloader');
const { InMemorySpanExporter, SimpleSpanProcessor, TracerProvider } = require('@opentelemetry/sdk-trace');
const DataLoader = require('dataloader');

const provider = new TracerProvider({
  spanProcessors: [new SimpleSpanProcessor({ exporter: memoryExporter })]
});
const instrumentation = new DataloaderInstrumentation();
instrumentation.setTracerProvider(provider);
instrumentation.enable();

// Case 1: Synchronous exception in batch function
const failingLoader = new DataLoader(keys => {
  throw new Error('Sync batch failure');
});

try {
  failingLoader.load('key-1');
} catch (err) {
  // Sync error caught
}

console.log('Finished spans count:', memoryExporter.getFinishedSpans().length); // Output: 0 (span leaked!)

// Case 2: Synchronous exception in custom cacheMap during prime()
class FailingMap extends Map {
  set(key, value) {
    throw new Error('Custom cache set failure');
  }
}
const customMapLoader = new DataLoader(async keys => keys, { cacheMap: new FailingMap() });

try {
  customMapLoader.prime('k', 'v');
} catch (err) {
  // Sync error caught
}

console.log('Finished spans count for prime:', memoryExporter.getFinishedSpans().length); // Output: 0 (span leaked!)
```

1. When any DataLoader operation errors synchronously, the started span should be ended via `span.end()`.
2. The span status should be set to `SpanStatusCode.ERROR` with the error message.
3. The exception should be recorded on the span via `span.recordException(err)`.

### What did you see instead?

1. `span.end()` is never called. The created span is left unended and leaked in memory.
2. `memoryExporter.getFinishedSpans()` returns 0 finished spans.
3. No error status or exception telemetry is recorded for the operation.

### Root Cause

In `packages/instrumentation-dataloader/src/instrumentation.ts`:

1. **`_wrapBatchLoadFn`**:
   `batchLoadFn.apply(this, args)` is called directly inside `context.with(...)`. If `batchLoadFn` throws a synchronous exception (e.g., synchronous setup/validation failure inside a non-async batch handler), the exception is thrown before `.catch()` can be attached to the returned Promise. The exception escapes `context.with(...)`, bypassing both `span.recordException(err)` and `span.end()`.

2. **`_getPatchedLoad` & `_getPatchedLoadMany`**:
   `original.call(this, ...args)` is called directly inside `context.with(...)`. If `original.call` throws synchronously (such as parameter validation errors in DataLoader or synchronous dispatch errors), the exception escapes before `.then()` or `.catch()` are chained onto the Promise, bypassing `span.end()`.

3. **`_getPatchedPrime`, `_getPatchedClear`, `_getPatchedClearAll`**:
   These wrappers call `original.call(this, ...args)` inside `context.with(...)` followed by `span.end()`. They lack `try...catch` or `try...finally` handling. When `original.call` throws a synchronous error (e.g. custom `cacheMap` throwing on `.set()`, `.delete()`, or `.clear()`), execution never reaches `span.end()`.

### Impact

- **Memory Leak**: In long-running GraphQL and Node.js services utilizing DataLoader, unended spans accumulate in memory exporters and span processors.
- **Telemetry Inconsistency**: Synchronous failures leave traces incomplete with missing child spans and unrecorded error metrics.

### Affected Components

- `packages/instrumentation-dataloader/src/instrumentation.ts`

### Acceptance Criteria

- All DataLoader instrumentation methods (`_wrapBatchLoadFn`, `_getPatchedLoad`, `_getPatchedLoadMany`, `_getPatchedPrime`, `_getPatchedClear`, `_getPatchedClearAll`) safely handle synchronous exceptions via `try...catch`.
- On synchronous error, `span.recordException(err)` is called, `span.setStatus({ code: SpanStatusCode.ERROR, message: err.message })` is set, and `span.end()` is guaranteed to run.
- Regression tests added in `packages/instrumentation-dataloader/test/dataloader.test.ts` covering synchronous throws across `load`, `loadMany`, `prime`, `clear`, `clearAll`, and `batchLoadFn`.

### Proposed High-Level Solution

Wrap `original.call` / `batchLoadFn.apply` execution in `try...catch` blocks inside the context wrappers:

```typescript
// Example for _wrapBatchLoadFn:
let result: Promise<unknown[]>;
try {
  result = batchLoadFn.apply(this, args) as Promise<unknown[]>;
} catch (err: any) {
  span.recordException(err);
  span.setStatus({
    code: SpanStatusCode.ERROR,
    message: err.message,
  });
  span.end();
  throw err;
}
return result
  .then(value => {
    span.end();
    return value;
  })
  .catch(err => {
    span.recordException(err);
    span.setStatus({
      code: SpanStatusCode.ERROR,
      message: err.message,
    });
    span.end();
    throw err;
  });
```

bhuvan-somisetty:

Fixes #3683

This PR addresses an issue in `@opentelemetry/instrumentation-dataloader` where synchronous errors thrown during DataLoader operations (`batchLoadFn`, `load`, `loadMany`, `prime`, `clear`, `clearAll`) cause active spans to remain un-ended, leaked in memory, and missing error status/exception metrics.

- Introduced `_handleError(span: Span, err: any)` helper that calls `span.recordException(err)`, sets `span.setStatus({ code: SpanStatusCode.ERROR, message: err?.message })`, and ends the span.
- Wrapped `batchLoadFn.apply` and `original.call` invocations inside `try...catch` blocks across all patched DataLoader methods (`_wrapBatchLoadFn`, `_getPatchedLoad`, `_getPatchedLoadMany`, `_getPatchedPrime`, `_getPatchedClear`, `_getPatchedClearAll`) to guarantee synchronous exceptions are properly recorded and spans are always ended.
- Added comprehensive regression tests in `packages/instrumentation-dataloader/test/dataloader.test.ts` covering synchronous throws across `batchLoadFn`, `prime`, `clear`, and `clearAll`.

Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>

### DIFF
--- a/packages/instrumentation-dataloader/src/instrumentation.ts
+++ b/packages/instrumentation-dataloader/src/instrumentation.ts
@@ -129,7 +129,12 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           promise = batchLoadFn.apply(this, args) as Promise<unknown[]>;
         } catch (err: any) {
           instrumentation._handleError(span, err);
-          throw err;
+          // Some dataloader versions (e.g. <2.1) call `_batchLoadFn`
+          // without a try/catch, so a synchronous throw here would
+          // become an uncaught exception instead of a batch rejection.
+          // Returning a rejected promise keeps behavior consistent
+          // across dataloader versions.
+          return Promise.reject(err);
         }
 
         return promise
--- a/packages/instrumentation-dataloader/src/instrumentation.ts
+++ b/packages/instrumentation-dataloader/src/instrumentation.ts
@@ -129,7 +129,7 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           promise = batchLoadFn.apply(this, args) as Promise<unknown[]>;
         } catch (err: any) {
           instrumentation._handleError(span, err);
-          return Promise.reject(err);
+          throw err;
         }
 
         return promise
@@ -139,7 +139,7 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           })
           .catch(err => {
             instrumentation._handleError(span, err);
-            return Promise.reject(err);
+            throw err;
           });
       });
     };
@@ -210,7 +210,7 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           promise = original.call(this, ...args);
         } catch (err: any) {
           instrumentation._handleError(span, err);
-          return Promise.reject(err);
+          throw err;
         }
 
         const result = promise
@@ -220,7 +220,7 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           })
           .catch(err => {
             instrumentation._handleError(span, err);
-            return Promise.reject(err);
+            throw err;
           });
 
         const loader = this as DataloaderInternal;
@@ -270,7 +270,7 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           promise = original.call(this, ...args);
         } catch (err: any) {
           instrumentation._handleError(span, err);
-          return Promise.reject(err);
+          throw err;
         }
 
         // .loadMany never rejects, as errors from internal .load
@@ -282,7 +282,7 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
           })
           .catch(err => {
             instrumentation._handleError(span, err);
-            return Promise.reject(err);
+            throw err;
           });
       });
     };
--- a/packages/instrumentation-dataloader/src/instrumentation.ts
+++ b/packages/instrumentation-dataloader/src/instrumentation.ts
@@ -12,6 +12,7 @@ import {
   trace,
   context,
   Link,
+  Span,
   SpanStatusCode,
   SpanKind,
 } from '@opentelemetry/api';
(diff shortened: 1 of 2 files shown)


packages/instrumentation-dataloader/src/instrumentation.ts
```diff
@@ -114,19 +124,22 @@ export class DataloaderInstrumentation extends InstrumentationBase<DataloaderIns
       );
 
       return context.with(trace.setSpan(parent, span), () => {
-        return (batchLoadFn.apply(this, args) as Promise<unknown[]>)
+        let promise: Promise<unknown[]>;
+        try {
+          promise = batchLoadFn.apply(this, args) as Promise<unknown[]>;
+        } catch (err: any) {
+          instrumentation._handleError(span, err);
+          return Promise.reject(err);
```
JacksonWeber: Could we rethrow here after ending the span instead? DataLoader catches synchronous batch-function errors and wraps them in its own descriptive `TypeError`. Converting the throw into a rejected promise bypasses that.

bhuvan-somisetty: Hi @JacksonWeber, thanks for the review! That makes total sense — I've updated the catch blocks to rethrow the original error after recording and ending the span so DataLoader can handle it natively. Updated the PR!

henrinormak (approved):

Issue #3686, opened 2026-08-20
[instrumentation-aws-sdk] Inject the AWS trace header, so S3 and EventBridge can propagate context


artkhod:
### Is your feature request related to a problem? Please describe

`@opentelemetry/instrumentation-aws-sdk` never writes `X-Amzn-Trace-Id` on an outgoing
request. It propagates context only where a service exposes a _command input_ field to carry
it:

| Service         | Channel             | Where                                |
| --------------- | ------------------- | ------------------------------------ |
| SQS, SNS        | `MessageAttributes` | `services/sqs.ts`, `services/sns.ts` |
| Lambda          | `ClientContext`     | `services/lambda.ts`                 |
| everything else | none                | —                                    |

Grepping the built package for `.headers` returns nothing — no code path touches an HTTP
request header for any service.

That leaves integrations with no message-attribute channel unable to propagate at all. An S3
object-created notification and an EventBridge rule reach their consumer carrying only what
the service itself forwards, and the only thing either forwards is the X-Ray tracing header.
So a Lambda woken by an S3 upload starts a trace disconnected from whatever wrote the object,
and no configuration changes that.

**The header is already being sent, with the wrong value.** The AWS SDK's own
`recursionDetectionMiddleware` (`@aws-sdk/core`, `submodules/client`) sets it on every AWS SDK
call from a Lambda:

```js
const traceId = traceIdFromInvokeStore ?? traceIdFromEnv;
if (nonEmptyString(functionName) && nonEmptyString(traceId)) {
  request.headers[X_AMZN_TRACE_ID] = traceId;
}
```

It populates it from the Lambda invoke store / `_X_AMZN_TRACE_ID` — the _platform's_ X-Ray
trace, unrelated to the active OTel trace. So this is not "add a header AWS services will now
see"; the header is already there on every call. The ask is that it carry the OTel trace.

Prior reports: #1311 (EventBridge propagation, closed as stale). #3267 is adjacent but
concerns which source wins on _extraction_.

### Describe the solution you'd like to see

Inject the AWS trace header on outgoing AWS SDK requests, as the Java instrumentation already
does:

- Compose it with `AWSXRayPropagator` — it already round-trips a span context losslessly into
  `Root`/`Parent`/`Sampled`.
- **Carry every field that is not `Root`/`Parent`/`Sampled`** from the header being replaced
  (or from `_X_AMZN_TRACE_ID`) onto the new one. See "Additional context" — this is the part
[...]

PR #3687, opened 2026-08-20
feat(instrumentation-aws-sdk): inject the AWS trace header

artkhod:
Implements #3686.

## Why

AWS managed services read context only from `X-Amzn-Trace-Id`, so it is the only channel for
an integration with no message attributes. Today this package propagates only where a service
exposes a command input field — `MessageAttributes` for SQS and SNS, `ClientContext` for
Lambda — so an S3 object notification and an EventBridge rule cannot propagate at all, and a
Lambda woken by an upload starts a trace disconnected from whatever wrote the object.

The Java instrumentation already does this unconditionally, and AWS's own JS distribution
patches it in from outside by overriding this package's private
`_getV3SmithyClientSendPatch`. #3686 has the detail and the measurements.

## What

A `build` step middleware added in `patchV3MiddlewareStack`, guarded by a `WeakSet` so each
stack gets it once. `build` is before signing, so the header is covered by the request
signature.

`Root`/`Parent`/`Sampled` come from `AWSXRayPropagator` — it already round-trips a span
context losslessly into that format, so there is no new encoding here.

**Everything else on the header is carried over**, from the header already on the request or
from `_X_AMZN_TRACE_ID`. This is the part worth reviewing. The AWS SDK's own
`recursionDetectionMiddleware` sits on the same `build` step at `low` priority — after this
one — and only acts `if (!request.headers.hasOwnProperty(traceIdHeader))`. Writing the header
here means it never runs, so this inherits what it would have forwarded:

| Field | Owner | Why it matters |
| --- | --- | --- |
| `Lineage` | Lambda | Counts the hops of one triggering event; a function is stopped after ~16 |
| `Self` | ALB | The trace id a load balancer stamped for the hop it handled |

Dropping `Lineage` restarts the chain on every hop, which silently disables recursive loop
detection through Lambda, S3, SQS and SNS. Measured across one S3 hop:

```
header replaced without carrying it : Lineage=1:aa20224c:0   (chain restarted)
header replaced carrying it forward : Lineage=2:aa20224c:0   (chain counted)
```

Unrecognised fields are carried too, so a field AWS adds later needs no change here. That is
the documented convention — an ALB "preserves these fields but does not use them".

## Behaviour change

This is on by default, matching Java. `injectTraceHeader: false` turns it off.

Two things worth being precise about:

- **Inside Lambda**, the header is already set on every AWS SDK call by
  `recursionDetectionMiddleware`, from the invoke store / `_X_AMZN_TRACE_ID` — the platform's
  X-Ray trace, unrelated to the active OTel trace. There the change is which trace it carries,
  not whether it is sent.
- **Outside Lambda**, that middleware needs `AWS_LAMBDA_FUNCTION_NAME` and so does nothing, and
  this adds a header where there was none. Harmless — AWS accepts it and it is what the Java
  instrumentation has always sent — but it is a new header on requests from ECS, EC2 and local
  runs, so worth calling out.

Happy to flip the default to opt-in for a release if you would rather ease it in.

## Alternatives

Covered in #3686. Briefly: `preRequestHook` and the service extensions are handed a
`NormalizedRequest` and never the request, and S3 `PutObject` has no command input field that
becomes this header; `suppressInternalInstrumentation: false` works but adds an HTTP span per
call and clobbers `Lineage` with nowhere to fix it; the trace id the SDK forwards cannot be
set, since `@aws/lambda-invoke-store` lists `X_RAY_TRACE_ID` in `PROTECTED_KEYS` where `set()`
throws.

## Notes for review

- `src/trace-header.ts` is package-internal — `index.ts` is unchanged, so no public API is
  added. Happy to fold it into `src/utils.ts` if you would prefer no new file.
- `@opentelemetry/propagator-aws-xray` is added as a dependency. It is a package in this repo,
  and reusing it keeps one definition of the format. The alternative is ~15 lines of inline
  formatting.
- No CHANGELOG entry, per CONTRIBUTING.

## Tests

`test/trace-header.test.ts` — 7 cases. Three unit-test the carry-through directly; four drive
a real `PutObject` through nock and assert on the header the request went out with.

- sets the tracing header from the active span
- carries `Lineage` from the invocation onto the header
- carries `Self` and unrecognised fields too
- leaves the header alone when disabled
- keeps the platform fields and drops the trace fields
- carries an unrecognised field rather than guess at it
- finds nothing in an absent or empty header

Package suite: **63 passing, 1 pending, 0 failing**. `tsc --noEmit` clean, `eslint` 0 errors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36712,6 +36712,7 @@
       "dependencies": {
         "@opentelemetry/core": "^2.0.0",
         "@opentelemetry/instrumentation": "^0.221.0",
+        "@opentelemetry/propagator-aws-xray": "^2.0.1",
         "@opentelemetry/semantic-conventions": "^1.34.0"
       },
       "devDependencies": {
--- a/packages/instrumentation-aws-sdk/package.json
+++ b/packages/instrumentation-aws-sdk/package.json
@@ -48,6 +48,7 @@
   "dependencies": {
     "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/instrumentation": "^0.221.0",
+    "@opentelemetry/propagator-aws-xray": "^2.0.1",
     "@opentelemetry/semantic-conventions": "^1.34.0"
   },
   "devDependencies": {
--- a/packages/instrumentation-aws-sdk/src/aws-sdk.ts
+++ b/packages/instrumentation-aws-sdk/src/aws-sdk.ts
@@ -23,6 +23,11 @@ import {
 /** @knipignore */
 import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
 import {
+  AWS_TRACE_HEADER,
+  existingTraceHeader,
+  traceHeaderFor,
+} from './trace-header';
+import {
   InstrumentationBase,
   InstrumentationModuleDefinition,
   InstrumentationNodeModuleDefinition,
@@ -58,6 +63,10 @@ export class AwsInstrumentation extends InstrumentationBase<AwsSdkInstrumentatio
   static readonly component = 'aws-sdk';
   // need declare since initialized in callbacks from super constructor
   declare private servicesExtensions: ServicesExtensions;
+  /** Stacks the trace header middleware is already on. */
+  private readonly _stacksWithTraceHeader = new WeakSet<
+    MiddlewareStack<any, any>
+  >();
 
   constructor(config: AwsSdkInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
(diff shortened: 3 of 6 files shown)


trentm (changes requested):
Propagating the AWS-specific trace header is the job of the https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/propagator-aws-xray/ package.
Please correct me if I'm missing something.

artkhod:
The propagator defines the header format, and this PR reuses it for exactly that, but a propagator only runs when an instrumentation calls `inject()` with the request as the carrier, and nothing does that for AWS SDK calls today (only `MessageAttributes` for SQS/SNS and `ClientContext` for Lambda). Even registering `AWSXRayPropagator` globally and injecting at the HTTP layer would write only `Root`/`Parent`/`Sampled`, dropping platform fields like `Lineage` (I raised aws-observability/aws-otel-js-instrumentation#528 for the same issue in the ADOT distro's injection), which is what breaks Lambda's recursion detection. So the goal here is to give the header an injection site that has access to the request and can carry those fields forward, same as the Java instrumentation does.

Issue #3471, opened 2026-04-08
Add sanitization as needed to all database instrumentations


JamieDanielson:
## Summary

Per the [Semantic Conventions for Database Client Spans](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#sanitization-of-dbquerytext), `db.query.text` SHOULD be collected by default only if there is sanitization that excludes sensitive information. Most of the database instrumentations in this repo do not currently sanitize query text, which means they either expose sensitive data by default or don't collect `db.query.text` at all.

This is a tracking issue to add sanitization support across all affected database instrumentations.

## Spec Requirements

From the [Sanitization of `db.query.text`](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#sanitization-of-dbquerytext) section:

- `db.query.text` SHOULD be collected by default only if there is sanitization that excludes sensitive information
- Sanitization SHOULD replace all literals with a placeholder value (string, numeric, date/time, boolean, interval, binary, hexadecimal literals)
- The placeholder value SHOULD be `?`, unless it already has a defined meaning in the given database system
- Parameterized query text SHOULD NOT be sanitized (the static part should be captured as-is)
- IN-clauses MAY be collapsed during sanitization
- Instrumentation MAY truncate the sanitized value for performance considerations

## Reference Implementation

PR #2732 (`instrumentation-mysql2`) is an example of this work being done for a single instrumentation. It added:
- A `maskStatement` boolean config option (to enable/disable sanitization)
- A `maskStatementHook` for user-provided custom sanitization logic
- A default regex-based sanitizer that replaces numeric literals and quoted strings with `?`

## Affected Instrumentations

Each instrumentation below needs to be evaluated and updated. Check the box when a PR has been merged.

### SQL-based (query text is a SQL string)

- [x] `@opentelemetry/instrumentation-mysql2` — #2732
- [ ] `@opentelemetry/instrumentation-mysql`
- [ ] `@opentelemetry/instrumentation-pg`
- [ ] `@opentelemetry/instrumentation-tedious` (MS SQL Server)
- [ ] `@opentelemetry/instrumentation-knex`
- [ ] `@opentelemetry/instrumentation-sequelize`
- [ ] `@opentelemetry/instrumentation-typeorm`
- [ ] `@opentelemetry/instrumentation-oracledb`

### NoSQL / other (query text or statement may contain sensitive values)

- [ ] `@opentelemetry/instrumentation-mongodb`
[...]

PR #3688, opened 2026-08-21
feat(instrumentation-pg)!: sanitize `db.query.text` by default

JamieDanielson:

- Updates #3471 

- This PR adds a new function, `sanitizeSql`, to `@opentelemetry/sql-common`. The function replaces literal values in a SQL string with `?`. It keeps identifiers and `$n` placeholders as-is.
- `instrumentation-pg` now uses this function by default. The instrumentation applies it only to non-parameterized queries, since a parameterized query never contains literal values in its text.
- Two new options control this: `skipQueryTextSanitization` (default `false`) turns sanitization off. `queryTextSanitizationHook` lets you supply your own function instead of the default.
- **This is a breaking change.** A query built by string concatenation will now show `?` instead of its literal values in `db.query.text`. Set `skipQueryTextSanitization: true` to keep the old behavior.
- Known gaps: the function assumes `standard_conforming_strings` is on (the PostgreSQL default). It also does not mask a sign that follows a non-reserved keyword, for example `LIMIT -1`.

### DIFF
--- a/packages/instrumentation-pg/README.md
+++ b/packages/instrumentation-pg/README.md
@@ -89,6 +89,14 @@ The `@opentelemetry/instrumentation-pg` versions 0.72.0 and later emit the stabl
 
 ### Sanitizing `db.query.text`
 
+> [!IMPORTANT]
+> Versions 0.74.0 and later sanitize `db.query.text` for non-parameterized
+> queries by default. Earlier versions recorded it verbatim. Queries built by
+> string concatenation therefore record `?` in place of their literals, which
+> changes the value of an attribute existing dashboards and saved queries may
+> match on. Set `skipQueryTextSanitization: true` to restore the previous
+> behaviour.
+
 The `pg` driver never interpolates parameter values into the query text, so a
 parameterized query records placeholders rather than values — which is why
 the specification advises against
--- a/packages/instrumentation-pg/README.md
+++ b/packages/instrumentation-pg/README.md
@@ -89,17 +89,18 @@ The `@opentelemetry/instrumentation-pg` versions 0.72.0 and later emit the stabl
 
 ### Masking `db.query.text`
 
-By default `db.query.text` records the query text exactly as it was passed to
-`client.query()`. The `pg` driver never interpolates parameter values into the
-query text, so a parameterized query records placeholders rather than values —
-which is why the specification advises against
+The `pg` driver never interpolates parameter values into the query text, so a
+parameterized query records placeholders rather than values — which is why
+the specification advises against
 [sanitizing](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#sanitization-of-dbquerytext)
-parameterized query text.
+parameterized query text. A parameterized call (one made with a `values`
+array) always has its `db.query.text` recorded exactly as passed to
+`client.query()`, regardless of `skipQueryTextSanitization`.
 
 Applications that build SQL by string concatenation put literals into the query
-text, and therefore onto the span. `db.query.text` is masked by default to
-strip those out; set `skipQueryTextSanitization: true` to record it verbatim
-instead. By default:
+text, and therefore onto the span. For a non-parameterized query, `db.query.text`
+is masked by default to strip those out; set `skipQueryTextSanitization: true`
+to record it verbatim instead. By default:
 
 - string, numeric, bit-string and dollar-quoted literals become `?`
 - `--` and (nestable) `/* */` comments are removed
--- a/packages/instrumentation-pg/README.md
+++ b/packages/instrumentation-pg/README.md
@@ -55,12 +55,15 @@ PostgreSQL instrumentation has few options available to choose from. You can set
 
 | Options | Type | Description |
 | ------- | ---- | ----------- |
-| [`enhancedDatabaseReporting`](./src/types.ts#L30) | `boolean` | If true, additional information about query parameters and results will be attached (as `attributes`) to spans representing database operations |
+| `enhancedDatabaseReporting` | `boolean` | If true, additional information about query parameters and results will be attached (as `attributes`) to spans representing database operations |
 | `requestHook` | `PgInstrumentationExecutionRequestHook` (function) | Function for adding custom span attributes using information about the query being issued and the db to which it's directed |
 | `responseHook` | `PgInstrumentationExecutionResponseHook` (function) | Function for adding custom span attributes from db response |
 | `requireParentSpan` | `boolean` | If true, requires a parent span to create new spans (default false) |
 | `addSqlCommenterCommentToQueries` | `boolean` | If true, adds [sqlcommenter](https://github.com/open-telemetry/opentelemetry-sqlcommenter) specification compliant comment to queries with tracing context (default false). _NOTE: A comment will not be added to queries that already contain `--` or `/* ... */` in them, even if these are not actually part of comments_ |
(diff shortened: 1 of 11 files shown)


packages/instrumentation-pg/src/utils.ts
```diff
@@ -202,39 +253,44 @@ export function handleConfigQuery(
   }
 
   // Set attributes
-  if (queryConfig.text) {
-    span.setAttribute(ATTR_DB_QUERY_TEXT, queryConfig.text);
```
trentm, line 206: `db.query.text` should still be set as is if a parameterized query was given, i.e. if `queryConfig.values` is an array, right?

Ah, it looks like `sanitizeSql` is attempting to be smart and handle query text with literals and parameters, e.g.:

```
> common.sanitizeSql("INSERT INTO tbl (foo, bar) VALUES ($1::text,2)", {dialect: 'postgresql'});
'INSERT INTO tbl (foo, bar) VALUES ($1::text,?)'
```

I'm not sure we need/want that. My read of the following from https://opentelemetry.io/docs/specs/semconv/db/database-spans/#sanitization-of-dbquerytext is that we want to *not run `sanitizeSql()` at all when the query is parameterized*:

> Parameterized query text SHOULD NOT be sanitized. Even though parameterized query text can potentially have sensitive data, by using a parameterized query the user is giving a strong signal that any sensitive data will be passed as parameter values, and the benefit to observability of capturing the static part of the query text by default outweighs the risk.

The benefits would be: (a) no need to take the performance hit of running `sanitizeSql()` for parameterized queries, (b) possibly the the sanitizer could be simpler (but I'm not sure about that).

packages/instrumentation-pg/src/types.ts
```diff
@@ -33,13 +33,47 @@ export interface PgInstrumentationExecutionRequestHook {
...
+   * recorded as the `db.query.text` attribute.
+   *
+   * This is independent of {@link enhancedDatabaseReporting}, which controls a
+   * separate attribute holding the raw parameter values. Enabling both records
+   * masked query text alongside unmasked values.
+   *
+   * @default false
+   * @see maskStatementHook
+   */
+  maskStatement?: boolean;
```
trentm: [Update: I see you've mentioned this in the PR description: `Note: Default value of false for now to avoid breaking changes`.]

nit (non-blocking, naming is hard): The eventual final state we'd like is that sanitization of (unparameterized) db.query.text should be *on* by default.  So the preference would probably be for a setting that is the inverse of this, e.g. `skipQueryTextSanitization`, so that the unset falsy value is the desired default behaviour.

Also, s/statement/queryText/ would be nice.
Also perhaps s/mask/sanitization/ would be nice as well given that is what the spec uses for language.

However, that's a pain because instr-mysql2 prior art added `maskStatement` already.
Hence I made this a nit.
(Perhaps this should be discussed more generally on the meta issue: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3471)

packages/instrumentation-pg/src/types.ts
```diff
@@ -33,13 +33,47 @@ export interface PgInstrumentationExecutionRequestHook {
...
 export interface PgInstrumentationConfig extends InstrumentationConfig {
   /**
    * If true, an attribute containing the query's parameters will be attached
    * the spans generated to represent the query.
    */
   enhancedDatabaseReporting?: boolean;
 
+  /**
+   * If true, the query text is masked by {@link maskStatementHook} before it is
+   * recorded as the `db.query.text` attribute.
```
trentm: If the impl changes to only do this for non-parameterized query text, then the description here should be updated to clarify that.

packages/sql-common/src/sanitize.ts
trentm, line 1: Phew, a complete (or complete enough for sanitization) lexer of PostgresSQL dialect. :)
I have not reviewed the logic in here at all.

(This is a kind of thing I'd have hoped the wider npm world would have a battle-tested package for, but I can understand if that isn't the case.)

packages/instrumentation-pg/src/types.ts
maryliag: I was thinking the same thing. I think it should be true by default or use the skip name as Trent suggested. We can make the update on instr-mysql2 accordingly.

I don't want to add a new feature that is not following the semantic, and then having to adjust. At the same time, since this would create a performance hit, having the skip option should be enough

packages/instrumentation-pg/README.md
```diff
@@ -84,6 +87,54 @@ The `@opentelemetry/instrumentation-pg` versions 0.72.0 and later emit the stabl
...
+
+So `SELECT * FROM users WHERE email = 'a@b.c' -- lookup` is recorded as
+`SELECT * FROM users WHERE email = ?`, while
+`SELECT * FROM users WHERE id = $1` is recorded unchanged.
+
+Limits worth knowing:
+
+- Only `db.query.text` is masked. The query sent to PostgreSQL is untouched, and
+  `requestHook` still receives the original text.
+- `db.operation.name` and the span name are derived from the raw query text, not
```
maryliag, line 144: I think you can solve this one by updating the normalize function to something like:

```
export function parseNormalizedOperationName(queryText: string) {
  const sqlCommand = (queryText.trim().match(/^\S+/)?.[0] ?? '').toUpperCase();
}
```

I always have a hard time with regex, so make sure this works with some tests:
```
describe('.parseNormalizedOperationName()', () => {
  it('splits on a tab as well as a space', () => {
    assert.strictEqual(
      utils.parseNormalizedOperationName("SELECT\t'x' FROM t"),
      'SELECT'
    );
  });

  it('splits on a newline', () => {
    assert.strictEqual(
      utils.parseNormalizedOperationName("SELECT\n'secret'"),
      'SELECT'
    );
  });

  it('splits on a carriage return', () => {
    assert.strictEqual(
      utils.parseNormalizedOperationName("UPDATE\r\nt SET c='x'"),
      'UPDATE'
    );
  });

  it('returns the whole trimmed input when it has no whitespace', () => {
    assert.strictEqual(
      utils.parseNormalizedOperationName('  SELECT  '),
      'SELECT'
    );
  });
});

File: packages/sql-common/src/sanitize.ts [hunk 5]
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  // An exponent marker belongs to the number only when digits follow it;
+  // otherwise it starts the next token.
+  if (sql[i] === 'e' || sql[i] === 'E') {
+    let j = i + 1;
+    if (sql[j] === '+' || sql[j] === '-') j++;
+    if (isDigit(sql[j])) {
+      i = j;
+      while (i < n && isDigit(sql[i])) i++;
```
JacksonWeber: Postgres permits underscores between exponent digits. `SELECT 1e123_456` currently becomes `SELECT ?_456`, leaking part of the literal. Could this use `isDigitOrSeparator` and add a regression test?

packages/instrumentation-pg/README.md
maryliag: if you don't wanna use regex, there was a recent update on the oracledb package to deal with the same situation:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/7eb70536040b3419f84a3b3c0f2d1c7995d1c40e/packages/instrumentation-oracledb/src/OracleTelemetryTraceHandler.ts#L87-L99

PR #3692, opened 2026-08-22
fix(instrumentation-kafkajs): record exceptions and handle synchronous errors to prevent span leaks

### LINKED ISSUE #3691: [instrumentation-kafkajs] Missing span.recordException and unended span leaks on synchronous errors

`@opentelemetry/instrumentation-kafkajs`: `v0.30.0` (main branch)

`v18.x`, `v20.x`, `v22.x`

### What did you do?

Used `@opentelemetry/instrumentation-kafkajs` to instrument a KafkaJS consumer or producer where:
1. `eachMessage` or `eachBatch` callback throws a synchronous exception (e.g. input validation failure or non-async error before returning a Promise), or
2. `producer.send` or `producer.sendBatch` throws a synchronous exception (e.g. topic validation error, payload formatting failure, or synchronous socket error), or
3. `eachMessage`, `eachBatch`, `send`, or `sendBatch` rejects asynchronously with an error.

1. For both synchronous and asynchronous errors, all spans created for the operation (`CONSUMER`, `PRODUCER`, or `CLIENT`) should be marked with `SpanStatusCode.ERROR`, have the exception recorded via `span.recordException(err)` (populating `exception.type`, `exception.message`, and stack trace events), and be cleanly ended (`span.end()`).
2. The original exception should be re-thrown without leaking spans in memory.

### What did you see instead?

1. **Unended Span Leak on Synchronous Errors**: When `eachMessage`, `eachBatch`, `send`, or `sendBatch` throws a synchronous exception inside `context.with(...)`, the exception escapes before `_endSpansOnPromise` is called. As a result, `span.end()` is never called, and active spans are leaked in memory.
2. **Missing `span.recordException` on Async Errors**: In `_endSpansOnPromise` (`packages/instrumentation-kafkajs/src/instrumentation.ts`), when a Promise rejects, `_endSpansOnPromise` sets `span.setAttribute(ATTR_ERROR_TYPE, ...)` and `span.setStatus({ code: SpanStatusCode.ERROR, ... })`, but **never calls `span.recordException(reason)`** on any of the spans. As a result, errored spans miss exception telemetry events and stack traces.

#### Affected Component
`packages/instrumentation-kafkajs` (`@opentelemetry/instrumentation-kafkajs`)

#### Root Cause Analysis

1. **Synchronous Exception Leak** in `src/instrumentation.ts`:
   In `_getConsumerEachMessagePatch`:
   ```ts
   const eachMessagePromise = context.with(
     trace.setSpan(propagatedContext, span),
     () => {
       return original!.apply(this, args); // <-- Throws synchronously, escaping context.with
     }
   );
   return instrumentation._endSpansOnPromise(
     [span],
     pendingMetrics,
     eachMessagePromise // <-- Never executed! span.end() is bypassed!
   );
   ```
   Similar unhandled synchronous execution paths exist in `_getConsumerEachBatchPatch`, `_getSendPatch`, and `_getSendBatchPatch`.

2. **Omission of `span.recordException`** in `_endSpansOnPromise`:
   In `_endSpansOnPromise` (lines 702-725):
   ```ts
   .catch(reason => {
     ...
     spans.forEach(span => {
       span.setAttribute(ATTR_ERROR_TYPE, errorType);
       span.setStatus({
         code: SpanStatusCode.ERROR,
         message: errorMessage,
       });
       // Missing: span.recordException(reason);
     });
     throw reason;
   })
   ```

#### Impact
- **Memory Leak**: Active spans remain un-ended in memory on synchronous errors.
- **Incomplete Telemetry**: Errored KafkaJS spans lack exception events and stack traces.

#### Acceptance Criteria
1. Synchronous errors in `eachMessage`, `eachBatch`, `send`, `sendBatch`, and `transaction` operations must be caught, marked with `SpanStatusCode.ERROR`, recorded via `span.recordException(err)`, and ended cleanly before re-throwing.
2. `_endSpansOnPromise` must record the exception on all active spans via `span.recordException(reason)`.
3. Comprehensive regression tests covering synchronous throws and exception recording should be added to `packages/instrumentation-kafkajs/test/kafkajs.test.ts`.

#### Proposed Solution Direction
1. Update `_endSpansOnPromise` to call `span.recordException(reason)` when `reason` is present.
2. Wrap `original.apply(this, args)` calls in `try...catch` blocks in patched methods to guarantee synchronous errors record exception telemetry, end spans, and re-throw cleanly.

bhuvan-somisetty:

- Fixes #3691

- Call `span.recordException(reason)` in `_endSpansOnPromise` when a promise rejects to record exception events, stack traces, and error type attributes in accordance with OpenTelemetry trace semantic conventions.
- Wrap `original.apply` invocations in `_getConsumerEachMessagePatch`, `_getConsumerEachBatchPatch`, `_getSendPatch`, `_getSendBatchPatch`, and `_getProducerTransactionPatch` in `try...catch` blocks so synchronous throws produce rejected promises passed to `_endSpansOnPromise`, preventing unended span leaks and memory leaks.
- Call `span.end()` inside `.then()` and `.catch()` callbacks in `_endSpansOnPromise` so spans end immediately upon promise settlement.
- Added comprehensive unit tests in `packages/instrumentation-kafkajs/test/kafkajs.test.ts` verifying synchronous throw error handling and exception event recording across `eachMessage`, `eachBatch`, producer `send`, and `sendBatch`.

### DIFF
--- a/packages/instrumentation-kafkajs/src/instrumentation.ts
+++ b/packages/instrumentation-kafkajs/src/instrumentation.ts
@@ -713,7 +713,6 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
     return Promise.resolve(sendPromise)
       .then(result => {
         pendingMetrics.forEach(m => m());
-        spans.forEach(span => span.end());
         return result;
       })
       .catch(reason => {
@@ -722,6 +721,7 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
         if (typeof reason === 'string' || reason === undefined) {
           errorMessage = reason;
         } else if (
+          reason !== null &&
           typeof reason === 'object' &&
           Object.prototype.hasOwnProperty.call(reason, 'message')
         ) {
@@ -739,10 +739,12 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
             code: SpanStatusCode.ERROR,
             message: errorMessage,
           });
-          span.end();
         });
 
         throw reason;
+      })
+      .finally(() => {
+        spans.forEach(span => span.end());
       });
   }
 
--- a/packages/instrumentation-kafkajs/src/instrumentation.ts
+++ b/packages/instrumentation-kafkajs/src/instrumentation.ts
@@ -387,12 +387,17 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
           }),
         ];
 
-        const eachMessagePromise = context.with(
-          trace.setSpan(propagatedContext, span),
-          () => {
-            return original!.apply(this, args);
-          }
-        );
+        let eachMessagePromise: Promise<void>;
+        try {
+          eachMessagePromise = context.with(
+            trace.setSpan(propagatedContext, span),
+            () => {
+              return original!.apply(this, args);
+            }
+          );
+        } catch (err: any) {
+          eachMessagePromise = Promise.reject(err);
+        }
         return instrumentation._endSpansOnPromise(
           [span],
           pendingMetrics,
@@ -485,10 +490,12 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
                 )
               );
             });
-            const batchMessagePromise: Promise<void> = original!.apply(
-              this,
-              args
-            );
+            let batchMessagePromise: Promise<void>;
+            try {
+              batchMessagePromise = original!.apply(this, args);
+            } catch (err: any) {
+              batchMessagePromise = Promise.reject(err);
+            }
             spans.unshift(receivingSpan);
             return instrumentation._endSpansOnPromise(
               spans,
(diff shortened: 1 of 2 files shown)


packages/instrumentation-kafkajs/src/instrumentation.ts
```diff
@@ -793,17 +731,18 @@ export class KafkaJsInstrumentation extends InstrumentationBase<KafkaJsInstrumen
         pendingMetrics.forEach(m => m(errorType));
 
         spans.forEach(span => {
+          if (reason !== undefined && reason !== null) {
```
JacksonWeber, line 734: I think this null check happens too late. `Promise.reject(null)` reaches `hasOwnProperty.call(null, 'message')` above, which throws before `span.end()`, replaces the original rejection, and leaks the spans now that `.finally()` is gone. This branch should be null-safe and retain finalization, with a regression test for rejecting `null`.

bhuvan-somisetty: Good catch — fixed. `typeof null === 'object'` was letting a `null` rejection reach `hasOwnProperty.call(null, 'message')`, which threw before `span.end()` and swallowed the original rejection. Restored the `.finally()` for span finalization (decoupled from the error-classification logic) and added a null check before the `hasOwnProperty` call, plus a regression test that rejects with `null` and asserts the span still ends. Pushed in 4f91a0f3.

PR #3699, opened 2026-08-25
feat(instrumentation-openai-agents): add instrumentation

eternalcuriouslearner:

Adds OpenTelemetry instrumentation for the `@openai/agents` JavaScript SDK so agent workflows and function tools can be observed using the GenAI semantic conventions.

- Adds `@opentelemetry/instrumentation-openai-agents` for `@openai/agents >=0.14.0 <1`.
- Registers through the Agents SDK tracing processor callbacks.
- Emits `invoke_workflow`, `invoke_agent`, and `execute_tool` spans while retaining ignored SDK spans for hierarchy propagation.
- Records SDK errors and supports opt-in function-tool argument/result capture.
- Supports preserving or replacing the native OpenAI trace exporter.
- Registers the package with `@opentelemetry/auto-instrumentations-node` and release tooling.
- Adds unit tests and recorded-response SDK integration tests.

## Tests

- `npm test` in `packages/instrumentation-openai-agents` (10 passing)
- `npm run compile` in `packages/instrumentation-openai-agents`
- `npm run lint:readme` in `packages/instrumentation-openai-agents`
- `npm run compile` in `packages/auto-instrumentations-node`
- `npm pack --dry-run --json` in `packages/instrumentation-openai-agents`

## Notes

This will be migrated once util-genai is commonly available. 

P.S: LLM help was used for writing this instrumentation.


packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,308 @@
...
+    this._safely('starting trace span', () => {
+      if (!this._enabled || this._traceRecords.has(trace)) {
+        return;
+      }
+
+      const name = trace.name?.trim();
+      const parentContext = context.active();
+      const attributes: Attributes = {
+        [ATTR_GEN_AI_OPERATION_NAME]:
+          GEN_AI_OPERATION_NAME_VALUE_INVOKE_WORKFLOW,
```
JacksonWeber: Every trace becomes `invoke_workflow`, including standalone `agents.run()`. The spec says not to report workflow spans for standalone agents: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md#invoke-workflow-span

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  onSpanEnd(span: OpenAIAgentsSpan): Promise<void> {
+    this._safely('ending span', () => {
+      const record = this._spanRecords.get(span);
+      if (!record) {
+        return;
+      }
+
+      if (record.span) {
```
JacksonWeber, line 175: Errors on unmapped task/turn spans are discarded, so workflow failures might appear as successful here.

packages/instrumentation-openai-agents/src/processor.ts
eternalcuriouslearner: I removed `invoke_workflow` span for now and I will introduce it in next pr.

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,314 @@
...
+        if (span.spanData.type === 'function' && this._captureMessageContent) {
+          if (span.spanData.input !== undefined) {
+            record.span.setAttribute(
+              ATTR_GEN_AI_TOOL_CALL_ARGUMENTS,
+              this._stringValue(span.spanData.input)
+            );
+          }
+          if (span.spanData.output !== undefined) {
```
JacksonWeber: Should we only set `gen_ai.tool.call.result` when there’s no error? The SDK can include fallback output for failed tool calls, but this attribute is meant to represent a successful result.

eternalcuriouslearner: Good catch!! we should only set `gen_ai.tool.call.result` when the tool execution succeeds. The Python GenAI instrumentation currently captures fallback output before checking span.error; that is a known gap we
need to fix there as well. The current js implementation intentionally checks !span.error, while still capturing the attempted tool arguments on failures. Lmk how it how this looks.

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,369 @@
...
+  onTraceEnd(trace: OpenAIAgentsTrace): Promise<void> {
+    this._safely('ending trace span', () => {
+      const record = this._traceRecords.get(trace);
+      if (!record) {
+        return;
+      }
+      record.span.end();
+      this._openSpans.delete(record.span);
+      this._traceRecords.delete(trace);
+      this._traceRecordsById.delete(trace.traceId);
```
JacksonWeber: The SDK skips `trace.end()` when a run rejects, leaving the run span open and its state in memory. Can we handle cleanup on failure too?

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,369 @@
...
+              ATTR_GEN_AI_TOOL_CALL_RESULT,
+              this._stringValue(span.spanData.output)
+            );
+          }
+        }
+        this._setError(record.span, span, true);
+        record.span.end(this._asTimeInput(span.endedAt));
+        this._openSpans.delete(record.span);
+      } else if (span.error && record.errorTarget) {
+        this._setError(record.errorTarget, span, span.spanData.type === 'task');
```
JacksonWeber, line 198: A failed model attempt marks the agent as ERROR even if the retry succeeds. Can we avoid propagating errors from attempts the SDK recovers from?

packages/instrumentation-openai-agents/src/processor.ts
JacksonWeber: Thanks for working on the above, this covers direct `Runner.run()` rejections. There’s still a case where the failure escapes a caller-managed trace:

```
await withTrace('workflow', async () => {
  await runner.run(agent, input); // rejects
});

```

Here, the trace starts before the runner’s token exists, so `_traceRecordsByRun` has no matching entry and `onRunError()` returns without cleanup. The SDK also skips `trace.end()` when the `withTrace()` callback rejects, leaving the `openai.agents.run` span open and its state retained.

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,469 @@
...
+      );
+      const spanContext = otelTrace.setSpan(parentContext, span);
+      const record: TraceRecord = {
+        context: spanContext,
+        span,
+        errorTarget: span,
+        trace,
+        runToken,
+      };
+      this._traceRecords.set(trace, record);
```
JacksonWeber, line 117: If a tool opens a nested `withTrace()`, this replaces the outer run’s mapping. When the nested trace ends, that mapping is deleted. If the tool then throws, `onRunError()` can’t find the outer trace, leaving its span open. Can we preserve the owning run’s mapping?

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,469 @@
...
+        errorTarget: otelSpan || parentRecord?.errorTarget,
+        agentName,
+        runToken: this._activeRunToken(),
+      };
+      this._spanRecords.set(span, record);
+      this._spanRecordsById.set(span.spanId, record);
+      if (otelSpan) {
+        this._openSpans.add(otelSpan);
+      }
+    });
```
JacksonWeber, line 179: A streamed run can fail after this promise resolves. With `includeTaskAndTurnSpans: false`, the agent gets ERROR, but the run ends with UNSET and no `error.type`. Can we also handle failures from stream completion?

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,498 @@
...
+    this._openSpans.delete(record.span);
+    this._traceRecords.delete(record.trace);
+    this._traceRecordsById.delete(record.trace.traceId);
+    if (record.runToken) {
+      if (this._traceRecordsByRun.get(record.runToken) === record) {
+        if (record.previousRunRecord) {
+          this._traceRecordsByRun.set(
+            record.runToken,
+            record.previousRunRecord
+          );
```
JacksonWeber: If overlapping nested traces A and B finish A-first, B restores the already-ended A rather than the outer trace. A subsequent tool failure leaves the outer run span open. Can we preserve the owning run when nested traces finish out of order?

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,498 @@
...
+        return;
+      }
+      this._recordRunError(record, error, 'OpenAI Agents run failed');
+      this._endTrace(record);
+    });
+  }
+
+  onRunStreamError(runToken: object, error: unknown): void {
+    this._safely('recording failed streamed run', () => {
+      const record = this._traceRecordsByRun.get(runToken);
```
JacksonWeber: With `withTrace(..., () => runner.run(..., { stream: true }))`, the trace starts before the runner token exists, so this lookup misses it. With task/turn spans disabled, a stream failure still leaves the run with `UNSET` and no `error.type`

packages/instrumentation-openai-agents/src/processor.ts
```diff
@@ -0,0 +1,498 @@
...
+      if (record) {
+        this._recordRunError(record, error, 'OpenAI Agents stream failed');
+      }
+    });
+  }
+
+  onTraceError(trace: OpenAIAgentsTrace, error: unknown): void {
+    this._safely('ending failed trace', () => {
+      const record =
+        this._traceRecords.get(trace) ??
```
JacksonWeber, line 310: When `getOrCreateTrace()` reuses an outer trace, a callback rejection ends that trace even if the outer scope catches the error and continues. Can we limit cleanup to traces this call creates and leave reused traces under their owner’s lifecycle?

JacksonWeber (approved):

PR #3702, opened 2026-08-26
fix(deps): move unused @types packages out of runtime dependencies

pichlermarc:

None of these packages are used in the public API of the consuming packages. A few others (`pg` most notably) still remain that need to depend on these types as they are used in the public API - these are not touched by this PR.

I'll follow up with changing the types on `@opentelemetry/instrumentation-pg` later as it's blocking #3538


aryamohanan (approved):

maryliag (approved):

seemk (approved):

pichlermarc:
`test-all-versions (24)` is failing due to `min-release-age: 3` (we're trying to install packages that are younger than that). I've opened https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3704 to fix that for now.

Issue #3697, opened 2026-08-24
redis-common: HGETALL spans all look the same because the key is not serialized


iblancasa:
### What version of OpenTelemetry are you using?

- `@opentelemetry/redis-common` 0.38.3
- `@opentelemetry/instrumentation-redis` 0.69.0
- `redis` 6.2.1

### What version of Node are you using?

v26.7.0

### What did you do?
I ran a few normal redis commands against a local redis 7 server with the redis instrumentation enabled and the default settings, then looked at the `db.query.text` attribute on the spans.

```js
await client.get('session:42');
await client.hGetAll('user:1');
await client.ttl('session:42');
await client.mGet(['session:42', 'user:1']);
```

### What did you expect to see?

That every one of them tells me which key it touched, the way `GET` already does. The key is the part I need when I am trying to work out which cache lookup is slow, or which hash is being read on a hot path.

### What did you see instead?

Some commands keep the key and some drop it:

```
GET session:42
HGETALL [1 other arguments]
TTL [1 other arguments]
MGET [2 other arguments]
```

`GET` tells me which key it read. `HGETALL` does not, so every hash read in my app produces the exact same string and I cannot tell them apart. If a request reads twenty different hashes, I get twenty identical spans. I cannot filter by key, group by key, or work out which one is the slow one.

The same thing happens to `TTL` and `MGET`, and once I started looking, to `HGET`, `HDEL`, `TYPE`, `KEYS`, `LPOP`, `RENAME`, `TOUCH` and `UNLINK` as well.

### Additional context

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. [Learn more here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

iblancasa:
I would like to propose a PR to fix this. Will do it as soon as possible.

PR #3703, opened 2026-08-26
feat(redis-common, instrumentation-redis, instrumentation-ioredis): add serializeKeys option

iblancasa:

This PR is to address #3697.

I implemented it in a way to avoid breaking current users. But I'm ok if you think we should implement this in a different way.

The list in `redis-common/src/index.ts` maps each Redis command to a number of arguments to serialize. Anything that does not match, removes all of them. Some commands were not taken into account.

Just adding them to the list would fix the inconsistency but it would also produce high cardinality output for commands that currently produce a constant string, which changes the cost and shape of existing metrics (or derived from spans like the ones generated by `span metrics` connector from the OTEL Collector) without anyone opting in. So this adds an option instead, and leaves the default exactly as it is.

A new `serializeKeys` option, off by default. When it is on, those 37 commands serialize their arguments, and the statement is masked so that the parts identifying a single entity are replaced with `?` (following semantic conventions).

```javascript
const { RedisInstrumentation } = require('@opentelemetry/instrumentation-redis');

const redisInstrumentation = new RedisInstrumentation({
  serializeKeys: true,
});
```

The same option is available on `IORedisInstrumentation`.

Given an application doing this:

```javascript
await client.hGetAll('player:1261821:stats');
await client.hGetAll('player:1261822:stats');
await client.hGetAll('tournament:leaderboard:sc:daily');
await client.ttl('auth_revoked:1261821');
await client.mGet(['player:1261821:stats', 'player:1261822:stats']);
await client.set('session:42', 'some value');
```

the `db.query.text` attribute comes out like this:

| default | `serializeKeys: true` |
| - | - |
| `HGETALL [1 other arguments]` | `HGETALL player:?:stats` |
| `HGETALL [1 other arguments]` | `HGETALL player:?:stats` |
| `HGETALL [1 other arguments]` | `HGETALL tournament:leaderboard:sc:daily` |
| `TTL [1 other arguments]` | `TTL auth_revoked:?` |
| `MGET [2 other arguments]` | `MGET player:?:stats player:?:stats` |
| `SET session:42 [1 other arguments]` | `SET session:? [1 other arguments]` |

### DIFF
--- a/packages/instrumentation-ioredis/README.md
+++ b/packages/instrumentation-ioredis/README.md
@@ -100,7 +100,7 @@ requestHook: function (
 By default, commands that accept only a key are serialized without it, so every
 call looks the same:
 
-```
+```text
 hgetall [1 other arguments]
 ttl [1 other arguments]
 ```
--- a/packages/instrumentation-ioredis/README.md
+++ b/packages/instrumentation-ioredis/README.md
@@ -95,6 +95,54 @@ requestHook: function (
 
 ```
 
+#### Serializing redis keys
+
+By default, commands that accept only a key are serialized without it, so every
+call looks the same:
+
+```
+hgetall [1 other arguments]
+ttl [1 other arguments]
+```
+
+Set `serializeKeys` to `true` to record the key instead. Segments that identify
+a single entity are replaced with `?`, so the key pattern is visible without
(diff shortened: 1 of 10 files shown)


iblancasa:
I can be a code owner for that package if needed.

PR #3706, opened 2026-08-28
fix(instrumentation-dns): preserve promisified lookup result

gentlelius:

When `dns.lookup` is patched, its non-enumerable symbol properties are not preserved. This drops the Node.js internal `Symbol(customPromisifyArgs)` metadata that `util.promisify()` uses for callbacks with multiple success values.

As a result, `promisify(dns.lookup)(hostname)` resolves with only the address string while DNS instrumentation is enabled, instead of the native `{ address, family }` object. Applications that rely on the documented promisified result shape can then observe `result.address` as `undefined`.

- Copy symbol property descriptors from the original `dns.lookup` function to the patched function.
- Add an integration test confirming that `util.promisify(dns.lookup)` preserves `{ address, family }` and still exports a DNS span.

## Testing

- `npm run compile:with-dependencies` in `packages/instrumentation-dns`
- `npm test` in `packages/instrumentation-dns` (48 passing)
- ESLint on both changed files (no errors)
- Prettier check on both changed files

## AI use disclosure

This contribution was prepared with Codex assistance for investigation, implementation, test drafting, and PR preparation. The reported behavior and the resulting patch were reproduced and verified with the package test suite.

### DIFF
--- a/packages/instrumentation-dns/src/instrumentation.ts
+++ b/packages/instrumentation-dns/src/instrumentation.ts
@@ -83,7 +83,19 @@ export class DnsInstrumentation extends InstrumentationBase<DnsInstrumentationCo
    */
   private _getLookup() {
     return (original: (hostname: string, ...args: unknown[]) => void) => {
-      return this._getPatchLookupFunction(original);
+      const patchedLookup = this._getPatchLookupFunction(original);
+
+      // Preserve symbols such as Node.js' internal customPromisifyArgs.
+      // util.promisify() relies on it to combine address and family into an
+      // object instead of resolving with only the address string.
+      for (const symbol of Object.getOwnPropertySymbols(original)) {
+        const descriptor = Object.getOwnPropertyDescriptor(original, symbol);
+        if (descriptor !== undefined) {
+          Object.defineProperty(patchedLookup, symbol, descriptor);
+        }
+      }
+
+      return patchedLookup;
     };
   }
 
--- a/packages/instrumentation-dns/test/integrations/dns-lookup.test.ts
+++ b/packages/instrumentation-dns/test/integrations/dns-lookup.test.ts
@@ -14,6 +14,7 @@ import * as dns from 'dns';
 import * as utils from '../utils/utils';
 import { assertSpan } from '../utils/assertSpan';
 import { SpanStatusCode } from '@opentelemetry/api';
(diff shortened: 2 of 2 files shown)


gentlelius:
This is a compatibility bug fix rather than a feature: enabling DNS instrumentation changes the documented `util.promisify(dns.lookup)` result from `{ address, family }` to an address string because the patched function drops Node.js symbol metadata. The PR includes a regression test for both the native result shape and span creation.

The unmaintained-package workflow exempts PRs carrying the `bug` label, but I do not have permission to apply it. Could a maintainer please add the `bug` label and approve the first-time-contributor Actions workflows?

gentlelius:
Hi @pichlermarc, would you be willing to sponsor or review this small compatibility bug fix for the unmaintained DNS instrumentation package? With instrumentation enabled, `util.promisify(dns.lookup)` no longer preserves Node.js native `{ address, family }` semantics because the wrapper drops symbol metadata. The patch copies the original symbol descriptors and adds an integration test that also verifies span creation; the package compile and all 48 tests pass locally.

If another approver would be a better fit, please let me know. Thank you.

gentlelius:
Hi @trentm, could you please help triage this compatibility bug fix before the unmaintained-package autoclose deadline? The PR preserves Node.js `util.promisify(dns.lookup)` return semantics after instrumentation and includes a regression test; compile and all package tests pass locally.

The remaining maintainer-only actions are applying the `bug` exemption label and approving the first-time-contributor CI workflows. I previously asked @pichlermarc but may have caught them at a busy time. Thanks for taking a look.

gentlelius:
Hi @david-luna, could you please help triage this small DNS instrumentation compatibility bug fix? The unmaintained-package 14-day autoclose window is now due.

The wrapper currently drops Node.js symbol metadata, changing `util.promisify(dns.lookup)` from `{ address, family }` to a string. This PR preserves the original symbol descriptors and adds a regression test; package compilation and all 48 tests pass locally.

The remaining maintainer-only steps are adding the `bug` exemption label and approving the first-time-contributor Actions workflows. A review or redirect to another approver would be greatly appreciated. Thank you.

david-luna (approved):
@gentlelius thanks for your contribution. I'll wait a bit is anyone else wants to review

PR #3709, opened 2026-08-28
feat(genai-util): scaffold package, types, semconv and core utils

psx95:
Re: #3681 

Generative AI instrumentations in OpenTelemetry JavaScript (such as OpenAI, Anthropic, and LangChain) require standardized utilities for GenAI semantic conventions, shared data models, content capture settings, and attribute extraction helpers.
This PR introduces the foundational package structure for `@opentelemetry/genai-util` to provide these shared utilities across GenAI instrumentation packages.

- **Package Scaffolding & Config**: Added `packages/genai-util` with `package.json`, `tsconfig.json`, `LICENSE`, `README.md`, and configured monorepo integration (`.github/component_owners.yml`, `release-please`, `knip.jsonc`).
- **Semantic Conventions (`src/semconv.ts`)**: Added constants for OpenTelemetry GenAI attributes, operation names, providers, finish reasons, and metric names.
- **Core Types (`src/types.ts`)**: Added TypeScript interfaces for GenAI messages, request options, token usage, and content capture modes.
- **Environment Variables (`src/environment-variables.ts`)**: Added parser and resolution helpers for `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`.
- **Utilities (`src/utils.ts`)**: Added helper functions for error classification (`getErrorType`), server URL attribute extraction (`getAttrsFromBaseURL`), request option attribute mapping, and message serialization.
- **Unit Tests**: Added test suite covering semantic conventions, environment variable resolution, and utility functions.

## Important Info
 - This is the **first PR** in a series that attempt to breakdown PR #3677. The files in this PR are taken from PR #3677 and did not require any modifications for the package to compile (except `README.md` - a note was added there).


packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,514 @@
...
+  [key: string]: unknown;
+}
+
+/**
+ * Binary / multimodal blob part (e.g. image, audio, video, document).
+ */
+export interface BlobPart {
+  type: 'blob';
+  /** Modality of the content (e.g. image, audio, video). */
+  modality?: string;
```
JacksonWeber: Could we align these message types with the sem conv data model: https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/model/gen-ai/gen-ai-input-messages.json)](https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/model/gen-ai/gen-ai-input-messages.json?

Just a few I noticed, `BlobPart` requires `modality` and `content`, `FilePart`/`UriPart` require `modality`; `OutputMessage` requires `finish_reason`.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,275 @@
...
+ */
+export function serializeContent(content: unknown): string {
+  if (typeof content === 'string') {
+    return content;
+  }
+  if (content === null || content === undefined) {
+    return '';
+  }
+  try {
+    return JSON.stringify(content);
```
JacksonWeber: Should we fallback here as well if the result is undefined but does not throw?

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * NOTE: Currently returns false until the OpenTelemetry JS Logs & Events SDK stabilizes.
+ *
+ * @experimental This function is experimental and subject to change.
+ */
+export function isEventContentCaptureEnabled(
+  _mode: ContentCaptureMode
+): boolean {
+  return false;
+}
```
lmolkova: do we need this method right away? We can always add it later when event support is added. This file would need to change then anyway.

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,514 @@
...
+/**
+ * Options for starting an inference invocation.
+ */
+export interface InferenceInvocationOptions {
+  /** Name of the provider (e.g. 'openai', 'anthropic', 'aws.bedrock'). */
+  providerName: string;
+  /** Operation name (e.g. 'chat', 'text_completion', 'generate_content'). Defaults to 'chat'. */
+  operationName?: string;
+  /** Model name requested. */
+  requestModel?: string;
```
lmolkova, line 354: not sure how this class will be used, but some of these parameters (model name, provider name, operation name, and maybe a few more should be provided at span start time and should not be settable after. We should not really populate all these attributes before span starts - e.g. we'd rather check if span is sampled before creating input messages and additional attributes.

So, it seems we need at least two types to separate start-time attributes from maybe-later attributes. Having one type would be confusing and error prone (some overriding operation or model name by accident). Same argument can be made for other options.

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,514 @@
...
+  /** Custom initial span attributes. */
+  attributes?: Attributes;
+}
+
+/**
+ * Options for starting an agent invocation (e.g. local in-process agent).
+ */
+export interface AgentInvocationOptions {
+  /** Unique ID of the agent. */
+  agentId?: string;
```
lmolkova: this and agent version only apply to remote agents

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,514 @@
...
+/**
+ * Options for starting a workflow invocation.
+ */
+export interface WorkflowInvocationOptions {
+  /** Name of the workflow. */
+  workflowName: string;
+  /** Parent context. */
+  parentContext?: Context;
+  /** Custom initial span attributes. */
+  attributes?: Attributes;
```
lmolkova: also conversation id, input messages https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md#invoke-workflow-span

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,275 @@
...
+
+/**
+ * Construct standardized span name following GenAI semantic conventions.
+ * E.g. "chat gpt-4o", "embeddings text-embedding-3-small", "generate_content gemini-1.5-pro".
+ *
+ * @param operationName - GenAI operation name (e.g. `'chat'`, `'embeddings'`).
+ * @param model - Optional model name or identifier.
+ * @returns Standardized span name in the format `"{operation} {model}"` or `"{operation}"`.
+ */
+export function getSpanName(operationName: string, model?: string): string {
```
lmolkova: i think it's over-generalization - there are spans that have different name format. Perhaps it should be a method on something inference and embeddings-specific?

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,275 @@
...
+      constructorName !== 'Object'
+    ) {
+      return constructorName;
+    }
+    return error.name || 'Error';
+  }
+  if (typeof error === 'string' && error.trim().length > 0) {
+    return error;
+  }
+  return 'Error';
```
lmolkova: the default value is _OTHER - https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/registry/attributes/error.md

```suggestion
  return '_OTHER';
```

packages/genai-util/README.md
```diff
@@ -0,0 +1,310 @@
...
+## Installation
+
+```bash
+npm install --save @opentelemetry/genai-util
+```
+
+## Features
+
+- **Semantic Conventions**: Complete constants and enumerations for OpenTelemetry GenAI semantic conventions (attributes, operation names, providers, finish reasons, token types, metrics, events).
+- **TelemetryHandler**: Central lifecycle façade managing spans, metrics, events, and completion hooks for LLM and GenAI operations.
```
lmolkova, line 24: I don't see this file in this PR, should it be here?

packages/genai-util/README.md
```diff
@@ -0,0 +1,310 @@
...
+  requestModel: 'gpt-4o',
+  requestOptions: {
+    temperature: 0.7,
+    maxTokens: 1000,
+  },
+  inputMessages: [
+    {
+      role: 'user',
+      parts: [{ type: 'text', content: 'What is OpenTelemetry?' }],
+    },
```
lmolkova: to my other comment - these friends should not be provided at start time (so we can optimize perf if sampling)

lmolkova (commented):
Thanks! left a few comments on the semconv compliance

packages/genai-util/src/environment-variables.ts
psx95: No, we could remove it for now. I left it as is because it was convenient during split.

packages/genai-util/README.md
psx95: No, I did not add that file yet. I copied this README as is from the final PR and added a Note at the very beginning of this README instead of going through this entire README and editing it in every subsequent PR.

packages/genai-util/src/types.ts
psx95: > not sure how this class will be used

From the original PR: [inference.ts](https://github.com/psx95/opentelemetry-js-contrib/blob/1f2178c141096a0f8679336afcbe0d58de537f40/packages/genai-util/src/invocations/inference.ts#L74). 

> but some of these parameters (model name, provider name, operation name, and maybe a few more should be provided at span start time and should not be settable after. 

Setters are **not provided for all of them** in the implementing class.

> We should not really populate all these attributes before span starts - e.g. we'd rather check if span is sampled before creating input messages and additional attributes.

Will there be absolutely no situation where such options cannot be provided in the beginning?

packages/genai-util/src/types.ts
psx95: Referred to the [table here](https://github.com/open-telemetry/semantic-conventions-genai/blob/5ca9052bc796ef1e497200b1d558fd87a201f335/docs/registry/attributes/gen-ai.md) - How do I infer that `gen_ai.agent.version` and `gen_ai.agent.id` only apply to remote agents?

packages/genai-util/src/types.ts
psx95: I'll add these. Thanks!

packages/genai-util/src/utils.ts
psx95: Hmm, I just looked into it again - my assumption was `content == null || content == undefined` was enough to make it safe. 

But someone could pass a lambda into the function and it would apparently  crash. It's rare, but could happen. I will update this.

packages/genai-util/src/utils.ts
psx95: This is meant to be used as a utility method, we could add inference specific methods in the inference classes  - but I could remove it as well.

packages/genai-util/README.md
psx95: Could you expand a bit on this? Are you talking about `requestOptions` ?

packages/genai-util/src/types.ts
psx95: I will go through the models again to make sure they're aligned.

psx95: `OutputMessage` requiring `finish_reason` seems to be [deprecated](https://github.com/open-telemetry/semantic-conventions-genai/blob/ac46a5d7bfe0b0f47e8ce393e2db3a2c3042f236/model/gen-ai/gen-ai-output-messages.json#L299).

I will add the others.

packages/genai-util/src/types.ts
lmolkova: this is a list of all attributes in GenAI namespace, this is the list of attributes (in all namespaces) that are applicable to internal agent span - https://github.com/open-telemetry/semantic-conventions-genai/blob/5ca9052bc796ef1e497200b1d558fd87a201f335/docs/gen-ai/gen-ai-agent-spans.md#invoke-agent-internal-span, same doc includes definition of the client span

packages/genai-util/README.md
lmolkova: some attributes on the spans are explicitly called out as sampling-relevant - they should be provided at start time (if provided at all). Otherwise they can't participate in head-based sampling decisions.

The API we design here should separate them from everything else that can be added after.

E.g. on inference span the following attributes are sampling-relevant:

> * [`gen_ai.operation.name`](/docs/registry/attributes/gen-ai.md)
> * [`gen_ai.provider.name`](/docs/registry/attributes/gen-ai.md)
> * [`gen_ai.request.model`](/docs/registry/attributes/gen-ai.md)
> * [`server.address`](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/registry/attributes/server.md)
> * [`server.port`](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/registry/attributes/server.md)

https://github.com/open-telemetry/semantic-conventions-genai/blob/5ca9052bc796ef1e497200b1d558fd87a201f335/docs/gen-ai/gen-ai-spans.md?plain=1#L337-L344

so the API should look like


```
invocation = start_inference(operationName, providerName, requestModel, serverAddress, serverPort)

// we can maybe shave some perf off by checking if span is sampled, etc and then 
// whenever before span is finished add other attributes

if (invocation.span.isRecorded()) {
  invocation.request_temperature = get_temp();  // not saying it has to be top-level property, just that it's ok to provide it later
  if (content_enabled) {
    invocation.input_messages = parse_input_messages()
  }
...
```

So however we design it, we need to find a good way to pass sampling-relevant attributes at start time. If we pass ALL start attributes at start time we won't be able to optimize depending on sampling decision. E.g.

```
request_temperature = get_temp()
input_messages = parse_input_messages()

invocation = start_inference(operationName, providerName, requestModel, serverAddress, serverPort, temperature, inputMessages)
......
```

would block the ability to optimize and postpone parsing inputs until span has started and we know someone would need them.

psx95: So the way current API would like is [shown here](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677/changes#diff-5079b28b1fbdb982cb18957cc3b5dc2ca15278ac4cdf843ad14b236e0da9a6f4R183) (From the original, un-split PR): 

```typescript
public startInference(
    options: InferenceInvocationOptions
  ): InferenceInvocation 
```
Now from [InferenceInvocationOptions defined in types](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/changes#diff-d1f6c20cf32b07382a979c2825e19ad382c1b843addadb7e257487a808be55bbR359), almost all of the options are optional so its not necessary to define all of them at the beginning. 

And the [InferenceInvocation can allow setting the attributes we need to set as configurable post-creation](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677/changes#diff-49519c3dfa6e862e94cec2371a165a1f9351b9d4a1905522e08d42d0a573b63cR196) (Again, from the original un-split PR). 

When we add support for these invocations, we can look more closely into what needs to be allowed at the beginning (and not settable post creation) vs what is optional at the beginning and configurable later on.

psx95: Discussed offline. Will limit invocation options to start-time attributes only. For other attributes, we can add setter functions  in the actual invocation classes.

packages/genai-util/src/types.ts
psx95: Discussed offline. This is addressed in commit [77af2e3](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/77af2e37b95af3d79e859ab62947d04a73f57e6a).

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,76 @@
...
+export const ENV_GENAI_COMPLETION_HOOK =
+  'OTEL_INSTRUMENTATION_GENAI_COMPLETION_HOOK' as const;
+
+/**
+ * Parse an arbitrary input value into a valid ContentCaptureMode.
+ *
+ * @experimental This function is experimental and subject to change.
+ */
+export function parseContentCaptureMode(
+  value?: string | boolean | ContentCaptureMode
```
DylanRussell: you sure you want to accept bool ?? doesn't matter that much

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,473 @@
...
+  /** A URI referencing attached data. */
+  uri: string;
+  /** Modality of the content (e.g. image, audio, video, document). */
+  modality: Modality;
+  /** The IANA MIME type of the attached data. */
+  mime_type?: string | null;
+  [key: string]: unknown;
+}
+
+/**
```
DylanRussell, line 185: do we need these ? we can't just use `GenericPart` ? I don't think we have equivalents in the python repo

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,473 @@
...
+ * List of input messages sent to the model.
+ */
+export type InputMessages = ChatMessage[];
+
+/**
+ * Output message generated by the model for a specific choice / candidate.
+ */
+export interface OutputMessage {
+  /** Role of the entity that created the message (typically 'assistant'). */
```
DylanRussell, line 272: the python equivalent has a finish reason field: https://github.com/open-telemetry/opentelemetry-python-genai/blob/main/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L318

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,277 @@
...
+  return attrs;
+}
+
+/**
+ * Serialize arbitrary data to a JSON string or string representation safely.
+ *
+ * @param content - Content to serialize (object, primitive, or nullish).
+ * @returns JSON string for objects, empty string for null/undefined, or plain string.
+ */
+export function serializeContent(content: unknown): string {
```
DylanRussell, line 38: Python uses dataclasses with asdict(). Any field with a None default is retained vs JSON.stringify strips them.. I think stripping them is better TBH but figured I'd point this out..

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+}
+
+/**
+ * Format input messages into a JSON string for span attribute storage.
+ *
+ * @param messages - Input messages payload to format.
+ * @returns JSON string representation of input messages, or `undefined` if empty/invalid.
+ */
+export function formatInputMessages(
```
DylanRussell, line 116: another difference to consider.. in python we base64 encode bytes to a string.. I think this does something different.. should we do the same thing here?

DylanRussell (approved):

packages/genai-util/src/types.ts
psx95: `OutputMessage` requiring `finish_reason` seems to be [deprecated](https://github.com/open-telemetry/semantic-conventions-genai/blob/ac46a5d7bfe0b0f47e8ce393e2db3a2c3042f236/model/gen-ai/gen-ai-output-messages.json#L299).

packages/genai-util/src/types.ts
psx95: Python is using `Any` (See [ref](https://github.com/open-telemetry/opentelemetry-python-genai/blob/b301b3227c1bf921c464b2d25affce22f4165e1b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L112)). Did not use `GenericPart` since it  semantically indicates a type of `MessagePart`. 

Although practically, it could be used here as well. I updated the code slightly to reduce duplication by using type alias.

packages/genai-util/src/utils.ts
psx95: Ack. This is a general purpose utility function - as long as it is consistent with the documentation, I'm ok with it. Might change it if most instrumentations require different behavior than what is being done here.

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,473 @@
...
+  /** Server port. */
+  serverPort?: number;
+  /** Custom initial span attributes. */
+  attributes?: Attributes;
+}
+
+/**
+ * Options for starting a tool execution invocation.
+ */
+export interface ToolInvocationOptions {
```
lmolkova, line 382: `toolArguments`, `toolDescription`, and `toolCallId` are not sampling-relevant. In particular, `toolArguments` is opt-in payload content and can be large.

Start options should only take sampling-relevant properties (`toolName`, `toolType`, `agentName`), `parentContext`, and custom initial `attributes`. Payload data and post-start attributes belong on the invocation instance.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,277 @@
...
+ * 1. Error `code` if present and non-empty (e.g., `'ECONNREFUSED'`, `404`, `'429'`)
+ * 2. Explicit `error.name` if set and not the default `'Error'` (e.g., `'RateLimitError'`)
+ * 3. Class / constructor name for subclasses (e.g., `class CustomAPIError extends Error`)
+ * 4. Non-empty string if `error` is a string
+ * 5. Fallback `'_OTHER'`
+ *
+ * @param error - The caught error, exception, or value.
+ * @returns The standardized error type string.
+ */
+export function getErrorType(error: unknown): string {
```
lmolkova, line 261: A few notes on `getErrorType`:

1. When `error` is a string (`typeof error === 'string'`), returning it directly as `error.type` causes cardinality explosion on thrown messages. The convention in this repo (see [instrumentation-kafkajs](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/20bbc8c8a9827d7add9e86aa734d8601932736fc/packages/instrumentation-kafkajs/src/instrumentation.ts#L782-L797)) is to fall back to `_OTHER` for strings/primitives and treat the string as the error message. The same attribute with the same value is used on metrics, that's why limited cardinality is important

2. (Nice to have): An optional resolver callback (e.g. `errorTypeResolver?: (err: unknown) => string | undefined`, matching Python's [`ErrorTypeResolver`](https://github.com/open-telemetry/opentelemetry-python-genai/blob/85f3fa0c2ecf667dc9ffc1fce9e599eb80886713/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L258-L262)) would give instrumentations a clean hook to map SDK-specific error types/status codes before falling back.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,277 @@
...
+  return '_OTHER';
+}
+
+/**
+ * Extract OpenTelemetry span attributes from GenAI request options.
+ *
+ * @param requestOptions - Optional GenAI request options to extract attributes from.
+ * @returns Attributes object populated with GenAI request semantic conventions.
+ */
+export function getRequestOptionsAttributes(
```
lmolkova, line 292: `getRequestOptionsAttributes` and `GenAIRequestOptions` seems to be dead code (only used in test) now that `requestOptions` was removed from `InferenceInvocationOptions`. s

packages/genai-util/src/utils.ts
psx95: Are you talking about [`_GenAiJsonEncoder`](https://github.com/open-telemetry/opentelemetry-python-genai/blob/b301b3227c1bf921c464b2d25affce22f4165e1b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/utils.py#L156)?

This function is used [like this](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677/changes#diff-11baa1ac0dd9aa430e584ed6ca7fefcff17ebbf052e1be21d1dc039c6243d136R253) to get a formatted string that gets attached as a span attribute. I guess technically the input could contain bytes. I'll look into it.

packages/genai-util/README.md
```diff
@@ -0,0 +1,310 @@
...
+  ],
```
lmolkova: should this be updated? we should set this 1. after span has started 2. input messages only when content is enabled. Even though we can (and should) drop inputMessages afterwards inside util, it's wasteful to populate it without the check

packages/genai-util/README.md
```diff
@@ -0,0 +1,310 @@
...
+    outputTokens: response.usage.completion_tokens,
+    totalTokens: response.usage.total_tokens,
+  });
+  invocation.addOutputMessages([
+    {
+      role: 'assistant',
+      parts: [{ type: 'text', content: response.content }],
+      finish_reason: 'stop',
+    },
+  ]);
```
lmolkova: please check if content is enabled before this

packages/genai-util/README.md
```diff
@@ -0,0 +1,310 @@
...
+---
+
+### 3. Agent Invocations (Local & Remote)
+
+The library provides dedicated support for Agent / Assistant spans:
+
+```ts
+// Local in-process agent (SpanKind.INTERNAL)
+const localAgent = handler.startLocalAgent({
+  agentId: 'research-agent-01',
```
lmolkova:
```suggestion
```

(also agent verion)

also conversation id is not start-time attribute

packages/genai-util/README.md
```diff
@@ -0,0 +1,310 @@
...
+---
+
+### 4. Tool Execution Spans
+
+```ts
+const toolInvocation = handler.startTool({
+  toolName: 'web_search',
+  toolType: 'function',
+  toolDescription: 'Searches the web for latest info',
+  toolCallId: 'call_abc123',
```
lmolkova: please update

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,473 @@
...
+}
+
+// ============================================================================
+// Invocation Options
+// ============================================================================
+
+/**
+ * Options for starting an inference invocation.
+ */
+export interface InferenceInvocationOptions {
```
lmolkova, line 348:
```suggestion
export interface InferenceStartInvocationOptions {
```

maybe?

packages/genai-util/src/utils.ts
psx95: I looked into it and addressed it in [de577a9](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/de577a923529f1424a0ea693f5b11f4f8e72af5f). 

The encoding to string was necessary to support `BlobPart`.

packages/genai-util/src/types.ts
psx95: I referenced the [Python constructor for ToolInvocation](https://github.com/open-telemetry/opentelemetry-python-genai/blob/b301b3227c1bf921c464b2d25affce22f4165e1b/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_tool_invocation.py#L33). 

I noticed  that `toolArgument` was opt-in, so I will remove that, but `toolDescription` and `toolCallId` were passed in Python.

packages/genai-util/README.md
psx95: I did not edit the README from the original PR, but I notice this is causing a lot of confusion. Since this PR does not add any usage and these classes don't exist, I'm going to remove all the usage examples for now - I think that makes the review easier as well.

packages/genai-util/src/types.ts
psx95: I think for now, `InferenceInvocationOptions` is ok. `InferenceStartInvocationOptions` indicate that there is going to be another type with options that encapsulates options that are not start-time attributes and I'm not sure if we would need to encapsulate them in a type.

IIUC, we could simply use setters to allow instrumentation authors to set them one-by-one at different times. If a need does arise to encapsulate them, we can do this renaming then.

packages/genai-util/README.md
psx95: Please see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709#discussion_r3919270896.

packages/genai-util/README.md
psx95: Please see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709#discussion_r3919270896

packages/genai-util/README.md
psx95: Please see https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709#discussion_r3919270896

packages/genai-util/src/utils.ts
psx95: Yes, for now it is, but I imagine we'd introduce a setter later in `inference.ts` - something like 
```typescript
public setRequestOptions(requestOptions: GenAIRequestOptions) {
 ...
}
```
where this could be used. I could remove it, but its gonna be used eventually in subsequent PRs given that inference invocation is something we plan on supporting initially.

packages/genai-util/src/utils.ts
psx95: 1. That makes sense to me, thanks for the explanation I will remove the if check for type = string to directly fallback to `_OTHER`. 

2. I can make this change when adding the base invocation class. That way would be able to add better tests for it as well.

packages/genai-util/src/types.ts
lmolkova: Given that there is no example of how you envision this to be used, I might be missing your point completely. What would the actual usage look like?
If we don't clearly distinguish start-time options from post-start options in the types, it creates a confusing API:
- If `InferenceInvocationOptions` is reused later, do you pass the same object twice - once to start, and then what?
- What happens if someone passes a new object later with different start parameters (e.g. changing `providerName` or `requestModel`)?
Separating them (e.g. `InferenceStartInvocationOptions` for start-time only, and dedicated setters or a separate options type for post-start parameters) makes the lifecycle unambiguous and prevents confusing states.

packages/genai-util/src/types.ts
lmolkova: Good catch! Fixing it in https://github.com/open-telemetry/opentelemetry-python-genai/pull/577

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Environment variable controlling capture of message content (prompt/completion).
+ */
+export const ENV_GENAI_CAPTURE_MESSAGE_CONTENT =
+  'OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT' as const;
+
+/**
+ * Environment variable specifying custom completion hook module.
+ */
+export const ENV_GENAI_COMPLETION_HOOK =
+  'OTEL_INSTRUMENTATION_GENAI_COMPLETION_HOOK' as const;
```
aabmass: Doesn't look like it's used in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677, can we remove?

packages/genai-util/src/environment-variables.ts
aabmass: Also would remove ContentCaptureMode since it all collapses to string, and remove the `?` since it looks like a param is always passed in the final PR.

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (typeof value === 'boolean') {
+    return value ? 'span_only' : 'none';
+  }
+
+  const strVal = String(value).trim().toLowerCase();
+  switch (strVal) {
+    case 'true':
+    case '1':
```
aabmass: Can we remove parsing for some of these special magic values? I think it's tech debt in the python repo at this point.

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Resolve effective ContentCaptureMode based on explicit config and environment variable.
+ *
+ * @experimental This function is experimental and subject to change.
+ */
+export function getContentCaptureMode(
+  configVal?: boolean | ContentCaptureMode,
+  envVarName: string = ENV_GENAI_CAPTURE_MESSAGE_CONTENT
+): ContentCaptureMode {
+  const envVal = process.env[envVarName];
```
aabmass: Since we're already depending on `@opentelemetry/core`, can use [`getStringFromEnv`](https://github.com/open-telemetry/opentelemetry-js/blob/52a9a5749776a6dd7bc7eee5766542b7bdb9ec76/packages/opentelemetry-core/src/platform/node/environment.ts#L41) helper which is cross platform ?

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (envVal !== undefined && envVal !== '') {
+    return parseContentCaptureMode(envVal);
+  }
+  return parseContentCaptureMode(configVal);
```
aabmass, line 57: nit could push this edge cases into parseContentCaptureMode (especially since it's already handling undefined)

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+import type {
+  GenAIRequestOptions,
+  InputMessages,
+  SystemInstructions,
+} from './types';
+
+const SERVER_PORT_FROM_URL_PROTOCOL: Record<string, number> = {
+  'https:': 443,
+  'http:': 80,
+};
```
aabmass:
```suggestion
const SERVER_PORT_FROM_URL_PROTOCOL = {
  'https:': 443,
  'http:': 80,
} as const;
```

generally prefer inference for typescript, it can infer the exact type. in this case it would force you to check that keys to this object are valid ([example](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAygUQEoDVkH0AKB5JAVdAMSWwFl0BVJAGSxL2wGFtqYBeGAbwCgYYByABZQoABwgAufhJgAWWQGYANLwHCxUmQA4ADCoC+MAIYQYoSFADc3bolQYc+IiXJVamekxYBtfgDMQEH4AXSA))

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+/**
+ * Extract `server.address` and `server.port` attributes from a client baseURL.
+ *
+ * @param baseURL - Base URL of the API client (e.g. `'https://api.openai.com/v1'`).
+ * @param diag_ - Optional diagnostic logger for debug logging.
+ * @returns Attributes object containing `server.address` and `server.port`, or `undefined` if baseURL is not provided or invalid.
+ */
+export function getAttrsFromBaseURL(
+  baseURL: string | undefined,
+  diag_: DiagLogger = diag
```
aabmass: i actually have no clue, is this the usual method or do most instrumentations just read the global diag?

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+};
+
+/**
+ * Extract `server.address` and `server.port` attributes from a client baseURL.
+ *
+ * @param baseURL - Base URL of the API client (e.g. `'https://api.openai.com/v1'`).
+ * @param diag_ - Optional diagnostic logger for debug logging.
+ * @returns Attributes object containing `server.address` and `server.port`, or `undefined` if baseURL is not provided or invalid.
+ */
+export function getAttrsFromBaseURL(
```
aabmass: this seems to be dead code in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+export function serializeContent(content: unknown): string {
+  if (typeof content === 'string') {
+    return content;
+  }
+  if (content === null || content === undefined) {
+    return '';
+  }
+
+  // Handle types we know JSON.stringify drops (returns undefined for)
+  if (typeof content === 'function' || typeof content === 'symbol') {
```
aabmass: this would serialize the entire function definition as a string, which seems like overkill

```
Welcome to Node.js v26.8.1.
> function foo() {
| // my cool func
| console.log("yo");
| }
undefined
> String(foo)
'function foo() {\r// my cool func\rconsole.log("yo");\r}'
>
```

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+  }
+  let binary = '';
+  for (let i = 0; i < bytes.byteLength; i++) {
+    binary += String.fromCharCode(bytes[i]);
+  }
+  return btoa(binary);
+}
+
+/**
+ * Format input messages into a JSON string for span attribute storage.
```
aabmass, line 111: Can you add a note that this only done because JS doesn't support complex attributes (maybe linking to the issue)

https://github.com/open-telemetry/semantic-conventions-genai/blob/fee465db333bdd6a7d2faa320edab5cf3101a4f4/docs/gen-ai/gen-ai-spans.md?plain=1#L1302-L1310

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+ * @param messages - Output messages payload to format.
+ * @returns JSON string representation of output messages, or `undefined` if empty/invalid.
+ */
+export function formatOutputMessages(messages: unknown): string | undefined {
+  if (!messages) {
+    return undefined;
+  }
+  if (typeof messages === 'string') {
+    return messages;
+  }
```
aabmass: is this special case needd?

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+ *
+ * @param error - The caught error, exception, or value.
+ * @returns The standardized error type string.
+ */
+export function getErrorType(error: unknown): string {
+  if (error instanceof Error) {
+    const code = (error as { code?: string | number }).code;
+    if (code !== undefined && code !== null && String(code).trim().length > 0) {
+      return String(code);
+    }
```
aabmass: I think `"code" in error` type narrowing is more idiomatic and doesn't require any `as` assertions

packages/genai-util/src/utils.ts
aabmass: You mean it's used in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677?

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,484 @@
...
+ *
+ * Rationale:
+ * 1. Per OpenTelemetry JavaScript Contrib repository guidelines (GUIDELINES.md),
+ *    packages must not depend on `@opentelemetry/semantic-conventions/incubating`
+ *    because incubating semantic conventions are experimental/unstable and subject
+ *    to breaking changes across minor releases.
+ * 2. Defining attributes, metric names, and event names as local `as const` string
+ *    literals isolates this utility package and all downstream GenAI instrumentations
+ *    from upstream breaking changes and prevents version-conflict / diamond-dependency
+ *    issues across packages in the ecosystem.
```
aabmass: I don't see anything like that in [GUIDELINES.md](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv) but I do see [this doc](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv). FWIW, npm allows multiple versions if there are conflicts (I'm guessing maintainers have good reason not to rely on that mechanism). can we just remove point 2?

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,484 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * Semantic conventions in this file are defined locally rather than imported from
+ * `@opentelemetry/semantic-conventions/incubating`.
```
aabmass, line 8: Can you add some note of which version of https://github.com/open-telemetry/semantic-conventions-genai this was "generated" from? It will be easier to keep it updated in the future.

I also found this script, but I don't think it would work for us anyway since the main semconv package doesn't generate from the semconv-genai repo. https://github.com/open-telemetry/opentelemetry-js-contrib/blob/20bbc8c8a9827d7add9e86aa734d8601932736fc/scripts/gen-semconv-ts.js#L7-L21

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,484 @@
...
+ * 1. Per OpenTelemetry JavaScript Contrib repository guidelines (GUIDELINES.md),
+ *    packages must not depend on `@opentelemetry/semantic-conventions/incubating`
+ *    because incubating semantic conventions are experimental/unstable and subject
+ *    to breaking changes across minor releases.
+ * 2. Defining attributes, metric names, and event names as local `as const` string
+ *    literals isolates this utility package and all downstream GenAI instrumentations
+ *    from upstream breaking changes and prevents version-conflict / diamond-dependency
+ *    issues across packages in the ecosystem.
+ *
+ * @see https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md
```
aabmass:
```suggestion
```

wrong repo

aabmass (approved):

packages/genai-util/src/utils.ts
psx95: Yes, but I'm using #3677 purely as a reference - I don't think that PR will be merged. It is meant to show what the final state of this util package *might* look like.

packages/genai-util/src/semconv.ts
psx95: FYI: I was made aware of this - https://github.com/open-telemetry/opentelemetry-js/pull/7014 which could potentially affect this file a lot. 

I can remove point 2 from the documentation.

packages/genai-util/src/environment-variables.ts
psx95: Yeah, I will remove the support for boolean here. 

Also filed [this issue](https://github.com/open-telemetry/semantic-conventions-genai/issues/497) to formally describe the behavior and supported values for content capture mode.

packages/genai-util/src/types.ts
psx95: I addressed this in [d478eac](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/d478eac6e3d3e186f2154087c8f6b32f697ddcf5).

packages/genai-util/src/environment-variables.ts
psx95: Addressed this in [b7b8f5f](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/b7b8f5f72f7a2606d32fd36d223f457ddb66d33e).

packages/genai-util/src/environment-variables.ts
psx95: Yeah, I had planned to add a hook similar to the upload hook in python, but since the code became too big there already, decided against it. 

I can remove it for now.

packages/genai-util/src/environment-variables.ts
psx95: I think this is now addressed in [b7b8f5f](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/b7b8f5f72f7a2606d32fd36d223f457ddb66d33e).  Feel free to re-open in case there's any other value for which we want to drop support.

packages/genai-util/src/environment-variables.ts
psx95: Checking `envVal !== undefined` verifies that the environment variable was not set and so its ok to fetch the value from explicit config. I can add some documentation to make this behavior clear.

packages/genai-util/src/environment-variables.ts
psx95: Removed in [a47bedb](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/a47bedb308ada98acb7b1ad026e1032910b18bab).

packages/genai-util/src/environment-variables.ts
psx95: Addressed in [82c1ca3](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/82c1ca324390eec5cfcb23e552e7392df527f6f7).

packages/genai-util/src/environment-variables.ts
psx95: Please see the change in [82c1ca3](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/82c1ca324390eec5cfcb23e552e7392df527f6f7). Feel free to re-open in case you think this needs further resolution.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,295 @@
...
+ * @returns JSON or plain string representation of system instructions, or `undefined` if empty/invalid.
+ */
+export function formatSystemInstructions(
+  instructions?: SystemInstructions | string
+): string | undefined {
+  if (!instructions) {
+    return undefined;
+  }
+  if (typeof instructions === 'string') {
+    return instructions;
```
JacksonWeber, line 214: Should we normalize strings to the semconv shape here? `gen_ai.system_instructions` expects an array of message parts, so returning the string directly would produce invalid telemetry.

packages/genai-util/src/semconv.ts
psx95: I did not use this script to generate the semconv used here. 

Can you explain what you men by "version" here? I might be missing something here. I see no releases or specific versioned branches in https://github.com/open-telemetry/semantic-conventions-genai. The semconv used here were just using latest from head of the repo. 

FYI: This file could change once https://github.com/open-telemetry/opentelemetry-js/pull/7014 goes through.

packages/genai-util/src/utils.ts
psx95: Took a pass at normalizing in [40f8c35](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/40f8c35892e3a55b5f7b8451a3a28530225fc592). 

If the passed argument is a string, the function attempts to check if can be parsed as a JSON object, if not, then   the entire string is returned as: 

```typescript
return [{ type: 'text', content: instructions }];
```

Let me know if you think the behavior should be altered.

JacksonWeber: Looks good to me, thanks!

packages/genai-util/src/semconv.ts
aabmass: Huh good point https://github.com/open-telemetry/semantic-conventions-genai#schema-url I guess we never released. Adding a commit SHA for semconv-genai repo that your agent read would be good.

At least before we migrated out of the main semantic-conventions repo, there were tagged releases. @lmolkova WDYT?

packages/genai-util/src/utils.ts
psx95: Agreed. I addressed this in [831d194](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/831d19419e77c19f0e7497423fed15798691f9c9).

packages/genai-util/src/types.ts
psx95: I don't think that `InferenceInvocationOptions` would be 'reused' later. This is meant to be passed in the constructor of `InferenceInvocation` (See [reference](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677/changes#diff-49519c3dfa6e862e94cec2371a165a1f9351b9d4a1905522e08d42d0a573b63cR77)).  
The invocations (e.g., `InferenceInvocation` here) would represent a single LLM operation. So it would not be re-used.

This constructor itself can be created using code like: 
```typescript
// handler.startInference takes in start time options encapsulated as `InferenceInvocationOptions`. 
const invocation = instrumentation._handler.startInference({
          providerName: GEN_AI_PROVIDER_NAME_VALUE_ANTHROPIC,
          operationName: GEN_AI_OPERATION_NAME_VALUE_CHAT,
          requestModel: params?.model,
          // any other start-time parameters
        });
``` 

Any other **non**-start time options can be set on the invocation itself using setters 

```typescript
// Just an example
invocation.setToolCallId(callId);
```
A rough sample on how genai-utils library might be used can be seen in [this PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3698/changes#diff-770b4e30e8fadc1b0376aaa798077f956b0c1dad6e1830e34ead919e25b2be7cR90). 

Let me know if this makes sense.

packages/genai-util/src/utils.ts
psx95: FYI: I think this code will be used in the upcoming PRs so I am keeping it for now. We can remove it if its not used later. Feel free to comment if this is not preferred for now.

packages/genai-util/src/types.ts
lmolkova: thanks for the explanation, this makes sense! I still think that `InferenceStartInvocationOptions` or some other name that reflects it's 'start' options would be more expressive, but that's a matter of taste and not substantical.

packages/genai-util/src/types.ts
```diff
@@ -0,0 +1,463 @@
...
+ */
+export interface GenAIInstrumentationConfig {
+  /**
+   * Controls capturing of prompt, completion, and tool call contents.
+   * - `false` or `'none'`: no content captured (default).
+   * - `true` or `'span_only'`: content captured on span attributes.
+   * - `'event_only'`: content captured in span events.
+   * - `'span_and_event'`: content captured in both spans and events.
+   */
+  captureMessageContent?: boolean | ContentCaptureMode;
```
lmolkova: Since boolean support was dropped from `parseContentCaptureMode` and `getContentCaptureMode`, should we drop `boolean` here as well?

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  let u: URL;
+  try {
+    u = new URL(baseURL);
+  } catch (ex) {
+    diag_.debug(
+      `could not determine server.address/server.port from baseURL: ${ex}`
+    );
+    return undefined;
+  }
```
lmolkova: Per OpenTelemetry semantic conventions, `server.port` should not be set when it is the default port for the protocol (80 for http, 443 for https).

Suggest omitting `server.port` unless a non-default port is explicitly provided in `baseURL`.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,370 @@
...
+ *
+ * @param messages - Output messages payload to format.
+ * @returns JSON string representation of output messages, or `undefined` if empty/invalid.
+ */
+export function formatOutputMessages(messages: unknown): string | undefined {
+  if (!messages) {
+    return undefined;
+  }
+  if (typeof messages === 'string') {
+    return messages;
```
lmolkova: If `messages` is a plain string, returning it directly produces a raw string rather than a serialized JSON array conforming to the `gen_ai.output.messages` schema.

Suggest normalizing string completions into `OutputMessage[]` (similar to `formatSystemInstructions`) and tightening the parameter type.

packages/genai-util/src/environment-variables.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * Returns true if content should be attached to span attributes.
+ *
+ * @experimental This function is experimental and subject to change.
+ */
+export function isSpanContentCaptureEnabled(mode: ContentCaptureMode): boolean {
```
lmolkova: we're dropping it from python in favor of `isContentCaptureEnabled()` - https://github.com/open-telemetry/opentelemetry-python-genai/pull/593

lmolkova (approved):
LGTM, just a few minor comments!

packages/genai-util/src/types.ts
psx95: Right. This should be removed. Addressed in [beb224c](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/beb224c00ee3acc31f83b3e7199f09e0e2233a0c).

packages/genai-util/src/semconv.ts
psx95: Addressed in [8b92607](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/8b92607967c88b2fb898724764dfc6b702e549d2).

packages/genai-util/src/semconv.ts
psx95: Addressed in [8b92607](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/8b92607967c88b2fb898724764dfc6b702e549d2).

packages/genai-util/src/utils.ts
psx95: Yeah - seems it was removed in a later refactor. I will remove it from here

packages/genai-util/src/utils.ts
psx95: Addressed in [4a737a9](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/4a737a992012e5a1b1f604608067a94222c7c9e9).

packages/genai-util/src/utils.ts
psx95: Addressed in [4a737a9](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/4a737a992012e5a1b1f604608067a94222c7c9e9).

packages/genai-util/src/utils.ts
psx95: Hmm, yes it would, but then again this is a general purpose utility function - if people using it want to serialize functions, then they can.

Are recommending to intentionally return "[Unserializable Content]" here?

packages/genai-util/src/utils.ts
psx95: I did not find a tracking issue in JS for this, but I added a note.

psx95: I couldn't find an existing issue tracking this in the JS SDK so opened one- https://github.com/open-telemetry/opentelemetry-js/issues/7071

packages/genai-util/src/utils.ts
psx95: I think we can just remove this special case.

packages/genai-util/src/utils.ts
psx95: Addressed in [c5af032](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/c5af032c4f73ad3e3b51c9316702976f111e938a).

packages/genai-util/src/utils.ts
psx95: Addressed in [c5af032](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/c5af032c4f73ad3e3b51c9316702976f111e938a).

packages/genai-util/src/utils.ts
psx95: This is removed now since it was a leftover part of some dead code.

packages/genai-util/src/utils.ts
psx95: The special handling of messages as a string is not required so I removed it. 

For the second part,  limiting to the input type `OutputMessage` is indeed a good idea, but I think normalization won't be required since the potential `BlobPart` of the `OutputMessage` is already handled by the special handling of `Uint8Array` type.

Addressed in [c5cda91](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/c5cda9135b0be618020a90aba34d43630d072b04).

psx95: Addressed in [c5af032](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/c5af032c4f73ad3e3b51c9316702976f111e938a) and [c5cda91](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/c5cda9135b0be618020a90aba34d43630d072b04).

packages/genai-util/src/environment-variables.ts
psx95: I think I can remove this function for now if its being dropped.  Addressed in [163568b](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/163568beca59b8e32ec9e035360e5d1b2b2aa3b8).

psx95:
I believe all comments on this PR are addressed, except the one about adding information about the semconv version used in `semconv.ts`. This is because the semconv package for genai did not have an official release (see [reference](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709#discussion_r3929685976)). IMO, its a non-blocker for merging this PR.

packages/genai-util/src/utils.ts
psx95: Explicitly avoid serializing function in [3ee1733](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/3ee17332e4288b4ba2367f5611569e832538d43f).

aabmass (approved):

packages/genai-util/src/semconv.ts
psx95: Added the commit hash at which the genai semconv repo was referenced to generate the `semconv.ts` file in 
[1df759b](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/1df759bcf84bf66487dc390973035eb53d6fae53).

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,327 @@
...
+ * @returns JSON string representation of input messages, or `undefined` if empty/invalid.
+ */
+export function formatInputMessages(
+  messages?: InputMessages
+): string | undefined {
+  if (!messages) {
+    return undefined;
+  }
+  try {
+    return JSON.stringify(messages, (_key, value) => {
```
JacksonWeber: Could we handle `Buffer` here too? `JSON.stringify` calls `Buffer.toJSON()` before the replacer, so it never reaches the `Uint8Array` check and gets serialized as `{ type: "Buffer", data: [...] }` instead of base64. The same issue affects `formatOutputMessages`.

psx95: I am not familiar with the internal implementation of `JSON.stringify` but the [unit test shows that it would work for blobs](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/changes#diff-aee31b0d2198b9d928f7987da79ed4b2a8f0a69c7f95377bb87837b719dc7a3bR88) (you might have to scroll to/search for `formatInputMessages` in the file).

Could you suggest a test case which you think would fail for the current implementation? Maybe that will help me understand the concern better.

JacksonWeber: The current test uses a `Uint8Array`. A `Buffer` extends `Uint8Array`, so it is also valid for `BlobPart.content`. Try changing it to a `Buffer` and keeping the same expected base64 string:
```
content: Buffer.from([72, 101, 108, 108, 111])
```
That should fail because `JSON.stringify` calls `Buffer.toJSON()` before our replacer runs. By then, the buffer has become `{ type: "Buffer", data: [...] }`, so it misses the `Uint8Array` check.

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,483 @@
...
+ *
+ * @example gpt-4o-2024-08-06
+ */
+export const ATTR_GEN_AI_RESPONSE_MODEL = 'gen_ai.response.model' as const;
+
+/**
+ * Array of reasons why the model stopped generating tokens for each response choice.
+ *
+ * @example ["stop"]
+ * @example ["tool_calls"]
```
hectorhdzg: The pinned GenAI schema and the `FinishReason` type in this package use the singular `tool_call` value.

```suggestion
 * @example ["tool_call"]

File: knip.jsonc [hunk 45]
```diff
@@ -6,6 +6,14 @@
       // Ensure register.ts is considered an entry-point (the `!` is to be
       // included with the `--production` flag).
       "entry": "src/register.ts!"
+    },
+    "packages/genai-util": {
+      // Temporary ignore during foundation phase before handlers and invocations are added in follow-up PRs.
+      // REMOVE this configuration once the exports and handlers are added to packages/genai-util/src/index.ts.
```
hectorhdzg:
```suggestion
      // TODO: Remove this configuration once exports and handlers are added to packages/genai-util/src/index.ts.
```

hectorhdzg (approved):
LGTM, just minor comments from my side

packages/genai-util/src/utils.ts
psx95: Addressed in [f629202](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/f629202a04d2bb102f2fae7fdc1bf2b2be1962ed). I added a test case to prevent regression.

Feel free to re-open in case there are any other concerns regarding the implementation.

packages/genai-util/src/utils.ts
```diff
@@ -0,0 +1,347 @@
...
+function serializeMessageReplacer(
+  this: Record<string, unknown>,
+  key: string,
+  value: unknown
+): unknown {
+  const rawValue = this[key];
+
+  if (
+    rawValue instanceof Uint8Array ||
+    (typeof Buffer !== 'undefined' && Buffer.isBuffer(rawValue))
```
aabmass: The second condition would be dead code since Buffer is a subclass?
```suggestion
    rawValue instanceof Uint8Array
```

aabmass (approved):

JacksonWeber (approved):
LGTM, thanks for addressing comments.

packages/genai-util/src/utils.ts
psx95: Yes, verified that this is not needed. The test passed without this OR condition. I addressed this in 
[4eb05ba](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/4eb05ba725ed7ac72da9ac3c8ec964c07e0ac2aa). 

Thanks.

knip.jsonc
psx95: Addressed in [4eb05ba](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/4eb05ba725ed7ac72da9ac3c8ec964c07e0ac2aa).

packages/genai-util/src/semconv.ts
psx95: Addressed in [4eb05ba](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3709/commits/4eb05ba725ed7ac72da9ac3c8ec964c07e0ac2aa).

psx95:
Thank you all for the reviews! Much appreciated! 
I think this PR can now be merged! 👍🏻

JacksonWeber (approved):

pichlermarc (approved):

PR #3712, opened 2026-08-29
feat(genai-util): add GenAI metric helpers and histograms

psx95:

Re: #3681

Generative AI instrumentations need to record standardized OpenTelemetry metrics such as operation duration, token usage (input and output tokens), and time-to-first-chunk/time-to-first-token with explicit bucket boundaries defined in the GenAI semantic conventions.
This PR introduces shared metric creation and recording helper functions for `@opentelemetry/genai-util`

- **Standard Histogram Creation**: Added helper functions with standard explicit bucket boundaries:
  - `createDurationHistogram` (`gen_ai.client.operation.duration`)
  - `createTokenUsageHistogram` (`gen_ai.client.token.usage`)
  - `createTimeToFirstChunkHistogram` (`gen_ai.client.operation.time_to_first_chunk`)
  - `createTimePerOutputChunkHistogram` (`gen_ai.client.operation.time_per_output_chunk`)
- **Unit Tests**: Added test suite covering bucket boundaries, histogram creation, and metric recording behavior with valid and edge-case inputs (`test/metrics.test.ts`).

## Important Info
 - This is the **second PR** in a series that attempt to breakdown PR #3677. The files in this PR are taken from PR #3677 and did not require any modifications for the package to compile. 
 - Based off:  #3709

### DIFF
--- a/packages/genai-util/src/metrics.ts
+++ b/packages/genai-util/src/metrics.ts
@@ -1,0 +1,244 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Attributes, Histogram, Meter } from '@opentelemetry/api';
+import {
+  ATTR_GEN_AI_TOKEN_TYPE,
+  GEN_AI_TOKEN_TYPE_VALUE_INPUT,
+  GEN_AI_TOKEN_TYPE_VALUE_OUTPUT,
+  METRIC_GEN_AI_CLIENT_OPERATION_DURATION,
+  METRIC_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK,
+  METRIC_GEN_AI_CLIENT_TOKEN_USAGE,
+  METRIC_GEN_AI_SERVER_TIME_TO_FIRST_TOKEN,
+} from './semconv';
+import type { TokenUsage } from './types';
+
+/**
+ * Standard explicit bucket boundaries for GenAI operation duration (in seconds).
+ *
+ * @experimental
+ */
+export const GENAI_OPERATION_DURATION_BUCKETS = [
+  0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12, 10.24, 20.48,
+  40.96, 81.92,
+];
+
+/**
+ * Standard explicit bucket boundaries for GenAI token usage.
+ *
+ * @experimental
+ */
+export const GENAI_TOKEN_USAGE_BUCKETS = [
+  1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304,
+  16777216, 67108864,
+];
+
+/**
+ * Standard explicit bucket boundaries for GenAI client time to first chunk (in seconds).
+ *
+ * @experimental
+ */
+export const GENAI_TIME_TO_FIRST_CHUNK_BUCKETS = [
+  0.001, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30,
+];
+
+/**
+ * Standard explicit bucket boundaries for GenAI time to first token (in seconds).
+ *
+ * Adopted from vLLM: https://github.com/vllm-project/vllm/blob/main/vllm/v1/metrics/loggers.py
+ *
+ * @experimental
+ */
+export const GENAI_SERVER_TIME_TO_FIRST_TOKEN_BUCKETS = [
+  0.001, 0.005, 0.01, 0.02, 0.04, 0.06, 0.08, 0.1, 0.25, 0.5, 0.75, 1.0, 1.25,
+  1.5, 1.75, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 6.0, 6.5, 7.0, 7.5, 8.0,
+  8.5, 9.0, 9.5, 10.0,
+];
+
+/**
+ * Options for creating GenAI metric instruments.
+ *
+ * @experimental
+ */
+export interface MetricCreationOptions {
+  /** Metric instrument name override. */
+  name?: string;
+  /** Metric description override. */
+  description?: string;
+  /** Metric unit override. */
+  unit?: string;
+}
+
+/**
+ * Create standard `gen_ai.client.operation.duration` histogram.
+ *
+ * @experimental
+ */
+export function createDurationHistogram(
+  meter: Meter,
+  options?: MetricCreationOptions
+): Histogram {
+  return meter.createHistogram(
+    options?.name ?? METRIC_GEN_AI_CLIENT_OPERATION_DURATION,
+    {
+      description: options?.description ?? 'GenAI operation duration',
+      unit: options?.unit ?? 's',
+      advice: {
+        explicitBucketBoundaries: GENAI_OPERATION_DURATION_BUCKETS,
+      },
+    }
+  );
+}
+
+/**
+ * Create standard `gen_ai.client.token.usage` histogram.
+ *
+ * @experimental
+ */
+export function createTokenUsageHistogram(
+  meter: Meter,
+  options?: MetricCreationOptions
+): Histogram {
+  return meter.createHistogram(
+    options?.name ?? METRIC_GEN_AI_CLIENT_TOKEN_USAGE,
+    {
+      description:
+        options?.description ??
+        'Measures number of input and output tokens used',
+      unit: options?.unit ?? '{token}',
+      advice: {
+        explicitBucketBoundaries: GENAI_TOKEN_USAGE_BUCKETS,
+      },
+    }
+  );
+}
+
+/**
+ * Create standard `gen_ai.client.operation.time_to_first_chunk` histogram.
+ *
+ * @experimental
+ */
+export function createTimeToFirstChunkHistogram(
+  meter: Meter,
+  options?: MetricCreationOptions
+): Histogram {
+  return meter.createHistogram(
+    options?.name ?? METRIC_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK,
+    {
+      description:
+        options?.description ??
+        'Time to first chunk for streaming response in seconds',
+      unit: options?.unit ?? 's',
+      advice: {
+        explicitBucketBoundaries: GENAI_TIME_TO_FIRST_CHUNK_BUCKETS,
+      },
+    }
+  );
+}
+
+/**
+ * Create standard `gen_ai.server.time_to_first_token` histogram.
+ *
+ * @experimental
+ */
+export function createServerTimeToFirstTokenHistogram(
+  meter: Meter,
+  options?: MetricCreationOptions
+): Histogram {
+  return meter.createHistogram(
+    options?.name ?? METRIC_GEN_AI_SERVER_TIME_TO_FIRST_TOKEN,
+    {
+      description:
+        options?.description ??
+        'Time to first token for streaming response in seconds',
+      unit: options?.unit ?? 's',
+      advice: {
+        explicitBucketBoundaries: GENAI_SERVER_TIME_TO_FIRST_TOKEN_BUCKETS,
+      },
+    }
+  );
+}
+
+/**
+ * Internal helper to validate and record a duration value in seconds.
+ */
+function _recordDuration(
+  histogram: Histogram | undefined,
+  durationSeconds: number,
+  attributes?: Attributes
+): void {
+  if (!histogram || durationSeconds < 0 || !isFinite(durationSeconds)) {
+    return;
+  }
+  histogram.record(durationSeconds, attributes);
+}
+
+/**
+ * Record operation duration metric.
+ *
+ * @experimental
+ */
+export function recordOperationDuration(
+  histogram: Histogram | undefined,
+  durationSeconds: number,
+  attributes?: Attributes
+): void {
+  _recordDuration(histogram, durationSeconds, attributes);
+}
+
+/**
+ * Record token usage metrics (both input and output tokens if present).
+ *
+ * @experimental
+ */
+export function recordTokenUsage(
+  histogram: Histogram | undefined,
+  usage: TokenUsage | undefined,
+  attributes?: Attributes
+): void {
+  if (!histogram || !usage) {
+    return;
+  }
+
+  if (typeof usage.inputTokens === 'number' && usage.inputTokens >= 0) {
+    histogram.record(usage.inputTokens, {
+      ...attributes,
+      [ATTR_GEN_AI_TOKEN_TYPE]: GEN_AI_TOKEN_TYPE_VALUE_INPUT,
+    });
+  }
+
+  if (typeof usage.outputTokens === 'number' && usage.outputTokens >= 0) {
+    histogram.record(usage.outputTokens, {
+      ...attributes,
+      [ATTR_GEN_AI_TOKEN_TYPE]: GEN_AI_TOKEN_TYPE_VALUE_OUTPUT,
+    });
+  }
+}
+
+/**
+ * Record time to first chunk metric.
+ *
+ * @experimental
+ */
+export function recordTimeToFirstChunk(
+  histogram: Histogram | undefined,
+  durationSeconds: number,
+  attributes?: Attributes
+): void {
+  _recordDuration(histogram, durationSeconds, attributes);
+}
+
+/**
+ * Record server time to first token metric.
+ *
+ * @experimental
+ */
+export function recordServerTimeToFirstToken(
+  histogram: Histogram | undefined,
+  durationSeconds: number,
+  attributes?: Attributes
+): void {
+  _recordDuration(histogram, durationSeconds, attributes);
+}
--- a/packages/genai-util/src/metrics.ts
+++ b/packages/genai-util/src/metrics.ts
@@ -25,7 +25,7 @@ import type { TokenUsage } from './types';
  *
  * @experimental
  */
-export const GENAI_OPERATION_DURATION_BUCKETS = [
+const GENAI_OPERATION_DURATION_BUCKETS = [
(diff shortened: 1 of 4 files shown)


packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * @experimental
+ */
+export interface MetricCreationOptions {
+  /** Metric instrument name override. */
+  name?: string;
+  /** Metric description override. */
+  description?: string;
+  /** Metric unit override. */
+  unit?: string;
```
lmolkova: why not histogram boundaries too ?

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * Create standard `gen_ai.client.operation.duration` histogram.
+ *
+ * @experimental
+ */
+export function createDurationHistogram(
```
lmolkova: this won't survive multiple metrics and the fallback to METRIC_GEN_AI_CLIENT_OPERATION_DURATION does not seem right. we should rather have a helper per individual metric and we should generate them, otherwise we'll spend all the time keeping this up to date. IF you want to have a hand-written version, consider writing in the way that it'd be generated later (boilerplace and repetitiveness is better than ad-hoc logic)

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,244 @@
...
+    });
+  }
+}
+
+/**
+ * Record time to first chunk metric.
+ *
+ * @experimental
+ */
+export function recordTimeToFirstChunk(
```
lmolkova: not sure if this method helps with anything, why expose it instead of exposing `_recordDuration` directly?

packages/genai-util/src/metrics.ts
psx95: IIUC, you're recommending to bring this more in-line with [Python's implementation](https://github.com/open-telemetry/opentelemetry-python-genai/blob/59e6efe6c0be45a9da78f8f3c21c5d6ac4488ed1/util/opentelemetry-util-genai/src/opentelemetry/util/genai/instruments.py#L42), where we don't allow users to pass optional parameters to the functions. 

I did it this way since I thought - it offered more flexibility, but I think we can be opinionated here - so I will change it to only accept `meter` as the argument. 

Could you explain what you meant by: 
> this won't survive multiple metrics

packages/genai-util/src/metrics.ts
psx95: This was just syntactic sugar in a way, but I can remove it.

packages/genai-util/src/metrics.ts
psx95: I think this will be moot now since we moved to opinionated methods where the users are not allowed to configure the histogram being created.

We can remove this type.

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,167 @@
...
+
+import type { Attributes, Histogram, Meter } from '@opentelemetry/api';
+import {
+  ATTR_GEN_AI_TOKEN_TYPE,
+  GEN_AI_TOKEN_TYPE_VALUE_INPUT,
+  GEN_AI_TOKEN_TYPE_VALUE_OUTPUT,
+  METRIC_GEN_AI_CLIENT_OPERATION_DURATION,
+  METRIC_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK,
+  METRIC_GEN_AI_CLIENT_TOKEN_USAGE,
+  METRIC_GEN_AI_SERVER_TIME_TO_FIRST_TOKEN,
```
lmolkova: there is no need for this metric here - it's for inference servers, we can always add it later if someone will need it

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,167 @@
...
+  16777216, 67108864,
+];
+
+/**
+ * Standard explicit bucket boundaries for GenAI client time to first chunk (in seconds).
+ *
+ * @experimental
+ */
+export const GENAI_TIME_TO_FIRST_CHUNK_BUCKETS = [
+  0.001, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30,
```
lmolkova: Per semconv, `gen_ai.client.operation.time_to_first_chunk` should use the same explicit bucket boundaries as operation duration: `[0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, 2.56, 5.12, 10.24, 20.48, 40.96, 81.92]`.

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,167 @@
...
+    {
+      description:
+        'Time to receive the first chunk, measured from when the client issues the generation request to when the first chunk is received in the response stream.',
+      unit: 's',
+      advice: {
+        explicitBucketBoundaries: GENAI_TIME_TO_FIRST_CHUNK_BUCKETS,
+      },
+    }
+  );
+}
```
lmolkova, line 111: Consider adding `createTimePerOutputChunkHistogram` and `GENAI_TIME_PER_OUTPUT_CHUNK_BUCKETS` for `gen_ai.client.operation.time_per_output_chunk` (defined in `semconv.ts`).

packages/genai-util/src/semconv.ts
```diff
@@ -0,0 +1,481 @@
...
+ */
+export const ATTR_GEN_AI_OPERATION_NAME = 'gen_ai.operation.name' as const;
+
+/**
+ * The Generative AI provider as identified by the client or server instrumentation.
+ *
+ * @example openai
+ * @example anthropic
+ * @example aws.bedrock
+ * @example google_genai
```
lmolkova: `google_genai` is not a valid provider name in semconv, should `gcp.gen_ai` or `gcp.gemini` 

https://github.com/verdie-g/semantic-conventions-genai/blob/ab523d099aa16bd01e35210c6f1766d2bc1845e4/docs/registry/attributes/gen-ai.md?plain=1#L437-L439

packages/genai-util/src/metrics.ts
psx95: Ok, I will leave it out for this PR.

packages/genai-util/src/metrics.ts
psx95: Sure, I will add that.

packages/genai-util/src/semconv.ts
psx95: Good catch, thank you!

packages/genai-util/src/metrics.ts
psx95: Good catch!

packages/genai-util/src/semconv.ts
psx95: Addressed in [4444cc6](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/4444cc680e521ea860de63a76df4f924528c9692).

packages/genai-util/src/metrics.ts
psx95: Addressed in [4444cc6](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/4444cc680e521ea860de63a76df4f924528c9692).

packages/genai-util/src/metrics.ts
psx95: Addressed in [4444cc6](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/4444cc680e521ea860de63a76df4f924528c9692).

packages/genai-util/src/metrics.ts
psx95: Addressed in [4444cc6](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/4444cc680e521ea860de63a76df4f924528c9692).

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,169 @@
...
+  });
+}
+
+/**
+ * Create standard `gen_ai.client.token.usage` histogram.
+ *
+ * @experimental
+ */
+export function createTokenUsageHistogram(meter: Meter): Histogram {
+  return meter.createHistogram(METRIC_GEN_AI_CLIENT_TOKEN_USAGE, {
```
hectorhdzg, line 84: Should this set `valueType: ValueType.INT`? Semconv defines token usage as an integer metric, but histograms default to `DOUBLE` when this is omitted.

hectorhdzg:
Looks like some of these helpers were removed in later commits. Could you update the PR description to reflect what’s currently included?

psx95:
> Looks like some of these helpers were removed in later commits. Could you update the PR description to reflect what’s currently included?

Updated the PR description. Thanks!

packages/genai-util/src/metrics.ts
psx95: Agreed, good catch! I overrode the `valueType` in [bf90bed](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/bf90bede591b29f03ecff22a1683fd4e1298c6b5).

JacksonWeber (approved):
LGTM

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,175 @@
...
+  GEN_AI_TOKEN_TYPE_VALUE_INPUT,
+  GEN_AI_TOKEN_TYPE_VALUE_OUTPUT,
+  METRIC_GEN_AI_CLIENT_OPERATION_DURATION,
+  METRIC_GEN_AI_CLIENT_OPERATION_TIME_PER_OUTPUT_CHUNK,
+  METRIC_GEN_AI_CLIENT_OPERATION_TIME_TO_FIRST_CHUNK,
+  METRIC_GEN_AI_CLIENT_TOKEN_USAGE,
+} from './semconv';
+import type { TokenUsage } from './types';
+
+/**
```
aabmass, line 23: nit I don't think we need to export all of these constants

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,175 @@
...
+export function recordOperationDuration(
+  histogram: Histogram | undefined,
+  durationSeconds: number,
+  attributes?: Attributes
+): void {
+  if (!histogram || durationSeconds < 0 || !isFinite(durationSeconds)) {
+    return;
+  }
+  histogram.record(durationSeconds, attributes);
+}
```
aabmass, line 146: IMO this function isn't adding much value, can the caller just do `histogram.record(...)`?

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  if (typeof usage.inputTokens === 'number' && usage.inputTokens >= 0) {
+    histogram.record(usage.inputTokens, {
+      ...attributes,
+      [ATTR_GEN_AI_TOKEN_TYPE]: GEN_AI_TOKEN_TYPE_VALUE_INPUT,
+    });
+  }
+
+  if (typeof usage.outputTokens === 'number' && usage.outputTokens >= 0) {
```
aabmass, line 169: I would have to double check what we do in python or define in semconv, but it feels weird to drop the zeros on the floor. Was there a reason for this?

packages/genai-util/src/metrics.ts
```diff
@@ -0,0 +1,175 @@
...
+ * @experimental
+ */
+export function recordTokenUsage(
+  histogram: Histogram | undefined,
+  usage: TokenUsage | undefined,
+  attributes?: Attributes
+): void {
+  if (!histogram || !usage) {
+    return;
+  }
```
aabmass, line 160:
```suggestion
  histogram: Histogram,
  usage: TokenUsage,
  attributes?: Attributes
): void {
```

unless there is a good reason, can we just forbid undefined?

packages/genai-util/test/metrics.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  beforeEach(() => {
+    metricReader = new TestMetricReader();
+    meterProvider = new MeterProvider({ readers: [metricReader] });
+  });
+
+  afterEach(async () => {
+    await meterProvider.shutdown();
+  });
+
+  it('should define standard bucket boundaries', () => {
```
aabmass: I guess this is why we export the constants, but not sure these tests provides much value. If you really wanted you could validate the buckets on the actual metric produced by the SDK

packages/genai-util/test/semconv.test.ts
```diff
@@ -100,6 +102,14 @@ describe('GenAI Semantic Conventions', () => {
       METRIC_GEN_AI_CLIENT_TOKEN_USAGE,
       'gen_ai.client.token.usage'
     );
+    assert.strictEqual(
```
aabmass, line 105: oop missed these before but testing the values of the constants is unnecessary imo

aabmass (changes requested):

packages/genai-util/src/metrics.ts
aabmass: Oops missed the `>=`. But still it stands, I don't think we need to validate the value and drop it on the floor. The SDK should already handle this.

packages/genai-util/test/metrics.test.ts
psx95: Yeah, I will clean up some unnecessary tests in the PR. Thanks for the review! 

Addressed in [434c2e9](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/434c2e9ddcf9e1e3c0138bcffa43e758ca9aa34b).

packages/genai-util/src/metrics.ts
psx95: Good point, they're used only within this file.

Addressed in [434c2e9](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/434c2e9ddcf9e1e3c0138bcffa43e758ca9aa34b).

packages/genai-util/src/metrics.ts
psx95: Agreed, actually it is going to be removed in the next PR - https://github.com/psx95/opentelemetry-js-contrib/pull/2/changes#diff-bb9bbaf5f970e3932f8afc63a2c8c6851da1b5827a5f55719a0b6c6ba4c930f8L132. 

Technically it could be removed in this PR as well, but I realized this function is extraneous while working on the next one. Let me know if you think I should remove it here instead.

packages/genai-util/src/metrics.ts
psx95: I can address this in the later PRs when this is actually used - in the core invocations. This function in its current form will be removed in the next PR.

packages/genai-util/src/metrics.ts
psx95: I'm not sure what is meant by 'Drop it to the floor'. Could you expand on this a little? 

The purpose of this is to only record token usage if its reported as a positive number.

I'm not sure if all SDKs handle it - I think its simple enough to check it here.

packages/genai-util/test/semconv.test.ts
psx95: I agree, but then `semconv.ts` has no testable logic. These tests keep the coverage numbers up which could go dramatically down as semconv.ts is huge. 

Technically, we could delete the entire `semconv.test.ts`. Let me know if you have any suggestions around this.

packages/genai-util/src/metrics.ts
aabmass: The SDK is intended to handle negatives ([spec](https://opentelemetry.io/docs/specs/otel/metrics/api/#:~:text=The%20value%20is%20expected%20to%20be%20non%2Dnegative.%20This%20API%20SHOULD%20be%20documented%20in%20a%20way%20to%20communicate%20to%20users%20that%20this%20value%20is%20expected%20to%20be%20non%2Dnegative.%20This%20API%20SHOULD%20NOT%20validate%20this%20value%2C%20that%20is%20left%20to%20implementations%20of%20the%20API.)). The JS SDK handles negatives and non finite values already https://github.com/open-telemetry/opentelemetry-js/blob/3d1d427a445d6d1b76b94785c3d0f4048e38cc69/packages/sdk-metrics/test/Instruments.test.ts#L283-L298

For the JS SDK, it also has some logging which is being bypassed here https://github.com/open-telemetry/opentelemetry-js/blob/3d1d427a445d6d1b76b94785c3d0f4048e38cc69/packages/sdk-metrics/src/Instruments.ts#L121-L126

packages/genai-util/src/metrics.ts
aabmass: > Let me know if you think I should remove it here instead.

SGTM

packages/genai-util/src/metrics.ts
psx95: Removed this function in [391f19e](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/391f19ef3c4fe936cb4f63c95f5942036ecfe1d8).

packages/genai-util/src/metrics.ts
psx95: Removed this function in [391f19e](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/391f19ef3c4fe936cb4f63c95f5942036ecfe1d8).

packages/genai-util/src/metrics.ts
psx95: Removed this function in [391f19e](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3712/commits/391f19ef3c4fe936cb4f63c95f5942036ecfe1d8).

hectorhdzg (approved):
LGTM, thanks for your contribution

packages/genai-util/test/semconv.test.ts
aabmass: OK can clean them up in a later PR

psx95: Will do a cleanup after the remaining PRs are merged. Thanks!

lmolkova (approved):
LGTM!

Please update PR description

eternalcuriouslearner (approved):

aabmass (approved):

PR #3714, opened 2026-08-31
fix(instrumentation-aws-lambda): force flush logger provider on invocation completion

vasireddy99:

The `AwsLambdaInstrumentation` force-flushes the tracer and meter providers when the wrapped handler settles (`_endInvocationSpanAndFlush`), but it has no awareness of the logger provider. Log records buffered in a `BatchLogRecordProcessor` are therefore dropped when AWS freezes the Lambda execution environment right after the handler returns, unless the application manually calls `forceFlush()`.

Related (downstream report): open-telemetry/opentelemetry-lambda#2537

- Add a `setLoggerProvider` override plus a `_logForceFlusher`, mirroring the existing `setMeterProvider` / `_metricForceFlush` handling.
- Include the logger force-flush in the invocation-end flush so buffered logs are exported alongside spans and metrics, with no application code changes.
- Add `@opentelemetry/api-logs` as a dependency (used by the new method signature); the added tests use `@opentelemetry/sdk-logs`.
- Extend `lambda-handler.force-flush.test.ts` with a `should force flush LoggerProvider` case and assert the logger provider is also flushed in the combined "complete handler after force flush providers" test.

### DIFF
--- a/packages/instrumentation-aws-lambda/src/instrumentation.ts
+++ b/packages/instrumentation-aws-lambda/src/instrumentation.ts
@@ -740,7 +740,18 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     if (!loggerProvider) return undefined;
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const currentProvider: any = loggerProvider;
+    let currentProvider: any = loggerProvider;
+
+    // A LoggerProvider obtained via @opentelemetry/api-logs is a delegating
+    // proxy (ProxyLoggerProvider), like ProxyTracerProvider, so unwrap it to
+    // reach the SDK provider that actually has forceFlush. The logs proxy
+    // exposes the delegate as _getDelegate(); fall back to getDelegate() for
+    // any provider that uses the non-underscore name.
+    if (typeof currentProvider._getDelegate === 'function') {
+      currentProvider = currentProvider._getDelegate();
+    } else if (typeof currentProvider.getDelegate === 'function') {
+      currentProvider = currentProvider.getDelegate();
+    }
 
     if (typeof currentProvider.forceFlush === 'function') {
       return currentProvider.forceFlush.bind(currentProvider);
--- a/packages/instrumentation-aws-lambda/src/instrumentation.ts
+++ b/packages/instrumentation-aws-lambda/src/instrumentation.ts
@@ -38,6 +38,7 @@ import {
   ATTR_SERVER_ADDRESS,
   ATTR_URL_FULL,
 } from '@opentelemetry/semantic-conventions';
+import { LoggerProvider } from '@opentelemetry/api-logs';
 import {
   ATTR_AWS_SQS_QUEUE_URL,
   ATTR_CLOUD_ACCOUNT_ID,
@@ -121,6 +122,7 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
   private readonly _xrayPropagator = new AWSXRayPropagator();
   declare private _traceForceFlusher?: () => Promise<void>;
   declare private _metricForceFlusher?: () => Promise<void>;
+  declare private _logForceFlusher?: () => Promise<void>;
 
   constructor(config: AwsLambdaInstrumentationConfig = {}) {
     super(PACKAGE_NAME, PACKAGE_VERSION, config);
@@ -729,6 +731,24 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     return undefined;
   }
 
+  override setLoggerProvider(loggerProvider: LoggerProvider) {
+    super.setLoggerProvider(loggerProvider);
+    this._logForceFlusher = this._logForceFlush(loggerProvider);
+  }
+
+  private _logForceFlush(loggerProvider: LoggerProvider) {
+    if (!loggerProvider) return undefined;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const currentProvider: any = loggerProvider;
+
+    if (typeof currentProvider.forceFlush === 'function') {
+      return currentProvider.forceFlush.bind(currentProvider);
+    }
+
+    return undefined;
+  }
+
   private _wrapCallback(
     original: Callback,
     span: Span,
@@ -798,6 +818,13 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     } else {
       diag.debug(
         'Metrics may not be exported for the lambda function because we are not force flushing before handler completion.'
+      );
+    }
+    if (this._logForceFlusher) {
+      flushers.push(this._logForceFlusher());
+    } else {
+      diag.debug(
+        'Logs may not be exported for the lambda function because we are not force flushing before handler completion.'
       );
     }
 
--- a/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
+++ b/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
@@ -25,11 +25,17 @@ import {
   MeterProvider,
   PeriodicExportingMetricReader,
 } from '@opentelemetry/sdk-metrics';
+import {
+  InMemoryLogRecordExporter,
+  LoggerProvider,
+  SimpleLogRecordProcessor,
+} from '@opentelemetry/sdk-logs';
 
 const traceMemoryExporter = new InMemorySpanExporter();
 const metricMemoryExporter = new InMemoryMetricExporter(
   AggregationTemporality.CUMULATIVE
 );
+const logMemoryExporter = new InMemoryLogRecordExporter();
 
 describe('force flush', () => {
   let instrumentation: AwsLambdaInstrumentation;
@@ -62,6 +68,16 @@ describe('force flush', () => {
     instrumentation.setMeterProvider(provider);
   };
 
+  const initializeHandlerLogging = (
+    handler: string,
+    provider: LoggerProvider
+  ) => {
+    process.env._HANDLER = handler;
+
+    instrumentation = new AwsLambdaInstrumentation();
+    instrumentation.setLoggerProvider(provider);
+  };
+
   const lambdaRequire = (module: string) =>
     require(path.resolve(__dirname, '..', module));
 
@@ -76,6 +92,7 @@ describe('force flush', () => {
 
     traceMemoryExporter.reset();
     metricMemoryExporter.reset();
+    logMemoryExporter.reset();
   });
 
   it('should force flush TracerProvider', async () => {
@@ -140,6 +157,26 @@ describe('force flush', () => {
     assert.strictEqual(forceFlushed, true);
   });
 
+  it('should force flush LoggerProvider', async () => {
+    const provider = new LoggerProvider({
+      processors: [
+        new SimpleLogRecordProcessor({ exporter: logMemoryExporter }),
+      ],
+    });
+    let forceFlushed = false;
+    const forceFlush = () =>
+      new Promise<void>(resolve => {
+        forceFlushed = true;
+        resolve();
+      });
+    provider.forceFlush = forceFlush;
+    initializeHandlerLogging('lambda-test/sync.handler', provider);
+
+    await lambdaRequire('lambda-test/sync').handler('arg', ctx);
+
+    assert.strictEqual(forceFlushed, true);
+  });
+
   it('should complete handler after force flush providers', async () => {
     const nodeTracerProvider = new TracerProvider({
       spanProcessors: [
@@ -169,15 +206,30 @@ describe('force flush', () => {
       });
     meterProvider.forceFlush = meterForceFlush;
 
+    const loggerProvider = new LoggerProvider({
+      processors: [
+        new SimpleLogRecordProcessor({ exporter: logMemoryExporter }),
+      ],
+    });
+    let loggerForceFlushed = false;
+    const loggerForceFlush = () =>
+      new Promise<void>(resolve => {
+        loggerForceFlushed = true;
+        resolve();
+      });
+    loggerProvider.forceFlush = loggerForceFlush;
+
     process.env._HANDLER = 'lambda-test/sync.handler';
 
     instrumentation = new AwsLambdaInstrumentation();
     instrumentation.setTracerProvider(tracerProvider);
     instrumentation.setMeterProvider(meterProvider);
+    instrumentation.setLoggerProvider(loggerProvider);
 
     await lambdaRequire('lambda-test/sync').handler('arg', ctx);
 
     assert.strictEqual(tracerForceFlushed, true);
     assert.strictEqual(meterForceFlushed, true);
+    assert.strictEqual(loggerForceFlushed, true);
   });
 });
--- a/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
+++ b/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
@@ -90,6 +90,8 @@ describe('force flush', () => {
   afterEach(() => {
     process.env = oldEnv;
     instrumentation.disable();
+    // Reset the global logger provider so the delegating-proxy test starts clean.
+    logs.disable();
 
     traceMemoryExporter.reset();
     metricMemoryExporter.reset();
@@ -179,10 +181,11 @@ describe('force flush', () => {
   });
 
   it('should force flush a LoggerProvider obtained via the API (delegating proxy)', async () => {
-    // Obtaining the provider from the API before a global is registered yields a
-    // delegating ProxyLoggerProvider. setGlobalLoggerProvider then wires the
-    // delegate, mirroring how instrumentations capture the provider at load time.
-    logs.disable();
+    // A LoggerProvider obtained from the logs API is a delegating
+    // ProxyLoggerProvider. We wire its delegate to a real LoggerProvider and
+    // then hand the proxy to the instrumentation, mirroring the
+    // ProxyTracerProvider test above. The delegate is wired before
+    // setLoggerProvider because the force-flusher is resolved at that point.
     const provider = logs.getLoggerProvider();
 
     const loggerProvider = new LoggerProvider({
@@ -206,7 +209,6 @@ describe('force flush', () => {
     await lambdaRequire('lambda-test/sync').handler('arg', ctx);
 
     assert.strictEqual(forceFlushed, true);
-    logs.disable();
   });
 
   it('should complete handler after force flush providers', async () => {
--- a/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
+++ b/packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
@@ -30,6 +30,7 @@ import {
   LoggerProvider,
   SimpleLogRecordProcessor,
 } from '@opentelemetry/sdk-logs';
+import { logs } from '@opentelemetry/api-logs';
 
 const traceMemoryExporter = new InMemorySpanExporter();
 const metricMemoryExporter = new InMemoryMetricExporter(
@@ -175,6 +176,37 @@ describe('force flush', () => {
     await lambdaRequire('lambda-test/sync').handler('arg', ctx);
 
     assert.strictEqual(forceFlushed, true);
+  });
+
+  it('should force flush a LoggerProvider obtained via the API (delegating proxy)', async () => {
+    // Obtaining the provider from the API before a global is registered yields a
+    // delegating ProxyLoggerProvider. setGlobalLoggerProvider then wires the
+    // delegate, mirroring how instrumentations capture the provider at load time.
+    logs.disable();
+    const provider = logs.getLoggerProvider();
+
+    const loggerProvider = new LoggerProvider({
+      processors: [
+        new SimpleLogRecordProcessor({ exporter: logMemoryExporter }),
+      ],
+    });
+    let forceFlushed = false;
+    const forceFlush = () =>
+      new Promise<void>(resolve => {
+        forceFlushed = true;
+        resolve();
+      });
+    loggerProvider.forceFlush = forceFlush;
+    logs.setGlobalLoggerProvider(loggerProvider);
+
+    process.env._HANDLER = 'lambda-test/sync.handler';
+    instrumentation = new AwsLambdaInstrumentation();
+    instrumentation.setLoggerProvider(provider);
+
+    await lambdaRequire('lambda-test/sync').handler('arg', ctx);
+
+    assert.strictEqual(forceFlushed, true);
+    logs.disable();
   });
 
   it('should complete handler after force flush providers', async () => {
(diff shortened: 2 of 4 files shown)


packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -729,6 +731,24 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+
+  private _logForceFlush(loggerProvider: LoggerProvider) {
+    if (!loggerProvider) return undefined;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const currentProvider: any = loggerProvider;
+
+    if (typeof currentProvider.forceFlush === 'function') {
+      return currentProvider.forceFlush.bind(currentProvider);
+    }
```
pichlermarc, line 758: A `LoggerProvider` obtained via `@opentelemetry/api` has the same delegating proxy that a `TracerProvider` obtained via the `@opentelemetry/api` has.

**suggestion:** add the same logic here as well + a test with a `LoggerProvider` obtained via the API.

vasireddy99: Agreed, pushed commits.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -729,6 +731,35 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     return undefined;
   }
 
+  override setLoggerProvider(loggerProvider: LoggerProvider) {
+    super.setLoggerProvider(loggerProvider);
+    this._logForceFlusher = this._logForceFlush(loggerProvider);
```
jj22ee: The unwrap added in `45a1780` runs eagerly here, so it captures whatever the proxy's delegate is *at registration time*. Under the standard `NodeSDK` ordering that's `NOOP_LOGGER_PROVIDER`, which has no `forceFlush`, so `_logForceFlusher` stays `undefined` **permanently** — even though the real provider is registered a moment later.

`NodeSDK.start()` does:
1. `registerInstrumentations({ instrumentations })` — no `loggerProvider`, so `autoLoader` defaults to `logs.getLoggerProvider()` → the proxy, delegate unset
2. → `instrumentation.setLoggerProvider(proxy)` → `_getDelegate()` → Noop → flusher `undefined`
3. *only then* `logs.setGlobalLoggerProvider(loggerProvider)`

Reproduced against real `api-logs` + `sdk-logs`:
```
Order A (global set first, as in the new test) -> flusher resolved: true
Order B (real NodeSDK order)                   -> flusher resolved: false
```

Note `sdk-node` has an explicit workaround loop re-calling `instrumentation.setMeterProvider(...)` after registering the global, with a TODO citing open-telemetry/opentelemetry-js#3609 — there's **no equivalent loop for loggers**, so this can't rely on being called a second time.

Resolving lazily at flush time fixes it; the proxy is a stable singleton that gets populated in place:

```ts
private _logForceFlush(loggerProvider: LoggerProvider) {
  if (!loggerProvider) return undefined;

  return async () => {
    // Resolve lazily: the provider handed to instrumentations at registration
    // time is often a not-yet-populated ProxyLoggerProvider, because NodeSDK
    // registers instrumentations before calling logs.setGlobalLoggerProvider().
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    let p: any = loggerProvider;
    if (typeof p.forceFlush !== 'function' && typeof p._getDelegate === 'function') {
      p = p._getDelegate();
    }
    if (typeof p?.forceFlush === 'function') return p.forceFlush();
    diag.debug(
      'Logs may not be exported for the lambda function because we are not force flushing before handler completion.'
    );
  };
}
```

With the debug moved inside the closure, the `else` branch at line 836 becomes dead and can go.

`_traceForceFlush` has the identical latent hole (`TRACE, NodeSDK order -> false`, `lazy -> true`), so applying the same shape there is a free bonus fix — fine as a follow-up if you'd rather keep this PR scoped.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -729,6 +731,35 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
...
+    // A LoggerProvider obtained via @opentelemetry/api-logs is a delegating
+    // proxy (ProxyLoggerProvider), like ProxyTracerProvider, so unwrap it to
+    // reach the SDK provider that actually has forceFlush. The logs proxy
+    // exposes the delegate as _getDelegate(); fall back to getDelegate() for
+    // any provider that uses the non-underscore name.
+    if (typeof currentProvider._getDelegate === 'function') {
+      currentProvider = currentProvider._getDelegate();
+    } else if (typeof currentProvider.getDelegate === 'function') {
+      currentProvider = currentProvider.getDelegate();
+    }
```
jj22ee: Two things about this unwrap:

1. `ProxyLoggerProvider._getDelegate()` is marked `@internal` with the doc comment *"Used by tests only"*, and `ProxyLoggerProvider` isn't exported from `@opentelemetry/api-logs` at all — the package only does `export type { LoggerProvider }`. That's asymmetric with traces, where `ProxyTracerProvider.getDelegate()` is public and exported. So this reaches into private API of an `experimental/` package that reserves the right to rename it in a patch.
2. The `getDelegate()` fallback is dead code — nothing in the logs API uses the non-underscore name, so it can only ever match a hypothetical third-party provider.

Checking `forceFlush` *before* reaching for `_getDelegate` (as in the lazy version I suggested on line 736) keeps the private-API surface to the minimum and lets you drop the dead branch. Worth also filing an upstream issue asking for a public `getDelegate()` on `ProxyLoggerProvider` and linking it here, so this can be cleaned up later.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -38,6 +38,7 @@ import {
   ATTR_SERVER_ADDRESS,
   ATTR_URL_FULL,
 } from '@opentelemetry/semantic-conventions';
+import { LoggerProvider } from '@opentelemetry/api-logs';
```
jj22ee: `@opentelemetry/api-logs` exports this as `export type { LoggerProvider }` — a value import of a type-only export breaks under `verbatimModuleSyntax`.

```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

Keeping `api-logs` in `dependencies` rather than `devDependencies` is correct though, since it appears in a public method signature and consumers need it to resolve the emitted `.d.ts`.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```diff
@@ -729,6 +731,35 @@ export class AwsLambdaInstrumentation extends InstrumentationBase<AwsLambdaInstr
     return undefined;
   }
 
+  override setLoggerProvider(loggerProvider: LoggerProvider) {
+    super.setLoggerProvider(loggerProvider);
+    this._logForceFlusher = this._logForceFlush(loggerProvider);
+  }
+
+  private _logForceFlush(loggerProvider: LoggerProvider) {
```
jj22ee: Nit: `_traceForceFlush` / `_metricForceFlush` / `_logForceFlush` are now three near-identical methods. Could collapse to one `private _resolveForceFlusher(provider: unknown, kind: string)` helper. Optional, but it'd mean the ordering fix on line 736 only has to be written once.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -140,6 +158,57 @@ describe('force flush', () => {
...
+
+    await lambdaRequire('lambda-test/sync').handler('arg', ctx);
+
+    assert.strictEqual(forceFlushed, true);
+  });
+
+  it('should force flush a LoggerProvider obtained via the API (delegating proxy)', async () => {
+    // Obtaining the provider from the API before a global is registered yields a
+    // delegating ProxyLoggerProvider. setGlobalLoggerProvider then wires the
+    // delegate, mirroring how instrumentations capture the provider at load time.
```
jj22ee: This test only exercises the ordering that already works: `setGlobalLoggerProvider` on line 200 runs *before* `setLoggerProvider` on line 204. Swap those two and it fails as written — and that's the ordering real `NodeSDK` users get (see my comment on `src/instrumentation.ts:736`).

Could you add the reverse-order case? Something like:

```ts
it('should force flush a LoggerProvider registered after the instrumentation', async () => {
  logs.disable();
  const proxy = logs.getLoggerProvider();

  process.env._HANDLER = 'lambda-test/sync.handler';
  instrumentation = new AwsLambdaInstrumentation();
  instrumentation.setLoggerProvider(proxy); // captured before the global exists

  const loggerProvider = new LoggerProvider({
    processors: [new SimpleLogRecordProcessor({ exporter: logMemoryExporter })],
  });
  let forceFlushed = false;
  loggerProvider.forceFlush = async () => { forceFlushed = true; };
  logs.setGlobalLoggerProvider(loggerProvider);

  await lambdaRequire('lambda-test/sync').handler('arg', ctx);

  assert.strictEqual(forceFlushed, true);
});
```

Also, the comment on lines 182-184 reads inverted — "mirroring how instrumentations capture the provider at load time" describes late binding, but the test wires the global up front, which is the opposite.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -140,6 +158,57 @@ describe('force flush', () => {
     assert.strictEqual(forceFlushed, true);
   });
 
+  it('should force flush LoggerProvider', async () => {
+    const provider = new LoggerProvider({
+      processors: [
+        new SimpleLogRecordProcessor({ exporter: logMemoryExporter }),
+      ],
+    });
```
jj22ee: Every new test overwrites `provider.forceFlush` with a stub, so the `SimpleLogRecordProcessor` + `logMemoryExporter` wired into each provider are never used, `logMemoryExporter` is never asserted on, and all of these would still pass if `forceFlush` were a no-op.

That matches the existing metric test, so I don't think it needs changing everywhere — but for logs there's a cheap end-to-end version of the exact bug this PR fixes. One test with a real `BatchLogRecordProcessor` + `InMemoryLogRecordExporter`, a log record emitted from the handler, **no stub**, and:

```ts
assert.strictEqual(logMemoryExporter.getFinishedLogRecords().length, 1);
```

That's the only assertion here that would catch "wired up but doesn't actually flush," which is the regression class this PR exists to prevent.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -140,6 +158,57 @@ describe('force flush', () => {
...
+    logs.setGlobalLoggerProvider(loggerProvider);
+
+    process.env._HANDLER = 'lambda-test/sync.handler';
+    instrumentation = new AwsLambdaInstrumentation();
+    instrumentation.setLoggerProvider(provider);
+
+    await lambdaRequire('lambda-test/sync').handler('arg', ctx);
+
+    assert.strictEqual(forceFlushed, true);
+    logs.disable();
```
jj22ee: If either assertion above throws, this trailing `logs.disable()` never runs and a registered global logger provider leaks into every subsequent test in the process. Please move it into the existing `afterEach` (line 90) — and adding it to `beforeEach` too would stop this file being polluted by other suites, since mocha runs these in one process.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -62,6 +69,16 @@ describe('force flush', () => {
...
 
+  const initializeHandlerLogging = (
+    handler: string,
+    provider: LoggerProvider
+  ) => {
+    process.env._HANDLER = handler;
+
+    instrumentation = new AwsLambdaInstrumentation();
+    instrumentation.setLoggerProvider(provider);
+  };
```
jj22ee: Nit: the proxy test on line 181 inlines exactly what this helper does (lines 202-204). Either reuse `initializeHandlerLogging` there or drop the helper — right now it's used once.

vasireddy99:
The tests are flaky at the moment and unrelated to the changes made.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

packages/instrumentation-aws-lambda/src/instrumentation.ts
jj22ee: This is evaluated eagerly, so on the auto path it unwraps the proxy before the delegate exists.

`registerInstrumentations()` defaults to `logs.getLoggerProvider()` and forwards it in the same call, and
`NodeSDK.start()` runs that *before* `logs.setGlobalLoggerProvider()`. So the proxy arrives with no
delegate, `_getDelegate()` returns `NOOP_LOGGER_PROVIDER`, and `_logForceFlusher` stays `undefined`
permanently — logs are still never flushed for anyone on bare `NodeSDK`.

The unwrap does help in one case: a proxy captured *before* registration and handed over *after*, which is
what the new test exercises. But that's the uncommon wiring. The two common ones are a concrete provider
(nothing to unwrap — `opentelemetry-lambda`'s wrapper and the ADOT distro both pass one) or the auto path
above (proxy arrives too early to unwrap).

Resolving at flush time covers all three, since the proxy is a stable singleton that gets populated in
place:

```ts
private _logForceFlush(loggerProvider: LoggerProvider) {
  if (!loggerProvider) return undefined;

  return async () => {
    // The proxy's delegate isn't set until logs.setGlobalLoggerProvider(),
    // which can happen after instrumentations are registered.
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    let p: any = loggerProvider;
    if (typeof p.forceFlush !== 'function' && typeof p._getDelegate === 'function') {
      p = p._getDelegate();
    }
    if (typeof p?.forceFlush === 'function') return p.forceFlush();
    diag.debug(
      'Logs may not be exported for the lambda function because we are not force flushing before handler completion.'
    );
  };
}
```

The flusher is then always truthy, so the `else` branch at line 836 can go. Checking `forceFlush` before
`_getDelegate` also keeps us off `_getDelegate` entirely on the common paths — worth doing, because upstream
removed the public `getDelegate()` in open-telemetry/opentelemetry-js#5789 specifically to stop consumers
reaching for it. That also means the `getDelegate()` fallback below can't match anything.

Non-blocking from my side.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: Non-blocking: every new test overwrites `provider.forceFlush` with a stub, so the `SimpleLogRecordProcessor` + `logMemoryExporter` wired into each provider are never used, `logMemoryExporter` is never asserted on, and all of these would still pass if `forceFlush` were a no-op.

That matches the existing metric test, so I don't think it needs changing everywhere — but for logs there's a cheap end-to-end version of the exact bug this PR fixes. One test with a real `BatchLogRecordProcessor` + `InMemoryLogRecordExporter`, a log record emitted from the handler, **no stub**, and:

```ts
assert.strictEqual(logMemoryExporter.getFinishedLogRecords().length, 1);
```

That's the only assertion here that would catch "wired up but doesn't actually flush," which is the regression class this PR exists to prevent.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: This test only exercises the ordering that already works: `setGlobalLoggerProvider` on line 200 runs *before* `setLoggerProvider` on line 204. Swap those two and it fails as written — and that's the ordering real `NodeSDK` users get (see my comment on `src/instrumentation.ts:736`).

Could you add the reverse-order case? Something like:

```ts
it('should force flush a LoggerProvider registered after the instrumentation', async () => {
  logs.disable();
  const proxy = logs.getLoggerProvider();

  process.env._HANDLER = 'lambda-test/sync.handler';
  instrumentation = new AwsLambdaInstrumentation();
  instrumentation.setLoggerProvider(proxy); // captured before the global exists

  const loggerProvider = new LoggerProvider({
    processors: [new SimpleLogRecordProcessor({ exporter: logMemoryExporter })],
  });
  let forceFlushed = false;
  loggerProvider.forceFlush = async () => { forceFlushed = true; };
  logs.setGlobalLoggerProvider(loggerProvider);

  await lambdaRequire('lambda-test/sync').handler('arg', ctx);

  assert.strictEqual(forceFlushed, true);
});
```

Also, the comment on lines 182-184 reads inverted — "mirroring how instrumentations capture the provider at load time" describes late binding, but the test wires the global up front, which is the opposite.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: `logs.disable()` can be moved into both/either `afterEach/beforeEach`.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: Non-blocking: every new test overwrites `provider.forceFlush` with a stub, so the `SimpleLogRecordProcessor` + `logMemoryExporter` wired into each provider are never used, `logMemoryExporter` is never asserted on, and all of these would still pass if `forceFlush` were a no-op.

That matches the existing metric test, so I don't think it needs changing everywhere — but for logs there's a cheap end-to-end version of the exact bug this PR fixes. One test with a real `BatchLogRecordProcessor` + `InMemoryLogRecordExporter`, a log record emitted from the handler, **no stub**, and:

```ts
assert.strictEqual(logMemoryExporter.getFinishedLogRecords().length, 1);
```

That's the only assertion here that would catch "wired up but doesn't actually flush," which is the regression class this PR exists to prevent.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: The comment here reads inverted — "mirroring how instrumentations capture the provider at load time"
describes late binding, but the test wires the global up front (line 200) before handing the proxy over
(line 204), which is the opposite of load-time capture.

Worth noting what that leaves untested, since it's the case the proxy handling was added for: on the
`NodeSDK` auto path, `registerInstrumentations()` hands over `logs.getLoggerProvider()` within the same call,
so the proxy arrives before any global is registered and `_getDelegate()` returns `NOOP_LOGGER_PROVIDER`.
Swapping lines 200 and 204 reproduces that ordering. Covering it would mean changing how the flusher is
resolved, which matches `_traceForceFlush` today — so entirely your call whether that's in scope here.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: `logs.disable()` can be moved into both/either `afterEach/beforeEach`.

packages/instrumentation-aws-lambda/src/instrumentation.ts
```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -140,6 +158,57 @@ describe('force flush', () => {
...
+        forceFlushed = true;
+        resolve();
+      });
+    provider.forceFlush = forceFlush;
+    initializeHandlerLogging('lambda-test/sync.handler', provider);
+
+    await lambdaRequire('lambda-test/sync').handler('arg', ctx);
+
+    assert.strictEqual(forceFlushed, true);
+  });
```
jj22ee: Since this test verifies the fix, it could assert the behaviour rather than a stub — as written it'd pass
even if `forceFlush` were a no-op.

```suggestion
  it('should force flush LoggerProvider', async () => {
    const provider = new LoggerProvider({
      processors: [
        new BatchLogRecordProcessor({
          exporter: logMemoryExporter,
          scheduledDelayMillis: 30000, // so only forceFlush can export
        }),
      ],
    });
    initializeHandlerLogging('lambda-test/sync.handler', provider);

    provider.getLogger('test').emit({ body: 'buffered' });
    await lambdaRequire('lambda-test/sync').handler('arg', ctx);

    assert.strictEqual(logMemoryExporter.getFinishedLogRecords().length, 1);
  });
```

(Needs `BatchLogRecordProcessor` imported; emitting here rather than in the handler because
`sync.handler` doesn't log.)

packages/instrumentation-aws-lambda/src/instrumentation.ts
```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

```suggestion
import type { LoggerProvider } from '@opentelemetry/api-logs';
```

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
```diff
@@ -140,6 +158,57 @@ describe('force flush', () => {
...
+      new Promise<void>(resolve => {
+        forceFlushed = true;
+        resolve();
+      });
+    loggerProvider.forceFlush = forceFlush;
+    logs.setGlobalLoggerProvider(loggerProvider);
+
+    process.env._HANDLER = 'lambda-test/sync.handler';
+    instrumentation = new AwsLambdaInstrumentation();
+    instrumentation.setLoggerProvider(provider);
```
jj22ee, line 207: Should these two be the other way round, with `instrumentation.setLoggerProvider` before `logs.setGlobalLoggerProvider`? The comment above describes the instrumentation receiving the provider before the delegate is wired, which is the opposite of what the test does.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
jj22ee: `logs.disable()` can be moved into both/either `afterEach/beforeEach`.

packages/instrumentation-aws-lambda/test/integrations/lambda-handler.force-flush.test.ts
vasireddy99: Good catch on the mismatch, I fixed the comment. The order is intentional and mirrors the ProxyTracerProvider test above. The log force flusher is resolved when setLoggerProvider is called, the same as the tracer and meter flushers, so the delegate has to be wired on the proxy first. If the delegate is wired after, getDelegate() returns the noop provider which has no forceFlush and nothing would be flushed.

jj22ee (approved):
lgtm!

vasireddy99:
Hi @pichlermarc can you help with review/merge.

PR #3724, opened 2026-09-01
fix(instrumentation-anthropic): harden async lifecycle

eternalcuriouslearner:
## Which problem is this PR solving?\n\nEager APIPromise observation can consume raw response bodies, and async lifecycle edge cases need explicit suppression, context, and stream error coverage.\n\n## Short description of the changes\n\n- Preserves APIPromise, asResponse, and withResponse behavior without eager parsing.\n- Respects trace suppression and active parent context.\n- Covers stream completion, tee, early close, stream errors, and messages.parse on SDK versions that provide it.\n- Validates Anthropic SDK versions 0.65.0 through 0.122.0.\n\nStack layer 6 of 6. Depends on the preceding client-metrics PR.


eternalcuriouslearner:
Closing this PR because the upstream repository rejected creation of the requested stack: Stacked PRs are not enabled here, and the preceding base branch exists only in the contributor fork. The branch and fork-local stacked PR remain available.

PR #3727, opened 2026-09-03
fix(instrumentation-document-load): replace deprecated otperformance

aryamohanan:
Replace usage of `otperformance` from `@opentelemetry/core` with the global `performance` object directly, following the removal of the deprecated export in opentelemetry-js core in upcoming SDK V3.

See: https://github.com/open-telemetry/opentelemetry-js/pull/7053/changes


david-luna (approved):
LGTM

PR #3731, opened 2026-09-03
fix(instrumentation-undici): add error.type to spans and metrics

### LINKED ISSUE #3729: [instrumentation-undici] error.type is not set when the response is 4xx or 5xx
`@opentelemetry/instrumentation-undici@0.31.0`, `undici@8.10.0`.

22.19.0

### What did you do?
Sent a request through `undici.request` to an endpoint that answers 500.

The client span carrying `error.type: "500"`, and `http.client.request.duration` recorded with the same attribute.

`error.type` is [Conditionally Required](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) on client spans when the request ends with an error, and SHOULD be the status code as a string when a response was received. It is also a required attribute of `http.client.request.duration` in the same case.

### What did you see instead?
The span has `status.code=error` and no `error.type`.

### Additional context
This finding is from the [semantic conventions conformance repo](https://github.com/open-telemetry/semantic-conventions-conformance/). See [here](https://github.com/open-telemetry/semantic-conventions-conformance/blob/main/scenarios/http/js/undici/opentelemetry-undici/client/data.json#L54-L60).

mwear:
This PR adds the conditionally required `error.type` attribute to both spans and metrics when a request has ended with an error. 

This was an issue identified by the [semantic conventions conformance](https://github.com/open-telemetry/semantic-conventions-conformance/) repo.

Fixes: #3729 

- `onResponseHeaders` now sets `error.type` to the status code as a string when the response is an error. 
- metrics inherit `error.type` from `spanAttributes`
- `assertSpan` now checks `error.type`

### DIFF
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -396,9 +396,14 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     }
 
     const { span, attributes } = record;
+    const isError = response.statusCode >= 400;
     const spanAttributes: Attributes = {
       [ATTR_HTTP_RESPONSE_STATUS_CODE]: response.statusCode,
     };
+
+    if (isError) {
+      spanAttributes[ATTR_ERROR_TYPE] = String(response.statusCode);
+    }
 
     const config = this.getConfig();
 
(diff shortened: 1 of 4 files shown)


david-luna (approved):
Thanks for your contribution @mwear

trentm (approved):

PR #3736, opened 2026-09-04
fix(instrumentation-undici): add `error.type` span attribute for requests that fail without a server response

trentm:
In earlier PRs `error.type` was added to spans for HTTP response codes
above 400, and also to HTTP client *metrics* for when a request fails
without a server response (e.g. the server crashes). This PR adds
the `error.type` attribute *to the span* for the latter case.

The HTTP span semconv (https://opentelemetry.io/docs/specs/semconv/http/http-spans/) also states:

> Instrumentations SHOULD document the list of errors they report.

so I've updated the semconv table in the README with some specifics.

### DIFF
--- a/packages/instrumentation-undici/README.md
+++ b/packages/instrumentation-undici/README.md
@@ -68,7 +68,7 @@ Attributes collected:
 
 | Attribute                      | Short Description                                                                                          |
 |--------------------------------|------------------------------------------------------------------------------------------------------------|
-| `error.type`                   | Describes a class of error the operation ended with.                                                       |
+| `error.type`                   | An HTTP status code, as a string, for HTTP responses >=400; or [an `'UND_ERR_*'` code](https://undici.nodejs.org/api/Errors) if the request errored without a server response. Excluded if the request suceeded. |
 | `http.request.method`          | HTTP request method.                                                                                       |
 | `http.request.method_original` | Original HTTP method sent by the client in the request line.                                               |
 | `http.response.status_code`    | [HTTP response status code](https://tools.ietf.org/html/rfc7231#section-6).                                |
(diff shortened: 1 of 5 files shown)


packages/instrumentation-undici/README.md
```diff
@@ -68,7 +68,7 @@ Attributes collected:
 
 | Attribute                      | Short Description                                                                                          |
 |--------------------------------|------------------------------------------------------------------------------------------------------------|
-| `error.type`                   | Describes a class of error the operation ended with.                                                       |
+| `error.type`                   | An HTTP status code, as a string, for HTTP responses >=400; or [an `'UND_ERR_*'` code](https://undici.nodejs.org/api/Errors) if the request errored without a server response. Excluded if the request suceeded. |
```
mwear:
```suggestion
| `error.type`                   | An HTTP status code, as a string, for HTTP responses >=400; or [an `'UND_ERR_*'` code](https://undici.nodejs.org/api/Errors) if the request errored without a server response. Excluded if the request succeeded. |
```

mwear (commented):
Is it worth adding an assertion to:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/914f249a07ba56b87d359b793340a3673acef32d/packages/instrumentation-undici/test/fetch.test.ts#L362

while we're here?

david-luna (approved):

maryliag (approved):
besides Matthew's comments, lgtm

trentm:
> Is it worth adding an assertion to:

Yup, thanks. commit 1279c3a9

mwear (approved):
Thanks @trentm. LGTM.

PR #3737, opened 2026-09-04
feat(instrumentation-undici): set network.protocol.version on client spans

### LINKED ISSUE #3730: [instrumentation-undici] set network.protocol.version on client spans
`@opentelemetry/instrumentation-undici` does not set `network.protocol.version` on client spans; `@opentelemetry/instrumentation-http` does. It would nice if this was consistent.

The attribute is [Recommended](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) on HTTP client spans.

Set it in `onRequestHeaders`, which already has the socket in hand:

```js
// src/undici.ts:351
private onRequestHeaders({ request, socket }: RequestHeadersMessage): void {
  const { remoteAddress, remotePort } = socket;
```

### Additional context
Found by [semantic-conventions-conformance](https://github.com/open-telemetry/semantic-conventions-conformance). See [here](https://github.com/open-telemetry/semantic-conventions-conformance/blob/main/scenarios/http/js/undici/opentelemetry-undici/client/data.json#L62-L81)

Yashisinghal285:

Fixes #3730

- `@opentelemetry/instrumentation-undici` was not setting `network.protocol.version` on HTTP client spans, unlike `@opentelemetry/instrumentation-http`.
- The `network.protocol.version` attribute is recommended on HTTP client spans according to OpenTelemetry HTTP Semantic Conventions, and was flagged as missing by `semantic-conventions-conformance`.

- **Set `network.protocol.version` in `onRequestHeaders`**: Retrieves the connection socket in `onRequestHeaders` and populates the `ATTR_NETWORK_PROTOCOL_VERSION` attribute on the span.
- **ALPN parsing & fallback (`getProtocolVersion`)**:
  - Normalizes negotiated TLS ALPN values from `socket.alpnProtocol` (`'http/1.1'` -> `'1.1'`, `'http/1.0'` -> `'1.0'`, `'h2'` -> `'2'`, `'h3'` -> `'3'`).
  - Defaults to `'1.1'` for standard HTTP/1.1 connections without ALPN (plain TCP or TLS without ALPN).
  - Safely returns `undefined` if socket information is absent.
- **Updated test helper**: Added `ATTR_NETWORK_PROTOCOL_VERSION` assertion to `assertSpan` when network peer attributes are expected.
- **Added unit tests**: Added tests in `undici.test.ts` for end-to-end client span attributes, `getProtocolVersion` ALPN parsing, and `onRequestHeaders` socket handling.
- **Updated documentation**: Added `network.protocol.version` to the "Attributes collected" table in `packages/instrumentation-undici/README.md`.

Assisted-by: Google Antigravity

### DIFF
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -34,6 +34,7 @@ import {
   ATTR_HTTP_RESPONSE_STATUS_CODE,
   ATTR_NETWORK_PEER_ADDRESS,
   ATTR_NETWORK_PEER_PORT,
+  ATTR_NETWORK_PROTOCOL_VERSION,
   ATTR_SERVER_ADDRESS,
   ATTR_SERVER_PORT,
   ATTR_URL_FULL,
@@ -363,6 +364,11 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
       [ATTR_NETWORK_PEER_PORT]: remotePort,
     };
 
+    const protocolVersion = this.getProtocolVersion(socket);
+    if (protocolVersion) {
+      spanAttributes[ATTR_NETWORK_PROTOCOL_VERSION] = protocolVersion;
+    }
+
     // After hooks have been processed (which may modify request headers)
     // we can collect the headers based on the configuration
     if (config.headersToSpanAttributes?.requestHeaders) {
@@ -380,6 +386,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     }
 
     span.setAttributes(spanAttributes);
+    record.attributes = Object.assign(record.attributes, spanAttributes);
   }
 
   // This is the 3rd message we get for each request and it's fired when the server
@@ -552,5 +559,27 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     }
 
     return '_OTHER';
+  }
+
+  private getProtocolVersion(
+    socket: { alpnProtocol?: string | boolean } | undefined | null
+  ): string | undefined {
+    if (!socket) {
+      return undefined;
+    }
+    if (typeof socket.alpnProtocol === 'string') {
+      const alpn = socket.alpnProtocol.toLowerCase();
+      if (alpn.startsWith('http/')) {
+        return alpn.slice(5);
+      }
+      if (alpn === 'h2') {
+        return '2';
+      }
+      if (alpn === 'h3') {
+        return '3';
+      }
+      return socket.alpnProtocol;
+    }
+    return '1.1';
   }
 }
--- a/packages/instrumentation-undici/src/undici.ts
+++ b/packages/instrumentation-undici/src/undici.ts
@@ -358,16 +358,14 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
 
     const config = this.getConfig();
     const { span } = record;
-    const { remoteAddress, remotePort } = socket;
+    const { remoteAddress, remotePort, alpnProtocol } = socket;
     const spanAttributes: Attributes = {
       [ATTR_NETWORK_PEER_ADDRESS]: remoteAddress,
       [ATTR_NETWORK_PEER_PORT]: remotePort,
+      // undici only ever offers `http/1.1` and, with `allowH2`, `h2` as ALPN
+      // protocols; a connection that negotiated neither is HTTP/1.1.
+      [ATTR_NETWORK_PROTOCOL_VERSION]: alpnProtocol === 'h2' ? '2' : '1.1',
     };
-
-    const protocolVersion = this.getProtocolVersion(socket);
-    if (protocolVersion) {
-      spanAttributes[ATTR_NETWORK_PROTOCOL_VERSION] = protocolVersion;
-    }
 
     // After hooks have been processed (which may modify request headers)
     // we can collect the headers based on the configuration
@@ -523,6 +521,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
       ATTR_SERVER_PORT,
       ATTR_URL_SCHEME,
       ATTR_ERROR_TYPE,
+      ATTR_NETWORK_PROTOCOL_VERSION,
     ];
     keysToCopy.forEach(key => {
       if (key in attributes) {
@@ -559,27 +558,5 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     }
 
     return '_OTHER';
-  }
-
-  private getProtocolVersion(
-    socket: { alpnProtocol?: string | boolean } | undefined | null
-  ): string | undefined {
-    if (!socket) {
-      return undefined;
-    }
-    if (typeof socket.alpnProtocol === 'string') {
-      const alpn = socket.alpnProtocol.toLowerCase();
-      if (alpn.startsWith('http/')) {
-        return alpn.slice(5);
-      }
-      if (alpn === 'h2') {
-        return '2';
-      }
-      if (alpn === 'h3') {
-        return '3';
-      }
-      return socket.alpnProtocol;
-    }
-    return '1.1';
   }
 }
--- a/packages/instrumentation-undici/test/undici.test.ts
+++ b/packages/instrumentation-undici/test/undici.test.ts
@@ -14,7 +14,11 @@ import {
   trace,
 } from '@opentelemetry/api';
 import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
-import { ATTR_ERROR_TYPE } from '@opentelemetry/semantic-conventions';
+import {
+  ATTR_ERROR_TYPE,
+  ATTR_NETWORK_PROTOCOL_VERSION,
+} from '@opentelemetry/semantic-conventions';
+import * as sinon from 'sinon';
 import {
   InMemorySpanExporter,
   SimpleSpanProcessor,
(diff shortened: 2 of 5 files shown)


mwear:
@Yashisinghal285, can you sign the CLA? The instructions are above.

Yashisinghal285:
Hi @mwear, I have signed the EasyCLA now and the check is green! Thank you.

packages/instrumentation-undici/src/undici.ts
```diff
@@ -363,6 +364,11 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
       [ATTR_NETWORK_PEER_PORT]: remotePort,
     };
 
+    const protocolVersion = this.getProtocolVersion(socket);
+    if (protocolVersion) {
+      spanAttributes[ATTR_NETWORK_PROTOCOL_VERSION] = protocolVersion;
+    }
```
mwear: From my research Undici only uses `h2` or `http1.1`. When `h2` is used it will be set on the socket. In all other cases it will be `http1.1`. So we can can simplify this to:
```suggestion
    const { remoteAddress, remotePort, alpnProtocol } = socket;
    const spanAttributes: Attributes = {
      [ATTR_NETWORK_PEER_ADDRESS]: remoteAddress,
      [ATTR_NETWORK_PEER_PORT]: remotePort,
      [ATTR_NETWORK_PROTOCOL_VERSION]: alpnProtocol === 'h2' ? '2' : '1.1',
    };
```

That means that we can delete the `getProtocolVersion` helper.

packages/instrumentation-undici/src/undici.ts
```diff
@@ -380,6 +386,7 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
     }
 
     span.setAttributes(spanAttributes);
+    record.attributes = Object.assign(record.attributes, spanAttributes);
```
mwear, line 387: There is one more thing that we need to do so this attribute will end up on metrics. We need to add it to `keysToCopy` in `recordRequestDuration`, e.g.:

```ts
  private recordRequestDuration(attributes: Attributes, startTime: HrTime) {
    // Time to record metrics
    const metricsAttributes: Attributes = {};
    // Get the attribs already in span attributes
    const keysToCopy = [
      ATTR_HTTP_RESPONSE_STATUS_CODE,
      ATTR_HTTP_REQUEST_METHOD,
      ATTR_SERVER_ADDRESS,
      ATTR_SERVER_PORT,
      ATTR_URL_SCHEME,
      ATTR_ERROR_TYPE,
      ATTR_NETWORK_PROTOCOL_VERSION,
    ];
```

Just make sure to update the tests to ensure `network.protocol.version` is recorded.

packages/instrumentation-undici/src/undici.ts
mwear: From my research undici only ever offers `http/1.1` and, with `allowH2`, `h2` as ALPN protocols. So the only value we need to look for is `h2`; anything else, including a plaintext socket with no `alpnProtocol`, means HTTP/1.1. We can simplify this to:
```ts
    const { remoteAddress, remotePort, alpnProtocol } = socket;
    const spanAttributes: Attributes = {
      [ATTR_NETWORK_PEER_ADDRESS]: remoteAddress,
      [ATTR_NETWORK_PEER_PORT]: remotePort,
      // undici only ever offers `http/1.1` and, with `allowH2`, `h2` as ALPN
      // protocols; a connection that negotiated neither is HTTP/1.1.
      [ATTR_NETWORK_PROTOCOL_VERSION]: alpnProtocol === 'h2' ? '2' : '1.1',
    };
```

That means that we can delete the `getProtocolVersion` helper and tests.

packages/instrumentation-undici/test/undici.test.ts
```diff
@@ -956,5 +960,70 @@ describe('UndiciInstrumentation `undici` tests', function () {
...
+        const attrs = fakeSpan.setAttributes.firstCall.args[0];
+        assert.strictEqual(attrs[ATTR_NETWORK_PROTOCOL_VERSION], '2');
+        assert.strictEqual(
+          fakeRecord.attributes[ATTR_NETWORK_PROTOCOL_VERSION],
+          '2'
+        );
+
+        (instrumentation as any)._recordFromReq.delete(fakeRequest);
+      });
+    });
```
mwear: All of this can go. You already updated `assertSpan` which covers the `1.1` case. For `h2`, you can simulate a request with `diagnostics_channel` by publishing to `undici:client:sendHeaders` with a socket that has `alpnProtocol: 'h2'`.

packages/instrumentation-undici/test/utils/assertSpan.ts
```diff
@@ -40,6 +41,7 @@ export const assertSpan = (
     query?: string | null;
     forceStatus?: SpanStatus;
     noNetPeer?: boolean; // we don't expect net peer info when request throw before being sent
+    protocolVersion?: string;
```
mwear: We can drop this. The mock server is plaintext, which undici always speaks HTTP/1.1 over.
```suggestion
```

packages/instrumentation-undici/test/utils/assertSpan.ts
```diff
@@ -149,6 +151,11 @@ export const assertSpan = (
       span.attributes[ATTR_NETWORK_PEER_PORT],
       `must have ${ATTR_NETWORK_PEER_PORT}`
     );
+    assert.strictEqual(
+      span.attributes[ATTR_NETWORK_PROTOCOL_VERSION],
+      validations.protocolVersion || '1.1',
```
mwear: And this can just be `1.1`
```suggestion
      '1.1',
```

packages/instrumentation-undici/test/undici.test.ts
mwear, line 987: All of this can go. You already updated `assertSpan`, which covers the `1.1` case. For `h2`, you can simulate a request with `diagnostics_channel` by publishing to `undici:client:sendHeaders` with a socket that has `alpnProtocol: 'h2'`.

mwear (changes requested):
Thanks for picking this up! Two changes: the ALPN handling can collapse to a single check for `h2`, and the attribute needs adding to `keysToCopy` so it lands on the `http.client.request.duration` metric.

Yashisinghal285:
Thanks for the feedback @mwear! I've updated the PR:

Collapsed the ALPN check to look for h2 vs 1.1 and removed getProtocolVersion.
Added ATTR_NETWORK_PROTOCOL_VERSION to keysToCopy so it's recorded on http.client.request.duration.
Cleaned up assertSpan and added test coverage for both the metrics and an h2 request via diagnostics_channel.

mwear (approved):
Thanks for the changes @Yashisinghal285. This looks great! A maintainer will have to approve the workflows to run, but this looks good to me.

packages/instrumentation-undici/src/undici.ts
Yashisinghal285: Thanks! Addressed in 92675ff by simplifying the ALPN handling and replacing the helper-based tests with the requested HTTP/2 diagnostics_channel coverage.

packages/instrumentation-undici/src/undici.ts
Yashisinghal285: Thanks! Addressed in 92675ff by adding ATTR_NETWORK_PROTOCOL_VERSION to keysToCopy and updating the related test coverage.

packages/instrumentation-undici/src/undici.ts
```diff
@@ -357,10 +358,13 @@ export class UndiciInstrumentation extends InstrumentationBase<UndiciInstrumenta
...
     const config = this.getConfig();
     const { span } = record;
-    const { remoteAddress, remotePort } = socket;
+    const { remoteAddress, remotePort, alpnProtocol } = socket;
     const spanAttributes: Attributes = {
       [ATTR_NETWORK_PEER_ADDRESS]: remoteAddress,
       [ATTR_NETWORK_PEER_PORT]: remotePort,
+      // undici only ever offers `http/1.1` and, with `allowH2`, `h2` as ALPN
+      // protocols; a connection that negotiated neither is HTTP/1.1.
+      [ATTR_NETWORK_PROTOCOL_VERSION]: alpnProtocol === 'h2' ? '2' : '1.1',
```
david-luna, line 367: maybe I am a little paranoid here but according to undici docs about this diagnostic channel
https://undici.nodejs.org/api/DiagnosticsChannel#event-undiciclientsendheaders

The observer receives the request and a `net.Socket` instance. In the documentation the class does not have an `alpnProtocol` property. I think what's happening is undici sends a `tls.TLSSocket` which extends `net.Socket` and [has the property](https://nodejs.org/api/tls.html#tlssocketalpnprotocol).

If we stick to the docs getting a `net.Socket` is possible and we do not have the property available. But does that mean the HTTP protocol is 1.1?

Yashisinghal285: Yes, exactly! In Undici:
1. For HTTPS requests, Undici creates a `tls.TLSSocket` (which inherits from `net.Socket`), which exposes `socket.alpnProtocol`.
2. For plaintext HTTP (`http://`), Undici uses a plain `net.Socket`. Undici only supports HTTP/1.1 over plaintext (it does not support cleartext `h2c`).
3. HTTP/2 in Undici is only negotiated via ALPN over TLS when `allowH2: true` is enabled.

So if the socket is a plain `net.Socket` where `alpnProtocol` is `undefined` (or a TLS socket where ALPN was not used), the connection is always HTTP/1.1.

Because `alpnProtocol === 'h2'` evaluates to `false` when `alpnProtocol` is `undefined`, it safely falls back to `'1.1'` for any plaintext `net.Socket`.

david-luna (approved):

PR #3739, opened 2026-09-06
feat(instrumentation-redis): add aggregate batch spans

ilansh:

`redis.multi().exec()` and `execAsPipeline()` currently emit one span per queued command. A large batch therefore creates thousands of Redis spans for one logical database operation, amplifying SDK and collector memory usage.

The database semantic conventions model a batch as one client span with [`db.operation.batch.size`](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#batch-operations).

- adds the opt-in `aggregateMultiCommandSpans` instrumentation option; the existing per-command behavior remains the default
- emits one client span when `exec()` or `execAsPipeline()` performs the batch
- sets `db.operation.batch.size` for multi-command and empty batches, while omitting it for one command
- names homogeneous batches `MULTI <command>` / `PIPELINE <command>` and mixed batches `MULTI` / `PIPELINE`
- models a single queued command as that command without `db.operation.batch.size`
- records command errors on the aggregate span and invokes `responseHook` once per aggregate operation
- clears retained command metadata after execution to avoid retaining it on reused objects
- patches and unpatches cluster `execAsPipeline()` consistently with the regular client
- omits `db.query.text` for multi-command batches to avoid replacing span amplification with a very large concatenated attribute

The implementation uses the instrumentation's existing node-redis patches, so the option works across supported node-redis v4-v6 clients without depending on v6-only diagnostics channels.

## Testing

- package TypeScript compile
- package test suite: 44 passing, 4 pending live-cluster integration tests
- `test-all-versions`: passing for the supported `redis` version matrix (`>=2.6.0 <7`)
- ESLint, Prettier, Markdownlint, and README link checks

### DIFF
--- a/packages/instrumentation-redis/README.md
+++ b/packages/instrumentation-redis/README.md
@@ -50,6 +50,14 @@ Redis instrumentation has a few options available to choose from. You can set th
 | `dbStatementSerializer` | `DbStatementSerializer` (function)                | Redis instrumentation will serialize the command to the `db.statement` attribute using the specified function. |
 | `responseHook`          | `RedisResponseCustomAttributeFunction` (function) | Function for adding custom attributes on db response. Receives params: `span, moduleVersion, cmdName, cmdArgs` |
 | `requireParentSpan`     | `boolean`                                         | Require parent to create redis span, default when unset is false.                                              |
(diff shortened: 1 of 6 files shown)


ilansh:
@blumamir would appreciate if you can spare some time to review this

PR #3740, opened 2026-09-07
feat(winston): support native OpenTelemetry log levels and custom severity mapping (#3345)

### LINKED ISSUE #3345: [instrumentation-winston] Support for native OpenTelemetry Log Levels and SeverityNumber mapping

Currently, the `@opentelemetry/instrumentation-winston` relies on [hardcoded mapping](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/d14f8e595c5ee31c0c8919cfe24be3cff6b3c928/packages/winston-transport/src/utils.ts#L58) of log level names to SeverityNumber.

When a user configures Winston levels to match the OpenTelemetry (or other custom levels), the instrumentation fails to assign a correct SeverityNumber, resulting in `undefined` value in the log record.

```js
const logger = createLogger({
  levels: {
    fatal: 0,
    error: 1,
    warn: 2,
    info: 3,
    debug: 4,
    trace: 5
  },
  ...
});

logger.trace('Test trace'); // SeverityNumber is undefined
logger.info('Test info'); // SeverityNumber is 9 as it is defined by internal npmLevels mapping
```

The instrumentation should export a predefined constant or helper that allows users to align Winston levels with the OpenTelemetry standard easily.

```js
import { otelLogLevels } from '@opentelemetry/instrumentation-winston';

const logger = createLogger({
  levels: otelLogLevels, // Predefined map: { fatal: 0, error: 1, warn: 2, info: 3, debug: 4, trace: 5 }
  level: 'info',
  transports: [new winston.transports.Console()],
});
```

If `otelLogLevels` is used, the instrumentation should automatically map these keys to the correct `SeverityNumber`

A more flexible approach would allow developers to explicitly define the severity mapping via the configuration

```js
new WinstonInstrumentation({
  severityMapping: {
    fatal: SeverityNumber.FATAL,
    emergency: SeverityNumber.FATAL3
  }
});

// mapping could be provided by instrumentation:

import {otelSeverityMapping} from '@opentelemetry/instrumentation-winsotn';

new WinstonInstrumentation({
  severityMapping: otelSeverityMapping
});
```

via SDK

```
const sdk = new NodeSDK({
  instrumentations: [getNodeAutoInstrumentations({
    '@opentelemetry/instrumentation-winston': {
       severityMapping: {...}
    }
  })],
});
```

### Additional context

Yashisinghal285:

Fixes #3345

Currently, Winston loggers using native OpenTelemetry log levels (`fatal`, `error`, `warn`, `info`, `debug`, `trace`) or custom logging levels end up with `SeverityNumber.UNSPECIFIED` (0) in the exported log records because the transport only maps Winston's default npm levels. Additionally, `instrumentation-winston`'s `logSeverity` threshold filtering only recognizes standard npm levels.

1. **`@opentelemetry/winston-transport`**:
   - Export `otelLogLevels` and `otelSeverityMapping` for out-of-the-box support with native OpenTelemetry log levels.
   - Add `severityMapping?: Record<string, SeverityNumber>` option to `OpenTelemetryTransportV3Options` so users can provide custom level-to-severity mappings.
   - Update `getSeverityNumber()` fallback order: custom `severityMapping` -> `otelSeverityMapping` -> standard npm `severityMap` -> `DEFAULT_SEVERITY_NUMBER`.
   - Update tests and README.

2. **`@opentelemetry/instrumentation-winston`**:
   - Export `otelLogLevels` and `otelSeverityMapping`.
   - Add `severityMapping?: Record<string, SeverityNumber>` to `WinstonInstrumentationConfig` and pass it through to `OpenTelemetryTransportV3` when injected.
   - Update `_winstonLevelFromSeverity` to detect `otelLogLevels` and evaluate custom `severityMapping` when determining the threshold level to capture on the logger.
   - Update tests and README.


packages/instrumentation-winston/src/instrumentation.ts
```diff
@@ -239,10 +243,47 @@ export class WinstonInstrumentation extends InstrumentationBase<WinstonInstrumen
...
   ): string | undefined {
     if (winstonLevels) {
-      if (isNpmLevels(winstonLevels)) {
+      if (severityMapping) {
+        let matchedLevel: string | undefined;
+        let maxPriority = -Infinity;
+        for (const [levelName, levelPriority] of Object.entries(
+          winstonLevels
+        )) {
+          const mappedSeverity = severityMapping[levelName];
```
JacksonWeber: Could we use the same default mappings here that the transport uses? If I only map `critical` and `caution`, I’d expect `info` to keep its normal severity. But with `logSeverity: INFO`, this picks `caution` as the cutoff and drops `info` logs.

Yashisinghal285: Hi @JacksonWeber, good catch, that makes total sense!

I've updated `_winstonLevelFromSeverity` to use `getSeverityNumber` with the same fallback chain as the transport (`severityMapping` -> `otelSeverityMapping` -> `npmLevels` -> `sysLoglevels` -> `cliLevels`). Unmapped standard levels now retain their normal severities, so `info` is properly preserved as the cutoff.

Added a unit test covering this exact fallback scenario in commit f50a3f8.

packages/instrumentation-winston/src/instrumentation.ts
```diff
@@ -239,10 +244,47 @@ export class WinstonInstrumentation extends InstrumentationBase<WinstonInstrumen
...
+        }
+        if (matchedLevel !== undefined) {
+          return matchedLevel;
+        }
+      }
+
+      if (isOtelLevels(winstonLevels)) {
+        if (severity >= SeverityNumber.FATAL) {
+          return 'fatal';
+        } else if (severity >= SeverityNumber.ERROR) {
```
hectorhdzg, line 258: `ERROR2` threshold resolves to Winston’s `error` level, which also allows plain `ERROR` logs through. Should intermediate thresholds select the next higher available level instead?

packages/instrumentation-winston/src/instrumentation.ts
```diff
@@ -239,10 +244,47 @@ export class WinstonInstrumentation extends InstrumentationBase<WinstonInstrumen
...
+        let maxPriority = -Infinity;
+        for (const [levelName, levelPriority] of Object.entries(
+          winstonLevels
+        )) {
+          const mappedSeverity = getSeverityNumber(levelName, severityMapping);
+          if (
+            mappedSeverity !== undefined &&
+            mappedSeverity >= severity &&
+            typeof levelPriority === 'number' &&
+            levelPriority > maxPriority
```
hectorhdzg: Custom severity mappings may not follow Winston’s numeric ordering, so one Winston cutoff cannot reliably enforce them. Could the transport filter each record using its mapped OTel severity instead?

packages/instrumentation-winston/src/instrumentation.ts
```diff
@@ -239,10 +244,47 @@ export class WinstonInstrumentation extends InstrumentationBase<WinstonInstrumen
...
+          ) {
+            maxPriority = levelPriority;
+            matchedLevel = levelName;
+          }
+        }
+        if (matchedLevel !== undefined) {
+          return matchedLevel;
+        }
+      }
```
hectorhdzg: Custom mapping produces no qualifying level, this falls back to the built-in mapping. That can override the user’s mapping and export records below their configured threshold

packages/instrumentation-winston/src/otel-levels.ts
```diff
@@ -0,0 +1,71 @@
...
+  data: SeverityNumber.INFO2,
+  info: SeverityNumber.INFO,
+  debug: SeverityNumber.DEBUG,
+  prompt: SeverityNumber.TRACE4,
+  verbose: SeverityNumber.TRACE3,
+  input: SeverityNumber.TRACE2,
+  silly: SeverityNumber.TRACE,
+};
+
+export function getSeverityNumber(
```
hectorhdzg: These mappings and `getSeverityNumber()` now exist in both `instrumentation-winston` and `winston-transport`. If they drift, the instrumentation could choose one threshold while the transport exports a different severity. Can we keep a single source of truth, perhaps by using the exports from the dynamically loaded transport module?

packages/instrumentation-winston/src/instrumentation.ts
Yashisinghal285: Good catch. Updated isOtelLevels so intermediate thresholds select the next higher level (e.g. ERROR2 selects 'fatal').

packages/instrumentation-winston/src/otel-levels.ts
Yashisinghal285: Agreed. Since the transport now filters each record directly using its own mappings, I removed getSeverityNumber() and the duplicate mapping tables from instrumentation-winston.

packages/instrumentation-winston/src/instrumentation.ts
Yashisinghal285: Handled with the transport-level filtering. When a custom severityMapping is provided, we let the transport filter per-record instead of falling back to a Winston cutoff.

packages/instrumentation-winston/src/instrumentation.ts
Yashisinghal285: That's a much better approach. Added logSeverity to OpenTelemetryTransportV3 so the transport filters each record directly against its mapped OTel severity. This avoids any reliance on Winston's numeric priority ordering when custom mappings are used.

packages/instrumentation-winston/src/instrumentation.ts
```diff
@@ -189,13 +189,21 @@ export class WinstonInstrumentation extends InstrumentationBase<WinstonInstrumen
...
-                const winstonLevel = instrumentation._winstonLevelFromSeverity(
-                  config.logSeverity,
-                  args[0].levels
-                );
-                transportOptions = { level: winstonLevel };
+              const transportOptions: Record<string, any> = {};
+              if (config.severityMapping) {
+                transportOptions.severityMapping = config.severityMapping;
+              }
+              if (config.logSeverity != null) {
```
JacksonWeber, line 196: With the logger at `info` and `logSeverity: DEBUG`, adding `severityMapping: {}` now drops debug logs. Leaving the transport’s `level` unset makes it inherit the logger’s cutoff, so those records never reach the new severity filter. We need a permissive transport cutoff when an explicit OTel threshold is set.

packages/instrumentation-winston/src/instrumentation.ts
```diff
@@ -305,6 +337,18 @@ export class WinstonInstrumentation extends InstrumentationBase<WinstonInstrumen
       );
     }
 
+    function isOtelLevels(arg: any): boolean {
```
JacksonWeber, line 340: Adding `fatal` and `trace` to npm levels makes this check pass, but the OTel cutoff ignores the other levels. For example, `logSeverity: DEBUG2` selects `info` and drops both `verbose` and `http`, even though they qualify. This shortcut should only apply to the exact OTel level map.

packages/instrumentation-winston/src/instrumentation.ts
Yashisinghal285: Good point! Added a permissive cutoff fallback (`_getPermissiveLevel`) for `transportOptions.level` so records aren't blocked by the logger level before reaching the transport. Added a test for this in 9cb16eb.

packages/instrumentation-winston/src/instrumentation.ts
Yashisinghal285: Nice catch. Added `Object.keys(arg).length === 6` so `isOtelLevels` only applies to the exact OTel map. Extended maps now route through transport filtering, preserving levels like `http` and `verbose`.

Issue #3742, opened 2026-09-08
Instrumentation request: ibmmq (IBM MQ MQI client for Node.js)


Aryan16138:
### Which library do you want to instrument?

[`ibmmq`](https://www.npmjs.com/package/ibmmq) ([ibm-messaging/mq-mqi-nodejs](https://github.com/ibm-messaging/mq-mqi-nodejs)), IBM's officially maintained MQI client binding for Node.js. If you talk to IBM MQ from Node, this is the package you use. There is no supported alternative.

Right now there is no OpenTelemetry instrumentation for it, and the effect on a trace is stark. A Node service that publishes to a queue simply appears to stop. You see the HTTP span that came in, you see the database calls, and then nothing, because the MQ put produces no span at all. On the other side, the consumer's work shows up as a trace with no visible cause. The message hop, which is usually the most interesting boundary in these systems and often the one people are actually trying to debug, is the one part that is invisible.

IBM MQ tends to sit exactly where that hurts most: banking, insurance, payments, retail order flow, government systems. It is frequently the backbone connecting a modern Node service to a mainframe or to a decades-old core system, and it is precisely those handoffs that teams cannot currently see.

Some of the groundwork is already done. The library ships an OpenTelemetry-aware module, [`lib/mqiotel.js`](https://github.com/ibm-messaging/mq-mqi-nodejs/blob/master/lib/mqiotel.js), which injects and extracts `traceparent` in the RFH2 message header, so trace context already survives a trip through a queue. What is missing is the spans on either end of that propagation. The plumbing exists; nothing is producing the telemetry that flows through it.

#### The queue manager identity problem

There is a second problem here that the generic messaging attributes do not solve, and it is the part I most want to explain properly, because it is easy to miss until it bites.

In IBM MQ, a queue manager is the broker process, and applications connect to one by name. Those names are not globally unique. A queue manager called `QM1` can exist on ten different hosts, and organisations really do run several with identical names across regions, data centres, and dev/test/prod. On top of that, a default IBM MQ install creates a queue named `DEV.QUEUE.1`, so that same queue name exists on all of them too.

[...]

PR #3743, opened 2026-09-08
feat: add @opentelemetry/instrumentation-ibmmq

Aryan16138:
Requested in #3742.

There is no OpenTelemetry instrumentation for [`ibmmq`](https://www.npmjs.com/package/ibmmq), IBM's MQI client binding for Node.js. A service that publishes to a queue currently just appears to stop: you get the incoming HTTP span, the database calls, then nothing. The consumer on the other side shows up as work with no visible cause.

The library already ships [`lib/mqiotel.js`](https://github.com/ibm-messaging/mq-mqi-nodejs/blob/master/lib/mqiotel.js), which carries `traceparent` in the RFH2 header, so context survives the queue. What is missing is spans on either end of it.

See #3742 for the fuller case, including why queue manager *names* are not enough to identify a broker.

Adds `@opentelemetry/instrumentation-ibmmq` and registers it with `auto-instrumentations-node`, so it works via `--require @opentelemetry/auto-instrumentations-node/register` with no application code change.

| MQI verb | span | kind |
|---|---|---|
| `Put`, `PutSync`, `Put1`, `Put1Sync` | `send <destination>` | `PRODUCER` |
| `GetSync` | `receive <destination>` | `CLIENT` |
| each delivery of the `Get` listener callback | `process <destination>` | `CONSUMER` |

`Conn`/`Connx`/`ConnSync`/`ConnxSync`, `Open`/`OpenSync` and `Ctl` are hooked for bookkeeping and context binding only, and emit no spans. Note that `Get` registers a listener rather than fetching one message, so a single call produces many spans, and nothing is delivered until `Ctl(hConn, MQOP_START)` runs.

### Two decisions worth reviewing

**A timed-out get is not an error.** A waiting get against an idle queue returns `MQRC_NO_MSG_AVAILABLE` (2033) on every timeout, which a polling consumer hits constantly, so those spans are left `UNSET`. Treating it as an error would report every idle consumer at a 100 percent error rate. All other reason codes still produce `ERROR`.

**The destination is the resolved queue, not the requested one.** MQ writes `ResolvedQName` into the caller's descriptor during `MQOPEN`, and for an alias, clustered or dynamic reply queue that differs from what was asked for. Since `Put`/`Get` only receive the opened `MQObject`, the resolved name is captured at open time in a `WeakMap`. `Put1`/`Put1Sync` are the exception: MQ only fills it in as the call completes, so those spans open on the requested name and are corrected once it settles.

### Queue manager identifier

Optionally records `messaging.ibmmq.queue_manager.id` (`MQCA_Q_MGR_IDENTIFIER`), which unlike the queue manager name is globally unique. One `MQINQ` per connection, never per message, cached in a `WeakMap` keyed on the connection because `MQQueueManager` is `Object.seal`ed. The entry is dropped on a reconnect reason code, since IBM MQ's automatic reconnect can return on a different queue manager and a stale value would be quietly wrong.

Not in `@opentelemetry/semantic-conventions` yet, so it is **off by default**:

```js
new IbmMqInstrumentation({ emitQueueManagerId: true });
```

It also honours `OTEL_INSTRUMENTATION_IBMMQ_EMIT_QUEUE_MANAGER_ID=true`, because `auto-instrumentations-node/register` exposes no per-instrumentation config hook and a config-only switch would be unreachable for zero-code-change users. The proposal is open-telemetry/semantic-conventions#4014, unratified; happy to drop the attribute from an initial release if you would rather not carry it yet.

### Notes

Propagation is left entirely to `mqiotel.js`; no propagation code here. Unstable semconv constants are copied into `src/semconv.ts` per the [semantic-conventions guidance](https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv).

10 unit tests cover all three span shapes, destination resolution including the alias case, the 2033 rule, and the identifier cache with its reconnect invalidation. Also verified against three real IBM MQ 9.4.5 queue managers side by side: all three had a queue named `DEV.QUEUE.1`, so every span came out identically named and only the queue manager identifier distinguished them.

Supported range is `ibmmq@>=2.0.0 <3`, with a `.tav.yml` for test-all-versions.

### DIFF
--- a/packages/instrumentation-ibmmq/src/ibmmq.ts
+++ b/packages/instrumentation-ibmmq/src/ibmmq.ts
@@ -1,0 +1,536 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Span, SpanKind, SpanStatusCode, context, trace } from '@opentelemetry/api';
+import {
+  InstrumentationBase,
+  InstrumentationNodeModuleDefinition,
+  isWrapped,
+} from '@opentelemetry/instrumentation';
+import {
+  IbmMqModuleExports,
+  MQObjectLike,
+  MQODLike,
+  MQQueueManagerLike,
+  MqCallback,
+} from './internal-types';
+import {
+  ATTR_MESSAGING_DESTINATION_NAME,
+  MESSAGING_OPERATION_TYPE_VALUE_PROCESS,
+  MESSAGING_OPERATION_TYPE_VALUE_RECEIVE,
+  MESSAGING_OPERATION_TYPE_VALUE_SEND,
+} from './semconv';
+import { DEFAULT_CONFIG, IbmMqInstrumentationConfig } from './types';
+import {
+  buildMessagingAttributes,
+  destinationFromOD,
+  destinationFromObject,
+  ensureQueueManagerId,
+  invalidateOnReconnect,
+  isRoutineNoMessage,
+  recordResolvedName,
+  resolveConnectionOnConnect,
+} from './utils';
+/** @knipignore */
+import { PACKAGE_NAME, PACKAGE_VERSION } from './version';
+
+const supportedVersions = ['>=2.0.0 <3'];
+
+export class IbmMqInstrumentation extends InstrumentationBase<IbmMqInstrumentationConfig> {
+  constructor(config: IbmMqInstrumentationConfig = {}) {
+    super(PACKAGE_NAME, PACKAGE_VERSION, { ...DEFAULT_CONFIG, ...config });
+  }
+
+  override setConfig(config: IbmMqInstrumentationConfig = {}) {
+    super.setConfig({ ...DEFAULT_CONFIG, ...config });
+  }
+
+  protected init() {
+    const module = new InstrumentationNodeModuleDefinition(
+      'ibmmq',
+      supportedVersions,
+      this.patch.bind(this),
+      this.unpatch.bind(this)
+    );
+    return module;
+  }
+
+  private patch(module: any) {
+    // `ibmmq` is CommonJS-only; under `--import`/ESM the loader hands us a
+    // namespace object instead of the raw exports.
+    const moduleExports: IbmMqModuleExports =
+      module[Symbol.toStringTag] === 'Module' ? module.default : module;
+    // `_wrap`'s generic signature demands the wrapper match the exact
+    // per-property type; our patch factories are written generically
+    // against `(...args: unknown[]) => unknown` instead (as amqplib and
+    // other free-function patches in this repo do), so wrap through an
+    // untyped view of the same object.
+    const anyExports: any = moduleExports;
+
+    if (!isWrapped(anyExports.Conn)) {
+      this._wrap(
+        anyExports,
+        'Conn',
+        this.getConnectPatch.bind(this, moduleExports)
+      );
+    }
+    if (!isWrapped(anyExports.Connx)) {
+      this._wrap(
+        anyExports,
+        'Connx',
+        this.getConnectPatch.bind(this, moduleExports)
+      );
+    }
+    if (!isWrapped(anyExports.ConnSync)) {
+      this._wrap(
+        anyExports,
+        'ConnSync',
+        this.getConnectPatch.bind(this, moduleExports)
+      );
+    }
+    if (!isWrapped(anyExports.ConnxSync)) {
+      this._wrap(
+        anyExports,
+        'ConnxSync',
+        this.getConnectPatch.bind(this, moduleExports)
+      );
+    }
+
+    if (!isWrapped(anyExports.Open)) {
+      this._wrap(anyExports, 'Open', this.getOpenPatch.bind(this));
+    }
+    if (!isWrapped(anyExports.OpenSync)) {
+      this._wrap(anyExports, 'OpenSync', this.getOpenPatch.bind(this));
+    }
+
+    if (!isWrapped(anyExports.Put)) {
+      this._wrap(
+        anyExports,
+        'Put',
+        this.getObjectOperationPatch.bind(this, moduleExports, 'send')
+      );
+    }
+    if (!isWrapped(anyExports.PutSync)) {
+      this._wrap(
+        anyExports,
+        'PutSync',
+        this.getObjectOperationPatch.bind(this, moduleExports, 'send')
+      );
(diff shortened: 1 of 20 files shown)


packages/instrumentation-ibmmq/src/ibmmq.ts
```diff
@@ -0,0 +1,536 @@
...
+    moduleExports: IbmMqModuleExports,
+    original: (...args: unknown[]) => unknown
+  ) {
+    const self = this;
+    return function patchedGet(this: unknown, ...args: unknown[]) {
+      const jsObject = args[0] as MQObjectLike;
+      const lastIndex = args.length - 1;
+      const originalCb = args[lastIndex] as MqCallback<unknown>;
+
+      const wrappedCb = context.bind(
```
JacksonWeber, line 384: I think we might be creating the `process` span too late here. `ibmmq` runs `getOtelTraceAfter` before calling this callback, and it only adds the producer link to whatever span is active at that point. Our new span won’t get that link, so the consumer still looks disconnected from the producer even when the message has a valid `traceparent`.

Aryan16138: You're right, and thanks for catching it. I reproduced it against a live queue manager before changing anything: every process span came out with no inbound context, and IBM's own debug log showed No current span to update on each delivery even though traceparent was present on every message.

The cause is the ordering you point at. preJsAppCB calls getOtelTraceAfter at mqigeta.js:307 and the application callback at 309. getTraceAfter ends with trace.getSpan(context.active()) and only adds the link if something is already current, so our span does not exist yet and the extracted context is dropped. context.bind does not help, since it applies when our callback runs, which is after the hook, and preJsAppCB is not exported so there is nothing to patch in between. The synchronous path is unaffected, because the same hook runs inside GetSync while our receive span is current.

So the async path now does its own extraction and uses the result as the process span's parent. It reads traceparent and tracestate from gmo.MsgHandle with InqMp when a handle exists, and otherwise parses the RFH2 through the public mq.MQRFH2 helpers. Both carriers are needed: getTraceBefore only creates a message handle when the application asked for MQGMO_NO_PROPERTIES, or for MQGMO_PROPERTIES_AS_Q_DEF with the queue's PROPCTL at MQPROP_NONE. In the default case the properties arrive inline, so a handle-only read finds nothing. The read is synchronous inside the callback, which is required, since a deferred one returns MQRC_HCONN_ASYNC_ACTIVE.

Every process span now shares its producer's trace, with parent.id matching the send span id. A message with no traceparent produces an explicit root rather than inheriting whatever was ambient. I added tests for both carriers and for the no-context case, and confirmed the RFH2 one fails against the previous implementation.

One open question for you. The messaging conventions treat links as the default correlation mechanism and allow the message creation context as a parent only for single-message process spans. One Get callback delivers exactly one message, so I took the parent option, which is what instrumentation-kafkajs does for eachMessage. Happy to switch to a link if you would rather stay on the spec default.

packages/instrumentation-ibmmq/src/semconv.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * The IBM MQ Queue Manager Identifier (QMID) - `MQCA_Q_MGR_IDENTIFIER`,
+ * globally unique, format `<QMNAME>_YYYY-MM-DD_HH.MM.SS`. Not yet part of
+ * `@opentelemetry/semantic-conventions` (semantic-conventions PR #4014,
+ * closed/unratified). See `semantic-conventions/model/messaging/ibmmq.yaml`
+ * and `registry.yaml` in the sibling `semantic-conventions` workstream repo.
+ */
+export const ATTR_MESSAGING_IBMMQ_QUEUE_MANAGER_ID =
+  'messaging.ibmmq.queue_manager.id' as const;
```
thompson-tomo, line 45:
```suggestion
  'ibm.mq.queue_manager.id' as const;
```
As per semconv rules

Aryan16138: Thanks for flagging this, though I think messaging is an exception to that rule.

naming.md does say system-specific attributes should start with the system name, and I can see why that reads as ibmmq.queue_manager.id. But messaging is listed under "Known exceptions" in that same file, as not following the system-specific guidance yet and getting updated one-by-one. And messaging-spans.md puts it the other way as a MUST: messaging system-specific attributes MUST be defined in the corresponding messaging.{system} namespace.

Every messaging system in the registry follows that today, messaging.kafka.*, messaging.rabbitmq.*, messaging.servicebus.*, messaging.eventhubs.*, and none of them has a root-namespace variant. Databases did get migrated out, db.cassandra.* to cassandra.*, but messaging hasn't yet.

So I'd rather keep the messaging. prefix and move along with the rest of messaging when that migration happens, instead of splitting off on my own now. Happy to change it as soon as it does.

thompson-tomo: It is listed as an exception due to legacy reasons and is yet to be updated, hence i am unsure why we would add something not compliant.

> Every messaging system in the registry follows that today, messaging.kafka., messaging.rabbitmq., messaging.servicebus., messaging.eventhubs., and none of them has a root-namespace variant.

Not the case, If you take aws sns as an example https://opentelemetry.io/docs/specs/semconv/messaging/sns/ this has already been updated. SQS https://opentelemetry.io/docs/specs/semconv/messaging/sqs/ is also using the correct pattern.

> So I'd rather keep the messaging. prefix and move along with the rest of messaging when that migration happens, instead of splitting off on my own now. Happy to change it as soon as it does.


The precedence is set in aws to follow the guidance for new attributes rather than introduce attributes which will need to be renamed.

Aryan16138: Fair point on SNS and SQS, checked and you're right, they already use the root-namespace pattern (aws.sns.topic.arn, aws.sqs.queue.url), so my "none of them has done this" claim doesn't hold.

Reason I'd still keep messaging.ibmmq.queue_manager.id here isn't that argument anymore, it's that this isn't the only IBM MQ instrumentation going up right now. Same join-key attribute is also in open-telemetry-java-instrumentation#19896 and opentelemetry-dotnet-instrumentation#5404. I would like to know which direction should I proceed in. In my opinion, keeping these consistent would take priority. 

The semconv proposal for this attribute, semantic-conventions#4080, is open but not settled either way yet. Happy to revisit the spelling here once that lands.

PR #3745, opened 2026-09-08
chore: regenerate component-label-map

maryliag:
Runs scripts/gen-component-label-map.mjs to bring the labeler config back in sync with component_owners.yml and the current workspace.

- Remove instrumentation-nestjs-core, instrumentation-tedious, and sql-common from pkg-status:unmaintained
- Add pkg:instrumentation-console and pkg:instrumentation-langchain labels
- Include propagator-aws-xray under pkg:instrumentation-aws-lambda via the dependency graph.


pichlermarc (approved):
thank you!

PR #3746, opened 2026-09-09
fix(instrumentation-aws-sdk): handle Uint8Array body in InvokeModel requestPreSpanHook

larowlan:

`requestPreSpanHookInvokeModel` calls `JSON.parse(request.commandInput.body)` assuming the body is
a `string`. With AWS SDK for JavaScript v3 ≥ ~3.600 the Smithy-based serde layer serialises the
`InvokeModelCommand` body to a `Uint8Array` _before_ instrumentation hooks run. Calling
`JSON.parse` on a `Uint8Array` coerces it via `.toString()`, producing a comma-separated list of
byte values (e.g. `"123,34,105,110,..."`) instead of JSON. This causes a
`SyntaxError: Unexpected non-whitespace character after JSON at position 3` that propagates out of
the hook and crashes the `client.send()` call entirely — breaking every `InvokeModel` invocation
for users on recent AWS SDK versions.

The same byte-array-to-string pattern is already handled correctly in `parseChunk()` in the same
file using `Buffer.from(bytes).toString('utf-8')`.

- **`src/services/bedrock-runtime.ts`** — In `requestPreSpanHookInvokeModel`, decode the request
  body before parsing: if `body` is a `Uint8Array` (new AWS SDK v3 serde), convert it with
  `Buffer.from(rawBody).toString('utf-8')` first; otherwise pass it through as-is (existing
  behaviour for callers that still supply a plain `string`).

- **`test/bedrock-runtime.test.ts`** — Adds a regression test inside `describe('InvokeModel')`
  that passes a `Uint8Array` body for `amazon.titan-embed-text-v2:0` (the exact shape produced by
  the new AWS SDK serialiser) and asserts the span is created without throwing.

- **`test/mock-responses/bedrock-invokemodel-does-not-throw-when-body-is-a-uint8array-aws-sdk-v3-serialization.json`**
  — nock-back fixture for the new test case.

### DIFF
--- a/packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
+++ b/packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
@@ -167,7 +167,12 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
     }
 
     if (request.commandInput?.body) {
-      const requestBody = JSON.parse(request.commandInput.body);
+      const rawBody = request.commandInput.body;
+      const requestBody = JSON.parse(
+        rawBody instanceof Uint8Array
+          ? Buffer.from(rawBody).toString('utf-8')
+          : rawBody
+      );
       if (modelId.includes('amazon.titan')) {
         if (requestBody.textGenerationConfig?.temperature !== undefined) {
           spanAttributes[ATTR_GEN_AI_REQUEST_TEMPERATURE] =
--- a/packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts
+++ b/packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts
@@ -632,6 +632,33 @@ describe('Bedrock', () => {
         [ATTR_GEN_AI_RESPONSE_FINISH_REASONS]: ['max_tokens'],
       });
     });
+
+    it('does not throw when body is a Uint8Array AWS SDK v3 serialization', async () => {
+      const modelId = 'amazon.titan-embed-text-v2:0';
+      const nativeRequest = {
+        inputText: 'hello world',
+        dimensions: 1024,
+        normalize: true,
+      };
+      // AWS SDK v3 >= ~3.600 serializes the body to Uint8Array before instrumentation
+      // hooks run. Passing it directly reproduces the real-world shape.
+      const command = new InvokeModelCommand({
+        modelId,
+        body: new TextEncoder().encode(JSON.stringify(nativeRequest)),
+      });
+      // Must not throw SyntaxError from JSON.parse(Uint8Array)
+      await client.send(command);
+
+      const testSpans: ReadableSpan[] = getTestSpans();
+      const invokeModelSpans: ReadableSpan[] = testSpans.filter(
+        (s: ReadableSpan) => s.name === 'BedrockRuntime.InvokeModel'
+      );
+      expect(invokeModelSpans.length).toBe(1);
+      expect(invokeModelSpans[0].attributes).toMatchObject({
+        [ATTR_GEN_AI_SYSTEM]: GEN_AI_SYSTEM_VALUE_AWS_BEDROCK,
+        [ATTR_GEN_AI_REQUEST_MODEL]: modelId,
+      });
+    });
   });
 
   describe('InvokeModelWithStreams', () => {
--- a/packages/instrumentation-aws-sdk/test/mock-responses/bedrock-invokemodel-does-not-throw-when-body-is-a-uint8array-aws-sdk-v3-serialization.json
+++ b/packages/instrumentation-aws-sdk/test/mock-responses/bedrock-invokemodel-does-not-throw-when-body-is-a-uint8array-aws-sdk-v3-serialization.json
@@ -1,0 +1,30 @@
+[
+    {
+        "scope": "https://bedrock-runtime.us-east-1.amazonaws.com:443",
+        "method": "POST",
+        "path": "/model/amazon.titan-embed-text-v2:0/invoke",
+        "body": {
+            "inputText": "hello world",
+            "dimensions": 1024,
+            "normalize": true
+        },
+        "status": 200,
+        "response": {
+            "embedding": [0.1, 0.2, 0.3],
+            "inputTextTokenCount": 2
+        },
+        "rawHeaders": [
+            "Date",
+            "Mon, 10 Feb 2025 01:39:12 GMT",
+            "Content-Type",
+            "application/json",
+            "Content-Length",
+            "52",
+            "Connection",
+            "keep-alive",
+            "x-amzn-RequestId",
+            "91139412-bc01-4c17-a89f-12534a3aca0b"
+        ],
+        "responseIsBinary": false
+    }
+]


packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -167,7 +167,12 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
     }
 
     if (request.commandInput?.body) {
-      const requestBody = JSON.parse(request.commandInput.body);
+      const rawBody = request.commandInput.body;
+      const requestBody = JSON.parse(
```
jj22ee: This `JSON.parse` is still unguarded, and a throw here propagates all the way out of the caller's `client.send()` — which is the actual crash this PR is fixing, just for one input shape.

`body` is typed `BlobPayloadInputTypes`, so `Readable`, `Blob`, and `ReadableStream` are all valid and all still land in the `: rawBody` branch and throw (`JSON.parse` stringifies them to `"[object Object]"`). A body that simply isn't JSON does the same.

There's already a precedent for the right shape a few hundred lines down in `parseChunk()`, which wraps the decode+parse in `try`/`catch` and falls back to `this._diag.warn(...)`. Could we do the same here? Instrumentation should degrade to missing attributes, never break the instrumented call:

```ts
      let requestBody: any;
      try {
        const rawBody = request.commandInput.body;
        requestBody = JSON.parse(
          rawBody instanceof Uint8Array
            ? Buffer.from(rawBody).toString('utf-8')
            : rawBody
        );
      } catch (err) {
        diag.warn('Failed to parse InvokeModel request body', err);
        requestBody = {};
      }
```

(If you take this, the `if (modelId.includes(...))` blocks below work unchanged against `{}`.)

`responseHookInvokeModel` has the same unguarded-parse exposure on the response side — fine to leave for a follow-up, but worth noting.

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -167,7 +167,12 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+        rawBody instanceof Uint8Array
+          ? Buffer.from(rawBody).toString('utf-8')
+          : rawBody
```
jj22ee: The decode itself is right, and `instanceof Uint8Array` correctly picks up `Buffer` too (subclass), so both byte shapes are covered.

One thing about the framing in the PR description though: I don't think the Smithy serde explanation is what's happening. This hook runs off `command.input` in the *initialize* step of the middleware stack (`aws-sdk.ts`, `_getV3MiddlewareStackResolvePatch` → `normalizeV3Request(..., command.input, ...)`), which is the caller's raw input — serialization happens later, in the serialize step, so the hook never sees a serde-produced value.

The simpler explanation: `InvokeModelCommandInput.body` is *typed* as bytes, so passing a `Uint8Array` is the normal, documented usage, and this code only ever handled the `string` form. The existing tests in this file that pass `JSON.stringify(...)` are the off-type ones that happen to work.

Worth correcting in the description, because as written it implies the `: rawBody` branch is legacy compatibility for old SDKs — it isn't, and someone will delete it later on that basis. A plain `string` body is still accepted at runtime today.

Nit while you're here: line 462 in this same file uses `new TextDecoder().decode(...)` for the identical job. Either is fine (this package is Node-only), just inconsistent within the file — `parseChunk()` uses `Buffer` like you have, so there's no clearly-correct answer, feel free to ignore.

packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts
```diff
@@ -632,6 +632,33 @@ describe('Bedrock', () => {
...
+
+    it('does not throw when body is a Uint8Array AWS SDK v3 serialization', async () => {
+      const modelId = 'amazon.titan-embed-text-v2:0';
+      const nativeRequest = {
+        inputText: 'hello world',
+        dimensions: 1024,
+        normalize: true,
+      };
+      // AWS SDK v3 >= ~3.600 serializes the body to Uint8Array before instrumentation
+      // hooks run. Passing it directly reproduces the real-world shape.
```
jj22ee: Same note as on the source change: this comment will mislead the next reader. The hook reads `command.input` before the serialize step, so nothing in the SDK has converted anything at this point — the `Uint8Array` here is a `Uint8Array` because *the test passed one*, which is exactly what the typed `InvokeModelCommandInput.body` contract asks for.

Suggest something closer to:

```suggestion
      // `InvokeModelCommandInput.body` is typed as bytes, so a Uint8Array is the
      // normal caller-supplied shape; the hook previously assumed a string.
```

packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts
```diff
@@ -632,6 +632,33 @@ describe('Bedrock', () => {
...
+
+      const testSpans: ReadableSpan[] = getTestSpans();
+      const invokeModelSpans: ReadableSpan[] = testSpans.filter(
+        (s: ReadableSpan) => s.name === 'BedrockRuntime.InvokeModel'
+      );
+      expect(invokeModelSpans.length).toBe(1);
+      expect(invokeModelSpans[0].attributes).toMatchObject({
+        [ATTR_GEN_AI_SYSTEM]: GEN_AI_SYSTEM_VALUE_AWS_BEDROCK,
+        [ATTR_GEN_AI_REQUEST_MODEL]: modelId,
+      });
```
jj22ee: These assertions don't actually cover the fix. `ATTR_GEN_AI_SYSTEM` is a constant and `ATTR_GEN_AI_REQUEST_MODEL` comes from `commandInput.modelId` — both are set before the body is touched, so they'd pass even if the decode produced garbage. And `amazon.titan-embed-text-v2:0` bodies have no `textGenerationConfig`, so no attribute is derived from the decoded body at all. The test only proves "didn't throw."

A body that yields attributes would pin down that the bytes decoded to usable JSON, not just that parsing survived — e.g. `amazon.titan-text-express-v1` with `{ inputText, textGenerationConfig: { temperature, topP, maxTokenCount } }`, then assert `ATTR_GEN_AI_REQUEST_TEMPERATURE` / `TOP_P` / `MAX_TOKENS` alongside what you have. That makes it a real regression test for the decode rather than for the absence of an exception. (Needs a re-recorded fixture, sorry.)

packages/instrumentation-aws-sdk/test/mock-responses/bedrock-invokemodel-does-not-throw-when-body-is-a-uint8array-aws-sdk-v3-serialization.json
```diff
@@ -0,0 +1,30 @@
...
+        "path": "/model/amazon.titan-embed-text-v2:0/invoke",
+        "body": {
+            "inputText": "hello world",
+            "dimensions": 1024,
+            "normalize": true
+        },
+        "status": 200,
+        "response": {
+            "embedding": [0.1, 0.2, 0.3],
+            "inputTextTokenCount": 2
```
jj22ee: This fixture looks hand-written rather than recorded — the `Date` and `x-amzn-RequestId` are the titan fixture's values with a couple of digits changed, and `Content-Length: 52` doesn't match the 50-byte response body.

The header comment at the top of the test file says these are recorded via `NOCK_BACK_MODE` against the real API, and the payload here isn't a real `titan-embed-text-v2:0` response shape (a real one returns a 1024-float `embedding`, plus `embeddingsByType` on v2). None of this breaks the test as written, since nock replays whatever is here, but it does mean the fixture can't be regenerated or trusted as a reference. Could you re-record it with `NOCK_BACK_MODE=record`?

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
jj22ee: Could wrap this in `try`/`catch` with a `diag.warn` fallback, the way `parseChunk()` already does, so a non-JSON body — or a `Readable`/`Blob` body, both valid `BlobPayloadInputTypes` — doesn't throw out of the caller's `client.send()`. Same crash this PR is fixing, just for the input shapes that aren't `Uint8Array`.

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
jj22ee: Decode looks right, and `instanceof Uint8Array` covers `Buffer` too. Minor: this hook runs on `command.input` before the serialize step, so the bytes aren't SDK serde output — `body` is just typed as bytes — worth correcting in the description so the `string` branch doesn't get deleted later as legacy compat.

packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts
jj22ee: Could frame this as "`body` is typed as bytes, so a `Uint8Array` is the normal caller-supplied shape" instead of attributing it to SDK serialization, since the hook reads `command.input` pre-serialize.

packages/instrumentation-aws-sdk/test/bedrock-runtime.test.ts
jj22ee: Both of these are set before the body is read, and titan-embed bodies yield no body-derived attributes, so this passes even if the decode produces garbage. Could use `amazon.titan-text-express-v1` with a `textGenerationConfig` and assert temperature/topP instead, so the test actually covers the decode rather than just the absence of a throw.

packages/instrumentation-aws-sdk/test/mock-responses/bedrock-invokemodel-does-not-throw-when-body-is-a-uint8array-aws-sdk-v3-serialization.json
jj22ee: This looks hand-written rather than recorded — `x-amzn-RequestId` and `Date` are the titan fixture's with digits changed, `Content-Length: 52` vs a 50-byte body, and not a real titan-embed-v2 response shape. Could re-record with `NOCK_BACK_MODE=record` so it stays regenerable.

packages/instrumentation-aws-sdk/src/services/bedrock-runtime.ts
```diff
@@ -167,7 +167,12 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+      );
```
jj22ee: Could wrap this in `try`/`catch` with a `diag.warn` fallback, the way `parseChunk()` already does, so a non-JSON body — or a `Readable`/`Blob` body, both valid `BlobPayloadInputTypes` — doesn't throw out of the caller's `client.send()`. Same crash this PR is fixing, just for the input shapes that aren't `Uint8Array`.

```suggestion
      let requestBody: any;
      try {
        const rawBody = request.commandInput.body;
        requestBody = JSON.parse(
          rawBody instanceof Uint8Array
            ? Buffer.from(rawBody).toString('utf-8')
            : rawBody
        );
      } catch (err) {
        diag.warn('Failed to parse InvokeModel request body', err);
        requestBody = {};
      }
```

PR #3753, opened 2026-09-11
fix(instrumentation-restify): end callback spans on next

### LINKED ISSUE #1988: [restify] Span ending prematurely with non-promise async handlers

`@opentelemetry/api@1.8.0`

`v16` - `v21`

### What did you do?

Given the following handler:

```js
server.get('/my_handler', (req, res, next) => {
  setTimeout(() => {
    // Currently, will output `true`
    console.log('span ended?', api.trace.getActiveSpan()._ended);

    res.end();
    next();
  }, 500);
});
```

`false` should be emitted on the console, given that `next` has not yet been called and therefore the span should still be active.

### What did you see instead?

`true`, indicating that the span is no longer active. In more handlers, this materialises as spans not being correctly nested, and events not emitting due to no span being active at that time.

### Additional context

It appears like the problematic point in the instrumentation is:

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-restify/src/instrumentation.ts#L248-L253

The `handler` is called with `patchedNext`. The handler immediately resolves (no promises or synchronous code in the handler), leading to `span.end()` executing on line 252. Once the `setTimeout` completes, we call the provided `next` function as per Restify's [documentation (see 'Using `next()`') header](http://restify.com/docs/home/) to continue the chain.

Similar to #1647, this causes the `You can only call end() on a span once.` error.

If I understand correctly, the `span.end()` call should be removed from line 252, and the responsibility of ending the span is differed to the patched next in order to be inline with Restify's handler convention. I would be more than happy to contribute this PR myself, just wanted some insight/clarification on whether this is actually the problem/solution. Specifically, I'm unsure how this may impact the context either...

---

On a side note, I'm wondering if this could be related to the nesting of middleware spans, as seen in the screenshot below.

image

Philipp-Perepelitsky:
Assisted-by: ChatGPT 5.6 Sol

Fixes #1988.

Callback-style Restify handlers can return before their asynchronous work is finished and call next() later. The instrumentation was ending the span as soon as the handler function returned, which made those spans too short.

- This change keeps callback-style handler spans open until next() is called, while using a shared endSpan() helper so Promise, error, and callback completion paths only end the span once.

- I also updated the shared test handler to call next() after sending the response, matching Restify's normal handler contract, and added a regression test for delayed next().

Validation:

- restify.test.ts: 12 passing
- npm run compile: passed
- Prettier and git diff --check: passed

### DIFF
--- a/packages/instrumentation-restify/src/instrumentation.ts
+++ b/packages/instrumentation-restify/src/instrumentation.ts
@@ -190,21 +190,42 @@ export class RestifyInstrumentation extends InstrumentationBase<RestifyInstrumen
           );
         }
 
+        let spanEnded = false;
+
+        const endSpan = () => {
+          if (!spanEnded) {
+            spanEnded = true;
+            span.end();
+          }
+        };
+
         const patchedNext = (err?: any) => {
-          span.end();
+          endSpan();
           next(err);
         };
-        patchedNext.ifError = next.ifError;
+
+        const nextIfError = next.ifError;
+
+        if (nextIfError) {
+          patchedNext.ifError = (err?: any) => {
+            if (err) {
+              endSpan();
+            }
+            return nextIfError.call(patchedNext, err);
+          };
+        } else {
+          patchedNext.ifError = nextIfError;
+        }
 
         const wrapPromise = (promise: Promise<unknown>) => {
           return promise
             .then(value => {
-              span.end();
+              endSpan();
               return value;
             })
             .catch(err => {
               span.recordException(err);
-              span.end();
+              endSpan();
               throw err;
             });
         };
@@ -221,11 +242,10 @@ export class RestifyInstrumentation extends InstrumentationBase<RestifyInstrumen
               if (isPromise(result)) {
                 return wrapPromise(result);
               }
-              span.end();
               return result;
             } catch (err: any) {
               span.recordException(err);
-              span.end();
+              endSpan();
               throw err;
             }
           },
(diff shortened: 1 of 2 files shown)


packages/instrumentation-restify/src/instrumentation.ts
```diff
@@ -190,21 +190,30 @@ export class RestifyInstrumentation extends InstrumentationBase<RestifyInstrumen
...
+            span.end();
+          }
+        };
+
         const patchedNext = (err?: any) => {
-          span.end();
+          endSpan();
           next(err);
         };
         patchedNext.ifError = next.ifError;
```
JacksonWeber: Could we also wrap `next.ifError`? In Restify 6 with `handleUncaughtExceptions: true`, an async `next.ifError(err)` makes Restify call the original `next`, bypassing `patchedNext` and our synchronous `catch`.

Philipp-Perepelitsky: Thank you for your comment. I addressed it in the following way:

I reproduced it with Restify 6.4.0 on Node 16 and added a regression test for an async `next.ifError(err)` call. Before the change, the test failed because the span was never ended.

I updated the wrapper so an actual error passed to `next.ifError` ends the span before delegating to Restify's original helper. I also verified the full test suite passes with both Restify 6.4.0 and Restify 11.1.0.

PR #3756, opened 2026-09-11
fix(resource-detector-gcp): add gcp.project_id attribute

torsrex:

The `gcpDetector` does not set the `gcp.project_id` resource attribute, despite the [OTLP migration guide](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/MIGRATION.md#resource-detection-recommended-for-all-signals) claiming it does. This causes 400 (`INVALID_ARGUMENT`) errors when exporting to `telemetry.googleapis.com`, which requires `gcp.project_id` to route telemetry data to the correct Google Cloud project.

The [Google Cloud Trace OTLP migration docs](https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints#node.js) explicitly instruct users to set this attribute:

> Add to the `OTEL_RESOURCE_ATTRIBUTES` environment variable the key-value pair that specifies your project. For the key, use `gcp.project_id`. For the value, use the ID of your Google Cloud project.
>
> Example: `export OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=PROJECT_ID"`

Users following the migration guide who install this detector as the recommended way to populate resource attributes are silently missing `gcp.project_id`, and must work around it manually.

Adds `gcp.project_id` as a new resource attribute populated from the GCP Metadata `project-id` field. Uses the same source and follows the same pattern as the existing `cloud.account.id`. The new attribute is added to `semconv.ts`, included in `ATTRIBUTE_NAMES`, and set in `makeResource()` alongside `cloud.account.id`.

### DIFF
--- a/packages/resource-detector-gcp/src/detectors/GcpDetector.ts
+++ b/packages/resource-detector-gcp/src/detectors/GcpDetector.ts
@@ -14,6 +14,7 @@ import {
   CLOUD_PLATFORM_VALUE_GCP_KUBERNETES_ENGINE,
   CLOUD_PROVIDER_VALUE_GCP,
   ATTR_CLOUD_ACCOUNT_ID,
+  ATTR_GCP_PROJECT_ID,
   ATTR_CLOUD_AVAILABILITY_ZONE,
   ATTR_CLOUD_PLATFORM,
   ATTR_CLOUD_PROVIDER,
@@ -52,6 +53,7 @@ const ATTRIBUTE_NAMES = [
   ATTR_HOST_NAME,
   ATTR_CLOUD_PROVIDER,
   ATTR_CLOUD_ACCOUNT_ID,
+  ATTR_GCP_PROJECT_ID,
   ATTR_FAAS_NAME,
   ATTR_FAAS_VERSION,
   ATTR_FAAS_INSTANCE,
@@ -188,6 +190,7 @@ async function makeResource(attrs: GcpResourceAttributes): Promise<Resource> {
   return resourceFromAttributes({
     [ATTR_CLOUD_PROVIDER]: CLOUD_PROVIDER_VALUE_GCP,
     [ATTR_CLOUD_ACCOUNT_ID]: project,
+    [ATTR_GCP_PROJECT_ID]: project,
     ...attrs,
   } satisfies GcpResourceAttributes);
 }
--- a/packages/resource-detector-gcp/src/semconv.ts
+++ b/packages/resource-detector-gcp/src/semconv.ts
@@ -20,6 +20,16 @@
 export const ATTR_CLOUD_ACCOUNT_ID = 'cloud.account.id' as const;
 
 /**
+ * The Google Cloud project ID.
+ *
+ * @example my-project
+ * @example 233510669999
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_GCP_PROJECT_ID = 'gcp.project_id' as const;
+
+/**
  * Cloud regions often have multiple, isolated locations known as zones to increase availability. Availability zone represents the zone where the resource is running.
  *
  * @example us-east-1c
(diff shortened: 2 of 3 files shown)


pichlermarc:
I'll let the component owners handle this (they're mostly Google employees to my knowledge), but wanted to note that this attribute does not seem to be part of semantic conventions, and this is why it's not included. 

It seems to me that the best way to get this included would be to have the [`cloud.resource_id`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-attributes) detected instead, which should include the project ID in the string.

Not sure what Google Cloud does with it though.

packages/resource-detector-gcp/src/semconv.ts
```diff
@@ -19,6 +19,16 @@
  */
 export const ATTR_CLOUD_ACCOUNT_ID = 'cloud.account.id' as const;
 
+/**
+ * The Google Cloud project ID.
+ *
+ * @example my-project
+ * @example 233510669999
+ *
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
```
dashpole: Since `gcp.project_id` is not part of `@opentelemetry/semantic-conventions`, could you update the docstring so it doesn't reference `@opentelemetry/semantic-conventions`? Maybe note that it is required for Google Cloud OTLP ingestion endpoints instead:
```suggestion
 * @note Required by Google Cloud OTLP ingestion endpoints (telemetry.googleapis.com).
```

packages/resource-detector-gcp/src/detectors/GcpDetector.ts
```diff
@@ -52,6 +53,7 @@ const ATTRIBUTE_NAMES = [
   ATTR_HOST_NAME,
   ATTR_CLOUD_PROVIDER,
   ATTR_CLOUD_ACCOUNT_ID,
+  ATTR_GCP_PROJECT_ID,
```
dashpole: Could you also update the attribute table in `README.md` to document `gcp.project_id` alongside `cloud.account.id`?

dashpole:
The gcpDetector detector should not detect gcp.project_id. You should set it using environment variables as described in the migration guide. If the migration guide doesn't work, please start with an issue in opentelemetry-operations-js so we can fix it.

cc @psx95

torsrex:
> The gcpDetector detector should not detect gcp.project_id. You should set it using environment variables as described in the migration guide. If the migration guide doesn't work, please start with an issue in opentelemetry-operations-js so we can fix it.
> 
> cc @psx95

Thank you for the clarification. I think there may be an inconsistency between what the migration guide says and the current detector behavior.

The migration guide currently says:

> "(...) installing the upstream GCP Resource Detector package (`@opentelemetry/resource-detector-gcp`) automatically populates Google Cloud resource attributes (such as `gcp.project_id`, `cloud.account.id`, `host.id`, `k8s.pod.name`, etc.) for your OpenTelemetry SDK."

It then shows `gcpDetector` being passed to `resourceDetectors`.

The current `gcpDetector`, however, does not populate `gcp.project_id`. It only populates `cloud.account.id` from the same GCP metadata source. So users following the documented resource-detection approach do not get `gcp.project_id` automatically.

I agree that `OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=..."` can be used to explicitly set the attribute, but this seems inconsistent with the guide's statement that `gcp.project_id` is automatically populated by the GCP resource detector.

I'm happy to open an issue in `opentelemetry-operations-js` to clarify whether this is a documentation or implementation issue, and close this PR until that's clarified.

psx95:
Yup, thanks for bringing this up @torsrex, that is indeed a documentation issue. I have a PR up for addressing this - https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/959

torsrex:
Sounds good, thanks for the updated documentation and for the clarification! Will close this and use the environment variable for populating `gcp.project_id`.

PR #3757, opened 2026-09-12
feat(genai-util): add base invocation lifecycle and TelemetryHandler

psx95:

Re: #3681

#### Implemented so far in genai-utils:
So far, `@opentelemetry/genai-util` only provides low-level building blocks — types, semantic-convention constants, and histogram factories. The lifecycle logic that turns those primitives into spec-compliant telemetry is left to each GenAI instrumentation to write itself.

#### The current problem:
So every instrumentation (OpenAI, Bedrock, Vertex, ...) re-implements the same plumbing: setting up a tracer and meter with the right schema URL, starting the span, propagating context across await boundaries and streams, recording duration, token-usage and streaming metrics, resolving content capture, mapping errors to error.type and span status, firing completion hooks, and ending the span exactly once on every path. Duplicating this is easy to get subtly wrong, and GenAI telemetry drifts apart between packages.

#### What this PR adds: 
This PR adds the missing shared lifecycle layer, so authors can construct an invocation, set data on it, and call stop()/fail() — with consistent spans, metrics, content-capture rules, and hook semantics handled for them.

- **Base Invocation (`src/invocations/base.ts`)**: Added `BaseInvocation` abstract class managing span lifecycle (`stop()`, `fail()`), error classification (`error.type`), exception recording, custom attribute updates, duration tracking, and extension hooks (`_recordMetrics`, `_emitContentEvents`, `_runCompletionHook`).
- **Completion Hook Manager (`src/completion-hook.ts`)**: Added `CompletionHookManager` providing registration, unregistration, and error-isolated execution of asynchronous completion hooks.
- **TelemetryHandler Foundation (`src/handler.ts`)**: Added `TelemetryHandler` managing tracer/meter lifecycles, initializing standard GenAI histograms (duration, token usage, TTFT), resolving content capture settings, and delegating metric recordings.
- **Unit Tests**: Added test suites covering base invocation lifecycle (`test/invocations/base.test.ts`), completion hook execution (`test/completion-hook.test.ts`), and handler configuration and delegation (`test/handler.test.ts`).

## Important Info
 - This is the **third PR** in a series that attempt to breakdown PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677. 
 - Based off: #3712


packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+   */
+  attributes?: Attributes;
+  /**
+   * Context used as the parent of this invocation.
+   *
+   * Defaults to the currently active context, so an invocation created inside
+   * {@link BaseInvocation.withContext} of another invocation is automatically nested
+   * under it.
+   */
+  parentContext?: Context;
```
aabmass: nit I would call it just `context` which is what the trace API uses

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+ */
+export interface BaseInvocationOptions {
+  /**
+   * Kind of the span created for this invocation.
+   *
+   * Defaults to {@link SpanKind.CLIENT}, which is the kind used by GenAI operations
+   * that call out to a model provider. Subclasses representing in-process operations
+   * (e.g. tool or workflow execution) should pass {@link SpanKind.INTERNAL}.
+   */
+  kind?: SpanKind;
```
aabmass: I don't understand why this is configurable, it seems like a property of the invocation class. I don't remember what we do in python though

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+ *   const response = await invocation.withContext(async () => {
+ *     const response = await client.chat(request);
+ *     invocation.setAttribute('custom.attr', 'value');
+ *     return response;
+ *   });
+ *   invocation.stop();
+ *   return response;
+ * } catch (error) {
+ *   invocation.fail(error);
+ *   throw error;
```
aabmass, line 110: Does the python impl handle exceptions within the context manager properly or does each instrumentation need to write this lifecycle boilerplate?

Since JS does not allow catching exceptions by pattern matching types (like in java/python), maybe the `withContext()` function can require a mandatory handler to normalize the exceptions

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+    );
+    this._context = trace.setSpan(parentContext, this._span);
+  }
+
+  /**
+   * Return the underlying OpenTelemetry Span.
+   *
+   * The span is owned by this invocation: use {@link stop} or {@link fail} to end it.
+   */
+  public getSpan(): Span {
```
aabmass: was there a use for this being public?

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * Return the OpenTelemetry Context holding this invocation's span.
+   *
+   * Pass it to `context.with()` or `context.bind()` to make the invocation the parent
+   * of downstream work that cannot be wrapped by {@link withContext}.
+   */
+  public getContext(): Context {
+    return this._context;
+  }
```
aabmass: Do we need this or they can use the `withContext()` function?

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+      this._span.end(endHr);
+    }
+  }
+
+  /**
+   * Complete the invocation with an error and end the underlying span.
+   *
+   * Ending an already ended invocation is a no-op.
+   */
+  public fail(error: Error | string | unknown, endTime?: TimeInput): void {
```
aabmass: can we be a bit more strict about accepted types for `error`? I believe this gets widened to `unknown`.

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+        : typeof error === 'string'
+          ? error
+          : String(error);
+
+    const errorType = getErrorType(error);
+    this._span.setAttribute(ATTR_ERROR_TYPE, errorType);
+
+    const errorObj = error instanceof Error ? error : new Error(errorMessage);
+
+    if (error instanceof Error) {
```
aabmass: nit this is the third time this same instanceof check happens

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+   * NOTE: Currently a no-op placeholder. Will be implemented with the OpenTelemetry Logs & Events API
+   * once `@opentelemetry/api-logs` and EventLogger reach stability in OpenTelemetry JavaScript.
+   */
+  protected _emitContentEvents(_endTime?: HrTime): void {}
+
+  /**
+   * Hook for subclasses to execute registered completion hooks on stop/fail.
+   */
+  protected _runCompletionHook(_durationSec: number, _error?: Error): void {}
+}
```
aabmass, line 365: Should these be abstract? None of the functions in this class are abstract

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+   *
+   * NOTE: Currently a no-op placeholder. Will be implemented with the OpenTelemetry Logs & Events API
+   * once `@opentelemetry/api-logs` and EventLogger reach stability in OpenTelemetry JavaScript.
+   */
+  protected _emitContentEvents(_endTime?: HrTime): void {}
+
+  /**
+   * Hook for subclasses to execute registered completion hooks on stop/fail.
+   */
+  protected _runCompletionHook(_durationSec: number, _error?: Error): void {}
```
aabmass: Is this the same completion hook as the spec documents for processing inference input/outputs?

packages/genai-util/src/handler.ts
```diff
@@ -0,0 +1,303 @@
...
+    return this._meter;
+  }
+
+  /**
+   * Register a completion hook.
+   */
+  public addCompletionHook(hook: CompletionHook): this {
+    this._hookManager.addHook(hook);
+    return this;
+  }
```
aabmass: Does we need this or it's sufficient to just accept hooks in the constructor?

packages/genai-util/src/handler.ts
```diff
@@ -0,0 +1,303 @@
...
+   */
+  public recordOperationDuration(
+    durationSeconds: number,
+    attributes?: Attributes,
+    context?: Context
+  ): void {
+    if (
+      !this._operationDurationHistogram ||
+      durationSeconds < 0 ||
+      !isFinite(durationSeconds)
```
aabmass: imo we don't need this validation since the metrics SDK handles it (and also handles diag logging info)

packages/genai-util/src/handler.ts
```diff
@@ -0,0 +1,303 @@
...
+export class TelemetryHandler {
+  private _tracer: Tracer;
+  private _meter: Meter;
+  private _diag: DiagLogger;
+  private _contentCaptureMode: ContentCaptureMode;
+  private readonly _hookManager: CompletionHookManager;
+  private _operationDurationHistogram?: Histogram;
+  private _tokenUsageHistogram?: Histogram;
+  private _timeToFirstChunkHistogram?: Histogram;
+  private _timePerOutputChunkHistogram?: Histogram;
```
aabmass: Why are these optional?

packages/genai-util/src/invocations/base.ts
aabmass: Discussed offline, for non streaming we could add it later.

packages/genai-util/src/invocations/base.ts
psx95: This is the base invocation class, when concrete implementations are used (which instrumentation authors would use), we could make it non-configurable in those implementations.

packages/genai-util/src/handler.ts
```diff
@@ -0,0 +1,303 @@
...
+
+  constructor(options: TelemetryHandlerOptions = {}) {
+    const tracerProvider = options.tracerProvider ?? trace.getTracerProvider();
+    this._tracer = tracerProvider.getTracer(PACKAGE_NAME, PACKAGE_VERSION, {
```
lmolkova: `TelemetryHandler` currently hardcodes `@opentelemetry/genai-util` as the instrumentation scope name and version when creating the `Tracer` and `Meter`.

Traces and metrics should be attributed to the instrumentation library using the handler (e.g. `@opentelemetry/instrumentation-openai`), not the utility library itself. Could we allow passing `instrumentationScopeName` and `instrumentationScopeVersion` in `TelemetryHandlerOptions` (falling back to the util package name/version if not provided)?

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * Return whether this invocation has already ended.
+   */
+  public isEnded(): boolean {
+    return this._isEnded;
+  }
+
+  /**
```
lmolkova, line 172: nit: Expose a `shouldCaptureContent` getter on `BaseInvocation` that delegates to `this._handler.shouldCaptureContent()` so subclasses and instrumentation authors do not have to reach into `_handler`.

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * Options shared by all GenAI invocations.
+ *
+ * @experimental This interface is experimental and subject to change.
+ */
+export interface BaseInvocationOptions {
+  /**
+   * Kind of the span created for this invocation.
+   *
+   * Defaults to {@link SpanKind.CLIENT}, which is the kind used by GenAI operations
```
lmolkova: do we need to default? I think we shouldn't - it will be just source of errors and does not save much typing.

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+  public isEnded(): boolean {
+    return this._isEnded;
+  }
+
+  /**
+   * Set a custom span attribute.
+   */
+  public setAttribute(key: string, value: AttributeValue): this {
+    this._customAttributes[key] = value;
+    this._span.setAttribute(key, value);
```
lmolkova, line 179: this would be misleading for metrics - some attributes come to spans and metrics, some only to spans (high cardinality). How would we solve it?

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+    this._isEnded = true;
+
+    const endHr = endTime != null ? timeInputToHrTime(endTime) : hrTime();
+    const durationSec = hrTimeToSeconds(hrTimeDuration(this._startTime, endHr));
+
+    this._recordMetrics(durationSec);
+    this._emitContentEvents(endHr);
+
+    try {
+      this._runCompletionHook(durationSec);
```
lmolkova: ideally hook should be able to modify event. Is it possible to prepare event, pass it to the hook and then emit the event?

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+          ? error
+          : String(error);
+
+    const errorType = getErrorType(error);
+    this._span.setAttribute(ATTR_ERROR_TYPE, errorType);
+
+    const errorObj = error instanceof Error ? error : new Error(errorMessage);
+
+    if (error instanceof Error) {
+      this._span.recordException(error);
```
lmolkova: this is legacy, we do log-based exceptions - https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-exceptions.md

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    this._emitContentEvents(endHr);
```
lmolkova: why plural?

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    const errorMessage =
+      error instanceof Error
+        ? error.message
+        : typeof error === 'string'
+          ? error
+          : String(error);
+
+    const errorType = getErrorType(error);
+    this._span.setAttribute(ATTR_ERROR_TYPE, errorType);
```
lmolkova: error type is reported on metrics too

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    this._emitContentEvents(endHr);
+
+    this._span.setStatus({
+      code: SpanStatusCode.ERROR,
+      message: errorMessage,
+    });
+
+    try {
+      this._runCompletionHook(durationSec, errorObj);
```
lmolkova: why error object is passed to the hook?

also same on the event - we should let hook modify the event before emission

packages/genai-util/src/invocations/base.ts
psx95: Its currently only facilitating testing, but since this is a getter, I think its ok to let it be public. 
I could rename it to `_getSpan()` to signal its only for testing.

packages/genai-util/src/invocations/base.ts
psx95: Keeping the class abstract will prevent the users from directly instantiating the `BaseInvocation` which is the desired behavior. 

I could keep them abstract, but using a no-op implementation instead of forcing every downstream invocation to implement a method helps preventing breaking changes in case this class evolves. Is there a benefit of explicitly declaring the methods abstract? 

During testing, BaseInvocation has been extended to create `CustomInvocation`s that facilitate testing and only certain methods that need to be tested have been overridden. If we make these methods abstract, all abstract would need to be implemented.

packages/genai-util/src/invocations/base.ts
psx95: @aabmass IIUC, we can resolve this comment for now and add a convenience method exclusively for non-streaming responses separately?

packages/genai-util/src/invocations/base.ts
psx95: I think `Error | string` here is decorative and can be removed, but I think we would need to use `unknown` - `getErrorType` already handles `unknown`s. 

Also, `catch` gives an `unknown` so if we don't handle `unknown` here, then the downstream users would have to manually ensure that an `Error` is being passed when attempting to `fail` the invocation. 

Let me know if this makes sense. FYI: I made this change in [f6ffc73](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/f6ffc732a028dea26a8b1f93675698241be84cae).

packages/genai-util/src/invocations/base.ts
psx95: Will refactor to preserve the result of the `instanceof` check and re-use it. Addressed in [f6ffc73](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/f6ffc732a028dea26a8b1f93675698241be84cae).

packages/genai-util/src/invocations/base.ts
psx95: Sounds good, I will make it non-optional. Addressed in [2dcf4ed](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/2dcf4edb741572ed2e2f856ba6b75ebedeec0ad6).

packages/genai-util/src/handler.ts
psx95: Ah, right - good point. I missed updating it. It should be passed in the options.

psx95: Addressed in [78490e1](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/78490e1968cb01f6bfb5d146fa2d755c29f1e097).

packages/genai-util/src/invocations/base.ts
aabmass: I see as long as it's not exposed to instrumentations. I don't have a strong feeling on this vs an abstract protected template method.

packages/genai-util/src/invocations/base.ts
psx95: Addressed in [f6ffc73](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/f6ffc732a028dea26a8b1f93675698241be84cae).

packages/genai-util/src/invocations/base.ts
aabmass: For testing, I believe it's common to do `foo['_privateProp']` e.g. https://github.com/open-telemetry/opentelemetry-js/blob/e98b904aee19e0c61c7a7c6ec3fd7b4fba7dcf72/packages/sdk-trace/src/TracerProvider.ts#L101

If we prefer to keep the span totally encapsulated maybe lets just remove this for now?

packages/genai-util/src/invocations/base.ts
psx95: I can avoid exporting the class in the next PR which will contain some concrete implementations.

packages/genai-util/src/invocations/base.ts
aabmass: Makes sense, but I'm wondering if we can force the caller to process it into a known type. In python, we accept general exceptions, but provide [Error dataclass](https://github.com/open-telemetry/opentelemetry-python-genai/blob/60c59fa6147b8828be4cd03b8b4a584165408667/util/opentelemetry-util-genai/src/opentelemetry/util/genai/types.py#L340-L349) and I think that is the intention. The idea is that instrumentations know how to classify their errors correctly and util will not.

@lmolkova hopefully I got that right.

packages/genai-util/src/invocations/base.ts
aabmass: It will mess up the type type refinement probably, you might need to restructure the branching

packages/genai-util/src/invocations/base.ts
psx95: Two things come to mind: 
1. Rename this method to `setSpanAttributes` - to make the intention of the method clear.
2. Currently, the invocations call `_recordMetrics` automatically on stop/fail. We don't allow users to pass custom attributes to metrics as of now, but we could add that - which is how [python has support for this](https://github.com/open-telemetry/opentelemetry-python-genai/blob/60c59fa6147b8828be4cd03b8b4a584165408667/util/opentelemetry-util-genai/src/opentelemetry/util/genai/_invocation.py#L74).

packages/genai-util/src/invocations/base.ts
aabmass: Gotcha, I think it's just a question of whether or not impls should be forced to implement these methods or not. For example I believe all invocations types have a relevant metric so typescript should probably show an error if they need to implement it

Feel free to resolve if it's not the case

packages/genai-util/src/invocations/base.ts
aabmass: @psx95 we discussed that the hooks in this util are more general than what we have in semconv and python's util-genai. Do you plan to remove these from base invocation and have it just on inference invocation with the specific API matching python?

psx95: @aabmass, yes - I plan on doing that. The hooks here are doing more than what's specified by the semconv which is leading to confusion.

Commit [3eab591](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/3eab591e1fc4a82a6266a9a9f95318a46d04ee2a) removed the current implementation of completion hook manager.

packages/genai-util/src/invocations/base.ts
psx95: The OpenTelemetry JS Logs API Is still experimental so we are not importing that to emit log events. So IIUC, this is ok for now?

packages/genai-util/src/invocations/base.ts
psx95: Sure, I added this convenience method in [7402b07](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/7402b072c469282a4964b806c456c857b3d44a83).

packages/genai-util/src/invocations/base.ts
psx95: Most likely they would, but its important to know that if this class is being extended by an instrumentation author, it would be for specific/custom use-cases - I think in that case having the flexibility would be better. For concrete implementations that make it into this library - accidentally missing out on implementing functions would be difficult/unlikely.

I updated to make `_recordMetrics` as abstract in [3e95331](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/3e95331e864c52e238fc5d7d99d330978e0073a8). I left `_emitContentEvent` as no-op since it is actually a placeholder and currently no invocations are expected to implement until logs api is stable.

packages/genai-util/src/handler.ts
psx95: I think it was mostly because the requirement level of the metrics tracked by these histograms in the spec was "recommended" instead of "required" - so I think it was an attempt to model that. 
But looking at it again, I could remove `?` because `_initMetrics` constructs the histograms unconditionally later on anyways - so they're always present at runtime. 

Addressed in [b725b82](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/b725b821aa471f2981cd82c04226f030a610cf2e).

packages/genai-util/src/invocations/base.ts
psx95: I was not aware of this pattern you mentioned. Span is a protected member and I noticed that the tests subclass `BaseInvocation` anyways - so it'll not be a problem.  

I don't have a strong justification to keep this method, so I removed it in [a354eb8](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/a354eb83afbe20422380a2a30f2ca9194761d1e4).

packages/genai-util/src/invocations/base.ts
lmolkova: forcing special `Error` type has pros and cons, but taking arbitrary string as an `error.type` is error prone. If someone wants to pass an error message without error type, what would they do? Naive implementation would pass it via `error` param and it'd explode metric cardinality. 

In python we take either exception or an object that has type and description, at least there is a bit more structure to it. There is also a 'resolver' thing that converts exception to error, but it's something that can be added in the future.

TL;DR: don't accept pure string, allow to pass exception instances (we'll record exception stack traces at some point) and structured errors (type + message). Fallback to [`_OTHER`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/error.md `error.type` when you know error has happened, but don't have low cardinality code.

packages/genai-util/src/invocations/base.ts
psx95: Took a pass at this in [eb48dd1](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/eb48dd185be7008eaeec250852bfbc4eb5433d6a) - which basically adds support for metric attributes. 

Let me know your thoughts on the approach in the linked commit,  @lmolkova!

packages/genai-util/src/invocations/base.ts
psx95: This gets addressed in [eb48dd1](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/eb48dd185be7008eaeec250852bfbc4eb5433d6a).

packages/genai-util/src/handler.ts
psx95: I could remove these extraneous checks in that case. Could you explain the reference to diag logging info here. The tagged lines do not provide enough context on this.

cc: @aabmass

packages/genai-util/src/handler.ts
psx95: @aabmass based on our offline discussions, I am thinking of removing this class completely and basically add the completion_hook as as property of specific invocation subclasses - like `InferenceInvocation`'s constructor would have `completion_hook` as a parameter.

packages/genai-util/src/invocations/base.ts
psx95: They can mostly use `withContext` function. This was to cover any use-cases that cannot be wrapped inside the callback function of `withContext`. 

However, it does break span encapsulation indirectly, so I removed it in [4f4e0f0](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/4f4e0f0fe272ef8e24602a17088278f68604cecc).

packages/genai-util/src/invocations/base.ts
aabmass: Apologies if it's already obvious, but part of the problem is you can throw literally anything in JS including string, number, etc. (I doubt any GenAI libraries are doing so though)

But the instrumentations should have a better idea what to expect rather than the util. It might be good to start strict here and later see if there is a lot of redundant boilerplate we could absorb into the function?

packages/genai-util/src/handler.ts
aabmass: https://github.com/open-telemetry/opentelemetry-js/blob/65c617068cbe3ef59f8a0653766a5f36364b5fc7/packages/sdk-metrics/src/Instruments.ts#L44-L63

packages/genai-util/src/handler.ts
aabmass: SG and I don't think we need to support lazy add/remove hooks or anything like that. If it's helpful I can send a PR to update the type and docs for `CompletionHook` we already have to make it more clear?

packages/genai-util/src/invocations/base.ts
lmolkova: we don't have to record exceptions as span events, I suggest to remove it since we know it'd need to go at some not-so-distant point in the future.

Exception stack trace will likely be caught and logged by the application anyway or handled gracefully. We don't record them in python and AFAIK nobody ever asked.

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  /**
+   * Set multiple custom span attributes.
+   *
+   * Span attributes are per-invocation and may be high cardinality. They are NOT
+   * recorded on metrics: use {@link setMetricAttributes} for that.
+   */
+  public setAttributes(attributes: Attributes): this {
+    Object.assign(this._customAttributes, attributes);
```
lmolkova: it seems `_customAttributes` are unused, are they necessary?

packages/genai-util/src/invocations/base.ts
lmolkova: thanks, looks good!

packages/genai-util/src/invocations/base.ts
lmolkova: > Apologies if it's already obvious, but part of the problem is you can throw literally anything in JS including string, number, etc

thanks! 🫠 


can we be strict? can the underlying lib throw a string? if it can, then it's util which should be making sense out of it rather than instrumentations. 

We know we have two general cases:

- exceptions | errors
- error codes and descriptions (e.g. https://developers.openai.com/api/reference/resources/responses#(resource)%20responses%20%3E%20(model)%20response_error%20%3E%20(schema)) 

Maybe the middle ground is something like 

```ts
export class GenAIError {
  readonly type: string;
  readonly message?: string;
  readonly exception?: Error;
  constructor(options: { type: string; message?: string; exception?: Error }) {
    this.type = options.type || '_OTHER';
    this.message = options.message;
    this.exception = options.exception;
  }
  static from(err: unknown): GenAIError {
    if (err instanceof GenAIError) {
      return err;
    }
    if (err instanceof Error) {
      return new GenAIError({
        type: getErrorType(err),
        message: err.message,
        exception: err,
      });
    }
    return new GenAIError({
      type: '_OTHER',
      message: err != null ? String(err) : undefined,
    });
  }
}
public fail(error: GenAIError, endTime?: TimeInput): void

// I don't know JS / TS API design - I have no idea if it's a good API
```


then instrumentation just needs to call this method and all the boilerplate handled centrally in utils. we also keep both the attributes and exception for the future when we start recording exceptions as logs.

packages/genai-util/src/invocations/base.ts
psx95: The completion hook manager has been removed in [3eab591](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/3eab591e1fc4a82a6266a9a9f95318a46d04ee2a) to make this PR simple.

packages/genai-util/src/handler.ts
psx95: Do you mean a PR to the semantic conventions repo or the python genai utils? I think either would be helpful.

FYI: [3eab591](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/3eab591e1fc4a82a6266a9a9f95318a46d04ee2a) removes the current implementation of CompletionHookManager. I will add it in the next PR on a per invocation basis as discussed.

packages/genai-util/src/invocations/base.ts
psx95: [3eab591](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/3eab591e1fc4a82a6266a9a9f95318a46d04ee2a) rendered this moot.

packages/genai-util/src/handler.ts
psx95: I just realized the `TokenUsage` type contains all fields as optional: 

```typescript
 inputTokens?: number;
 outputTokens?: number;
 reasoningTokens?: number;
 ...
```

So we'd have to put this check here anyways otherwise we get: 

```
Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
```

Took a pass at removing redundant checks in [3776f00](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/3776f00ee81030b7b846bb17769b78b6a8a8d338).

packages/genai-util/src/invocations/base.ts
psx95: I see, I removed it in [24122dd](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/24122dd96f7c8dc4f16606a80a1e7889f376b961).

packages/genai-util/src/invocations/base.ts
psx95: Yes, it should be singular as per semconv. Addressed in [24122dd](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/24122dd96f7c8dc4f16606a80a1e7889f376b961).

packages/genai-util/src/invocations/base.ts
psx95: No, not after all the refactorings done in this PR. Good catch - I removed it in [286ec44](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/286ec441b8874a6f8178e3d7468c0a6d455fbf73).

packages/genai-util/src/invocations/base.ts
psx95: IIUC, the cardinality concerns are only for the error's `type`. In the current implementation if what ever is passed is of type `Error`, then the implementation is already similar to what's shared above and so there's no issue. If the passed object is anything other than `Error` type is set as `_OTHER`. 

So for instance if the type of error being passed is a `string`, then the passed string becomes the error message with the error type `_OTHER` - this is the desired behavior already, right? 

Or am I missing something here?

packages/genai-util/src/handler.ts
aabmass: No I meant to this repo to update the CompletionHook's type to be in line with python

packages/genai-util/src/invocations/base.ts
aabmass: I'll file an issue and we can resolve this in a later PR.

I agree the current has the the same result as what Liudmila proposed so it seems fine.. My understanding was that the we [shouldn't record arbitrary strings in `SpanStatus.description`](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/general/recording-errors.md#recording-errors-on-spans) which is getting populated from `message` in the code snippet

packages/genai-util/src/handler.ts
psx95: I see,  I can tackle this in the next PR then - I removed it for now so I think its ok. Feel free to reopen in case there's anything else.

packages/genai-util/src/invocations/base.ts
psx95: Sounds good to me, in that case - I will mark this conversation as resolved and I can tackle the issue you file in the subsequent PRs. Feel free to link the issue here for posterity.

aabmass: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3769

packages/genai-util/src/handler.ts
```diff
@@ -0,0 +1,271 @@
...
+   * @param attributes - Metric attributes.
+   * @param context - Context used to associate an exemplar with the
+   *   measurement. Defaults to the currently active context.
+   */
+  public recordTimeToFirstChunk(
+    durationSeconds: number,
+    attributes?: Attributes,
+    context?: Context
+  ): void {
+    if (durationSeconds < 0) {
```
aabmass: Can we remove all the negative checks? The SDK has logging for negative measurements in Histogram and Counter?

https://github.com/open-telemetry/opentelemetry-js/blob/bf28c1107e6ed1331d862a64a08b62e7a2874dad/packages/sdk-metrics/src/Instruments.ts#L90-L94

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,326 @@
...
+      error instanceof Error
+        ? error.message
+        : typeof error === 'string'
+          ? error
+          : String(error);
+
+    // Resolved once here so that the span and the metrics always agree on `error.type`,
+    // and so that subclasses do not each have to derive it from the raw error.
+    const errorType = getErrorType(error);
+    this._metricAttributes[ATTR_ERROR_TYPE] = errorType;
```
aabmass: Please remove `this._metricAttributes[ATTR_ERROR_TYPE] = errorType` here and instead pass `errorType?: string` directly to `_recordMetrics(durationSec: number, errorType?: string)`.

Semconv only defines `error.type` on `gen_ai.client.operation.duration`. Putting it into the shared metric attributes object means subclasses will also attach `error.type` to `gen_ai.client.token.usage` (e.g. when a failed stream records partial usage), where it isn't a valid attribute. Subclasses should only attach `error.type` when recording `gen_ai.client.operation.duration`.

packages/genai-util/src/handler.ts
```diff
@@ -0,0 +1,271 @@
...
+    if (usage.outputTokens !== undefined && usage.outputTokens >= 0) {
+      this._tokenUsageHistogram.record(
+        usage.outputTokens,
+        {
+          ...attributes,
+          [ATTR_GEN_AI_TOKEN_TYPE]: GEN_AI_TOKEN_TYPE_VALUE_OUTPUT,
+        },
+        context
+      );
+    }
```
aabmass: Related to the negative check comment: please also make sure zero values aren't recorded by default.

1. In the `TokenUsage` JSDoc, state explicitly that `inputTokens` / `outputTokens` must be left `undefined` when the provider didn't return usage (e.g. failed requests), rather than defaulting to `0`. The `gen_ai.client.token.usage` histogram buckets start at `1`, so recording `0` on failures skews the distribution.
2. For the optional breakdown fields in #3762 (`reasoningTokens`, `cacheReadTokens`, `cacheCreationTokens`), change the check in `setUsage` from `!== undefined` to `> 0`. Semconv marks these "When applicable", and the reference scenarios omit them when the count is `0`.

aabmass (approved):

packages/genai-util/src/handler.ts
psx95: 1. Yup, sounds good. 
2. setUsage is not being added in this PR, will fix it in the future PRs.

Addressed in [fc663cb](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/fc663cb42d7e1d7d5cbe62799a18b7bc3de22956).

packages/genai-util/src/handler.ts
psx95: Addressed in [fc663cb](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/fc663cb42d7e1d7d5cbe62799a18b7bc3de22956).

packages/genai-util/src/invocations/base.ts
```diff
@@ -0,0 +1,330 @@
...
+  public stop(endTime?: TimeInput): void {
+    if (this._isEnded) {
+      return;
+    }
+    this._isEnded = true;
+
+    const endHr = endTime != null ? timeInputToHrTime(endTime) : hrTime();
+    const durationSec = hrTimeToSeconds(hrTimeDuration(this._startTime, endHr));
+
+    this._recordMetrics(durationSec);
```
JacksonWeber: Can we make sure the span still ends if one of these hooks throws? We've already set `_isEnded = true`, so we'd skip `span.end()` and any later attempt to finish the invocation would do nothing.

psx95: Hmm, I guess wrapping in a try-finally block could prevent the span from being leaked, but this seems like a very defensive code style to me. Is this idiomatic in JS/TS ?

Technically, the hooks like - `_recordMetrics` &  `_emitContentEvents` would be implemented by instrumentation authors, so they should not be throwing errors from those hooks.

I don't have a strong opinion on this so I can wrap the `span.end` calls in both `fail()` and `stop()`, but I'm not sure if we need to be this defensive. Let me know your thoughts.

packages/genai-util/src/invocations/base.ts
psx95: Good catch, its not defined on token usage metric. However, it seems to be defined on all duration metrics, not just `gen_ai.client.operation.duration`. 

I addressed this in [c330d70](https://github.com/psx95/opentelemetry-js-contrib/pull/2/commits/c330d701c4ca2acf57ae2109de424107b2b12ec0).

packages/genai-util/src/invocations/base.ts
psx95: I took a pass at this in [36fc41d](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3757/commits/36fc41daf02d368e925a1f39853c9df9082a0aa4). I took it as an opportunity to refactor code to reduce some code duplicity and adhere to semconv when it comes to setting the description in `span.setStatus`. 

Feel free to un-resolve in case there's any concerns about the approach here.

cc: @aabmass

JacksonWeber (approved):

PR #3762, opened 2026-09-14
feat(genai-util): add support for core invocations

psx95:

Re: #3681

#### Implemented so far in genai-utils:
The `@opentelemetry/genai-util` package provides generic lifecycle scaffolding — `TelemetryHandler` and the abstract `BaseInvocation` — but nothing that actually knows what a GenAI operation is. 

#### The current problem
There's no concrete invocation for the operations the GenAI semantic conventions define, so every instrumentation (OpenAI, Bedrock, Vertex, ...) still has to hand-assemble the per-operation semantics itself: which `gen_ai.*` attributes apply to a chat vs. an embeddings call, correct span naming (chat {model}, embeddings {model}, execute_tool {name}), span kind, request-option mapping, response model/ID and finish reasons, token usage and time-to-first-chunk for streaming, tool call arguments/results gated on content-capture mode, and matching metric attribute sets for duration and token usage.

#### What this PR adds: 
This PR adds the concrete operation layer on top of that scaffolding — `InferenceInvocation`, `EmbeddingInvocation`, and `ToolInvocation` — plus `startInference()` / `startEmbedding()` / `startTool()` factories on the handler, so instrumentation authors get spec-correct, consistent GenAI spans and metrics out of the box instead of re-deriving the conventions per package.

- **Inference Invocation (`src/invocations/inference.ts`)**: Added `InferenceInvocation` subclass of `BaseInvocation` handling chat and text completion spans, model request/response attributes, finish reasons, token usage metrics, TTFT metrics, and content capture modes.
- **Embedding Invocation (`src/invocations/embedding.ts`)**: Added `EmbeddingInvocation` subclass of `BaseInvocation` handling embedding vector creation spans (`SpanKind.CLIENT`), server coordinates, and input token metrics.
- **Tool Invocation (`src/invocations/tool.ts`)**: Added `ToolInvocation` subclass of `BaseInvocation` handling tool execution spans (`SpanKind.INTERNAL`), tool parameters, safe circular reference serialization, and tool output capture.
- **Handler Factory Methods (`src/handler.ts`)**: Added `startInference()`, `startEmbedding()`, and `startTool()` to `TelemetryHandler`.
- **Span Name Utility (`src/utils.ts`)**: Added `getSpanName()` helper formatting `<operation.name> <request.model>` per GenAI semantic conventions.
- **Unit Tests**: Added test suites covering inference lifecycle (`test/invocations/inference.test.ts`), embedding operations (`test/invocations/embedding.test.ts`), and tool execution edge cases (`test/invocations/tool.test.ts`).

## Important Info
 - This is the **fourth PR** in a series that attempt to breakdown PR https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3677. 
 - Based off: #3757


packages/genai-util/src/invocations/inference.ts
```diff
@@ -0,0 +1,396 @@
...
+      metricAttrs[ATTR_GEN_AI_RESPONSE_MODEL] = this._responseModel;
+    }
+    if (this._serverAddress) {
+      metricAttrs[ATTR_SERVER_ADDRESS] = this._serverAddress;
+    }
+    if (this._serverPort !== undefined) {
+      metricAttrs[ATTR_SERVER_PORT] = this._serverPort;
+    }
+    if (error) {
+      metricAttrs[ATTR_ERROR_TYPE] = getErrorType(error);
```
JacksonWeber: The base class already passes the resolved error-type string. Calling `getErrorType()` again turns it into `_OTHER`, losing the actual error classification. You can use the string directly in both inference and embedding.

packages/genai-util/src/invocations/inference.ts
```diff
@@ -0,0 +1,396 @@
...
+      );
+    }
+    return this;
+  }
+
+  /**
+   * Build the metric attributes shared by all metrics recorded for this invocation.
+   */
+  private _getMetricAttributes(error?: unknown): Attributes {
+    const metricAttrs: Attributes = {
```
JacksonWeber, line 321: Please include `this._metricAttributes` here and in embedding. Without it, attributes added through `setMetricAttribute()` or `setMetricAttributes()` are silently dropped.

packages/genai-util/src/invocations/inference.ts
psx95: Good catch, the overridden `_recordMetrics` should be accepting a resolved `errorType: string` - it's unknown here.

Addressed in [2fb4413](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3762/commits/2fb44132e7da4c49c589a0973460dd87dc82203a).

packages/genai-util/src/invocations/inference.ts
psx95: Good catch, thanks! 👍🏻 

Addressed in [2fb4413](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3762/commits/2fb44132e7da4c49c589a0973460dd87dc82203a).

JacksonWeber (approved):

packages/genai-util/src/invocations/embedding.ts
```diff
@@ -0,0 +1,195 @@
...
+    return this._responseModel;
+  }
+
+  public setDimensionCount(count: number): this {
+    this._dimensionCount = count;
+    this._span.setAttribute(ATTR_GEN_AI_EMBEDDINGS_DIMENSION_COUNT, count);
+    return this;
+  }
+
+  public getDimensionCount(): number | undefined {
```
aabmass, line 133: do we need getters?

wondering if we can just set stuff in attributes and forego the private member variables. also InferenceInvocation doesn't have getters

packages/genai-util/src/invocations/inference.ts
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  if (options.serverAddress) {
+    attrs[ATTR_SERVER_ADDRESS] = options.serverAddress;
+    if (options.serverPort !== undefined) {
+      attrs[ATTR_SERVER_PORT] = options.serverPort;
+    }
+  }
+
+  Object.assign(attrs, getRequestOptionsAttributes(options.requestOptions));
```
aabmass, line 81: nit I noticed throughout it's a bit inconsistent on using spread `...attrs` vs `Object.assign()`

consider just assigning this on L63 as well, unless you expect it to override the attributes set above.

packages/genai-util/src/invocations/inference.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (options.systemInstructions) {
+      const formatted = formatSystemInstructions(options.systemInstructions);
+      if (formatted) {
+        attrs[ATTR_GEN_AI_SYSTEM_INSTRUCTIONS] = formatted;
+      }
+    }
+
+    if (options.inputMessages && options.inputMessages.length > 0) {
+      const formatted = formatInputMessages(options.inputMessages);
+      if (formatted) {
```
aabmass, line 93: nit Do we need this check too? There's also the check on L91 if the inputMessages are empty or not

packages/genai-util/src/handler.ts
```diff
@@ -262,4 +304,29 @@ export class TelemetryHandler {
...
+   *   measurement. Pass the invocation's context explicitly, since the
+   *   measurement is often recorded after the invocation's context is no
+   *   longer active. Defaults to the currently active context.
+   */
+  public recordExecuteToolDuration(
+    durationSeconds: number,
+    attributes?: Attributes,
+    context?: Context
+  ): void {
+    if (durationSeconds < 0 || !isFinite(durationSeconds)) {
```
aabmass, line 323:
```suggestion
    if (durationSeconds < 0) {
```

covered by the SDK

aabmass (approved):

packages/genai-util/src/invocations/embedding.ts
psx95: Hmm,  yes - I can remove the getters. There's no use of them at the moment and they can always be added later.

packages/genai-util/src/invocations/inference.ts
psx95: `formatInputMessages` can technically return `string | undefined` . So this check guards against the case where formatting returns undefined (in case of an error during stringify op).

packages/genai-util/src/invocations/inference.ts
psx95: Addressed in [230cdd1](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3762/commits/230cdd1a269c4fb09515c3d9edcaaf80cfa9f41b).

packages/genai-util/src/handler.ts
psx95: Addressed in [230cdd1](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3762/commits/230cdd1a269c4fb09515c3d9edcaaf80cfa9f41b).

psx95:
Note: This PR should ideally be merged **after https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3772 is merged** since this PR will require a rebase.

PR #3766, opened 2026-09-15
refactor(build): migrate from tsc to tsdown with dual CJS/ESM exports

david-luna:

Packages of this repository are having the same issue as the core repo https://github.com/open-telemetry/opentelemetry-js/issues/3989 which was fixed in https://github.com/open-telemetry/opentelemetry-js/pull/6293

This PR follows the same approach to have dual exports (CJS & ESM formats) working properly.

- Replace `tsc` with `tsdown` 0.22.14 for compilation across all packages, using a shared root `tsdown.config.ts` that per-package configs extend with their `entry` list
- Generate dual CJS/ESM outputs in `dist/` with proper `.cjs`/`.mjs` extensions and `.d.cts`/`.d.mts` type declarations; rewrite `main`/`module`/`types`/`exports` accordingly
- Build with `unbundle: true` so per-module files are preserved: deep imports keep working and `package.json#browser` path swaps stay meaningful; platform barrels are kept as explicit entries so the node/browser indirection survives tree-shaking (dead browser-map entries for inlined barrels removed)
- Use `rimraf` for cross-platform clean scripts
- Align nx with the new layout: `dist` outputs, cache inputs covering tsdown configs, `package.json`, and the lockfile
- Lint the `tsdown.config.ts` build configs through the root `tsconfig` (they sit outside each package's `include` globs)
- Validate every package with `publint` at compile time
- Migrate `watch` scripts to `tsdown --watch --no-clean`
- Compile on Node `24` in CI while tests continue to run on the full supported matrix (build-on-26, run-on-matrix)
- Add a `verify-pack` CI job that packs all 43 publishable packages, loads the `require`/`import` target of every `exports` entry from the extracted tarball, and existence-checks `types`/`browser`/`main`/`module` files

> ⚠️ **Note on the `build/` directory**: published pre-tsdown tarballs ship `build/src` + `build/esnext`, so tooling that compares against released versions (changelog generator) still understands that layout. The local tree only ever produces `dist/`.

## Why the test runners are unchanged

Mocha runs Node tests via ts-node against TypeScript source, so it never touches the build output and is unaffected by the migration. Karma is kept for browser tests: it bundles each package's `src/` directly through Webpack, with a small `NormalModuleReplacementPlugin` in `karma.webpack.js` performing the same node→browser platform swap at the source level that `package.json#browser` performs on `dist/` for real bundlers. Replacing the runners would have been scope creep with no functional benefit.

| Environment | Runner | Relationship to build output |
|---|---|---|
| Node (`npm test`) | Mocha + NYC | Runs TS source via ts-node, independent of `dist/` |
| Browser (`test:browser`) | Karma + Webpack | Bundles `src/` with a source-level platform swap |
| Packed artifact | `verify-pack` CI job | Resolves every published entry from the real tarball |

## Why builds run on Node `24`

tsdown 0.22 dropped support for Node < 22.18, but the unit-test matrix must keep testing every supported runtime (18.19.0 through 24). Those jobs now compile once on Node `24` and then switch to the matrix Node version for tests.

## Why not add Node `26` as it was in the core repo

There are some dependencies that need to be built in a `postinstall` script in order to run tests. Some of them are not compatible with Node 26 and need to be updated. That would be scope creep on this PR. It can be done in a follow up PR

## Test plan

- [x] `npm run compile` succeeds across all packages (73 projects)
- [x] `npm run lint` passes (73 projects)
- [ ] `npm run test` passes on the full Node matrix (18.19.0 → 24)
- [ ] `npm run test:browser`
- [x] `npm run verify-pack` resolves all 71 publishable packages cleanly


packages/id-generator-aws-xray/package.json
```diff
@@ -2,29 +2,59 @@
...
+        "types": "./dist/platform/browser/index.d.cts",
+        "default": "./dist/platform/browser/index.cjs"
+      }
+    }
+  },
   "browser": {
-    "./src/platform/index.ts": "./src/platform/browser/index.ts",
-    "./build/esm/platform/index.js": "./build/esm/platform/browser/index.js",
-    "./build/src/platform/index.js": "./build/src/platform/browser/index.js"
+    "./dist/platform/node/index.mjs": "./dist/platform/browser/index.mjs"
```
dashpole: Because tsdown with `unbundle: true` resolves the barrel re-export in `dist/index.mjs` and `dist/index.cjs` directly to `./platform/node/AWSXRayIdGenerator.mjs` (and `.cjs`), mapping only `./dist/platform/node/index.mjs` causes browser bundlers to bundle the Node implementation (`Buffer.allocUnsafe`) and fail at runtime. We should map the implementation files and both `.mjs`/`.cjs` formats like in `opentelemetry-js` PR #6293:
```suggestion
    "./dist/platform/node/AWSXRayIdGenerator.mjs": "./dist/platform/browser/AWSXRayIdGenerator.mjs",
    "./dist/platform/node/AWSXRayIdGenerator.cjs": "./dist/platform/browser/AWSXRayIdGenerator.cjs",
    "./dist/platform/index.mjs": "./dist/platform/browser/index.mjs",
    "./dist/platform/index.cjs": "./dist/platform/browser/index.cjs"
```

scripts/verify-pack.mjs
```diff
@@ -0,0 +1,232 @@
...
+    // would rebuild). The caller is responsible for running `npm run compile`
+    // first.
+    const tgz = execFileSync(
+      'npm',
+      ['pack', '--ignore-scripts', '--pack-destination', scratch, '--json'],
+      { cwd: dir, encoding: 'utf8' }
+    );
+    const tarball = JSON.parse(tgz)[0].filename;
+    execFileSync('tar', ['xzf', path.join(scratch, tarball), '-C', scratch]);
+    const extracted = path.join(scratch, 'package');
```
dashpole: Extracting into `tmpdir()` means there is no `node_modules` in parent directories, so 65 of 71 packages fail on their very first external import (like `@opentelemetry/api`) and get skipped by `handleLoadError` before the module body or internal imports are evaluated. Symlinking `node_modules` into `scratch` keeps tarball files isolated while letting external deps resolve, which immediately uncovers runtime ESM load crashes in `host-metrics`, `instrumentation-host-metrics`, and `instrumentation-winston`:
```suggestion
    const extracted = path.join(scratch, 'package');
    symlinkSync(path.join(REPO_ROOT, 'node_modules'), path.join(scratch, 'node_modules'), 'dir');
```

packages/auto-instrumentations-node/package.json
```diff
@@ -11,25 +11,44 @@
...
+      "import": {
+        "types": "./dist/index.d.mts",
+        "default": "./dist/index.mjs"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
+    },
+    "./register": {
```
dashpole: When loading `@opentelemetry/auto-instrumentations-node` (or `./register`) via ESM `import()`, Node throws `ERR_MODULE_NOT_FOUND: Cannot find module '.../systeminformation/lib/network'` because `@opentelemetry/instrumentation-host-metrics` (`dist/stats/si.mjs`) and `@opentelemetry/host-metrics` import `systeminformation/lib/network` without the `.js` file extension required by Node's native ESM resolver. Also note that `require('./register')` returns `{}` since `register.ts` is side-effect-only, which triggers the empty-module check in `verify-pack.mjs` once deps resolve.

packages/instrumentation-winston/package.json
```diff
@@ -2,16 +2,29 @@
...
-  "main": "build/src/index.js",
-  "types": "build/src/index.d.ts",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.cts",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.mts",
+        "default": "./dist/index.mjs"
```
dashpole: Because `@opentelemetry/winston-transport` is in `devDependencies` instead of `dependencies`/`peerDependencies`, tsdown bundles it and its transitive `readable-stream` tree into `dist/node_modules/`. Loading `dist/index.mjs` in Node ESM then crashes immediately with `TypeError: require__stream_writable is not a function`. Consider marking `@opentelemetry/winston-transport` as external in `tsdown.config.ts` so it is not bundled into `dist/`.

trentm:
I'm not sure if you/we want to do some of the following in separate PRs. Probably do a "bump min-supported node" in a PR before this one.

- The tests for node 18, 20 are, of course failing. It is super-unlikely that we'll do a release from the contrib repo before we have SDK 3.0 released from the core repo, so I guess we could go ahead *now* and move the contrib repo packages to `engines: { node: ">=22.15.0" }` as in the core repo. And then update CI.
- Perhaps also add Node.js 26 testing to CI.
- We'll have to update the "required" checks in open-telemetry/admin, similar to this one: https://github.com/open-telemetry/admin/pull/905

* * *

Separately:

- test-all-versions (22) is failing: Have you looked into that? Looks like knex and something with its `sqlite3` driver:

```
[5529](https://github.com/open-telemetry/opentelemetry-js-contrib/actions/runs/35599393430/job/107255550210?pr=3766#step:10:25530)

  2) Knex instrumentation
       Instrumenting
         should catch better-sqlite3 errors:

0 !== 1

      + expected - actual

      -0
      +1
      
      at /home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/packages/instrumentation-knex/test/index.test.ts:289:18
      at async Context.<anonymous> (test/index.test.ts:272:7)
```

Guesses: Possibly related to npm not running postinstall scripts by default any more. I'm not at all sure though.

Ah:

```
     Error: The module '/home/runner/work/opentelemetry-js-contrib/opentelemetry-js-contrib/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 137. This version of Node.js requires
NODE_MODULE_VERSION 127. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```

PR #3767, opened 2026-09-17
feat(instrumentation-langchain): migrate donated framework coverage

JacksonWeber:

Contributes the LangChain brownfield migration tracked in open-telemetry/opentelemetry-js-contrib#3668, following the [assignment](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3668#issuecomment-5625045100). It augments the existing package rather than introducing competing instrumentation.

The reference is the [sanctioned donated LangChain instrumentation](https://github.com/open-telemetry/donation-openinference/tree/6cdd644d79fccf50aedcb614187f924ddfcafb7b/js/packages/openinference-instrumentation-langchain) at commit `6cdd644d79fccf50aedcb614187f924ddfcafb7b`.

**[Complete migration report, all 81 donor-test dispositions, and deep-review corrections](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3767#issuecomment-5737151931)**

## Changes

- Instrument composed `RunnableSequence`/`RunnableMap` workflows, including batch and nested streaming transforms; modern agents; structured/dynamic tools; direct/nested LangGraph workflows; and owned `MemoryVectorStore` retrieval, MMR, and retriever/RAG adapters.
- Preserve original SDK Promises, streams, results, errors, callbacks, cancellation and active contexts. Capture content only when explicitly enabled.
- Observe effective agent configuration after SDK defaults are merged. Record descriptions and JSON output type; avoid misleading model attribution when middleware selects or falls back to another model.
- Summarize invocation-local usage, with SDK estimates only when actual usage is absent, and preserve repeated finish reasons in generation order.
- Add deterministic public-SDK/CJS/ESM regressions, the supported-version matrix, upstream conformance, and a reproducible real-Collector command.

### Deep-review follow-up

Commit `c5f19a52` fixes all seven subsequently reproduced defects: authoritative empty agent snapshots now clear removed output; configured streaming stages retain sibling ancestry; inherited callbacks exclude suppressed model runs; explicit recursion gets distinct error spans; SDK prompt values and string message shorthand are normalized; standard server-tool result IDs are preserved; and nested graph namespaces are handled across current and older SDK envelopes.

Regressions reuse existing helpers and test files. Existing conformance/Collector scenarios were strengthened for prompt values, sibling parentage and suppressed usage rather than adding another runner.

## Semantic conventions and intentional scope

Uses [GenAI conventions at `c88d504ab3d9879f8e50d3cc87e69775e11db234`](https://github.com/open-telemetry/semantic-conventions-genai/tree/c88d504ab3d9879f8e50d3cc87e69775e11db234/docs/gen-ai). Workflow/agent/tool spans are `INTERNAL`; retrieval uses the official `CLIENT` convention. Success remains `UNSET`; failures record `error.type`, not exception messages or stacks.

No OpenInference dependencies or runtime incubating-semconv imports are introduced. Provider inference, embeddings and provider-backed retrieval remain provider-owned. Standalone lambdas, prompts, parsers and internal graph adapters are not fabricated into workflows. Legacy pre-1.x SDKs are out of scope.

This is a span migration matching donated signal coverage. Additional **Recommended** framework duration/call-count and owned-retrieval duration metrics are documented as separate features. Retrieval captures only standardized IDs/scores, not arbitrary document bodies/metadata. Graph output is captured only when a complete supported message state is observable, without replaying reducers.

The pinned conformance policy expects `server.address` on retrieval client spans, but in-memory retrieval has no truthful endpoint. Four memory-retrieval scenarios declare only `genai_expected_attribute_missing` with context `{operation: retrieval, missing_attribute: server.address}`. Other findings fail; the expectation also fails when stale. No fabricated addresses or modified upstream policies are used.

No shared public GenAI utility package exists on this checkout. Helpers remain LangChain-specific. Existing owners (`hectorhdzg`, `yiyuan-he`, `mxiamxia`), private/experimental status, release configuration and absence from the auto-instrumentation bundle are unchanged.

## Validation

- Canonical `compile:with-dependencies`: **passed**.
- Full local package suite: **243 passing**, including the seven deep-review regressions.
- Linux Node **24.21.0** `test-all-versions`: **all 20 selected SDK versions passed**, including earliest/latest releases across LangChain, core, classic and LangGraph. Older SDKs without `withConfig` explicitly skip seven applicable tests; their other **236 tests pass**. No structured-output or namespace-output assertion is skipped.
- Strict upstream runner `f8c9b1b55e34e4cea362260626b5a478a7245b09`, Weaver **0.26.1**, pinned registry above: **25 scenarios passed**, only the documented in-memory endpoint finding. No unexpected violations; generated coverage remains in `test/conformance/data.json`.
- Real Collector Contrib **0.161.0**, OTLP/gRPC to Collector JSON: **all seven review reproducers corrected**, verified across **9 cases and 15 received spans**. Committed `npm run test:collector` also passes **25/25 strengthened scenarios** with an official image pinned by digest.
- ESLint, Prettier, README/Markdown lint, production-dependency checks, semantic-convention dependency checks and diff whitespace: **passed**.

All model/tool/embedding responses are deterministic in-process test data. No live model API, credentials or copied provider recording is used. Test containers are stopped after validation.


JacksonWeber:
# LangChain OpenInference donation migration report

**Completed and validated against the donated framework-owned span coverage.**

Fix/completion commit:
[`a5f38c3a`](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3767/commits/a5f38c3a97b746e07c9b8a9044ac5936ef498655).

Deep-review correction:
[`c5f19a52`](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3767/commits/c5f19a52fdfb52eca7145908915badea99dd51d1).
This follow-up fixes all seven subsequently reproduced defects:

| Defect | Correction |
| --- | --- |
| Empty final agent snapshots retained removed responses | Complete values snapshots clear prior output and delta fallbacks; empty incremental updates remain no-ops. |
| Configured streaming siblings acquired incorrect parents | Upstream iterator consumption retains its producer context, separate from the downstream transform. |
| Inherited observers counted suppressed model calls | Suppression is tracked by model run ID and checked at completion without suppressing application callbacks. |
| Same-instance recursion lost error spans | Only explicit one-use SDK method handoffs are deduplicated; recursive public invocations get distinct spans. |
| Valid string prompt values/message shorthand disappeared | Public SDK prompt values and string message entries are normalized without serializing arbitrary state. |
| Standard server-tool results lost correlation IDs | Camel-case `toolCallId` maps to the official ID, retaining the legacy snake-case fallback. |
| Nested graph snapshots were discarded | Owning namespaces are distinguished from descendants, including older `[""]` root envelopes and trailing checkpoint metadata. |

All seven original reproductions now pass positive assertions against a real
Collector: **9 cases, 15 received spans**, including invoke/stream controls.
The final version matrix caught and verified the older LangGraph envelope
variation; no output assertions were weakened to make it pass.

This report accounts for every donated test and distinguishes intentional
OpenTelemetry semantic-model conversions from retained framework coverage.
It is maintained outside the source tree for publication as a PR comment.

## 1. Provenance, assignment, and migration mode

- Tracking issue: [open-telemetry/opentelemetry-js-contrib#3668][tracking].
- Assignment: [explicit approval for the LangChain brownfield migration][assignment].
- Pull request: [open-telemetry/opentelemetry-js-contrib#3767][pr].
- Source package: `@arizeai/openinference-instrumentation-langchain`, version
  `4.0.12`, exclusively from [the sanctioned donation package][donor].
- Exact source commit: `6cdd644d79fccf50aedcb614187f924ddfcafb7b`.
- No pre-donation Arize repository or OpenInference npm tarball was used as a
  migration reference.
- Target: `@opentelemetry/instrumentation-langchain`, existing package
  `packages/instrumentation-langchain`, version `0.10.0`.
- Mode: **augment**, not greenfield. Before the PR, the target contained
  configuration normalization and module-definition scaffolding, but no working
  patches or metrics. The baseline inspected was `20bbc8c8`.
- The initial audited PR revision was
  `3b63d40d9fcc5af23fafa35f5b6aefa4fccba194`. Subsequent fixes and framework
  coverage additions are not covered by that revision's passing test totals.

The original request is to reconcile donated coverage with the existing package,
retain relevant SDK scenarios, use official GenAI conventions, remove
OpenInference dependencies, and include conformance tests. It is not a request to
reproduce every OpenInference attribute or create competing provider spans.

## 2. Existing, native, and shared-infrastructure investigation

### Existing target

The baseline target already owned its package metadata, exports,
`captureMessageContent` configuration, release configuration, and component-owner
entry. Those capabilities remain the starting point. The original
`InstrumentationNodeModuleDefinition` contained TODO patch/unpatch callbacks, so
adding operation coverage does not replace a pre-existing implementation.

### Native SDK tracing

The initial executable audit used `langchain@1.2.29` and
`@langchain/core@1.1.30`. Final local validation uses langchain1.2.29,
core1.2.10, classic1.0.47 and langgraph1.4.14. The complete Linux matrix below
also exercises the earliest and latest published supported 1.x releases.

Investigation performed:

1. Inspected the SDK's callback manager, runnable operation/stream paths, and
   `ReactAgent` invocation/configuration paths. These expose LangChain callbacks
   and LangSmith integration rather than an enabled-by-default OTel provider.
2. Searched the installed `langchain` and `@langchain/core` JS/CJS implementation
   files for `@opentelemetry`, `OTEL`, and `gen_ai.`. No matches were found in
   those implementation directories in the inspected installation.
3. Read the public callback and agent API declarations. The supported SDK has
   callback extension points, public invoke/stream boundaries, model middleware,
   and an optional agent `description`; these are real SDK APIs, not invented
   instrumentation interfaces. The version-specific [ReactAgent source][sdk-agent]
   provides a reproducible upstream reference.
4. The implementation owner ran real-provider/collector disabled-instrumentation
   controls and reported no target operation spans. The existing public-SDK test
   `does not emit native OTel spans when disabled` also checks workflow/tool
   behavior with the OTel test provider.

Conclusion: the examined SDK paths do not supply native conformant GenAI
instrumentation that can replace this package. LangSmith callback support, or
an OTel dependency in a transitive package, is not evidence that these public
SDK operations natively emit the required OTel spans. This conclusion is scoped
to the versions and paths inspected; it is not a claim that every possible
LangSmith deployment/export configuration was evaluated. Recheck this decision
when the supported SDK adds native tracing.

The donation's callback-only tracer is not carried over wholesale. Public
boundaries establish active OTel context; callbacks provide agent usage
summaries without mutating application handler arrays/managers. Stream
observation preserves SDK objects, returned promises, and native cancellation
behavior.

### Shared GenAI utilities

No shared `genai` utility package exists in the inspected `packages/` tree.
Consequently, this PR keeps LangChain-specific extraction and accumulation
helpers local rather than introducing a speculative reusable framework or
importing another package's private implementation. This is an explicit
availability constraint against the tracking issue's shared-utility goal, not a
claim that a shared utility is already used.

### Ownership, release, and automatic registration

- Existing owners remain `hectorhdzg`, `yiyuan-he`, and `mxiamxia`; no new owner
  consent is implied.
- The package remains `private: true` with its existing release manifest entry
  and `skip-github-release` configuration.
- It is intentionally not added to `getNodeAutoInstrumentations()`. Explicit
  `registerInstrumentations`/SDK registration is tested. Preserving the existing
  private-package status is not a missing greenfield-registration step.
- npm workspaces, Nx, the repository's TypeScript/Mocha tooling, and
  release-please remain authoritative. No pnpm workspace, donation changeset,
  or copied generated build output is needed.
- The source SDK remains a development/type dependency; runtime hooks receive
  the SDK module instances from the instrumentation loader.
- Explicit development dependencies and five-version matrix entries cover both
  the owned-retrieval and graph surfaces. Runtime SDK imports remain type-only.

## 3. Donated inventory and operation disposition

The donor patches `@langchain/core/dist/callbacks/manager.cjs` and injects a
`LangChainTracer` through `_configureSync`. It also offers manual callback-manager
instrumentation and custom tracer-provider plumbing. Its tracer observes
LangChain run callbacks, assigning OpenInference run-type categories and
flattening input/output, messages, invocation parameters, tools, documents,
token usage, metadata, and session identifiers.

The donor advertises core `^1.0.0 || ^0.3.0 || ^0.2.0`; its README says only
1.x is tested and directs full legacy support to the separate `langchain-v0`
package. This migration targets the existing target's 1.x support rather than
silently importing the separately assigned legacy package.

| Surface | Donor / baseline | Target decision and verification status |
| --- | --- | --- |
| Composed workflows | Callback chain runs; no implemented baseline target | `RunnableSequence` and `RunnableMap` invoke/stream, optimized sequence batch, and nested streaming transform paths. Regression tests, strict conformance and real-Collector assertions pass. |
| Modern agents | Donor example uses `createAgent`; callback tracing sees resulting runs | `createAgent`/`ReactAgent` invoke/stream, active context, effective conversation defaults, description, system instructions, output-format attribution, and dynamic/fallback-model safeguards. Regression tests and collector assertions pass. |
| Tools | Public `tool()` and callback tool runs | Structured/dynamic tools, invoke and legacy call, call IDs, opt-in arguments/results/description, errors, parentage. |
| Library-owned in-memory retrieval | Donor's retriever integration uses `MemoryVectorStore.asRetriever()` in a retrieval chain | **Implemented:** `@langchain/classic` MemoryVectorStore public `similaritySearch`, `similaritySearchWithScore`, `similaritySearchVectorWithScore`, and `maxMarginalRelevanceSearch`. `asRetriever()` and RAG chains delegate to these owned-class boundaries. One retrieval span records query/top-k and opt-in document IDs/scores; generic/provider VectorStore classes are not patched. Unit, CJS/ESM, conformance and real-Collector checks pass. |
| Direct application graphs | Donor callback mechanism can observe graph runs; no dedicated donor graph test | **Implemented:** Pregel public invoke/stream as configured graph workflows, nested application graphs, errors/cancellation, final-state handling, and suppression. The agent's delegated graph call is observed for effective configuration without a duplicate workflow span. Internal node adapters remain excluded. Unit, CJS/ESM, conformance and real-Collector checks pass. |
| Provider inference and embeddings | Donor traces model/embedding callback runs and provider attributes | Deliberately not re-emitted by a framework instrumentation. Underlying instrumentable provider SDK owns inference/embedding spans and client metrics. Agent-level invocation-local usage is separately summarized. |
| Provider-backed retrieval | Donor generic retriever callbacks can see these | Provider/vector-store instrumentation owns those operations. New tests explicitly require no local-memory instrumentation on a different provider-store implementation. |
| Prompt templates / parsers / standalone lambdas | Donor can produce prompt/chain-category spans and OI template fields | Not classified as workflows on their own. Actual composed workflows retain telemetry; OI template text/variable attributes are not renamed into fabricated GenAI attributes. |
| Usage and finish reasons | Message usage metadata, `tokenUsage`, `estimatedTokenUsage`, `usage`, total/detail fields | Input/output usage is summarized from this invocation's callbacks; no checkpoint-history summation. Actual usage objects take precedence over explicit SDK estimates. Only nonnegative integral counts are recorded, and repeated finish reasons retain generation order. Focused alternative-format, malformed, missing and precedence tests pass. |
| Arbitrary metadata/tags/OI context | Donor serializes these and custom OI context attributes | Not copied. Only recognized application conversation identifiers map to official `gen_ai.conversation.id`. |
| OI masking/provider configuration | `TraceConfig`, `OITracer`, custom constructor provider | Replaced with standard OTel instrumentation/provider configuration and opt-in content capture, not preserved as a compatibility API. |

### Official retrieval and graph semantics

The pinned [retrieval convention][retrieval-spec] defines `retrieval` as a
`CLIENT` span for a service or framework retrieval request. The target uses
that official kind even for framework-owned memory retrieval; it does not
invent `gen_ai.retrieval.internal`. Data-source identity and provider identity
are conditional and must not be fabricated for an anonymous memory store.
Top-k, textual query, and retrieved documents use the official attributes.
The convention's name template includes a data-source ID; for an absent ID,
the implementation uses the bare operation name `retrieval`. The expanded
strict upstream run did not report a name, kind, document-ID/score, query, or
top-k violation. Its only finding for the four owned-memory retrieval scenarios
was `genai_expected_attribute_missing` for `server.address`. An in-process
MemoryVectorStore has no network endpoint; fabricating `localhost`, a URL,
or an IP address would make the telemetry incorrect. Section 6 records the
precisely scoped expected violation and actionable upstream-policy gap.

The pinned [workflow convention][agent-spec] explicitly identifies graphs as
workflow examples and says application-defined nested workflows should be
retained while internal implementation workflows should not. Direct graphs
therefore use `invoke_workflow` / `INTERNAL`, not a new operation name. Complete
root-state snapshots from `values` streams can be captured when they contain
actual SDK messages or supported `input`/`output` message shapes. Arbitrary
state is not serialized as fabricated chat messages. `updates`, `debug`, and
`custom` modes do not expose the full reduced final state, so output content
is omitted for those modes rather than replaying application reducers or
claiming that a delta is a complete result.

## 4. Request/response, content, and control-flow coverage

| Area | Shapes / behavior represented |
| --- | --- |
| Inputs and outputs | String messages; real SDK BaseMessages; role/content objects; role/content tuples; explicit `messages`, `input`, and `output` containers. Arbitrary chain state is not fabricated into a message. |
| Roles | System, developer, human/user, ai/assistant, tool, and SDK custom roles. |
| Tool messages | Normalized `tool_calls`, legacy `additional_kwargs.tool_calls`, legacy `function_call`, invalid calls, streamed argument fragments, tool-call IDs and tool responses. |
| Multimodal parts | Text, reasoning/thinking, images including validated inline data URLs, audio/video/file parts, URI/blob/file references, server tool calls/results, and preserved provider-specific refusal/citation/custom parts. |
| Agent streams | Updates, values, messages, combined modes, namespaced subgraph envelopes, interleaved/missing message IDs, and `returnDirect` tool output. Do not re-record input history as the final agent response. |
| Graph streams | Final root-state `values` snapshots capture actual messages or supported SDK `input`/`output` shapes; arbitrary state is not converted into fake messages. Updates/debug/custom, SSE and native transformations omit output when they do not expose a complete logical result. Unit, strict conformance and real-Collector scenarios pass. |
| Tools | String/JSON/scalar/array arguments and results serialized into official object-shaped tool content; SDK ToolMessage content extraction; circular output does not change application return values. |
| Retrieval | MemoryVectorStore text/vector queries, configured/default k, scored/unscored documents, empty results, MMR, filters/errors, callbacks, and workflow/stream parentage. |
| Usage | `message.usage_metadata`, `llmOutput.tokenUsage`, `llmOutput.usage`, then SDK `estimatedTokenUsage` only when no actual usage object exists; deduplicate run IDs, not distinct generations. The instrumentation does not calculate estimates or invent a total-token attribute. |
| Module/API contracts | CJS/ESM hooks, explicit registration, enable/disable idempotence, multiple module copies, original promises/results/errors, concurrent active contexts, and inherited callbacks. |
| Streaming lifecycle | Async iterator and reader consumption, values/preventCancel, cancellation races, native pipeTo/pipeThrough/tee, exact producer outcome association, failed branches with active siblings, and borrowed method receivers. |

Content capture defaults to false; the documented boolean
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` overrides the constructor
option. Content on messages, system instructions, tool arguments/results and
tool descriptions is opt-in. Arbitrary metadata is not copied. Retrieval tests
require query/documents to obey capture configuration and require document text
and customer metadata not to leak through the official ID/score representation.

The official [retrieval document schema][retrieval-schema] names `id` and
`score` and permits extensions. It does not define a standard mapping for
LangChain `pageContent` or arbitrary `metadata`. Omitting those OI-only
document fields is an intentional standardization/privacy decision, not a
schema prohibition: the report does **not** claim the schema forbids extensions.
These omissions are not missing retrieval coverage or expected conformance
violations. The owned retrieval operation, textual query, requested result
count, and standardized document identity/relevance information are retained;
only the unstandardized document payload/metadata mapping is intentionally
excluded. Query text and document ID/score arrays remain opt-in.

Native stream pipelines and SSE omit output content rather than reporting
transformed bytes or a misleading partial prefix. Their lifecycle remains
observable. SDK early cancellation does not necessarily immediately close an
upstream generator; baseline experiments established that this behavior is not
introduced by the instrumentation.

## 5. Complete source-test disposition: all 81 tests

Inventory is exactly **22 tracer + 54 utility + 5 instrumentation utility
tests = 81**. Every original test name appears individually below, including
duplicate names distinguished by source location. Line numbers refer to the
exact donation commit, not a moving branch.

Disposition vocabulary:

- **Rewrite**: preserve the real SDK scenario under the OTel boundary/models;
  a target source mapping is supplied. This is not a post-fix pass claim.
- **Skip**: only the stated OI/internal-representation assertion is dropped for
  the specific reason given; related real behavior is mapped when applicable.
- No donor test is left without either a concrete replacement assertion or a
  narrow ownership/semantic-model reason for intentionally dropping its old
  representation.

Target path aliases are all relative to `packages/instrumentation-langchain/`:

| Alias | Target file |
| --- | --- |
| CONFIG | `test/instrumentation.test.ts` |
| LC | `test/langchain.test.ts` |
| CONTENT | `test/content.test.ts` |
| USAGE | `test/agent-usage.test.ts` |
| STREAM | `test/agent-stream.test.ts` |
| RUNTIME | `test/runtime-regressions.test.ts` |
| LIFECYCLE | `test/stream-lifecycle.test.ts` |
| MODULE | `test/module-loading.test.ts` |
| REGISTER | `test/fixtures/register.cjs` |
| MIGRATION | `test/migration-regressions.test.ts` |
| FRAMEWORK | `test/framework-coverage.test.ts` |
| CONF | `test/conformance/conformance.yaml` and `scenario.cjs` |

### 5.1 [test/tracer.test.ts][donor-tracer] — 22 tests

| ID | Source line and exact name | Disposition and target scenario / narrow reason |
| --- | --- | --- |
| T01 | 152 — `should patch the callback manager module` | **Rewrite:** CONFIG `patches and unpatches every loaded copy of a module`; MODULE `automatically instruments ${extension} entry points`. Public behavior replaces the OI-global patched flag. |
| T02 | 157 — `should trace an llm call` | **Skip provider-span assertion:** a direct ChatOpenAI call belongs to OpenAI instrumentation, not a second framework inference span. LC `does not instrument standalone lambdas, prompts or delegated model calls` tests the ownership boundary. |
| T03 | 185 — `should properly nest spans` | **Rewrite:** LC active/concurrent parentage tests, MIGRATION `preserves nested ${kind} workflow boundaries (streaming=${streaming})`, FRAMEWORK nested-graph test, and real-Collector parent-ID assertions. |
| T04 | 238 — `should add attributes to llm spans` | **Skip OI/provider inference assertion:** provider owns model request parameters/content/usage on inference spans. CONTENT mappings and LC agent/tool tests preserve relevant message shapes on framework-owned spans without copying OI raw input/output/metadata attributes. |
| T05 | 273 — `should add attributes to llm spans when streaming` | **Skip duplicate provider span; rewrite framework streaming:** LC `keeps the SDK stream and completes only after consumption`; RUNTIME supported stream-mode tests; STREAM message/tool-chunk cases. Donor estimated-usage branch is separately tracked as U41, not implicitly covered. |
| T06 | 307 — `should add documents to retriever spans` | **Rewrite:** FRAMEWORK `traces owned ${name} retrieval once (capture=${capture})`, `retains retrieval callbacks, workflow parentage and stream results`, strict conformance and actual Collector checks. Deterministic embeddings replace external OpenAI calls; official document IDs/scores replace OI document text/metadata. |
| T07 | 383 — `should add a prompt template to a span if found ` | **Skip OI prompt-span/template assertions:** no official prompt-template text/variables span mapping is introduced. Prompt composition remains inside a workflow; separately provided system instructions use the official message-parts model. |
| T08 | 417 — `should add function calls to spans` | **Rewrite parsing, skip duplicate provider span:** CONTENT legacy function/tool-call mappings, LC modern SDK messages, and FRAMEWORK `preserves legacy function-call content at a public workflow boundary` cover input/output. Provider inference attributes remain provider-owned. |
| T09 | 490 — `should capture tool json schema in llm spans for bound tools` | **Skip duplicate provider request assertion:** definitions bound to ChatOpenAI belong on the underlying model request, using official `gen_ai.tool.definitions` there. This target does not invent OI `llm.tools.*.tool.json_schema` fields or treat defining a tool as executing it. |
| T10 | 539 — `should add tool information to tool spans` | **Rewrite:** LC `preserves tool results, call IDs and active context without duplicate call spans`, `supports legacy tool.call and default content privacy`; CONF tool content on/off. |
| T11 | 565 — `should capture context attributes and add them to spans` | **Skip:** OI `setAttributes`, `setSession`, arbitrary custom context propagation, and OI `session.id` are not target APIs. OTel active-parent propagation is independently covered in LC/MIGRATION. Recognized SDK conversation configuration maps to the official attribute. |
| T12 | 615 — `should extract session ID from run metadata with session_id` | **Rewrite:** LC `maps ${key} metadata without copying arbitrary metadata` with `session_id`; MIGRATION effective immutable agent conversation defaults. |
| T13 | 631 — `should extract session ID from run metadata with thread_id` | **Rewrite:** LC metadata-alias parameterization with `thread_id`; MIGRATION defaults/invocation override tests. |
| T14 | 647 — `should extract session ID from run metadata with conversation_id` | **Rewrite:** LC metadata-alias parameterization with `conversation_id`. |
| T15 | 663 — `should prioritize session_id over thread_id and conversation_id` | **Rewrite:** FRAMEWORK `preserves donor metadata-alias precedence and ignores unrelated metadata` invokes a real workflow with all three aliases and asserts `session_id` wins. |
| T16 | 681 — `should handle missing session identifiers in metadata` | **Rewrite:** FRAMEWORK metadata-alias test explicitly invokes a workflow with unrelated-only metadata and asserts the conversation attribute is absent. MIGRATION also checks the original unconfigured agent. |
| T17 | 729 — `should patch the callback manager module` | **Rewrite:** same loader/module lifecycle behavior as T01 under normal OTel configuration. The donor's separate TraceConfig construction context is not a new target configuration API. |
| T18 | 734 — `should respect trace config options` | **Rewrite privacy goal; skip OI masking API:** CONFIG environment/config normalization and invalid-value tests; LC content off/on; CONF workflow/tool/agent capture cases. Default omission replaces OI placeholder/masking assertions. |
| T19 | 810 — `should use the provided tracer provider instead of the global one` (constructor provider suite) | **Rewrite provider selection; skip donor constructor shape:** LC `honors an explicitly supplied tracer provider`; REGISTER explicit OTel registration. Standard `setTracerProvider`/registration replaces `OITracer` constructor injection. |
| T20 | 863 — `should use the provided tracer provider instead of the global one` (setter suite) | **Rewrite:** LC `honors an explicitly supplied tracer provider`, with a separate exporter and no global-provider spans. |
| T21 | 915 — `should use the provided tracer provider instead of the global one` (registerInstrumentations suite) | **Rewrite:** REGISTER passes `tracerProvider` to `registerInstrumentations`; MODULE runs real CJS/ESM SDK operations through that provider/exporter. |
| T22 | 936 — `should delete runs after they are ended` | **Skip obsolete run-map internals; rewrite lifecycle outcome:** target has no donor run-ID-to-span map. LC idempotent enable/disable/in-flight completion and LIFECYCLE end-once/error/cancellation cases cover completion without requiring the removed private data structure. |

### 5.2 [test/utils.test.ts][donor-utils] — 54 tests

| ID | Source line and exact name | Disposition and target scenario / narrow reason |
| --- | --- | --- |
| U01 | 49 — `should flatten attributes with nested objects` | **Skip:** OI dotted attribute flattening is removed; official structured message attributes are serialized as JSON rather than inventing dotted keys. CONTENT validates resulting models. |
| U02 | 54 — `should flatten attributes with nested arrays` | **Skip:** same OI flattening removal; array order is retained in official JSON messages, not numeric dotted attribute names. |
| U03 | 68 — `should ignore null and undefined values` | **Rewrite missing-content behavior:** CONTENT `does not fabricate messages from arbitrary chain state` and `produces object-shaped tool content for every supported value`. Do not copy OI flattening's representation-specific behavior. |
| U04 | 79 — `should return OpenInferenceSpanKind.AGENT when runType includes 'AGENT'` | **Skip:** arbitrary substring classification of OI run types has no official mapping. Instrument actual supported agent boundaries as `invoke_agent`. |
| U05 | 85 — `should return the corresponding OpenInferenceSpanKind value when runType is a valid key` | **Skip:** OI span-kind enum removed; operation and OTel kind are asserted at real boundaries in LC/FRAMEWORK/CONF. |
| U06 | 91 — `should be case-insensitive` | **Skip:** pertains only to OI run-type string classification; target hooks concrete SDK types/methods. |
| U07 | 97 — `should return 'UNKNOWN' when runType is not recognized` | **Skip:** no fabricated unknown GenAI operation is emitted. Unsupported/OI-only run types do not become new spans. |
| U08 | 113 — `should return an empty object when output is undefined` | **Rewrite:** CONTENT `does not fabricate messages from arbitrary chain state`; `toolContent(undefined)` assertion in `produces object-shaped tool content for every supported value`. |
| U09 | 117 — `should return the formatted the stringified IO when given an object with more than one key` | **Skip OI raw IO assertion:** arbitrary object state is not official chat content. CONTENT `does not fabricate messages from arbitrary chain state`; tool object/scalar content has its own official representation. |
| U10 | 129 — `should return the io value as TEXT type when there is only one value in the io` | **Rewrite supported content, skip OI MIME fields:** LC `captures workflow content only when enabled` maps text into message parts. CONTENT tool serialization preserves strings/scalars without OI `input.mime_type`/`output.mime_type`. |
| U11 | 171 — `should return null if input messages is an empty array` | **Rewrite:** CONTENT missing/invalid-message test explicitly asserts both `[]` and `{messages: []}` omit content. |
| U12 | 178 — `should return null if the first set of messages is not an array` | **Skip donor callback-envelope requirement; rewrite invalid-message handling:** nested `Run.inputs.messages` arrays are provider callback serialization, not the new public input contract. CONTENT `does not fabricate messages from arbitrary chain state` rejects objects without roles while valid public flat message arrays remain supported. |
| U13 | 185 — `should parse and return the input messages if they are valid` | **Rewrite:** CONTENT `preserves all message roles and tuple inputs`; LC modern SDK messages test and captured workflow/agent input scenarios. |
| U14 | 194 — `should ignore non-object messages and return the valid ones` | **Rewrite:** CONTENT `preserves valid messages among malformed input and output entries` surrounds a real SDK message with invalid entries and asserts its unchanged position/content. |
| U15 | 203 — `should parse add tool call information` | **Rewrite:** CONTENT legacy additional_kwargs calls, normalized calls, duplicate content/tool_calls suppression, and invalid argument preservation; LC modern-message public representations. |
| U16 | 250 — `should parse add function call information` | **Rewrite:** CONTENT legacy `arguments` and `args` cases plus FRAMEWORK `preserves legacy function-call content at a public workflow boundary` assert official tool-call parts on both input and output. |
| U17 | 283 — `should return null if generations is an empty array` | **Rewrite public output omission:** CONTENT empty/undefined message omission and USAGE `omits missing or malformed usage without changing callback behavior` explicitly cover empty generations. Provider `Run.outputs.generations` envelopes are not framework message input. |
| U18 | 290 — `should return null if the first set of generations is not an array` | **Skip provider-serialized envelope assertion; rewrite safety:** this old envelope is not the framework response contract. USAGE malformed callback cases explicitly include invalid generation entries and require harmless omission rather than an OI output map. |
| U19 | 297 — `should parse and return the generations if they are valid` | **Rewrite:** LC captured final agent/workflow output; CONTENT `extracts final agent output without recapturing input history`; STREAM complete SDK-message accumulation. Provider multiple-choice inference remains provider-owned. |
| U20 | 306 — `should ignore non-object messages and return the valid ones` | **Rewrite:** CONTENT mixed-message case tests both input/output default roles and preserves only valid messages. |
| U21 | 332 — `should return null if run_type is not 'retriever'` | **Rewrite boundary ownership, not run-type strings:** FRAMEWORK `does not instrument delegated vector store implementations`; only owned memory-store entry points create the new retrieval telemetry. |
| U22 | 337 — `should return null if outputs is not an object` | **Rewrite:** FRAMEWORK `rejects malformed retrieval output without inventing documents` checks null; common operation extraction errors are caught and diagnosed without replacing SDK results. |
| U23 | 342 — `should return null if outputs.documents is not an array` | **Rewrite representation:** direct SDK retrieval returns a document array, not an OI callback `{documents}` envelope. FRAMEWORK malformed-output case rejects non-array output; no OI envelope parser is retained. |
| U24 | 352 — `should return an object with parsed documents if outputs.documents is an array` | **Rewrite:** FRAMEWORK owned retrieval capture cases parse documents into official ID/score objects. The donor's document text is intentionally not copied: no standardized `pageContent` mapping exists, and no custom content field is invented. |
| U25 | 361 — `should filter out non object documents` | **Rewrite safety, representation intentionally differs:** FRAMEWORK rejects malformed retrieval output rather than emitting a guessed partial document list; actual MemoryVectorStore outputs are SDK Documents. The old generic provider-callback envelope is not an accepted target response shape. Common extraction failures are diagnosed without replacing SDK results. |
| U26 | 373 — `should add document metadata` | **Skip OI document metadata fields intentionally:** official retrieval schema names ID/score, not arbitrary LangChain metadata. FRAMEWORK asserts private document text/metadata are absent. Extensions are technically allowed by the schema, but this package intentionally does not invent a custom mapping. This omission is not an implementation gap, an accepted deferral, or an expected conformance violation. |
| U27 | 399 — `should return null if runExtra is not an object or runExtra.invocation_params is not an object` | **Skip provider/OI callback representation:** target does not read `Run.extra.invocation_params`. No arbitrary invocation-parameter blob is copied. |
| U28 | 405 — `should return swallow errors stringifying invocation params, but still add model_name if possible` | **Skip provider-parameter serialization; rewrite extraction safety:** LC `does not let content serialization failures change application output`; model attribution uses typed SDK options, not stringified invocation parameters. |
| U29 | 420 — `should return the formatted LLMParameterAttributes object` | **Skip duplicate provider parameters/tool-definition map:** underlying model instrumentation owns inference request parameters. MIGRATION static/dynamic model attribution covers the narrower officially supported agent model field. |
| U30 | 434 — `should use 'model' if 'model_name' is not a string` | **Skip OI callback-key fallback:** `model_name`/`model` inside old serialized invocation params are not the target input. MIGRATION `keeps statically known model attribution...` tests the actual SDK model option; dynamic selection must not be mislabeled. |
| U31 | 448 — `should return null if invocation_params is missing` | **Skip removed provider parameter blob:** there is no target invocation_params serializer; optional agent model/content attributes are omitted when unavailable. |
| U32 | 456 — `should return null if run type is not 'prompt'` | **Skip OI-only prompt-kind guard:** prompt callbacks are not separately emitted GenAI operations. LC negative ownership test covers no standalone prompt/model span. |
| U33 | 461 — `should return the run input as prompt template variables for a prompt run` | **Skip OI template-variable attribute:** do not rename arbitrary template variables into invented GenAI fields. Explicit system instructions are separately covered by CONTENT. |
| U34 | 472 — `should not return the template from the runs serialized field if it the path to the template has the wrong type` | **Skip:** private serialized prompt-template path is not read by the target; no official prompt-template field is fabricated. |
| U35 | 495 — `should not return the template from the runs serialized field if it the path to the template has the wrong type` | **Skip:** separate donor malformed-template case with the same name; same removed serialized-template representation as U34. |
| U36 | 517 — `should return null if outputs is undefined` | **Rewrite:** USAGE `omits missing or malformed usage without changing callback behavior` invokes the real SDK handler with undefined output and asserts no attributes or propagated error. |
| U37 | 522 — `should return null if llmOutput or tokenUsage is not an object` | **Rewrite:** same USAGE negative parameterization includes null/non-object tokenUsage and llmOutput. |
| U38 | 530 — `should return the token counts if llmOutput and tokenUsage are valid` | **Rewrite:** USAGE `records before ending a span behind a slow background ${kind} callback` uses `tokenUsage.promptTokens/completionTokens`; `deduplicates model run IDs and sums distinct runs` verifies aggregation. No unofficial total-token attribute is emitted. |
| U39 | 548 — `should return null if llmOutput or tokenUsage are missing` | **Rewrite:** USAGE negative parameterization includes missing llmOutput, empty llmOutput and empty generations, with explicit absent-attribute assertions. |
| U40 | 556 — `should only add numeric token values` | **Rewrite:** USAGE `records only finite nonnegative integral token counts` rejects strings, NaN, infinity, negatives and fractional values while retaining zero and positive integers. |
| U41 | 573 — `should add estimated token counts if actual usage is not present` | **Rewrite:** SDK `estimatedTokenUsage` is the last fallback after actual usage objects. USAGE source-shape parameterization and `prefers actual usage over SDK estimates and accepts missing totals` verify both fallback and precedence. No new nonstandard provenance attribute or locally calculated estimate is emitted; this policy is documented in README. |
| U42 | 591 — `should add token counts from generations message usage metadata if present` | **Rewrite:** LC `parents real agent tool execution without duplicating inference` sums current invocation's SDK `usage_metadata`, ignoring old message history; RUNTIME slow-callback usage cases and CONF agent cases. |
| U43 | 615 — `should add token counts from llmOutput.usage if present adding together input and output tokens if total tokens is not present` | **Rewrite:** USAGE missing-total/actual-usage test verifies input_tokens/output_tokens. Only the OI total-token arithmetic assertion is dropped because no official total-token field is introduced. |
| U44 | 632 — `should add token counts from llmOutput.usage if present using total_tokens if present` | **Rewrite:** USAGE source-shape test supplies contradictory total_tokens=99 and asserts the exact two correct official input/output attributes, with no total attribute. |
| U45 | 653 — `should return null if outputs is empty` | **Rewrite:** CONTENT arbitrary/missing-state output omission; the removed provider `generations` wrapper is not reconstructed. |
| U46 | 658 — `should return null if the first output generation is not an object or does not have a message property` | **Rewrite public invalid-message safety:** CONTENT empty/invalid-state and mixed-valid/invalid-message tests. Provider-generation wrapper parsing itself belongs to provider telemetry. |
| U47 | 677 — `should return null if additional_kwargs or function_call is not an object` | **Rewrite:** CONTENT `supports legacy function args and ignores malformed call envelopes` tests undefined/null/non-object envelopes and preserves valid text without an invented call. |
| U48 | 713 — `should return the formatted function call if it is valid` | **Rewrite:** CONTENT legacy `function_call`/`tool_calls` case maps into official `tool_call` parts; no standalone OI `llm.function_call` JSON attribute is emitted. |
| U49 | 741 — `should return null if the normalized run type is not 'tool'` | **Skip OI callback run-type classification:** concrete StructuredTool boundaries determine execution; workflows/models must not acquire tool attributes. LC default workflow exact-attribute test and provider-ownership negative test cover that distinction. |
| U50 | 747 — `should get tool name from run name if not present in serialized field` | **Rewrite public naming:** LC `supports legacy tool.call and default content privacy` and tool-result test use SDK tool.name directly. No donor serialized.run-name fallback is needed. |
| U51 | 755 — `should return the tool attributes with the serialized name and description if available` | **Rewrite:** LC tool result/call-ID test and CONF tool content on/off use SDK name/description, with description opt-in rather than copied by default. |
| U52 | 771 — `should not return name or description if they are not strings` | **Skip malformed donor serialization only:** this case supplies numeric `run.serialized` fields while a valid run name exists. Target does not read that object; SDK tool name/description are typed string fields. Standard boundary/error safety remains tested without preserving the deleted serializer. |
| U53 | 788 — `should return null if run.extra or run.extra.metadata is not an object` | **Skip arbitrary OI metadata serializer:** arbitrary metadata is never emitted; recognized conversation identifiers have separate type guards and configuration tests. |
| U54 | 798 — `should return the formatted metadata if run.extra.metadata is an object` | **Skip OI metadata attribute:** LC `maps ${key} metadata without copying arbitrary metadata` explicitly verifies the intentional privacy/official-semantics replacement. |

### 5.3 [test/instrumentationUtils.test.ts][donor-instrumentation-utils] — 5 tests

| ID | Source line and exact name | Disposition and target scenario / narrow reason |
| --- | --- | --- |
| I01 | 10 — `should add a tracer if there are no handlers` | **Rewrite:** USAGE `creates a real SDK handler that retains its scheduling when copied`; normal agent invocation receives a usage observer without requiring application callbacks. |
| I02 | 21 — `should add a handler to a pre-existing array of handlers` | **Rewrite without mutation:** USAGE `preserves user method objects and callback identities in a new array`; donor same-array mutation is not preserved. |
| I03 | 33 — `should add a handler to a callback handler class' handlers` | **Rewrite without mutation:** USAGE `preserves parent ID, handler inheritance, tags and metadata on a manager copy`; LC agent tool-loop callback-manager test. |
| I04 | 48 — `should not add a handler if it already exists in an array of handlers` | **Rewrite deduplication:** USAGE span-specific observer names/copies and run-ID deduplication; LC real agent tool-loop asserts exact usage and user callback counts. Donor `instanceof LangChainTracer` identity is removed. |
| I05 | 61 — `should not add a handler if it already exists in a callback handler class' handlers` | **Rewrite:** USAGE copy/name/run-ID cases plus LC real CallbackManager agent invocation with unchanged original handlers. Parent IDs, inheritance and scheduling must survive. |

### 5.4 Donor parity closure

The final targeted parity run passed **59 tests** across usage callbacks, content
mapping and framework coverage. It explicitly closes the previously missing
estimated-usage, alternate/negative usage, empty/mixed messages, legacy-function
args, malformed envelopes and metadata-precedence assertions.

The agent summary intentionally provides aggregate input/output token counts,
not provider inference cache/audio/reasoning breakdowns or an invented total-token
field. Provider instrumentations own those detailed inference attributes.
Malformed old callback serialization is not treated as a new valid SDK response
shape; extraction safety and official output models replace those OI helpers.

No additional donor operation family was identified beyond the owned retrieval,
workflow/graph, modern agent, and tool surfaces implemented here. Provider
inference/embedding/remote retrieval and OI-only helper semantics are
intentionally excluded for the ownership/standardization reasons above, not
postponed work disguised as completion.

## 6. Semantic conventions and conformance

Signal scope is **spans**, as in the donated package. The pinned conventions
also define **Recommended**, not required, workflow/agent/tool duration and
agent-call-count metrics, plus client retrieval duration. These are not emitted
by this span migration and are explicitly documented as additional metric
features; they are not incorrectly attributed to provider instrumentation.
No donor metric implementation or test was dropped. `metrics: []` in the
conformance scenarios documents this scope rather than claiming metric support.

- Registry/convention pin:
  `open-telemetry/semantic-conventions-genai@c88d504ab3d9879f8e50d3cc87e69775e11db234`.
- Upstream runner pin:
  `open-telemetry/semantic-conventions-conformance@f8c9b1b55e34e4cea362260626b5a478a7245b09`.
- Weaver: `0.26.1`.
- Unstable constants are copied locally with provenance. Stable `error.type`
  comes from the normal semantic-conventions package, not the incubating entry.
- Success remains `UNSET`; real SDK errors retain original identity and set
  `ERROR`/`error.type`, without copying error message/stack content.
- All emitted content must satisfy the official JSON models. Unknown
  provider-specific message parts are preserved where the model permits them,
  not silently flattened into OI attribute maps.
- The original 11 scenarios cover workflow invoke/stream, map invoke, tool
  call/invoke/error, agent invoke/stream, and independent content modes.
- **Original-head result only:** the implementation owner reports all 11 passed
  without `--report-only` and `findings: []`.
- **Final expanded strict-run result:** **25 scenarios passed**, including
  structured-output and estimated-usage scenarios, with the four precisely
  expected memory-retrieval endpoint occurrences below and no unexpected
  findings. This was a strict run, not `--report-only`.
- A preceding strict run without the expectations failed only on those four
  retrieval endpoint findings; the non-retrieval additions passed.
- `data.json` is regenerated and includes the deduplicated endpoint-policy
  finding rather than falsely claiming an empty findings list.

### Narrow expected violation: endpoint requirement on in-memory retrieval

| Field | Exact scope |
| --- | --- |
| Policy finding ID | `genai_expected_attribute_missing` |
| Required matching context | `operation: retrieval`, `missing_attribute: server.address` |
| Scenarios | `retrieval_content_off`, `retrieval_content_on`, `retrieval_mmr`, `retrieval_error` |
| Instrumented implementation | Owned `@langchain/classic` MemoryVectorStore; no remote/vector-store provider or network request |
| Reason | The pinned retrieval convention expressly includes service **or framework** retrieval, but the upstream conformance policy requires a server endpoint even for local memory retrieval. There is no truthful `server.address` value. |
| Narrow treatment | An expected violation is declared separately for each listed scenario through the shared `in_memory_retrieval_endpoint` YAML anchor. Other missing attributes, operations, scenarios, and findings remain failures. |
| Not done | No invented localhost/server identity; no upstream policy edits; no disabling validation; no broad ignore of retrieval findings. |
| Verification status | Strict pre-expectation failure was observed. Final strict run: 25 scenarios passed, with precisely these four expected occurrences and no unexpected findings. |
| Actionable upstream resolution | Make the conformance endpoint requirement conditional on an actual remote endpoint, consistently with framework-owned in-process retrieval; add an upstream local-memory retrieval case. No upstream issue/PR has been filed by this reporting task. |

This is a concrete conformance-policy applicability gap, not incomplete
retrieval instrumentation and not a reason to omit the owned operation.
Document payload/metadata omissions are a separate intentional
standardization decision; they do not need an expected violation.

## 7. Privacy and fixture review

The donor tests contain generated-looking provider responses, SDK run fixtures,
sample document strings, placeholder API keys, provider object IDs and response
timestamps. They were reviewed as coverage references, not copied as live
provider recordings. The target scenarios use deterministic in-process fake
models/tools/embeddings and application-generated IDs.

- No live model API was called in the reported collector/conformance exercises.
- No copied donor credential, account identifier, customer document, or provider
  recording is needed.
- Default spans omit prompts, completions, tool input/output, system instructions
  and arbitrary metadata; tests exercise content-on separately.
- Local retrieval cases deliberately include sentinel private document
  text/metadata and require those values to remain absent.
- Exception message/stack content is not recorded.
- Generated in-process model results are not represented as sanitized real
  provider recordings. The skill's follow-up requirement for replacing
  synthetic provider *recordings* is not invoked for these local fake SDK
  implementations.
- Native pipeline omission and final-state limitations are documented rather
  than claimed to contain complete outputs.

## 8. Validation ledger

The historical column applies to original PR head
`3b63d40d9fcc5af23fafa35f5b6aefa4fccba194`. Final results were executed on the
completed changes, including the full supported-version matrix on Linux.

| Check | Original-head evidence | Final post-fix result |
| --- | --- | --- |
| Compile with dependencies | Root `nx run-many -t compile -p @opentelemetry/instrumentation-langchain --verbose` passed. Package-cwd workspace script failed project-graph processing. The script was added by the PR; revised script now invokes Nx from root. | Canonical workspace `compile:with-dependencies` command now **passes** after the cwd fix. |
| Unit/public-SDK/module tests | Full suite: **167 passing**. The PR body's 51-test total was stale. | **243 passing** on Node24.15.0 with langchain1.2.29, core1.2.10, classic1.0.47 and langgraph1.4.14. |
| Upstream conformance | **11 scenarios passed**, Weaver0.26.1, pinned runner/registry, no `--report-only`, findings empty. | **25 scenarios pass strictly** with only the four expected retrieval `server.address` occurrences. Generated `test/conformance/data.json` retains the policy finding. |
| Real Collector Contrib | Official Collector Contrib **0.161.0**, OTLP gRPC receiver and file JSON exporter. Expanded original-head audit: **28 cases, 23 passed, 5 failed, 72 received spans**. | Earlier extended harness: **36/36 cases, 79 spans**. Final seven-finding positive reproductions: **9/9 cases, 15 spans**. The committed `npm run test:collector` command also passes **25/25 strengthened scenarios**. Image digest is pinned; containers are stopped afterward. |
| Confirmed original-head failures | Scoped suppressed stream emitted tool/application children; nested streaming workflow lost inner span; configured agent thread ID absent; available agent description absent; dynamically selected model B attributed to initial A. | All five original reproducers now pass in the real-collector run. Keep regression tests and revalidate after any relevant follow-on source changes. |
| README lint | Initial local failure was CRLF versus the linter's LF substring. | Passes on the final LF-normalized changed documentation, together with Markdown lint. |
| test-all-versions | Windows registry TLS/spawn restrictions blocked the local matrix. | **All 20 selected versions pass** with the unchanged TAV runner in an isolated Linux Node24.21.0 container. Older LangChain releases without the public `withConfig` API explicitly skip seven applicable API tests; their other **236 tests pass**. Current releases pass **243 tests**. The synthetic structured-output model advertises the SDK's supported testing capability; no structured-output case is skipped. |
| Lint/format/dependency/semconv checks | Earlier PR description claimed success. | Targeted ESLint, Prettier, Markdown, README, root production-dependency checker and semantic-convention dependency checker pass. The Windows root `lint:deps` script contains a POSIX comment; its exact `knip --dependencies --production --tags=-knipignore` command passes directly. |
| Final dependency/fixture/diff review | No OpenInference dependencies or incubating runtime imports. | Repeated source/test/manifest search is empty; SDK imports added to runtime source are type-only; generated lock metadata uses portable public npm URLs and is validated by npm. |

Reproduction commands:

```text
npm run compile:with-dependencies -w @opentelemetry/instrumentation-langchain
npm test -w @opentelemetry/instrumentation-langchain
npm run test-all-versions -w @opentelemetry/instrumentation-langchain
npm run lint:readme -w @opentelemetry/instrumentation-langchain
npm run test:conformance -w @opentelemetry/instrumentation-langchain -- --registry <pinned-registry-model>
npm run test:collector -w @opentelemetry/instrumentation-langchain
```

The Linux matrix used the repository's `.tav.yml` and `test-all-versions` runner
against isolated copies of the current package and test utilities. It did not
alter the host dependency installation or source to make old SDK cases pass.

| SDK | Successful selected versions |
| --- | --- |
| `langchain` | 1.0.1, 1.2.11, 1.2.36, 1.5.9, 1.5.11 |
| `@langchain/core` | 1.0.0, 1.1.14, 1.1.36, 1.2.8, 1.2.10 |
| `@langchain/classic` | 1.0.0, 1.0.14, 1.0.29, 1.0.45, 1.0.47 |
| `@langchain/langgraph` | 1.0.0, 1.1.3, 1.3.3, 1.4.12, 1.4.14 |

## 9. Finalization checklist

- [x] Resolve all donor-test assertions and estimated-token fallback.
- [x] Verify owned memory retrieval and direct graph coverage; neither remains deferred.
- [x] Verify all five original collector-confirmed bugs with the real collector.
- [x] Complete finish-reason/output-format verification and all 20 version-fixture runs.
- [x] Record the four precisely scoped in-memory retrieval `server.address` findings and actionable upstream-policy gap.
- [x] Verify the 25-scenario strict conformance run accepts only the narrow endpoint finding and retains all other checks.
- [x] Record final commands, totals and versions separately from historical evidence.
- [x] Retain explicit native tracing/shared-utility investigation and dependency decisions.
- [x] Keep this report outside source commits for publication with the PR update.

[tracking]: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3668
[assignment]: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3668#issuecomment-5625045100
[pr]: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3767
[donor]: https://github.com/open-telemetry/donation-openinference/tree/6cdd644d79fccf50aedcb614187f924ddfcafb7b/js/packages/openinference-instrumentation-langchain
[donor-tracer]: https://github.com/open-telemetry/donation-openinference/blob/6cdd644d79fccf50aedcb614187f924ddfcafb7b/js/packages/openinference-instrumentation-langchain/test/tracer.test.ts
[donor-utils]: https://github.com/open-telemetry/donation-openinference/blob/6cdd644d79fccf50aedcb614187f924ddfcafb7b/js/packages/openinference-instrumentation-langchain/test/utils.test.ts
[donor-instrumentation-utils]: https://github.com/open-telemetry/donation-openinference/blob/6cdd644d79fccf50aedcb614187f924ddfcafb7b/js/packages/openinference-instrumentation-langchain/test/instrumentationUtils.test.ts
[sdk-agent]: https://github.com/langchain-ai/langchainjs/blob/langchain%401.2.29/libs/langchain/src/agents/ReactAgent.ts
[agent-spec]: https://github.com/open-telemetry/semantic-conventions-genai/blob/c88d504ab3d9879f8e50d3cc87e69775e11db234/docs/gen-ai/gen-ai-agent-spans.md
[retrieval-spec]: https://github.com/open-telemetry/semantic-conventions-genai/blob/c88d504ab3d9879f8e50d3cc87e69775e11db234/docs/gen-ai/gen-ai-spans.md#retrievals
[retrieval-schema]: https://github.com/open-telemetry/semantic-conventions-genai/blob/c88d504ab3d9879f8e50d3cc87e69775e11db234/model/gen-ai/gen-ai-retrieval-documents.json

PR #3770, opened 2026-09-19
chore(examples): remove deprecated exporter-jaeger

aryamohanan:

The `@opentelemetry/exporter-jaeger` package is deprecated and should no longer be used by the examples. This PR migrates the affected examples to the standard OTLP HTTP exporter and removes outdated Jaeger Thrift configuration.

* Replaced `@opentelemetry/exporter-jaeger` with `@opentelemetry/exporter-trace-otlp-http` in the Redis and Koa examples.
* Updated the tracing setup in the Redis and Koa examples to use `OTLPTraceExporter`.

ref https://github.com/open-telemetry/opentelemetry-js/issues/6900


examples/koa/src/tracer.ts
```diff
@@ -19,8 +19,8 @@ const EXPORTER = process.env.EXPORTER || '';
 
 export const setupTracing = (serviceName: string) => {
   let exporter;
-  if (EXPORTER === 'jaeger') {
-    exporter = new JaegerExporter();
+  if (EXPORTER === 'jaeger' || EXPORTER === 'otlp') {
```
david-luna: why keeping the `jaeger` option in this conditional?

examples/express/README.md
```diff
@@ -25,8 +25,6 @@ docker run --rm --name jaeger \
   -p 16686:16686 \
   -p 4317:4317 \
   -p 4318:4318 \
-  -p 14250:14250 \
```
david-luna, line 28: why remove these ports? also if the goal of this PR is to remove jaeger from examples maybe we want to provide a different image to receive the tracing

david-luna (commented):
@aryamohanan thanks for working on this. I think we might want to update the `README.md` files as well :)

examples/express/README.md
aryamohanan: This PR removes the deprecated Jaeger exporter and migrates to standard OTLP. Jaeger v2 is used as a convenient  OTLP receiver and UI, avoiding the need for a separate collector and backend.

14250 and 14268 are legacy Jaeger Thrift ports used by the deprecated @opentelemetry/exporter-jaeger. Since the example now uses standard OTLP on 4317/4318, they are no longer needed.

examples/koa/src/tracer.ts
aryamohanan: ah nice catch, I will update this

trentm (approved):

